PR #1, opened 2017-04-27
wl_pointer is sketched in, requires wl_surface to implement

zog34:
Code compiles, wl_pointer is sketched in. Methods require wl_surface to continue.


SirCmpwn:
Please revert your changes to the readme.

Your functions should have a more fleshed out signature and you can implement them at least somewhat.

Also, we don't need to pass the client around.

PR #2, opened 2017-05-02
DRM backend + Session interface + EGL

ascent12:
This adds the initial support for running on DRM. It also does a lot of the other related stuff such as opening a session with D-Bus, monitoring devices with Udev, and creating a rendering context with GBM/EGL.

~~This adds a hard dependency on systemd, as logind and sd-bus is being used for the session. This might not be something that  some people like, but it can be made optional if/when an alternate way of handling sessions is implemented, like weston and wlc (optionally) uses.~~


SirCmpwn:
Thanks for all the work! This is a big one, so lots of comments.

Simple fix: style. Braces go on the same line for functions and if statements have mandatory braces (I'll be lenient on the latter only in the case of `if (condition) return;`).

Less simple: some of the approach is wrong. The code needs to be restructured a bit. The goal of wlroots isn't just to be a drop-in wlc replacement, but to be much more modular and flexible. It'll be up to the user to piece the modules together. More specifically, the udev and logind code should be moved to separate libraries in the top level (or maybe one `session` library).

Building off of this, the backend should not be setting up resources. Rather than initialize the session, the caller should deal with it and pass `wl_backend_drm_init` the necessary references. It should also not be initializing the event loop. The ideal function signature is probably `wlr_backend_drm_init(wlr_display *display, int gpu_fd)`. Obviously it also shouldn't release the session, udev, etc in `wlr_backend_drm_free`.

One advantage to this approach is that, for example, you could initialize two DRM backends to support multiple GPUs. You could also start a DRM backend and a RDP backend, for example. That kind of flexibility is the main goal of wlroots.

Other feedback inline.

CMakeLists.txt
```diff
@@ -48,9 +48,11 @@ find_package(WaylandProtocols REQUIRED)
 find_package(EGL REQUIRED)
 find_package(GLESv2 REQUIRED)
 find_package(DRM REQUIRED)
+find_package(GBM REQUIRED)
 find_package(LibInput REQUIRED)
-find_package(Udev)
+find_package(Udev REQUIRED)
 find_package(Dbus)
+find_package(Systemd REQUIRED)
```
SirCmpwn: These should be easy to make optional with the approach I described.

backend/drm/backend.c
```diff
@@ -0,0 +1,118 @@
...
+		goto error_session;
+	}
+
+	backend->fd = wlr_udev_find_gpu(&backend->udev, &backend->session);
+	if (backend->fd == -1) {
+		wlr_log(L_ERROR, "Failed to open DRM device");
+		goto error_udev;
+	}
+
+	if (!wlr_drm_renderer_init(&backend->renderer, backend, backend->fd)) {
```
SirCmpwn: At some point I would like to detach this somewhat from the backend as well, but for now it's fine.

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+	[DRM_MODE_CONNECTOR_DisplayPort] = "DP",
+	[DRM_MODE_CONNECTOR_HDMIA]       = "HDMI-A",
+	[DRM_MODE_CONNECTOR_HDMIB]       = "HDMI-B",
+	[DRM_MODE_CONNECTOR_TV]          = "TV",
+	[DRM_MODE_CONNECTOR_eDP]         = "eDP",
+	[DRM_MODE_CONNECTOR_VIRTUAL]     = "Virtual",
+	[DRM_MODE_CONNECTOR_DSI]         = "DSI",
+};
+
+static const char *egl_error(void)
```
SirCmpwn: Is EGL applicable to other backends?

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+			wlr_drm_display_free(disp, false);
+			wlr_log(L_INFO, "Sending destruction signal for '%s'", disp->name);
+			wl_signal_emit(&backend->signals.display_rem, disp);
+		}
+
+		drmModeFreeConnector(conn);
+	}
+
+	drmModeFreeResources(res);
+}
```
SirCmpwn: Can you split the mostly-DRM code from the mostly-EGL code into two files?

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+
+			disp->renderer = &backend->renderer;
+			disp->state = DRM_DISP_DISCONNECTED;
+			disp->connector = res->connectors[i];
+			snprintf(disp->name, sizeof disp->name, "%s-%"PRIu32,
+				 conn_name[conn->connector_type],
+				 conn->connector_type_id);
+
+			list_add(backend->displays, disp);
+			wlr_log(L_INFO, "Found display '%s'", disp->name);
```
SirCmpwn: Something I'd like to do here is have the backend register the wl_output globals in the display's registry, and have ownership over those resources.

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+	drmModeAddFB(fd, gbm_bo_get_width(bo), gbm_bo_get_height(bo), 24, 32,
+		     gbm_bo_get_stride(bo), gbm_bo_get_handle(bo).u32, id);
+
+	gbm_bo_set_user_data(bo, id, free_fb);
+
+	return *id;
+}
+
+static bool display_init_renderer(struct wlr_drm_renderer *renderer,
+		struct wlr_drm_display *disp)
```
SirCmpwn: Wait is display here an output, not a wl_display? Let's use the output terminology strictly to avoid confusion.

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+	for (size_t i = 0; i < num_modes; ++i) {
+		if (modes[i].hdisplay == width &&
+		    modes[i].vdisplay == height &&
+		    (!rate || modes[i].vrefresh == rate))
+			return &modes[i];
+	}
+
+	wlr_log(L_ERROR, "Unable to find mode %ux%u@%u", width, height, rate);
+	return NULL;
+}
```
SirCmpwn: Export this function and have the top level call it. If you need to pick one during init, use the preferred mode. Otherwise, make a backend-ambiguous way for the caller to set the output mode.

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+		goto error;
+	}
+
+	disp->num_modes = conn->count_modes;
+	disp->modes = malloc(sizeof *disp->modes * disp->num_modes);
+	if (!disp->modes) {
+		wlr_log(L_ERROR, "Allocation failed: %s", strerror(errno));
+		goto error;
+	}
+	memcpy(disp->modes, conn->modes, sizeof *disp->modes * disp->num_modes);
```
SirCmpwn: style nit: `sizeof(*disp->mode)`

backend/drm/drm.c
```diff
@@ -0,0 +1,597 @@
...
+		    modes[i].vdisplay == height &&
+		    (!rate || modes[i].vrefresh == rate))
+			return &modes[i];
+	}
+
+	wlr_log(L_ERROR, "Unable to find mode %ux%u@%u", width, height, rate);
+	return NULL;
+}
+
+bool wlr_drm_display_modeset(struct wlr_drm_display *disp, const char *str)
```
SirCmpwn: Err, this should probably take a `wlr_output_mode` or whatever the struct is called.

SirCmpwn: This is meant to be the public function I was asking about, right?

backend/drm/session.c
```diff
@@ -0,0 +1,219 @@
...
+	sd_bus_error error = SD_BUS_ERROR_NULL;
+
+	struct stat st;
+	if (stat(path, &st) < 0) {
+		wlr_log(L_ERROR, "Failed to stat '%s'", path);
+		return -1;
+	}
+
+	ret = sd_bus_call_method(session->bus,
+				 "org.freedesktop.login1",
```
SirCmpwn: Style: don't indent with respect to the `(`, just add a single tab on subsequent lines. Also merge parameters until you hit 80 columns before breaking.

backend/drm/session.c
```diff
@@ -0,0 +1,219 @@
...
+		goto error;
+	}
+
+	ret = sd_session_get_seat(session->id, &session->seat);
+	if (ret < 0) {
+		wlr_log(L_ERROR, "Failed to get seat id: %s", strerror(-ret));
+		goto error;
+	}
+
+	// This could be done using asprintf, but I don't want to define _GNU_SOURCE
```
SirCmpwn: Better to leave these comments in the PR review, not the source. :+1: to not defining `_GNU_SOURCE`, though.

backend/drm/udev.c
```diff
@@ -0,0 +1,193 @@
...
+#include "backend/drm/udev.h"
+#include "backend/drm/session.h"
+#include "backend/drm/drm.h"
+#include "common/log.h"
+
+static bool device_is_kms(struct wlr_udev *udev,
+		struct wlr_session *session,
+		struct udev_device *dev,
+		int *fd_out)
+{
```
SirCmpwn: Merge parameters until you hit 80 columns on function defs too

backend/drm/udev.c
```diff
@@ -0,0 +1,193 @@
...
+	return true;
+
+out_res:
+	drmModeFreeResources(res);
+out_fd:
+	wlr_session_release_device(session, fd);
+	return false;
+}
+
+int wlr_udev_find_gpu(struct wlr_udev *udev, struct wlr_session *session)
```
SirCmpwn: Better if this function enumerated GPUs instead of just finding the first one.

ascent12:
@SirCmpwn 
Thanks for the feedback.
I figured that there was a bit too much work going on in the backend, and a lot of it would be more generally useful to other parts of the code. I wasn't really sure how the library was going to be structured, so I just left them in inside the backend, expecting them to be split eventually.
I'll get started separating the various parts into their own sections.

SirCmpwn:
Sounds good, thanks!

backend/drm/drm.c
ascent12: Yeah, EGL can be used to create a renderer for a wl_display.

backend/drm/drm.c
ascent12: I was just copying weston here, which uses a string. However, I think having a string argument would be quite useful if a compositor just wants to hand a user-defined string (possible straight from a config file) and not have to worry about the details themselves.
We could just have multiple interfaces, or a string to wlr_output_mode function which doesn't do any actual modesetting.

backend/drm/drm.c
SirCmpwn: Then let's split some of the code into the backend/. I'm okay with the backends creating an OpenGL context for you, but we should still have them share among themselves where appropriate.

backend/drm/drm.c
SirCmpwn: No thanks. String parsing doesn't belong here. It wouldn't be difficult to take a string and pick it out of the list of modes.

SirCmpwn: s/parsing/handling/

ascent12:
@SirCmpwn 
I think I've split out and cleaned up as much of the code as I could sanely do right now.

Currently, the udev part still lives inside the DRM backend. The things it does is very DRM specific, and I cannot think of an appropriate way to move it out and have it generally be useful. It also still just picks the "primary" GPU instead of letting a user choose (however, 99.9% of the time, it's probably the right choice). Again, I couldn't think of an appropriate way to have any sort of useful information exposed in some API. Perhaps instead of handling udev ourselves, we could just let the user do it, and they just hand in an fd and/or a path.

Some parts of the DRM renderer _may_ be able to factored out into a more general renderer. However, to implement this properly, I would need to know a bit more about how the other backends would need to use the renderer, so it would probably be premature to do that right now.

The session code now has a "direct" interface, which removes the previously hard dependency on systemd. It would probably be very limited and only really considered a fallback when all else fails. The session code isn't complete either, as there are still several responsibilities that it's not handling (proper VT setup, VT switching, DRM master, DBus signals, etc.), but that can come later.

SirCmpwn:
>Perhaps instead of handling udev ourselves, we could just let the user do it, and they just hand in an fd and/or a path.

This

Will give this a new review shortly

SirCmpwn:
>The session code isn't complete either, as there are still several responsibilities that it's not handling (proper VT setup, VT switching, DRM master, DBus signals, etc.), but that can come later.

I think part of wlc's problem is that it was rushed into being a compositor, skipping some of the low level details, then filling them in when there was already a lot of complexity involved. I'd like to get the backends into tip-top shape early on.

ascent12:
> I'd like to get the backends into tip-top shape early on.
Yeah, that would be good idea. It's not something you would put off for a long time, but I figured this PR is massive enough as it is, and I don't really understand all of the details.

SirCmpwn:
Yeah, for sure.

SirCmpwn:
Going to push some changes to your branch

ascent12:
That's fine.

SirCmpwn:
Eventually, at least. I need to put some thought into this.

SirCmpwn:
https://github.com/SirCmpwn/wlroots/pull/2/commits/15b1ce9e6ced22a90ab0df7a6b86c6d35c40ca93

Started moving towards what I'd like to see. Currently working on similarly generalizing the output code in drm.c and its headers.

backend/backend.c
```diff
@@ -0,0 +1,37 @@
+#include <wayland-server.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include "common/log.h"
+#include "backend.h"
+
+struct wlr_backend *wlr_backend_create(const struct wlr_backend_impl *impl,
+		struct wlr_backend_state *state) {
```
SirCmpwn: Note `wlr_backend_autocreate` will eventually be made to automatically create the most appropriate backend. It will handle udev and logind and things like that, and will invoke the most appropriate backend's `wlr_*_backend_create` function. This function, `wlr_backend_create`, is an internal function used by the backends to create the user-facing backend object.

backend/drm/backend.c
```diff
@@ -8,106 +8,101 @@
...
 #include "backend/drm/drm.h"
 #include "backend/drm/udev.h"
 #include "common/log.h"
 
-struct wlr_drm_backend *wlr_drm_backend_init(struct wl_display *display,
-	struct wlr_session *session, struct wl_listener *add, struct wl_listener *rem,
-	struct wl_listener *render) {
+static bool wlr_drm_backend_init(struct wlr_backend_state *state) {
+	wlr_drm_scan_connectors(state);
+	return true;
```
SirCmpwn: Can this fail?

SirCmpwn: Also, I wonder if it's a good idea to move the EGL init stuff into here

backend/drm/backend.c
```diff
@@ -8,106 +8,101 @@
...
 
 error_event:
-	wl_event_source_remove(backend->drm_event);
+	wl_event_source_remove(state->drm_event);
 error_fd:
-	wlr_session_close_file(backend->session, backend->fd);
+	wlr_session_close_file(state->session, state->fd);
 error_udev:
-	wlr_udev_free(&backend->udev);
+	wlr_udev_free(&state->udev);
```
SirCmpwn: udev and session code will be removed from this file shortly

include/wlr/backend.h
```diff
@@ -1,7 +1,30 @@
...
+struct wlr_backend {
+	const struct wlr_backend_impl *impl;
+	struct wlr_backend_state *state;
+
+	struct {
+		struct wl_signal output_add;
+		struct wl_signal output_remove;
+		struct wl_signal output_frame;
+		struct wl_signal keyboard_add;
+		struct wl_signal keyboard_remove;
```
SirCmpwn: Note that I don't expect the drm backend to handle these. There will eventually be a libinput backend that will emit these events, but not output events. Probably.

SirCmpwn:
https://github.com/SirCmpwn/wlroots/pull/2/commits/00931f2f8fa4098c5e70b5ececd6d2dcb8786f90

Getting closer. Still have a few more things to change.

include/wlr/backend.h
```diff
@@ -1,7 +1,29 @@
...
-struct wlr_backend *wlr_backend_init();
-void wlr_backend_free(struct wlr_backend *backend);
+#include <wayland-server.h>
+
+struct wlr_backend_impl;
+struct wlr_backend_state;
+
+struct wlr_backend {
+	const struct wlr_backend_impl *impl;
+	struct wlr_backend_state *state;
```
ascent12: This is a DRM-specific type. Perhaps this should be changed to a void *, or we could have 'wlr_backend' as a member of 'wlr_\*_backend' and use wl_container_of() to get to a backend-specific structure (the same way the session code currently does it).

include/wlr/wayland.h
```diff
@@ -3,47 +3,42 @@
...
 };
 
-struct wlr_wl_output {
-	struct wl_output *wl_output;
+struct wlr_output_impl;
+struct wlr_output_state;
+
+struct wlr_output {
+	const struct wlr_output_impl *impl;
+	struct wlr_output_state *state;
```
ascent12: This is the same as my comment regarding wlr_backend: this is a DRM type.

backend/drm/drm.c
```diff
@@ -0,0 +1,419 @@
...
+			}
+
+			output->state = DRM_OUTPUT_NEEDS_MODESET;
+			wlr_log(L_INFO, "Sending modesetting signal for '%s'", output->name);
+			wl_signal_emit(&state->backend->events.output_add, wlr_output);
+		} else if (output->state == DRM_OUTPUT_CONNECTED &&
+			conn->connection != DRM_MODE_CONNECTED) {
+
+			wlr_log(L_INFO, "'%s' disconnected", output->name);
+			// TODO: Destroy
```
ascent12: This was originally intended to just free some of the resources valid for an connected output (renderer and modes), and leave the rest of the structure in place, so we can reuse it if it gets connected again. If we're going to keep this design, we can leave the rest of the info there (name, id, etc.) and not have to free anything else.

include/wlr/backend.h
SirCmpwn: No - it's not a fully defined type. I figure each backend can define the type itself in a private header, and then we get better type safety. A bit unorthodox but I think I like this approach.

ascent12: That seems unnecessarily misleading, in my opinion.

SirCmpwn: I'll think about it a bit more, then.

backend/drm/drm.c
SirCmpwn: Want to push some commits towards this end?

ascent12: I think the way it currently is (that is, not freeing outputs until the very end) is fine. Connectors and outputs are very static things, It's extremely unlikely that the number of them will change throughout the program, and there will only be a small number of them (4 or so on average, maybe 6-8 on the high end), so its not like its wasting a colossal amount of memory.

We just need to free the renderer and modes because they become useless after a connector is disconnected.

SirCmpwn:
Thanks! Will keep iterating on things in mainline.

PR #3, opened 2017-05-13
DPMS support

ascent12:
Initial support for DPMS. We don't support it per-output, as I don't really see the point of doing that.

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -101,3 +101,11 @@ error_backend:
 	free(backend);
 	return NULL;
 }
+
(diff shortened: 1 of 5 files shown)


nloewen:
I was actually just wishing I could do per output DPMS a couple days ago.

Example use case: Turn off all but one display while watching a movie. In this scenario, only the display containing the video player would have power management inhibited, all others would respect the set timeout.

SirCmpwn:
Reason enough for me, let's support it.

ascent12:
The code for it is already there. It just needs to be made part of the public interface, then.

PR #4, opened 2017-05-13
More accurate refresh rates (mHz)

ascent12:


include/wlr/wayland.h
```diff
@@ -10,8 +10,8 @@ struct wlr_output_mode_state;
 struct wlr_output_mode {
 	struct wlr_output_mode_state *state;
 	uint32_t flags; // enum wl_output_mode
-	int32_t width, height;
-	int32_t refresh; // mHz
+	uint32_t width, height;
+	uint32_t refresh; // mHz
```
SirCmpwn: This is actually taken straight from the wayland protocol, which uses signed integers. Why? I have no clue.

ascent12: Yeah, it makes absolutely no sense for them to be negative, so I changed them. They can be changed back to signed though; it wouldn't hurt anything.

SirCmpwn: Please do. Also address the merge conflicts that were introduced by merging your other PR.

SirCmpwn:
Thanks!

PR #5, opened 2017-05-14
Logind D-Bus signals and VT switching

ascent12:


SirCmpwn:
Cursory glance looks good. Will take a closer look tomorrow.

PR #6, opened 2017-06-03
Udev event device matching

ascent12:
This just allows devices to not need to reconfigure themselves if they get a udev event not intended for them.
This should allow multiple DRM backends to share a wlr_udev.

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <assert.h>
 #include <wayland-server.h>
+#include <xf86drm.h>
 #include <sys/stat.h>
 #include <wlr/session.h>
 #include <wlr/types.h>
@@ -30,7 +31,6 @@ static void wlr_drm_backend_destroy(struct wlr_backend_state *state) {
 	}
 	wlr_drm_renderer_free(&state->renderer);
 	wlr_session_close_file(state->session, state->fd);
-	wlr_session_finish(state->session);
 	wl_event_source_remove(state->drm_event);
 	free(state);
 }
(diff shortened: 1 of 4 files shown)


backend/drm/backend.c
```diff
@@ -73,14 +73,31 @@ static void device_resumed(struct wl_listener *listener, void *data) {
 	}
 }
 
-static void drm_invalidated(struct wl_listener *listener, void *state) {
+static void drm_invalidated(struct wl_listener *listener, void *data) {
 	struct wlr_backend_state *drm = wl_container_of(listener, drm, drm_invalidated);
-	wlr_drm_scan_connectors(drm);
+	dev_t *dev = data;
```
SirCmpwn: Where is dev_t declared?

backend/udev.c
```diff
@@ -143,21 +138,16 @@ static int udev_event(int fd, uint32_t mask, void *data) {
...
-	}
-
 	if (!action || strcmp(action, "change") != 0) {
 		goto out;
 	}
 
-	// TODO: Specify the GPU that's being invalidated
-	wl_signal_emit(&udev->invalidate_drm, udev);
+	dev_t devnum = udev_device_get_devnum(dev);
+	wl_signal_emit(&udev->invalidate_drm, &devnum);
```
SirCmpwn: We definitely need to provide people with access to the wlr_udev instance here. You should make a new struct for the event that contains both the dev_t and the wlr_udev.

SirCmpwn:
I would prefer an approach where wlr_udev has the discretion over who gets events rather than everyone getting events and then deciding if they need them.

Something like `wlr_udev_get_listener` or `wlr_udev_signal_add(int fd, struct wl_listener *listener)`

backend/drm/backend.c
ascent12: sys/types.h

SirCmpwn: Oh, alright.

SirCmpwn:
Thanks!

PR #7, opened 2017-06-05
Logging improvements

ascent12:
Changes the logging to actually use the logging callback that the library user can set. It defaults to the current log function (what was _wlr_vlog).
The output message for the default log function has changed slightly: it uses the ISO 8601 date format instead of a locale dependant one (but the time zone is still locale dependent).
The filename is now relative to the wlroots directory instead of just the basename. We have multiple "backend.c"s, so this is to distinguish them.

This implementation uses preprocessor string concatenation, so we can't pass a non-string literal as the format string to the logging functions anymore, but I think that's fine. Passing a non-string literal would be a mistake, anyway.

There is also a static inline function inside the header. This is so that the compiler can optimise the offset calculation out, so it doesn't have to be recalculated every time a logging function is called. I've viewed the assembly output with optimisations enabled, and it does work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ endif()
 
 add_definitions(-DWLR_GIT_VERSION=\"${GIT_COMMIT_HASH}\")
 add_definitions(-DWLR_GIT_BRANCH=\"${GIT_BRANCH}\")
+add_definitions(-DWLR_SRC_DIR=\"${CMAKE_SOURCE_DIR}\")
 
 string(TIMESTAMP CURRENT_DATE "%Y-%m-%d" UTC)
 add_definitions(-DWLR_VERSION_DATE=\"${CURRENT_DATE}\")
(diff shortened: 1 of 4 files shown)


SirCmpwn:
I like it, but no thanks to the inline function.

ascent12:
Ok, I changed it to use a sneaky strlen.

SirCmpwn:
Wait, what does the sneaky strlen do? I don't get it.

bb010g:
Are you worried that it won't get inlined without that annotation? C compilers tend to just kinda do what they think is best with inlining. If you want it to be memoized, just actually write memoization code.

ascent12:
It's just moving the pointer past the "wlroots" in \__FILE__.
Since \__FILE__ is an absolute path, and WLR_SRC_DIR is the path the the wlroots dir.
`"/home/user/wlroots/backend/backend.c + strlen("/home/user/wlroots") + 1`

I originally didn't do it that way, because it seemed a little too "tricky".

SirCmpwn:
Ah, I see. Sounds good. Thanks!

PR #8, opened 2017-06-05
Output rotation

ascent12:
I was looking into output rotation, but due to how trivial it is to implement with OpenGL, I don't really think there is a lot of value wlroots could add by "supporting" this directly; the library user can do it themselves.
Maybe there is a flag somewhere that we needs to be set, just so we know that an output is rotated, but until I know that we actually need that, I won't bother adding it.

On that note, I wrote another example program which supports output rotation, and is configurable with its command line options. It also renders something different, so that you can actually tell if the rotation is working.


example/rotation.c
```diff
@@ -0,0 +1,402 @@
...
+}
+
+int main(int argc, char *argv[]) {
+	if (getenv("DISPLAY")) {
+		fprintf(stderr, "Detected that X is running. Run this in its own virtual terminal.\n");
+		return 1;
+	} else if (getenv("WAYLAND_DISPLAY")) {
+		fprintf(stderr, "Detected that Wayland is running. Run this in its own virtual terminal.\n");
+		return 1;
+	}
```
SirCmpwn: Can you remove this? From the other example too.

ascent12: Just the Wayland one, or the xorg one too?

SirCmpwn: Both.

SirCmpwn:
Some of this code needs to be moved into wlr but I'm not sure exactly how much or how it should be done. Will be tinkering with this branch before merging.

SirCmpwn:
TODO:

* [ ] Configure GL so that output coordinates are [<0,0>,<width,height>)
* [ ] Move shaders into separate files and use CMake to embed them in the executable
* [ ] Try to handle transform in the DRM code before it reaches the compositor code

SirCmpwn:
53dd622 looks much better, thanks!

SirCmpwn:
Currently working on a module for matrix operations, will be using it to make quad rendering more cool.

SirCmpwn:
Alright, I've made all of the changes I'd like to make, but I'm still not 100% happy with it (and some cleanup is necessary). I'm going to work on cleanup and stew on this for a bit, then get it merged probably later today.

render/renderer.c
```diff
@@ -0,0 +1,138 @@
...
+"	gl_FragColor = vec4(texture2D(tex, v_texcoord).rgb, 1.0);\n"
+"}\n";
+
+static const GLchar frag_src_RGBA[] =
+"precision mediump float;\n"
+"varying vec2 v_texcoord;\n"
+"uniform sampler2D tex;\n"
+"void main() {\n"
+"	gl_FragColor = texture2D(tex, v_texcoord);\n"
+"}\n";
```
SirCmpwn: TODO: shaders for other pixel formats

SirCmpwn:
The main goals of the rendering subsystem are to:

* Individual wl_surfaces could have different renderers if we wanted them to for whatever reason
* Eliminate or reduce use of GL calls in compositors
* Not necessarily tied to Wayland surfaces, just a simple abstraction over GLES

SirCmpwn:
As I worked on it, I was split between making this follow the same interface pattern in use everywhere else and making this a special thing. Clearly the interface pattern is the right choice, so I will also be refactoring this a bit towards that goal.

SirCmpwn:
Sorry for hijacking your PR with a lot of unrelated work, I just didn't want to merge more sophisticated GL code without making a proper subsystem for it :)

PR #11, opened 2017-06-09
[WIP] libinput backend

SirCmpwn:
TODO:

* [x] Establish backend
* [x] Add new types
* [x] Implement container for multiple backends
* [x] Split events.c into several files
* [ ] Devices
  * [x] Keyboards
    * [x] Events
    * [x] Key maps
    * [x] LEDs
  * [x] Pointers
    * [ ] DRM hardware cursor - implemented but doesn't work, don't know why
  * [x] Touch
    * [ ] ~~Get mapped output from udev~~ not among the udev properties on my hardware
  * [x] Tablet tool
  * [x] Tablet pad
  * [ ] Gesture
  * [ ] Switch
* [x] Device cleanup
* [ ] Demos
  * [x] Keyboard
  * [x] Pointer
  * [x] Touch
  * [x] Tablets
  * [ ] Switch
* [x] Put more thought into the design

I own hardware for everything, so it's just a matter of writing the code.


SirCmpwn:
Point for discussion: libinput devices are often several kinds of devices in one (my mouse, for example, shows up as a pointer and keyboard because the buttons can be bound to keypresses). However, this is not exposed beyond the libinput backend - one libinput device might become two wlr_input_devices rather than one wlr_input_device with keyboard+pointer capabilities.

SirCmpwn:
I feel good about merging this after I hook up tablet pad buttons. We can do the remaining input devices whenever. Maybe @ascent12 has some comments?

SirCmpwn:
Oh, and I want to sort out LEDs and DRM hardware cursors too.

ascent12:
I'm not really a fan of using wlr_backend for this. It seems too conceptually different than what the DRM backend is doing. I think a "wlr_input_backend" or something would be more appropriate.
However, if we're going to try and expose more information about the backend and its internals to the user, I think we need to rethink wlr_backend and how it works anyway. But this isn't really related to this PR itself, and is something we can do later.

In terms of the *actual* libinput related stuff, I don't really have much to say. I don't really know a lot of the details of libinput, but it looks fine.

SirCmpwn:
The single biggest reason backends are doing input is because some of them are going to have to anyway - Wayland, X11, RDP backends are all going to have to provide input devices.

SirCmpwn:
Pushed some initial support for hardware cursors, but for some reason it doesn't work. If anyone wants to take a look at it, run the pointer demo and you should get an extra cat pic showing up that doesn't move with the mouse (the mouse-bound cat is still a software cursor).

ascent12:
>Pushed some initial support for hardware cursors, but for some reason it doesn't work

Most hardware doesn't support particularly large hardware cursors, and I think the cat you have is too big. You should call drmGetCap() with DRM_CAP_CURSOR_WIDTH and DRM_CAP_CURSOR_HEIGHT to get a cursor size. I think 64x64 is always supposed to be safe, though.

SirCmpwn:
I tried 32x32 and got similar results, but I'll double check.

ascent12:
I've been looking slightly more in-depth into the hardware cursor's design, and I really think the multiple buffers aren't necessary. I don't really think tearing is an issue for cursors; they're small and they move around quickly, so I really don't think anyone could notice it. Also, Weston only uses a single buffer.
The current design doesn't make use of the second buffer, anyway.

SirCmpwn:
Weston uses two buffers. I used the Weston implementation heavily as a reference. And we do make use of both buffers.

PR #16, opened 2017-06-19
Make logging ninja friendly

nyorain:
When building with ninja instead of make ```__FILE__``` will be relative
and logging therefore break. This moves stripping the path
from ```__FILE__``` to runtime and handles a relative filepath as
well as a full one.


SirCmpwn:
Thanks!

PR #18, opened 2017-06-21
Fix VT switching with libinput backend

ascent12:
The session interface was also changed to only emit a signal only when a DRM device is paused/resumed. The old per-device interface wasn't really appropriate for libinput, as it acquires dozens of devices, and we can't tell it to close them individually. This brings it more in line with the way Weston handles this.

The logind backend currently makes some assumptions that there is only a single DRM device being used. If/when we get multi-GPU support, this will need to be changed, but until I know what using multiple GPUs and DRM backends would look like, I can't really see how this needs to be designed.

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -54,8 +54,10 @@ static void session_signal(struct wl_listener *listener, void *data) {
 	} else {
 		wlr_log(L_INFO, "DRM fd paused");
 
-		// TODO: Actually pause the renderer or something.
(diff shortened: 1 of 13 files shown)


SirCmpwn:
Thanks!

SirCmpwn:
Regarding multi-GPU support, we can tackle that once we're rendering wayland client surfaces and sharing their resources between GPUs. That way we can deal with all of the hard problem at once.

PR #19, opened 2017-06-22
Basic wayland backend input

nyorain:
This implements pointer/keyboard input support for the wayland backend.
Touch is not supported yet in any way (cannot test it, probably not as important).
Axis support for pointers could also be improved and it does currently not handle the received keymap (there is no way to do it, xkbcommon would have to be handled by the library).
Due to the current pointer event api, pointer movement events are only handled relatively (not sure about this, should it use wlr_event_pointer_motion_absolute and just imagine some fake mm values?).

### DIFF
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -33,15 +33,21 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 	assert(dev && dev->pointer && dev->pointer->state);
 	struct wlr_pointer_state *state = dev->pointer->state;
 
+	double x = wl_fixed_to_double(surface_x);
+	double y = wl_fixed_to_double(surface_y);
+
+	if (x == state->surface_x && y == state->surface_y)
+		return;
+
 	struct wlr_event_pointer_motion wlr_event;
 	wlr_event.time_sec = time / 1000;
 	wlr_event.time_usec = time * 1000;
-	wlr_event.delta_x = wl_fixed_to_double(surface_x) - state->surface_x;
-	wlr_event.delta_y = wl_fixed_to_double(surface_y) - state->surface_y;
+	wlr_event.delta_x = x - state->surface_x;
+	wlr_event.delta_y = y - state->surface_y;
 	wl_signal_emit(&dev->pointer->events.motion, &wlr_event);
 
-	state->surface_x = surface_x;
-	state->surface_y = surface_y;
+	state->surface_x = x;
+	state->surface_y = y;
 }
 
 static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -11,12 +11,109 @@
 #include <wlr/util/log.h>
 #include "backend/wayland.h"
 
-static void wlr_wl_device_destroy(struct wlr_input_device_state *state) {
+static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface, wl_fixed_t surface_x,
+		wl_fixed_t surface_y) {
+
+}
+
+static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface) {
+
+}
+
+static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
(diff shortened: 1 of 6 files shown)


SirCmpwn:
>not sure about this, should it use wlr_event_pointer_motion_absolute and just imagine some fake mm values?)

Yes

SirCmpwn:
Or could use real mm values based on the wl_output's size compared to our surface's size

backend/wayland/wl_seat.c
```diff
@@ -8,34 +8,188 @@
...
+	struct wlr_input_device *dev = data;
+	assert(dev && dev->pointer && dev->pointer->state);
+	struct wlr_pointer_state *state = dev->pointer->state;
+	state->surface_x = wl_fixed_to_double(surface_x);
+	state->surface_y = wl_fixed_to_double(surface_y);
+}
+
+static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface) {
+
```
SirCmpwn: Could destroy the input device and recreate it when it re-enters. Would cause compositors to remove the cursor. Not a blocker but food for thought.

backend/wayland/wl_seat.c
```diff
@@ -8,34 +8,188 @@
...
+		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
+	struct wlr_input_device *dev = data;
+	assert(dev && dev->pointer && dev->pointer->state);
+	struct wlr_pointer_state *state = dev->pointer->state;
+
+	double x = wl_fixed_to_double(surface_x);
+	double y = wl_fixed_to_double(surface_y);
+
+	if (x == state->surface_x && y == state->surface_y)
+		return;
```
SirCmpwn: style nit - always use parenthesis. wlroots uses sway style FYI https://github.com/SirCmpwn/sway/blob/master/CONTRIBUTING.md

backend/wayland/wl_seat.c
```diff
@@ -8,34 +8,188 @@
...
+		uint32_t time, uint32_t axis, wl_fixed_t value) {
+	struct wlr_input_device *dev = data;
+	assert(dev && dev->pointer);
+
+	struct wlr_event_pointer_axis wlr_event;
+	wlr_event.delta = value;
+	wlr_event.orientation = axis;
+	wlr_event.time_sec = time / 1000;
+	wlr_event.time_usec = time * 1000;
+	wlr_event.source = WLR_AXIS_SOURCE_CONTINUOUS; // TODO
```
SirCmpwn: Drop the TODO, we can't really expect this to ever change

SirCmpwn: Wait, we can

SirCmpwn: You could just store the source information in wlr_input_device->pointer->state?

backend/wayland/wl_seat.c
```diff
@@ -8,34 +8,188 @@
...
-	struct wlr_input_device_state *devstate =
-		calloc(1, sizeof(struct wlr_input_device_state));
-	if(!devstate) {
+	struct wlr_input_device_state *devstate;
+	if (!(devstate = calloc(1, sizeof(struct wlr_input_device_state)))) {
 		wlr_log(L_ERROR, "Allocation failed: %s", strerror(errno));
 		return NULL;
 	}
 
 	// TODO: any way to retrieve those information?
```
SirCmpwn: Nope

SirCmpwn:
Looks good. I would at least take care of absolute motion before this is merged.

SirCmpwn:
So far as keymaps are concerned, leave that to me. I'll be doing some refactoring in that department later.

nyorain:
ok thanks for the input, am working on it.

nyorain:
Fixed most of the issues and made the pointer example work with it.
Does not recreate the input device on leaving the window (this would introduce some new complexity of how to manage input devices, it might be better to use some additional 'visible' flag or something).
The mm values are now just the width and height in pixels of the output, trying to compute it somehow
from the remote wayland output information can probably not be done in a reliable manner (we would have to know for sure on which remote output or window is displayed).

The current implementation of motion_absolute in the shared example compositor is far from perfect, it just assumes the x and y mm values are the pixels of the output. Doing this correctly would need to keep track of the current output in the compositors pointer state and then convert from mm values to pixel coordinates but this would probably need some further changes.

SirCmpwn:
Sounds good for the time being. I'll spend some time working on improvements in this domain shortly. Thanks!

PR #22, opened 2017-06-25
Show and move the hardware cursor

crondog:
The size of the cat is 128x128 so pass that exact size down drm. It still
doesn't render correctly as the channels are wrong. Not sure how to fix that
though.


SirCmpwn:
This is intentionally left unfinished because the hardware cursor doesn't work on my system. I'm going to have to implement a fallback software cursor of some sort.

crondog:
https://github.com/SirCmpwn/wlroots/pull/23 is better :)

PR #23, opened 2017-06-26
Software cursor + Hardware cursor improvements

ascent12:
Draw a cursor using wlr_renderer if a backend fails to draw a cursor. It also gives the option for a backend that doesn't want/need to handle cursors in a special way to just let this handle it.

There are also a couple of other changes, mainly to make this easier to implement.
- Changed the wlr_output 'frame' signal to *not* automatically make the EGL/OpenGL context current. The user must now explicitly call the new `wlr_output_make_current` and `wlr_output_swap_buffers` functions. This is mainly so the SW cursor can be drawn in swap buffers, but it also gives the user the option to draw something to an output whenever they want to.
- Added stride argument to `wlr_renderer_attach_pixels`. Also the GL_EXT_unpack_subimage extension is used to deal with the stride, so we don't have no needlessly copy a bunch of stuff ourselves. It currently doesn't check for the extension, but it's probably widely supported enough.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -296,54 +296,60 @@ static void wlr_drm_output_transform(struct wlr_output_state *output,
 	output->wlr_output->transform = transform;
 }
 
-static void wlr_drm_cursor_bo_update(struct wlr_output_state *output,
-		uint32_t width, uint32_t height) {
-	if (output->cursor_width == width && output->cursor_height == height) {
-		return;
-	}
-	wlr_log(L_DEBUG, "Allocating new cursor bos");
-	struct wlr_backend_state *state =
-		wl_container_of(output->renderer, state, renderer);
-	for (size_t i = 0; i < 2; ++i) {
-		output->cursor_bo[i] = gbm_bo_create(state->renderer.gbm,
-				width, height, GBM_FORMAT_ARGB8888,
-				GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE);
-		if (!output->cursor_bo[i]) {
-			wlr_log(L_ERROR, "Failed to create cursor bo");
-			return;
-		}
-		if (!get_fb_for_bo(state->fd, output->cursor_bo[i])) {
-			wlr_log(L_ERROR, "Failed to create cursor fb");
-			return;
-		}
-	}
-}
-
 static bool wlr_drm_output_set_cursor(struct wlr_output_state *output,
 		const uint8_t *buf, int32_t stride, uint32_t width, uint32_t height) {
-	struct wlr_backend_state *state =
-		wl_container_of(output->renderer, state, renderer);
+	struct wlr_backend_state *state = wl_container_of(output->renderer, state, renderer);
 	if (!buf) {
 		drmModeSetCursor(state->fd, output->crtc, 0, 0, 0);
 		return true;
 	}
-	wlr_drm_cursor_bo_update(output, width, height);
+
+	uint64_t bo_width, bo_height;
+	int ret;
+
+	ret = drmGetCap(state->fd, DRM_CAP_CURSOR_WIDTH, &bo_width);
+	bo_width = ret ? 64 : bo_width;
+	ret = drmGetCap(state->fd, DRM_CAP_CURSOR_HEIGHT, &bo_height);
+	bo_height = ret ? 64 : bo_height;
+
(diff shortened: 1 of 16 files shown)


ascent12:
I also updated the hardware cursor code to make use of drmGetCap, to make sure we get a valid hardware cursor size. Now, if you mess around with the texture width/height in the pointer example, it should display properly (Values less than 64 should be safe for getting a hardware cursor).

backend/drm/drm.c
```diff
@@ -296,54 +296,65 @@ static void wlr_drm_output_transform(struct wlr_output_state *output,
...
+		return false;
+	}
+
+	for (int i = 0; i < 2; ++i) {
+		if (output->cursor_bo[i]) {
+			continue;
+		}
+
+		output->cursor_bo[i] = gbm_bo_create(state->renderer.gbm, bo_width, bo_height,
+			GBM_FORMAT_RGBA8888, GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE);
```
crondog: If GBM_FORMAT_RGBA8888 is used, then on my box it falls back to sw. If it is GBM_FORMAT_ARGB8888 then it correctly uses the hw cursor, but the cat is blue

ascent12: Yeah, the cat is in RBGA format, so it would be reading the colours incorrectly as ARBG. I think only ARGB8888 and XRBG8888 are the only guaranteed formats to be available, so I guess some code to convert the pixel format is needed, or the cat could be changed.

crondog:
Also if i do not use 128 as the image w/h then it only shows the top corner of the image with lines through it (with hw and sw)

ascent12:
@crondog It'll only show the top left corner of the image, because it's not being scaled; it's being cropped. Also, are you messing with the stride? if you do, the image will get messed up. It should still be cat_tex.width.

crondog:
No i didn't mess with the stride. I did this:
```
cat_tex = {
  64, 64, 4, ...
```
So it would use 64 for stride, width and height.

Maybe it is just the image

ascent12:
@crondog Modify the call to wlr_output_set_cursor, not the texture itself. https://github.com/SirCmpwn/wlroots/pull/23/files#diff-9d4506d3a89647f631b477b678644bfcR102
`wlr_output_set_cursor(wlr_output, cat_tex.pixel_data, cat_tex.width, 64, 64)` or something.

crondog:
Yup that worked. I have the cropped sw image, and a cropped hw image (obviously still blue until the format is worked out )

backend/drm/drm.c
crondog: I figured as much. Just wanted to make a note of it :)

SirCmpwn:
Tried implementing this for the wayland backend and it did not work, not sure why. It renders one frame and then the frame_callback is never run again. Mind taking a closer look?

backend/drm/drm.c
SirCmpwn: Let's convert the cat to ARGB8888 and document that as the only supported pixel format for cursors?

SirCmpwn: Ideally we should also have a mini subsystem to grab Xcursors like wayland-cursor does, maybe we can just drop the cat in favor of a proper cursor image

types/wlr_output.c
```diff
@@ -182,16 +186,44 @@ void wlr_output_transform(struct wlr_output *output,
...
+	}
+
+	wlr_log(L_INFO, "Falling back to software cursor");
+
+	output->cursor.is_sw = true;
+	output->cursor.width = width;
+	output->cursor.height = height;
+
+	if (!output->cursor.renderer) {
+		output->cursor.renderer = wlr_gles2_renderer_init();
```
SirCmpwn: Instead of this, let's add `wlr_output_attach_renderer` and let the user provide one. Software cursor would just fail with a warning if there is no renderer attached.

ascent12:
>Mind taking a closer look?

I don't know a huge amount about the wayland protocol, but does `wl_callback_add_listener` and all that need to be called inside 'swap buffers'?

SirCmpwn:
Doesn't seem to have any effect.

SirCmpwn:
Nevermind, that totally works. I was doing it wrong.

SirCmpwn:
Thanks!

Issue #24, opened 2017-07-03
Add detection of no input devices


ascent12:
Currently with the libinput backend, if you cannot open /dev/input/event* for whatever reason (e.g. no permissions), the rest of the compositor will open up, leaving you stuck and unable to close it. It's pretty annoying, as you need to restart your computer using the power button when it happens.

Libinput doesn't report any errors about being unable to open files, so an explicit check needs to happen somewhere to make sure we actually have an input device, and close the compositor if we don't.

PR #25, opened 2017-07-03
Forking session backend

ascent12:
Updated the 'direct' backend to now fork, so that the child (with privileges) can call drmSetMaster/drmDropMaster. It's up to the library user to drop any extra privileges after they initialise wlroots. It also initialises the TTY properly and handles VT switching.

Libcap was added as an optional dependency, so we can have processes with CAP_SYS_ADMIN instead of needing to seteuid as root. If logind is being used, there is no need to do this.

There are a couple of outstanding issues:
- Just like the logind code, it still makes a lot of assumptions about there being a single GPU per session.
- It's very Linux specific. I know that WLC had some things to get their TTY code to work with FreeBSD, but I'm not going to blindly copy that code if I can't test it. Someone else can take a look at fixing this if they want to.
- The examples haven't been updated to use setcap/seteuid. You must do it manually.

If you're going to run this code, make sure you have permissions for /dev/input/event* (be part of the input group) or run as root, otherwise you'll run into #24.

### DIFF
--- a/session/direct-ipc.c
+++ b/session/direct-ipc.c
@@ -1,0 +1,241 @@
+#define _POSIX_C_SOURCE 200809L
+#include <errno.h>
+#include <stdio.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/sysmacros.h>
+#include <sys/wait.h>
+#include <xf86drm.h>
+#include <linux/major.h>
+#include <wlr/util/log.h>
+#include "session/direct-ipc.h"
+
+enum { DRM_MAJOR = 226 };
+
+#ifdef HAS_LIBCAP
+#include <sys/capability.h>
+
+static bool have_permissions(void) {
+	cap_t cap = cap_get_proc();
+	cap_flag_value_t val;
+
+	if (!cap || cap_get_flag(cap, CAP_SYS_ADMIN, CAP_PERMITTED, &val) || val != CAP_SET) {
+		wlr_log(L_ERROR, "Do not have CAP_SYS_ADMIN; cannot become DRM master");
+		cap_free(cap);
+		return false;
+	}
+
+	cap_free(cap);
+	return true;
+}
+#else
+static bool have_permissions(void) {
+	if (geteuid() != 0) {
+		wlr_log(L_ERROR, "Do not have root privileges; cannot become DRM master");
+		return false;
+	}
+
+	return true;
+}
+#endif
+
+static void send_msg(int sock, int fd, void *buf, size_t buf_len) {
+	char control[CMSG_SPACE(sizeof(fd))] = {0};
+	struct iovec iovec = { .iov_base = buf, .iov_len = buf_len };
+	struct msghdr msghdr = {0};
+
+	if (buf) {
+		msghdr.msg_iov = &iovec;
+		msghdr.msg_iovlen = 1;
+	}
+
+	if (fd >= 0) {
+		msghdr.msg_control = &control;
+		msghdr.msg_controllen = sizeof(control);
+
+		struct cmsghdr *cmsg = CMSG_FIRSTHDR(&msghdr);
+		*cmsg = (struct cmsghdr) {
+			.cmsg_level = SOL_SOCKET,
+			.cmsg_type = SCM_RIGHTS,
+			.cmsg_len = CMSG_LEN(sizeof(fd)),
+		};
+		*(int *)CMSG_DATA(cmsg) = fd;
+	}
+
+	ssize_t ret;
+	do {
+		ret = sendmsg(sock, &msghdr, 0);
+	} while (ret < 0 && errno == EINTR);
+}
+
+static ssize_t recv_msg(int sock, int *fd_out, void *buf, size_t buf_len) {
+	char control[CMSG_SPACE(sizeof(*fd_out))] = {0};
+	struct iovec iovec = { .iov_base = buf, .iov_len = buf_len };
+	struct msghdr msghdr = {0};
+
+	if (buf) {
+		msghdr.msg_iov = &iovec;
+		msghdr.msg_iovlen = 1;
+	}
+
+	if (fd_out) {
+		msghdr.msg_control = &control;
+		msghdr.msg_controllen = sizeof(control);
+	}
+
+	ssize_t ret;
+	do {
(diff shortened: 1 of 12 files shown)


SirCmpwn:
I thought I said that this design wasn't going to work. We need to have the user do the forking or setuid-ing or anything else they need tod do.

ascent12:
Currently the user is the one who needs to use setuid. We just expect having the proper permissions going in, we fork to keep them, and then the user can drop the permissions themselves later.
I'm not really sure why we would want the user to fork for us, because that means they would have to manage the IPC for us, and probably doesn't have any real advantages. It would also massively change the session interface, and cause a weird situation where the user only needs to do it when this session is used, but not for the logind session.

SirCmpwn:
I thought I explained this. The user may want to do additional things in the privledged program, including Sway.

Timidger:
Unless I'm misunderstanding @ascent12, doesn't the user keep the privileges and still has to drop it manually (like how wlc is currently)? Is there something about the user forking that affects a use case I'm not seeing? I don't see why the user should have to do the forking themselves if wlroots can do it for them and you'll always need that.

SirCmpwn:
Because then the user is just going to have to fork again and drop privs in a third instance of the process if they're going to keep root permissions for something.

Timidger:
Ok, that's a fair use case and something I hadn't considered. Shouldn't really affect me, because I can hide the forking boilerplate behind my interface in my Rust wrapper if I find it not to be useful for me.

ascent12:
I don't see why needing to fork more than once is a bad thing. A child process is just internal state to the library, just like anything else. I cannot think of any real reason why a user would care about this. It's not like it's another thread which can cause potentials for race conditions that the user needs to deal with; it's just a process which is going to be sitting idle 99% of the time, except when we change VT. I've deliberately coded it so there is as little happening in the child process as possible.

And as I was saying before, I think something different leads to a really stupid interface that the user has to deal with, which would basically be "proxy this data for the library, but only sometimes". I just cannot see the justification for something like that.

SirCmpwn:
Well, let me think about it.

session/direct-ipc.c
```diff
@@ -0,0 +1,241 @@
...
+enum { DRM_MAJOR = 226 };
+
+#ifdef HAS_LIBCAP
+#include <sys/capability.h>
+
+static bool have_permissions(void) {
+	cap_t cap = cap_get_proc();
+	cap_flag_value_t val;
+
+	if (!cap || cap_get_flag(cap, CAP_SYS_ADMIN, CAP_PERMITTED, &val) || val != CAP_SET) {
```
SirCmpwn: This should check if we're root and proceed if so. We should support both that and caps.

SirCmpwn: CAP_SYS_ADMIN is basically being root anyway

session/direct-ipc.c
```diff
@@ -0,0 +1,241 @@
...
+		ret = recvmsg(sock, &msghdr, MSG_CMSG_CLOEXEC);
+	} while (ret < 0 && errno == EINTR);
+
+	if (fd_out) {
+		struct cmsghdr *cmsg = CMSG_FIRSTHDR(&msghdr);
+		*fd_out = cmsg ? *(int *)CMSG_DATA(cmsg) : -1;
+	}
+
+	return ret;
+}
```
SirCmpwn: This seems complicated. Why not just communicate over a pipe?

SirCmpwn: Oh, presumably because we're transferring ownership of file descriptors.

session/direct-ipc.c
```diff
@@ -0,0 +1,241 @@
...
+	int drm_fd = -1;
+	bool running = true;
+
+	while (running && recv_msg(sock, NULL, &msg, sizeof(msg)) >= 0) {
+		switch (msg.type) {
+		case MSG_OPEN:
+			errno = 0;
+
+			// These are the same flags that logind opens files with
+			int fd = open(msg.path, O_RDWR|O_CLOEXEC|O_NOCTTY|O_NONBLOCK);
```
SirCmpwn: We probably want to have a whitelist of files that you can open so that a rogue child process can't have free reign over the system.

ascent12: The part further down where we check for INPUT_MAJOR or DRM_MAJOR already does that.
It'll only match /dev/dri/* or /dev/input/*.

SirCmpwn: Ah, cool.

session/direct-ipc.c
ascent12: If you're root, you automatically get all of the caps anyway.
Checking for CAP_SYS_ADMIN explicitly means that while we still have some pretty serious permissions, we don't have free reign over the filesystem to be able to mess anything up.
Having just this cap and being part of the input group was enough for me to be able to run the examples.

SirCmpwn: I thought that cap checks would only check caps, rather than shortcutting if you're root. Is that not the case? As for CAP_SYS_ADMIN, you can do *tons* of shit with it. Is there not a more suitable cap?

ascent12: Capabilities are just splitting up the special things root can do, so naturally, root has all of them. I would say try the examples as root, but it's currently it's broken at the moment and will probably require you to restart your computer. The breakage isn't part of this code though: it successfully gets past it, and also breaks on the logind backend.

Unfortunately it has to be CAP_SYS_ADMIN. The DRM code in linux explicitly checks for it, and if we don't handle being the DRM master properly, we will be unable to change VTs.

SirCmpwn: Bleh, alright.

SirCmpwn: Should I wait for you to fix this brokenness before merging?

ascent12: The last line that appears the in the log is https://github.com/SirCmpwn/wlroots/blob/master/examples/shared.c#L480 with socket printing as (null), so I think it's completely unrelated to this. It probably has something to do with environment variables or something that wl_display_add_socket_auto expects, but I'm not really sure.

It's up to you if you want that to be part of that to be part of this pull request or a separate fix.

ascent12: I did some testing, and it seems sudo is unsetting XDG_RUNTIME_DIR, which causes wl_display_add_socket_auto to fail. Running in a proper root shell or using setuid works fine.
I'll submit the code to exit cleanly when it fails, but this is something we can fix later and would be out of scope for this PR.

SirCmpwn: Cool, thanks for investigating.

PR #27, opened 2017-07-10
[Proposal] Use the Meson build system

ascent12:
I've been curious about this build system lately, so I thought I would try to learn it by applying it to wlroots. It seems to be catching on with a few other projects, including libinput and a bunch of GNOME stuff.

I think it's a lot nicer to use than CMake to me, and it seems we can get away with writing a lot less. It also builds a lot faster than CMake.

What I have here is roughly equivalent to what the current CMake implementation does. It builds the library and the examples, but doesn't install/distribute the library yet.
Also, it doesn't use subdirectory build files; everything exists in the top-level. I can be written to use subdirectories, but I wrote it this way because it was simpler.

I'm not really pushing this too hard. If you're not interested in changing off of CMake, just close this PR.
It was just something I threw together pretty quickly, and I thought it was promising.


SirCmpwn:
You've piqued my interest. I'd like to see what this looks like with subdirectories.

FSMaxB:
The thing about CMake being faster is mostly because of make, if you use Ninja as a backend for CMake, most of the bad performance goes away.

But meson is definitely faster **by default**. It also trivially allows enabling features like Link Time Optimization, building with AddressSanitizer/UndefinedBehaviorSanitizer, running the tests with Valgrind, code coverage etc. without writing any additional build system code (in most cases). And its a hell of a lot easier to reason about than CMake (I still have nightmares about CMake, especially since I'm still using it for most of my projects).

I'm not a contributor to wlroots and I'm probably not going to become one in the foreseeable future, but I am definitely in favor of this change. With developers having to focus less on the build system overall, the code can only get better.

SirCmpwn:
Thanks!

PR #28, opened 2017-07-11
Add null session

ascent12:
This is just a session backend which does nothing.
It's so when we're using a wayland/xorg/whatever (anything that isn't DRM, really) backend, the user can be given this session and not worry about checking what backend is actually being used.

This fixes a bug where closing one of the examples using the wayland backend would absolutely destroy your session (massive visual glitches, unable to change VT), and forces you to restart your computer. This happened to me when using an example under weston under X.

This PR doesn't add "session/null.c" to the build system. I'll add it when/if the meson PR is accepted.

Instead of doing this sort of patchwork fix, we might want to consider whether we want wlr_session to be part of the public API. Maybe all of the session stuff could be handled in `wlr_backend_autocreate`, so we don't accidentally try to open a logind/direct session when we're not using DRM.

### DIFF
--- a/session/null.c
+++ b/session/null.c
@@ -1,0 +1,63 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <wayland-server.h>
+#include <wayland-util.h>
+#include <wlr/session/interface.h>
+#include <wlr/util/log.h>
+
+const struct session_impl session_null;
+
+static bool null_session_change_vt(struct wlr_session *session, unsigned vt) {
+	return false;
+}
+
+static void null_session_close(struct wlr_session *session, int fd) {
+	// Calling this function is a programming error
+	wlr_log(L_ERROR, "Invalid call to wlr_session_close with null session");
+	exit(1);
+}
+
+static int null_session_open(struct wlr_session *session, const char *path) {
+	// Calling this function is a programming error
+	wlr_log(L_ERROR, "Invalid call to wlr_session_open with null session");
+	exit(1);
+}
+
+static void null_session_finish(struct wlr_session *session) {
+	free(session);
+}
+
+static struct wlr_session *null_session_start(struct wl_display *disp) {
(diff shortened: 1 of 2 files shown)


session/null.c
```diff
@@ -0,0 +1,63 @@
...
+	wlr_log(L_ERROR, "Invalid call to wlr_session_open with null session");
+	exit(1);
+}
+
+static void null_session_finish(struct wlr_session *session) {
+	free(session);
+}
+
+static struct wlr_session *null_session_start(struct wl_display *disp) {
+	if (!getenv("WAYLAND_DISPLAY") && !getenv("_WAYLAND_DISPLAY") && !getenv("DISPLAY")) {
```
SirCmpwn: Not sure how I feel about this. It ties too strongly into the backends that *we* implement and has no consideration for the backends the *user* implements.

SirCmpwn:
>Instead of doing this sort of patchwork fix, we might want to consider whether we want wlr_session to be part of the public API. Maybe all of the session stuff could be handled in wlr_backend_autocreate, so we don't accidentally try to open a logind/direct session when we're not using DRM.

We want as much as possible to be in the public API. We want the user to have as much control over the system as we can give them.

session/null.c
ascent12: Yeah, it is pretty crappy. I cannot think of a reliable way to test "is going to be drm", though.

ascent12:
>We want as much as possible to be in the public API. We want the user to have as much control over the system as we can give them.

Yes, but wlr_session is tied so ridiculously close to the DRM backend, and isn't generally useful for other backends at all.
I think session should just be bundled into DRM and then the state about the session could be accessed through the 'backend-specific state' functions, whenever those get added.

SirCmpwn:
That's fair. I wouldn't be opposed to rolling this more closely into DRM.

ascent12:
Ok. This PR isn't really needed then.
I'll look into the best place it fits at some point.

PR #29, opened 2017-07-11
Moved session into backend/session and changed ownership

ascent12:
The session stuff is tied very closely to the DRM/libinput backend, so it didn't really make sense for it to be sitting at the top level of project. I moved it into backend/, where it can live alongside other similar things like wlr_udev (also very DRM/libinput specific).

Also, the user no longer needs to pass a wlr_session to `wlr_backend_autocreate`: it will create one on its own.
The multi backend has been modified to accept a wlr_session and a wlr_udev. It takes "ownership" of them, and will clean them up properly when it is destroyed (previously, we leaked the wlr_udev). If the multi backend is ever used for something that doesn't need a wlr_session/wlr_udev, it's perfectly safe to just pass it NULL.

This fixes the broken session issue I outlined in #28.
~~Again, the build system code is not committed. I'll wait for #27.~~

Currently, this breaks the ability to change VT in the examples, because the user doesn't have access to the wlr_session any more. We need to decide how the VT changing function and how the other backend-specific functions are going to look.


SirCmpwn:
bump

ascent12:
I've rebased this against master and added the appropriate build files.

I also added a backend specific function `wlr_multi_get_session` which the examples now use to get access to the session so they can change VTs. The backend specific functions probably need to be fleshed out more and have more thought put into them, but currently this works are a temporary solution.
Now XKB_KEY_XF86Switch_VT_[0-12] is used, so Ctrl + Alt + F[0-12] is used to change VT.

SirCmpwn:
Pushed a couple of small changes. Looks good otherwise, thanks!

PR #31, opened 2017-07-17
Use meson as build system

nyorain:
This picks up the work done by @ascent12, but should fix (and improve) the meson build scripts.
It fixes a bug in wlr_output.c along the way that made the pointer example fail (partially) for me.
I also tried to modify the travis integration, let's see if this works...


nyorain:
Seems like it does! should the commits be squashed?

SirCmpwn:
Yes please

nyorain:
Also updated the README for meson now.

SirCmpwn:
I'm on board with meson, but this needs to be updated.

nyorain:
Rebased it.
As a notice it would be really a good idea to add the required versions to the dependencies in future so meson can output an error instead of getting a compile error (noticed this when testing on ubuntu, seems like you already use features not in libinput 1.6).
Not sure though which versions are required and it isn't too important until this gets ready for the first release.

ascent12:
Yeah, this stuff gets tested with very up-to-date libraries and recent kernel versions.
Basically, if it's in Arch's stable repo, it's good to go, or at least that's how I've thought of it.

Sorting out which library versions work is something that can wait until much later.

README.md
```diff
@@ -23,8 +23,5 @@ Install dependencies:
 
 Run these commands:
 
-    mkdir build
-    cd build
-    cmake -DCMAKE_BUILD_TYPE=Release ..
-    make
-    sudo make install
+    meson build
+	ninja -C build
```
SirCmpwn: Indent issue

types/wlr_output.c
```diff
@@ -206,7 +206,7 @@ bool wlr_output_set_cursor(struct wlr_output *output,
 		output->cursor.texture = wlr_render_surface_init(output->cursor.renderer);
 	}
 
-	wlr_surface_attach_pixels(output->cursor.texture, WL_SHM_FORMAT_XBGR8888,
+	wlr_surface_attach_pixels(output->cursor.texture, WL_SHM_FORMAT_ABGR8888,
```
SirCmpwn: What's this doing here?

acrisci:
Works for me.

types/wlr_output.c
nyorain: As mentioned in the first pr comment i stumbled across the same bug that was fixed by @acrisci a few days ago. I used abgr which i guess makes more sense than xbgr (since we are attaching a cursor which will usuallly have an alpha channel)? i know this does not belong here but could not really test the meson build without this when i wrote it.

nyorain: Could change it to xbgr here though since it currently does not matter for any example if this stands in the way of merging the pr.

PR #36, opened 2017-07-27
Use ARGB8888 pixel format for cursor bo

acrisci:
The previous pixel format (RGBA8888) is not supported on all devices for cursors.

Weston uses ARGB8888 exclusively for its cursor buffer object which leads me to
believe this pixel format is better supported.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -304,6 +304,13 @@ static bool wlr_drm_output_set_cursor(struct wlr_output_state *output,
 		return true;
 	}
 
+	if (!gbm_device_is_format_supported(state->renderer.gbm,
+				GBM_FORMAT_ARGB8888, GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE)) {
+		wlr_log(L_ERROR, "Failed to create cursor bo: ARGB8888 pixel format is "
+				"unsupported on this device");
+		return false;
+	}
+
 	uint64_t bo_width, bo_height;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


acrisci:
see [drm_output_init_cursor_egl()](https://github.com/wayland-project/weston/blob/master/libweston/compositor-drm.c#L2068)

acrisci:
I don't know if that check for support is strictly necessary. I put it in there to illustrate my point so you could check what is supported on your device.

backend/drm/drm.c
```diff
@@ -304,6 +304,11 @@ static bool wlr_drm_output_set_cursor(struct wlr_output_state *output,
 		return true;
 	}
 
+	if (!gbm_device_is_format_supported(state->renderer.gbm, GBM_FORMAT_ARGB8888, GBM_BO_USE_CURSOR | GBM_BO_USE_WRITE)) {
```
SirCmpwn: Can you 80 columns this

SirCmpwn:
Can't hurt to have the check.

ascent12:
The only reason I used RGBA8888 to begin with was because that is the format that the cat image was in, so this will mess up the colours.
@SirCmpwn can we get a ARGB version of the cat image?

SirCmpwn:
Gimp can't export ARGB, so the conversion is going to have to be done manually.

ascent12:
I suppose we could use GL for this? Requiring users to have their images in the somewhat uncommon format seems like it would just be annoying.

I've been looking at adding DRM plane support (which includes cursors) as part of the atomic modesetting work I've been looking into. This will probably remove most of the cursor-specific rendering stuff we have going on here, and makes it easy to just use GL.
Though I've been pretty busy lately and probably will be for the foreseeable future, so it may be quite a while before this work is finished.

acrisci:
Or instead of using a cat picture you could use a, you know, normal cursor set.

ascent12:
>Or instead of using a cat picture you could use a, you know, normal cursor set.

I just looked up the Xcursor spec and apparently it is stored as ARGB. I though it would have been RGBA, but I guess I was mistaken.
It still might be nice being able to attach a PNG or something.

SirCmpwn:
A little Xcursor utility library is something I want to see in wlroots at some point.

SirCmpwn:
Works for me. Thanks!

ascent12:
>A little Xcursor utility library is something I want to see in wlroots at some point.

Does wayland-cursor would do this for us? I don't really know a lot about it, but I think it works for compositors and not just clients.

SirCmpwn:
Last I checked, it was pretty client specific. I may have missed something, though.

PR #39, opened 2017-07-31
Use correct pixel format enum in touch example

acrisci:
Fix a call to `wlr_surface_attach_pixels()` in the main method of the touch
example to use the correct enum for this method (wayland instead of gl).


SirCmpwn:
Thanks!

PR #41, opened 2017-08-02
Add surface interface stubs in compositor example

acrisci:
Add the wayland surface interface to the example compositor.

Implement the create_surface method to create a new wlr surface from the
wayland surface and add the interface.

### DIFF
--- a/examples/compositor/wl_compositor.c
+++ b/examples/compositor/wl_compositor.c
@@ -78,7 +78,28 @@ struct wl_surface_interface surface_interface = {
 };
 
 static void destroy_surface(struct wl_resource *resource) {
-	wlr_log(L_DEBUG, "TODO: destroy surface");
+	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+	wlr_surface_destroy(surface);
+}
+
+static void destroy_surface_listener(struct wl_listener *listener, void *data) {
+	struct wl_compositor_state *state;
+	struct wlr_surface *surface = data;
+	state = wl_container_of(listener, state, destroy_surface_listener);
+
+	struct wl_resource *res = NULL, *_res;
+	wl_list_for_each(_res, &state->surfaces, link) {
+		if (_res == surface->resource) {
+			res = _res;
+			break;
+		}
+	}
+
+	if (!res) {
+		return;
+	}
+
+	wl_list_remove(&res->link);
 }
 
 static void wl_compositor_create_surface(struct wl_client *client,
@@ -87,10 +108,12 @@ static void wl_compositor_create_surface(struct wl_client *client,
 	struct wl_resource *surface_resource = wl_resource_create(client,
 			&wl_surface_interface, wl_resource_get_version(resource), id);
 	struct wlr_surface *surface = wlr_render_surface_init(state->renderer);
+	surface->resource = surface_resource;
 	wl_resource_set_implementation(surface_resource, &surface_interface,
 			surface, destroy_surface);
 	wl_resource_set_user_data(surface_resource, surface);
 	wl_list_insert(&state->surfaces, wl_resource_get_link(surface_resource));
+	wl_signal_add(&surface->destroy_signal, &state->destroy_surface_listener);
 }
 
 static void wl_compositor_create_region(struct wl_client *client,
@@ -137,6 +160,7 @@ void wl_compositor_init(struct wl_display *display,
 		&wl_compositor_interface, 4, state, wl_compositor_bind);
 	state->wl_global = wl_global;
 	state->renderer = renderer;
+	state->destroy_surface_listener.notify = destroy_surface_listener;
 	wl_list_init(&state->wl_resources);
 	wl_list_init(&state->surfaces);
 }
--- a/examples/compositor/wl_compositor.c
+++ b/examples/compositor/wl_compositor.c
@@ -3,13 +3,95 @@
 #include <wlr/util/log.h>
 #include "compositor.h"
 
+static void surface_destroy(struct wl_client *client, struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static void surface_attach(struct wl_client *client,
+		struct wl_resource *resource,
+		struct wl_resource *buffer_resource, int32_t sx, int32_t sy) {
+	wlr_log(L_DEBUG, "TODO: surface attach");
+}
+
+static void surface_damage(struct wl_client *client,
+		struct wl_resource *resource,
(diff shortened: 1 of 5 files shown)


acrisci:
I had to add a `wlr_renderer` to the `wl_compositor` struct to create the `wlr_surface` in the `create_surface` method but I don't know if that was your intent.

I think the `wlr_surface` needs to be saved in a list somewhere and then accessed in the `handle_output_frame` method for rendering but I'm not sure exactly what the best way to do that is.

I also have a semi-working implementation locally of attaching an shm buffer to the surface where I can at least access the buffer contents and the gl calls don't fail but I have yet to successfully paint anything on the screen. I'm working on that now.

My test client is located in my notes repository [here](https://github.com/acrisci/scribbles/tree/master/wayland-window).

SirCmpwn:
- I recommend `weston-simple-shm` for testing.
- It makes sense that wl_compositor would need a wlr_renderer to work
- Yes the surface needs to be stored in a list and rendered each frame

SirCmpwn:
There are some style issues here, but GitHub fucked up my web browser so I can't comment inline. Put `{` on the same line and wrap to 80 columns.

acrisci:
Ok I cleaned it up a bit. I think this is a good starting point for the surface implementation.

Next I will add the surfaces to a list that can be accessed when they need to be rendered and implement attaching the shm buffer.

examples/compositor/wl_compositor.c
```diff
@@ -3,9 +3,91 @@
...
+		surface_destroy,
+		surface_attach,
+		surface_damage,
+		surface_frame,
+		surface_set_opaque_region,
+		surface_set_input_region,
+		surface_commit,
+		surface_set_buffer_transform,
+		surface_set_buffer_scale,
+		surface_damage_buffer
```
SirCmpwn: Double indented?

SirCmpwn:
LGTM

SirCmpwn:
Are you expecting this to be merged before you attach buffers?

Also, side note, in the future wlr_surface is going to need to handle damage, likely in cooperation with wlr_renderer. A surface is going to have to keep track of what parts of it are damaged and bubble it up to the user somehow, and partial rerendering is something to be considered.

acrisci:
> Are you expecting this to be merged before you attach buffers?

However you want to work is fine. We can do one big pull request or I can just request comments on every commit. I have about a month off work and this is my main project right now so I should be moving fairly quickly.

> Also, side note, in the future wlr_surface is going to need to handle damage

I'll look at the weston and wlc implementation and use my best judgement on how to do this.

SirCmpwn:
Let's implement this up to the point of attaching buffers, and do the rest separately. Right now there's not really much going on here, not really worth merging.

SirCmpwn:
Also, the usual wisdom is to track damage with pixman. To some degree it will be necessary for compositors to combine this with their own information about where views are being rendered to composite the damaged regions with underlying transparent surfaces and so on.

acrisci:
Ok, I implemented attaching the buffer to the surface. This commit actually makes shit show up on the screen.

SirCmpwn:
Pic?

SirCmpwn:
LGTM codewise

acrisci:
image

This is made using my [example client](https://github.com/acrisci/scribbles/tree/master/wayland-window) which attaches a buffer that consists of all green pixels. The weston-simple-shm won't work yet because we don't implement a shell interface.

When you close the window, the server segfaults.

SirCmpwn:
Can you address the segfault before this is merged?

render/gles2/surface.c
```diff
@@ -34,6 +34,35 @@ static bool gles2_surface_attach_pixels(struct wlr_surface_state *surface,
...
+	uint8_t *pixels = wl_shm_buffer_get_data(buffer);
+	int width = wl_shm_buffer_get_width(buffer);
+	int height = wl_shm_buffer_get_height(buffer);
+	int stride = wl_shm_buffer_get_stride(buffer);
+	surface->wlr_surface->width = width;
+	surface->wlr_surface->height = height;
+	surface->wlr_surface->format = format;
+	surface->pixel_format = fmt;
+
+	GL_CALL(glActiveTexture(GL_TEXTURE0));
```
acrisci: I'm not exactly sure why I need this call to glActiveTexture, but it seems to be necessary. Here is my analysis.

`GL_TEXTURE0` is the default active texture when the compositor starts. Upon the first render, the surface is bound with `gles2_surface_bind()` which changes the active texture with `GL_CALL(glActiveTexture(GL_TEXTURE0 + 1));`.

So the first surface is attached with active texture 0 and then it is rendered. Then the second texture is attached with active texture 1 which doesn't seem to work and I just get a black window. In fact, if I attach the first surface with texture 1, I get a black window as well so it seems we must do this with texture 0 active.

Anyway, this works but maybe there is a better way to manage this state. I guess I need to learn more about opengl.

render/gles2/surface.c
```diff
@@ -34,6 +34,35 @@ static bool gles2_surface_attach_pixels(struct wlr_surface_state *surface,
...
+	const struct pixel_format *fmt = gl_format_for_wl_format(format);
+	if (!fmt || !fmt->gl_format) {
+		wlr_log(L_ERROR, "No supported pixel format for this surface");
+		return false;
+	}
+	wl_shm_buffer_begin_access(buffer);
+	uint8_t *pixels = wl_shm_buffer_get_data(buffer);
+	int width = wl_shm_buffer_get_width(buffer);
+	int height = wl_shm_buffer_get_height(buffer);
+	int pitch = wl_shm_buffer_get_stride(buffer) / (fmt->bpp / 8);
```
acrisci: Weston calls this variable `pitch` although I don't know if that term is the same as `stride`. This division I added is necessary for setting `GL_UNPACK_ROW_LENGTH_EXT` because otherwise I get memory errors from stride being too large and the opengl skipping too many pixels.

Wayland seems to include some sanity checks for us so the stride will never cause these memory errors as long as we do this calculation.

acrisci:
The next commit implements destroying surfaces and is ready for review.

I hate to add another callback for surface destruction (there are 3 now), but it was the best way I could find to get both the compositor state and the wlr_surface in the same callback.

SirCmpwn:
Great work, thanks!

render/gles2/surface.c
SirCmpwn: FTR both wlc and weston do some special shit with GL textures that I don't fully understand and have not replicated here.

acrisci: This is a really good explanation of texture units that I read: https://open.gl/textures#Textureunits

PR #42, opened 2017-08-05
Remove unnecessary calls to glActiveTexture

acrisci:
Since wlroots shaders only use one texture at a time (ie there is only one
sampler2D variable in any shader), it is unnecessary to switch between active
texture units at this time.


SirCmpwn:
I wonder if at some point it would make a performance impact if we prepared lots of surfaces for rendering and render them all in one go on the shader.

PR #44, opened 2017-08-06
DRM plane support, refactoring, and other changes

ascent12:
This is part of an effort to get atomic modesetting and multi-GPU support for DRM. This PR doesn't do those, but it's laying the groundwork for it.
Instead of just dumping what would probably be 4000+ lines of code when I'm finished, I'm submitting this now to get opinions and make it less daunting to review (though this is still pretty big).

# Significant changes
## DRM planes
These are hardware-backed rendering surfaces that we can use to efficiently layer different things over the main surface. They currently aren't exposed to users, but they can be used for some optimisations that a compositor might do (e.g. putting a fullscreen client on an overlay plane, so they don't need to discard whatever was drawn under them). They're needed for atomic modesetting.
As a side effect of this, it's allowed me to generalise some of the cursor rendering code.
## Upgraded CRTC <-> Encoder matching
We now have a MUCH more complicated way of doing this (it's an NP-complete problem), instead of the greedy/naive method that we were doing before.
Hopefully this will make hotplugging and modesetting work properly for a wider range of hardware.
I would like to know if this actually fixes anything for anyone else (as compared to master currently).
## DRM properties
I've written some infrastructure to allow us to easily read/set drm properties. Before, we only needed like 2 properties and we could just if-else-if them, but with atomic modesetting, there are significantly more.
This code is doing some funny tricks with anonymous unions and arrays, but it's all there to get of code duplication.


SirCmpwn:
Weston renders the software cursor in a plane, should we do something similar?

SirCmpwn:
>As a side effect of this, it's allowed me to generalise some of the cursor rendering code.

Ah, maybe you have. I will read the code.

backend/drm/drm-properties.h
```diff
@@ -0,0 +1,59 @@
...
+		uint32_t rotation;
+		uint32_t scaling_mode;
+	};
+	uint32_t props[2];
+};
+
+union wlr_drm_plane_props {
+	struct {
+		uint32_t type;
+		uint32_t rotation; // Not guranteed to exist
```
SirCmpwn: What do you think about adding a bitfield to indicate which properties are valid?

backend/drm/drm-properties.h
```diff
@@ -56,4 +56,7 @@ bool wlr_drm_get_connector_props(int fd, uint32_t id, union wlr_drm_connector_pr
 bool wlr_drm_get_crtc_props(int fd, uint32_t id, union wlr_drm_crtc_props *out);
 bool wlr_drm_get_plane_props(int fd, uint32_t id, union wlr_drm_plane_props *out);
 
+bool wlr_drm_get_prop(int fd, uint32_t obj, uint32_t prop, uint64_t *ret);
+void *wlr_drm_get_prop_blob(int fd, uint32_t obj, uint32_t prop, size_t *ret_len);
```
SirCmpwn: If we have this `get_prop` et al do we need to have the other 3 functions and the structs made public? I'd rather hide some of that cruft if we can get away with it.

SirCmpwn:
Looks good overall. Can you do a pass on style? Make sure you put `{` on the same line as function defs and always include `{}` in if statements.

backend/drm/drm-properties.h
ascent12: Missing properties will be set to zero (assuming the whole thing was allocated with calloc), so it would be unnecessary.

SirCmpwn: Are there no situations in which a property could genuinely be set to zero?

ascent12: Yeah, I'm pretty sure that basically all DRM objects ids (including properties) can't genuinely be zero.

backend/drm/drm-properties.h
ascent12: We still need the union wlr_drm_*_props, because they contain the prop argument. I would consider these all to be internal to the DRM backend and probably not particularly useful for a user, so I feel these don't need to be exposed.
Even if a user wanted to muck around with DRM properties for whatever reason, as long as we give them the DRM fd, they can just do it themselves.

SirCmpwn: Then let's just keep the union internal and return a magic uint32_t black box.

ascent12: I agree that the union looks strange in the header (most uses of unions usually raise a few eyebrows), but due to anonymous structs/unions, people don't even need to think of that detail; it just looks like using a struct in the code. It gives a nice, safe, clean-looking interface, and overall I don't think it's particularly 'offensive'.

Also do we even want to expose the wlr_drm_{crtc,plane} types directly to users? I not really sure how useful they would be.

SirCmpwn: Until someone external asks for it, let's keep it internal.

ascent12:
Unfortunately it looks like the cursor rendering code will need to be changed back. I tried this code on my laptop and it seems like (GBM_BO_USE_RENDERING | GBM_BO_USE_CURSOR) isn't guaranteed to be supported.
That's a massive pain in the ass, because now we can't use OpenGL to manage rotations and share common rendering code with the other plane types.

ascent12:
Actually, I managed to fix it while keeping the common rendering code.
It now creates a separate gbm_bo (with GBM_BO_USE_CURSOR) and then whatever we drew is copied in with glDrawPixels/gbm_bo_map.
It leads to an extra layer of copying, but we can now still use OpenGL and easily be able to rotate things.

SirCmpwn:
Looks good, thanks!

PR #48, opened 2017-08-08
Rename wlr_surface -> wlr_texture; attach -> upload

nyorain:
Compiles without warnings and all examples still run the same for me.
Does not do anything else than renaming.


backend/drm/drm.c
```diff
@@ -605,15 +605,15 @@ static bool wlr_drm_output_set_cursor(struct wlr_output_state *output,
...
+		wlr_matrix_texture(plane->matrix, plane->width, plane->height,
 			output->base->transform ^ WL_OUTPUT_TRANSFORM_FLIPPED_180);
 
 		plane->wlr_rend = wlr_gles2_renderer_init();
 		if (!plane->wlr_rend) {
 			return false;
 		}
 
-		plane->wlr_surf = wlr_render_surface_init(plane->wlr_rend);
+		plane->wlr_surf = wlr_render_texture_init(plane->wlr_rend);
```
SirCmpwn: Grepping for `surf` should be enough, I don't think there's any other shorthand in use.

nyorain: already tried to do it but misssed this one. Will search for 'surf' as well then

nyorain: updated it

SirCmpwn:
LGTM, thanks!

PR #49, opened 2017-08-09
Atomic modesetting

ascent12:
This adds support for the new atomic modesetting DRM interface, where available.
It's supposed to be quicker and prevent screen tearing, especially when working with planes.

Most DRM drivers don't have support for this yet, but the latest Linux version enables it by default for the i915 (intel) driver, which is what I tested it on. amdgpu is supposed to get it when their DC/DAL code lands. I have no idea when radeon or nouveau is going to get it.

This splits the DRM functionality between 'drm-legacy' and 'drm-atomic' and creates an interface for the common DRM code to use.


ascent12:
I looked just now, and it seems nouveau already has atomic modesetting support, but it's currently not enabled by default.

SirCmpwn:
Thanks!

PR #50, opened 2017-08-09
Add first wlr_surface example implementation

nyorain:
Implements no real new functionality for it but makes sure buffer attaching already follows the wayland protocol and updates the compositor example. The base for further work on implementing wl_surface.

### DIFF
--- a/examples/compositor/wlr_surface.c
+++ b/examples/compositor/wlr_surface.c
@@ -1,0 +1,120 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <wayland-server.h>
+#include <wlr/util/log.h>
+#include <wlr/render/interface.h>
+#include "wlr_surface.h"
+
+static void surface_destroy(struct wl_client *client, struct wl_resource *resource) {
+	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+	wl_signal_emit(&surface->signals.destroy, surface);
+	wl_resource_destroy(resource);
+}
+
+static void surface_attach(struct wl_client *client,
+		struct wl_resource *resource,
+		struct wl_resource *buffer, int32_t sx, int32_t sy) {
+	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+	surface->pending_buffer = buffer;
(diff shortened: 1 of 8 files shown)


examples/compositor/wlr_surface.c
```diff
@@ -0,0 +1,120 @@
...
+	struct wlr_surface *surface = calloc(1, sizeof(struct wlr_surface));
+	surface->texture = wlr_render_texture_init(renderer);
+	surface->resource = res;
+	wl_signal_init(&surface->signals.commit);
+	wl_signal_init(&surface->signals.destroy);
+	wl_resource_set_implementation(res, &surface_interface,
+			surface, destroy_surface);
+
+	return surface;
+}
```
SirCmpwn: This can live inside of wlroots (make a new sublibrary called `wayland` full of implementations of wayland interfaces).

SirCmpwn: Better yet, just put it in `types`?

nyorain: will put it in ```types```

types/wlr_surface.c
```diff
@@ -0,0 +1,119 @@
...
+}
+
+static void surface_commit(struct wl_client *client,
+		struct wl_resource *resource) {
+	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+
+	// apply pending state
+	if (surface->pending_attached) {
+		surface->attached = surface->pending_buffer;
+		if (surface->pending_buffer) {
```
SirCmpwn: I would change this a bit. Drop the `pending_attached` variable and just test if `pending_buffer` is not NULL here. I would also rename `attached` to `buffer`.

nyorain: the problem is that attaching a null buffer is valid and simply removes the currently attached buffer. And attached is a bool.

SirCmpwn: I see. Well, that's fine then. If you rebase this I'll go ahead and merge it.

SirCmpwn:
Thanks!

PR #51, opened 2017-08-09
Implement wlr_region using pixman

nyorain:
Rather small pr, makes wlr dependent on pixman.
Will be needed to further implement wl_surface.
Not sure if it is ok to not use something wlr_region and instead directly setting the pixman region as the regions resource user data.

### DIFF
--- a/include/wlr/types/wlr_region.h
+++ b/include/wlr/types/wlr_region.h
@@ -1,0 +1,10 @@
+#ifndef _WLR_TYPES_REGION_H
+#define _WLR_TYPES_REGION_H
+
+struct wl_resource;
+
+// Implements the given resource as region.
+// Sets the associated pixman_region32_t as userdata.
+void wlr_region_create(struct wl_resource *res);
+
+#endif
--- a/.build.yml
+++ b/.build.yml
@@ -7,6 +7,7 @@ packages:
   - libinput
   - libxkbcommon
   - libcap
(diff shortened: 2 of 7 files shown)


include/wlr/types/wlr_region.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _WLR_TYPES_REGION_H
+#define _WLR_TYPES_REGION_H
+
+struct wl_resource;
+
+// Implements the given resource as region.
+// Sets the associated pixman_region32_t as userdata.
+void wlr_region_create(struct wl_resource *res);
```
SirCmpwn: Can you return the `pixman_region32_t`?

nyorain: thought about, are you sure this is a good idea? the region is directly connected with the resource so why should it be needed on creation? changing it independently from the client will result in weird behavior and it should only be read when it is passed by the client as parameter somewhere i guess

SirCmpwn: In case the user wants to stow it away somewhere to access it without relying on the user data.

nyorain: But it cannot know when the region is destroyed. The region is probably only created by the compositor implementation.

SirCmpwn: Alright, w/e.

PR #43, opened 2017-08-05
shell surface interface

acrisci:
Add stubs for the wl_shell_surface interface.

Implement wl_shell_get_shell_surface by creating the shell surface and settings
its implementation to these stubs.

PR #52, opened 2017-08-09
Implement surface frame callback

acrisci:
The surface frame callback lets a window know when it is a good time to show the next frame if it is animating. In particular, this callback is used by weston-simple-shm to throttle drawing.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -35,22 +35,22 @@ static void surface_frame(struct wl_client *client,
 	struct wlr_frame_callback *cb;
 	struct wlr_surface *surface = wl_resource_get_user_data(resource);
 
-	cb = malloc(sizeof *cb);
+	cb = malloc(sizeof(struct wlr_frame_callback));
 	if (cb == NULL) {
 		wl_resource_post_no_memory(resource);
 		return;
 	}
 
-	cb->resource = wl_resource_create(client, &wl_callback_interface, 1,
-					  callback);
+	cb->resource = wl_resource_create(client,
+			&wl_callback_interface, 1, callback);
 	if (cb->resource == NULL) {
 		free(cb);
 		wl_resource_post_no_memory(resource);
 		return;
 	}
 
-	wl_resource_set_implementation(cb->resource, NULL, cb,
-				       destroy_frame_callback);
+	wl_resource_set_implementation(cb->resource,
+			NULL, cb, destroy_frame_callback);
 
 	wl_list_insert(surface->frame_callback_list.prev, &cb->link);
 }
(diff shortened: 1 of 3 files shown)


acrisci:
This along with #43 will make weston-simple-shm actually work.

types/wlr_surface.c
```diff
@@ -23,9 +23,36 @@ static void surface_damage(struct wl_client *client,
...
+	free(cb);
+}
+
 static void surface_frame(struct wl_client *client,
 		struct wl_resource *resource, uint32_t callback) {
-	wlr_log(L_DEBUG, "TODO: surface frame");
+	struct wlr_frame_callback *cb;
+	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+
+	cb = malloc(sizeof *cb);
```
SirCmpwn: Add parens

types/wlr_surface.c
```diff
@@ -55,6 +82,7 @@ static void surface_commit(struct wl_client *client,
 			} else {
 				uint32_t format = wl_shm_buffer_get_format(buffer);
 				wlr_texture_upload_shm(surface->texture, format, buffer);
+				wl_resource_queue_event(surface->pending_buffer, WL_BUFFER_RELEASE);
```
nyorain: i would have expected this to use wl_buffer_send_release, have never seen a manual call to wl_resource_queue_event before actually... any reasons to use this here?

acrisci: It might work. I'm following the example of weston here. They could be using old apis though.

Try grepping `wl_buffer_send_release` in the weston source and there are no results. Try grepping `WL_BUFFER_RELEASE` and you will see two results.

SirCmpwn: When in doubt I would prefer the Weston approach.

nyorain: looked the functions up and it boils down to whether to write the release message immediately or write it later (like in "queue"). So i guess both work, nvm go with the weston approach then.

SirCmpwn:
Tested it and it seems good.

types/wlr_surface.c
```diff
@@ -23,9 +23,36 @@ static void surface_damage(struct wl_client *client,
...
+	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+
+	cb = malloc(sizeof *cb);
+	if (cb == NULL) {
+		wl_resource_post_no_memory(resource);
+		return;
+	}
+
+	cb->resource = wl_resource_create(client, &wl_callback_interface, 1,
+					  callback);
```
SirCmpwn: Same alignment problem I mentioned in the other PR

SirCmpwn:
Looks good aside from style nits, which I can pick myself.

SirCmpwn:
Thanks!

PR #53, opened 2017-08-09
Implement drm (egl) buffer attaching

nyorain:
This adds support for attaching egl buffers (or more general and correct: buffers coming from wl_drm).
It expands wlr_egl to load the required extensions and makes the gles renderer dependent on the egl
object.

An unresolved problem that probably needs to be fixed until this can get merged is that i had to disable software cursor support in wlr_output.c since i don't think we currently have access to an wlr_egl object there. Is it ok by design that there are multiple renderer objects? Then i could add support for gles renderers that simply cannot attach drm buffers since they were initialized without a wlr_egl object.

Another quick workaround i used here that may be problematic is the fact that wlr_egl is now exposed from wlr_backend and this leads to an somewhat ugly situation in the multi backend. I don't think this can be handled in a better way with the current backend/egl design though.

Also it currently simply uses the rgba pixel format for the texture, not sure how this could be handled correctly (but it works, so...).

### DIFF
--- a/backend/egl.c
+++ b/backend/egl.c
@@ -1,9 +1,16 @@
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
-#include <GLES3/gl3.h>
+#include <GLES2/gl2.h>
 #include <gbm.h> // GBM_FORMAT_XRGB8888
+#include <stdlib.h>
 #include <wlr/util/log.h>
 #include "backend/egl.h"
+
+// Extension documentation
+// https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_image_base.txt.
+// https://cgit.freedesktop.org/mesa/mesa/tree/docs/specs/WL_bind_wayland_display.spec
+
+struct wlr_egl *egl_global;
 
 const char *egl_error(void) {
 	switch (eglGetError()) {
@@ -45,6 +52,12 @@ const char *egl_error(void) {
 // EGL extensions
 PFNEGLGETPLATFORMDISPLAYEXTPROC get_platform_display;
 PFNEGLCREATEPLATFORMWINDOWSURFACEEXTPROC create_platform_window_surface;
+
+PFNEGLCREATEIMAGEKHRPROC eglCreateImageKHR;
+PFNEGLDESTROYIMAGEKHRPROC eglDestroyImageKHR;
+PFNEGLQUERYWAYLANDBUFFERWL eglQueryWaylandBufferWL;
+PFNEGLBINDWAYLANDDISPLAYWL eglBindWaylandDisplayWL;
+PFNEGLUNBINDWAYLANDDISPLAYWL eglUnbindWaylandDisplayWL;
 
 static bool egl_exts() {
 	get_platform_display = (PFNEGLGETPLATFORMDISPLAYEXTPROC)
@@ -109,7 +122,8 @@ static bool egl_get_config(EGLDisplay disp, EGLConfig *out, EGLenum platform) {
 	return false;
 }
 
-bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *display) {
+bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform,
+		void *remote_display) {
 	if (!egl_exts()) {
 		return false;
 	}
@@ -119,7 +133,7 @@ bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *display) {
 		goto error;
 	}
 
-	egl->display = get_platform_display(platform, display, NULL);
+	egl->display = get_platform_display(platform, remote_display, NULL);
 	if (egl->display == EGL_NO_DISPLAY) {
 		wlr_log(L_ERROR, "Failed to create EGL display: %s", egl_error());
 		goto error;
@@ -147,11 +161,31 @@ bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *display) {
 	}
 
 	eglMakeCurrent(egl->display, EGL_NO_SURFACE, EGL_NO_SURFACE, egl->context);
+	egl->egl_exts = eglQueryString(egl->display, EGL_EXTENSIONS);
+	if (strstr(egl->egl_exts, "EGL_WL_bind_wayland_display") == NULL ||
+		strstr(egl->egl_exts, "EGL_KHR_image_base") == NULL) {
+		wlr_log(L_ERROR, "Required egl extensions not supported");
+		goto error;
+	}
+
+	eglCreateImageKHR = (PFNEGLCREATEIMAGEKHRPROC)
+		eglGetProcAddress("eglCreateImageKHR");
+	eglDestroyImageKHR = (PFNEGLDESTROYIMAGEKHRPROC)
+		eglGetProcAddress("eglDestroyImageKHR");
+	eglQueryWaylandBufferWL = (PFNEGLQUERYWAYLANDBUFFERWL)
+		(void*) eglGetProcAddress("eglQueryWaylandBufferWL");
+	eglBindWaylandDisplayWL = (PFNEGLBINDWAYLANDDISPLAYWL)
+		(void*) eglGetProcAddress("eglBindWaylandDisplayWL");
+	eglUnbindWaylandDisplayWL = (PFNEGLUNBINDWAYLANDDISPLAYWL)
+		(void*) eglGetProcAddress("eglUnbindWaylandDisplayWL");
+
+	egl_global = egl;
+
+	egl->gl_exts = (const char*) glGetString(GL_EXTENSIONS);
 	wlr_log(L_INFO, "Using EGL %d.%d", (int)major, (int)minor);
-	wlr_log(L_INFO, "Supported EGL extensions: %s", eglQueryString(egl->display,
-		EGL_EXTENSIONS));
+	wlr_log(L_INFO, "Supported EGL extensions: %s", egl->egl_exts);
 	wlr_log(L_INFO, "Using %s", glGetString(GL_VERSION));
-	wlr_log(L_INFO, "Supported OpenGL ES extensions: %s", glGetString(GL_EXTENSIONS));
+	wlr_log(L_INFO, "Supported OpenGL ES extensions: %s", egl->gl_exts);
 	return true;
 
 error:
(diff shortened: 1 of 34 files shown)


nyorain:
btw this should fix issue number 47 (don't want to auto fix it even if this is merged) if i understood it correctly.

VincentVanlaer:
A solution to the problem of the software cursor might be to move the creation of the EGLImage and  the querying of its attributes completely to the EGL backend. The GLES2 renderer would no longer need to be created with the wlr_egl structure. I would change the arguments of the wlr_texture_upload_drm function to the buffer and the height, width and format of the buffer. This also makes the renderer less coupled with the egl backend. It might also be more sensible to then call that function wlr_texture_attach_egl.

nyorain:
decoupling renderer and egl sounds good imo but how to get the glEGLImageTargetTexture2DOES function pointer in the gles renderer then? and this would require access to the wlr_egl object when attaching a buffer to a texture (so e.g. when implementing wlr_surface) which we don't want since it is  even worse i guess.

VincentVanlaer:
It seems that you are only using it to determine whether the device supports GL_OES_EGL_image_external. That can be done in the drm and wayland backend before calling wlr_egl_bind_display.

nyorain:
sadly it's not that simple, the function pointer is used in texture.c when the drm buffer is bound. Stuff cannot work without it, this function pointer glues the EGLImageKHR to the the gl texture.

VincentVanlaer:
That's true, but I would get the function pointer regardless of whether GL_OES_EGL_image_external is supported or not and put it in a global variable like we do now with the shaders.

nyorain:
hm shifting the function pointer loading into egl and introduce a global (or something like it) would work.
Might bring new problems but probably better than the current approach.

VincentVanlaer:
I would keep the function pointer in the gles renderer, since the egl backend doesn't really need it, but I guess that doens't matter that much.

SirCmpwn:
A casual scroll through the code doesn't turn up a new shader - weston and wlc both use [a special shader](https://cgit.freedesktop.org/wayland/weston/tree/libweston/gl-renderer.c#n2446) for EGL surfaces (less casual stroll to come).

A casual test of this also shows some problems - weston-simple-egl displays but does not update on the wayland backend, and aborts on the drm backend.

SirCmpwn:
:+1: to @VincentVanlaer's suggestions

backend/egl.c
```diff
@@ -147,26 +166,83 @@ bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *display) {
...
+
+	egl->api->eglCreateImageKHR = (PFNEGLCREATEIMAGEKHRPROC)
+		eglGetProcAddress("eglCreateImageKHR");
+    egl->api->eglDestroyImageKHR = (PFNEGLDESTROYIMAGEKHRPROC)
+		eglGetProcAddress("eglDestroyImageKHR");
+    egl->api->eglQueryWaylandBufferWL = (PFNEGLQUERYWAYLANDBUFFERWL)
+		(void*) eglGetProcAddress("eglQueryWaylandBufferWL");
+    egl->api->eglBindWaylandDisplayWL = (PFNEGLBINDWAYLANDDISPLAYWL)
+		(void*) eglGetProcAddress("eglBindWaylandDisplayWL");
+    egl->api->eglUnbindWaylandDisplayWL = (PFNEGLUNBINDWAYLANDDISPLAYWL)
```
SirCmpwn: Indent with tabs

backend/egl.c
```diff
@@ -147,26 +166,83 @@ bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *display) {
...
+	if (!egl->api->eglQueryWaylandBufferWL) {
+		return false;
+	}
+
+	return egl->api->eglQueryWaylandBufferWL(egl->display, buf, attrib, value);
+}
+
+EGLImage wlr_egl_create_image(struct wlr_egl *egl, EGLenum target,
+		EGLClientBuffer buffer, const EGLint *attribs)
+{
```
SirCmpwn: Braces on the same line

backend/multi/backend.c
```diff
@@ -116,6 +116,9 @@ void wlr_multi_backend_add(struct wlr_backend *multi,
 	wl_signal_add(&backend->events.output_remove, &sub->output_remove);
 
 	list_add(multi->state->backends, sub);
+
+	if(backend->egl)
+		multi->egl = backend->egl;
```
SirCmpwn:
```c
if (backend->egl) {
\tmulti->egl = backend->egl;
}
```

cc @ascent12: this'll probably throw a bit of a wrench into your multi-GPU plans

include/backend/egl.h
```diff
@@ -2,17 +2,55 @@
...
+	const char *egl_exts;
+	const char *gl_exts;
+
+	struct wl_display *wl_display;
+	struct wlr_egl_api *api;
 };
 
-const char *egl_error(void);
+// Initializes an egl context for the given platform and remote display.
+// Will attempt to load all possibly required api functions.
```
SirCmpwn: Doc args format is

```c
/**
  * Does things to the stuff
  * And some other stuff 
  */
```

render/gles2/texture.c
```diff
@@ -10,6 +10,17 @@
 #include <wlr/render/matrix.h>
 #include <wlr/util/log.h>
 #include "render/gles2.h"
+#include "backend/egl.h"
+
+static void gles2_texture_gen_texture(struct wlr_texture_state *surface) {
+	if (surface->tex_id)
+		return;
```
SirCmpwn: Braces are mandatory

backend/egl.c
```diff
@@ -99,7 +112,7 @@ static bool egl_get_config(EGLDisplay disp, EGLConfig *out, EGLenum platform) {
 			continue;
 		}
 
-		if (gbm_format == GBM_FORMAT_ARGB8888) {
+		if (gbm_format == GBM_FORMAT_XRGB8888) {
```
ascent12: This is going to mess with transparency for the DRM hardware cursor.

SirCmpwn:
This is definitely a good start but it also definitely needs some work.

backend/egl.c
nyorain: Thanks for the notice, wrote this a while ago and did not realize that the egl code had changed in the meantime.

nyorain:
Ok, will implement @VincentVanlaer suggestion, fix style issues and use the special shader for external targets i forgot. This was meant as some sort of WIP pr so keep suggestions and issues coming.

nyorain:
Rebased it and fixed the mentioned style issues. Also removed the coupling of the gles renderer and wlr_egl, at the cost of introducing an (internal) wlr_egl global. We just have to know the egl display when attaching the drm buffer and like said before dealing with wlr_egl in the wlr_surface implementation seems like an even worse idea, but i still don't really like this.

I also added the external shader but i am not sure how to test it. The weston sample egl program does not seem to use it (since it worked before), so when exactly is it used?

weston-simple-egl works (the triangle also spins now thanks to the frame callback pr).
The compositor crashed when closing it, turned out it was a bug in the destruction listener from wlr_surface and wlr_compositor so i changed that bit a little in the last commit.
The code was written with the assumption that a wl_listener can listen to multiple signals which is not the case. The current implementation is not too beautiful again (reserving a listener and void pointer for the compositor in the wlr_surface struct) but since the whole compositor surface list should not be needed later on anyways i guess it's ok.

@SirCmpwn, You mentioned weston-simple-egl aborting on the drm backend. Somehow when using egl it is not ok to start the wayland client from another terminal (try starting it from a another tty than the one sway runs in) due to drm authentication issues. I use ```./examples/compositor & sleep 2 && weston-simple-egl``` to test the drm backend.

ascent12:
>at the cost of introducing an (internal) wlr_egl global

That's not going to work. I'm going to be using multiple EGL/OpenGL contexts for multi-GPU support in DRM. We can't really have any global state besides function pointers, and even then, I'm not 100% sure that's ok.

>I also added the external shader but i am not sure how to test it. The weston sample egl program does not seem to use it (since it worked before), so when exactly is it used?

It's not the original intention, but I will need this shader for multi-GPU as well.

VincentVanlaer:
It seems like you might have missed a part of my suggestions. I would the calls to wlr_egl_query_buffer and wlr_egl_create_image out of gles2_texture_upload_drm and into the compostior or the egl backend. Then only the compositor needs the know wlr_egl. It I recall correctly, exposing the egl object to the compositor was something that we would want to do anyway.

nyorain:
Ok, should be easy to move. But are you sure that it is a good idea to use wlr_egl from the compositor implementation? We lock the whole renderer interface to gl/egl since one of its function takes an EGLImageKHR as parameter.

SirCmpwn:
I think it makes sense for the compositor to have an egl reference.

VincentVanlaer:
Isn't wl_drm already tied to egl? If that is the case, then having EGLImageKHR as a parameter to wlr_texture_upload_drm is sensible.

SirCmpwn:
Aye.

nyorain:
ok will be moving it. Btw it's wlr_surface that will then need a referece to wlr_egl since this is where attaching takes place (and wlr_surface cannot know anything of the compositor implementation)

SirCmpwn:
I mean just change this:

```c
bool wlr_texture_upload_drm(struct wlr_texture *tex,
 	struct wl_resource *drm_buffer);
```

To this:

```c
bool wlr_texture_upload_drm(struct wlr_texture *tex,
 	struct wlr_egl *egl, struct wl_resource *drm_buffer);
```

nyorain:
ok that should work too. wlr_surface will still need a reference to wlr_egl but at least we get rid of the global.

SirCmpwn:
Nah, wait a few minutes for my next commit to see how EGL should work with wlr_surface.

SirCmpwn:
https://git.io/v7Mj9 I was going to suggest that you change wlr_surface_flush_damage to take a wlr_egl, but on second thought I think it just makes sense to initialize the wlr_surface with a wlr_egl.

nyorain:
uh, implemented it that far and then noticed again that this is not going to work. If wlr_surface needs the wlr_egl object, so does the current wl_compositor implementation. Where should it get it from in e.g. the example compositor main.c? wlr_backend would have to expose wlr_egl and we are again where i started (e.g. the multi backend mess).

nyorain:
To be honest i still think we should do it like weston and couple the gles/egl stuff tightly. Then we would not have any of this problems, it cannot be done with the current design (without introducing globals or some other ugly stuff).

SirCmpwn:
I'll take a look at this tonight and play around with it, make some decisions and push a fixup commit.

SirCmpwn:
Just resolved merge conflicts, didn't fix everything yet.

Also, trying to run wlroots inside of wlroots with this fails.

SirCmpwn:
Correction: the simple example works, the rest don't.

SirCmpwn:
Fixed nested issues

include/wlr/egl.h
```diff
@@ -10,12 +10,22 @@ struct wlr_egl {
...
+	PFNEGLBINDWAYLANDDISPLAYWL eglBindWaylandDisplayWL;
+	PFNEGLUNBINDWAYLANDDISPLAYWL eglUnbindWaylandDisplayWL;
+
 	const char *egl_exts;
 	const char *gl_exts;
 
 	struct wl_display *wl_display;
 };
 
+// TODO: Allocate and return a wlr_egl
```
SirCmpwn: Low hanging fruit if anyone wants to make this refactor happen

SirCmpwn:
c243516 makes wlr_egl public, but mostly so we can share it around more internally - it shouldn't be necessary for users to handle it. Open to feedback on this approach, but in future PRs - going to go ahead and merge this now.

Regarding multi-GPU support, this should work for the initial "all on one GPU and copied to the other" approach, but I imagine in the future we'll want to have some sort of container for multiple EGL contexts that's exposed by the backends, and a solution for matching them with respect to outputs or whatever else makes sense. Right now the multi backend just picks the first backend that provides an EGL context (it also might make sense for the initial pass on multi-GPU support to let the user pick which GPU they want doing the heavy lifting).

PR #59, opened 2017-08-10
release buffer on surface flush when no damage

acrisci:
This fixes a crash in `weston-simple-damage` when there is no damage because the damage buffer isn't implemented yet.


SirCmpwn:
Thanks!

PR #66, opened 2017-08-11
Leak plumbing

martinetd:
Three parts:
 - change libinput events to be on the stack. I went for minimal change with wlr_event = { 0 } to replace calloc, compiler should optimize it out for the fields we set right afterwards anyway -- happy to do slightly different.
 - A little bit of cleanup of our destroy functions
 - Still missing something obvious in the example code though, the remove functions are never called. Might use a hint there, not sure what is supposed to trigger these events - should our own code in the libinput backend when we destroy notify every device of a remove event?
I still need to test on a computer where I can actually unplug the mouse (e.g. not a laptop with a trackpad), not sure if it's called on unplugging either at this point...

### DIFF
--- a/backend/libinput/backend.c
+++ b/backend/libinput/backend.c
@@ -78,7 +78,19 @@ static bool wlr_libinput_backend_init(struct wlr_backend_state *state) {
 }
 
 static void wlr_libinput_backend_destroy(struct wlr_backend_state *state) {
-	// TODO
+	if (!state) {
+		return;
+	}
+
+	for (size_t i = 0; i < state->devices->length; i++) {
+		struct wlr_input_device *wlr_device = state->devices->items[i];
+
+		wlr_input_device_destroy(wlr_device);
+	}
+	list_free(state->devices);
+	libinput_unref(state->libinput);
+
+	free(state);
 }
 
 static struct wlr_backend_impl backend_impl = {
--- a/backend/libinput/events.c
+++ b/backend/libinput/events.c
@@ -24,6 +24,11 @@ struct wlr_input_device *get_appropriate_device(
 }
 
 static void wlr_libinput_device_destroy(struct wlr_input_device_state *state) {
+	list_t *devices = libinput_device_get_user_data(state->handle);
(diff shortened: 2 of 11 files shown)


backend/libinput/backend.c
```diff
@@ -78,7 +78,19 @@ static bool wlr_libinput_backend_init(struct wlr_backend_state *state) {
...
+
+	for (size_t i = 0; i < state->devices->length; i++) {
+		struct wlr_input_device *wlr_device = state->devices->items[i];
+
+		wlr_input_device_destroy(wlr_device);
+	}
+	list_free(state->devices);
+	libinput_unref(state->libinput);
+
+	free(state);
```
SirCmpwn: This function has an excessive amount of whitespace imo

backend/libinput/events.c
```diff
@@ -24,6 +24,11 @@ struct wlr_input_device *get_appropriate_device(
 }
 
 static void wlr_libinput_device_destroy(struct wlr_input_device_state *state) {
+	list_t *devices = libinput_device_get_user_data(state->handle);
+	// devices themselves are freed in wlr_libinput_backend_destroy
+	// this list only has a part of the same elements so just free list
+	list_free(devices);
```
SirCmpwn: Is this necessary at all?

SirCmpwn: Nevermind.

SirCmpwn:
Thanks!

backend/libinput/events.c
martinetd: If we don't free devices here, the devices we create line 65 (added with libinput_device_set_user_data) is never freed.

Turns out that also causes a double-free on different hardware now I'm testing on a different machine, trying to figure why before it explodes for too many others... Ah, I see, the same devices (list) can be pointed to in multiple wlr_input_device_state if the device has multiple capabilities.
Not sure how to free exactly once without tracking all the lists at this point, will have to think it through - state->devices could be a list to all the devices list but it's not particularily elegant

PR #69, opened 2017-08-11
Add user data to wlr_* structs

### LINKED ISSUE #68: Add user data storage mechanism to wlr_* types
We should let the user attach something to `wlr_*` types, especially after moving wl_shell/xdg_shell/etc into wlr.

4e554c4c:
Resolves #68

### DIFF
--- a/include/wlr/types/wlr_input_device.h
+++ b/include/wlr/types/wlr_input_device.h
@@ -41,6 +41,8 @@ struct wlr_input_device {
 		struct wlr_tablet_tool *tablet_tool;
 		struct wlr_tablet_pad *tablet_pad;
 	};
+
+	void *data;
 };
 
 #endif
--- a/include/wlr/types/wlr_keyboard.h
+++ b/include/wlr/types/wlr_keyboard.h
@@ -20,6 +20,8 @@ struct wlr_keyboard {
 	struct {
 		struct wl_signal key;
 	} events;
+
+	void *data;
 };
 
 void wlr_keyboard_led_update(struct wlr_keyboard *keyboard, uint32_t leds);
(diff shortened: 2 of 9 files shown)


types/wlr_keyboard.c
```diff
@@ -9,6 +9,7 @@ struct wlr_keyboard *wlr_keyboard_create(struct wlr_keyboard_impl *impl,
 	struct wlr_keyboard *kb = calloc(1, sizeof(struct wlr_keyboard));
 	kb->impl = impl;
 	kb->state = state;
+	kb->data = NULL;
```
SirCmpwn: This sort of thing is not necessary since we use calloc

examples/compositor/wl_compositor.c
```diff
@@ -8,7 +8,7 @@
 
 static void destroy_surface_listener(struct wl_listener *listener, void *data) {
 	struct wlr_surface *surface = wl_resource_get_user_data(data);
-	struct wl_compositor_state *state = surface->compositor_data;
```
SirCmpwn: Hold off on this change

4e554c4c: Should I create both a `compositor_data` and `data` field in wlr_surface then? I think it's harmless until wl_compositor gets merged into wlroots. Also I created #71 for this

SirCmpwn: Yes, leave it how it was before, with special fields in wlr_surface for the compositor.

types/wlr_keyboard.c
4e554c4c: Thanks, I didn't notice this

SirCmpwn:
Thanks!

PR #70, opened 2017-08-11
Fix wlr_surface invalid bitflag values

nyorain:
Seems like a copy-paste error. Fixing it before someone spends much time tracking it down when things get weird.


SirCmpwn:
Good catch.

PR #73, opened 2017-08-11
Fix libinput destroy for devices with multiple capabilities

martinetd:
Didn't find anything better than making state->devices a list of list.

Thanksfully it doesn't seem to be used for anything aside of final free, might mess with it more to handle unplug events but this fixes the crash


SirCmpwn:
Weird, but aight.

PR #74, opened 2017-08-11
More leaks

martinetd:
With that I think we're finally clean, well, ASAN/valgrind only pick a couple more unreachable regions that I didn't see how to fix (eglGetProcAddress and wl_event_loop_add_fd both allocate something and meh - we're down from 1MB of leak to around a hundred, that'll do)

libinput_event_destroy and the GLES stuff are the two big ones. I'm not too familiar with opengl but some wiki says we should always detach/free after linking, and I make a point of always trusting what I read on the internet.
Might be worth noting that the fragment_src_external program never links for me - but the error doesn't come back all the way up so this doesn't change anything at this point.


SirCmpwn:
Thanks!

PR #75, opened 2017-08-12
Fix egl image leak in gles renderer

nyorain:
Going for small fixes at the moment.
EGL images were never destroyed. And we would probably get a gl error if a wlr_texture would be created and then destroyed without ever being used since we then destroy a gl texture never created.


SirCmpwn:
Thanks!

PR #77, opened 2017-08-12
Fix external shader

nyorain:
Some pr in the last days added support for shader compile error messages that revealed the external shader does not link correctly. This fixes it.

We would have noticed it if we could test it somehow but after some research and if i understand everything correctly it seems the external image/shader stuff is only used in some uncommon situations (yuv buffers) so no idea how to test it.


SirCmpwn:
Weston might have some demo clients, but I'm not sure.

nyorain:
Could not find one. I guess it's not too important then anyways.

PR #78, opened 2017-08-12
libinput: fix segfault when handling non-wlr device removal

martinetd:
Devices is null for some devices. . . There's way too many things called device or devices in there!


SirCmpwn:
Feel free to sort it out and send a patch disambiguating the names if you like

PR #79, opened 2017-08-12
Libinput renaming

martinetd:
You said feel free, so here it is! Hopefully won't get too much in the way of anyone, might be a pain if they have changes pending in libinput.

Also sneaked in a change that makes wayland backend not hog CPU needlessly, happy to resubmit that separately if you think it is worth discussing.

copy/paste from commit message for summary:
    - backend state's 'libinput' to li_context
    - libinput_device ('device') to li_dev
    - wlr_device ('dev') to wlr_dev
    - 'devices' lists tangling of libinput devices to wlr_devices
    - 'devices' list of wlr_devices in backend state to wlr_device_lists

### DIFF
--- a/backend/libinput/backend.c
+++ b/backend/libinput/backend.c
@@ -25,20 +25,20 @@ static const struct libinput_interface libinput_impl = {
 
 static int wlr_libinput_readable(int fd, uint32_t mask, void *_backend) {
 	struct wlr_libinput_backend *backend = _backend;
-	if (libinput_dispatch(backend->libinput) != 0) {
+	if (libinput_dispatch(backend->libinput_context) != 0) {
 		wlr_log(L_ERROR, "Failed to dispatch libinput");
 		// TODO: some kind of abort?
 		return 0;
 	}
 	struct libinput_event *event;
-	while ((event = libinput_get_event(backend->libinput))) {
+	while ((event = libinput_get_event(backend->libinput_context))) {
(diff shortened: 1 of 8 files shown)


SirCmpwn:
I don't like shortening libinput to li, but device -> dev is fine.

wlr_device should not become wlr_dev (I considered and rejected this, reason being that wlr_device is more exposed to the public API and should be clearer (and not that long anyway))

Please separate that wayland backend commit into another PR.

SirCmpwn:
@Timidger we've missed you in the IRC, you gonna come hang out and help us work on wlroots?

martinetd:
> I don't like shortening libinput to li, but device -> dev is fine.

I was hesitant about it, re-changed to use libinput as full prefix all the time

> wlr_device should not become wlr_dev (I considered and rejected this, reason being that wlr_device is more exposed to the public API and should be clearer (and not that long anyway))

I'm not talking about any exposed wlr_device (no struct or function argument were changed), but the 'dev' variables inside our functions.
Happy to go with full wlr_device but I do not think there is a problem there.

> Please separate that wayland backend commit into another PR.

Done. I just can't get used to github PR and thought I could get away with that, won't try again :P

SirCmpwn:
Thanks!

SirCmpwn:
If you are not satisfied with libinput -> li, I would also merge `libinput_ctx`.

PR #83, opened 2017-08-13
Buffer damage and scaling (attempt 2)

acrisci:
Keep track of the scale of the buffer set by surface.set_buffer_scale.

Calculate the height and width of the texture from the buffer depending on the
scale when it is attached.

Use buffer damage to determine the damage of the buffer when flushing damage.

Convert surface damage to buffer damage and union to the surface buffer_damage
prior to flushing damage.

supercedes #65

### DIFF
--- a/render/gles2/texture.c
+++ b/render/gles2/texture.c
@@ -218,6 +218,18 @@ static void gles2_texture_get_matrix(struct wlr_texture *_texture,
 	wlr_matrix_mul(projection, matrix, matrix);
 }
 
+static void gles2_texture_get_buffer_size(struct wlr_texture *texture, struct
+		wl_resource *resource, int *width, int *height) {
+	struct wl_shm_buffer *buffer = wl_shm_buffer_get(resource);
+	if (!buffer) {
+		wlr_log(L_ERROR, "getting buffer size is only implemented for shm buffers");
+		return;
+	}
+
+	*width = wl_shm_buffer_get_width(buffer);
+	*height = wl_shm_buffer_get_height(buffer);
+}
+
 static void gles2_texture_bind(struct wlr_texture *_texture) {
 	struct wlr_gles2_texture *texture = (struct wlr_gles2_texture *)_texture;
 	GL_CALL(glBindTexture(GL_TEXTURE_2D, texture->tex_id));
@@ -247,6 +259,7 @@ static struct wlr_texture_impl wlr_texture_impl = {
 	.update_shm = gles2_texture_update_shm,
 	.upload_drm = gles2_texture_upload_drm,
 	.get_matrix = gles2_texture_get_matrix,
+	.get_buffer_size = gles2_texture_get_buffer_size,
 	.bind = gles2_texture_bind,
 	.destroy = gles2_texture_destroy,
 };
--- a/render/gles2/texture.c
+++ b/render/gles2/texture.c
@@ -222,7 +222,19 @@ static void gles2_texture_get_buffer_size(struct wlr_texture *texture, struct
 		wl_resource *resource, int *width, int *height) {
 	struct wl_shm_buffer *buffer = wl_shm_buffer_get(resource);
 	if (!buffer) {
-		wlr_log(L_ERROR, "getting buffer size is only implemented for shm buffers");
+		struct wlr_gles2_texture *tex = (struct wlr_gles2_texture *)texture;
+		if (!glEGLImageTargetTexture2DOES) {
+			return;
+		}
+		if (!wlr_egl_query_buffer(tex->egl, resource, EGL_WIDTH,
+				(EGLint*)&width)) {
+			wlr_log(L_ERROR, "could not get size of the buffer "
+				"(no buffer found)");
+			return;
+		};
+		wlr_egl_query_buffer(tex->egl, resource, EGL_HEIGHT,
+			(EGLint*)&height);
+
 		return;
 	}
 
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -14,8 +14,11 @@ static void surface_attach(struct wl_client *client,
 		struct wl_resource *resource,
 		struct wl_resource *buffer, int32_t sx, int32_t sy) {
 	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+	int scale = surface->current.scale;
 	surface->pending.invalid |= WLR_SURFACE_INVALID_BUFFER;
 	surface->pending.buffer = buffer;
+	surface->texture->height_from_buffer = surface->texture->height / scale;
+	surface->texture->width_from_buffer = surface->texture->width / scale;
 }
 
 static void surface_damage(struct wl_client *client,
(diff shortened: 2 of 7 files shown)


acrisci:
This one actually makes the little green ball work.

render/gles2/texture.c
```diff
@@ -207,7 +207,8 @@ static void gles2_texture_get_matrix(struct wlr_texture_state *texture,
 	wlr_matrix_identity(matrix);
 	wlr_matrix_translate(&world, x, y, 0);
 	wlr_matrix_mul(matrix, &world, matrix);
-	wlr_matrix_scale(&world, _texture->width, _texture->height, 1);
+	wlr_matrix_scale(&world, _texture->width_from_buffer,
+		_texture->height_from_buffer, 1);
```
SirCmpwn: Involving wlr_render in this seems like the wrong approach imo. Better to let the caller adjust the matrix as necessary. Maybe we shouldn't use this function at all and instead just let the caller sling matricies themselves? Or maybe we can just let the user pass in another matrix to multiply the world by.

types/wlr_surface.c
```diff
@@ -14,8 +14,11 @@ static void surface_attach(struct wl_client *client,
 		struct wl_resource *resource,
 		struct wl_resource *buffer, int32_t sx, int32_t sy) {
 	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+	int scale = surface->current.scale;
 	surface->pending.invalid |= WLR_SURFACE_INVALID_BUFFER;
 	surface->pending.buffer = buffer;
+	surface->texture->height_from_buffer = surface->texture->height / scale;
+	surface->texture->width_from_buffer = surface->texture->width / scale;
```
SirCmpwn: The variable name is also misleading imo. The width from the buffer comes from `wl_shm_buffer_get_width` (or a similar EGL thing), which is what gets encoded into `surface->texture->width`.

types/wlr_surface.c
```diff
@@ -90,23 +93,64 @@ static void surface_set_input_region(struct wl_client *client,
...
+static void wlr_surface_to_buffer_region(struct wlr_surface *surface,
+		pixman_region32_t *surface_region, pixman_region32_t *buffer_region) {
+	pixman_box32_t *src_rects, *dest_rects;
+	int nrects, i;
+	int scale = surface->current.scale;
+
+	src_rects = pixman_region32_rectangles(surface_region, &nrects);
+	dest_rects = malloc(nrects * sizeof(*dest_rects));
+	if (!dest_rects)
+		return;
```
SirCmpwn: style nit, add braces

types/wlr_surface.c
```diff
@@ -90,23 +93,64 @@ static void surface_set_input_region(struct wl_client *client,
...
+		struct wl_shm_buffer *buffer = wl_shm_buffer_get(surface->current.buffer);
+		pixman_region32_intersect_rect(&surface->current.buffer_damage,
+			&surface->current.buffer_damage, 0, 0,
+			wl_shm_buffer_get_width(buffer),
+			wl_shm_buffer_get_height(buffer));
+
 		// TODO: Surface sizing is complicated
 		//pixman_region32_intersect_rect(&surface->current.surface_damage,
 		//		&surface->current.surface_damage,
 		//		0, 0, surface->width, surface->height);
```
SirCmpwn: This change is the appropriate one to resolve this TODO in

SirCmpwn: This commented out code is just to ensure that the damage falls within the buffer size, so if you're already handling it (seems you are, but incorrectly), then this can just be removed.

types/wlr_surface.c
```diff
@@ -90,23 +93,64 @@ static void surface_set_input_region(struct wl_client *client,
...
+		wlr_surface_to_buffer_region(surface, &surface->current.surface_damage,
+			&buffer_damage);
+		pixman_region32_union(&surface->current.buffer_damage,
+			&surface->current.buffer_damage, &buffer_damage);
+
+		struct wl_shm_buffer *buffer = wl_shm_buffer_get(surface->current.buffer);
+		pixman_region32_intersect_rect(&surface->current.buffer_damage,
+			&surface->current.buffer_damage, 0, 0,
+			wl_shm_buffer_get_width(buffer),
+			wl_shm_buffer_get_height(buffer));
```
SirCmpwn: What if the buffer isn't shm? Use the `buffer->{width,height}`

SirCmpwn: We should probably have a `wlr_surface_get_effective_size` similar to `wlr_output_get_effective_size` that figures it out with respect to all of the various configuration details going on.

SirCmpwn:
LGTM overall but the sizing issues need to be handled better.

types/wlr_surface.c
acrisci: > What if the buffer isn't shm? Use the buffer->{width,height}

Did you mean texture->{width,height}? it's not attached at this point. It gets attached after the damage is flushed.

SirCmpwn: Right. Hmm...

acrisci:
Ok I think this is ready for review.

types/wlr_surface.c
```diff
@@ -91,11 +91,33 @@ static void surface_set_input_region(struct wl_client *client,
 	}
 }
 
+static void wlr_surface_get_effective_size(struct wlr_surface *surface,
+		int swidth, int sheight, int *width, int *height) {
+	int scale = surface->current.scale;
+	enum wl_output_transform transform = surface->current.transform;
+	int _width = swidth / scale;
+	int _height = sheight / scale;
```
SirCmpwn: I don't think the scale should be used here, no? This is also meant to be public, and swidth/sheight should be removed. If you need those to work with a partially committed surface, then keep this function around but add a new public function that just grabs it from current.

render/gles2/texture.c
```diff
@@ -218,6 +218,18 @@ static void gles2_texture_get_matrix(struct wlr_texture *_texture,
 	wlr_matrix_mul(projection, matrix, matrix);
 }
 
+static void gles2_texture_get_buffer_size(struct wlr_texture *texture, struct
+		wl_resource *resource, int *width, int *height) {
+	struct wl_shm_buffer *buffer = wl_shm_buffer_get(resource);
+	if (!buffer) {
+		wlr_log(L_ERROR, "getting buffer size is only implemented for shm buffers");
```
SirCmpwn: This isn't getting merged until it does DRM buffers too. Check out how the upload function handles it if you don't know how.

types/wlr_surface.c
acrisci: Yeah you need this division here for it to work correctly.

We could get this a bit nicer with a little refactoring of the surface implementation and texture interface and I think that's worth doing eventually. Maybe I'll make an issue for it.

If it were up to me, I would cache this on surface.attach. That's what the letter of the spec says to do.

SirCmpwn: Yeah, I agree. File that as an issue to deal with eventually.

SirCmpwn:
Thanks!

PR #84, opened 2017-08-14
Fixes

martinetd:
Had a few more issues

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -178,13 +178,16 @@ void wlr_output_destroy(struct wlr_output *output) {
 	wlr_texture_destroy(output->cursor.texture);
 	wlr_renderer_destroy(output->cursor.renderer);
 
-	output->impl->destroy(output);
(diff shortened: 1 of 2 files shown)


types/wlr_output.c
```diff
@@ -178,13 +178,12 @@ void wlr_output_destroy(struct wlr_output *output) {
...
 	wlr_renderer_destroy(output->cursor.renderer);
 
-	output->impl->destroy(output);
 	for (size_t i = 0; output->modes && i < output->modes->length; ++i) {
 		struct wlr_output_mode *mode = output->modes->items[i];
 		free(mode);
-		free(mode);
 	}
 	list_free(output->modes);
+	output->impl->destroy(output);
```
SirCmpwn: Wrap this in an if statement, only call it if this exists

```c
if (output->impl->destory) {
    output->impl->>destroy(output);
}
```

martinetd:
Done and added dangling else that just frees if no impl, as I'm adding to others with refactoring

SirCmpwn:
Thanks!

PR #85, opened 2017-08-14
Refactor states

### LINKED ISSUE #26: -Wlto-type-mismatch
Building wlroots with link time optimisation (-flto) causes this error to be raised by GCC.
It's being caused by the whole overloaded "wlr_backend_state" thing, where it had multiple, different definitions.

I feel we should change how all of that is handled, so we can safely build with LTO.

martinetd:
This should be all of them, plus the final cleanup (add a free call if no impl->destroy)

Good enough to close #26 ?

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -32,8 +32,10 @@ bool wlr_backend_start(struct wlr_backend *backend) {
 }
 
 void wlr_backend_destroy(struct wlr_backend *backend) {
-	if (backend->impl->destroy) {
+	if (backend->impl && backend->impl->destroy) {
 		backend->impl->destroy(backend);
+	} else {
+		free(backend);
 	}
 }
 
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -17,31 +17,33 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 		wl_fixed_t surface_y) {
 	struct wlr_input_device *dev = data;
 	struct wlr_wl_input_device *wlr_wl_dev = (struct wlr_wl_input_device *)dev;
-	assert(dev && dev->pointer && dev->pointer->state);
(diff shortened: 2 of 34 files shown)


backend/backend.c
```diff
@@ -32,8 +32,10 @@ bool wlr_backend_start(struct wlr_backend *backend) {
 }
 
 void wlr_backend_destroy(struct wlr_backend *backend) {
-	if (backend->impl->destroy) {
+	if (backend->impl && backend->impl->destroy) {
```
SirCmpwn: If you're going to start testing that impl itself is not NULL you should probably be doing it everywhere.

backend/wayland/wl_seat.c
```diff
@@ -160,46 +166,34 @@ static struct wl_keyboard_listener keyboard_listener = {
...
-static void input_device_destroy(struct wlr_input_device_state *state) {
-	wl_signal_emit(&state->backend->backend.events.input_remove, state->wlr_device);
-	if (state->resource)
-		wl_proxy_destroy(state->resource);
-	free(state);
+static void input_device_destroy(struct wlr_input_device *_dev) {
+	struct wlr_wl_input_device *dev = (struct wlr_wl_input_device *)_dev;
+	wl_signal_emit(&dev->backend->backend.events.input_remove, &dev->wlr_input_device);
+	if (dev->resource)
+		wl_proxy_destroy(dev->resource);
```
SirCmpwn: Braces required. I know this isn't your fault, but fix it anyway.

backend/backend.c
martinetd: I am. This one is in the final cleanup commit, this "need" came from the last libinput types that do not have impl at all.

SirCmpwn: Gotcha, my bad.

SirCmpwn:
Looks good aside from the minor braces issue.

SirCmpwn:
Thanks!

PR #86, opened 2017-08-14
Fix style issues

nyorain:
Inspired by the many cleanup contributions in the last days i fixed some (often self-caused) code style issues.

### DIFF
--- a/types/wlr_input_device.c
+++ b/types/wlr_input_device.c
@@ -23,7 +23,10 @@ void wlr_input_device_init(struct wlr_input_device *dev,
 }
 
 void wlr_input_device_destroy(struct wlr_input_device *dev) {
-	if (!dev) return;
+	if (!dev) {
+		return;
+	}
+	
 	if (dev->_device) {
 		switch (dev->type) {
 		case WLR_INPUT_DEVICE_KEYBOARD:
--- a/types/wlr_keyboard.c
+++ b/types/wlr_keyboard.c
@@ -11,7 +11,10 @@ void wlr_keyboard_init(struct wlr_keyboard *kb,
 }
 
 void wlr_keyboard_destroy(struct wlr_keyboard *kb) {
-	if (!kb) return;
+	if (!kb) {
+		return;
+	}
+	
 	if (kb->impl && kb->impl->destroy) {
 		kb->impl->destroy(kb);
 	} else {
--- a/types/wlr_keyboard.c
+++ b/types/wlr_keyboard.c
@@ -11,11 +11,7 @@ void wlr_keyboard_init(struct wlr_keyboard *kb,
 }
 
 void wlr_keyboard_destroy(struct wlr_keyboard *kb) {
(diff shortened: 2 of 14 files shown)


types/wlr_keyboard.c
```diff
@@ -11,7 +11,10 @@ void wlr_keyboard_init(struct wlr_keyboard *kb,
 }
 
 void wlr_keyboard_destroy(struct wlr_keyboard *kb) {
-	if (!kb) return;
+	if (!kb) {
+		return;
+	}
```
SirCmpwn: Better to update the if statement below to

```c
if (kb && kb->impl && kb->impl->destroy)
```

imo

martinetd:
(Heh, the failures are connexion timeout to pkgbuild.com... No idea how to get it to rerun but the site looks down for me as well atm)

That aside, do we want travis to run some checkpatch to catch new style issues?

nyorain:
That would be nice imo. Do you know a style-checking bot for c?

martinetd:
Not really any great fit for us - I've used checkpatch.pl (kernel's) but some rules are hard-coded for kernel style and don't fit us (e.g. they don't want optional braces like we have here, and there is no option to change that)

Some other project use astyle but I can't get it to ignore some of our quirks (like case foo: bar; break; in a single line), and it doesn't understand patch format so would need to fix everything in one go before we start using it. (it's easy to make travis complain if we have any file that changed, but not so much to tell if it was because of our patch)
I've also seen indent being used, it has a gazillion options so probably something that would match our current style.

I'm not fussy, will play with indent options for a while longer and if it's not a perfect fit I can try to hammer something into checkpatch.pl ; unless you know other alternatives?

SirCmpwn:
I investigated automated style enforcement for Sway with clang-format and found it unsuitable for our needs.

types/wlr_tablet_pad.c
```diff
@@ -13,7 +13,10 @@ void wlr_tablet_pad_init(struct wlr_tablet_pad *pad,
 }
 
 void wlr_tablet_pad_destroy(struct wlr_tablet_pad *pad) {
-	if (!pad) return;
+	if (!pad) {
+		return;
+	}
```
SirCmpwn: Merge this with the next condition

types/wlr_input_device.c
```diff
@@ -23,7 +23,10 @@ void wlr_input_device_init(struct wlr_input_device *dev,
 }
 
 void wlr_input_device_destroy(struct wlr_input_device *dev) {
-	if (!dev) return;
+	if (!dev) {
```
SirCmpwn: Merge with next condition

nyorain: not really possible in this case since this function does more stuff

SirCmpwn: My bad, the diff only expands up to the end of the if statement on GitHub.

SirCmpwn:
Thanks!

PR #87, opened 2017-08-14
Implement wayland backend cursor

### LINKED ISSUE #46: Wayland backend hardware cursor should use wayland cursors

nyorain:
Branched this off the style issue fix branch by accident, can rebase it as soon as that is merged.
Tested on weston and sway. Had to add os-compatibility.c for the shm buffer (stripped everything unneeded).

There is one (minor) issue, i had to use a void* instead of wl_buffer* in the wlr_wl_backend_output since otherwise the compiler warns (and therefore errors) about wl_buffer being deprecated (since there are wl_buffer structs in the client and server side, deprecated on server side, we actually refer to the client side definition here...). Not sure if this is a problem or if it can be handled better.
Resolves #46.

### DIFF
--- a/backend/meson.build
+++ b/backend/meson.build
@@ -22,6 +22,7 @@ backend_files = files(
   'wayland/output.c',
   'wayland/registry.c',
   'wayland/wl_seat.c',
+  'wayland/os-compatibility.c',
 )
 
 if systemd.found()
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Pushed a few style fixes, rebased out the other commit, and implemented a move_cursor that just returns true with a TODO to only return true if the supplied coordinates match the current ones.

PR #88, opened 2017-08-15
Prevent alloc errors from crashing

### LINKED ISSUE #76: alloc error handling
In some places wlroots check if an alloc function returns NULL, in others it does not (and will simply crash).
In most places when alloc returns NULL things are going to blow up anyways (or things are getting weird), i guess the best thing to do would output an error and exit.
We could remove a lot of repetitive checks and boilerplate and at the same time transform in other places a segmentation fault on alloc error into an error message and exit by introducing a macro/function that simply logs the allocation error and exits on failure and otherwise is guaranteed to return a valid pointer.
[This](https://github.com/wayland-project/weston/blob/master/shared/xalloc.h) is how weston does it, would something like this be acceptable in wlroots?

4e554c4c:
Resolves #76

### DIFF
--- a/backend/libinput/events.c
+++ b/backend/libinput/events.c
@@ -39,8 +39,10 @@ static struct wlr_input_device *allocate_device(
 	int vendor = libinput_device_get_id_vendor(libinput_dev);
 	int product = libinput_device_get_id_product(libinput_dev);
 	const char *name = libinput_device_get_name(libinput_dev);
-	struct wlr_libinput_input_device *wlr_libinput_dev =
-		calloc(1, sizeof(struct wlr_libinput_input_device));
+	struct wlr_libinput_input_device *wlr_libinput_dev;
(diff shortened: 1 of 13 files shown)


backend/libinput/events.c
```diff
@@ -108,6 +148,12 @@ static void handle_device_added(struct wlr_libinput_backend *backend,
 	} else {
 		list_free(wlr_devices);
 	}
+	return;
+
+fail:
+	wlr_log(L_ERROR, "Could not allocate new device");
+	list_foreach(wlr_devices, free);
```
SirCmpwn: This will segfault

4e554c4c: When `wlr_devices` is NULL? It won't segfault, list_foreach does sanity checks.

SirCmpwn: Eh? So it seems... not sure if I like that approach.

PR #91, opened 2017-08-15
Move wl_shell into wlroots

martinetd:
This is just a trivial move that "works" e.g. mpv will create a wl_shell surface and play in it.

BUT there is a lot wrong with the code as it was. I'm getting segfaults when destroying xdg_shell, segfaults when destroying wl_shells, segfaults accessing freed elements in examples/compositor/wl_compositor.c.... (these aren't new, I'm getting them with the old code, but I can't test what I'm doing and I don't like it)

I have more commits that fix part of them, but I need more time to understand what makes xdg_shell xdg_shell and what makes wl_shell wl_shell and what I'm free to modify (e.g. changing how we store a list of surfaces for example)
Unless we really can't move forward without this I'd rather just have another day or so to finish that

### DIFF
--- a/examples/compositor/main.c
+++ b/examples/compositor/main.c
@@ -33,6 +33,26 @@ static inline int64_t timespec_to_msec(const struct timespec *a) {
 	return (int64_t)a->tv_sec * 1000 + a->tv_nsec / 1000000;
 }
 
+void output_frame_handle_surface(struct sample_state *sample,
+		struct wlr_output *wlr_output, struct timespec *ts,
+		struct wl_resource *_res) {
+	struct wlr_surface *surface = wl_resource_get_user_data(_res);
+	float matrix[16];
+	float transform[16];
+	wlr_surface_flush_damage(surface);
+	if (surface->texture->valid) {
+		wlr_matrix_translate(&transform, 200, 200, 0);
+		wlr_surface_get_matrix(surface, &matrix,
+			&wlr_output->transform_matrix, &transform);
+		wlr_render_with_matrix(sample->renderer, surface->texture, &matrix);
+
+		struct wlr_frame_callback *cb, *cnext;
+		wl_list_for_each_safe(cb, cnext, &surface->frame_callback_list, link) {
+			wl_callback_send_done(cb->resource, timespec_to_msec(ts));
+			wl_resource_destroy(cb->resource);
+		}
+	}
+}
 void handle_output_frame(struct output_state *output, struct timespec *ts) {
 	struct compositor_state *state = output->compositor;
 	struct sample_state *sample = state->data;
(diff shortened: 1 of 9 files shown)


SirCmpwn:
>Unless we really can't move forward without this I'd rather just have another day or so to finish that

Sure, keep us up to date on your findings in the IRC. I think we can find stuff for everyone to work on regardless.

examples/compositor/main.c
```diff
@@ -39,22 +58,13 @@ void handle_output_frame(struct output_state *output, struct timespec *ts) {
...
-			wlr_render_with_matrix(sample->renderer, surface->texture, &matrix);
-
-			struct wlr_frame_callback *cb, *cnext;
-			wl_list_for_each_safe(cb, cnext, &surface->frame_callback_list, link) {
-				wl_callback_send_done(cb->resource, timespec_to_msec(ts));
-				wl_resource_destroy(cb->resource);
-			}
-		}
+	struct wlr_wl_shell_surface *wl_shell_surface;
+	wl_list_for_each(wl_shell_surface, &sample->wl_shell.surfaces, link) {
```
martinetd: I don't see how to make the list point to the wl_resource as wl_compositor.c uses the surface's wl_resource's link, so didn't add a get_wlr_wl_shell_surface_from_resource function at this point

types/wlr_xdg_shell_v6.c
```diff
@@ -178,22 +179,18 @@ static void xdg_shell_bind(struct wl_client *wl_client, void *_xdg_shell,
...
-		return NULL;
-	}
+void wlr_xdg_shell_v6_init(struct wlr_xdg_shell_v6 *xdg_shell,
+		struct wl_display *display) {
 	struct wl_global *wl_global = wl_global_create(display,
 		&zxdg_shell_v6_interface, 1, xdg_shell, xdg_shell_bind);
 	if (!wl_global) {
 		wlr_xdg_shell_v6_destroy(xdg_shell);
-		return NULL;
+		// TODO: return failure somehow
```
martinetd: Unsure about this, tempted to rename the function as create instead of make it respect init rules

types/wlr_wl_shell.c
```diff
@@ -108,10 +109,10 @@ static void wl_shell_destroy(struct wl_resource *resource) {
 	}
```
martinetd: This is one of the "browse list to remove" pattern I don't see the point of.
resource is being remove, so even if it's not part of this list it should be removed from any list it belongs to.
Not sure how to check if someone else just removed the element from the list without adding it anywhere else though, but I don't know if we should expect users to mess with this list more than just read-looping over it like we do in handle_output_frame

martinetd: github context is bad, requoting code:
```static void wl_shell_destroy(struct wl_resource *resource) {
        struct wlr_wl_shell *wl_shell = wl_resource_get_user_data(resource);
        struct wl_resource *_resource = NULL;
        wl_resource_for_each(_resource, &wl_shell->wl_resources) {
                if (_resource == resource) {
                        struct wl_list *link = wl_resource_get_link(_resource);
                        wl_list_remove(link);
                        break;
                }
        }
```

examples/compositor/main.c
SirCmpwn: wl_shell's resource list should be a list of wl_shell_surfaces, not a list of wl_surfaces. The link conflict should not be a problem.

types/wlr_wl_shell.c
SirCmpwn: But yeah we could avoid iterating over it anyway.

types/wlr_xdg_shell_v6.c
SirCmpwn: I wouldn't change this to the init pattern. Rename this to create instead.

types/wlr_wl_shell.c
martinetd: It can't belong to multiple lists, but it can be taken out and put somewhere else - which is what I'm worrying about.
Let's just say users shouldn't do that and I'll unlink directly

examples/compositor/main.c
martinetd: That's the case right now. I was referring to your comment on irc earlier, but I guess I misunderstood it then. Leaving as is for now unless you want to change something

types/wlr_wl_shell.c
SirCmpwn: Yeah, just tell users not to.

martinetd:
I don't see anything else to change here right now, review anytime please :)

SirCmpwn:
LGTM, thanks!

PR #94, opened 2017-08-16
libinput: fail if no input found on init

### LINKED ISSUE #24: Add detection of no input devices

martinetd:
This runs through events pending at init on initialization so we can
tell if some devices are available.

Note that with the way wlr_device_lists is managed, this checks that
there is at least one device we handle - it doesn't have to be a
keyboard, but there is at least a mouse or tablet_pad or something
that we care about.

Instead of failing inconditionally it might be better to leave the
decision to the user, e.g. add a "backend_has_devices" function to
call later.

(Tested by moving /dev/input off)

Fixes #24.

### DIFF
--- a/backend/libinput/backend.c
+++ b/backend/libinput/backend.c
@@ -63,14 +63,22 @@ static bool wlr_libinput_backend_start(struct wlr_backend *_backend) {
 	libinput_log_set_handler(backend->libinput_context, wlr_libinput_log);
 	libinput_log_set_priority(backend->libinput_context, LIBINPUT_LOG_PRIORITY_ERROR);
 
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #95, opened 2017-08-16
Prevent alloc errors from crashing in `list_t`

4e554c4c:
This commit changes the `list_t` api so that alloc errors can be
detected and worked around. Also fixes errors not found in 5cc7342


SirCmpwn:
Thanks!

PR #97, opened 2017-08-16
Add wlr_output_layout implementation

acrisci:
An output layout consists of a mapping of outputs to their position in a global
coordinate system that usually cooresponds to the output position in physical
space in front of the user.

Add an example that allows configuration of an output layout and demonstrates
its boundaries with a bouncing image.

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -33,7 +33,7 @@ void wlr_output_layout_add(struct wlr_output_layout *layout,
 	wl_list_insert(&layout->outputs, &layout_output->link);
 }
 
-static struct wlr_output_layout_output *wlr_output_layout_get(
+struct wlr_output_layout_output *wlr_output_layout_get(
 		struct wlr_output_layout *layout, struct wlr_output *reference) {
 	struct wlr_output_layout_output *ret = NULL;
 	struct wlr_output_layout_output *_output;
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -35,16 +35,13 @@ void wlr_output_layout_add(struct wlr_output_layout *layout,
 
 struct wlr_output_layout_output *wlr_output_layout_get(
 		struct wlr_output_layout *layout, struct wlr_output *reference) {
-	struct wlr_output_layout_output *ret = NULL;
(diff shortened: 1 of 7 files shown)


types/wlr_output_layout.c
```diff
@@ -0,0 +1,104 @@
...
+			}
+		}
+	}
+
+	return ret;
+}
+
+void wlr_output_layout_move(struct wlr_output_layout *layout,
+    struct wlr_output *output, int x, int y) {
+	struct wlr_output_layout_output *layout_output =
```
nyorain: The style here is inconsistent. I believe it should be as above: 2 tabs for the second argument line and 1 tab for the function body.

SirCmpwn: Aye, @nyorain is correct.

acrisci:
Ok I think this is good.

acrisci:
The problem with running this in the wayland backend under sway is that when the output is added, the advertised width and height is that of the sway container without a border. This is the time that I add the output to the layout. The layout is arranged so that they are lined up from left to right.

Once the window is managed by sway, a border is added between the containers which is about 4px. This creates a gap between the outputs in the layout. You can prevent this with by setting `border none`.

If the collision detection in the example worked perfectly, the image should not cross the screen in this case. I do not believe it is the correct behavior to adjust the x and y coordinates when an output is resized. Width and height will adjust appropriately in the current implementation.

Right now, I am working to improve the collision detection in the example for the correct behavior described above, and I will add some more helper methods. Also, I need to handle the case of overlapping outputs in the layout which is very tricky.

SirCmpwn:
When the width/height changes, you should probably invalidate the layout and recalculate it. Maybe keep track of outputs that are manually configured versus automatically laid out and recalculate with respect to that.

SirCmpwn:
Overlapping outputs should just render the same content multiple times.

acrisci:
ok I have massively improved the output-layout example.

SirCmpwn:
Tested it and it works much better, in every situation I tried it in. Great work!

SirCmpwn:
Rebased and pushed a cleanup+fixup commit that should fix the build. Will merge when the build is good.

SirCmpwn:
Thanks!

PR #98, opened 2017-08-16
Wayland backend xdg

martinetd:
A few notes:
 - I made a new list for protos to only generate some as client headers, we might want to always generate them and simplify that.
 - as described in 2nd commit message, failing to create any display does not fail the backend, it probably should.
 - I didn't find how to set a class, closest I could find was app_id. Should I? (there is a comment in the code)

Rest looks like it works ok.

### DIFF
--- a/backend/wayland/output.c
+++ b/backend/wayland/output.c
@@ -11,6 +11,7 @@
 #include <wlr/interfaces/wlr_output.h>
 #include <wlr/util/log.h>
 #include "backend/wayland.h"
+#include "xdg-shell-unstable-v6-client-protocol.h"
 
 int os_create_anonymous_file(off_t size);
 
@@ -136,7 +137,7 @@ static void wlr_wl_output_destroy(struct wlr_output *_output) {
 	}
 	eglDestroySurface(output->backend->egl.display, output->surface);
 	wl_egl_window_destroy(output->egl_window);
-	wl_shell_surface_destroy(output->shell_surface);
+	// xdg_surface/toplevel destroy
 	wl_surface_destroy(output->surface);
 	free(output);
 }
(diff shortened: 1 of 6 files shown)


backend/wayland/output.c
```diff
@@ -216,16 +231,30 @@ struct wlr_output *wlr_wl_output_create(struct wlr_backend *_backend) {
...
+		goto error;
+	}
 
-	wl_shell_surface_set_class(output->shell_surface, "sway");
-	wl_shell_surface_set_title(output->shell_surface, "sway-wl");
-	wl_shell_surface_add_listener(output->shell_surface,
-			&shell_surface_listener, output);
-	wl_shell_surface_set_toplevel(output->shell_surface);
+	// class? app_id?
+	zxdg_toplevel_v6_set_title(output->xdg_toplevel, "sway-wl");
```
SirCmpwn: Don't set it to sway-wl, set it to wlroots. app_id too.

backend/wayland/registry.c
```diff
@@ -4,6 +4,18 @@
 #include <wayland-client.h>
 #include <wlr/util/log.h>
 #include "backend/wayland.h"
+#include "xdg-shell-unstable-v6-client-protocol.h"
+
+
+static void xdg_shell_handle_ping(void *data, struct zxdg_shell_v6 *shell, uint32_t serial)
+{
+        zxdg_shell_v6_pong(shell, serial);
```
SirCmpwn: style

nyorain:
If there is a wayland compositor (i.e. WAYLAND_DISPLAY set in the current env) but it does not implement xdg shell it's ok to simply fail IMO, using the drm backend will probably not work anyways in this case (since there is a wayland compositor).
But we could at least add an error log (i don't think there is currently one) e.g. simply check in wlr_wl_registry_poll if all globals we need (and assume to be present later on) are present and output an error (or warning depending on the importance of the global) if it is not there so we know at least what went wrong from looking into a log.

backend/wayland/output.c
martinetd: Hey don't blame me for the old code! I mean, done; and fixed style as well.

nyorain:
Oh and this does not work for on weston.
It outputs 
```
zxdg_surface_v6@12: error 3: xdg_surface has never been configured
```
and does not show anything. 
xdg_surface is a pain to implement correctly for all compositors.
We should test this on all major compositors.

martinetd:
Ah, only tested with sway. Will test others.. But not sure what to change, I used weston-terminal as xdg example heh. Will have a look later.

martinetd:
Soo.. the way wayland works is that their xdg shell struct ('window') has a inhibit_redraw flag, and they're setting that to inhibit on shell surface creation and back to 0 on xdg_surface_handle_configure...
I guess we're going to need something similar :/

Didn't look far but gnome wayland seems to have a similar error ("wl_surface@10: error 3: buffer committed to unconfigred xdg_surface")

martinetd:
hm I'm actually confusing our test compositor code with that, no clue where in the wayland compositor output code we say that we want draw the surface. Will try to figure what exactly prints that message and see if we can delay it somehow.

martinetd:
Haven't/not able to test kde or enlightenment, but it's finally starting to shape up.
Regarding fail, we could fail in wlr_wl_registry_poll but does having the global guarantee that it will work later? I'd rather check if an output has been init in attempt_wl_backend really.
The problem as I said is that this function is called twice during our init, and the first time will not set up outputs (it's just counting the number of requested outputs the first time) so I need a way to tell which pass it is.

SirCmpwn:
Thanks!

PR #99, opened 2017-08-17
wlr_seat

nyorain:
This fixes the remaining seat issues and adds at least basic keyboard support to the example compositor.
It will simply focus the first surface and send keyboard events if it has a keyboard bound (works/tested with weston-terminal). The example is overall a bit hacky but pretty much all of it will probably change when we implement the shell correctly (and therefore focus semantics).

I fixed up the destroy/release handling, the previous implementation was a bit problematic. The release function is part of the interface, it's pretty much useless all we can do there is destroy the resource.
The resource destructor is guaranteed to be called (release is not) so we perform all cleanup there now.

I'll comment on a few places in a diff that might need explanation.

### DIFF
--- a/types/wlr_seat.c
+++ b/types/wlr_seat.c
@@ -6,6 +6,11 @@
 #include <wlr/types/wlr_seat.h>
 #include <wlr/util/log.h>
 
+static void resource_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
 static void wl_pointer_set_cursor(struct wl_client *client,
 			   struct wl_resource *resource,
 			   uint32_t serial,
@@ -15,29 +20,22 @@ static void wl_pointer_set_cursor(struct wl_client *client,
 	wlr_log(L_DEBUG, "TODO: wl_pointer_set_cursor");
 }
 
+static const struct wl_pointer_interface wl_pointer_impl = {
+	.set_cursor = wl_pointer_set_cursor,
+	.release = resource_destroy
+};
+
 static void wl_pointer_destroy(struct wl_resource *resource) {
 	struct wlr_seat_handle *handle = wl_resource_get_user_data(resource);
 	if (handle->pointer) {
-		wl_resource_destroy(handle->pointer);
 		handle->pointer = NULL;
 	}
 }
 
-static void wl_pointer_release(struct wl_client *client,
-		struct wl_resource *resource) {
-	wl_pointer_destroy(resource);
-}
-
-struct wl_pointer_interface wl_pointer_impl = {
-	.set_cursor = wl_pointer_set_cursor,
-	.release = wl_pointer_release
-};
-
 static void wl_seat_get_pointer(struct wl_client *client,
 		struct wl_resource *_handle, uint32_t id) {
 	struct wlr_seat_handle *handle = wl_resource_get_user_data(_handle);
-	if (!(handle->wlr_seat->capabilities & WL_SEAT_CAPABILITY_KEYBOARD)) {
-		// TODO: Show error?
+	if (!(handle->wlr_seat->capabilities & WL_SEAT_CAPABILITY_POINTER)) {
 		return;
 	}
 	if (handle->pointer) {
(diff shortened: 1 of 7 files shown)


types/wlr_seat.c
```diff
@@ -67,33 +115,28 @@ static void wl_seat_destroy(struct wl_resource *resource) {
...
-	.release = wl_seat_release
+	.release = resource_destroy
 };
 
 static void wl_seat_bind(struct wl_client *wl_client, void *_wlr_seat,
 		uint32_t version, uint32_t id) {
 	struct wlr_seat *wlr_seat = _wlr_seat;
 	assert(wl_client && wlr_seat);
-	if (version > 5) {
+	if (version > 6) {
```
nyorain: Was it intentional to only implement version 5? There is already version 6 of wl_seat (adds a pointer function or something iirc)

types/wlr_seat.c
```diff
@@ -146,11 +196,18 @@ struct wlr_seat_handle *wlr_seat_handle_for_client(struct wlr_seat *wlr_seat,
 
 void wlr_seat_set_capabilities(struct wlr_seat *wlr_seat,
 		uint32_t capabilities) {
-	// TODO: If e.g. pointer was removed, destroy all client pointer resources
```
nyorain: The wayland spec says we simply update the capabilities. The client must then destroy the resource. Simply destroying a resource on our side is usually not a good idea.

types/wlr_seat.c
```diff
@@ -46,19 +44,69 @@ static void wl_seat_get_pointer(struct wl_client *client,
...
 
 static void wl_seat_get_keyboard(struct wl_client *client,
 		struct wl_resource *_handle, uint32_t id) {
-	wlr_log(L_DEBUG, "TODO: wl_seat_get_keyboard");
+	struct wlr_seat_handle *handle = wl_resource_get_user_data(_handle);
+	if (!(handle->wlr_seat->capabilities & WL_SEAT_CAPABILITY_KEYBOARD)) {
+		return;
+	}
+	if (handle->keyboard) {
+		// TODO: this is probably a protocol violation but it simplifies our
```
nyorain: I added this note to the other functions. It is indeed problematic, weston keeps a list of keyboard/pointer/touch resources. But this can wait for later i guess since it complicates things that have no real use case (as stated).

types/wlr_seat.c
```diff
@@ -15,29 +20,22 @@ static void wl_pointer_set_cursor(struct wl_client *client,
...
-struct wl_pointer_interface wl_pointer_impl = {
-	.set_cursor = wl_pointer_set_cursor,
-	.release = wl_pointer_release
-};
-
 static void wl_seat_get_pointer(struct wl_client *client,
 		struct wl_resource *_handle, uint32_t id) {
 	struct wlr_seat_handle *handle = wl_resource_get_user_data(_handle);
-	if (!(handle->wlr_seat->capabilities & WL_SEAT_CAPABILITY_KEYBOARD)) {
-		// TODO: Show error?
```
nyorain: There is no error code for this and making up own codes isn't a thing i guess. If i did not oversee something weston does not do it either.

types/wlr_seat.c
```diff
@@ -0,0 +1,213 @@
...
+		struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static void wl_pointer_set_cursor(struct wl_client *client,
+			   struct wl_resource *resource,
+			   uint32_t serial,
+			   struct wl_resource *surface,
+			   int32_t hotspot_x,
+			   int32_t hotspot_y) {
```
SirCmpwn: Indented incorrectly

types/wlr_seat.c
```diff
@@ -0,0 +1,213 @@
...
+	if (handle->pointer) {
+		// TODO: this is probably a protocol violation but it simplifies our
+		// code and it'd be stupid for clients to create several pointers for
+		// the same seat
+		wl_resource_destroy(handle->pointer);
+	}
+	handle->pointer = wl_resource_create(client, &wl_pointer_interface,
+		wl_resource_get_version(_handle), id);
+	wl_resource_set_implementation(handle->pointer, &wl_pointer_impl,
+			handle, &wl_pointer_destroy);
```
SirCmpwn: Indented incorrectly

types/wlr_seat.c
```diff
@@ -0,0 +1,213 @@
...
+	if (handle->keyboard) {
+		// TODO: this is probably a protocol violation but it simplifies our
+		// code and it'd be stupid for clients to create several pointers for
+		// the same seat
+		wl_resource_destroy(handle->keyboard);
+	}
+	handle->keyboard = wl_resource_create(client, &wl_keyboard_interface,
+		wl_resource_get_version(_handle), id);
+	wl_resource_set_implementation(handle->keyboard, &wl_keyboard_impl,
+			handle, &wl_keyboard_destroy);
```
SirCmpwn: .

types/wlr_seat.c
```diff
@@ -0,0 +1,213 @@
...
+	if (handle->touch) {
+		// TODO: this is probably a protocol violation but it simplifies our
+		// code and it'd be stupid for clients to create several pointers for
+		// the same seat
+		wl_resource_destroy(handle->touch);
+	}
+	handle->touch = wl_resource_create(client, &wl_touch_interface,
+		wl_resource_get_version(_handle), id);
+	wl_resource_set_implementation(handle->touch, &wl_touch_impl,
+			handle, &wl_touch_destroy);
```
SirCmpwn: .

types/wlr_seat.c
SirCmpwn: Maybe we should add a client_assert helper that behaves like assert but kills the client on error instead of aborting.

SirCmpwn:
I tried this and control doesn't seem to be sent through to weston-terminal.

nyorain:
No modifiers are currently sent (shift won't work either).
Wasn't sure when/how to do this. Might require an additional callback.
Will figure it out then.

SirCmpwn:
Nice, good work!

Issue #100, opened 2017-08-17
gnome-calculator causes the example compositor to segfault


SirCmpwn:
```
[render/gles2/texture.c:136] GL error 1281 Invalid value
```

Shortly before the crash.

nyorain:
It does not crash for me but simply not show anything.
Read the [glTexSubImage2D spec](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexSubImage2D.xhtml) it can only be the third reason for invalid value.
I guess this occurs since we use update_shm too often. If e.g. the buffer size changes we must use upload_shm instead.
I could come up with this small fix that makes gnome-calculator appear for me.
But this is rather a hack at the moment and there may be other related issues not fixed by this.
We should carefully check whether we can really use update_shm. Maybe such a check should take place in the texture implementation instead?

```diff
diff --git a/include/wlr/types/wlr_surface.h b/include/wlr/types/wlr_surface.h
index 9b13d9c..f076516 100644
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -38,6 +38,7 @@ struct wlr_surface {
 
 	float buffer_to_surface_matrix[16];
 	float surface_to_buffer_matrix[16];
+	bool reupload_buffer;
 
 	struct {
 		struct wl_signal commit;
diff --git a/types/wlr_surface.c b/types/wlr_surface.c
index 50de2e4..fd37c35 100644
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -96,7 +96,7 @@ static void wlr_surface_update_size(struct wlr_surface *surface) {
 	enum wl_output_transform transform = surface->current.transform;
 
 	wlr_texture_get_buffer_size(surface->texture, surface->current.buffer,
-		&surface->current.buffer_width, &surface->current.buffer_height);
+		&surface->pending.buffer_width, &surface->current.buffer_height);
 
 	int _width = surface->current.buffer_width / scale;
 	int _height = surface->current.buffer_height / scale;
@@ -206,8 +206,13 @@ static void surface_commit(struct wl_client *client,
 		surface->current.buffer = surface->pending.buffer;
 	}
 	if ((surface->pending.invalid & WLR_SURFACE_INVALID_SURFACE_DAMAGE)) {
+		int32_t oldw = surface->current.buffer_width;
+		int32_t oldh = surface->current.buffer_height;
 		wlr_surface_update_size(surface);
 
[...]

PR #101, opened 2017-08-17
Fix surface buffer uploading

nyorain:
The patch from #100.
I only tested this with the wayland backend (where it works now), gnome-calculator seems to still
crash on the drm backend i noticed (due to something else though).
There is no gl error anymore.
I would be still ok with not merging this yet, maybe someone else comes up with something better (or finds the new reason for the crash).

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -38,6 +38,7 @@ struct wlr_surface {
 
 	float buffer_to_surface_matrix[16];
 	float surface_to_buffer_matrix[16];
+	bool reupload_buffer;
 
 	struct {
 		struct wl_signal commit;
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -206,7 +206,12 @@ static void surface_commit(struct wl_client *client,
 		surface->current.buffer = surface->pending.buffer;
 	}
 	if ((surface->pending.invalid & WLR_SURFACE_INVALID_SURFACE_DAMAGE)) {
+		int32_t oldw = surface->current.buffer_width;
+		int32_t oldh = surface->current.buffer_height;
 		wlr_surface_update_size(surface);
+
+		surface->reupload_buffer = oldw != surface->current.buffer_width ||
+			oldh != surface->current.buffer_height;
 
 		pixman_region32_union(&surface->current.surface_damage,
 			&surface->current.surface_damage,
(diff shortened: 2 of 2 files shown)


acrisci:
I'm still getting a lot of errors like this

```
2017-08-17 12:11:36 - [types/wlr_surface.c:261] Unknown buffer handle attached
2017-08-17 12:11:36 - [types/wlr_surface.c:261] Unknown buffer handle attached
2017-08-17 12:11:36 - [types/wlr_surface.c:261] Unknown buffer handle attached
```

gnome-calculator runs for about 10 seconds then dies like this

```
(gnome-calculator:17257): Gdk-CRITICAL **: gdk_seat_get_keyboard: assertion 'GDK_IS_SEAT (seat)' failed

(gnome-calculator:17257): Gdk-CRITICAL **: gdk_seat_get_keyboard: assertion 'GDK_IS_SEAT (seat)' failed

(gnome-calculator:17257): Gdk-CRITICAL **: gdk_seat_get_keyboard: assertion 'GDK_IS_SEAT (seat)' failed

(gnome-calculator:17257): Gdk-WARNING **: (gdkwindow-wayland.c:791):buffer_release_callback: runtime check failed: (impl->staging_cairo_surface != cairo_surface)
Gdk-Message: Error reading events from display: Broken pipe
```

nyorain:
That is probably a different issue though.
I am still not too familiar with the wlr_surface implementation but i figured it may happen because wlr_surface_flush_damage is called multiple times on the same buffer, i.e. on a buffer that was already released?

SirCmpwn:
Surface damage is cleared when you flush, so I doubt it.

nyorain:
But the buffer is still accessed (e.g. wl_shm_buffer_get) in every call.
And the release event is sent again in every call (which seems to trigger the failure assertion).
This is a protocol violation (a rather bad one) we have to not access the buffer anymore in any way
after we send the release event once (and we must not send it multiple times).
Not sure though if this causes gnome-calculator to fail on the drm backend.

nyorain:
And also not sure how to fix it. Either store if have already released the buffer or redesign the whole damage flushing concept.

SirCmpwn:
We could just not access the buffer unless there's damage? FTR Weston has a similar flush procedure.

SirCmpwn:
Let's figure out more improvements separately.

PR #105, opened 2017-08-18
fix uninitialized warning

acrisci:
`index` might be uninitialized. Warning can be produced with gcc and `-01` flag.


martinetd:
That's probably not the behavior we want, I'd say we should init to -1 and fail if it didn't get set.
It really has no reason not to find one, maybe just assert..

SirCmpwn:
Thanks!

SirCmpwn:
Send a follow up PR if you guys come to an agreement I guess.

martinetd:
This definitely isn't worse than before, but it's just sweeping it under the rug when we can do better :)

I'll post a follow-up that just assert we found one.

PR #106, opened 2017-08-18
Fix wild pointer to buffer

acrisci:
After the buffer is released from the release event, it should no longer be
accessed.

Fixes a GTK crash.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -293,6 +293,7 @@ release:
 	pixman_region32_init(&surface->current.buffer_damage);
 
 	wl_resource_queue_event(surface->current.buffer, WL_BUFFER_RELEASE);
+	surface->current.buffer = NULL;
 }
 
 static void surface_set_buffer_transform(struct wl_client *client,


acrisci:
This demonstrates the problem but might not be the best solution.

I think Weston is setting the buffer reference to NULL in a destroy callback. I think this would require us to have a buffer abstraction with a destroy listener so we can get the container.

types/wlr_surface.c
```diff
@@ -293,6 +293,7 @@ void wlr_surface_flush_damage(struct wlr_surface *surface) {
 	pixman_region32_init(&surface->current.buffer_damage);
 
 	wl_resource_queue_event(surface->current.buffer, WL_BUFFER_RELEASE);
+	surface->current.buffer = NULL;
```
nyorain: Probably ok for a temporary workaround.
But keep in mind that this will lead to issues as soon as we implement the TODO at the top of the function, i.e. empty the texture when there was a null buffer attached. It would be triggered from this i guess (when damage_flush is called multiple times).

SirCmpwn:
Thanks!

types/wlr_surface.c
SirCmpwn: ack

PR #107, opened 2017-08-18
Implement wlr_data_{source,device,device_manager}

nyorain:
This implements the most important interfaces for data exchange.
It does more than just stub functions, it basically finishes all implementations already.
wl_data_offer is left completely undone since most of it will be handled by the user and it will only be needed when we want to get clipboard to work. I could therefore also only test this in a theoretical way (creating and setting the selection using weston-terminal) which works.

A little explanation since there may be a few unexpected places and this is handled different (better) than wlc does it:
- it does already correctly implement multi-seat support
   - since wl_data_device is basically the child of wl_data_device_manager and wl_seat those resources are added to the seat_handle struct. This might sound weird first but it is closely connected to it. When a data_device start a dnd operation it grabs the seat. When the keyboard resource is focused, the user will send a data offer using the associated data device
   - it associates the wlr_data_device directly with a wlr_seat to make this work
- it currently suffers from the same simplification as wlr_seat: there could theoretically be multiple wl_data_device resources for one client and seat but we ignore it at the moment (that should be fixed as well when it is fixed on day for the other seat resources) 
- wlr_data_source is a virtual interface to allow stuff like x11 data sources later on or allow the user to set their own data sources (where ever they get them from, like e.g. in sway the set_clipboard functionality)

With this architecture it should be pretty easy to implement dnd later on (basically only surface roles + seat grab).

### DIFF
--- a/examples/compositor/main.c
+++ b/examples/compositor/main.c
@@ -17,6 +17,7 @@
 #include <wlr/types/wlr_wl_shell.h>
 #include <wlr/types/wlr_xdg_shell_v6.h>
 #include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_data_device_manager.h>
 #include <xkbcommon/xkbcommon.h>
 #include <wlr/util/log.h>
 #include "shared.h"
@@ -31,6 +32,7 @@ struct sample_state {
 	struct wlr_wl_shell *wl_shell;
 	struct wlr_seat *wl_seat;
 	struct wlr_xdg_shell_v6 *xdg_shell;
+	struct wlr_data_device_manager *data_device_manager;
 	struct wl_resource *focus;
 	struct wl_listener keyboard_bound;
 	int keymap_fd;
@@ -130,6 +132,9 @@ static void handle_keyboard_bound(struct wl_listener *listener, void *data) {
 
 	wl_keyboard_send_keymap(handle->keyboard, WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1,
 		state->keymap_fd, state->keymap_size);
+
+	if (wl_resource_get_version(handle->keyboard) >= 2)
+		wl_keyboard_send_repeat_info(handle->keyboard, 660, 25);
 }
 
 int main() {
@@ -150,6 +155,7 @@ int main() {
 	wl_compositor_init(compositor.display, &state.compositor, state.renderer);
 	state.wl_shell = wlr_wl_shell_create(compositor.display);
 	state.xdg_shell = wlr_xdg_shell_v6_create(compositor.display);
+	state.data_device_manager = wlr_data_device_manager_create(compositor.display);
 
 	state.wl_seat = wlr_seat_create(compositor.display, "seat0");
 	state.keyboard_bound.notify = handle_keyboard_bound;
--- a/examples/compositor/main.c
+++ b/examples/compositor/main.c
@@ -133,8 +133,9 @@ static void handle_keyboard_bound(struct wl_listener *listener, void *data) {
 	wl_keyboard_send_keymap(handle->keyboard, WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1,
 		state->keymap_fd, state->keymap_size);
 
-	if (wl_resource_get_version(handle->keyboard) >= 2)
-		wl_keyboard_send_repeat_info(handle->keyboard, 660, 25);
+	if (wl_resource_get_version(handle->keyboard) >= 2) {
+		wl_keyboard_send_repeat_info(handle->keyboard, 25, 600);
+	}
 }
 
 int main() {
--- a/types/wlr_data_device_manager.c
+++ b/types/wlr_data_device_manager.c
@@ -1,0 +1,185 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wayland-server.h>
+#include <wlr/util/log.h>
+#include <wlr/types/wlr_data_device_manager.h>
+#include <wlr/types/wlr_data_source.h>
+#include <wlr/types/wlr_seat.h>
+
+static void resource_destroy(struct wl_client *client, struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static void wl_cb_data_device_start_drag(struct wl_client *client,
+		struct wl_resource *res, struct wl_resource *src_res,
+		struct wl_resource *origin_res, struct wl_resource *icon_res,
+		uint32_t serial) {
+	wlr_log(L_DEBUG, "implement data_device:start_drag");
+
+	// Will probably look like this:
+	// struct wlr_seat_handle *handle = wl_resource_get_user_data(res);
+	// struct wlr_data_device *device = handle->wlr_seat->data_device;
+	// struct wlr_data_source *src = wl_resource_get_user_data(src_res);
+	// struct wlr_surface *origin = wl_resource_get_user_data(origin_res);
+	// struct wlr_surface *icon;
+	// if (icon_res)
+	// 	icon = wl_resource_get_user_data(icon_res);
+	// wlr_seat_start_drag(serial, device->seat, src,
+	// 	origin, icon); // will set surface roles and emit signal for user
+}
+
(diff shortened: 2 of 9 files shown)


nyorain:
Ah and it does also fix a small bug in the example compositor, forgot to send a repeat rate which seemed to create issues with gtk sometimes. I am using now simply the default wlc repeat rates there.

martinetd:
Haven't looked at code yet, but use-after-free when I closed a client:
```
==31774==ERROR: AddressSanitizer: heap-use-after-free on address 0x6060000bff88 at pc 0x7f4711c71ee9 bp 0x7ffd1de488b0 sp 0x7ffd1de488a0
WRITE of size 8 at 0x6060000bff88 thread T0
    #0 0x7f4711c71ee8 in data_device_destroy ../types/wlr_data_device_manager.c:74
    #1 0x7f4711a19f7f  (/lib64/libwayland-server.so.0+0x7f7f)
    #2 0x7f4711a1e808 in wl_map_for_each (/lib64/libwayland-server.so.0+0xc808)
    #3 0x7f4711a1a06c in wl_client_destroy (/lib64/libwayland-server.so.0+0x806c)
    #4 0x7f4711a1a127  (/lib64/libwayland-server.so.0+0x8127)
    #5 0x7f4711a1bc51 in wl_event_loop_dispatch (/lib64/libwayland-server.so.0+0x9c51)
    #6 0x7f4711a1a499 in wl_display_run (/lib64/libwayland-server.so.0+0x8499)
    #7 0x40821a in compositor_run ../examples/shared.c:535
    #8 0x404063 in main ../examples/compositor/main.c:180
    #9 0x7f47114214d9 in __libc_start_main (/lib64/libc.so.6+0x204d9)
    #10 0x4026b9 in _start (/home/shared/wayland/wlroots/build/examples/compositor+0x4026b9)

0x6060000bff88 is located 40 bytes inside of 64-byte region [0x6060000bff60,0x6060000bffa0)
freed by thread T0 here:
    #0 0x7f4711f894b8 in __interceptor_free (/lib64/libasan.so.4+0xde4b8)
    #1 0x7f4711c76e34 in wl_seat_destroy ../types/wlr_seat.c:125
    #2 0x7f4711a19f7f  (/lib64/libwayland-server.so.0+0x7f7f)

    #0 0x7f4711f89a38 in __interceptor_calloc (/lib64/libasan.so.4+0xdea38)
    #1 0x7f4711c76ed7 in wl_seat_bind ../types/wlr_seat.c:144
    #2 0x7f470efeebdd in ffi_call_unix64 (/lib64/libffi.so.6+0x5bdd)

SUMMARY: AddressSanitizer: heap-use-after-free ../types/wlr_data_device_manager.c:74 in data_device_destroy
  0x0c0c8000ffa0: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
  0x0c0c8000ffb0: fd fd fd fd fd fd fd fd fa fa fa fa fd fd fd fd
  0x0c0c8000ffc0: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c8000ffd0: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
  0x0c0c8000ffe0: fd fd fd fd fd fd fd fd fa fa fa fa fd fd fd fd
=>0x0c0c8000fff0: fd[fd]fd fd fa fa fa fa 00 00 00 00 00 00 00 fa
  0x0c0c80010000: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0c80010010: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0c80010020: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0c80010030: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0c80010040: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==31774==ABORTING
```


handle was freed/destroyed before the data_device it looks like.

nyorain:
Thanks! I hope it's fixed now

SirCmpwn:
What's the best way to test this?

examples/compositor/main.c
```diff
@@ -130,6 +132,9 @@ static void handle_keyboard_bound(struct wl_listener *listener, void *data) {
 
 	wl_keyboard_send_keymap(handle->keyboard, WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1,
 		state->keymap_fd, state->keymap_size);
+
+	if (wl_resource_get_version(handle->keyboard) >= 2)
+		wl_keyboard_send_repeat_info(handle->keyboard, 660, 25);
```
SirCmpwn: I shouldn't have to tell you about style at this point

types/wlr_data_device_manager.c
```diff
@@ -0,0 +1,185 @@
...
+		struct wl_resource *res, uint32_t id, struct wl_resource *seat_res) {
+	struct wlr_data_device_manager *manager = wl_resource_get_user_data(res);
+	struct wlr_seat_handle *seat_handle = wl_resource_get_user_data(seat_res);
+	struct wlr_data_device *device;
+	if (!(device = seat_ensure_data_device(manager, seat_handle->wlr_seat))) {
+		wl_resource_post_no_memory(res);
+		return;
+	}
+
+	if (seat_handle->data_device) {
```
SirCmpwn: We should figure out at some point a good way of adding state to objects without their involvement.

types/wlr_data_device_manager.c
```diff
@@ -0,0 +1,185 @@
...
+}
+
+void wlr_data_device_manager_destroy(struct wlr_data_device_manager *manager) {
+	if (!manager) {
+		return;
+	}
+
+	// TODO: destroy remaining resources? cancel current selection?
+	//  if this is called why there are still resources active we will
+	//  always get problems
```
SirCmpwn: Yeah tear it all down here

SirCmpwn:
LGTM overall codewise

martinetd:
Looks good to me as well. This fixes the crash alright.
There's a couple of new leaks but there was a lot of other new ones while I wasn't looking (probably just gtk app working now so I never had a chance to see them), so I'll do another pass on these tomorrow :P

types/wlr_data_device_manager.c
nyorain: I don't even think it would make sense it this case. As tried to describe above data_device and seat are really closely connected.

SirCmpwn: Only from data_device's perspective, though.

nyorain: hm, right. But the user has to retrieve its handle from the seat e.g. when the keyboard is focused. I guess this can wait for later?

types/wlr_data_device_manager.c
nyorain: oh this todo is actually wrong, data_device_manager does not manage any selections anymore (was moved with multi seat support). Will make sure to move/fix it.

nyorain:
The way i tested this was simply adding a logging call to the end of wl_cb_data_device_set_selection and then use weston-terminal ctrl+c with an empty selection (which already sets a selection).
Since this does not implement the offer side yet that's probably all we can test.

types/wlr_data_device_manager.c
SirCmpwn: Yeah, it can wait.

nyorain:
Btw this fixes now also resource leaks when a wlr_seat is destroyed.
But as mentioned in this one TODO things will probably not end well if we simply destruct a global (and its compositor-side representation) that still has resources bound (and then continue running the wl_display). We should at least add a warning/documentation somewhere that this will result in problems.

SirCmpwn:
Thanks!

PR #108, opened 2017-08-18
drm realloc_crtcs: check we found an output

martinetd:
This follows-up on #105 that just initialized to 0.
Instead, make sure we actually found an output that matches


SirCmpwn:
Thanks!

Issue #102, opened 2017-08-18
Performance issues


SirCmpwn:
I get some pretty bad stuttering with mpv playing 1080p video on the example compositor using any backend. We should profile this and figure it out.

martinetd:
Been looking at that bit, this really isn't a cpu/performance problem, but there's something inconsistant somewhere.

Sorry for long logs ahead, feel free to skip through, this is just me thinking aloud - just skip to the end for the important part/question.

Looking with WAYLAND_DEBUG=1 on both mpv and our test compositor, I have this kind of logs on mpv:
```[1268950.206]  -> wl_surface@3.attach(wl_buffer@19, 0, 0)
[1268950.806]  -> wl_surface@3.damage(0, 0, 2147483647, 2147483647)
[1268951.545]  -> wl_surface@3.commit()
[1268951.720]  -> wl_display@1.sync(new id wl_callback@24)
[1268953.166] wl_display@1.delete_id(24)
[1268963.138] wl_buffer@19.release()
[1268963.419] wl_callback@24.done(0)
[1268967.665]  -> wl_surface@3.attach(wl_buffer@19, 0, 0)
[1268968.220]  -> wl_surface@3.damage(0, 0, 2147483647, 2147483647)
[1268968.935]  -> wl_surface@3.commit()
[1268969.085]  -> wl_display@1.sync(new id wl_callback@24)
[1268970.767] wl_display@1.delete_id(24)
[1268975.286] wl_buffer@19.release()
[1268975.482] wl_callback@24.done(0)
[1268979.953]  -> wl_surface@3.attach(wl_buffer@19, 0, 0)
[1268980.532]  -> wl_surface@3.damage(0, 0, 2147483647, 2147483647)
[1268981.263]  -> wl_surface@3.commit()
[1268981.430]  -> wl_display@1.sync(new id wl_callback@24)
[1268982.689] wl_display@1.delete_id(24)
[1268986.769] wl_buffer@19.release()
[1268986.971] wl_callback@24.done(0)
[1268997.055]  -> wl_surface@3.attach(wl_buffer@19, 0, 0)
[1268997.702]  -> wl_surface@3.damage(0, 0, 2147483647, 2147483647)
[1268998.809]  -> wl_surface@3.commit()
[1268998.999]  -> wl_display@1.sync(new id wl_callback@24)
[1269000.588] wl_display@1.delete_id(24)
[1269008.048] wl_buffer@19.release()
[...]

PR #109, opened 2017-08-18
wlr_surface: use post_event instead of queue

martinetd:
This fixes not sending anything back to the client if the only event
it's waiting for is a buffer release, and nothing else is in the pipe.

Workaround for #102 - there probably is a better solution as weston and
wlc do not need to use post_event here.


SirCmpwn:
Thanks!

PR #113, opened 2017-08-18
rename wlr_cursor to wlr_xcursor

acrisci:
This is for the implementation of another type that should be called wlr_cursor.


SirCmpwn:
Thanks!

PR #114, opened 2017-08-19
Cleanups

martinetd:
Lots of small cleanups:
 - rename init to create for some init functions that did allocs for us
 - more memory leaks, mainly examples/compositor not calling destroy functions.
 - the biggest part is probably wlr_surface, I've taken the liberty to reorder things a bit. There should not be any functional change (except that we no longer union with an empty buffer needlessly), but can be worth looking at.

### DIFF
--- a/backend/wayland/output.c
+++ b/backend/wayland/output.c
@@ -224,7 +224,6 @@ struct wlr_output *wlr_wl_output_create(struct wlr_backend *_backend) {
 
 	wlr_output->width = 640;
 	wlr_output->height = 480;
-	wlr_output->scale = 1;
 	strncpy(wlr_output->make, "wayland", sizeof(wlr_output->make));
 	strncpy(wlr_output->model, "wayland", sizeof(wlr_output->model));
 	snprintf(wlr_output->name, sizeof(wlr_output->name), "WL-%zd",
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -204,20 +204,20 @@ static void wlr_surface_to_buffer_region(struct wlr_surface *surface,
 static void surface_commit(struct wl_client *client,
 		struct wl_resource *resource) {
 	struct wlr_surface *surface = wl_resource_get_user_data(resource);
-	int update_size = 0;
-	int update_damage = 0;
+	bool update_size = false;
+	bool update_damage = false;
 
 	if ((surface->pending.invalid & WLR_SURFACE_INVALID_SCALE)) {
 		surface->current.scale = surface->pending.scale;
-		update_size = 1;
+		update_size = true;
 	}
 	if ((surface->pending.invalid & WLR_SURFACE_INVALID_TRANSFORM)) {
 		surface->current.transform = surface->pending.transform;
-		update_size = 1;
+		update_size = true;
 	}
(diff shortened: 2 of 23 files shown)


backend/wayland/output.c
```diff
@@ -224,7 +224,6 @@ struct wlr_output *wlr_wl_output_create(struct wlr_backend *_backend) {
 
 	wlr_output->width = 640;
 	wlr_output->height = 480;
-	wlr_output->scale = 1;
```
SirCmpwn: Why did you remove this?

types/wlr_surface.c
```diff
@@ -67,6 +67,9 @@ static void surface_set_opaque_region(struct wl_client *client,
 		struct wl_resource *resource,
 		struct wl_resource *region_resource) {
 	struct wlr_surface *surface = wl_resource_get_user_data(resource);
+	if ((surface->pending.invalid & WLR_SURFACE_INVALID_OPAQUE_REGION)) {
+		pixman_region32_clear(&surface->pending.opaque);
```
SirCmpwn: Is this correct? Other regions accumulate according to the protocol.

types/wlr_surface.c
```diff
@@ -199,46 +204,69 @@ static void wlr_surface_to_buffer_region(struct wlr_surface *surface,
...
+	int update_size = 0;
+	int update_damage = 0;
 
+	if ((surface->pending.invalid & WLR_SURFACE_INVALID_SCALE)) {
+		surface->current.scale = surface->pending.scale;
+		update_size = 1;
+	}
+	if ((surface->pending.invalid & WLR_SURFACE_INVALID_TRANSFORM)) {
+		surface->current.transform = surface->pending.transform;
+		update_size = 1;
```
SirCmpwn: Make these stdbool.h booleans

backend/wayland/output.c
ascent12: wlr_output_init sets it already/

martinetd: Yeah, it's just a follow-up on that earlier fix from yesterday. We no longer need it here.

types/wlr_surface.c
martinetd: It's what we do below anyway (we just copy over, while the other two damage callback do unions)
The clear only frees memory if that is called twice. I'll check what the protocol wants us to do.

martinetd: I think it's fine to copy.
wayland-server-protocol.h says this for damage, which is clearly union:
```
         * The initial value for pending damage is empty: no damage.
         * wl_surface.damage adds pending damage: the new pending damage is
         * the union of old pending damage and the given rectangle.
```

On the other hand, opaque and input have nothing about accumulating, instead they just say 'copy'
For opaque
```
         * The initial value for an opaque region is empty. Setting the
         * pending opaque region has copy semantics [...]
```
And input
```
         * The initial value for an input region is infinite. That means
         * the whole surface will accept input. Setting the pending input
         * region has copy semantics...
```

types/wlr_surface.c
martinetd: Done!

SirCmpwn:
Thanks!

PR #115, opened 2017-08-19
move wl_compositor into wlroots as wlr_compositor

martinetd:


examples/compositor.c
```diff
@@ -153,7 +153,7 @@ int main() {
 		exit(EXIT_FAILURE);
 	}
 	wl_display_init_shm(compositor.display);
-	wl_compositor_init(compositor.display, &state.compositor, state.renderer);
+	wlr_compositor_init(&state.compositor, compositor.display, state.renderer);
```
SirCmpwn: This should probably be converted to a create

SirCmpwn:
Thanks!

Issue #93, opened 2017-08-15
Code consistency


ascent12:
Jumping around the code a bit, I can see a few different styles for how different things are handled, particularly error messages, and how certain situations are handled.
These currently aren't covered by CONTRIBUTING.md, so I think we should expand it.

### Error Messages

There are several different formats being used, ranging from "Cannot X, reason", "Failed to X: reason", "X failed: reason", etc.
We should mandate that a single style is used. More use of wlr_log_errno (where appropriate) would help.

### Assertions

There are a few non-NULL asserts around. We should use them more consistently, at least over the public API.
On a related note, any "destruction" function should be able to accept a NULL pointer and exit gracefully, which simplifies error handling a lot.

### Function names

We seem to have functions named *_init, *_create, *_finish, *_destroy. While it's mostly like this already (but not fully), we should document that init/finish should NOT handle allocations, and create/destroy do.

SirCmpwn:
All of these are correct, though I'm not too picky about error message consistency.

acrisci:
The rule of one tab after multi-line functions is inconsistent too.

SirCmpwn:
It's two tabs after multi-line declarations and one tab after multi-line calls.

martinetd:
Not doing this because I'm in the middle of something else (I might later, but feel free to if someone else is doing cleanups), but:
 - struct wl_signal should be in an "events" substruct
These need fixing: include/backend/udev.h, include/wlr/backend/session.h, include/wlr/types/wlr_surface.h, and include/wlr/render.h

ascent12:
It's a pretty minor point, the names we use for include guards is a bit all over the place. It ranges from DRM_BACKEND_H to _WLR_BACKEND_LIBINPUT_INTERNAL_H.

I would suggest we just take the file path relative to include/ and just uppercase everything and replace bad characters with underscores (/, -, . etc.).
include/backend/libinput.h becomes BACKEND_LIBINPUT_H, include/wlr/types/wlr_box.h becomes WLR_TYPES_WLR_BOX_H, etc.

PR #116, opened 2017-08-20
Updated CONTRIBUTING.md

ascent12:
Changed the contribution guidelines to be more consistent with the style that wlroots has evolved into, partially mentioned in #93.
This is my current interpretation of what the style is actually supposed to be. Correct me if I'm wrong.

I've removed some of the things which was duplicating information from the kernel style, and am mainly highlighting the differences from it. We just defer to the kernel style in cases not mentioned here.

I removed the code example, as it had references to very sway specific things, and it didn't actually follow the style being talked about. Not to mention that I think the code was probably a bit too complicated for an example.  
I looked around the wlroots codebase for a replacement, but I couldn't find a reasonably self-contained example that showcased the style and not have some other problem with it.
We can add the example back when the code gets cleaned up a little bit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,8 +2,8 @@
 
 Contributing just involves sending a pull request. You will probably be more
 successful with your contribution if you visit
-[#sway-devel](https://webchat.freenode.net/?channels=sway-devel) on freenode
-upfront and discuss your plans.
+[#sway-devel](https://webchat.freenode.net/?channels=sway-devel) on
+irc.freenode.net upfront and discuss your plans.
 
 ## Pull Requests
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,8 +2,7 @@
 
 Contributing just involves sending a pull request. You will probably be more
 successful with your contribution if you visit the [IRC
-channel](http://webchat.freenode.net/?channels=sway-devel&uio=d4) upfront and discuss
-your plans.
+channel](irc://chat.freenode.net/sway-devel) upfront and discuss your plans.
 
 ## Pull Requests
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -2,203 +2,138 @@
...
+they should be written as a pair of one of two forms:
+* `init`/`finish`: These initialize/deinitialize a type, but are **NOT**
+responsible for allocating it. They should accept a pointer to some
+pre-allocated memory (e.g. a member of a struct).
+* `create`/`destroy`: These also initialize/deinitialize, but will return a
+pointer to a `malloc`ed chunk of memory, and will `free` it in `destroy`.
+
+A destruction function should always be able to accept a NULL pointer or an
+otherwise uninitialised value and exit cleanly; this simplifies error handling
+a lot.
```
martinetd: uninitialized is a bit rough. Can't really handle garbage. zeroed?

CONTRIBUTING.md
```diff
@@ -2,203 +2,138 @@
...
-	registry_teardown(registry);
-	return 0;
+For long lines that need to be broken, the continuation line should be indented
+with an additional tab.  
+If the line being broken is opening a new block (functions, if, while, etc.),
+the continuation line should be indented with two tabs, so they can't be
+misread as being part of the block.
+```c
+really_long_function(argument1, argument2, ...,
+	argument3, argument4);
```
martinetd: hm I was using two tabs for that as well - that's what we have in most headers.
Reasoning (for me) is that it allows a straight copy from definition to declaration.

ascent12: I've seen both in use, so it'll be good if we can decide on a style and have everybody consistently use it.

CONTRIBUTING.md
ascent12: Yeah, I should probably mention the zeroing/calloc somewhere.

CONTRIBUTING.md
SirCmpwn: Definition -  two tabs

Usage - one tab

ascent12: That seems like an awfully specific rule. I would prefer two tabs across the board.

SirCmpwn: There's a reason for it. Definitions indent an additional time to make the indent distinct from the body:

```c
void foobar(...,
        ...) {
    ...
}
```

Whereas calls use only one indent so they can fit more arguments on one line.

ascent12: Oh, so you just mean function definitions?
I thought you meant all definitions, including normal variables.
```c
int really_long_variable_name = really_long_initialiser(...,
        ...);
really_long_function_usage(...,
    ...);
```
Which I thought was kind of pointless.

If it's just functions, then that's covered by the "If the line being broken is opening a new block".

SirCmpwn: Yeah just function defs. Thanks for clarifying.

SirCmpwn:
bump

ascent12:
Updated.

SirCmpwn:
Thanks!

PR #117, opened 2017-08-20
Change meson to be more configurable

ascent12:
We move the warning options into the project's default_options, meaning
a user can configure them with `meson configure` and not need to edit
the file.

We also make it possible to disable logind/libcap even if they're
present.


SirCmpwn:
Thanks!

PR #118, opened 2017-08-20
example compositor: handle keyboard after we're done with init

martinetd:
the libinput backend does wl_display roundtrips, during which there is a
small window where keystrokes can be handled before the rest of the example
compositor is ready.
Setting the callback later ensures we're not called at this point


SirCmpwn:
Thanks!

PR #119, opened 2017-08-20
Xwayland

martinetd:
This whole mess starts an Xwayland server and listens to the bare minimum of events to get the surface to display.

Still need to pass keystrokes and quite a few other events, I didn't want to copy the whole logic out of wlc/weston so I've been adding events one at a time whenever they poped up.
The architecture is globally similar but there are a few variations here and there.

### DIFF
--- a/examples/compositor.c
+++ b/examples/compositor.c
@@ -36,7 +36,7 @@ struct sample_state {
 	struct wlr_data_device_manager *data_device_manager;
 	struct wl_resource *focus;
 	struct wl_listener keyboard_bound;
-	struct wlr_xwayland wlr_xwayland;
+	struct wlr_xwayland *wlr_xwayland;
 	int keymap_fd;
 	size_t keymap_size;
 	uint32_t serial;
@@ -178,13 +178,13 @@ int main() {
 		free(keymap);
 		break;
 	}
-	wlr_xwayland_init(&state.wlr_xwayland, compositor.display, state.wlr_compositor);
+	state.wlr_xwayland = wlr_xwayland_create(compositor.display, state.wlr_compositor);
 
 	compositor.keyboard_key_cb = handle_keyboard_key;
 
 	wl_display_run(compositor.display);
 
-	wlr_xwayland_finish(&state.wlr_xwayland);
+	wlr_xwayland_destroy(state.wlr_xwayland);
 	close(state.keymap_fd);
 	wlr_seat_destroy(state.wl_seat);
 	wlr_data_device_manager_destroy(state.data_device_manager);
--- a/examples/compositor.c
+++ b/examples/compositor.c
@@ -36,7 +36,7 @@ struct sample_state {
 	struct wlr_data_device_manager *data_device_manager;
 	struct wl_resource *focus;
 	struct wl_listener keyboard_bound;
-	struct wlr_xwayland *wlr_xwayland;
+	struct wlr_xwayland *xwayland;
 	int keymap_fd;
 	size_t keymap_size;
 	uint32_t serial;
@@ -84,6 +84,10 @@ static void handle_output_frame(struct output_state *output, struct timespec *ts
 	struct wlr_xdg_surface_v6 *xdg_surface;
 	wl_list_for_each(xdg_surface, &sample->xdg_shell->surfaces, link) {
 		output_frame_handle_surface(sample, wlr_output, ts, xdg_surface->surface);
+	}
+	struct wlr_x11_window *x11_window;
+	wl_list_for_each(x11_window, &sample->xwayland->displayable_windows, link) {
+		output_frame_handle_surface(sample, wlr_output, ts, x11_window->surface);
 	}
 
 	wlr_renderer_end(sample->renderer);
@@ -178,13 +182,13 @@ int main() {
 		free(keymap);
 		break;
 	}
-	state.wlr_xwayland = wlr_xwayland_create(compositor.display, state.wlr_compositor);
+	state.xwayland = wlr_xwayland_create(compositor.display, state.wlr_compositor);
 
 	compositor.keyboard_key_cb = handle_keyboard_key;
 
 	wl_display_run(compositor.display);
 
-	wlr_xwayland_destroy(state.wlr_xwayland);
+	wlr_xwayland_destroy(state.xwayland);
 	close(state.keymap_fd);
 	wlr_seat_destroy(state.wl_seat);
 	wlr_data_device_manager_destroy(state.data_device_manager);
--- a/examples/compositor.c
+++ b/examples/compositor.c
@@ -178,7 +178,7 @@ int main() {
 		free(keymap);
 		break;
 	}
-	wlr_xwayland_init(&state.wlr_xwayland, compositor.display);
+	wlr_xwayland_init(&state.wlr_xwayland, compositor.display, state.wlr_compositor);
 
 	compositor.keyboard_key_cb = handle_keyboard_key;
 
--- a/examples/compositor.c
+++ b/examples/compositor.c
@@ -19,6 +19,7 @@
 #include <wlr/types/wlr_seat.h>
 #include <wlr/types/wlr_data_device_manager.h>
 #include "wlr/types/wlr_compositor.h"
+#include <wlr/xwayland.h>
 #include <xkbcommon/xkbcommon.h>
 #include <wlr/util/log.h>
 #include "shared.h"
@@ -35,6 +36,7 @@ struct sample_state {
 	struct wlr_data_device_manager *data_device_manager;
 	struct wl_resource *focus;
 	struct wl_listener keyboard_bound;
+	struct wlr_xwayland wlr_xwayland;
 	int keymap_fd;
 	size_t keymap_size;
 	uint32_t serial;
@@ -176,11 +178,13 @@ int main() {
 		free(keymap);
 		break;
 	}
+	wlr_xwayland_init(&state.wlr_xwayland, compositor.display);
 
 	compositor.keyboard_key_cb = handle_keyboard_key;
 
 	wl_display_run(compositor.display);
 
+	wlr_xwayland_finish(&state.wlr_xwayland);
 	close(state.keymap_fd);
 	wlr_seat_destroy(state.wl_seat);
 	wlr_data_device_manager_destroy(state.data_device_manager);
--- a/include/wlr/types/wlr_compositor.h
+++ b/include/wlr/types/wlr_compositor.h
@@ -9,6 +9,8 @@ struct wlr_compositor {
 	struct wlr_renderer *renderer;
 	struct wl_list surfaces;
 	struct wl_listener destroy_surface_listener;
+
+	struct wl_signal create_surface_signal;
 };
 
 void wlr_compositor_destroy(struct wlr_compositor *wlr_compositor);
--- a/include/wlr/types/wlr_compositor.h
+++ b/include/wlr/types/wlr_compositor.h
@@ -10,7 +10,9 @@ struct wlr_compositor {
 	struct wl_list surfaces;
 	struct wl_listener destroy_surface_listener;
 
-	struct wl_signal create_surface_signal;
+	struct {
+		struct wl_signal create_surface;
+	} events;
 };
 
 void wlr_compositor_destroy(struct wlr_compositor *wlr_compositor);
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -1,5 +1,11 @@
 #ifndef _WLR_XWAYLAND_H
 #define _WLR_XWAYLAND_H
+#include <time.h>
+#include <sys/types.h>
+#include <stdbool.h>
+#include <wlr/types/wlr_compositor.h>
+
+struct wlr_xwm;
 
 struct wlr_xwayland {
         pid_t pid;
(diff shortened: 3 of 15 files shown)


types/wlr_surface.c
```diff
@@ -386,6 +386,7 @@ struct wlr_surface *wlr_surface_create(struct wl_resource *res,
 		wl_resource_post_no_memory(res);
 		return NULL;
 	}
+	wlr_log(L_DEBUG, "New wlr_surface %p (res %p)", surface, res);
```
martinetd: Not sure about these added logs, as said in the commit message it's probably not something we want all the time, but it helped me debug a bit.

xwayland/sockets.c
```diff
@@ -0,0 +1,147 @@
+#define _XOPEN_SOURCE 700
```
martinetd: We have quite a few #define _XOPEN_SOURCE around, should we put it in some global configuration? (in meson as -D or a global header?)

xwayland/xwayland.c
```diff
@@ -0,0 +1,227 @@
...
+	struct wl_client *client = data;
+	struct wlr_xwayland *wlr_xwayland = wl_container_of(client, wlr_xwayland, client);
+
+	/* don't call client destroy */
+	wlr_xwayland->client = NULL;
+	wlr_xwayland_finish(wlr_xwayland);
+
+	if (wlr_xwayland->server_start - time(NULL) > 5) {
+		wlr_xwayland_init(wlr_xwayland, wlr_xwayland->wl_display,
+				wlr_xwayland->compositor);
```
martinetd: Haven't tested the restart bit yet. wlc does that and it seemed like a good idea.

xwayland/xwm.c
```diff
@@ -0,0 +1,341 @@
...
+			return window;
+		}
+	}
+	return NULL;
+}
+static struct wlr_x11_window *lookup_window_any(struct wlr_xwm *xwm, xcb_window_t window_id) {
+	struct wlr_x11_window *window;
+	if ((window = lookup_window(&xwm->xwayland->displayable_windows, window_id)) ||
+			(window = lookup_window(&xwm->unpaired_windows, window_id)) ||
+			(window = lookup_window(&xwm->new_windows, window_id))) {
```
martinetd: I'm not sure where these three lists should be, they all used to be in xwm but the compositor needed to access the first one.
Basically an wlr_x11_window is always in a list, either new_windows just after creation; for a very short time unpaired (depending on the order in which we get the wayland and the X11 notifications for surface creation) then all will be in displayable_windows

include/wlr/xwayland.h
```diff
@@ -0,0 +1,41 @@
...
+
+struct wlr_x11_window {
+	xcb_window_t window_id;
+	uint32_t surface_id;
+	struct wl_list link;
+
+	struct wl_resource *surface;
+	struct wl_listener surface_destroy_listener;
+	int16_t x, y;
+	uint16_t width, height;
```
martinetd: Actually not using x/y/width/height there, I thought I would need them but didn't.
Not sure if I will later yet, but these come from Xwayland on the wayland client side anyway so probably safe to remove again unless we need some notifications.

examples/compositor.c
```diff
@@ -83,6 +85,10 @@ static void handle_output_frame(struct output_state *output, struct timespec *ts
 	wl_list_for_each(xdg_surface, &sample->xdg_shell->surfaces, link) {
 		output_frame_handle_surface(sample, wlr_output, ts, xdg_surface->surface);
 	}
+	struct wlr_x11_window *x11_window;
+	wl_list_for_each(x11_window, &sample->xwayland->displayable_windows, link) {
+		output_frame_handle_surface(sample, wlr_output, ts, x11_window->surface);
+	}
```
martinetd: What do we gain out of iterating through different surface types, instead of iterating on wlr_surfaces?

xwayland/xwm.h
```diff
@@ -0,0 +1,84 @@
+#ifndef XWAYLAND_INTERNALS_H
+#define XWAYLAND_INTERNALS_H
+#include <wayland-server-core.h>
+#include <wlr/xwayland.h>
+
+/* wlc's atom list:
```
martinetd: Will remove that once we're handling a few more events, it's useful to have right now.

examples/compositor.c
SirCmpwn: Per the spec we should only be rendering views that have a role assigned to them. Each role may also need to be handled subtly differently. Iterating over each of the containers of surfaces with roles attached also separates concerns better.

include/wlr/types/wlr_compositor.h
```diff
@@ -9,6 +9,8 @@ struct wlr_compositor {
 	struct wlr_renderer *renderer;
 	struct wl_list surfaces;
 	struct wl_listener destroy_surface_listener;
+
+	struct wl_signal create_surface_signal;
```
SirCmpwn:
```c
struct {
    struct wl_signal surface_created;
} events;
```

include/wlr/xwayland.h
SirCmpwn: These may be useful to the compositor.

types/wlr_surface.c
SirCmpwn: *shrug* can never have too much debug logging

xwayland/sockets.c
SirCmpwn: No, I prefer that each file defines the minimum feature macros it requires.

xwayland/xwayland.c
```diff
@@ -0,0 +1,227 @@
+#define _XOPEN_SOURCE 700
+#define _GNU_SOURCE
```
SirCmpwn: _GNU_SOURCE is not permitted in wlroots

xwayland/xwayland.c
```diff
@@ -0,0 +1,227 @@
...
+			unset_cloexec(wlr_xwayland->x_fd[1]) ||
+			unset_cloexec(wlr_xwayland->wm_fd[1]) ||
+			unset_cloexec(wlr_xwayland->wl_fd[1])) {
+		exit(EXIT_FAILURE);
+	}
+
+	/* Make Xwayland signal us when it's ready */
+	signal(SIGUSR1, SIG_IGN);
+
+	char *argv[11] = { 0 };
```
SirCmpwn: You could probably set most of these upfront and add the formatted ones separately, would make the code clearer.

SirCmpwn: Bonus points if you comment the formatted ones to give people an idea of how they'll end up looking upfront.

xwayland/xwayland.c
```diff
@@ -0,0 +1,227 @@
...
+				wlr_xwayland->compositor);
+	}
+}
+
+static struct wl_listener xwayland_destroy_listener = {
+	.notify = xwayland_destroy_event,
+};
+
+static void wlr_xwayland_finish(struct wlr_xwayland *wlr_xwayland) {
+
```
SirCmpwn: style - this line doesn't need to be here

xwayland/xwayland.c
```diff
@@ -0,0 +1,227 @@
...
+	safe_close(wlr_xwayland->x_fd[0]);
+	safe_close(wlr_xwayland->x_fd[1]);
+	safe_close(wlr_xwayland->wl_fd[0]);
+	safe_close(wlr_xwayland->wl_fd[1]);
+	safe_close(wlr_xwayland->wm_fd[0]);
+	safe_close(wlr_xwayland->wm_fd[1]);
+
+	unlink_display_sockets(wlr_xwayland->display);
+	unsetenv("DISPLAY");
+	/* kill Xwayland process? */
```
SirCmpwn: Probably smart

xwayland/xwayland.c
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	wl_event_source_remove(wlr_xwayland->sigusr1_source);
+	wlr_xwayland->sigusr1_source = NULL;
+
+	char display_name[16];
+	snprintf(display_name, sizeof(display_name), ":%d", wlr_xwayland->display);
+	setenv("DISPLAY", display_name, true);
+
+	return 1;
```
SirCmpwn: Did you mean to return 0 here? I don't know what calls this I guess

xwayland/xwm.c
```diff
@@ -0,0 +1,341 @@
...
+
+	wlr_log(L_ERROR, "xcb call failed in %s:%u, x11 error code %d",
+			func, line, error->error_code);
+	free(error);
+	return false;
+}
+
+static void map_shell_surface(struct wlr_xwm *xwm, struct wlr_x11_window *window,
+		struct wlr_surface *surface) {
+
```
SirCmpwn: Another extra newline here

xwayland/xwm.c
```diff
@@ -0,0 +1,341 @@
...
+			break;
+		case XCB_PROPERTY_NOTIFY:
+			handle_property_notify(xwm, (xcb_property_notify_event_t *)event);
+			break;
+		case XCB_CLIENT_MESSAGE:
+			handle_client_message(xwm, (xcb_client_message_event_t *)event);
+			break;
+		default:
+			wlr_log(L_DEBUG, "X11 event: %d",
+					event->response_type & XCB_EVENT_RESPONSE_TYPE_MASK);
```
SirCmpwn: you should break here, and only indent once

xwayland/xwm.c
```diff
@@ -0,0 +1,341 @@
...
+		}
+	}
+
+	xcb_flush(xwm->xcb_conn);
+	return count;
+}
+
+static void create_surface_handler(struct wl_listener *listener, void *data)
+{
```
SirCmpwn: Braces on the same line

xwayland/xwm.c
```diff
@@ -0,0 +1,341 @@
...
+		if (window->surface_id == surface_id) {
+			map_shell_surface(xwm, window, surface);
+			xcb_flush(xwm->xcb_conn);
+			return;
+		}
+	}
+}
+
+static void xcb_get_resources(struct wlr_xwm *xwm) {
+	int i;
```
SirCmpwn: This can probably be size_t

xwayland/xwm.c
```diff
@@ -0,0 +1,341 @@
...
+}
+
+static void xcb_init_wm(struct wlr_xwm *xwm) {
+	xcb_screen_iterator_t screen_iterator;
+	screen_iterator = xcb_setup_roots_iterator(xcb_get_setup(xwm->xcb_conn));
+	xwm->screen = screen_iterator.data;
+
+	xwm->window = xcb_generate_id(xwm->xcb_conn);
+
+	uint32_t values[] = { XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT | XCB_EVENT_MASK_PROPERTY_CHANGE,
```
SirCmpwn: Split this up onto several lines

SirCmpwn:
This looks quite good so far, nice work.

xwayland/xwayland.c
martinetd: this is a wayland event loop processing function, so we return "count" like others which happens to be always 1 here.
To be fair in that case wayland ignores it (it only uses count for 'check' mode), so we could always return 0 if that is clearer

xwayland/xwayland.c
martinetd: I'm worried that another process takes the same pid so we'd kill some random thing instead.
For now, Xwayland dies anyway because of broken pipe when we close up, so we do clean up indirectly and it feels safer than checking the process name beforehand.

SirCmpwn: Seems reasonable to me. Change the comment to an explanation of this.

xwayland/xwayland.c
SirCmpwn: Gotcha, we should stick with 1 then.

xwayland/xwayland.c
martinetd: Yeah I actually agonized a bit over this one. I thought of that but didn't like doing the argument that need formating out of order because argv[4] (for example) could move easily.
Hmm, now I've slept on it maybe place-hold with NULL and fill in there to automatically adjust if someone adds static strings. Will come up with something, and remove asprintf (that's the gnu extension here)

xwayland/xwm.h
nyorain: weston uses a struct with those atoms as members iirc, which makes more sense than what wlc does. Should not really be more complicated but way cleaner than those globals values.

martinetd: Weston does have the advantage that there is no enum in the .h, but there is that huge temporary struct within a function that I don't like at all.
There also is an hard-coded list of atoms in the header rightin the weston_wm struct when we just have an ATOM_LAST-sized array there -- I guess it's down to taste, but I prefer our approach that just defines the enum and strings, so we have them all at the same place, and it's easy to loop over.
Wish I could define the two at the same time to ensure no mismatch problem though, will check if there's any trick to that I think I might have seen it somewhere...

martinetd: Looked a bit and didn't find anything that looks better to me here. Obviously a difference in opinion, I'll leave as is as it is my personal preference and defer to @SirCmpwn if he cares (pick one of current enum + static string array all in our header, or weston-style long xwm struct defining all atoms we use in header + array with a substruct that holds string/positions in that struct defined within the code later)

martinetd:
I've pushed more cleanup/round edges. Killing Xwayland now restarts properly, can't find any more problems with shutdown etc.
I've removed the WIP tag - there's still way more X11 events to handle but at this point I think the base is solid enough to merge and continue later - many of the missing events have to do with window positioning/fullscreen/title/etc and frankly I'd rather not start with that on the Xwayland side. It's probably much better to do that in xdg/wl shell first.
Other missing events are cursor which is pretty much the same, and some stuff like clipboard that aren't critical yet.

xwayland/xwm.h
SirCmpwn: In this case I'm inclined to leave it up to you.

SirCmpwn:
LGTM, thanks! Can you file an issue for the remaining xwayland tasks, or post it in the wlroots status issue?

Issue #112, opened 2017-08-18
Output layouts should handle hotplugging


acrisci:
When you unplug a monitor, it should be removed from the layout.

PR #120, opened 2017-08-20
wlr_cursor

acrisci:
It's very rough right now, but it should give you an idea of where it is heading.

I'm pretty sure that "committed with" is a github bug. The log doesn't show me as an author.


SirCmpwn:
Seems like a good start.

examples/pointer.c
```diff
@@ -42,112 +54,156 @@ static void handle_output_frame(struct output_state *output, struct timespec *ts
...
 
+	wlr_cursor_set_xcursor(state.cursor, state.xcursor);
+
 	compositor_init(&compositor);
 	wl_display_run(compositor.display);
 	compositor_fini(&compositor);
 
 	wlr_xcursor_theme_destroy(theme);
+	example_config_destroy(state.config);
+	wlr_cursor_destroy(state.cursor);
```
martinetd: You need to be careful with the order in which you do that, I get a use-after-free when looping over cur->state->devices if you do that before compositor_fini -- probably means there's some uncomplete cleanup?

types/wlr_cursor.c
```diff
@@ -0,0 +1,206 @@
...
+	struct wlr_output *output;
+	output = wlr_output_layout_output_at(cur->state->layout, new_x, new_y);
+
+	if (output) {
+		int output_x = new_x;
+		int output_y = new_y;
+
+		// TODO fix double to int rounding issues
+		wlr_output_layout_output_coords(cur->state->layout, output, &output_x, &output_y);
+		wlr_output_move_cursor(output, output_x - hotspot_x, output_y - hotspot_y);
```
martinetd: this call segfaulted for me after unplugging an output while the cursor was still in there. Took a couple of unplugs but got this, haven't investigated at all:
```==29197==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7ff9ec59c2c8 bp 0x7ffd7e20e560 sp 0x7ffd7e20e540 T0)
==29197==The signal is caused by a READ memory access.
==29197==Hint: address points to the zero page.
    #0 0x7ff9ec59c2c7 in legacy_crtc_move_cursor ../backend/drm/drm-legacy.c:51
    #1 0x7ff9ec597f09 in wlr_drm_output_move_cursor ../backend/drm/drm.c:680
    #2 0x7ff9ec5bc901 in wlr_output_move_cursor ../types/wlr_output.c:171
    #3 0x7ff9ec5befdd in wlr_cursor_move ../types/wlr_cursor.c:97
    #4 0x402c17 in handle_cursor_motion ../examples/pointer.c:104
    #5 0x7ff9ec5be67c in wl_signal_emit /usr/include/wayland-server-core.h:388
    #6 0x7ff9ec5bf152 in handle_pointer_motion ../types/wlr_cursor.c:107
    #7 0x7ff9ec5a1346 in wl_signal_emit /usr/include/wayland-server-core.h:388
    #8 0x7ff9ec5a16de in handle_pointer_motion ../backend/libinput/pointer.c:37
    #9 0x7ff9ec5a0aa3 in wlr_libinput_event ../backend/libinput/events.c:202
    #10 0x7ff9ec59f0f8 in wlr_libinput_readable ../backend/libinput/backend.c:35
    #11 0x7ff9ec360c51 in wl_event_loop_dispatch (/lib64/libwayland-server.so.0+0x9c51)
    #12 0x7ff9ec35f499 in wl_display_run (/lib64/libwayland-server.so.0+0x8499)
    #13 0x4040fa in main ../examples/pointer.c:203
    #14 0x7ff9ebb584d9 in __libc_start_main (/lib64/libc.so.6+0x204d9)
    #15 0x402189 in _start (/home/shared/wayland/wlroots/build/examples/pointer+0x402189)
```

acrisci: This should be fixed by #112

martinetd: Makes sense, ok.

backend/drm/drm.c
```diff
@@ -598,6 +598,8 @@ static bool wlr_drm_output_set_cursor(struct wlr_output *_output,
 		wlr_matrix_texture(plane->matrix, plane->width, plane->height,
 			output->output.transform ^ WL_OUTPUT_TRANSFORM_FLIPPED_180);
 
+		// TODO the image needs to be rotated depending on the output rotation
```
acrisci: I've been working on this for awhile and I don't yet have a good explanation for this bug. Rotating the image will not solve the problem like I thought it would.

For a rotated monitor, the cursor image must be rotated in some 90 degree angle increment. But whenever the output transform is given to rotate 90 or 270 degrees, it puts the cursor in the wrong position on the screen translated about 100 pixels to the left or right. If I try to translate it to the right position, it seems to go out of bounds of some clipping box.

acrisci: I think what could be happening is that if I simply multiply by a rotation matrix, it will rotate around the origin when what I really want is rotation around the center but I'm still not sure what's going on.

SirCmpwn: That sounds like the likely cause.

SirCmpwn: You'll want to rotate around the hotspot of the cursor before you translate.

SirCmpwn: This can be done by translating [-hotspot, -hotspot], rotating, translating [+hotspot, +hotspot], then translating to the final place on screen.

SirCmpwn: It may be worthwhile to add some extensions to our matrix code to reduce the number of steps there. idrc

examples/pointer.c
acrisci: It needed `wl_list_for_each_safe`. You have to use that for all your cleanup loops I guess.

acrisci:
There is kind of a pathological case of a layout having like a 1px border that a mouse move event may or may not "jump over" if the new point from the delta is high enough. Should we handle that case and bound the mouse or leave it? I don't think that setup will be common.

SirCmpwn:
We can probably avoid weird edge cases for now, but file bugs for them please.

acrisci:
Does it really make sense to use wlr_cursor without an output layout? If there is no layout, how do we get the wlr_output that is needed to move the pointer? It seems like we would need two different code paths for every function to handle this case.

SirCmpwn:
No, it doesn't make sense.

types/meson.build
```diff
@@ -16,6 +16,7 @@ lib_wlr_types = static_library('wlr_types', files(
     'wlr_xdg_shell_v6.c',
     'wlr_wl_shell.c',
     'wlr_compositor.c',
+	'wlr_geometry.c',
```
SirCmpwn: Indent

martinetd:
This is starting to be a huge PR! :)

Just started playing with this, a couple of feedbacks:
 * starting pointer with an empty configuration file works as expected, perhaps do that if no file were found (and log something about file not found)?
Also, just a nitpick but I don't really like that it looks in current directory, would be happier if it always looked at the same place relative to binary wherever that is. I've been starting the test programs from all over the place (my home, wlroots root, build dir, example dir in the build dir..) and I'm going to need some practice to get used to it :P I probably won't mind as much if no file works-ish.
 * Just one drm output for now, with default settings pointer looks ok but if I rotate=90 the "input box" looks wrong - it cannot reach the bottom by quite a bit. It looks like the top is off by a similar margin so looks like positioning issue? I have similar problems with all other rotate values, flipped or not, as well. flipped adds an extra layer of wrongnes I think, just flipped is already wrong and others have a different feel than no flip.

I'll get onto reading code now, will repost once it's all read (might be tomorrow morning or much later tonight)

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+		"\n"
+		" -C <FILE>      Path to the configuration file (default: wlr-example.ini).\n"
+		"                See `examples/wlr-example.ini.example` for config file documentation.\n", name);
+
+	exit(ret);
+}
+
+static struct wlr_geometry *parse_geometry(const char *str) {
+	// format: {width}x{height}+{x}+{y}
+	if (strlen(str) > 255l) {
```
martinetd: Where does 2551 come from?

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+		} else if (!has_y) {
+			geo->y = val;
+			has_y = true;
+		} else {
+			break;
+		}
+		pch = strtok(NULL, "x+");
+	}
+
+	if (!has_width || !has_height || !has_x || !has_y) {
```
martinetd: +x+y could be assumed to be 0 if omitted?

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+	free(geo);
+	return NULL;
+}
+
+static const char *output_prefix = "output:";
+static const char *device_prefix = "device:";
+
+static int config_ini_handler(void *user, const char *section, const char *name, const char *value) {
+	struct example_config *config = user;
+	if (strncmp(output_prefix, section, strlen(output_prefix)) == 0) {
```
martinetd: calling strncmp with strlen of either argument is pointless, use strcmp?

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+			} else {
+				wlr_log(L_ERROR, "got unknown transform value: %s", value);
+			}
+		}
+	} else if (strcmp(section, "cursor") == 0) {
+		if (strcmp(name, "map-to-output") == 0) {
+			config->cursor.mapped_output = strdup(value);
+		} else if (strcmp(name, "geometry") == 0) {
+			if (config->cursor.mapped_geo) {
+				free(config->cursor.mapped_geo);
```
martinetd: (nitpick) feels weird to have this one freed on multi-value but not others (e.g. mapped_output just above)
(also, free can be called on NULL values, not sure if we have a policy about checking for null before free)

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+			break;
+		case 'h':
+		case '?':
+			usage(argv[0], c != 'h');
+		}
+	}
+
+	if (!config->config_path) {
+		// get the config path from the current directory
+		char cwd[1024];
```
martinetd: Probably want to make that 4k as that is the max on most system, or use MAXPATHLEN from sys/param.h

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+		if (getcwd(cwd, sizeof(cwd)) != NULL) {
+			char buf[1024];
+			sprintf(buf, "%s/%s", cwd, "wlr-example.ini");
```
martinetd: snprintf (and ditto on length)

examples/config.c
```diff
@@ -0,0 +1,304 @@
...
+	if (config->cursor.mapped_geo) {
+		free(config->cursor.mapped_geo);
+	}
+	free(config);
+}
+
+struct wlr_output_layout *configure_layout(struct example_config *config, struct wl_list *outputs) {
+	struct wlr_output_layout *layout = wlr_output_layout_init();
+	int max_x = INT_MIN;
+	int max_x_y = INT_MIN; // y value for the max_x output
```
martinetd: (nitpick) y_at_max_x? If you feel the need to comment a variable name, there might be a better fit -- not sure mine is :P

examples/wlr-example.ini.example
```diff
@@ -0,0 +1,57 @@
...
+# Device Configuration
+# ~~~~~~~~~~~~~~~~~~~~
+# Each device is specified in a section named [device:{NAME}] where NAME is the
+# name given to this device. See a log file for device names.
+#
+# Value "map-to-output" specifies the output to which the device is constrained.
+#
+# Value "geometry" specifies the geometry (widthxheight+x+y) to which the device
+# is constrained.
+[device:Razer Razer DeathAdder 2013]
```
martinetd: It's ok for now, but might need an invocation command that just lists outputs/devices at some point later... Gathering that from debug logs is not really good even if we want testers.

include/wlr/types/wlr_output_layout.h
```diff
@@ -4,14 +4,20 @@
 #include <wayland-util.h>
 #include <stdbool.h>
 
+struct wlr_output_layout_state;
+
 struct wlr_output_layout {
     struct wl_list outputs;
+	struct wlr_output_layout_state *state;
```
martinetd: You're not to blame now but doesn't match previous indentation, I'd say fix that while you're here... Mixed indentation is evil.

include/wlr/types/wlr_output_layout.h
```diff
@@ -38,12 +44,28 @@ void wlr_output_layout_remove(struct wlr_output_layout *layout,
  * coordinates relative to the given reference output.
  */
 void wlr_output_layout_output_coords(struct wlr_output_layout *layout,
-		struct wlr_output *reference, int *x, int *y);
+		struct wlr_output *reference, double *x, double *y);
```
martinetd: Why do x and y have to be doubles? within the function it's comparing with struct wlr_output_layout_output x and y fields, which are int.

. . . okay, I see later that everything else got changed to double. I don't understand why yet, but maybe change that in the type as well?

types/wlr_output_layout.c
```diff
@@ -103,26 +131,120 @@ void wlr_output_layout_move(struct wlr_output_layout *layout,
...
+			*x = src_x - (double)_output->x;
+			*y = src_y - (double)_output->y;
 			return;
 		}
 	}
 }
+
+static double get_distance(double x1, double y1, double x2, double y2) {
+	double distance;
+	distance = sqrt((x1 - x2) * (x1 - x2) + (y1 - y2) * (y1 - y2));
```
martinetd: If we just want the distance to compare, just work with squared-distance as sqrt preserve ordering.
Ignore me if it's used as distance proper.

acrisci:
> starting pointer with an empty configuration file works as expected, perhaps do that if no file were found (and log something about file not found)?

Yeah the configuration file should always be optional. I'll have to fix that.

> Just one drm output for now, with default settings pointer looks ok but if I rotate=90 the "input box" looks wrong - it cannot reach the bottom by quite a bit

Yeah that's an issue that I've discussed in the chat and in some conversation above. I think that the issue is that the point of rotation of the cursor texture is around the origin rather than the center which would cause the issue but I still don't have a great understanding of what's going on because of some gaps in my knowledge about 2D rendering. I'm not sure if I can fix it myself without dedicating a lot of time to it so if you want to look into it go ahead.

examples/config.c
acrisci: I misread that today too. That's an L. :) I don't even know if there's a real need for that length check though honestly.

examples/config.c
acrisci: Yeah that's probably a good idea.

examples/config.c
SirCmpwn: Uppercase L perhaps?

examples/config.c
acrisci: Yeah I think that's leftover from when I was doing it a different way.

martinetd:
> Yeah that's an issue that I've discussed in the chat and in some conversation above. I think that the issue is that the point of rotation of the cursor texture is around the origin rather than the center which would cause the issue but I still don't have a great understanding of what's going on because of some gaps in my knowledge about 2D rendering. I'm not sure if I can fix it myself without dedicating a lot of time to it so if you want to look into it go ahead.

Ah, I had misunderstood the discussions there to be that the actual point of the cursor is off (so instead of sending click event at proper x/y it would go to another x/y within the cursor block). I was actually wondering how you found that out with the current test program, as it doesn't have a button area to click or something to aim at :P
I guess that could still be the case but if so the cursor block would have to be huuuge given what I'm seeing. And wrongly shaped with e.g. flipped-270. I'm not much better at geometry but will have a look later this week, this particular point shouldn't block us from merging this PR in anyway.

backend/drm/drm.c
ascent12: The buffer that the cursor sits on (the DRM plane) may be much larger than the cursor actually is. The cursor is sitting in the top-left corner.
On some intel hardware I own, the cursor plane is 256x256 (although 64x64 is more typical for other hardware), so just naively rotating the cursor is going to put it in the wrong place by potentially quite a long way.

You're going to need to translate by the plane dimensions. The cursor plane dimensions are found in output->crtc->cursor->{width,height}.

examples/config.c
ascent12: The L is pretty pointless. There is no reason to force that to be a long type.

types/wlr_output_layout.c
ascent12: You can use hypot from math.h instead of doing this.

examples/pointer.c
```diff
@@ -42,112 +88,291 @@ static void handle_output_frame(struct output_state *output, struct timespec *ts
...
+		s_device = calloc(1, sizeof(struct sample_input_device));
+		s_device->device = device;
+
+		wl_list_insert(&sample->devices, &s_device->link);
+		wlr_cursor_attach_input_device(sample->cursor, device);
+		configure_devices(sample);
+	}
+}
+
+static void handle_input_remove(struct compositor_state *state, struct
```
SirCmpwn: Don't split parameter declarations across lines, put the whole thing on the next line

SirCmpwn: (this is not the only place this is a problem)

include/wlr/types/wlr_cursor.h
```diff
@@ -0,0 +1,113 @@
...
+		struct wl_signal touch_up;
+		struct wl_signal touch_down;
+		struct wl_signal touch_motion;
+		struct wl_signal touch_cancel;
+
+		struct wl_signal tablet_tool_axis;
+		struct wl_signal tablet_tool_proximity;
+		struct wl_signal tablet_tool_tip;
+		struct wl_signal tablet_tool_button;
+    } events;
```
SirCmpwn: This indentation is fucky

SirCmpwn: Also, I think it would make sense for wlr_cursor to expose a pointer-like API for all of the devices its using. IMO exposing touch/tool events like this is a bit of a non-sequiter since using touch or tool devices normally will completely circumvent the wayland pointer API this is meant to abstract.

include/wlr/types/wlr_cursor.h
```diff
@@ -0,0 +1,113 @@
...
+#include <wlr/types/wlr_input_device.h>
+#include <wlr/types/wlr_geometry.h>
+#include <wlr/xcursor.h>
+
+struct wlr_cursor_state;
+//struct wlr_cursor_impl *;
+
+struct wlr_cursor {
+    struct wlr_cursor_state *state;
+    //struct wlr_cursor_impl *impl;
```
SirCmpwn: Don't leave this commented out, just don't add it at all unless we need it.

include/wlr/types/wlr_cursor.h
```diff
@@ -0,0 +1,113 @@
...
+		struct wl_signal touch_cancel;
+
+		struct wl_signal tablet_tool_axis;
+		struct wl_signal tablet_tool_proximity;
+		struct wl_signal tablet_tool_tip;
+		struct wl_signal tablet_tool_button;
+    } events;
+};
+
+struct wlr_cursor *wlr_cursor_init();
```
SirCmpwn: Use _create convention when you're doing the allocating

include/wlr/types/wlr_cursor.h
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * - WLR_INPUT_DEVICE_POINTER
+ * - WLR_INPUT_DEVICE_TOUCH
+ * - WLR_INPUT_DEVICE_TABLET_TOOL
+ */
+void wlr_cursor_attach_input_device(struct wlr_cursor *cur,
+    struct wlr_input_device *dev);
+
+void wlr_cursor_detach_input_device(struct wlr_cursor *cur,
+    struct wlr_input_device *dev);
```
SirCmpwn: Further indent issues. Do a style pass on the whole PR please

include/wlr/types/wlr_geometry.h
```diff
@@ -0,0 +1,20 @@
...
+
+void wlr_geometry_closest_boundary(struct wlr_geometry *geo, double x, double y,
+		int *dest_x, int *dest_y, double *distance);
+
+bool wlr_geometry_intersection(struct wlr_geometry *geo_a,
+		struct wlr_geometry *geo_b, struct wlr_geometry **dest);
+
+bool wlr_geometry_contains_point(struct wlr_geometry *geo, int x, int y);
+
+bool wlr_geometry_empty(struct wlr_geometry *geo);
```
SirCmpwn: You were going to merge this into wlr_region, right?

types/wlr_cursor.c
```diff
@@ -0,0 +1,469 @@
...
+ * 3. cursor geometry mapping
+ * 4. cursor output mapping
+ *
+ * Absolute movement for touch and pen devices will be relative to this box and
+ * pointer movement will be constrained to this box.
+ *
+ * If none of these are set, returns NULL and absolute movement should be
+ * relative to the extents of the layout.
+ */
+static struct wlr_geometry *get_mapping(struct wlr_cursor *cur,
```
SirCmpwn: const?

types/wlr_geometry.c
```diff
@@ -0,0 +1,86 @@
...
+#include <stdlib.h>
+#include <math.h>
+#include <stdbool.h>
+#include <wlr/types/wlr_geometry.h>
+#include <wlr/util/log.h>
+
+static double get_distance(double x1, double y1, double x2, double y2) {
+	double distance;
+	distance = sqrt((x1 - x2) * (x1 - x2) + (y1 - y2) * (y1 - y2));
+	return distance;
```
SirCmpwn: This could be one line

types/wlr_geometry.c
```diff
@@ -0,0 +1,86 @@
...
+
+	// calculate distance
+	if (distance) {
+		*distance = get_distance(*dest_x, *dest_y, x, y);
+	}
+}
+
+#ifndef max
+#define max(a,b) ((a) > (b) ? (a) : (b))
+#endif
```
SirCmpwn: fmin/fmax? math.h?

types/wlr_output_layout.c
```diff
@@ -1,36 +1,60 @@
...
 	wl_list_init(&layout->outputs);
 	return layout;
 }
 
+static void wlr_output_layout_output_destroy(
+		struct wlr_output_layout_output *l_output) {
+		wl_list_remove(&l_output->link);
+		free(l_output->state->_geo);
+		free(l_output->state);
+		free(l_output);
```
SirCmpwn: indent

SirCmpwn:
Overall, looks great. Grabbing some hardware to test it on now.

SirCmpwn:
Output config should let you specify a mode, and the file should be optional.

SirCmpwn:
Bug: if I have an absolute input device (like my drawing tablet) and a relative one (like my mouse), the following procedure demonstrates a bug:

1. Move the relative device around
2. Move the absolute device around
3. Move the relative device around

At step 3, it warps back to where it was at the end of step 1. It should instead pick up from where the absolute device left off.

SirCmpwn:
Verified that it works with touch screens and pushed a fix for tablet tools. Great work!

include/wlr/types/wlr_cursor.h
acrisci: so are you suggesting to merge some of the events together?

acrisci:
Now I'm a bit unsure that wlr_geometry should be merged into wlr_region because a region data structure doesn't really fit our use case very well because it's too general.

Most often we want to work with boxes not regions. An output is a box (not a region). Same for an input geometry. Only the output layout as a whole is a region but it's not really inconvenient to work with it as a collection of boxes.

Maybe the best thing to do is to rename wlr_geometry to wlr_box.

SirCmpwn:
I'm fine with whichever approach to regions/geometry/boxes you deem most suitable.

include/wlr/types/wlr_cursor.h
SirCmpwn: No, on further thought this approach is fine.

acrisci:
Thanks for the reviews.

I think I've addressed everybody's issues.

SirCmpwn:
Looks good, thanks!

PR #121, opened 2017-08-21
wlr_seat: fix destroy with multiple handles

martinetd:
Need to use for_each_safe because wl_resource_destroy's callback will
remove the handle from the list itself


SirCmpwn:
Thanks!

PR #123, opened 2017-08-26
Session Multi-GPU

ascent12:
This is part of the effort to get multi-GPU support in the DRM backend.
It removes the single-GPU assumptions from wlr_session code.

In order to do this, I needed to maintain a list of all of the devices we have open, which wlr_udev already maintained. Instead of maintaining two different lists of devices, or trying to create an extra interface between wlr_udev and wlr_session, I just merged them together.
They are doing conceptually similar things and are always used together anyway.


SirCmpwn:
LGTM, thanks!

PR #124, opened 2017-08-27
Xwayland double fork & leak fix

martinetd:
Double-forking xwayland was a bit more complicated than I had hoped (because the SIGUSR1 signal when xserver is ready only goes to parent), but I guess it's not too bad.

Also used the opportunity to move some code around and only try to init xwm if the Xwayland process started properly, that wasn't checked correctly before.


SirCmpwn:
LGTM, thanks!

Issue #127, opened 2017-08-29
Output layout autoconfiguration


acrisci:
An auto configured output is positioned in a sensible place on the screen. The user does not specify coordinates for an auto configured output, rather the coordinates are adjusted dynamically whenever the layout changes. An output that is not auto configured is called manually configured.

The use case is for compositors that don't want to require user configuration for outputs within the layout.

Auto configuration is presently implemented in `examples/config.c:configure_layout()` but should be moved into wlroots.

```c
/**
* Add an auto configured output to the layout. This will place the output in a
* sensible location in the layout. The coordinates of the output in the layout
* may adjust dynamically when the layout changes. If the output is already in
* the layout, it will become auto configured. If the position of the output is
* set such as with `wlr_output_layout_move()`, the output will become manually
* configured.
*/
void wlr_output_layout_add_auto(struct wlr_output_layout *layout, struct wlr_output *output);
```

acrisci:
Maybe the auto configuration strategy should be configurable. Right now it adds all the outputs to the right of the rightmost output, but users might want to add the outputs to the left, above, below, or something else.

SirCmpwn:
Not very important IMO. They can just manually arrange them at that point.

PR #128, opened 2017-08-30
implement output layout auto configuration

acrisci:
#127

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -52,13 +52,13 @@ void wlr_output_layout_destroy(struct wlr_output_layout *layout) {
 
 static struct wlr_box *wlr_output_layout_output_get_box(
 		struct wlr_output_layout_output *l_output) {
-		l_output->state->_box->x = l_output->x;
-		l_output->state->_box->y = l_output->y;
-		int width, height;
-		wlr_output_effective_resolution(l_output->output, &width, &height);
-		l_output->state->_box->width = width;
-		l_output->state->_box->height = height;
-		return l_output->state->_box;
+	l_output->state->_box->x = l_output->x;
+	l_output->state->_box->y = l_output->y;
+	int width, height;
+	wlr_output_effective_resolution(l_output->output, &width, &height);
+	l_output->state->_box->width = width;
+	l_output->state->_box->height = height;
+	return l_output->state->_box;
 }
 
 /**
@@ -150,20 +150,13 @@ struct wlr_output_layout_output *wlr_output_layout_get(
 	return NULL;
 }
 
-static bool output_contains_point( struct wlr_output_layout_output *l_output,
-		int x, int y, int width, int height) {
-	return x >= l_output->x && x <= l_output->x + width &&
-		y >= l_output->y && y <= l_output->y + height;
-}
-
 bool wlr_output_layout_contains_point(struct wlr_output_layout *layout,
 		struct wlr_output *reference, int x, int y) {
 	if (reference) {
-		struct wlr_output_layout_output *layout_output =
+		struct wlr_output_layout_output *l_output =
 			wlr_output_layout_get(layout, reference);
-		int width, height;
-		wlr_output_effective_resolution(layout_output->output, &width, &height);
-		return output_contains_point(layout_output, x, y, width, height);
+		struct wlr_box *box = wlr_output_layout_output_get_box(l_output);
+		return wlr_box_contains_point(box, x, y);
 	} else {
 		return !!wlr_output_layout_output_at(layout, x, y);
 	}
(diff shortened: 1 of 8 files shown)


acrisci:
To properly complete this implementation, I'm going to need to know whether or not an output is connected and get events when output connection status changes. The layout must be reconfigured when an output is disconnected. Right now, I'm hooking into the `output_remove` event from the backend and removing the output from the layout then (which triggers reconfiguration) but this can be handled automatically in a nicer way for users.

I think a disconnected output should rather stay within the layout but be invisible to any geometry calculations and queries.

I think we already discussed reasons for not wanting to put the backend within the high level api, so I think the events and connection queries should go on the wlr_output itself.

Connection status presently seems to be implementation specific, with the drm backend having its own enum for this. So I think we should add a function to the wlr_output interface called `wlr_output_is_connected()` that queries connection state.

As for who emits the connection status events, I think we have two options. One is to require the backend to emit the events on the output (essentially at the same time as the backend input_add/remove event) or for the backend to hook into its own input_add/remove event and emit the connection signal then. The latter is appealing because it takes some responsibility off of the backend.

ascent12:
> Connection status presently seems to be implementation specific

It might be a good idea to move the DRM state-machine (that enum) into wlr_output. This was actually mentioned ages ago, around the time the DRM backend was first written, but nobody ever got around to it.

SirCmpwn:
I prefer the approach where disconnected outputs are removed rather than bubbling the connector model up to the API surface.

SirCmpwn:
To clarify, I don't think a disconnected output should stay in the layout. That would be weird.

acrisci:
I got bored so I went ahead and did the implementation the way I described it to see what it looked like :)

As you see, it is a little bit weird. The nice thing is that you aren't required to re-add the output when it reconnects which is one less thing for users to keep track of, but I will defer to your judgement.

I think putting the DRM enum on the wlr_output is probably a good idea too.

SirCmpwn:
The problem is that connectedness is a property of the connector, not the output. If you were to do this then there would have to be a wlr_connector which has ownership over the wlr_output. This would only be in use by the DRM backend and would add a fair amount of complexity.

The behavior you want is pretty weird, too. When a display is unplugged, its resolution and other properties become undefined until plugged in again, when it might have changed. People don't unplug their monitors and leave them there turned off, either - they unplug them and remove them from their desk, or unplug themselves from a projector, or undock their laptop and take it somewhere else.

:-1:

ascent12:
>The nice thing is that you aren't required to re-add the output when it reconnects which is one less thing for users to keep track of

I don't think that's really a helpful feature. The way that we've designed it, the user should be able to able to handle displays being added and removed at any time. This somewhat specific special case needlessly breaks that "model".

Also, they can easily place that output in the right place, based on whatever their config is. We don't need to store that state for them.

types/wlr_output_layout.c
```diff
@@ -47,16 +58,115 @@ void wlr_output_layout_destroy(struct wlr_output_layout *layout) {
...
+static struct wlr_box *wlr_output_layout_output_get_box(
+		struct wlr_output_layout_output *l_output) {
+		l_output->state->_box->x = l_output->x;
+		l_output->state->_box->y = l_output->y;
+		int width, height;
+		wlr_output_effective_resolution(l_output->output, &width, &height);
+		l_output->state->_box->width = width;
+		l_output->state->_box->height = height;
+		return l_output->state->_box;
+}
```
SirCmpwn: Check indent

SirCmpwn:
What's this new "main" output for? Also, we might want to talk about auto layouts with KDE.

acrisci:
The "main" output exists only so the physics don't calculate at a different rate depending on how many monitors you have but it actually might be unnecessary.

SirCmpwn:
Try to get rid of it, please.

acrisci:
That all makes goos sense. You can consider this feature complete minus the last commit then.

acrisci:
The "main" output here is necessary because otherwise, the animation speed will be proportional to the number of outputs which is not correct behavior. The problem is that each output has its own independent frame event so this gets called more times for more outputs.

This is actually going to be a problem for anyone who wants to do animations in the output frame callback (and where else would you do animations?). This solution is very simple and works for the example, but I believe that we will have to come up with some sort of framework to do animations. I think that's out of the scope of this PR, but definitely something that needs to be done.

acrisci:
I think this is ready. This should fix #112 and #127. Please test hotplugging.

ascent12:
> This is actually going to be a problem for anyone who wants to do animations in the output frame callback

I would actually consider that a bug in the application's code. Consistent frame events are not something we should promise, and I don't think other wayland compositors do that either. There are technologies like freesync which would also mess with that.

SirCmpwn:
If you want to do animation in the frame callback you should get the current monotonic time and base it on that. Pull it out.

acrisci:
I don't do many animation code so that wasn't immediately obvious to me like it should have been. Thanks.

acrisci:
Ok that works much better.

SirCmpwn:
Nice, thanks!

PR #133, opened 2017-09-04
wlr_cursor: automatically handle layout changes

acrisci:
Automatically set the cursor on outputs when a layout is attached.

Add a `destroy` event to wlr_output_layout and listen for it in wlr_cursor to
detach the output layout cleanly.

Add a `change` event to wlr_output_layout and listen for it in wlr_cursor to
set the cursor on the outputs in the layout.

Reconfigure the cursor on the outputs when the xcursor changes.

### DIFF
--- a/examples/pointer.c
+++ b/examples/pointer.c
@@ -151,15 +151,15 @@ static void handle_output_add(struct output_state *ostate) {
 
 	configure_devices(sample);
 
-	// TODO move to wlr_cursor
+	// TODO the cursor must be set depending on which surface it is displayed
+	// over which should happen in the compositor.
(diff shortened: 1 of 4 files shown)


acrisci:
Should we initialize the wlr_cursor with a sensible default xcursor to remove some boilerplate from the pointer examples?

SirCmpwn:
>Automatically set the cursor on outputs when a layout is attached.

I would rather not do this here, because the cursor is going to have to change depending on which surface it's displayed over, which is a compositor concern.

acrisci:
Ok i took that out but left in the good parts.

SirCmpwn:
Thanks!

PR #134, opened 2017-09-05
pointer.c: refactor device configuration

acrisci:
It's a bit more readable now, but I'm afraid there's no way we can avoid copy/pasting this into each example that needs a pointer unless they can share `sample_state`.

### DIFF
--- a/examples/config.c
+++ b/examples/config.c
@@ -278,3 +278,70 @@ struct device_config *example_config_get_device(struct example_config *config,
 
 	return NULL;
 }
+
+/**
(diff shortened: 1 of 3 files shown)


SirCmpwn:
I don't understand. Why can't this live in shared.c?

acrisci:
Actually I think I can put it in shared.c. Let me look at it again when I get home today.

acrisci:
Ok moving that to config.h should reduce boilerplate for the cursor by a lot.

SirCmpwn:
Thanks!

PR #136, opened 2017-09-05
Rename wlr_output_layout_init()

acrisci:
Rename wlr_output_layout_init() to wlr_output_layout_create() to be consistent
with the rest of the api.


SirCmpwn:
Thanks!

PR #137, opened 2017-09-06
Add gamma_control interface

emersion:
This adds the `gamma_control` interface from orbital: https://github.com/giucam/orbital/blob/master/protocol/gamma-control.xml

For reference, this is the PR that added gamma control support to wlc: https://github.com/Cloudef/wlc/pull/216/files. And this is the sway implementation of the protocol: https://github.com/SirCmpwn/sway/blob/4e663ab44f5320f20ec2376f24ea1eb3b458f2c2/sway/extensions.c#L251

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -301,6 +301,23 @@ static void wlr_drm_output_swap_buffers(struct wlr_output *_output) {
 	output->pageflip_pending = true;
 }
 
+static void wlr_drm_output_set_gamma(struct wlr_output *_output,
+		uint16_t size, uint16_t *r, uint16_t *g, uint16_t *b) {
+	struct wlr_drm_output *output = (struct wlr_drm_output *)_output;
+	struct wlr_drm_backend *backend =
+		wl_container_of(output->renderer, backend, renderer);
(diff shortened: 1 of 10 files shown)


SirCmpwn:
Looks like it's still WIP? The wlr_output still needs to actually implement this.

SirCmpwn:
Ah, nevermind, I misread the changes.

SirCmpwn:
Thanks!

acrisci:
In `wlr_drm_output_set_gamma()`, what is `size`?

emersion:
@acrisci: The size of the `r`, `g` and `b` `uint16_t` arrays.

acrisci:
Oh ok I've only ever understood gamma as a single color value. What happens if I pass arrays with size > 1?

acrisci:
I can't get `wlr_drm_output_set_gamma()` to work. The libdrm function returns -22, which I think is invalid argument, whenever I call it.

Can you provide an example call that will make the function work correctly?

acrisci:
Oh sorry I just looked at the call you posted above. I think this api is extremely confusing.

SirCmpwn:
Try using the guicam redshift fork and letting it use the protocol to set the gamma.

PR #138, opened 2017-09-06
compositor example basic cursor and layout

acrisci:
TODO:

* render surfaces within the correct place in the layout
* focus a window with the pointer
* move a window with the pointer


acrisci:
Now I realize I have to do some work fleshing out the desktop shells to accomplish my goals here and that should probably be done in a separate PR.

If you want to merge this, you can because it's work that needs to be done anyway.

SirCmpwn:
Sure, looks like a good start.

PR #140, opened 2017-09-07
Add screenshooter

### LINKED ISSUE #139: Implement screenshot protocol
So the original idea for the screenshooter is to implement weston's protocol: https://github.com/wayland-project/weston/blob/master/protocol/weston-screenshooter.xml

Though, this protocol has some issues, discussed here: https://lists.freedesktop.org/archives/wayland-devel/2013-December/thread.html#12571. Also, this protocol is just a demo, not a standard.

Orbital has fixed one of them, by adding a `screenshot` interface: https://github.com/giucam/orbital/blob/gamma/protocol/screenshooter.xml

So we have 3 solutions:

1. Implement weston's protocol
2. Implement orbital's protocol
3. Design another protocol, and propose it as a standard

emersion:
This adds the `orbital_screenshooter` interface from https://github.com/giucam/orbital/blob/gamma/protocol/screenshooter.xml

~~Security hooks will come in another PR (#141).~~

Fixes #139


types/wlr_screenshooter.c
```diff
@@ -0,0 +1,66 @@
...
+	struct wlr_screenshot *screenshot;
+	if (!(screenshot = calloc(1, sizeof(struct wlr_screenshot)))) {
+		return;
+	}
+	screenshot->output = _output;
+	screenshot->resource = wl_resource_create(client,
+		&orbital_screenshot_interface, wl_resource_get_version(_screenshooter), id);
+	wlr_log(L_DEBUG, "new screenshot %p (res %p)", screenshot, screenshot->resource);
+	wl_resource_set_implementation(screenshot->resource, NULL, screenshot, NULL);
+	// TODO: orbital_screenshot_send_done(screenshot->resource);
```
SirCmpwn: When you do implement this, make sure you sync it up with the output's frame event.

SirCmpwn:
You probably want to add a post-frame callback for this purpose, rather than sharing the frame callback. I'm guessing that your callback is getting run *before* the normal frame callback, i.e. the one that renders the scene.

emersion:
Oh, I see, that explains why I get transparent screenshots. Any ideas how to do this properly? Add another event?

SirCmpwn:
Yeah, a new event. I'm not sure if this is your actual problem but it's a lead I'd chase down.

emersion:
Hum, that doesn't seem to be the problem after all, still getting transparent screenshots. Will think about it.

SirCmpwn:
Note: it seems wlc reads pixels between rendering everything and swapping the buffer.

SirCmpwn:
It also appears to use a "fake framebuffer", not sure what for. @Cloudef?

Cloudef:
The fakefb is only for write_pixels it has no other purpose. The implementation isn't very effecient, see: https://github.com/SirCmpwn/sway/issues/1004#issuecomment-268947734

emersion:
The screenshooter interface is now working properly. I'll begin to cleanup the screenshooter executable in `examples/`. Do you want to ditch the cairo dependency?

SirCmpwn:
If you can, yes.

emersion:
Okay, so I spent some time figuring out what are our options. The goal is to replace cairo by spawning imagemagick and feed it with raw data. However, imagemagick doesn't support ARGB. I tried to play with the CLI (`-color-matrix`, `-channel`, `-swap`) but I couldn't get the right flags. I then saw `wl_shm_buffer` supports other formats, but `wl_shm` will refuse to create a buffer with an unsupported format, and ARGB is the only supported format. That's a shame because directly requesting RGB data would have been cool.

So I think I'll end up only supporting ARGB in the screenshooter interface (since we don't support anything else anyway) and swap the channels in the client before passing the data to imagemagick.

SirCmpwn:
https://github.com/swaywm/sway/blob/master/swaygrab/main.c#L73

SirCmpwn:
I'm okay with keeping cairo if necessary.

emersion:
Yeah, I started all this work by looking at sway's source.

I've just discovered that ES 2.0 doesn't support ARGB. So we'll end up reading pixels as RGBA, copying pixels to the shared buffer as ARGB, and converting back pixels to RGBA in the client.

emersion:
All right, this is ready for review.

examples/screenshot.c
```diff
@@ -0,0 +1,288 @@
...
+		if (output->buffer == NULL) {
+			return -1;
+		}
+		struct orbital_screenshot *screenshot = orbital_screenshooter_shoot(
+			screenshooter, output->output, output->buffer);
+		orbital_screenshot_add_listener(screenshot, &screenshot_listener, screenshot);
+		buffer_copy_done = 0;
+		while (!buffer_copy_done)
+			wl_display_roundtrip(display);
+	}
```
SirCmpwn: Wait, how does this work if there are several outputs?

SirCmpwn: I guess you're trying to stitch them together?

types/wlr_output.c
```diff
@@ -247,3 +250,20 @@ uint16_t wlr_output_get_gamma_size(struct wlr_output *output) {
...
+		uint32_t v = data[i];
+		uint32_t rgb = (v & 0xffffff00) >> 8;
+		uint32_t a = v & 0x000000ff;
+		data[i] = rgb | (a << 24);
+	}
+}
+
+void wlr_output_read_pixels(struct wlr_output *output, void *out_data) {
+	wlr_output_make_current(output);
+	glReadPixels(0, 0, output->width, output->height, GL_RGBA, GL_UNSIGNED_BYTE,
```
SirCmpwn: This should go through the renderer instead of putting GL calls in wlr_output. And do the rgba->argb translation in there too, the pixel format is specific to the renderer so it needs to be converted to the standard format by the renderer.

types/wlr_screenshooter.c
```diff
@@ -0,0 +1,148 @@
...
+	screenshooter->wl_global = wl_global;
+	return screenshooter;
+}
+
+void wlr_screenshooter_destroy(struct wlr_screenshooter *screenshooter) {
+	if (!screenshooter) {
+		return;
+	}
+	// TODO: this segfault (wl_display->registry_resource_list is not init)
+	// wl_global_destroy(screenshooter->wl_global);
```
SirCmpwn: .

examples/screenshot.c
emersion: This part is from weston's screenshooter. If there are multiple outputs, take a screenshot of each one, and then put them together in one image according to each output's position and size. `set_buffer_size` computes the global buffer size, `write_image` combines each screenshot.

SirCmpwn: Gotcha.

SirCmpwn:
Thanks!

PR #142, opened 2017-09-08
add a global filter example to examples/compositor

Ongy:
Adds example for filtering wl_global structs exposed to clients.
This can be used to give access to a specific protocol implemented to
selected clients.
The current implementation is to deny every fifth client.

This is a proof of concept I made after writing https://github.com/SirCmpwn/wlroots/issues/141


SirCmpwn:
Thanks for the patch, but I don't think this needs to be upstreamed. If someone asks how to do this we'll just refer them to this PR.

PR #143, opened 2017-09-08
fixes use after free caused by signal lists

Ongy:
A few structs throughout the code use implementation specific free
functions.
When those functions are not used, they simply call free() on their
data, but this leaves around wl_signals linked into listeners.
When those listeners try to remove themself from the list, they write
into the now free memory.

This commit adds calls to remove the signals from those lists, so the
listeners can safely call wl_list_remove


SirCmpwn:
Thanks!

Issue #130, opened 2017-08-30
DisplayPort MST


ascent12:
This is a tracking issue for DisplayPort MST (Multi-Stream Transport), which is the ability for multiple monitors to come off the same DisplayPort connector. It's also used in Dual-panel 4K monitors and many modern laptop docks.

Unfortunately, I don't have a lot of the hardware to implement this all properly or even be certain about the requirements for it. I DO own a monitor that is MST compatible, but the radeon drivers are messed up with my GPU, so I can't get it working.

Based on my understanding (I may be wrong), what we need to do is
- [x] Handle DRM connectors being able to appear at any point
- [ ] Handle DRM connectors being able to disappear at any point
- [ ] Naming the connectors properly (e.g. DP-1-1, DP-1-2)
- [ ] Being able to handle the DRM tiling property (for dual-panel 4K screens), possibly exposing dual-panels as a single logical monitor?

SirCmpwn:
Does daisy chaning need special consideration?

SirCmpwn:
Also, it might be worthwhile to put down a hardware wishlist of some cheap hardware with these capabilities that someone interested in support for these features might donate to you.

ascent12:
I believe daisy chaining works the same as a MST hub. My bought a Dell U2715H recently, which is daisy chaining compatible, but the problems are with my GPU (HD 7950) and its drivers now. Radeon SHOULD support MST, but it's currently experimental, and doesn't currently work properly for me.

The only other GPUs I have on hand is some old Nvidia (570?) card and an AMD HD 6950. It doesn't have a displayport, and I don't even have a clue about the status of MST for nouveau anyway.

Once the AMDGPU DC stuff lands in the kernel (this may be several releases away?), I may pick up a new cheapish AMD GPU, and that should be able to handle MST properly.

kasicka:
I have a few thinkpads/laptops laying around at home if needed, but it's not bleeding edge hardware.

ascent12:
Regarding naming the connectors, the DRM system seems to handle this pretty well for us. It'll just keep counting on them like other DisplayPort connectors (DP-2, DP3, ...).
There are some weird cases where it doesn't un-number the old connectors when they get unplugged, meaning when I plugged it back in, I got DP-5 and DP-6, when I really should have gotten DP-3 and DP-4. I'm not 100% how to reproduce it, though.

[...]

PR #144, opened 2017-09-09
DisplayPort MST

ascent12:
This implements the basics of DisplayPort MST, as mentioned in #130. Fortunately, due to the way we were already doing things, not a lot of code actually needed to be added.

It _seems_ to work on my specific set of hardware, but this is something I really want others to test. I don't know if there are some assumptions I'm accidentally making based on my setup.

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -26,6 +26,9 @@ static void wlr_drm_backend_destroy(struct wlr_backend *_backend) {
 		return;
 	}
 	struct wlr_drm_backend *backend = (struct wlr_drm_backend *)_backend;
+
+	wlr_drm_restore_outputs(backend);
+
 	for (size_t i = 0; backend->outputs && i < backend->outputs->length; ++i) {
 		struct wlr_drm_output *output = backend->outputs->items[i];
 		wlr_output_destroy(&output->output);
--- a/backend/drm/drm-atomic.c
+++ b/backend/drm/drm-atomic.c
@@ -57,8 +57,7 @@ static bool atomic_commit(int drm_fd, struct atomic *atom,
 
 		// Try to commit without new changes
 		drmModeAtomicSetCursor(atom->req, atom->cursor);
-		ret = drmModeAtomicCommit(drm_fd, atom->req, flags, output);
-		if (ret) {
+		if (drmModeAtomicCommit(drm_fd, atom->req, flags, output)) {
 			wlr_log_errno(L_ERROR, "%s: Atomic commit failed (%s)",
 				output->output.name, modeset ? "modeset" : "pageflip");
 		}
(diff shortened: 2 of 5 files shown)


SirCmpwn:
LGTM. Does anyone have hardware to test this?

kasicka:
Tried on Thinkpad T440s/Fedora Rawhide.
Log: https://pastebin.com/TuZQv5HC
Observation: I started wlroots with docked laptop, monitors were connected. Then I redocked and the monitors didn't get on, only eDP-1 was on. After that I wanted to kill wlroots with Escape, I ended up in tty, but I didn't get prompt nor was I able to switch to different tty and I had to reboot.

ascent12:
Ah, those damn `Atomic commit failed: Device or Resource Busy. I don't know what's actually causing that, although it's not necessarily going to break anything. `Atomic commit failed: Invalid argument` is there too, but that's not actually that big of a deal: it's just we're tried to set a connector that doesn't exist (anymore). Maybe there are some checks we could add to catch that earlier.

Interestingly, there is no `'DP-3' disappeared` or anything like that, which means we never properly closed the renderer for these non-existent displays. When it comes time to clean up, it's getting stuck in an infinite loop trying to stop a renderer which isn't even running anymore.

kasicka:
New log: https://pastebin.com/7EVB9BBM
Observation: Docking/undocking works, killing wlroots does not.

ascent12:
I've made some changes to hopefully deal with errors better. It seems to work fine for me, but I only have a single piece of hardware to test it for.

If you test this, can you also run an example with WLR_DRM_NO_ATOMIC to force the legacy interface, which may give better results in some situations.

SirCmpwn:
Code LGTM. @kasicka mind giving it a shot?

kasicka:
Here are some logs: https://github.com/kasicka/logs/tree/master/wlroots/t450s/docked
I didn't try to set WLR_DRM_NO_ATOMIC for any of the tests.
Also the mouse pointer didn't show up.

ascent12:
These logs seem to imply that some things outside the DRM backend (xwayland in particular) have some use-after-frees/null defere with the wlr_outputs, which could lead to crashes, particularly when you're trying to close wlroots. That's obviously a bug, but it's not related to this PR.

Can you try using `examples/simple` and seeing if that works better? Most of the other things won't end up getting loaded with that, so it will only test the DRM stuff.

ascent12:
We've confirmed that it works with `examples/simple`. I'm pretty confident that the MST handling is working properly, so I think we can go ahead and merge this.

SirCmpwn:
Thanks!

kasicka:
Tried with t460p + nouveau / f27: https://github.com/kasicka/logs/tree/master/wlroots/t460p
My observation: independently on how I start it (docked/undocked), works fine until first undock. After that the screen keeps the color, after redock it still keeps the color and monitors are black and connected and after another redock screen goes black, monitors still connected, stays black even after I kill it from ssh.

ascent12:
@kasicka It looks like wlroots is crashing, considering there is no clean exit in the logs. Can you try with valgrind or gdb and see which line it gets to?

SirCmpwn:
Maybe look for a core dump too?

PR #145, opened 2017-09-12
xdg shell toplevel

acrisci:
In this PR I want to flesh out the xdg shell toplevel.

We can use this as a basis to decide the abstractions for the other shells and what interface is provided to the user.

Our shell implementation will consist simply of a desktop surface state that is changed by the client through the protocol and convenience functions for common things like configure requests. Library users can choose to interpret these values how they want.

Open questions:
* How are users notified of desktop surface state changes?
* Is there enough similarity between the shells (title, geometry, etc.) to make them a subclass of a `wlr_desktop_surface` to make them easier to work with?

### DIFF
--- a/types/wlr_xdg_shell_v6.c
+++ b/types/wlr_xdg_shell_v6.c
@@ -104,6 +104,7 @@ static void xdg_surface_destroy(struct wlr_xdg_surface_v6 *surface) {
 	wl_resource_set_user_data(surface->resource, NULL);
 	wl_list_remove(&surface->link);
 	wl_list_remove(&surface->surface_destroy_listener.link);
+	wl_list_remove(&surface->surface_commit_listener.link);
 	free(surface);
 }
 
@@ -118,9 +119,8 @@ static void xdg_surface_get_toplevel(struct wl_client *client,
 		struct wl_resource *resource, uint32_t id) {
 	// TODO: Flesh out
 	struct wlr_xdg_surface_v6 *surface = wl_resource_get_user_data(resource);
-	struct wlr_surface *wsurface = wl_resource_get_user_data(surface->surface);
 
-	if (wlr_surface_set_role(wsurface, wlr_desktop_xdg_toplevel_role,
+	if (wlr_surface_set_role(surface->surface, wlr_desktop_xdg_toplevel_role,
 			resource, ZXDG_SHELL_V6_ERROR_ROLE)) {
 		return;
 	}
@@ -173,6 +173,11 @@ static void handle_wlr_surface_destroyed(struct wl_listener *listener,
 	xdg_surface_destroy(xdg_surface);
 }
 
+static void handle_wlr_surface_committed(struct wl_listener *listener,
+		void *data) {
+	wlr_log(L_DEBUG, "TODO: handle wlr surface committed");
+}
+
 static void xdg_shell_get_xdg_surface(struct wl_client *client,
 		struct wl_resource *_xdg_shell, uint32_t id,
 		struct wl_resource *_surface) {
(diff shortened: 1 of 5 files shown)


types/wlr_xdg_shell_v6.c
```diff
@@ -91,19 +120,40 @@ static const struct zxdg_toplevel_v6_interface zxdg_toplevel_v6_implementation =
...
 }
 
+static void xdg_surface_resource_destroy(struct wl_resource *resource) {
+	struct wlr_xdg_surface_v6 *surface = wl_resource_get_user_data(resource);
+	if (surface != NULL) {
+		xdg_surface_destroy(surface);
+	}
+}
+
+static const char* wlr_desktop_xdg_toplevel_role = "xdg_toplevel";
```
SirCmpwn: Style nit, move the * over

types/wlr_xdg_shell_v6.c
```diff
@@ -91,19 +120,40 @@ static const struct zxdg_toplevel_v6_interface zxdg_toplevel_v6_implementation =
...
+
 static void xdg_surface_get_toplevel(struct wl_client *client,
 		struct wl_resource *resource, uint32_t id) {
 	// TODO: Flesh out
```
SirCmpwn: Is this TODO still valid?

SirCmpwn:
LGTM.

>How are users notified of desktop surface state changes?

With wl_signals on struct wlr_xdg_surface_v6?

>Is there enough similarity between the shells (title, geometry, etc.) to make them a subclass of a wlr_desktop_surface to make them easier to work with?

I'd rather not.

acrisci:
> With wl_signals on struct wlr_xdg_surface_v6?

There's a certain point where that becomes awkward but we may not actually reach that point. We'll try it first and see how it feels.

> I'd rather not.

I think once we do the second shell you might see why I'm bringing this up.

types/wlr_xdg_shell_v6.c
acrisci: I think the only thing left to do here is initialize the toplevel state.

acrisci:
Actually, now that I'm looking at it, almost everything on the xdg-surface is double buffered, so we could just have a single event that triggers after the xdg state is committed (after the wlr_surface is committed). Although that wouldn't tell people exactly what changed.

acrisci:
Now we can do configure requests to set the size and alert a toplevel to state changes.

Next we need a way to know when new desktop shell surfaces (toplevel or popup) appears so we can begin to configure them. Maybe an event on the shell?

SirCmpwn:
Yep, event sounds like the right approach.

acrisci:
Here is a test script for the request events. I can't do one for the window menu yet because I have to figure out how to simulate a button press, but it should be the same.

```python
#!/usr/bin/env python3

import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk, GLib, Gdk
import signal

signal.signal(signal.SIGINT, signal.SIG_DFL)

w = Gtk.Window()
w.connect('destroy', Gtk.main_quit)
w.props.title = 'test-window'
w.show()

def timeout_func(data):
	w.begin_move_drag(1, 0, 0, 0)
	w.begin_resize_drag(Gdk.WindowEdge.NORTH_WEST, 1, 0, 0, 0)
	w.iconify()

	GLib.timeout_add(1000, timeout_func, None)

GLib.timeout_add(1000, timeout_func, None)

Gtk.main()
```

acrisci:
Ok, I think this is ok for now and ready for review. I'm sure we will find some bugs once we do the integrations though.

types/wlr_xdg_shell_v6.c
```diff
@@ -166,23 +630,73 @@ static struct zxdg_shell_v6_interface xdg_shell_impl = {
...
+static void wlr_xdg_client_v6_destroy(struct wl_resource *resource) {
+	struct wlr_xdg_client_v6 *client = wl_resource_get_user_data(resource);
+	struct wl_list *list = &client->surfaces;
+	struct wl_list *link, *tmp;
+
+	for (link = list->next, tmp = link->next;
+			link != list;
+			link = tmp, tmp = link->next) {
+		wl_list_remove(link);
+		wl_list_init(link);
```
SirCmpwn: Any particular reason this isn't just wl_list_foreach?

SirCmpwn:
Looks great, awesome work!

types/wlr_xdg_shell_v6.c
acrisci: I lifted this from Weston, but it does look pretty much identical to `wl_list_for_each_safe()`.

acrisci: That seems to work ok.

SirCmpwn:
Great, thanks!

PR #149, opened 2017-09-20
wlr-seat pointer events

acrisci:
Implement the wlr-seat pointer events and an example usage in compositor.c.

Feedback wanted on the api.

Test it out by clicking the buttons in gnome-calculator. (if you hover for too long on a button, it will crash but that is likely not related to pointer events.)

### DIFF
--- a/examples/compositor.c
+++ b/examples/compositor.c
@@ -17,6 +17,7 @@
 #include <wlr/types/wlr_wl_shell.h>
 #include <wlr/types/wlr_xdg_shell_v6.h>
 #include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_input_device.h>
 #include <wlr/types/wlr_data_device_manager.h>
 #include <wlr/types/wlr_gamma_control.h>
 #include "wlr/types/wlr_compositor.h"
@@ -27,6 +28,7 @@
 #include <wlr/util/log.h>
 #include "config.h"
 #include "shared.h"
+#include <assert.h>
 
 // TODO: move to common header?
 int os_create_anonymous_file(off_t size);
@@ -58,6 +60,8 @@ struct sample_state {
 	struct wl_listener cursor_axis;
 
 	struct wl_listener new_xdg_surface_v6;
+
+	struct wlr_xdg_surface_v6 *focused_surface;
 };
 
 struct example_xdg_surface_v6 {
@@ -75,6 +79,25 @@ struct example_xdg_surface_v6 {
 	struct wl_listener request_resize_listener;
 	struct wl_listener request_show_window_menu_listener;
 };
+
+static void example_set_focused_surface(struct sample_state *sample,
+		struct wlr_xdg_surface_v6 *surface) {
+	if (sample->focused_surface == surface) {
+		return;
+	}
+
+	// set activated state of the xdg surfaces
+	struct wlr_xdg_surface_v6 *xdg_surface;
+	struct wlr_xdg_client_v6 *xdg_client;
+	wl_list_for_each(xdg_client, &sample->xdg_shell->clients, link) {
+		wl_list_for_each(xdg_surface, &xdg_client->surfaces, link) {
+			wlr_xdg_toplevel_v6_set_activated(xdg_surface,
+				xdg_surface == surface);
+		}
+	}
+
+	sample->focused_surface = surface;
+}
 
 /*
  * Convert timespec to milliseconds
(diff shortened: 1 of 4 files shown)


SirCmpwn:
Hmm, gnome-calculator causes the compositor to crash for me.

SirCmpwn:
```
Thread 1 "compositor" received signal SIGSEGV, Segmentation fault.
0x0000555555557b4e in handle_output_frame (output=0x5555558d30b0, ts=0x7fffffffd9d0) at ../examples/compositor.c:259
259					esurface->position.lx + width, esurface->position.ly + height);
(gdb) bt
#0  0x0000555555557b4e in handle_output_frame (output=0x5555558d30b0, ts=0x7fffffffd9d0) at ../examples/compositor.c:259
#1  0x000055555555a089 in output_frame_notify (listener=0x5555558d30c0, data=0x55555588f7c0) at ../examples/shared.c:437
#2  0x00007ffff7bae27f in wl_signal_emit (signal=0x55555588f8a0, data=0x55555588f7c0) at /usr/include/wayland-server-core.h:388
#3  0x00007ffff7bae4b7 in surface_frame_callback (data=0x55555588f7c0, cb=0x5555558fd360, time=3048196)
    at ../backend/wayland/output.c:23
#4  0x00007ffff52351c8 in ffi_call_unix64 () at /usr/lib/libffi.so.6
#5  0x00007ffff5234c2a in ffi_call () at /usr/lib/libffi.so.6
#6  0x00007ffff7180bad in  () at /usr/lib/libwayland-client.so.0
#7  0x00007ffff717d679 in  () at /usr/lib/libwayland-client.so.0
#8  0x00007ffff717e9b4 in wl_display_dispatch_queue_pending () at /usr/lib/libwayland-client.so.0
#9  0x00007ffff7bad88b in dispatch_events (fd=7, mask=1, data=0x5555557e60a0) at ../backend/wayland/backend.c:20
#10 0x00007ffff79879b2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#11 0x00007ffff798614c in wl_display_run () at /usr/lib/libwayland-server.so.0
#12 0x0000555555558afa in main (argc=1, argv=0x7fffffffe3e8) at ../examples/compositor.c:562
```

SirCmpwn:
Looks like esurface is NULL sometimes. Adding a NULL check that just continues fixes that issue.

SirCmpwn:
Tested this on the DRM backend and it worked, but I couldn't click the close button on gnome-calculator's titlebar (it unfocused the surface instead). On the Wayland backend, I couldn't focus gnome-calculator at all.

SirCmpwn:
Code and API looks good.

acrisci:
> Looks like esurface is NULL sometimes.

Toplevels (the only surfaces we are concerned about right now) should always have an esurface because it's assigned after the toplevel is created, so I'll have to look into that.

> Tested this on the DRM backend and it worked, but I couldn't click the close button on gnome-calculator's titlebar

The close button works for me, but I think the coordinates we are sending might be off, possibly due some rounding that happens in the cursor coordinates. I need to fix that.

acrisci:
Here is a little [gtk app](https://github.com/acrisci/scribbles/blob/master/gtk-window-helpers/pointer-helper.py) you can use to test the pointer. Compare the output to other window managers.

This reveals some problems with the pointer coordinates we are sending.

acrisci:
Ok I think I fixed all the issues. There's probably some cleanup that needs to be done, and implement sending the pointer events for absolute pointer motion.

The pointer crossing region is now flush against the window geometry. This is how wlc does things. Weston seems to be leaving about a 10px padding around the window geometry for resize handles, but I'm not sure if I like that.

acrisci:
The segfault was caused by unconfigured toplevel windows. We send the toplevel new event after the first ack configure because this is the point at which the window should be mapped. So it is possible for a window to have the toplevel role and be considered unmapped, which is something to be aware of.

SirCmpwn:
Still does not work on the Wayland backend.

SirCmpwn:
>The pointer crossing region is now flush against the window geometry. This is how wlc does things. Weston seems to be leaving about a 10px padding around the window geometry for resize handles, but I'm not sure if I like that.

I agree with your decision. Since this is in the reference compositor, other compositors can do it differently if they like.

acrisci:
Ok it should work on the wayland backend now.

examples/compositor.c
```diff
@@ -431,6 +437,31 @@ static void handle_cursor_button(struct wl_listener *listener, void *data) {
 		event->button, event->state);
 }
 
+static void handle_tool_axis(struct wl_listener *listener, void *data) {
+	struct sample_state *sample =
+		wl_container_of(listener, sample, tool_axis);
+	struct wlr_event_tablet_tool_axis *event = data;
+	if ((event->updated_axes & WLR_TABLET_TOOL_AXIS_X) &&
```
acrisci: Can we get a case where either one or the other axis is updated?

SirCmpwn: Not that I've ever observed. I suppose it could happen in theory.

PR #152, opened 2017-09-22
wlr-seat cursor axis events

acrisci:
Test by doing a mouse wheel over this [pointer test window](https://github.com/acrisci/scribbles/blob/master/gtk-window-helpers/pointer-helper.py).


SirCmpwn:
LGTM, will merge when Travis agrees

PR #154, opened 2017-09-22
Add elogind support

### LINKED ISSUE #146: elogind support
I noticed cloudef/wlc#279 and we could easily do that as well.
I don't want to blindly add code I can't test, so if someone on a non-systemd system could try this and see if it doesn't break anything, that would be great.

Dudemanguy911:
Resolves #146

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Install dependencies:
 * udev
 * pixman
 * systemd (optional, for logind support)
-* elogind (optional, for logind support)
+* elogind (optional, for logind support on systems without systemd)
 * libcap (optional, for capability support)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -25,6 +25,7 @@ Install dependencies:
 * udev
 * pixman
 * systemd (optional, for logind support)
+* elogind (optional, for logind support)
```
SirCmpwn: Can you clarify that this is an xor dependency?

Dudemanguy911:
Is that OK?

SirCmpwn:
Yep, thanks!

PR #160, opened 2017-09-24
Subcompositor

acrisci:
Test with `weston-subsurfaces` and clicking menus on gnome-calculator.

Currently crashes with `weston-subsurfaces -r1` and on synchronized subsurfaces on surface commits (the green circle) because I'm forgetting to clean up a buffer somewhere, but it's pretty close.

One of the buttons on gnome-calculator will try to get an xdg positioning anchor, so after this is ready, we can actually do the xdg-toplevel.

I wonder if there is any client that actually uses arbitrarily nested subsurfaces and depends on all the little rules.

TODO:
- [x] fix `weston-subsurfaces -r1 -t1`
- [x] treat xdg-toplevel and subsurfaces as one window for input/seat events in Rootston
- [x] <s>find and test a client with subsurface depth > 1</s>
- [x] clean up surface commit
- [x] subsurface stacking requests
- [x] subsurface interface errors
- [x] subsurface parent destroy

### DIFF
--- a/include/rootston/desktop.h
+++ b/include/rootston/desktop.h
@@ -52,7 +52,8 @@ struct roots_desktop *desktop_create(struct roots_server *server,
 void desktop_destroy(struct roots_desktop *desktop);
 
 void view_destroy(struct roots_view *view);
-struct roots_view *view_at(struct roots_desktop *desktop, int x, int y);
+struct roots_view *view_at(struct roots_desktop *desktop, double lx, double ly,
+		struct wlr_surface **surface, double *sx, double *sy);
 void view_activate(struct roots_view *view, bool activate);
 
 void output_add_notify(struct wl_listener *listener, void *data);
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -18,6 +18,8 @@ struct wlr_frame_callback {
 #define WLR_SURFACE_INVALID_INPUT_REGION 16
 #define WLR_SURFACE_INVALID_TRANSFORM 32
 #define WLR_SURFACE_INVALID_SCALE 64
+#define WLR_SURFACE_INVALID_SUBSURFACE_POSITION 128
+#define WLR_SURFACE_INVALID_FRAME_CALLBACK_LIST 256
 
 struct wlr_surface_state {
 	uint32_t invalid;
@@ -29,6 +31,12 @@ struct wlr_surface_state {
 	int32_t scale;
 	int width, height;
 	int buffer_width, buffer_height;
+
+	struct {
+		int32_t x, y;
+	} subsurface_position;
+
+	struct wl_list frame_callback_list; // wl_surface.frame
 };
 
 struct wlr_subsurface {
@@ -36,25 +44,19 @@ struct wlr_subsurface {
 	struct wlr_surface *surface;
 	struct wlr_surface *parent;
 
-	struct wlr_surface_state cached;
-
-	struct {
-		int32_t x, y;
-	} position;
-
-	struct {
-		int32_t x, y;
-		bool set;
-	} pending_position;
+	struct wlr_surface_state *cached;
+	bool has_cache;
 
 	bool synchronized;
+
+	struct wl_list parent_link;
 };
 
 struct wlr_surface {
 	struct wl_resource *resource;
 	struct wlr_renderer *renderer;
 	struct wlr_texture *texture;
-	struct wlr_surface_state current, pending;
+	struct wlr_surface_state *current, *pending;
 	const char *role; // the lifetime-bound role or null
 
 	float buffer_to_surface_matrix[16];
@@ -66,13 +68,13 @@ struct wlr_surface {
 		struct wl_signal destroy;
 	} signals;
 
-	struct wl_list frame_callback_list; // wl_surface.frame
 	// destroy listener used by compositor
 	struct wl_listener compositor_listener;
 	void *compositor_data;
 
 	// subsurface properties
 	struct wlr_subsurface *subsurface;
+	struct wl_list subsurface_list; // wlr_subsurface::parent_link
 
 	void *data;
 };
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -19,6 +19,12 @@ struct wlr_frame_callback {
 #define WLR_SURFACE_INVALID_TRANSFORM 32
 #define WLR_SURFACE_INVALID_SCALE 64
 
+struct wlr_subsurface {
+	struct wl_resource *resource;
+	struct wlr_surface *surface;
+	struct wlr_surface *parent;
+};
+
 struct wlr_surface_state {
 	uint32_t invalid;
 	struct wl_resource *buffer;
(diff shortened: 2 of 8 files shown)


SirCmpwn:
Initial test looks good, but subsurfaces aren't receiving pointer events (presumably this is just missing from your todo list). I also got gnome-calculator to crash by clicking the decimal/hex/binary dropdown in programming mode.

acrisci:
> Initial test looks good, but subsurfaces aren't receiving pointer events

I consider that to be Rootston's responsibility because we don't have a wlr view abstraction. That's todo number two.

> I also got gnome-calculator to crash by clicking the decimal/hex/binary dropdown in programming mode.

That's the xdg anchor request. That button needs xdg-toplevel to not crash.

SirCmpwn:
>I consider that to be Rootston's responsibility because we don't have a wlr view abstraction. That's todo number two.

So I guess you're not putting that into the example compositor? Fair enough I suppose.

acrisci:
This is a pretty heavy refactoring of the surface logic to handle all the new edge cases that weston-subsurfaces presented me.

Now we have three surface states.

1. `current` is effective state of the surface.
2. `pending` is the state that the client gives us through the surface interface.
3. `cached` is a cached state for synchronized subsurfaces.

A synchronized subsurface is "synchronized" to its parent surface commit cycle. When a synchronized subsurface, commits its state, it has to wait for the parent to commit its state for that state to apply. We save state from pending state in the cached state when the subsurface commits. Then when the parent commits, we move the cached state back into the pending state and then commit the pending state (sum of the cached and pending state).

To facilitate state movement, we have a new function called `wlr_surface_move_state()` to move the states around. Note that frame callbacks are now double buffered and get moved around with the state.

I found a bug in `wlr_surface_flush_damage()` where it was uploading the pending buffer (which wasn't being cleared when pending was moved to current) instead of the current buffer which I fixed.

Another edge case is when a client commits a buffer to the subsurface cache twice. This edge case might also be present if the client commits two buffers very quickly and we don't get a chance to flush the current buffer (that case is complicated and still needs some testing). Weston solves the problem with a fairly complicated reference counting scheme, but I managed to avoid that with the centralized move function that should guarantee that the buffer is only in one state at a time. If the backend ever needs to hold onto buffers, we will need to start using reference counting.

To avoid leaking buffers, it is necessary to release a buffer when moving a buffer into a state that already has a buffer. This is because weston-subsurfaces client will attempt to commit a buffer to the subsurface cache twice in a row. This is allowed by the spec and we are required to handle this situation. The only time we don't have to send release is if the client attaches a buffer twice without a commit in between.

acrisci:
Ok I think this is ready to go.

I don't know if it will work for arbitrarily nested subsurfaces because I can't find test clients. I am leaving that testing to my xdg-toplevel implementation where I believe there is a good chance I will find nested subsurfaces (in menu trees). If I don't, I will write a test client and sort out the mess later.

SirCmpwn:
Doesn't any GTK3 menu or context menu with nested entries qualify? I think gnome-terminal has some. If you compile GTK from source there's also a bunch of GTK test programs available.

SirCmpwn:
Sorry, can you clarify the role of the various kinds of surfaces in an xdg shell? toplevel et al

acrisci:
> Doesn't any GTK3 menu or context menu with nested entries qualify? I think gnome-terminal has some.

I think it will but I'm not sure. If you click those menus in gnome-terminal, you crash because it needs `zxdg_shell_v6_create_positioner` to give it a positioner. This is part of the xdg-toplevel interface.

> GTK from source there's also a bunch of GTK test programs available.

Thanks, I've been mostly writing my own test clients so this will help.

> Sorry, can you clarify the role of the various kinds of surfaces in an xdg shell? toplevel et al

The documentation lives [here](https://github.com/wayland-project/wayland-protocols/blob/master/unstable/xdg-shell/xdg-shell-unstable-v6.xml).

A toplevel is a "main window". I can't really compare/contrast it to a popup because I haven't completely read through the interface yet. I also don't know the relationship between subsurfaces and popups if there is one, but they seem very similar in scope.

An xdg-popup is something like a tooltip or a right click context menu. I will do a more detailed write-up of them when I do that interface like I did here.

SirCmpwn:
Alright, cool. I will review+test+merge this PR in a little while. Thanks!

SirCmpwn:
I'm having some issues with weston-subsurfaces while testing this. If I pass -r1 -t1, it can segfault rootston when the client quits.

```
hread 1 "rootston" received signal SIGSEGV, Segmentation fault.
0x00007ffff73fdfd9 in malloc_consolidate () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff73fdfd9 in malloc_consolidate () at /usr/lib/libc.so.6
#1  0x00007ffff7400f52 in _int_malloc () at /usr/lib/libc.so.6
#2  0x00007ffff740367a in calloc () at /usr/lib/libc.so.6
#3  0x00007fffef6df653 in  () at /usr/lib/xorg/modules/dri/nouveau_dri.so
#4  0x00007fffef66dfad in  () at /usr/lib/xorg/modules/dri/nouveau_dri.so
#5  0x00007fffef674a95 in  () at /usr/lib/xorg/modules/dri/nouveau_dri.so
#6  0x00007fffef650b49 in  () at /usr/lib/xorg/modules/dri/nouveau_dri.so
#7  0x00007fffef650bf2 in  () at /usr/lib/xorg/modules/dri/nouveau_dri.so
#8  0x00007fffef53d1d1 in  () at /usr/lib/xorg/modules/dri/nouveau_dri.so
#9  0x00007ffff7bafbdd in wlr_gles2_begin (_renderer=0x555555801be0, output=0x5555558bc760) at ../render/gles2/renderer.c:129
#10 0x00007ffff7bb1399 in wlr_renderer_begin (r=0x555555801be0, o=0x5555558bc760) at ../render/wlr_renderer.c:19
#11 0x000055555555a7ce in output_frame_notify (listener=0x5555558bb040, data=0x5555558bc760) at ../rootston/output.c:72
#12 0x00007ffff7ba8b44 in wl_signal_emit (signal=0x5555558bc840, data=0x5555558bc760) at /usr/include/wayland-server-core.h:388
#13 0x00007ffff7ba8d7c in surface_frame_callback (data=0x5555558bc760, cb=0x5555558dd810, time=4245498)
    at ../backend/wayland/output.c:23
#14 0x00007ffff4cee1c8 in ffi_call_unix64 () at /usr/lib/libffi.so.6
#15 0x00007ffff4cedc2a in ffi_call () at /usr/lib/libffi.so.6
#16 0x00007ffff7179bad in  () at /usr/lib/libwayland-client.so.0
#17 0x00007ffff7176679 in  () at /usr/lib/libwayland-client.so.0
#18 0x00007ffff71779b4 in wl_display_dispatch_queue_pending () at /usr/lib/libwayland-client.so.0
#19 0x00007ffff7ba8143 in dispatch_events (fd=4, mask=1, data=0x55555576ffa0) at ../backend/wayland/backend.c:20
#20 0x00007ffff79809b2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#21 0x00007ffff797f14c in wl_display_run () at /usr/lib/libwayland-server.so.0
#22 0x000055555555a2ff in main (argc=1, argv=0x7fffffffe3e8) at ../rootston/main.c:44
```

Pretty weird stacktrace.

SirCmpwn:
I can't reproduce this on the DRM backend. Probably a separate problem.

SirCmpwn:
LGTM, thanks!

PR #161, opened 2017-09-25
unset egl current before terminating display

acrisci:
fixes segfault when closing the compositor with an egl window.


SirCmpwn:
Thanks!

PR #162, opened 2017-09-26
Add wl_shell to rootston

emersion:
Test client: https://gist.github.com/emersion/7e453eb466beb3c2958c042ccf239ba7

### DIFF
--- a/include/wlr/types/wlr_wl_shell.h
+++ b/include/wlr/types/wlr_wl_shell.h
@@ -103,5 +103,8 @@ struct wlr_wl_shell *wlr_wl_shell_create(struct wl_display *display);
 void wlr_wl_shell_destroy(struct wlr_wl_shell *wlr_wl_shell);
 
 void wlr_wl_shell_surface_ping(struct wlr_wl_shell_surface *surface);
+void wlr_wl_shell_surface_configure(struct wlr_wl_shell_surface *surface,
+	uint32_t edges, int32_t width, int32_t height);
+void wlr_wl_shell_surface_popup_done(struct wlr_wl_shell_surface *surface);
 
 #endif
--- a/include/wlr/types/wlr_wl_shell.h
+++ b/include/wlr/types/wlr_wl_shell.h
@@ -1,6 +1,7 @@
 #ifndef WLR_TYPES_WLR_WL_SHELL_H
 #define WLR_TYPES_WLR_WL_SHELL_H
 
+#include <stdbool.h>
 #include <wayland-server.h>
 
 struct wlr_wl_shell {
@@ -23,6 +24,18 @@ struct wlr_wl_shell_surface_transient_state {
 	uint32_t flags;
 };
 
+struct wlr_wl_shell_surface_popup_state {
+	struct wlr_seat_handle *seat_handle;
+	uint32_t serial;
+};
+
+enum wlr_wl_shell_surface_role {
+	WLR_WL_SHELL_SURFACE_ROLE_NONE,
+	WLR_WL_SHELL_SURFACE_ROLE_TOPLEVEL,
+	WLR_WL_SHELL_SURFACE_ROLE_TRANSCIENT,
+	WLR_WL_SHELL_SURFACE_ROLE_POPUP,
+};
+
 struct wlr_wl_shell_surface {
 	struct wlr_wl_shell *shell;
 	struct wl_client *client;
@@ -33,8 +46,10 @@ struct wlr_wl_shell_surface {
 	uint32_t ping_serial;
 	struct wl_event_source *ping_timer;
 
-	bool toplevel;
+	enum wlr_wl_shell_surface_role role;
 	struct wlr_wl_shell_surface_transient_state *transient_state;
+	struct wlr_wl_shell_surface_popup_state *popup_state;
+
 	char *title;
 	char *class_;
 
@@ -45,10 +60,9 @@ struct wlr_wl_shell_surface {
 		struct wl_signal request_move;
 		struct wl_signal request_resize;
 		struct wl_signal request_set_fullscreen;
-		struct wl_signal request_set_popup;
 		struct wl_signal request_set_maximized;
-		struct wl_signal set_toplevel;
-		struct wl_signal set_transient;
+
+		struct wl_signal set_role;
 		struct wl_signal set_title;
 		struct wl_signal set_class;
 	} events;
--- a/include/wlr/types/wlr_wl_shell.h
+++ b/include/wlr/types/wlr_wl_shell.h
@@ -53,6 +53,8 @@ struct wlr_wl_shell_surface {
 	char *title;
 	char *class;
 
+	struct wl_listener surface_destroy_listener;
+
 	struct {
 		struct wl_signal destroy;
 		struct wl_signal ping_timeout;
--- a/include/wlr/types/wlr_wl_shell.h
+++ b/include/wlr/types/wlr_wl_shell.h
@@ -7,20 +7,35 @@ struct wlr_wl_shell {
 	struct wl_global *wl_global;
 	struct wl_list wl_resources;
 	struct wl_list surfaces;
+	uint32_t ping_timeout;
+
+	struct {
+		struct wl_signal new_surface;
+	} events;
 
 	void *data;
 };
 
 struct wlr_wl_shell_surface {
+	struct wlr_wl_shell *shell;
+	struct wl_client *client;
 	struct wl_resource *surface;
 	struct wlr_texture *wlr_texture;
 	struct wl_list link;
 
+	uint32_t ping_serial;
+	struct wl_event_source *ping_timer;
+
+	struct {
+		struct wl_signal ping_timeout;
+	} events;
+
 	void *data;
 };
 
-
 struct wlr_wl_shell *wlr_wl_shell_create(struct wl_display *display);
 void wlr_wl_shell_destroy(struct wlr_wl_shell *wlr_wl_shell);
+
+void wlr_wl_shell_surface_ping(struct wlr_wl_shell_surface *surface);
 
 #endif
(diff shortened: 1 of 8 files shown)


include/wlr/types/wlr_wl_shell.h
```diff
@@ -26,13 +33,58 @@ struct wlr_wl_shell_surface {
 	uint32_t ping_serial;
 	struct wl_event_source *ping_timer;
 
+	bool toplevel;
+	struct wlr_wl_shell_surface_transient_state *transient_state;
+	char *title;
+	char *class_;
```
SirCmpwn: Why the underscore?

rootston/wl_shell.c
```diff
@@ -10,17 +10,30 @@
...
 void handle_wl_shell_surface(struct wl_listener *listener, void *data) {
 	struct roots_desktop *desktop =
 		wl_container_of(listener, desktop, wl_shell_surface);
 
 	struct wlr_wl_shell_surface *surface = data;
-	wlr_log(L_DEBUG, "new wl_shell surface");
-	//wlr_wl_shell_surface_ping(surface);
+	wlr_log(L_DEBUG, "new shell surface: title=%s, class=%s",
+		surface->title, surface->class_);
+	//wlr_wl_shell_surface_ping(surface); // TODO: segfaults
```
SirCmpwn: FWIW I think you could implement the whole ping/pong workflow entirely in wlr_wl_shell, no need to expose this function (though I understand this wasn't part of your changes).

types/wlr_wl_shell.c
```diff
@@ -18,32 +20,128 @@ static void shell_surface_pong(struct wl_client *client, struct wl_resource
...
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	event->client = client;
+	event->surface = surface;
+	event->seat_handle = seat_handle;
+	event->serial = serial;
+
+	wl_signal_emit(&surface->events.request_move, event);
```
SirCmpwn: Reminder that you should have the rootston side of things listen to this and attach the view to the cursor. This may be further down in your changes, I'll delete this if I find out.

include/wlr/types/wlr_wl_shell.h
emersion: It was there in the argument list. I guess it breaks the build if using wlroots from C++.

SirCmpwn: I don't care about C++, change it.

rootston/wl_shell.c
emersion: Hmm, I had the feeling compositors might want to call `wlr_wl_shell_surface_ping` to check that a surface is responsive. When is the compositor supposed to call ping? Also, this is exposed in xdg shell too. Should we remove it from xdg shell and wl_shell?

SirCmpwn: We can instead expose an event for when we believe the surface has become unresponsive.

SirCmpwn:
LGTM. @acrisci may have some feedback, but let's address it in a separate PR so we can get rootston merged.

SirCmpwn:
Thanks!

include/wlr/types/wlr_wl_shell.h
```diff
@@ -1,26 +1,112 @@
...
+
+struct wlr_wl_shell_surface_popup_state {
+	struct wlr_seat_handle *seat_handle;
+	uint32_t serial;
+};
+
+enum wlr_wl_shell_surface_role {
+	WLR_WL_SHELL_SURFACE_ROLE_NONE,
+	WLR_WL_SHELL_SURFACE_ROLE_TOPLEVEL,
+	WLR_WL_SHELL_SURFACE_ROLE_TRANSCIENT,
```
acrisci: typo `s/TRANSCIENT/TRANSIENT`

types/wlr_wl_shell.c
```diff
@@ -1,61 +1,264 @@
...
+		wl_resource_get_user_data(seat_resource);
+	struct wlr_wl_shell_surface *parent =
+		wl_resource_get_user_data(parent_resource);
+	// TODO: check if parent_resource == NULL?
+
+	if (surface->role != WLR_WL_SHELL_SURFACE_ROLE_NONE) {
+		return;
+	}
+
+	struct wlr_wl_shell_surface_transient_state *transcient_state =
```
acrisci: typo with transcient

include/wlr/types/wlr_wl_shell.h
```diff
@@ -1,26 +1,112 @@
...
+	int32_t y;
+	uint32_t flags;
+};
+
+struct wlr_wl_shell_surface_popup_state {
+	struct wlr_seat_handle *seat_handle;
+	uint32_t serial;
+};
+
+enum wlr_wl_shell_surface_role {
```
acrisci: There are not "roles" but rather "states".

types/wlr_wl_shell.c
```diff
@@ -1,61 +1,264 @@
...
+	free(event);
 }
 
 static void shell_surface_set_toplevel(struct wl_client *client,
 		struct wl_resource *resource) {
-	wlr_log(L_DEBUG, "TODO: implement shell surface set_toplevel");
+	wlr_log(L_DEBUG, "got shell surface toplevel");
+	struct wlr_wl_shell_surface *surface = wl_resource_get_user_data(resource);
+
+	if (surface->role != WLR_WL_SHELL_SURFACE_ROLE_NONE) {
```
acrisci: You should be able to set a surface that was previously in another state to a toplevel. It should undo the previous state things like the parent and all that.

types/wlr_wl_shell.c
```diff
@@ -1,61 +1,264 @@
...
-	wlr_log(L_DEBUG, "TODO: implement shell surface set_transient");
+		struct wl_resource *resource, struct wl_resource *parent_resource,
+		int32_t x, int32_t y, uint32_t flags) {
+	wlr_log(L_DEBUG, "got shell surface transient");
+	struct wlr_wl_shell_surface *surface = wl_resource_get_user_data(resource);
+	struct wlr_wl_shell_surface *parent =
+		wl_resource_get_user_data(parent_resource);
+	// TODO: check if parent_resource == NULL?
+
+	if (surface->role != WLR_WL_SHELL_SURFACE_ROLE_NONE) {
```
acrisci: We could probably use a function that changes the state to handle changes between the states. This is allowed.

types/wlr_wl_shell.c
emersion: Okay. I suppose it's also possible to transform a toplevel to a popup/transcient?

types/wlr_wl_shell.c
emersion: What do you mean?

acrisci: We need a way to change from one state to any other state.

include/wlr/types/wlr_wl_shell.h
```diff
@@ -1,26 +1,112 @@
...
+	struct wlr_wl_shell_surface *surface;
+	struct wlr_seat_handle *seat_handle;
+	uint32_t serial;
+	uint32_t edges;
+};
+
+struct wlr_wl_shell_surface_set_fullscreen_event {
+	struct wl_client *client;
+	struct wlr_wl_shell_surface *surface;
+	uint32_t method;
```
acrisci: `wl-server-protocol.h` provides enums for this sort of thing. In this case `enum wl_shell_surface_fullscreen_method`. We should probably be using them.

acrisci:
What test client did you land on for the wl_shell_surface?

emersion:
This one: https://gist.github.com/emersion/7e453eb466beb3c2958c042ccf239ba7

types/wlr_wl_shell.c
```diff
@@ -71,27 +274,86 @@ struct wl_shell_surface_interface shell_surface_interface = {
 	.set_class = shell_surface_set_class,
 };
 
-static void destroy_shell_surface(struct wl_resource *resource) {
-	struct wlr_wl_shell_surface *state = wl_resource_get_user_data(resource);
-	wl_list_remove(&state->link);
-	free(state);
+static void wl_shell_surface_destroy(struct wlr_wl_shell_surface *surface) {
```
acrisci: need to clean up the ping timer

types/wlr_wl_shell.c
```diff
@@ -71,27 +274,86 @@ struct wl_shell_surface_interface shell_surface_interface = {
...
+	return 1;
 }
 
 static void wl_shell_get_shell_surface(struct wl_client *client,
 		struct wl_resource *resource, uint32_t id,
-		struct wl_resource *surface) {
-	struct wlr_texture *wlr_texture = wl_resource_get_user_data(surface);
-	struct wlr_wl_shell *wlr_wl_shell = wl_resource_get_user_data(resource);
-	struct wlr_wl_shell_surface *state =
+		struct wl_resource *surface_resource) {
```
acrisci: need to use `wlr_surface_set_role()` to set the role to "wl_shell_surface".

types/wlr_wl_shell.c
```diff
@@ -1,61 +1,264 @@
...
+	event->method = method;
+	event->framerate = framerate;
+	event->output = output;
+
+	wl_signal_emit(&surface->events.request_set_fullscreen, event);
+
+	free(event);
 }
 
 static void shell_surface_set_popup(struct wl_client *client,
```
acrisci: We need to do a pointer grab here, but that's not implemented in the seat yet so I don't want to worry about it. Maybe just add a TODO for it.

rootston/wl_shell.c
```diff
@@ -0,0 +1,84 @@
...
+}
+
+void handle_wl_shell_surface(struct wl_listener *listener, void *data) {
+	struct roots_desktop *desktop =
+		wl_container_of(listener, desktop, wl_shell_surface);
+
+	struct wlr_wl_shell_surface *surface = data;
+	wlr_log(L_DEBUG, "new shell surface: title=%s, class=%s",
+		surface->title, surface->class);
+	wlr_wl_shell_surface_ping(surface); // TODO: segfaults
```
acrisci: remove todo?

PR #164, opened 2017-09-28
X11 backend

ascent12:
An X11 backend.
It's deliberately been written to be pretty basic. I don't really care enough to write something fancy for this.

There is a segfault that sometimes happens when you click on a blank `examples/compositor`. I don't really know enough about that code to fix it.

### DIFF
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -297,7 +297,7 @@ static bool wlr_x11_backend_start(struct wlr_backend *backend) {
 	wl_signal_emit(&x11->backend.events.input_add, &x11->keyboard_dev);
 	wl_signal_emit(&x11->backend.events.input_add, &x11->pointer_dev);
 
-	signal_frame(x11);
+	wl_event_source_timer_update(x11->frame_timer, 16);
 
 	return true;
 }
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -1,6 +1,8 @@
+#define _POSIX_C_SOURCE 199309L
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <time.h>
 #include <EGL/egl.h>
 #include <wayland-server.h>
 #include <xcb/xcb.h>
@@ -9,10 +11,16 @@
 #include <wlr/backend/x11.h>
 #include <wlr/egl.h>
 #include <wlr/interfaces/wlr_output.h>
+#include <wlr/interfaces/wlr_input_device.h>
+#include <wlr/interfaces/wlr_keyboard.h>
+#include <wlr/interfaces/wlr_pointer.h>
 #include <wlr/util/log.h>
 #include "backend/x11.h"
 
 static struct wlr_backend_impl backend_impl;
+static struct wlr_input_device_impl input_impl;
+static struct wlr_keyboard_impl keyboard_impl;
+static struct wlr_pointer_impl pointer_impl;
 
 int x11_event(int fd, uint32_t mask, void *data) {
 	struct wlr_x11_backend *x11 = data;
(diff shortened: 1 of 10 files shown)


backend/x11/backend.c
```diff
@@ -0,0 +1,374 @@
...
+		xcb_query_pointer_reply_t *pointer =
+			xcb_query_pointer_reply(x11->xcb_conn, cookie, NULL);
+		if (!pointer) {
+			break;
+		}
+
+		struct wlr_event_pointer_motion_absolute abs = {
+			.device = &x11->pointer_dev,
+			//.time_sec = ev->time / 1000,
+			//.time_usec = (ev->time % 1000) * 1000,
```
SirCmpwn: Commented out?

SirCmpwn: If you don't have anything from x11 here, just grab the current monotonic time.

backend/x11/backend.c
```diff
@@ -0,0 +1,374 @@
...
+}
+
+static struct wlr_backend_impl backend_impl = {
+	.start = wlr_x11_backend_start,
+	.destroy = wlr_x11_backend_destroy,
+	.get_egl = wlr_x11_backend_get_egl,
+};
+
+static void output_transform(struct wlr_output *output, enum wl_output_transform transform) {
+	// TODO
```
SirCmpwn: output_transform should be as easy as just doing this: https://github.com/SirCmpwn/wlroots/blob/master/backend/wayland/output.c#L51

backend/x11/backend.c
```diff
@@ -0,0 +1,374 @@
...
+static struct wlr_output_impl output_impl = {
+	.transform = output_transform,
+	.destroy = output_destroy,
+	.make_current = output_make_current,
+	.swap_buffers = output_swap_buffers,
+};
+
+static void input_destroy(struct wlr_input_device *input_dev) {
+	// Do nothing
+}
```
SirCmpwn: You don't have to provide implementations if they don't actually do anything. Each type should be checking if the impl exists and the function exists before calling it.

meson.build
```diff
@@ -45,6 +45,7 @@ udev           = dependency('libudev')
 pixman         = dependency('pixman-1')
 xcb            = dependency('xcb')
 xcb_composite  = dependency('xcb-composite')
+x11_xcb        = dependency('x11-xcb')
```
SirCmpwn: Let's make this optional.

SirCmpwn:
I was unable to run gnome-calculator on this backend, it says it can't connect to the display. weston-info seems to manage it, though. weston-simple-egl and weston-simple-shm work as well.

backend/x11/backend.c
```diff
@@ -0,0 +1,374 @@
...
+	}
+}
+
+static bool handle_x11_event(struct wlr_x11_backend *x11, xcb_generic_event_t *event) {
+	struct wlr_x11_output *output = &x11->output;
+
+	switch (event->response_type) {
+	case XCB_EXPOSE: {
+		wl_signal_emit(&output->wlr_output.events.frame, output);
+		break;
```
acrisci: dead code?

acrisci:
Seems to work for me. Tested gnome-calculator, weston-simple-{egl,shm,damage} and weston-terminal.

backend/x11/backend.c
ascent12: Wow, how did I not notice that?
I merged XCB_KEY_PRESS and XCB_KEY_RELEASE together and I guess I forgot to delete the old one.

backend/x11/backend.c
ascent12: I'm not sure what the x server bases its time off of. It would probably be better to deliver time from a single source. so I guess we should replace all of the events with monotonic time.

meson.build
ascent12: Of course. We need to go through and make all of this X11 stuff optional.
I figured it would happen at the same time as Xwayland.

SirCmpwn: Alright, can be done separately

ascent12:
Rebased against master and addressed some of the issues.
I still haven't looked into why some clients aren't starting yet.

SirCmpwn:
*shrug*

PR #170, opened 2017-09-29
Improve xwayland, batch 2

emersion:
See https://github.com/Cloudef/wlc/blob/master/src/xwayland/xwm.c#L135

### DIFF
--- a/include/wlr/types/wlr_wl_shell.h
+++ b/include/wlr/types/wlr_wl_shell.h
@@ -106,7 +106,7 @@ void wlr_wl_shell_destroy(struct wlr_wl_shell *wlr_wl_shell);
 
 void wlr_wl_shell_surface_ping(struct wlr_wl_shell_surface *surface);
 void wlr_wl_shell_surface_configure(struct wlr_wl_shell_surface *surface,
-	uint32_t edges, int32_t width, int32_t height);
+	enum wl_shell_surface_resize edges, int32_t width, int32_t height);
 void wlr_wl_shell_surface_popup_done(struct wlr_wl_shell_surface *surface);
 
 #endif
--- a/rootston/xwayland.c
+++ b/rootston/xwayland.c
@@ -28,8 +28,8 @@ static void handle_configure(struct wl_listener *listener, void *data) {
 	xwayland_surface->width = event->width;
 	xwayland_surface->height = event->height;
 
-	roots_surface->view->x = (double) event->x;
-	roots_surface->view->y = (double) event->y;
+	roots_surface->view->x = (double)event->x;
+	roots_surface->view->y = (double)event->y;
 
 	wlr_xwayland_surface_configure(roots_surface->view->desktop->xwayland,
 		xwayland_surface);
@@ -60,8 +60,8 @@ void handle_xwayland_surface(struct wl_listener *listener, void *data) {
 
 	struct roots_view *view = calloc(1, sizeof(struct roots_view));
 	view->type = ROOTS_XWAYLAND_VIEW;
-	view->x = (double) surface->x;
-	view->y = (double) surface->y;
+	view->x = (double)surface->x;
+	view->y = (double)surface->y;
 	view->xwayland_surface = surface;
 	view->roots_xwayland_surface = roots_surface;
 	view->wlr_surface = surface->surface;
--- a/rootston/xwayland.c
+++ b/rootston/xwayland.c
@@ -17,7 +17,25 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
 	free(roots_surface);
 }
 
-static void x11_activate(struct roots_view *view, bool active) {
+static void handle_configure(struct wl_listener *listener, void *data) {
+	struct roots_xwayland_surface *roots_surface =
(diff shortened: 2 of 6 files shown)


SirCmpwn:
wlr_xwayland_surface_configure should probably raise an event and let rootston set the x/y.

include/wlr/types/wlr_wl_shell.h
```diff
@@ -106,7 +106,7 @@ void wlr_wl_shell_destroy(struct wlr_wl_shell *wlr_wl_shell);
 
 void wlr_wl_shell_surface_ping(struct wlr_wl_shell_surface *surface);
 void wlr_wl_shell_surface_configure(struct wlr_wl_shell_surface *surface,
-	uint32_t edges, int32_t width, int32_t height);
+	enum wl_shell_surface_resize edges, int32_t width, int32_t height);
```
SirCmpwn: This should be a uint32_t, these are flags iirc. Enum value types cannot have more than one value.

rootston/xwayland.c
```diff
@@ -17,7 +17,25 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
...
+	struct wlr_xwayland_surface_configure_event *event = data;
+	struct wlr_xwayland_surface *xwayland_surface = event->surface;
+
+	xwayland_surface->x = event->x;
+	xwayland_surface->y = event->y;
+	xwayland_surface->width = event->width;
+	xwayland_surface->height = event->height;
+
+	roots_surface->view->x = (double) event->x;
+	roots_surface->view->y = (double) event->y;
```
SirCmpwn: Drop the spaces after `(double)`

rootston/xwayland.c
```diff
@@ -35,15 +53,20 @@ void handle_xwayland_surface(struct wl_listener *listener, void *data) {
...
+	wl_list_init(&roots_surface->request_configure.link);
+	roots_surface->request_configure.notify = handle_configure;
+	wl_signal_add(&surface->events.request_configure,
+		&roots_surface->request_configure);
 
 	struct roots_view *view = calloc(1, sizeof(struct roots_view));
 	view->type = ROOTS_XWAYLAND_VIEW;
-	view->x = view->y = 200;
+	view->x = (double) surface->x;
+	view->y = (double) surface->y;
```
SirCmpwn: Here as well

include/wlr/types/wlr_wl_shell.h
emersion: I believe all values are enumerated: https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_shell_surface-enum-resize (but this shouldn't be part of this PR, agreed)

rootston/xwayland.c
SirCmpwn: Can you also open a ticket with a list of x11 clients you've tested? We'll also fill it with clients we *should* test and try to deal with any weird ones.

SirCmpwn:
Can you have rootston log the window title when it gets set? Also the class and such. Does it make sense to merge the title and class into one signal? Is there other metadata we should make available?

include/wlr/types/wlr_wl_shell.h
SirCmpwn: You're right, no big deal then.

xwayland/xwm.c
```diff
@@ -80,12 +89,124 @@ static bool xcb_call(struct wlr_xwm *xwm, const char *func, uint32_t line,
...
+	xcb_window_t *xid = xcb_get_property_value(reply);
+	if (xid != NULL) {
+		surface->parent = lookup_surface_any(xwm, *xid);
+	} else {
+		surface->parent = NULL;
+	}
+
+	wlr_log(L_DEBUG, "XCB_ATOM_WM_TRANSIENT_FOR: %p", xid);
+	wl_signal_emit(&surface->events.set_parent, surface);
+}
```
SirCmpwn: What is the WM supposed to do with the parent? Anything?

SirCmpwn:
LGTM overall.

xwayland/xwm.c
emersion: I know some compositors move the parent to foreground (but under the child) when the child is activated. Some might also add a semi-transparent overlay to the parent while it has some children. I added the signal for completeness, but that's maybe not needed (at least for now).

emersion:
>Does it make sense to merge the title and class into one signal?

Maybe. I don't really know what compositors might do with the class. The class should only be set one time, and the title can change a lot, so I'm not sure about it. Maybe the `set_class` event is not really useful.

>Is there other metadata we should make available?

wlc also exposes:
* The state (used for fullscreen, modals)
* The PID
* The window type (tooltip, utility, DnD, dropdown, popup, combo, dialog, splash)
* A `close()` function

If you think any of these are of interest, I'll add 'em. Maybe there are things wlc doesn't implement but that we should implement.

SirCmpwn:
Hmm, I think all of those could be of value.

emersion:
Okay, let's add them.

SirCmpwn:
Thanks!

PR #175, opened 2017-09-30
rootston: add xdg shell resize support

emersion:
- [x] There's an issue with `get_input_bounds` - not large enough for resize handles
- [x] Properly implement resizing in `cursor.c`
- [x] Offset issues when resizing
- [ ] `wlr_xdg_toplevel_v6_set_resizing`

### DIFF
--- a/include/rootston/input.h
+++ b/include/rootston/input.h
@@ -78,7 +78,7 @@ struct roots_input {
 	enum roots_cursor_mode mode;
 	struct roots_view *active_view;
 	int offs_x, offs_y;
-	int view_width, view_height;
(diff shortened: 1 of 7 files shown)


emersion:
Resizing now works correctly. I ended up removing `get_input_bounds` (to let the client send resize requests) and add `get_size` (to send the correct size). Feedback is appreciated, not sure it's a good way to solve the issue.

We may want to change `view_begin_move` to use `view_x` _et al_.

SirCmpwn:
Thanks! I think we can deal with input bounds separately.

PR #177, opened 2017-09-30
rootston: add view->resize for xwayland

emersion:
Client requests for resize/move are already processed by `handle_request_configure`. I tested that with this program: https://gist.github.com/emersion/cc4909ba3581928871192822fc96d00b

It's the compositor's responsibility to detect mouse clicks near window borders and resize windows, if the compositor wants to have this feature. This is different from xdg shell (since xdg shell clients draw their own decorations).

This adds `view->resize` that will be used to force resize with meta+rclick.

### DIFF
--- a/rootston/xwayland.c
+++ b/rootston/xwayland.c
@@ -9,6 +9,14 @@
 #include "rootston/desktop.h"
 #include "rootston/server.h"
 
+static void resize(struct roots_view *view, uint32_t width, uint32_t height) {
+	assert(view->type == ROOTS_XWAYLAND_VIEW);
+	struct wlr_xwayland_surface *xwayland_surface = view->xwayland_surface;
+	xwayland_surface->width = width;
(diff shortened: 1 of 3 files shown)


rootston/xwayland.c
```diff
@@ -9,6 +9,14 @@
 #include "rootston/desktop.h"
 #include "rootston/server.h"
 
+static void resize(struct roots_view *view, uint32_t width, uint32_t height) {
+	assert(view->type == ROOTS_XWAYLAND_VIEW);
+	struct wlr_xwayland_surface *xwayland_surface = view->xwayland_surface;
+	xwayland_surface->width = width;
+	xwayland_surface->height = height;
```
SirCmpwn: Xwayland should handle setting the width and height itself.

SirCmpwn: I mean the wlroots xwayland code should.

PR #179, opened 2017-09-30
rootston: add resize support for wl_shell

emersion:
Need a client to test this.

Is it really important to set `WL_SHELL_SURFACE_RESIZE_NONE` to the real value when calling `wlr_wl_shell_surface_configure`?


emersion:
Tested with https://gist.github.com/emersion/7e453eb466beb3c2958c042ccf239ba7, it works.

SirCmpwn:
Thanks!

PR #182, opened 2017-10-01
DRM Multi-GPU

ascent12:
Finally got this shit sorted.
This introduces the ability for multiple DRM backends to be running at the same time, and have each of them driving their own outputs.

This performs all of the rendering on the primary GPU ("boot_vga" or whichever was found first), and copies it over to the other GPU to be displayed.
It uses zero-copy (with DMA-BUFs and EGLImages), so it's not too inefficient. I didn't notice any performance problems in my testing.

We'll eventually need to have a way to specify which GPU should be doing the rendering, and maybe even have a way at runtime to switch between them (e.g. unplugging laptop will use iGPU instead of discrete GPU) if it's even possible. That can come later, though.

There is also various bits of refactoring in this PR.


backend/meson.build
```diff
@@ -5,10 +5,10 @@ backend_files = files(
 	'session/session.c',
 	'drm/backend.c',
 	'drm/drm.c',
-	'drm/drm-atomic.c',
-	'drm/drm-legacy.c',
-	'drm/drm-properties.c',
-	'drm/drm-util.c',
+	'drm/iface_atomic.c',
+	'drm/iface_legacy.c',
```
SirCmpwn: I would just call these atomic.c and legacy.c

backend/drm/drm.c
```diff
@@ -20,12 +20,12 @@
...
-#include "backend/drm.h"
-#include "backend/drm-util.h"
+#include "backend/drm/drm.h"
+#include "backend/drm/util.h"
 
 bool wlr_drm_check_features(struct wlr_drm_backend *backend) {
-	extern const struct wlr_drm_interface legacy_iface;
-	extern const struct wlr_drm_interface atomic_iface;
+	extern const struct wlr_drm_interface iface_legacy;
+	extern const struct wlr_drm_interface iface_atomic;
```
SirCmpwn: And the old names for these variables made more sense imo.

backend/backend.c
```diff
@@ -91,37 +91,46 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
...
+		if (!primary_drm) {
+			primary_drm = drm;
+		}
+
+		wlr_multi_backend_add(backend, drm);
+	}
+
+	if (!primary_drm) {
+		wlr_log(L_ERROR, "Failed to open any DRM device");
+		goto error_multi;
```
SirCmpwn: Something that might be useful in the future is to specify a list of GPUs to use - then maybe you can run a different compositor on each GPU, for example.

SirCmpwn:
Overall, this looks good. Excellent work. Thanks!

backend/meson.build
ascent12: I wanted it to be obvious that they were related to iface.h.

backend/backend.c
ascent12: Yes, being able to explicitly pick the GPUs would be helpful.
Maybe a colon-separated environment variable would be helpful, where the first one specified is the primary GPU:
WLR_DRM_DEVICES=/dev/dri/card0:/dev/dri/card3

backend/meson.build
SirCmpwn: I think that was already obvious.

SirCmpwn: iface_atomic reads like hungarian notation imo

SirCmpwn:
Excellent, thanks!

PR #183, opened 2017-10-01
xdg-popup

acrisci:
Xdg popups are meant for user interface elements (mostly for menus I guess). They are similar to subsurfaces, but not related.

Popups must have a parent surface which can be a toplevel or another popup which gives them a tree structure. This structure determines the z-order of the popup, and the protocol forbids destroying popups that are not leaves.

Popups are created with a "positioner" that hints on how the popup should be positioned. This positioner is used only once when the popup is created and will then be destroyed by the client.

Popups can "grab" the seat. When a popup has a grab, it has keyboard focus. The client who has the grab will get events on their surfaces as normal, but no other client should get seat events until the grab is broken. The grab is broken when the popup is dismissed, or there is a button event outside of the popup and none of the child popups have a grab (I think). At that time the popup is closed.

Examples of xdg-popup:

* `weston-flower` right click menu (the simplest example)
* `gnome-terminal` menu bar menus (nested example)
* `gnome-calculator` has various menus

TODO:

* [x] cancel grab
* [x] popup keyboard grab
* [x] popup pointer grab
* [x] popup input events
* [x] render popups in the right position
* [x] popup tree data structures in the view


SirCmpwn:
Jesus christ, it really is a GTK+ protocol. Disgusting.

acrisci:
I didn't like the way Weston did the data structures for the popup, because they seem to store various parts of the popup state in all of the weston_surface, view, and weston_desktop_surface. Wlroots will contain all of this within the xdg-surface implementation.

The xdg-surface has a list of popups. A popup is added to the list upon popup role assignment (it may be within the list in an unconfigured state).

A popup will not have its own view, but will rather be considered part of the view of the xdg-toplevel it is a descendent of.

I think this is a good way to do it, but there are probably different ways someone could suggest that might have a different set of complexity tradeoffs.

SirCmpwn:
I think that will be fine. If we encounter problems with this approach later we can address them then.

rootston/desktop.c
```diff
@@ -93,6 +93,36 @@ static struct wlr_subsurface *subsurface_at(struct wlr_surface *surface,
 	return NULL;
 }
 
+static struct wlr_xdg_surface_v6 *xdg_v6_popup_at(
```
acrisci: `xdg_v6_popup_at()` was done by intuition and trial and error. Maybe someone smarter than me can simplify this.

acrisci:
The only useful thing I can see about a popup grab is that if you click outside of the window, it will close the popup. I can see how this behavior actually does require a grab.

Here are some other things the popup grab does that don't seem important:

1. When the grab is active, other windows do not get any pointer events. When I click in another window when the grab is active, it will close the popup but the button event is swallowed. This is also he behavior in i3.
2. The popup always has keyboard focus. This doesn't seem important because of how keyboard focus is usually handled. If the window loses activation, it will close the popup, and this is the time we change keyboard focus.

I think some of the usefulness of those might be for right click context menus on the "root window" or maybe there is some use case I'm not thinking of.

acrisci:
Ok this is ready to go.

types/wlr_xdg_shell_v6.c
```diff
@@ -12,12 +12,505 @@
...
+	xdg_grab = calloc(1, sizeof(struct wlr_xdg_popup_grab_v6));
+	if (!xdg_grab) {
+		return NULL;
+	}
+
+	xdg_grab->pointer_grab.data = xdg_grab;
+	xdg_grab->pointer_grab.interface = &xdg_pointer_grab_impl;
+	xdg_grab->keyboard_grab.data = xdg_grab;
+	xdg_grab->keyboard_grab.interface = &xdg_keyboard_grab_impl;
+	// TODO: keyboard grab
```
SirCmpwn: This todo looks outdated?

SirCmpwn:
Thanks!

Issue #158, opened 2017-09-23
Handle/support xkb in wlroots


Ongy:
The compositor examples currently use [`xkb_state_update_key`](https://github.com/SirCmpwn/wlroots/blob/master/examples/shared.c#L39).
The [documentation of this function](https://xkbcommon.org/doc/current/group__state.html#gac554aa20743a621692c1a744a05e06ce) states that `xkb_state_update_mask` should be used when input comes from a master process, like a wayland compositor.

This requires to either detect the current backend in a user, oder handle (part of) xkb inside wlroots.

PR #184, opened 2017-10-02
Keyboard modifiers

emersion:
Test config file:

```ini
[bindings]
Ctrl+q=exit
Ctrl+t=xterm
```

### DIFF
--- a/rootston/config.c
+++ b/rootston/config.c
@@ -250,7 +250,7 @@ struct roots_config *parse_args(int argc, char *argv[]) {
 		wl_list_insert(&config->bindings, &bc->link);
 		bc->command = strdup("exit");
 		bc->keysyms_len = 2;
-		bc->keysyms = calloc(1, sizeof(xkb_keysym_t));
+		bc->keysyms = calloc(1, bc->keysyms_len * sizeof(xkb_keysym_t));
 		bc->keysyms[0] = XKB_KEY_Meta_L;
 		bc->keysyms[1] = XKB_KEY_q;
 	} else if (result == -2) {
@@ -279,7 +279,12 @@ void roots_config_destroy(struct roots_config *config) {
 		free(dc);
 	}
 
-	// TODO: free bindings
+	struct binding_config *bc, *btmp = NULL;
+	wl_list_for_each_safe(bc, btmp, &config->bindings, link) {
+		free(bc->keysyms);
+		free(bc->command);
+		free(bc);
+	}
 
 	free(config->config_path);
 	free(config->cursor.mapped_output);
--- a/rootston/config.c
+++ b/rootston/config.c
@@ -216,14 +216,6 @@ struct roots_config *parse_args(int argc, char *argv[]) {
 	wl_list_init(&config->devices);
 	wl_list_init(&config->bindings);
 
-	// TEMPORARY, probably
-	struct binding_config *bc = calloc(1, sizeof(struct binding_config));
-	wl_list_insert(&config->bindings, &bc->link);
-	bc->command = strdup("exit");
-	bc->keysyms_len = 1;
-	bc->keysyms = calloc(1, sizeof(xkb_keysym_t));
-	bc->keysyms[0] = XKB_KEY_Escape;
-
 	int c;
 	while ((c = getopt(argc, argv, "C:h")) != -1) {
 		switch (c) {
@@ -253,6 +245,14 @@ struct roots_config *parse_args(int argc, char *argv[]) {
 
 	if (result == -1) {
 		wlr_log(L_DEBUG, "No config file found. Using empty config.");
+
+		struct binding_config *bc = calloc(1, sizeof(struct binding_config));
+		wl_list_insert(&config->bindings, &bc->link);
+		bc->command = strdup("exit");
+		bc->keysyms_len = 2;
+		bc->keysyms = calloc(1, sizeof(xkb_keysym_t));
+		bc->keysyms[0] = XKB_KEY_Meta_L;
+		bc->keysyms[1] = XKB_KEY_q;
 	} else if (result == -2) {
 		wlr_log(L_ERROR, "Could not allocate memory to parse config file");
 		exit(1);
--- a/rootston/config.c
+++ b/rootston/config.c
@@ -177,28 +177,49 @@ static int config_ini_handler(void *user, const char *section, const char *name,
 
 		bc->command = strdup(value);
 
-		bc->keysyms_len = 1;
+		size_t keysyms_len = 1;
 		char *symnames = strdup(name);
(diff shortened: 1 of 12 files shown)


SirCmpwn:
I would rather bind to shell commands and interpret "quit" as special.

rootston/config.c
```diff
@@ -182,6 +221,15 @@ struct roots_config *parse_args(int argc, char *argv[]) {
 	struct roots_config *config = calloc(1, sizeof(struct roots_config));
 	wl_list_init(&config->outputs);
 	wl_list_init(&config->devices);
+	wl_list_init(&config->bindings);
+
+	// TEMPORARY, probably
```
SirCmpwn: Temporary?

rootston/config.c
```diff
@@ -238,6 +286,8 @@ void roots_config_destroy(struct roots_config *config) {
 		free(dc);
 	}
 
+	// TODO: free bindings
```
SirCmpwn: Do this please

rootston/config.c
SirCmpwn: Oh, I misunderstood this. Take this out, please.

SirCmpwn: Let's only add a sane default quit binding if there's no config file (make it logo+q though).

SirCmpwn:
I like your approach. Good shit. I would introduce modifiers properly, though.

rootston/keyboard.c
```diff
@@ -21,12 +21,10 @@ static ssize_t keyboard_pressed_keysym_index(struct roots_keyboard *keyboard,
 }
 
 static void keyboard_binding_execute(struct roots_keyboard *keyboard,
-		struct binding_config *bc) {
+		char *command) {
 	struct roots_server *server = keyboard->input->server;
-	switch (bc->action) {
-	case BINDING_CONFIG_ACTION_QUIT:
+	if (strcmp(command, "exit") == 0) {
```
SirCmpwn: else exec it with the shell

rootston/keyboard.c
```diff
@@ -10,32 +11,95 @@
...
+	struct roots_server *server = keyboard->input->server;
+	if (strcmp(command, "exit") == 0) {
+		wl_display_terminate(server->wl_display);
+	} else {
+		pid_t pid = fork();
+		if (pid < 0) {
+			wlr_log(L_ERROR, "cannot execute binding command: fork() failed");
+			return;
+		} else if (pid == 0) {
+			execl("/bin/sh", "/bin/sh", "-c", command, (void *)NULL);
```
SirCmpwn: You'll probably want to close FDs here, right

SirCmpwn:
Also update the example config with docs on this feature and some sane defaults.

rootston/keyboard.c
emersion: Ah, right

SirCmpwn: It's also reasonable to just double check CLOEXEC. Up to you.

types/wlr_keyboard.c
```diff
@@ -13,19 +13,45 @@ int os_create_anonymous_file(off_t size);
...
+			group == keyboard->modifiers.group) {
+		return;
+	}
+
+	keyboard->modifiers.depressed = depressed;
+	keyboard->modifiers.latched = latched;
+	keyboard->modifiers.locked = locked;
+	keyboard->modifiers.group = group;
+
+	// TODO: wl_keyboard_send_modifiers
```
SirCmpwn: This isn't your job, it's already being handled by the seat. But maybe you should look for other places modifiers are being handled and attempt to deduplicate them?

emersion: Oh, right, thanks! (wlr_seat.c:405)

types/wlr_keyboard.c
```diff
@@ -21,37 +21,12 @@ static void keyboard_led_update(struct wlr_keyboard *keyboard) {
...
-			group == keyboard->modifiers.group) {
-		return;
-	}
-
-	keyboard->modifiers.depressed = depressed;
-	keyboard->modifiers.latched = latched;
-	keyboard->modifiers.locked = locked;
-	keyboard->modifiers.group = group;
-
-	// TODO: wl_keyboard_send_modifiers
```
SirCmpwn: I was just referring to wl_keyboard_send_modifiers when I said you don't have to do this. You still have to update depressed/latched/locked/group, or else wlr_keyboard_get_modifiers doesn't work without a seat. You should update seat to not duplicate the work, rather than pulling it out of here. Maybe raise a modifiers changed signal?

SirCmpwn:
Once the modifier handling code is good, update keybindings to specify modifiers rather than keysyms. Something like `Control+q` would be better than `Control_L+q`. Basically just hardcode a list of modifier names and check that before you attempt to parse the keysym.

types/wlr_keyboard.c
emersion: Yeah, this is now done in the seat: https://github.com/swaywm/wlroots/pull/184/files#diff-259a69ede9a79f9859eda604dd141090R417

Am I missing sth?

SirCmpwn: Yes.

What should be done in the seat: calling wl_keyboard_send_modifiers.

What should be done in wlr_keyboard: updating keyboard->modifiers.

emersion:
Some backends handle modifiers (wayland), some don't (libinput). That's why there's a public `wlr_keyboard_update_modifiers` and that's why `wlr_keyboard_update_state` calls `wlr_keyboard_update_modifiers`.

emersion:
If I understand correctly, the flow is: backend -> wlr_keyboard -> wlr_seat, wl_keyboard. I should probably:
* Remove `wlr_keyboard_update_modifiers`, that will be done according to the keymap in `wlr_keyboard_update_state` (in the wayland backend, modifiers events will be ignored)
* Add a `modifiers` event as you suggested in `wlr_keyboard`, `wlr_seat` will listen and call `wl_keyboard_send_modifiers`

SirCmpwn:
Is this issue relevant here? https://github.com/swaywm/wlroots/issues/158

emersion:
I checked that all FDs have CLOEXEC set (https://gist.github.com/emersion/cd54a1694644cb2ab1488cd8dd265a4f).

emersion:
Concerning #158, it would be better to use `xkb_state_update_mask` when using an X11 or Wayland backend, and fallback to `xkb_state_update_key` otherwise. That would prevent inconsistent keyboard states (pressing a modifier, focusing another window, releasing the modifier). However that could be done in a separate PR, as this PR doesn't change the current behavior.

SirCmpwn:
Cool. Can you rebase the erronous changes away?

emersion:
Done!

SirCmpwn:
Thanks!

PR #186, opened 2017-10-02
Fix cursor positioning on rotated output

versusvoid:
When output is rotated by 90 degrees, same transformation is applied to whole cursor texture, which on my system have 256x256 pixels. As result visible cursor movements are bounded to (-256, 1664) interval on y axis on FullHD monitor.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -618,7 +618,7 @@ static bool wlr_drm_output_set_cursor(struct wlr_output *_output,
 			return false;
 		}
 
-		plane->cursor_bo = gbm_bo_create(renderer->gbm, w, h, GBM_FORMAT_ARGB8888,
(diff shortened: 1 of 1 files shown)


acrisci:
It's not working for me on Rootston. It just puts a bunch of junk in the cursor plane.

versusvoid:
Works for me with this config:
```
[output:HDMI-A-1]
x=0
y=0

[output:VGA-1]
x=1920
rotate=90
```

Can you post yours?

acrisci:
I have one monitor with normal rotation, one with 90, and one with 270. The cursor does not display correctly on any monitor.

SirCmpwn:
This doesn't work for me, either. I have monitor at 90 and one at 0. It shows up corrupted on the 0 and not at all on the 90.

However, on master, it works fine - the cursor is shown rotated on the rotated display. Maybe something in #182 fixed it?

SirCmpwn:
Something to note is that the cursor is not rotated on the wayland backend on either branch, but I don't really care about that.

PR #187, opened 2017-10-02
Add examples/wlr-example.ini.example

versusvoid:
with option descriptions for examples


SirCmpwn:
Can you move this to rootston/rootston.ini.example?

versusvoid:
And change referenced file name at https://github.com/SirCmpwn/wlroots/blob/master/examples/config.c#L22 ?

SirCmpwn:
Hmm. Yes, I think so.

At some point we'll want to make the examples actually use rootston/config.c to get their work done.

SirCmpwn:
Thanks!

PR #189, opened 2017-10-02
xwayland: add support for window hints

### LINKED ISSUE #178: Add support for X11 _MOTIF_WM_HINTS and WM_NORMAL_HINTS
This includes features such as window minimum size and window decorations.

See https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/wm-normal-hints.html

wlc doesn't support these.

emersion:
- [x] Expose WM_NORMAL_HINTS
- [x] Add `libxcb-icccm` as an optional dependency
- [x] Parse _MOTIF_HINTS
- [x] Expose WM_HINTS
- [x] Do not expose `xcb_icccm_wm_hints_t` directly, create our own struct

Urgency can be retrieved with `xcb_icccm_wm_hints_get_urgency`.

Fixes #178


include/wlr/xwayland.h
```diff
@@ -53,6 +57,11 @@ struct wlr_xwayland_surface {
 	xcb_atom_t *protocols;
 	size_t protocols_len;
 
+	uint32_t motif_hints[5];
+	#ifdef HAS_XCB_ICCCM
+	xcb_size_hints_t size_hints;
```
emersion: Not sure how to do this.

SirCmpwn: You'll want to leave the data in the struct even if the library isn't in use. Figure out how big xcb_size_hints_t is and set aside the same amount of space.

emersion:
The library exposes raw XCB atoms and hints data, it's the minimum required to make this metadata available to compositors. But we could do more, especially concerning the non-standard `motif_hints` (we could parse hints into a struct). Do we want to do that?

SirCmpwn:
What are some use-cases for that?

include/wlr/xwayland.h
```diff
@@ -53,6 +63,16 @@ struct wlr_xwayland_surface {
...
 
+	uint32_t decorations;
+
+	#ifdef HAS_XCB_ICCCM
+	xcb_icccm_wm_hints_t *hints;
+	xcb_size_hints_t *size_hints;
+	#else
+	void *hints;
+	void *size_hints;
+	#endif
```
SirCmpwn: You know, as I think about it more, it probably makes more sense to convert the xcb structure into a wlr structure and expose that type regardless of whether or not we have support for it. Then it can just always be NULL if we don't support it and users don't have to put their code accessing the hints behind ifdefs

SirCmpwn:
Otherwise this looks good.

include/wlr/xwayland.h
emersion: Yeah, true. This bugs me for two reasons:
- We'll have a mix of `xcb` and `wlr` data types
- Users won't be able to use functions like `xcb_icccm_wm_hints_get_urgency`, because they take `xcb_icccm_wm_hints_t` as an argument

That said, we can probably call these for the user, and put the result in our struct.

SirCmpwn: Aye, we can do just that.

emersion: I ended up just copying these two structs into wlroots, and adding an extra field for urgency.

SirCmpwn:
Thanks!

PR #198, opened 2017-10-03
rootston: force move, resize and rotate

emersion:
- [x] Force move/resize
- [x] Force rotate
- [x] Set meta key in config file
- [x] Add rotation to pointer logic
- [x] Rotated subsurfaces

### DIFF
--- a/rootston/config.c
+++ b/rootston/config.c
@@ -86,6 +86,28 @@ invalid_input:
 	return NULL;
 }
 
+static uint32_t parse_modifier(const char *symname) {
+	if (strcmp(symname, "Shift") == 0) {
+		return WLR_MODIFIER_SHIFT;
+	} else if (strcmp(symname, "Caps") == 0) {
+		return WLR_MODIFIER_CAPS;
+	} else if (strcmp(symname, "Ctrl") == 0) {
+		return WLR_MODIFIER_CTRL;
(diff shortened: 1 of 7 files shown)


rootston/config.c
```diff
@@ -190,22 +221,9 @@ static int config_ini_handler(void *user, const char *section, const char *name,
...
-				bc->modifiers |= WLR_MODIFIER_MOD2;
-			} else if (strcmp(symname, "Mod3") == 0) {
-				bc->modifiers |= WLR_MODIFIER_MOD3;
-			} else if (strcmp(symname, "Logo") == 0) {
-				bc->modifiers |= WLR_MODIFIER_LOGO;
-			} else if (strcmp(symname, "Mod5") == 0) {
-				bc->modifiers |= WLR_MODIFIER_MOD5;
+			uint32_t modifier = parse_modifier(symname);
+			if (modifier != 0) {
+				bc->modifiers |= modifier;
```
ascent12: Bitwise ORing with zero isn't going to change anything, so you can remove this if.

emersion: Yeah, but I find it more readable this way: "if the symname is a modifier, add it to the list, otherwise try to parse it as a regular key name". I can change that if you prefer removing the if, though.

SirCmpwn:
Thanks!

PR #199, opened 2017-10-03
wlr-seat: honor seat version

### LINKED ISSUE #191: weston-simple-egl aborts
```
~/s/wlroots > env WAYLAND_DISPLAY=wayland-1 weston-simple-egl
has EGL_EXT_buffer_age and EGL_EXT_swap_buffers_with_damage
listener function for opcode 1 of wl_seat is NULL
fish: “env WAYLAND_DISPLAY=wayland-1 w…” terminated by signal SIGABRT (Abort)
```

Someone said this might be because of the order of globals, but it's not - I made the ordering match sway to no avail, and the abort comes during the dispatch anyway.

acrisci:
fixes #191

### DIFF
--- a/types/wlr_seat.c
+++ b/types/wlr_seat.c
@@ -12,6 +12,13 @@ static void resource_destroy(struct wl_client *client,
 	wl_resource_destroy(resource);
 }
 
+static void pointer_send_frame(struct wl_resource *resource) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #201, opened 2017-10-04
Set `DISPLAY` environment variable in XWayland

### LINKED ISSUE #200: Rootston does not the Xwayland DISPLAY in the environment
Can be fixed in either wlroots xwayland or in rootston, I don't care.

4e554c4c:
Fixes #200

### DIFF
--- a/xwayland/xwayland.c
+++ b/xwayland/xwayland.c
@@ -215,6 +215,12 @@ static bool wlr_xwayland_init(struct wlr_xwayland *wlr_xwayland,
 		wlr_xwayland_finish(wlr_xwayland);
 		return false;
 	}
+
+	// set $DISPLAY
+	char dpy[4] = {0};
+	snprintf(dpy, 4, ":%d", wlr_xwayland->display);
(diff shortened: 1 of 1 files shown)


4e554c4c:
@martinetd Are you suggesting I move this after the fork?

martinetd:
No, there already is code that should be setting DISPLAY in xserver_handle_ready()

I can reproduce so I definitely agree something isn't working :) But I don't think trying to add another setenv for DISPLAY is a proper solution.
The way xserver starting works is that we fork and start it, and when it feels like it it'll send a SIGUSR1 to notify us it's ready -- the variable only makes sense to be set at this point, that's why I had put the original setenv there.

I'll do some digging tonight if you'd like (in ~8 hours), but if you want to do something I'd like to understand why the current setup isn't working in the first place.
If we're fine with setting the envvar early, at least remove the old code that sets it :P

4e554c4c:
Yeah I see that now, thanks :)

4e554c4c:
After discussion on IRC, this turned out to not be an issue.

PR #206, opened 2017-10-05
Client provided cursors

emersion:
- [x] Update cursor on surface commit
- [x] Give surface the cursor role
- [x] The cursor actually changes only if the pointer focus for this device is one of the requesting client's surfaces or the surface parameter is the current pointer surface.
- [x] If surface is NULL, the pointer image is hidden.
- [x] On surface.attach requests to the pointer surface, hotspot_x and hotspot_y are decremented by the x and y parameters passed to the request. Attach must be confirmed by wl_surface.commit as usual.
- [x] The hotspot can also be updated by passing the currently set pointer surface to this request with new values for hotspot_x and hotspot_y.
- [x] Rootston: set default pointer when leaving a surface
- [x] Not working properly on xwayland

Docs: https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_pointer-request-set_cursor

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -1,7 +1,9 @@
+#define _POSIX_C_SOURCE 199309L
 #include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 #include <tgmath.h>
+#include <time.h>
 #include <wayland-server.h>
 #include <wlr/types/wlr_output.h>
 #include <wlr/types/wlr_surface.h>
@@ -171,12 +173,12 @@ bool wlr_output_set_cursor(struct wlr_output *output,
 	return set_cursor(output, buf, stride, width, height, hotspot_x, hotspot_y);
 }
 
-static void handle_cursor_surface_commit(struct wl_listener *listener,
-		void *data) {
-	struct wlr_output *output = wl_container_of(listener, output,
-		cursor.surface_commit);
-	struct wlr_surface *surface = data;
+static inline int64_t timespec_to_msec(const struct timespec *a) {
+	return (int64_t)a->tv_sec * 1000 + a->tv_nsec / 1000000;
+}
 
+static void commit_cursor_surface(struct wlr_output *output,
+		struct wlr_surface *surface) {
 	struct wl_shm_buffer *buffer = wl_shm_buffer_get(surface->current->buffer);
 	if (buffer == NULL) {
 		return;
@@ -196,6 +198,25 @@ static void handle_cursor_surface_commit(struct wl_listener *listener,
 		output->cursor.hotspot_x - surface->current->sx,
 		output->cursor.hotspot_y - surface->current->sy);
 	wl_shm_buffer_end_access(buffer);
+}
+
+static void handle_cursor_surface_commit(struct wl_listener *listener,
+		void *data) {
+	struct wlr_output *output = wl_container_of(listener, output,
+		cursor.surface_commit);
+	struct wlr_surface *surface = data;
+
+	commit_cursor_surface(output, surface);
+
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+
+	struct wlr_frame_callback *cb, *cnext;
+	wl_list_for_each_safe(cb, cnext, &surface->current->frame_callback_list,
+			link) {
+		wl_callback_send_done(cb->resource, timespec_to_msec(&now));
+		wl_resource_destroy(cb->resource);
+	}
 }
 
 static void handle_cursor_surface_destroy(struct wl_listener *listener,
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -172,24 +172,24 @@ bool wlr_output_set_cursor(struct wlr_output *output,
 
 bool wlr_output_set_cursor_surface(struct wlr_output *output,
 		struct wlr_surface *surface, int32_t hotspot_x, int32_t hotspot_y) {
-	if (output->impl->set_cursor) {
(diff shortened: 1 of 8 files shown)


types/wlr_output.c
```diff
@@ -168,6 +170,28 @@ bool wlr_output_set_cursor(struct wlr_output *output,
...
+	output->cursor.hotspot_x = hotspot_x;
+	output->cursor.hotspot_y = hotspot_y;
+
+	wlr_texture_destroy(output->cursor.texture);
+	output->cursor.texture = surface->texture;
+
+	wlr_renderer_destroy(output->cursor.renderer);
+	output->cursor.renderer = surface->renderer;
+
+	return true;
```
emersion: This outputs `attempt to render invalid texture`, I don't understand why.

types/wlr_output.c
```diff
@@ -168,6 +170,28 @@ bool wlr_output_set_cursor(struct wlr_output *output,
 				WL_SHM_FORMAT_ARGB8888, stride, width, height, buf);
 }
 
+bool wlr_output_set_cursor_surface(struct wlr_output *output,
+		struct wlr_surface *surface, int32_t hotspot_x, int32_t hotspot_y) {
+	if (output->impl->set_cursor) {
+		output->impl->set_cursor(output, NULL, 0, 0, 0, 0, 0);
```
emersion: How to implement this with a `wlr_surface` as input? Some backends like DRM already draw the cursor data to a `wlr_texture`, so converting the `wlr_surface` to a buffer seems redundant in this case.

types/wlr_output.c
SirCmpwn: This is a return statement. It doesn't output anything.

types/wlr_output.c
SirCmpwn: Should be able to by digging into the surface's buffer.

types/wlr_output.c
emersion: Yeah, that's exactly why I don't understand why it outputs something.

(I was talking about the entire function)

SirCmpwn: Well what line does the output actually *happen* on?

emersion: So actually this isn't the place where the cursor is rendered on screen. This just sets `output->cursor`, the cursor is rendered in `wlr_output_swap_buffers`.

SirCmpwn: Aye. And it's there where the error occurs?

emersion: Yeah, the error occurs when calling `wlr_render_with_matrix` from `wlr_output_swap_buffers`. I don't get why.

SirCmpwn: It means the texture hasn't had pixels attached to it yet.

emersion: Hmm, thanks.

emersion:
Okay, so reading the surface's buffer actually works. However, I get strange behaviors with gnome-calculator: the text cursor is not set when the pointer hovers the text input.

SirCmpwn:
Is it possible that they're reusing the surface and committing it again with an updated buffer

emersion:
In fact, yes. The hotspot can also be updated.

emersion:
Okay, this works much better now that I listen to commit events. Finishing the remaining TODOs will be much easier.

types/wlr_output.c
```diff
@@ -172,24 +172,24 @@ bool wlr_output_set_cursor(struct wlr_output *output,
...
+	if (format != WL_SHM_FORMAT_ARGB8888) {
+		return false;
+	}
 
-	return true;
+	void *data = wl_shm_buffer_get_data(buffer);
+	int32_t width = wl_shm_buffer_get_width(buffer);
+	int32_t height = wl_shm_buffer_get_height(buffer);
+	wl_shm_buffer_begin_access(buffer);
+	bool ok = wlr_output_set_cursor(output, data, width, width, height,
```
ascent12: You need to pass the stride properly.

emersion: Since we only support `ARGB8888`, this isn't a big deal. Also, `wl_shm_buffer_get_stride` returns the stride in bytes, and we need to pass it in pixels (so is it really useful?). See https://github.com/swaywm/wlroots/blob/master/backend/wayland/output.c#L61

ascent12: The stride has to do with the distance between different rows. There is no guarantee that stride == width (i.e. tightly packed).

emersion: Hmmm, you're right, I haven't thought about that. I've seen this done somewhere else in the code, I guess I'll change that too if possible.

emersion:
This PR is ready for review.

SirCmpwn:
This doesn't seem to work for Xwayland views.

emersion:
Which client? Works well with `xterm`.

SirCmpwn:
qutebrowser

emersion:
It seems that it does work, but the client somewhat updates the surface without sending a commit. Is it possible?

SirCmpwn:
Hmm, not that I'm aware of.

emersion:
Ah, maybe they're calling `set_cursor` again with the same surface, updated with a new buffer. I've planned to fix that, but forgot it.

emersion:
Okay, it turns out I'm doing something wrong with the `commit` even. It gets triggered but `handle_cursor_surface_commit` doesn't get called.

emersion:
All right, just fixed cursors for xwayland, after struggling a few hours.

SirCmpwn:
Thanks!

PR #220, opened 2017-10-06
Make Xwayland optional

### LINKED ISSUE #147: Make Xwayland optional

versusvoid:
Fix #147

- [x] runtime
- [x] compile time

### DIFF
--- a/include/rootston/config.h
+++ b/include/rootston/config.h
@@ -27,6 +27,7 @@ struct binding_config {
 };
 
 struct roots_config {
+	bool xwayland;
 	// TODO: Multiple cursors, multiseat
 	struct {
 		char *mapped_output;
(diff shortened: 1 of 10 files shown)


emersion:
I think the issue was about making xwayland an optional dependency at compile-time, not disabling it at runtime.

SirCmpwn:
Yeah, but I don't mind making it optional both ways.

SirCmpwn:
I think the API should stay the same but if compiled without xwayland support then wlr_xwayland_create should return NULL (and log the issue).

versusvoid:
So whole xwayland directory will be compiled in library, but `_create`, `_destroy` and `_activate` won't work? What the point?

SirCmpwn:
Eh, you're right.

PR #221, opened 2017-10-06
Use xkb_state_update_mask() with Wayland backend

### LINKED ISSUE #158: Handle/support xkb in wlroots

versusvoid:
Fix #158

### DIFF
--- a/backend/libinput/keyboard.c
+++ b/backend/libinput/keyboard.c
@@ -67,5 +67,6 @@ void handle_keyboard_key(struct libinput_event *event,
 		wlr_event.state = WLR_KEY_PRESSED;
 		break;
 	}
-	wlr_keyboard_update_state(wlr_dev->keyboard, &wlr_event);
+	wlr_event.update_state = true;
+	wlr_keyboard_notify_key(wlr_dev->keyboard, &wlr_event);
 }
--- a/include/wlr/interfaces/wlr_keyboard.h
+++ b/include/wlr/interfaces/wlr_keyboard.h
@@ -11,7 +11,10 @@ struct wlr_keyboard_impl {
 
 void wlr_keyboard_init(struct wlr_keyboard *keyboard, struct wlr_keyboard_impl *impl);
 void wlr_keyboard_destroy(struct wlr_keyboard *keyboard);
-void wlr_keyboard_update_state(struct wlr_keyboard *keyboard,
+void wlr_keyboard_notify_key(struct wlr_keyboard *keyboard,
 		struct wlr_event_keyboard_key *event);
+void wlr_keyboard_notify_modifiers(struct wlr_keyboard *keyboard,
+		uint32_t mods_depressed, uint32_t mods_latched, uint32_t mods_locked,
+		uint32_t group);
 
 #endif
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -149,13 +149,16 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
 	wlr_event.state = state;
 	wlr_event.time_sec = time / 1000;
 	wlr_event.time_usec = time * 1000;
-	wlr_keyboard_update_state(dev->keyboard, &wlr_event);
+	wlr_keyboard_notify_key(dev->keyboard, &wlr_event);
 }
 
 static void keyboard_handle_modifiers(void *data, struct wl_keyboard *wl_keyboard,
 		uint32_t serial, uint32_t mods_depressed, uint32_t mods_latched,
 		uint32_t mods_locked, uint32_t group) {
-
+	struct wlr_input_device *dev = data;
+	assert(dev && dev->keyboard);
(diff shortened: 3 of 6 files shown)


backend/libinput/keyboard.c
```diff
@@ -23,9 +23,17 @@ static void wlr_libinput_keyboard_destroy(struct wlr_keyboard *wlr_kb) {
 	libinput_device_unref(wlr_libinput_kb->libinput_dev);
 }
 
+void wlr_libinput_keyboard_update_state(struct wlr_keyboard *keyboard,
+		struct wlr_event_keyboard_key *event) {
+	uint32_t keycode = event->keycode + 8;
+	xkb_state_update_key(keyboard->xkb_state, keycode,
+		event->state == WLR_KEY_PRESSED ? XKB_KEY_DOWN : XKB_KEY_UP);
```
emersion: I don't like how xkb stuff is mixed with libinput stuff.

include/wlr/interfaces/wlr_keyboard.h
```diff
@@ -7,11 +7,16 @@
 struct wlr_keyboard_impl {
 	void (*destroy)(struct wlr_keyboard *keyboard);
 	void (*led_update)(struct wlr_keyboard *keyboard, uint32_t leds);
+	void (*update_state)(struct wlr_keyboard *keyboard,
+			struct wlr_event_keyboard_key *event);
```
emersion: Maybe we could add a `bool sends_modifiers_update` field instead? Not sure if it's a good idea to put a `bool` in a `*_impl` struct.

emersion:
Also, we need to call `wlr_keyboard_update_modifiers` from xwayland.

include/wlr/interfaces/wlr_keyboard.h
versusvoid: Nah, I think better solution will be `source` field in `struct wlr_event_keyboard_key` and condition on it.

versusvoid:
And why we need it? It does not call `wlr_keyboard_update_state()` now and works fine.

include/wlr/interfaces/wlr_keyboard.h
emersion: You mean, a field which indicates if keyboard state has already been updated?

emersion:
The reason why it's better to call `wlr_keyboard_update_modifiers` when we don't have ownership over the keyboard (i.e. when we're not the primary compositor, when we're an X/wayland client) is that we can miss some keys (e.g. catch <kbd>Ctrl</kbd> pressed but miss <kbd>Ctrl</kbd> released) and end up with a corrupted keyboard state.

emersion:
Also, it may be better to rename `wlr_keyboard_update_state` to something else (`wlr_keyboard_send_key`?) as this function won't even always update the keyboard state anymore.

include/wlr/interfaces/wlr_keyboard.h
versusvoid: Something like
```cpp
 void wlr_keyboard_update_state(struct wlr_keyboard *keyboard,
  		struct wlr_event_keyboard_key *event) {
	if (event->source == LIBINPUT) {
		uint32_t keycode = event->keycode + 8;
		xkb_state_update_key(keyboard->xkb_state, keycode,
			event->state == WLR_KEY_PRESSED ? XKB_KEY_DOWN : XKB_KEY_UP);
	}
```

emersion: libinput may or may not be the only backend that doesn't call `wlr_keyboard_update_modifiers`. I would prefer a more descriptive name, such as `bool update_state`. And as soon as the event will be forwarded to the compositor and the compositor doesn't care about this, maybe make this an additional argument to this function. But you're right, it's a better idea.

backend/libinput/keyboard.c
SirCmpwn: I agree. This shouldn't be here.

include/wlr/interfaces/wlr_keyboard.h
SirCmpwn: -1 to having wlr_keyboard test for which backed the event came from

versusvoid:
By looking at keyboard event handling in weston's X11 backend it can be concluded that correct approach isn't as simple as calling `wlr_keyboard_update_modifiers`. Committed modification make it work for now, but unlikely is complete solution.

SirCmpwn:
Looks good, thanks! If you're concerned about the X11 backend, file a ticket, but I don't personally really care about that backend.

PR #223, opened 2017-10-06
Activate only active xwayland views

### LINKED ISSUE #217: Focus lags behind for xwayland views
Clicking focus away to another view will send that click to the previously focused view, and you will have to click again to give the new view focus.

versusvoid:
Fix #217


SirCmpwn:
Should we also be deactivating windows if active == false? Does that make sense in xorg?

PR #224, opened 2017-10-06
Add command to close views

### LINKED ISSUE #216: Add "special" rootston command to close the focused window
Special meaning similar to the quit command.

emersion:
Fixes #216

### DIFF
--- a/rootston/rootston.ini.example
+++ b/rootston/rootston.ini.example
@@ -32,5 +32,5 @@ meta-key = Logo
 # Maps key combinations with commands to execute
 # Use the prefix "exec" to execute a shell command
 [bindings]
-Logo+q = exit # Stops the compositor
-Logo+w = close # Closes the current view
+Logo+Shift+e = exit # Stop the compositor
+Logo+q = close # Close the current view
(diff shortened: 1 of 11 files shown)


emersion:
wlc [destroys the `wl_client`](https://github.com/Cloudef/wlc/blob/704c56bf66f87f8788ab352a60d4ad0108b1920c/src/compositor/view.c#L503) if the surface implements wl_shell. Should we do that? Should we close `wl_client` for surfaces that aren't wl_shell too?

SirCmpwn:
That seems reasonable to me. wl_shell clients are going to become rarer and rarer anyway.

rootston/rootston.ini.example
```diff
@@ -30,6 +30,7 @@ meta-key = Logo
 
 # Keybindings
 # Maps key combinations with commands to execute
-# The special command "exit" stops the compositor
+# Use the prefix "exec" to execute a shell command
 [bindings]
-Logo+q = exit
+Logo+q = exit # Stops the compositor
+Logo+w = close # Closes the current view
```
SirCmpwn: Just for fun let's match the default sway keybindings. mod+q closes the current view and mod+shift+e exits sway.

SirCmpwn:
LGTM

SirCmpwn:
Thanks!

PR #228, opened 2017-10-07
When output has no modes, send one anyway

emersion:
The [docs](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output-enum-mode) say that:

>The event is sent when binding to the output object and there will always be one mode, the current mode.

However, some backends such as Wayland or X11 don't set any mode. This PR sends a mode in this case.

This fixes support for the screenshooter, which needs to know how much space it will allocate when creating a shared buffer.


SirCmpwn:
Nice, thanks!

SirCmpwn:
At some point we need to update the mode when the window size changes.

emersion:
Yeah, we should be doing this. When the currently selected mode changes (either if another one is selected under the DRM backend, or if the window is resized under the Wayland backend) we should send it. That implies keeping a list of resources to send the new mode to.

PR #234, opened 2017-10-07
fix clipboard

acrisci:
The clipboard was mostly there but I fixed some bugs that it works.

Also fixed a related crash with drag and drop, but that support is lacking.

Also did some refactoring in preparation for some further work I need to do to this interface.


SirCmpwn:
Hmm.

1. Open gedit
2. Open gnome-calculator
3. Copy answer from gnome-calculator
4. Attempt to paste into gedit

acrisci:
Yeah it's still broken, but it used to segfault when you did that before.

This will be my next project after I get wl_shell to work correctly and fix some other stuff with the surfaces.

SirCmpwn:
Ah, cool.

PR #236, opened 2017-10-08
Add GL/EGL extension loader generator

ascent12:
This is trying to remove a lot of the boilerplate and inconsistency for using GL/EGL extension functions, by automatically generating code to load it and check errors for us.

I've written the shell script to be POSIX compatible, so hopefully it will be portable enough. It passes `checkbashisms`.

It takes a text file with each function we want to load on its own line. If a function begins with a -, it won't fail if it's not loaded.


SirCmpwn:
Thanks!

PR #239, opened 2017-10-08
Center views in rootston

### LINKED ISSUE #212: Spawn rootston windows centered on-screen

emersion:
- [x] Center xdg shell views
- [x] Center wl shell views
- [ ] Center xwayland views
- [x] Activate views when spawned?

Fixes #212

### DIFF
--- a/rootston/xdg_shell_v6.c
+++ b/rootston/xdg_shell_v6.c
@@ -73,10 +73,7 @@ static void handle_commit(struct wl_listener *listener, void *data) {
 	struct roots_view *view = roots_xdg_surface->view;
 
 	if (!roots_xdg_surface->initialized) {
-		bool centered = view_center(view);
-		if (centered) {
-			roots_xdg_surface->initialized = true;
-		}
+		roots_xdg_surface->initialized = view_initialize(view);
 	}
 }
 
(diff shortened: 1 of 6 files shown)


rootston/xdg_shell_v6.c
```diff
@@ -67,6 +67,17 @@ static void handle_request_resize(struct wl_listener *listener, void *data) {
...
 
+static void handle_commit(struct wl_listener *listener, void *data) {
+	struct roots_xdg_surface_v6 *roots_xdg_surface =
+		wl_container_of(listener, roots_xdg_surface, commit);
+	struct roots_view *view = roots_xdg_surface->view;
+
+	bool centered = view_center(view);
+	if (centered) {
+		wl_list_remove(&listener->link);
+	}
```
SirCmpwn: Why remove the link if the view is off-screen? Also, won't this center it every time the view configures itself? Like when it's resizing?

emersion: The problem is that we get multiple `commit` events from the view before it gets a size. The goal of this is to try to center the window until it succeeds. Once the window has been centered, we remove the listener.

SirCmpwn: I see. I think it would be better to store an "initialized" variable in the rootston struct, perhaps. We may at some point want to do more with configure events post-initialize, no?

emersion:
The only missing piece is centering xwayland clients. Not sure if we want/can do it, since clients send an explicit position. Thoughts?

SirCmpwn:
I think we should override the explicit position Xwayland clients send. That's what other WMs do iiuc. But that can be separate.

PR #246, opened 2017-10-08
[wip] wl-shell fixes (QT support)

acrisci:
Now I will try to make wl-shell work correctly for QT clients.

Beware that there are many crashes on the DRM backend that will require you to restart your computer because you lose the ability to switch VTs because the keyboard dies.

QT test clients known to use wl-shell:

* qtcreator
* qgit

Start with e.g., `qtcreator -platform wayland`.

TODO:

* [x] render popups in the right place
* [ ] popup grab
* [ ] popup input events
* [ ] bug: window jumps around sometimes on pointer enter
* [ ] bug: weird repaint on resize bug


acrisci:
It's not pretty, but it works now, which is more than can be said for weston's implementation of this interface.

Not tested is the extremely weird situation where a popup can transition into a toplevel, which is allowed by the letter of the spec, but probably will never happen.

Note that you need QT ~5.7 to be able to test this. As of QT 5.9 (the arch version) at least, they have migrated to the xdg-shell.

This interface might not be around in a few years, but it's a good idea to support it because it's a core wayland protocol.

SirCmpwn:
I don't have any way to test this, but the code looks good. Thanks!

PR #247, opened 2017-10-08
Fix glgen to be more POSIX corformant

ascent12:
I tried running this with dash (used by debian/ubuntu for their /bin/sh) and apparently this wasn't conforming.

### DIFF
--- a/glgen.sh
+++ b/glgen.sh
@@ -30,28 +30,36 @@ CHECK_FMT='if (!%s) {
 }'
 
 while read -r COMMAND; do
-	[ ${COMMAND::1} = "-" ]
-	OPTIONAL=$?
+	OPTIONAL=0
+	FUNC_PTR_FMT='PFN%sPROC'
+
+	case $COMMAND in
+	-*)
+		OPTIONAL=1
+		;;
+	esac
+
+	case $COMMAND in
+	*WL)
+		FUNC_PTR_FMT='PFN%s'
+		;;
+	esac
 
 	COMMAND=${COMMAND#-}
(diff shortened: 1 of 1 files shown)


SirCmpwn:
How about more like

```sh
case $COMMAND in
    -*)
        # ...
        ;;
    *WL)
        ;;
esac
```

SirCmpwn:
[POSIX shell specification](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html)

ascent12:
I suppose that would work, but it's awfully verbose. Also, would -eglFooBarWL execute both of them?

emersion:
[Shellcheck](https://github.com/koalaman/shellcheck) is pretty good at catching this kind of issues (there's a POSIX mode).

SirCmpwn:
More verbose than defining two functions? More intented, perhaps. And yes, but are we going to do that?

ascent12:
-eglUnbindWaylandDisplayWL is there currently, although it would be easy to argue that a lot of the "optional" things there really shouldn't be optional.

SirCmpwn:
How about making it more POSIXY in general? `mandatory mandatory -o optional mandatory`

SirCmpwn:
(though strictly speaking POSIX specifies all options come before arguments)

ascent12:
It didn't turn out as verbose as I thought it would.
I don't think it's worth messing with the arguments. Meson was giving me shit when I tried to build both the .c and .h in the same script, so I made them separate.

PR #248, opened 2017-10-09
Fixed include in wlr_keyboard.h

Timidger:
The build was failing on my machine until I include these headers directly in the keyboard header file.


SirCmpwn:
Thanks!

PR #252, opened 2017-10-09
Add FreeBSD compatibility

myfreeweb:
Hi, it's [me](https://github.com/swaywm/sway/commit/da26d69cb1b21d582a81af0cad7342fab6596eed), again :)

(Do not merge this just yet!)

So, it works on X11, can't test on DRM/KMS for now because kernel panic. (well, I could install a couple month old kernel where wlc/sway worked, but I'll look into fixing the problem instead)

- `-Wno-error=deprecated-declarations` because `warning: 'wl_buffer' is deprecated`. This comes from `EGL/eglmesaext.h`. I have Mesa 17.2.2 and Wayland 1.13.0, but seems like it was deprecated a [long long time ago](https://mail.gnome.org/archives/commits-list/2013-October/msg08157.html). You didn't encounter this on Linux? o_0
- the libinput ifdefs are for compatibility with version 1.6.0 (damn we have to catch up…)
- There's the `environ` linking issue (undefined symbol). Maybe because of some linker flag that meson uses…
- I'm not sure about the raw terminal thing (from https://github.com/Cloudef/wlc/commit/ef3c443b3ab97a3a900de386973a3d1a5ccb065c) — @johalun why did wlc both save/restore termios state and set the raw flag?
- The elephant in the room. The ugly part of the patch. **The `_POSIX_C_SOURCE` / `_XOPEN_SOURCE` stuff.** What does it do on Linux? On FreeBSD, it actually makes all the headers compliant with old POSIX specifications and nothing more (`__BSD_VISIBLE` gets disabled) — so we lose `CLOCK_MONOTONIC`, `minor`/`major`, `snprintf` and a bunch of other stuff I already forgot. Currently I have `ifdef`'d these defines to Linux where they caused problems, but that's really bad. What do you think about them? Are they actually necessary anywhere at all?

### DIFF
--- a/backend/session/direct-freebsd.c
+++ b/backend/session/direct-freebsd.c
@@ -1,0 +1,217 @@
+#include <errno.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <signal.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <termios.h>
+#include <fcntl.h>
+#include <sys/types.h>
+#include <dev/evdev/input.h>
+#include <sys/kbio.h>
+#include <sys/consio.h>
+#include <wayland-server.h>
+#include <wlr/backend/session/interface.h>
+#include <wlr/util/log.h>
+#include "backend/session/direct-ipc.h"
+
+const struct session_impl session_direct;
+
+struct direct_session {
+	struct wlr_session base;
+	int tty_fd;
+	int old_kbmode;
+	int sock;
+	pid_t child;
+
+	struct wl_event_source *vt_source;
+};
+
+static int direct_session_open(struct wlr_session *base, const char *path) {
+	struct direct_session *session = wl_container_of(base, session, base);
+
+	int fd = direct_ipc_open(session->sock, path);
+	if (fd < 0) {
+		wlr_log(L_ERROR, "Failed to open %s: %s%s", path, strerror(-fd),
+			fd == -EINVAL ? "; is another display server running?" : "");
+		return fd;
+	}
+
+	struct stat st;
+	if (fstat(fd, &st) < 0) {
+		close(fd);
+		return -errno;
+	}
+
+	return fd;
+}
+
+static void direct_session_close(struct wlr_session *base, int fd) {
+	struct direct_session *session = wl_container_of(base, session, base);
+
+	struct stat st;
+	if (fstat(fd, &st) < 0) {
+		wlr_log_errno(L_ERROR, "Stat failed");
+		close(fd);
+		return;
+	}
+
+	close(fd);
+}
+
+static bool direct_change_vt(struct wlr_session *base, unsigned vt) {
+	struct direct_session *session = wl_container_of(base, session, base);
+	return ioctl(session->tty_fd, VT_ACTIVATE, (int)vt) == 0;
+}
+
+static void direct_session_destroy(struct wlr_session *base) {
+	struct direct_session *session = wl_container_of(base, session, base);
+	struct vt_mode mode = {
+		.mode = VT_AUTO,
+	};
+
+	errno = 0;
+
+	ioctl(session->tty_fd, KDSKBMODE, session->old_kbmode);
+	ioctl(session->tty_fd, KDSETMODE, KD_TEXT);
+	ioctl(session->tty_fd, VT_SETMODE, &mode);
+
+	if (errno) {
+		wlr_log(L_ERROR, "Failed to restore tty");
+	}
+
+	direct_ipc_finish(session->sock, session->child);
+	close(session->sock);
+
+	wl_event_source_remove(session->vt_source);
+	close(session->tty_fd);
+	free(session);
+}
+
+static int vt_handler(int signo, void *data) {
+	struct direct_session *session = data;
+
+	if (session->base.active) {
+		session->base.active = false;
+		wl_signal_emit(&session->base.session_signal, session);
+		ioctl(session->tty_fd, VT_RELDISP, 1);
+	} else {
+		ioctl(session->tty_fd, VT_RELDISP, VT_ACKACQ);
+		session->base.active = true;
+		wl_signal_emit(&session->base.session_signal, session);
+	}
+
+	return 1;
+}
+
+static bool setup_tty(struct direct_session *session, struct wl_display *display) {
+	int fd = -1, tty = -1, tty0_fd = -1;
+	if ((tty0_fd = open("/dev/ttyv0", O_RDWR | O_CLOEXEC)) < 0) {
+		wlr_log_errno(L_ERROR, "Could not open /dev/ttyv0 to find a free vt");
+		goto error;
+	}
+	if (ioctl(tty0_fd, VT_OPENQRY, &tty) != 0) {
+		wlr_log_errno(L_ERROR, "Could not find a free vt");
+		goto error;
+	}
+	close(tty0_fd);
+	char tty_path[64];
+	snprintf(tty_path, sizeof(tty_path), "/dev/ttyv%d", tty - 1);
+	wlr_log(L_INFO, "Using tty %s", tty_path);
+	fd = open(tty_path, O_RDWR | O_NOCTTY | O_CLOEXEC);
+
+	if (fd == -1) {
+		wlr_log_errno(L_ERROR, "Cannot open tty");
+		return false;
+	}
+
+	ioctl(fd, VT_ACTIVATE, tty);
+	ioctl(fd, VT_WAITACTIVE, tty);
+
+	int old_kbmode;
+	if (ioctl(fd, KDGKBMODE, &old_kbmode)) {
+		wlr_log_errno(L_ERROR, "Failed to read tty %d keyboard mode", tty);
+		goto error;
+	}
+
+	if (ioctl(fd, KDSKBMODE, K_CODE)) {
+		wlr_log_errno(L_ERROR, "Failed to set keyboard mode K_CODE on tty %d", tty);
+		goto error;
+	}
+
+	if (ioctl(fd, KDSETMODE, KD_GRAPHICS)) {
+		wlr_log_errno(L_ERROR, "Failed to set graphics mode on tty %d", tty);
+		goto error;
+	}
+
+	struct vt_mode mode = {
+		.mode = VT_PROCESS,
+		.relsig = SIGUSR1,
+		.acqsig = SIGUSR1,
+		.frsig = SIGIO, // has to be set
+	};
+
+	if (ioctl(fd, VT_SETMODE, &mode) < 0) {
+		wlr_log(L_ERROR, "Failed to take control of tty %d", tty);
+		goto error;
+	}
+
+	struct wl_event_loop *loop = wl_display_get_event_loop(display);
+	session->vt_source = wl_event_loop_add_signal(loop, SIGUSR1,
+		vt_handler, session);
+	if (!session->vt_source) {
+		goto error;
+	}
+
+	session->base.vtnr = tty;
+	session->tty_fd = fd;
+	session->old_kbmode = old_kbmode;
+
+	return true;
+
+error:
+	// Drop back to tty 1, better than hanging in a useless blank console
+	ioctl(fd, VT_ACTIVATE, 1);
+	close(fd);
+	return false;
+}
+
+static struct wlr_session *direct_session_create(struct wl_display *disp) {
+	struct direct_session *session = calloc(1, sizeof(*session));
+	if (!session) {
+		wlr_log_errno(L_ERROR, "Allocation failed");
+		return NULL;
+	}
+
+	session->sock = direct_ipc_init(&session->child);
+	if (session->sock == -1) {
+		goto error_session;
+	}
+
+	if (!setup_tty(session, disp)) {
+		goto error_ipc;
+	}
+
+	wlr_log(L_INFO, "Successfully loaded direct session");
+
+	snprintf(session->base.seat, sizeof(session->base.seat), "seat0");
+	session->base.impl = &session_direct;
+	return &session->base;
+
+error_ipc:
+	direct_ipc_finish(session->sock, session->child);
+	close(session->sock);
+error_session:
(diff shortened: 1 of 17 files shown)


SirCmpwn:
We definitely need CLOCK_MONOTONIC. Surely FreeBSD has an equivalent?

SirCmpwn:
Also I'm 1000% sure BSD has snprintf somehow. If it doesn't then you'll have to write up an snprintf implementation here.

ascent12:
snprintf is C99 standard. Surely the freeBSD libc is at least compatible with that?
Maybe the correct macros aren't being set somewhere.

myfreeweb:
We do have all that. I guess I didn't make that clear — the declarations/defines are lost because of the `_POSIX_C_SOURCE` stuff! And because of that, I put `_POSIX_C_SOURCE` into `ifdef __linux__` which is ugly.

SirCmpwn:
I don't think you fully understand how feature test macros work. `man feature-test-macros`, if it's installed on your system.

ascent12:
With the session code, I would honestly prefer that an entire new implementation was written instead of being bolted on top of the existing direct session.
There would be a bit of code duplication, but I feel the linux and freeBSD terminal/DRM handling are different enough to justify it.

myfreeweb:
Well, the macro situation is weird indeed — looks like *something* is setting `_ANSI_SOURCE`, `_C99_SOURCE` or `_C11_SOURCE` which actually disables all the things:

https://github.com/freebsd/freebsd/blob/1bfe7bd2dfc44082539777ab9d03ecdaffb41d95/sys/sys/cdefs.h#L748-L773

But I can't find these strings *anywhere*…

SirCmpwn:
I suspect it comes from a BSD header somewhere? Some macros activate other macros.

ascent12:
Compile with strict standards conformance off (`meson configure -Dc_std=gnu11`) and see if it will build then.

SirCmpwn:
Note that GNUC will not be merged, you'll have to find a different solution eventually.

johalun:
>    the libinput ifdefs are for compatibility with version 1.6.0 (damn we have to catch up…)

I've been meaning to look over all wayland related ports updating wayland port to 1.14 but it seems to take an awful lot of time for the 1.14 release... Maybe we need to go ahead and update related stuff like libinput before that? 

>    I'm not sure about the raw terminal thing (from Cloudef/wlc@ef3c443) — @johalun why did wlc both save/restore termios state and set the raw flag?

I'm definitely no master of virtual console management and this was the solution I could find by looking at how other projects did this. Do you have a better/different solution? Maybe it's a security thing, not sending ascii codes when not needed...(?) 

>  The elephant in the room. The ugly part of the patch. The _POSIX_C_SOURCE / _XOPEN_SOURCE stuff. What does it do on Linux? On FreeBSD, it actually makes all the headers compliant with old POSIX specifications and nothing more (__BSD_VISIBLE gets disabled) — so we lose CLOCK_MONOTONIC, minor/major, snprintf and a bunch of other stuff I already forgot. Currently I have ifdef'd these defines to Linux where they caused problems, but that's really bad. What do you think about them? Are they actually necessary anywhere at all?

For sway/wlc I upped the value on _XOPEN_SOURCE to make some headers visibly but if I understand correctly these shouldn't be used at all on FreeBSD because all they do is disable headers, right? I think it's OK to ifdef linux if they are required there.

johalun:
> With the session code, I would honestly prefer that an entire new implementation was written instead of being bolted on top of the existing direct session.

Agreed. The way its done in wlc is messy. I was planning to make a separate backend for freebsd but it hasn't had high priority...

myfreeweb:
Okay, looks like just increasing the POSIX/XOPEN values works for most things! e.g. `200809L` enables `CLOCK_MONOTONIC`.

`__BSD_VISIBLE` is only needed for `SOCK_CLOEXEC`.

I "fixed" the kernel panic (by reverting https://github.com/freebsd/freebsd/commit/24eb9f47135d89569c0fa92c46e5e82532249808 in the kernel), tested rootston on DRM/KMS, it works!

Split the direct session code into a separate file, yeah looks nicer than a ton of ifdefs.

myfreeweb:
Pushed all the changes, looks better now!

Virtual console management might be improved in the future, but It Works™.

I removed the libinput ifdefs because these were enum values, not defines. Oh well — [updated our libinput](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=222905).

backend/session/direct-ipc.c
```diff
@@ -36,8 +42,12 @@ static bool have_permissions(void) {
 #else
 static bool have_permissions(void) {
 	if (geteuid() != 0) {
+#ifdef __FreeBSD__
+		wlr_log(L_INFO, "Do not have root privileges; make sure you have write access to /dev/ttyv*, /dev/drm/* and /dev/input/*");
```
SirCmpwn: How about testing for access instead? Can do this on both platforms.

meson.build
```diff
@@ -34,6 +34,9 @@ if cc.get_id() == 'clang'
 	add_project_arguments('-Wno-missing-braces', language: 'c')
 endif
 
+# Wayland deprecated wl_buffer, Mesa EGL still uses it, at least on FreeBSD
+add_project_arguments('-Wno-error=deprecated-declarations', language: 'c')
```
SirCmpwn: Can you only add this on FreeBSD?

xwayland/xwayland.c
```diff
@@ -86,10 +90,12 @@ static void exec_xwayland(struct wlr_xwayland *wlr_xwayland) {
 		_exit(EXIT_FAILURE);
 	}
 
+#ifdef __linux__
 	if (clearenv()) {
 		wlr_log_errno(L_ERROR, "clearenv failed");
 		_exit(EXIT_FAILURE);
 	}
+#endif
```
SirCmpwn: You can't do this on FreeBSD? We should put in some other kind of mechanism, then.

myfreeweb: > There's the environ linking issue (undefined symbol).

I removed the `environ = NULL` because of this. I guess removing the `no-undefined` linker flag is also an option…

SirCmpwn: No, no, don't add that flag o_o

Simply figure out another way to clear the environment on BSD.

meson.build
myfreeweb: Are you sure you can't see that on Linux? Does `EGL/eglmesaext.h` not mention `wl_buffer`?

SirCmpwn: It does, but it also doesn't raise a deprecation error. We would have heard about it by now.

xwayland/xwayland.c
myfreeweb: There's no way other than `environ = NULL`.

And `environ` is in `crt1.o` instead of `libc.so` — which means the linker thinks it's undefined. https://bugs.webkit.org/show_bug.cgi?id=138420

Why is the environment cleared for Xwayland?

SirCmpwn: I'm not sure. Maybe @martinetd or @emersion knows.

meson.build
myfreeweb: Weird… hmm, I guess your compiler does not include `deprecated-declarations` in the default warning set! clang 5.0.0 has a lot more warnings enabled by default

SirCmpwn: We run builds on Clang, too, so... https://builds.sr.ht/job/320

backend/session/direct-freebsd.c
```diff
@@ -0,0 +1,253 @@
...
+#include <sys/types.h>
+#include <dev/evdev/input.h>
+#include <sys/kbio.h>
+#include <sys/consio.h>
+#include <wayland-server.h>
+#include <wlr/backend/session/interface.h>
+#include <wlr/util/log.h>
+#include "backend/session/direct-ipc.h"
+
+enum { DRM_MAJOR = 226 };
```
ascent12: Are you sure this is correct on FreeBSD?
It was my understanding that this number is linux-specific.

myfreeweb: Why would it be different? The drm/kms subsystem is copied from linux :) We don't use major/minor numbers generally, major is 0 for everything else.

meson.build
ascent12: I think that only the wl_buffer inside wayland-server.h (where you can see its actual declaration) is deprecated.
The opaque wl_buffer from wayland-server/client-protocol.h works fine.
Add -DWL_HIDE_DEPRECATED to the build system or something.

myfreeweb: Yeah, that works, thanks @ascent12

xwayland/xwayland.c
emersion: I don't know why the environment is cleared. But [wlc does it too](https://github.com/Cloudef/wlc/blob/master/src/xwayland/xwayland.c#L295), without having to ifdef linux.

myfreeweb: Oh, right, check out where that clearenv comes from https://github.com/Cloudef/wlc/blob/328edc089915aeb5cf26e5fc0e81b307ec707d62/src/internal.h#L395-L401 :)

CMake does not add `-Wl,--no-undefined` by default like Meson does, so wlc builds fine.

backend/session/direct-ipc.c
ascent12: Linux definitely needs root (or CAP_SYS_ADMIN) to become the DRM master, so it's a bit redundant to check anything else in this case.
The libcap version is the one that should be checking these sorts of permissions, but I think they currently will fail gracefully somewhere else if you don't.

@myfreeweb @johalun Does FreeBSD have similar requirements for becoming DRM master?

myfreeweb: No, no requirements. Like I wrote in the log string — all you need is permissions on tty and drm devices. I have been launching rootston as a normal user.

myfreeweb: The problem with testing access is that `have_permissions(void)` only has access to the void :) not to variables that store which drm/tty device that was selected. I think just showing a general "make sure you have permissions" message is fine.

xwayland/xwayland.c
johalun: > check out where that clearenv comes from https://github.com/Cloudef/wlc/blob/328edc089915aeb5cf26e5fc0e81b307ec707d62/src/internal.h#L395-L401 
Yeah, I did that...

backend/session/direct-ipc.c
ascent12: This function was written purely to test if the process was able to become the DRM master, which is not device-specific in Linux.

backend/session/direct-freebsd.c
johalun: I don't think you can get the same drm major value on freebsd. At least not with drm-next kernel iirc. In wlc I think I just skip that check.

myfreeweb: Oh — the check currently does the `setmaster` stuff. I guess I can remove some code :)

myfreeweb:
Meson is being an inflexible pile of garbage, it's impossible to set `b_lundef=false` conditionally (in `override_options` it does nothing) >_<

So, pick an option:

- just don't clear the environment, no one knows why it's cleared anyway
- set `b_lundef=false` as a default option (on any platform), it's not necessary
- tell users manually specify it on FreeBSD: `meson build -D b_lundef=false`

SirCmpwn:
Just do what @johalun did?

myfreeweb:
In wlc? Again — wlc's CMake based build did not set that `no-undefined` flag in the first place!

SirCmpwn:
Right. I think you should do the 3rd option and submit a ticket with meson.

myfreeweb:
Alright, done that.

xwayland/xwayland.c
martinetd: Yeah I've added that clearenv because wlc (and weston iirc) do it ; I think we could do without if we clear some key values instead but clearenv() sounds a lot less painful to me... My initial implementation used execvep but that was a GNU extension, I guess we could try to execve for every component of the PATH instead but that's a bit unsightly... There's not the problem you link from execve because we don't need to retrieve the whole previous env.

That said, I'm tempted to say it's easiest to do like wlc: define environ as extern and call it a day (either clear directly or redefine clearenv like them)

martinetd: Ah, I should learn to read before replying ; extern won't work with -Wl,--no-undefined?

That's gonna be tricky... Can try execve then, I like the idea of messing with warnings or linked objects quite less!

SirCmpwn:
Where's the meson ticket?

backend/session/session.c
```diff
@@ -263,6 +263,11 @@ size_t wlr_session_find_gpus(struct wlr_session *session,
 		return explicit_find_gpus(session, ret_len, ret, explicit);
 	}
 
+#ifdef __FreeBSD__
+	// XXX: libudev-devd does not return any GPUs (yet?)
```
SirCmpwn: Did you file a bug to follow up on this?

meson.build
```diff
@@ -34,6 +34,9 @@ if cc.get_id() == 'clang'
 	add_project_arguments('-Wno-missing-braces', language: 'c')
 endif
 
+# Avoid wl_buffer deprecation warnings
+add_project_arguments('-DWL_HIDE_DEPRECATED', language: 'c')
```
SirCmpwn: File a bug with EGL too please

SirCmpwn:
LGTM, just file bugs with the appropriate projects ASAP. Thanks!

SirCmpwn:
And follow-up here when they fix it so we can remove our workarounds, please.

meson.build
myfreeweb: https://bugs.freedesktop.org/show_bug.cgi?id=103225

myfreeweb:
Meson: https://github.com/mesonbuild/meson/issues/2388

backend/session/session.c
myfreeweb: There's no issue tracker on https://github.com/FreeBSDDesktop/libudev-devd — these projects aren't taken as seriously as you think :)

SirCmpwn: They're still overseen by someone. Maybe there's a parent project or a mailing list or if all else fails you can just email some recent committers directly.

PR #254, opened 2017-10-10
Destroy callbacks before callback list reinitialization

versusvoid:
Prevents accessing freed memory.


SirCmpwn:
Thanks!

acrisci:
I was looking at this yesterday and I don't understand why we free the callbacks here.

PR #255, opened 2017-10-10
Upload texture on surface commit

### LINKED ISSUE #226: Xwayland clients resize in a janky, laggy fashion

versusvoid:
An attempt to fix #226.
Quick testing didn't show any regressions, but I'm not very familiar with wayland architecture and thus do not know if any additional changes are necessary.


SirCmpwn:
Thanks!

emersion:
I believe this is breaking client provided cursors. It sets the surface buffer to NULL when committed. Any idea how to fix it? Rendering the surface to get back pixels again seems weird.

versusvoid:
Leave upload here and return buffer release to `render_surface()`?

PR #256, opened 2017-10-10
Fix wlr_output_layout_intersects

### LINKED ISSUE #238: Disappearing windows when moving in top right corner
1. Spawn a window
2. Move it so that the top corner of the window is above the top corner of the screen, it works
3. Move it so that the right corner of the window is at the right of the right corner of the screen, it works too
4. Move it so that the top right corner of the window is hidden off screen, the window disappears
5. If you don't release the mouse button and continue to move the window inside the screen, the window re-appears

Happens with the Wayland and the DRM backend.

emersion:
Fixes #238


SirCmpwn:
Thanks!

PR #257, opened 2017-10-11
[WIP] Fix client cursors

### LINKED ISSUE #253: Cursor jumps near window border
Tested on `weston-subsurfaces`. When cursor touches lower edge of border it jumps inside window. If move it further, cursor will change to larger (application provided?) icon.

emersion:
- [x] in request_set_cursor, to tell if there is pointer focus don't check what's under the cursor, check if the seat handle has a focused surface for the pointer, look in pointer_state, ~~that will probably fix the jump~~

Test clients:
- [ ] qutebrowser
- [ ] gnome-calculator, gnome-terminal
- [ ] weston-subsurfaces

See https://github.com/swaywm/wlroots/pull/255#issuecomment-335585656

Fixes #253

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -359,7 +359,7 @@ static void wlr_surface_flush_damage(struct wlr_surface *surface,
 		if (wlr_renderer_buffer_is_drm(surface->renderer,
 					surface->current->buffer)) {
 			wlr_texture_upload_drm(surface->texture, surface->current->buffer);
-			goto release;
+			goto clear_damage;
 		} else {
 			wlr_log(L_INFO, "Unknown buffer handle attached");
 			return;
@@ -372,7 +372,7 @@ static void wlr_surface_flush_damage(struct wlr_surface *surface,
 	} else {
 		pixman_region32_t damage = surface->current->buffer_damage;
 		if (!pixman_region32_not_empty(&damage)) {
-			goto release;
+			goto clear_damage;
 		}
 		int n;
 		pixman_box32_t *rects = pixman_region32_rectangles(&damage, &n);
@@ -388,7 +388,7 @@ static void wlr_surface_flush_damage(struct wlr_surface *surface,
 		}
 	}
 
-release:
+clear_damage:
 	pixman_region32_clear(&surface->current->surface_damage);
 	pixman_region32_clear(&surface->current->buffer_damage);
 }
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -429,7 +429,8 @@ static void wlr_surface_commit_pending(struct wlr_surface *surface) {
 
 	// Release the buffer after calling commit, because some listeners
 	// might need it (e.g. for cursor surfaces)
-	wlr_surface_state_release_buffer(surface->current);
+	// TODO: breaks weston-subsurfaces
+	//wlr_surface_state_release_buffer(surface->current);
 }
 
 static bool wlr_subsurface_is_synchronized(struct wlr_subsurface *subsurface) {
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -391,8 +391,6 @@ static void wlr_surface_flush_damage(struct wlr_surface *surface,
 release:
 	pixman_region32_clear(&surface->current->surface_damage);
 	pixman_region32_clear(&surface->current->buffer_damage);
-
-	wlr_surface_state_release_buffer(surface->current);
 }
 
 static void wlr_surface_commit_pending(struct wlr_surface *surface) {
@@ -428,6 +426,8 @@ static void wlr_surface_commit_pending(struct wlr_surface *surface) {
 
 	// TODO: add the invalid bitfield to this callback
 	wl_signal_emit(&surface->events.commit, surface);
+
+	wlr_surface_state_release_buffer(surface->current);
 }
 
 static bool wlr_subsurface_is_synchronized(struct wlr_subsurface *subsurface) {
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -427,8 +427,9 @@ static void wlr_surface_commit_pending(struct wlr_surface *surface) {
 	// TODO: add the invalid bitfield to this callback
 	wl_signal_emit(&surface->events.commit, surface);
 
-	// TODO: call this
-	//wlr_surface_state_release_buffer(surface->current);
+	// Release the buffer after calling commit, because some listeners
+	// might need it (e.g. for cursor surfaces)
+	wlr_surface_state_release_buffer(surface->current);
 }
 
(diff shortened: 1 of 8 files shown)


types/wlr_surface.c
```diff
@@ -391,8 +391,6 @@ static void wlr_surface_flush_damage(struct wlr_surface *surface,
 release:
```
acrisci: need to rename this now.

acrisci:
It helps a lot and I think this change is ok, but the cursor still jumps like 5px when entering the surface.

acrisci:
On weston-subsurfaces, the cursor jumps at the edges and gets much larger when inside the surface.

emersion:
This problem with `weston-subsurfaces` also happen in GNOME Shell. Not sure if we can do something about it.

versusvoid:
Does not happen in weston though =)
But I don't see why cursor jumps are related to this commit. It jumped before `flush_damage()` call was moved.

SirCmpwn:
The hotspot probably changes and we don't address that.

emersion:
Yeah, I can fix that too.

SirCmpwn:
Sweet.

emersion:
The `weston-subsurfaces` issue might be related to the surface's scale, which isn't applied (as well as other surface transformations).

acrisci:
> But I don't see why cursor jumps are related to this commit. It jumped before flush_damage() call was moved.

Only because the branch is called `fix-client-cursors` :)

emersion:
Okay, the jumping cursor issue is fixed when removing `wlr_surface_state_release_buffer`. This happens because:

1. The client sends a `set_cursor` request with a new cursor surface
2. The client draws and commits the surface
3. The client calls again `set_cursor` with the same surface without drawing and committing again

The surface in step 3 doesn't have a populated buffer anymore, so the cursor doesn't get updated.

emersion:
Possible solutions to this:
1. On commit, copy buffer into `wlr_output.cursor.pixels`. Doesn't work when a client creates a surface, commits it and then sets it as the cursor. Not sure if this can happen however.
2. Add a way to say to the output impl `set_cursor` function "change the hotspot without changing the cursor image". Maybe a new parameter `bool update_image`?
3. Add `wlr_texture_read_pixels`. Not sure how this would work though. Surface transformations (scale, transform and so on) would work (they don't work for now).
4. Make the output impl `set_cursor` function accept a `wlr_texture` or a `wlr_surface` instead of a buffer. Same problems and pros as (3).

SirCmpwn:
>Add a way to say to the output impl set_cursor function "change the hotspot without changing the cursor image". Maybe a new parameter bool update_image?

This

emersion:
Tried it, this isn't enough. After investigation, this is because `weston_subsurfaces` commits the surface before sending `set_cursor`.

This doesn't leave a lot of options.

1. Doesn't work anymore: enter a surface, the client sets a cursor and commits the cursor surface. Leave it, the default server cursor is enabled again. Enter it again, the client sends set_cursor without committing the cursor surface again. Pixels are lost because they are overwritten by the server cursor.
2. Not working.
3. Still working, but meh.
4. Still working, but meh.
5. The easiest option would be to revert #255: do not automatically upload the buffer. Or rather, let the compositor decide if a surface should be automatically uploaded on commit. Maybe add a `bool upload_on_commit` field to `wlr_surface`? (Or `bool no_upload_on_commit` so that it does it by default)

emersion:
Also, I don't know if not releasing the cursor surface buffer until a new buffer is received is a big deal.

SirCmpwn:
I guess we should go with 3 or 4, then. Up to you.

acrisci:
FYI another way they can set client cursors is through the data-device protocol. The surface is given through the protocol via a resource at the beginning of a drag operation and we have to make that surface be the pointer.

SirCmpwn:
Does that set the cursor or does it just display the surface underneath the cursor?

acrisci:
Would it work to overload the set_cursor function to work from a texture instead of buffer data (number 4 in your list)? I think that might be the way to go but I'm not familiar enough with the renderer to know the issues involved with that.

acrisci:
> Does that set the cursor or does it just display the surface underneath the cursor?

Yeah actually it just does that. It doesn't seem to be a pointer thing, so probably not related to this actually.

emersion:
>Would it work to overload the set_cursor function to work from a texture instead of buffer data

Yeah, that's what's planned.

I'm trying to read pixels from a texture, but I've trouble getting it right. Here are the result and the code, if someone finds a mistake somewhere... (May be related to `GL_RGBA4`, but GLES2 doesn't support `GL_RGBA8` so...)

screenshot from 2017-10-15 11-50-22

```c
	int width = surface->current->width;
	int height = surface->current->height;

	GLuint fbo, render_buf;
	glGenFramebuffers(1, &fbo);
	glBindFramebuffer(GL_FRAMEBUFFER, fbo);
	glGenRenderbuffers(1, &render_buf);
	glBindRenderbuffer(GL_RENDERBUFFER, render_buf);
	glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA4, width, height);
	glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_RENDERBUFFER, render_buf);

	glBindFramebuffer(GL_FRAMEBUFFER, fbo);
	glViewport(0, 0, width, height);
	glEnable(GL_BLEND);
	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);

	glClearColor(0, 0, 0, 1);
	float matrix[16];
	wlr_texture_get_matrix(surface->texture, &matrix, &output->transform_matrix, 0, 0);
	wlr_render_with_matrix(surface->renderer, surface->texture, &matrix);

	void *pixels = malloc(4 * width * height);
	wlr_renderer_read_pixels(surface->renderer, 0, 0, width, height, pixels);

	glDeleteFramebuffers(1, &fbo);
	glDeleteRenderbuffers(1, &render_buf);

	wlr_output_set_cursor(output, pixels, width, width, height,
		output->cursor.hotspot_x - surface->current->sx,
		output->cursor.hotspot_y - surface->current->sy);

	free(pixels);
```

emersion:
I disabled hardware client cursors for now. Tested with gnome-calculator, weston-subsurfaces and qutebrowser.

acrisci:
It's not completely broken anymore, but still jumps a little on the borders.

emersion:
On which client(s)? Can you be more specific? (Can't reproduce)

It's normal if it _blinks_ because it's switching between hardware and software cursors.

acrisci:
gnome-terminal drm or X11. Move the cursor slowly across a border and I see a jump.

> It's normal if it blinks because it's switching between hardware and software cursors.

It should not blink.

SirCmpwn:
File a new bug for blinkiness.

PR #258, opened 2017-10-11
[WIP] Implement surface-layers protocol

emersion:
- [x] Implement surface-layers in wlroots
- [x] Implement surface-layers in rootston
  - [x] Render at correct position
  - [x] Input types (pointer)
  - [x] Exclusive types: pointer
  - [x] Exclusive types: keyboard
  - [x] Render and send inputs at correct z-depth
  - [ ] ~~Exclusive zones~~ N/A
- [x] Build a test client
  - [x] Basic version
  - [x] CLI flags to control everything
- [x] Protocol changes
  - [x] Negative margins
  - [x] Errors
  - [x] Double-buffered state
  - [x] Fix non-exclusive pointer events
  - [ ] ~~Fix non-exclusive keyboard events~~
  - [x] Add a configure event
- [x] Bug fixes
  - [x] Invalid margin: -100,100
  - [x] More than 2 anchors sends a configure request
- [ ] Push back protocol changes to sway-protocols

Protocol spec: https://github.com/swaywm/sway-protocols#surface-layers


types/wlr_surface_layers.c
```diff
@@ -0,0 +1,129 @@
...
+#include <wayland-server.h>
+#include <wlr/types/wlr_surface.h>
+#include <wlr/types/wlr_surface_layers.h>
+#include <wlr/util/log.h>
+#include "surface-layers-protocol.h"
+
+static const char *surface_layers_role = "layer_surface";
+
+// TODO
+#define SURFACE_LAYERS_ERROR_ROLE 1
```
acrisci: These are normally defined within the protocol definition.

acrisci:
What is this intended to do and who is using it?

types/wlr_surface_layers.c
emersion: Yeah, exactly. Since it's not defined in the XML file yet, I wanted to talk with @SirCmpwn before adding it.

emersion:
See https://github.com/swaywm/sway-protocols#surface-layers

types/wlr_surface_layers.c
SirCmpwn: Yeah stick that in there please.

SirCmpwn:
LGTM so far

emersion:
All right, this is ready for review.

include/wlr/types/wlr_surface_layers.h
```diff
@@ -0,0 +1,74 @@
...
+	WLR_LAYER_SURFACE_INPUT_DEVICE_TOUCH = 3,
+};
+
+enum wlr_layer_surface_anchor {
+	WLR_LAYER_SURFACE_ANCHOR_NONE = 0,
+	WLR_LAYER_SURFACE_ANCHOR_TOP = 1,
+	WLR_LAYER_SURFACE_ANCHOR_BOTTOM = 2,
+	WLR_LAYER_SURFACE_ANCHOR_LEFT = 4,
+	WLR_LAYER_SURFACE_ANCHOR_RIGHT = 8,
+};
```
emersion: I should probably just include `surface_layers_server_protocol.h` instead.

SirCmpwn: Yes

examples/layer-surface.c
```diff
@@ -0,0 +1,434 @@
...
+enum layer_surface_anchor parse_anchor(const char *s) {
+	if (strcmp(s, "top") == 0) {
+		return LAYER_SURFACE_ANCHOR_TOP;
+	} else if (strcmp(s, "bottom") == 0) {
+		return LAYER_SURFACE_ANCHOR_BOTTOM;
+	} else if (strcmp(s, "left") == 0) {
+		return LAYER_SURFACE_ANCHOR_LEFT;
+	} else if (strcmp(s, "right") == 0) {
+		return LAYER_SURFACE_ANCHOR_RIGHT;
+	}
```
SirCmpwn: This is a bitmask - you can be top-right, bottom-right, etc.

examples/layer-surface.c
```diff
@@ -0,0 +1,434 @@
...
+enum layer_surface_input_device parse_input_device(const char *s) {
+	if (strcmp(s, "pointer") == 0) {
+		return LAYER_SURFACE_INPUT_DEVICE_POINTER;
+	} else if (strcmp(s, "keyboard") == 0) {
+		return LAYER_SURFACE_INPUT_DEVICE_KEYBOARD;
+	} else if (strcmp(s, "touch") == 0) {
+		return LAYER_SURFACE_INPUT_DEVICE_TOUCH;
+	}
+	fprintf(stderr, "Unknown input device: %s\n", s);
+	exit(EXIT_FAILURE);
```
SirCmpwn: This is a bitfield too

examples/layer-surface.c
```diff
@@ -0,0 +1,434 @@
...
+		if (opt == -1) {
+			break;
+		}
+
+		switch (opt) {
+		case 'l':
+			layer_index = parse_layer(optarg);
+			break;
+		case 'a':
+			anchor |= parse_anchor(optarg);
```
SirCmpwn: Oh I see, sorry

SirCmpwn:
Test notes:

- Surface is getting really weird pointer events
- Exclusive input does not work (exclusive input events should not also be passed to surfaces underneath the layer, and should cause a focus change probably)

include/wlr/types/wlr_surface_layers.h
emersion: Hum, I have trouble doing so because it's a header file not installed on the host system but generated on build. Any idea how to solve this?

SirCmpwn: You should get it via https://github.com/swaywm/wlroots/blob/master/types/meson.build#L26

emersion: Okay, so I shouldn't include it in the header file. I wanted to avoid this because you loose the `enum wlr_layer_surface_layer layer` field (you have to replace that by `uint32_t layer`).

examples/layer-surface.c
```diff
@@ -0,0 +1,434 @@
...
+		return LAYER_SURFACE_INPUT_DEVICE_KEYBOARD;
+	} else if (strcmp(s, "touch") == 0) {
+		return LAYER_SURFACE_INPUT_DEVICE_TOUCH;
+	}
+	fprintf(stderr, "Unknown input device: %s\n", s);
+	exit(EXIT_FAILURE);
+}
+
+void parse_pair(const char *s, uint32_t *x, uint32_t *y) {
+	char *buf = strdup(s);
```
ascent12: Why are you duplicating (and then leaking) the string here?
You don't need to worry about the comma messing with atoi. It'll immediately stop once it encounters a non-digit.

ascent12: Actually, thinking about it, you can just use `sscanf(s, "%u,%u", x, y)`.

SirCmpwn: -1 to sscanf. +1 to fixing the leak

examples/layer-surface.c
```diff
@@ -0,0 +1,434 @@
...
+static void draw() {
+	if (eglMakeCurrent(egl_display, egl_surface, egl_surface, egl_context)) {
+		fprintf(stderr, "Made current\n");
+	} else {
+		fprintf(stderr, "Made current failed\n");
+	}
+
+	glClearColor(1.0, 1.0, 0.0, 0.1);
+	glClear(GL_COLOR_BUFFER_BIT);
+	glFlush();
```
ascent12: It's a very minor point, but this call is not needed. eglSwapBuffers will do it for us.

examples/layer-surface.c
ascent12: sscanf does literally the exact same thing with no unnecessary and actually allows errors to checked a bit better than the current implementation.
```c
errno = 0;
if (sscanf(s, "%u,%u", x, y) != 2 || errno != 0) {
    fprintf(stderr, "Invalid format\n");
    exit(EXIT_FAILURE);
}
```
I agree that scanf (the stdin one) is shit, but sscanf is fine.

SirCmpwn: It works, but I don't really like it.

emersion:
Fixed both issues.

Not sure about the focus change. When a lockscreen is enabled, I don't want all surfaces to be unfocused after it's unlocked. Also, surface_layers doesn't have the concept of focus.

examples/layer-surface.c
acrisci: can you be top-bottom or right-left?

emersion: Yes, just specify `-a` multiple times (e.g. `-a top -a left`).

rootston/surface_layers.c
```diff
@@ -0,0 +1,59 @@
...
+
+	struct roots_layer_surface *roots_surface =
+		calloc(1, sizeof(struct roots_layer_surface));
+	if (roots_surface == NULL) {
+		return;
+	}
+	roots_surface->desktop = desktop;
+	roots_surface->layer_surface = surface;
+
+	wl_list_init(&roots_surface->destroy.link);
```
acrisci: Do we need these list initializers here like this? Won't signal_add handle it for you?

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,212 @@
...
+#include "surface-layers-protocol.h"
+
+static const char *surface_layers_role = "layer_surface";
+
+static void layer_surface_set_interactivity(struct wl_client *client,
+		struct wl_resource *resource, uint32_t input_types,
+		uint32_t exclusive_types) {
+	struct wlr_layer_surface *surface = wl_resource_get_user_data(resource);
+	surface->input_types = input_types;
+	surface->exclusive_types = exclusive_types;
```
acrisci: I think we should handle invalid input types with a protocol error.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,212 @@
...
+	wl_signal_emit(&surface->events.set_interactivity, surface);
+}
+
+static void layer_surface_set_anchor(struct wl_client *client,
+		struct wl_resource *resource, uint32_t anchor) {
+	struct wlr_layer_surface *surface = wl_resource_get_user_data(resource);
+	surface->anchor = anchor;
```
acrisci: Maybe a protocol error for invalid anchors (like orthogonal anchors or anchors outside the enum).

rootston/surface_layers.c
emersion: According to https://people.freedesktop.org/~whot/wayland-doxygen/wayland/Server/structwl__listener.html, we don't need to do it. Right now we do it in a lot of places, we should clean up the whole codebase.

types/wlr_surface_layers.c
emersion: I don't know if it's a good idea. More input types may be added in the future.

types/wlr_surface_layers.c
emersion: Again, I don't know: the rest of the wayland protocol isn't sending errors for things like this (e.g. [`edges` in `wl_shell_surface::resize`](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_shell_surface-request-resize)).

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,212 @@
...
+static void layer_surface_set_exclusive_zone(struct wl_client *client,
+		struct wl_resource *resource, uint32_t zone) {
+	struct wlr_layer_surface *surface = wl_resource_get_user_data(resource);
+	surface->exclusive_zone = zone;
+	wl_signal_emit(&surface->events.set_exclusive_zone, surface);
+}
+
+static void layer_surface_set_margin(struct wl_client *client,
+		struct wl_resource *resource, uint32_t horizontal,
+		uint32_t vertical) {
```
acrisci: Does it make sense for margins to be negative like for drop-down animations?

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,212 @@
...
+	struct wlr_layer_surface *surface = wl_resource_get_user_data(resource);
+	surface->margin_horizontal = horizontal;
+	surface->margin_vertical = vertical;
+	wl_signal_emit(&surface->events.set_margin, surface);
+}
+
+void wlr_layer_surface_get_position(struct wlr_layer_surface *layer_surface,
+		double *x, double *y) {
```
acrisci: I think we should allow the layers to "stack" on top of each other (similar to X11 `STRUT_PARTIAL`). So if two layers both request to be at the bottom of the screen (like two bars), they will appear on top of each other without the clients having to know each others positions.

types/wlr_surface_layers.c
emersion: Shouldn't the client handle animations with a translation matrix on his side? I don't think we want users to use margins to animate things. Although, it might still be useful to have negative margins for other things.

acrisci:
Do we want to double buffer the properties?

This would allow us to for instance set the anchor and the margin at the same time without flashing unrendered content.

types/wlr_surface_layers.c
acrisci: > I don't think we want users to use margins to animate things.

Why not? This would be such a cool feature for guake-style dropdowns and totally within the scope of the compositor.

types/wlr_surface_layers.c
acrisci: This is why interfaces are versioned.

types/wlr_surface_layers.c
acrisci: Many of them do this, like xdg-shell and data-device.

types/wlr_surface_layers.c
acrisci: So for instance, if I am desktop icons, I will want to put my icons in the top left corner but in a position relative to the bar height.

protocol/surface-layers.xml
```diff
@@ -0,0 +1,145 @@
...
+        actually be occluded.
+
+        This value is only meaningful if the surface is anchored to an edge,
+        rather than a corner. The zone is the number of pixels from the edge
+        that are considered exclusive.
+      </description>
+      <arg name="zone" type="uint"/>
+    </request>
+
+    <request name="set_margin">
```
acrisci: The drag icon interface uses `surface.attach` `sx` and `sy` properties to set its margin, so that might be the canonical wayland way of doing things which we should be aware of. But on the other hand, this is much more explicit.

rootston/cursor.c
```diff
@@ -263,7 +294,11 @@ static void do_cursor_button_press(struct roots_input *input,
 		input->input_events_idx = (i + 1)
 			% (sizeof(input->input_events) / sizeof(input->input_events[0]));
 		set_view_focus(input, desktop, view);
-		if (view) {
+
+		struct wlr_layer_surface *layer_surface =
+			wlr_surface_layers_get_exclusive(desktop->surface_layers,
+			LAYER_SURFACE_INPUT_DEVICE_KEYBOARD);
+		if (view && !layer_surface) {
```
acrisci: This looks really strange to me. You should never be in a state where window manager focus and keyboard focus are different (window manager focus is essentially an indication of where keyboard focus is). If the surface layer accepts keyboard input (not handled here?) then it should be considered the focused window. If there is no keyboard focus, then there should be no window manager focus.

emersion: The problem here is that surface layers are not views, thus can't have focus (and don't have the concept of focus). Also a surface layer can have exclusive keyboard but no exclusive pointer (or vice versa).

acrisci: These issues seem like they need to be addressed in the protocol.

acrisci:
Do these things support subsurfaces and popups?

What if I make desktop icons and I want to have a context menu when I right click on a file?

What if I have a bar and I want to show a drop down menu when the user clicks on an icon (like the windows button)?

rootston/surface_layers.c
SirCmpwn: Make sure to file a ticket for this

types/wlr_surface_layers.c
SirCmpwn: Yeah protocol error is fine and we can test the version later.

types/wlr_surface_layers.c
SirCmpwn: Protocol error here. And we should put them everywhere.

types/wlr_surface_layers.c
SirCmpwn: Negative margins sounds good to me

types/wlr_surface_layers.c
SirCmpwn: No, that's intentionally not part of the design (for now). But that's somewhat possible with exclusivity - the idea is that a bar or something similar can set its exclusive zone to the top 10 pixels of the screen, then anything at or above that layer will not be shown on top of it (e.g. notifications).

SirCmpwn:
>Do we want to double buffer the properties?

Not a bad idea.

>Do these things support subsurfaces and popups?

Not for v1, propose this for v2

rootston/cursor.c
SirCmpwn: No, this is by design. A surface layer can choose to *receive* certain input events, and also choose to receive them *exclusively*. In the former something else will still have focus and both will get the event. The binary idea of "focus" is a bit too simple for this protocol's design.

SirCmpwn:
We should also add a configure event.

SirCmpwn:
Sorry, rootston has some reoccuring issues it seems.

We should add a configure event so the compositor can tell the client how much space it ended up being allocated. If it attaches to all edges, for example, this will be the size of the output minus any exclusive zones set by other surfaces.

rootston/cursor.c
emersion: Oh, I didn't understood that both get the event.

emersion:
Okay, so the only thing missing from this PR is now keyboard events for layer surfaces. Since surface layers is a _subtle_ protocol, this PR [is also _subtle_](https://github.com/swaywm/wlroots/pull/258/files#diff-fdc0e2c92f2086c22c357b7879ff1f1cR81).

Blocked on #287.

emersion:
This PR is ready for review.

acrisci:
How do I test it? What are the edge cases?

emersion:
You can try several things:

1. One layer surface, one anchor, some margin
2. Two anchors
3. One layer surface at the top/overlay layer receiving pointer events, one view
4. One layer surface at the bottom/background layer receiving pointer events, one view
5. One layer surface at the top/overlay layer receiving exclusive pointer events, one view
6. Two overlapping layer surfaces, one at the top layer, one at the overlay layer, both receiving pointer events
7. Two overlapping layer surfaces, one at the top layer receiving exclusive pointer events, one at the overlay layer receiving non-exclusive pointer events
8. Two overlapping layer surfaces at the same layer, both receiving pointer events

acrisci:
> You can try several things:

Great. How do I do these things?

emersion:
Ahah, just use `examples/layer-surface`:

```
layer-surface [-l <layer>] [-a <anchor>]... [-i <input-type>]... [-e <exclusive-type>]... [-m <margin_h,margin_v>] [-s <width,height>] [-h]
```

1. `layer-surface -l overlay -a top -m 100,100`
2. `layer-surface -l overlay -a top -a right`
3. `layer-surface -l overlay -a top -i pointer` and spawn a view
4. `layer-surface -l background -a top -i pointer` and spawn a view
5. `layer-surface -l overlay -a top -i pointer -e pointer` and spawn a view
6. `layer-surface -l overlay -a top -i pointer -s 200,100` and `layer-surface -l top -a top -i pointer -s 100,200`

And so on.

acrisci:
`surface-layers` gives invalid margin for negative numbers: `Invalid margin: -100,100`

acrisci:
Overlay windows give the window behind it pointer focus. Is this a bug?

emersion:
>Overlay windows give the window behind it pointer focus even with -i.

Overlay windows with -i receive events, but act as if they weren't there. Adding the -e flag prevents the event from being sent to surfaces underneath. So this isn't a bug.

acrisci:
Should exclusive pointer break grabs when going over its surface?

I can start an interactive move of a shell window, move the titlebar through the overlay with exclusive pointer, and the overlay won't receive any pointer motion events.

SirCmpwn:
>I can start an interactive move of a shell window, move the titlebar through the overlay with exclusive pointer, and the overlay won't receive any pointer motion events.

This behavior seems fine imo

emersion:
That's a good question. For a lockscreen, I guess its better if it does break the grab, otherwise you could end up in a state where you think the screen is locked but there's a stupid keyboard grab that allows a bad guy to type things in a view.

SirCmpwn:
I don't think "locking the screen while dragging a window around" is going to happen often enough to care.

acrisci:
Also, the logic of keyboard grabs exists entirely within our code, so there is no security threat possible from grabs.

acrisci:
This however is  a pretty serious issue. If I have a popup window open, the overlay won't get pointer events.

Popup grabs take the keyboard too. So if I have a right-click context menu open and my lock screen goes off, I cannot interact with my lock screen.

SirCmpwn:
>right-click context menu open and my lock screen goes off

This also seems like it's not going to happen in practice.

acrisci:
I am allowed to pass three or four anchors, but what happens when I do doesn't make any sense.

For instance, if I pass all four anchors, the layer is drawn in the top left corner.

SirCmpwn:
FTR the desired behavior is for things like wallpapers to pass all four anchors.

acrisci:
> FTR the desired behavior is for things like wallpapers to pass all four anchors.

So if they want to be in the top-left corner, they should just specify the top-left corner and draw the width/height of the output read from the output info they get from wl_output.

What's the point of being able to pass three or four anchors if it doesn't do anything new?

SirCmpwn:
Strictly speaking, Wayland avoids the idea that views are allowed to know the monitor dimensions in pixels.

acrisci:
Oh, I see now. So if I pass all four anchors, I'll get a configure request for the size of the output I guess.

Then let's have the example client do that so we can demonstrate the configure request to see if it works correctly.

SirCmpwn:
+1

acrisci:
Am I not allowed to change my layer value after I make surface layer object?

SirCmpwn:
No.

include/rootston/desktop.h
```diff
@@ -52,19 +61,24 @@ struct roots_desktop {
...
 
 void view_destroy(struct roots_view *view);
 struct roots_view *view_at(struct roots_desktop *desktop, double lx, double ly,
-		struct wlr_surface **surface, double *sx, double *sy);
+	struct wlr_surface **surface, double *sx, double *sy);
 void view_activate(struct roots_view *view, bool activate);
 
+bool layer_surface_is_at(struct roots_desktop *desktop,
+	struct wlr_layer_surface *layer_surface, double lx, double ly,
+	double *sx, double *sy);
```
acrisci: This is a layer_surface centric function and has prefix `layer_surface` but has desktop as the first parameter, only uses the layout of the desktop, and lives in desktop.h.

acrisci: And also for consistency, should be `layer_surface_at` as in other parts of the code that give something from a position.

include/wlr/types/wlr_surface_layers.h
```diff
@@ -0,0 +1,71 @@
...
+	} events;
+
+	void *data;
+};
+
+enum wlr_layer_surface_invalid {
+	WLR_LAYER_SURFACE_INVALID_INTERACTIVITY = 1,
+	WLR_LAYER_SURFACE_INVALID_ANCHOR = 2,
+	WLR_LAYER_SURFACE_INVALID_EXCLUSIVE_ZONE = 4,
+	WLR_LAYER_SURFACE_INVALID_MARGIN = 8,
```
acrisci: We should really start using bitwise operators for things like this.

include/wlr/types/wlr_surface_layers.h
```diff
@@ -0,0 +1,71 @@
...
+
+	struct wl_listener surface_destroy_listener;
+	struct wl_listener surface_commit_listener;
+
+	struct {
+		struct wl_signal destroy;
+		struct wl_signal commit;
+	} events;
+
+	void *data;
```
acrisci: What is this for?

rootston/cursor.c
```diff
@@ -76,27 +78,153 @@ static void cursor_set_xcursor_image(struct roots_input *input,
...
+			focused_layer_surface = NULL;
+		}
+
+		struct wl_client *client =
+			wl_resource_get_client(layer_surface->resource);
+		struct wlr_seat_handle *handle =
+			wlr_seat_handle_for_client(input->wl_seat, client);
+
+		double sx, sy;
+		bool ok = layer_surface_is_at(input->server->desktop, layer_surface,
```
acrisci: Variable names like "ok" should be avoided. They always trip me up when I'm reading code.

SirCmpwn: That's a bit pedantic.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+	}
+	if (next->invalid & WLR_LAYER_SURFACE_INVALID_EXCLUSIVE_ZONE) {
+		state->exclusive_zone = next->exclusive_zone;
+	}
+	if (next->invalid & WLR_LAYER_SURFACE_INVALID_MARGIN) {
+		state->margin_horizontal = next->margin_horizontal;
+		state->margin_vertical = next->margin_vertical;
+	}
+
+	state->invalid |= next->invalid;
```
acrisci: probably better to call this `changed`.

SirCmpwn: I think we use invalid elsewhere

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+	.set_anchor = layer_surface_set_anchor,
+	.set_exclusive_zone = layer_surface_set_exclusive_zone,
+	.set_margin = layer_surface_set_margin,
+};
+
+static void handle_layer_surface_commit(struct wl_listener *listener,
+		void *data) {
+	struct wlr_layer_surface *surface =
+		wl_container_of(listener, surface, surface_commit_listener);
+	wlr_layer_surface_move_state(surface->pending, surface->current);
```
acrisci: There's probably no reason to put state moving into its own function since there are only two kinds of state and moving the state only happens here. I did that in the surface implementation because there are three states that need to be moved around.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+
+void wlr_layer_surface_configure(struct wlr_layer_surface *layer_surface,
+		int32_t width, int32_t height) {
+	layer_surface_send_configure(layer_surface->resource, width, height);
+}
+
+static void layer_surface_destroy(struct wlr_layer_surface *surface) {
+	wl_signal_emit(&surface->events.destroy, surface);
+	wl_list_remove(&surface->link);
+	wl_resource_set_user_data(surface->resource, NULL);
```
acrisci: I think you forgot to clean up your parent surface listeners here.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+	wl_signal_emit(&surface_layers->events.new_surface, layer_surface);
+}
+
+static const struct surface_layers_interface surface_layers_impl = {
+	.get_layer_surface = surface_layers_get_layer_surface,
+};
+
+static void surface_layers_bind(struct wl_client *wl_client,
+		void *_surface_layers, uint32_t version, uint32_t id) {
+	struct wlr_screenshooter *surface_layers = _surface_layers;
```
acrisci: How did a screenshooter get into this protocol?

types/wlr_surface_layers.c
acrisci: It's definitely wrong here, and probably wrong elsewhere too.

SirCmpwn: How do you figure? Invalidate is a common term in systems for when some kind of state is changed and has to be flushed somewhere.

acrisci: I guess you're right. Like "cache invalidation". These are the hard things in computer science I guess.

SirCmpwn: It's only hard when you invent new terms, using established ones is easy :wink:

rootston/cursor.c
```diff
@@ -76,27 +78,153 @@ static void cursor_set_xcursor_image(struct roots_input *input,
...
-		view = view_at(desktop, input->cursor->x, input->cursor->y,
-			&surface, &sx, &sy);
-		bool set_compositor_cursor = !view && input->cursor_client;
-		if (view) {
-			struct wl_client *view_client =
-				wl_resource_get_client(view->wlr_surface->resource);
-			set_compositor_cursor = view_client != input->cursor_client;
+		// Send events to non-exclusive layer surfaces, check if a layer surface
+		// gets exclusive events
+		remaining_layer_surfaces = &desktop->surface_layers->surfaces;
```
acrisci: You really have to drill "layer surface" into your head to get this straight. `surface_layers` really sounds like it should be a list of layer surfaces. What about calling the global the "layer surface shell"?

rootston/desktop.c
```diff
@@ -200,6 +201,35 @@ struct roots_view *view_at(struct roots_desktop *desktop, double lx, double ly,
...
+	double x, y;
+	wlr_layer_surface_get_position(layer_surface, &x, &y);
+	double layer_sx = lx - output_box->x - x;
+	double layer_sy = ly - output_box->y - y;
+
+	struct wlr_box box = {
+		.x = 0,
+		.y = 0,
+		.width = layer_surface->surface->current->buffer_width,
+		.height = layer_surface->surface->current->buffer_height,
```
acrisci: We make this mistake everywhere in the code and it's going to lead to a lot of bugs. But I think to actually get the width/height, you want to use `current->[width|height]` which takes into account the scale and transform of the buffer.

rootston/output.c
```diff
@@ -133,23 +136,52 @@ static void render_view(struct roots_view *view, struct roots_desktop *desktop,
 	}
 }
 
+struct wl_list *render_layer_surfaces(struct wl_list *layer_surfaces,
+		struct roots_desktop *desktop, struct wlr_output *wlr_output,
+		struct timespec *when, enum surface_layers_layer until_layer) {
```
acrisci: I think it would be a whole lot cleaner to have a function that renders surfaces by layer so you don't have to do math on the enum or return anything. But I'm probably doing a big rootston refactor pretty soon where I'll clean up a lot of stuff.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+	int output_width, output_height;
+	wlr_output_effective_resolution(layer_surface->output, &output_width,
+		&output_height);
+
+	if (state->anchor & LAYER_SURFACE_ANCHOR_LEFT) {
+		*x = state->margin_horizontal;
+	} else if (state->anchor & LAYER_SURFACE_ANCHOR_RIGHT) {
+		*x = output_width - width - state->margin_horizontal;
+	} else {
+		*x = (double)(output_width - width) / 2;
```
acrisci: According to the spec, you should consider the margin here when placing in the center.

rootston/cursor.c
SirCmpwn: To clarify the naming convention I had in mind when I originally wrote the protocol: *surface layers* refer to the layers in which surfaces may reside and the protocol as a whole. Layer surfaces is just short for "surface layer surface". I'm not torn up about how we name anything, just clarifying.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+			inserted = true;
+			wl_list_insert(&ls->link, &layer_surface->link);
+			break;
+		}
+	}
+	if (!inserted) {
+		wl_list_insert(&surface_layers->surfaces, &layer_surface->link);
+	}
+
+	wl_signal_emit(&surface_layers->events.new_surface, layer_surface);
```
acrisci: We probably don't have a buffer at this point. You should send this event on the first commit with a valid texture.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	layer_surface->resource = wl_resource_create(client,
+		&layer_surface_interface, wl_resource_get_version(resource), id);
+	wlr_log(L_DEBUG, "new layer_surface %p (res %p)", layer_surface,
+		layer_surface->resource);
+	wl_resource_set_implementation(layer_surface->resource, &layer_surface_impl,
+		layer_surface, layer_surface_resource_destroy);
+
+	layer_surface->surface_layers = surface_layers;
```
acrisci: If my shell gets destroyed, all these references dangle, right?

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+	layer_surface->surface = surface;
+	layer_surface->output = output;
```
acrisci: This will become invalid when the output is hotplugged, which will lead to memory errors.

types/wlr_surface_layers.c
```diff
@@ -0,0 +1,287 @@
...
+};
+
+static void surface_layers_bind(struct wl_client *wl_client,
+		void *_surface_layers, uint32_t version, uint32_t id) {
+	struct wlr_screenshooter *surface_layers = _surface_layers;
+	assert(wl_client && surface_layers);
+	if (version > 1) {
+		wlr_log(L_ERROR, "Client requested unsupported surface_layers version,"
+			"disconnecting");
+		wl_client_destroy(wl_client);
```
acrisci: This will never be reached. The `wl_registry` will give a protocol error and disconnect the client for you.

rootston/cursor.c
acrisci: > surface layers refer to the layers in which surfaces may reside and the protocol as a whole

So we have one term with three meanings. The protocol, an object in the protocol, and an attribute of an object in the protocol.

acrisci:
A big question on my mind right now is will any of this work with mult-seat configurations?

include/wlr/types/wlr_surface_layers.h
emersion: User-defined data, as in [other](https://github.com/swaywm/wlroots/blob/master/include/wlr/types/wlr_xdg_shell_v6.h#L18) [parts](https://github.com/swaywm/wlroots/blob/master/include/wlr/types/wlr_xdg_shell_v6.h#L135) [of](https://github.com/swaywm/wlroots/blob/master/include/wlr/types/wlr_output.h#L64) [the](https://github.com/swaywm/wlroots/blob/master/include/wlr/types/wlr_screenshooter.h#L19) [code](https://github.com/swaywm/wlroots/search?utf8=%E2%9C%93&q=void+*data&type=).

rootston/cursor.c
emersion: (@SirCmpwn edited his reply)

So yeah basically I'm trying to use terms consistently:
* _Surface layers_ for the protocol
* _Layer surface_ for a surface
* _Layer surfaces_ for multiple surfaces (such as in a list of surfaces)

rootston/output.c
emersion: That's what I did the first time, and that caused a lot of issues, in particular when managing pointer focus. Using the sorted list is cool because it allows to iterate through the list of layer surfaces only once.

types/wlr_surface_layers.c
emersion: Should this be added in the spec? (Something like _It's a protocol error to assign the layer surface role to a committed surface_?)

emersion:
>will any of this work with mult-seat configurations?

I've tried to think about it when implementing the feature. I think the only thing the could not work is how pointer events (and after that keyboard events) are sent to surfaces. Since there's one cursor per seat and since the list of cursor-focused layer surfaces is in `roots_input`, assuming there's one `roots_input` per seat, I think we're fine.

include/rootston/desktop.h
emersion: For the consistency point, I don't agree: `view_at` returns the view which is at a position, and this function does something different: it tells whether or not a surface layer is at a position. Hence the different names.

rootston/output.c
emersion: We can make this function only render one layer if we want to, though. Alternatively, maybe a `wlr_surface_layers_layer_cmp` function might be useful. What do you think?

types/wlr_surface_layers.c
emersion: >Requests that the surface be placed some distance away from the anchor point on the output, in pixels.

If there's no anchor point for an axis, I don't think it makes sense to take into account a margin.

emersion:
Okay, I think I addressed all comments. To test with more than 2 anchors, try:

```
layer-surface -i pointer -a right -a left -l bottom -m 10,0
```

acrisci:
Now when I unplug my monitor and plug it back in, my layer-surface disapears forever. The example needs to cover this case to be considered a complete implementation.

Also, the configure event doesn't work. It flashes unrendered content. The layer surface is a small box for one frame and then expands to the whole screen on the next frame.

rootston/output.c
acrisci: > Using the sorted list is cool because it allows to iterate through the list of layer surfaces only once.

If you're trying to optimize the code, please spend your time on the renderer instead :).

Yeah, I think rendering one layer at a time would help readability of this section quite a bit.

include/rootston/desktop.h
acrisci: That's true.

acrisci:
Also, it would be nice if the example had a random color everytime I opened (maybe based on the time) it so I can test the stack ordering.

emersion:
>The example needs to cover this case to be considered a complete implementation.

The example uses global variables to ease implementation. Covering this case would mean make the example more complex by removing these globals (making it also more difficult to understand). I don't know if it's worth it?

rootston/output.c
emersion: I don't think rendering one layer at a time helps readability. Also, this isn't the most complex function - there's another one in `cursor.c` that is harder to understand, and also returns a list of remaining layer surfaces. Ironically, I think making these functions more "simple" would complexify them.

acrisci:
> Covering this case would mean make the example more complex by removing these globals (making it also more difficult to understand).

Your example should be considered documentation on how to make an application that uses layer surfaces. This is an edge case that application developers will be required to handle and understand to make an implementation. A complicated basic example indicates a weakness in the protocol.

types/wlr_surface_layers.c
SirCmpwn: Don't worry about it in the center. We can update the spec if it's confusing.

SirCmpwn:
>Now when I unplug my monitor and plug it back in, my layer-surface disapears forever. The example needs to cover this case to be considered a complete implementation.

I agree with acrisci, without this we can't effectively test that the compositor side of things is handling hotplugging well, either.

>Also, the configure event doesn't work. It flashes unrendered content. The layer surface is a small box for one frame and then expands to the whole screen on the next frame.

Yeah definitely fix this too.

emersion:
>without this we can't effectively test that the compositor side of things is handling hotplugging well, either.

The compositor doesn't do anything about hotplugging and surface layers. The whole work is on the client side. But I see that it will be useful to have a good starting point for implementors.

>Also, the configure event doesn't work.

How does that work elsewhere? Right now: the client creates a surface, requests a surface layer, sets its state, and commits the surface with a buffer. Should it commit the surface with a null buffer to have a chance to get the configure event before starting to render?

SirCmpwn:
>How does that work elsewhere? Right now: the client creates a surface, requests a surface layer, sets its state, and commits the surface with a buffer. Should it commit the surface with a null buffer to have a chance to get the configure event before starting to render?

Hmm. What are we doing for xdg shell?

acrisci:
We send it a first configure request don't map it until it ackknowledged it.

emersion:
Added random colors and hotplugging (managed to do it while keeping globals).

>We send it a first configure request don't map it until it ackknowledged it.

We don't have an ack request on this interface.

acrisci:
When the output changes modes, you need to send a configure request to tell the layer surface what the new resolution of the screen is.

emersion:
You're right, that's a good idea.

Do you think adding a new ack request is necessary? Or a commit with an empty buffer is enough?

acrisci:
I'm not familiar enough with the client side to know what is most convenient for app developers.

See if you can get away with sending a configure to the client immediately after it creates the layer surface and then make it a protocol violation to have a buffer on your surface before the configure is sent.

emersion:
>See if you can get away with sending a configure to the client immediately after it creates the layer surface

The problem with this solution is that rootston is responsible for sending configure events, but rootston won't know about the layer surface before the first commit request is sent by the client. That's too late.

We could instead emit `wlr_surface_layers.new_surface` before the first commit when the layer surface is created, and check `wlr_layer_surface.configured` to prevent flashes of unrendered content. That would be simpler.

acrisci:
> The problem with this solution is that rootston is responsible for sending configure events

Move the responsibility of sending configure events entirely within wlroots.

SirCmpwn:
Are we maybe going to want the surface to send us its desired size in 0-2 dimensions so we can combine that with the anchors we do know to get the missing dimensions and know the final size of the surface?

SirCmpwn:
I think that would be without precedent in Wayland, fwiw

emersion:
>Move the responsibility of sending configure events entirely within wlroots.

That's possible right now because the current implementation is simple, but that would become more difficult with exclusive zones. I'm also thinking about compositors that draw themselves UI elements on the screen and want some of them not to be occluded.

Though, it would be also possible to add APIs to `wlr_surface_layers` to let the compositor specify additional zones that cannot be occluded (when that's needed).

I'm not sure whether it's better to let the compositor have full control over this or not.

acrisci:
> I'm not sure whether it's better to let the compositor have full control over this or not.

Keep in mind that the reason for letting the compositor have control over anything is because there is a hook for them to customize the behavior of something. This protocol is so tight about the size of things that there is exactly one correct configure request to send at exactly the right times.

In any case if the compositor's only choice is to violate the protocol, then the compositor doesn't need to make any decisions and wlroots can handle it.

SirCmpwn:
```
emersion   SirCmpwn:
emersion   >Are we maybe going to want the surface to send us its desired size in 0-2 dimensions so we can combine that with the anchors we do know to get the missing
           dimensions and know the final size of the surface?
emersion   what do you mean
SirCmpwn   uh, what I said
SirCmpwn   can you be more specific
emersion   the problem is to send to the surface its actual size
emersion   are you talking about a separate feature?
SirCmpwn   okay let me explain
emersion   or a way to takle this problem?
SirCmpwn   rootston is told where to anchor the surface to
emersion   tackle
SirCmpwn   this can mean that rootston is resposnible for assigning anywhere between 0 and 2 of the dimensions
SirCmpwn   0, for example, if anchoring to the left or bottom right or centor
SirCmpwn   1, for example, if anchoring to left top bottom
SirCmpwn   2, if anchored on all sides
SirCmpwn   any dimensions it isn't responsible it needs to know from the client
SirCmpwn   but the client needs to know the dimension rootston is picking, too
SirCmpwn   all of this needs to be sorted before we render shit so we don't get an FOUC
SirCmpwn   correct?
emersion   correct
SirCmpwn   so, here's a proposed flow for this - though I don't know if it's the right approach
SirCmpwn   1. surface is assigned the layer surface role with a layer and a set of anchors
SirCmpwn   2. surface lets rootston know the dimension(s) it's responsible for
SirCmpwn   3. rootston lets surface know the dimension(s) it's responsible for
SirCmpwn   4. buffer attached and committed
SirCmpwn   5. rootston renders it
SirCmpwn   slightly more complicated to allow for double buffering the state xdg-style so we can change anchors and shit later with FOUC similarly accounted for
SirCmpwn   also 2 and 3 can be whatever order
emersion   i'm maybe mistaken, but that's more or less what we want to do too:
emersion   1. surface is assigned the layer surface role with a layer and a set of anchors
emersion   2. rootston or wlroots (we discussing about that) is sending configure with the correct size for dimensions it's responsible for (zero for others)
SirCmpwn   ftr wlroots should do this, not rootston. my bad
emersion   3. surface fills missing dimensions with whatever it wants
emersion   4. buffer attached and commited
emersion   5. rootston renders it
SirCmpwn   how specifically does step 3 happen
emersion   3. surface receives a configure event with zeroes for dimensions it's responsible for, and non-zeroes for dimensions picked by the compositor
emersion      it fills the zeroes, and creates a surface with this size
SirCmpwn   you mean a buffer?
emersion   ugh, yeah
SirCmpwn   okay cool
SirCmpwn   and later the client changes the anchor and steps 3-5 happen again
SirCmpwn   right?
emersion   exactly
SirCmpwn   sounds good, +1 to that flow
emersion   so now the question is:
emersion   do we want wlroots or rootston to decide how the configure event is sent
SirCmpwn   wlroots
emersion   okay
SirCmpwn   do we need rootston to make some decisions for wlroots during this exchange?
SirCmpwn   I don't think so
emersion   so, if a compositor draws some UI
emersion   and wants that UI not to be occluded by an exclusive zone
emersion   that makes things more complicated
SirCmpwn   wlroots just bubbles up the exclusive zone to rootston
SirCmpwn   it's rootston's business to deal with that, right?
SirCmpwn   since it handles window placement and rendering
SirCmpwn   it can just stir its own exclusive zone into the mix
emersion   so, that's another part of the problem
emersion   if wlroots doesn't care about exclusive zones
emersion   it cannot send the configure event
emersion   there are issues with: two bars, one bar and a background
SirCmpwn   how do you mean?
emersion   hmm, two bars is fine
SirCmpwn   ftr exclusive zone only applies to shell views
SirCmpwn   well
emersion   but, let's take for instance a bar and a desktop with icons
SirCmpwn   hmmmmmmmmmmmmm
SirCmpwn   this is a quandry
SirCmpwn   I want wallpapers to be not affected by exclusive zones (they would extend under a panel)
SirCmpwn   but things like desktop icons or notifications probably _should_ be affected
emersion   yeah, wallpapers are probably fine
SirCmpwn   I think maybe we define exclusive zones to only be effective in the middle two layers?
SirCmpwn   we don't have to define that in the protocol, btw
SirCmpwn   I think it's explicitly said that the compositor can do whatever it wants with the exclusive zone
emersion   i don't see how that solves the issue
SirCmpwn   well
SirCmpwn   wallpaper goes in layer 0
SirCmpwn   bar goes in layer 1
SirCmpwn   notifications go in layer 2
SirCmpwn   lock screen layer 3
SirCmpwn   only 1 and 2 care about exclusive zones
SirCmpwn   it solves an issue I just thought of, perhaps not the issue you're asking about
emersion   ahah
emersion   bar in layer 1?
SirCmpwn   yeah
emersion   under windows?
SirCmpwn   yep
SirCmpwn   BUT
SirCmpwn   it has an exclusive zone defined
SirCmpwn   so maximized or tiled shell windows don't occlude it
SirCmpwn   if they're floated though you could drag them over it
emersion   okay
emersion   not all compositors will want that
SirCmpwn   sure, but they're allowed to do whatever they want with exclusive zones per the spec
emersion   yeah
emersion   so, back to the problem
emersion   should the compositor decide the dimension of a layer surface? if not, how does wlroots deals with exclusive zones?
emersion   a desktop with icons would go in layer 0 or 1
SirCmpwn   1
emersion   yeah, 1
SirCmpwn   it would be anchored on all sides
emersion   and a bar would also be in layer 1
emersion   anchored on top
SirCmpwn   and the bottom anchor (or w/e) would be pushed out due to the bar's exclusive zone
SirCmpwn   top, then
SirCmpwn   plus any margins
emersion   yeah
SirCmpwn   anyway, yes, the compositor should keep track of this and make the calls
SirCmpwn   bubble it up
```

emersion:
Updated the protocol to reflect these changes.

emersion:
Status recap:

* The protocol needs a redesign to support subsurfaces, popups (to be able to support a full-featured bar for instance)
* Make it more simple to implement
* Rendering issues where unrendered content is flashed during configuration have been fixed
* Think about multiseat support: the protocol will probably need to be seat-aware

emersion:
Sorry, deleted the branch by mistake.

emersion:
We'll start form scratch.

PR #259, opened 2017-10-11
Fixed wrong function name

Enerccio:
Apparently I am first guy to ever call this...


ascent12:
I'm pretty sure I was the one who wrote that, which would have been several months ago. Thanks for catching that.

Enerccio:
No problem

PR #260, opened 2017-10-11
Destroy surfaces on client destruction (XDG shell)

versusvoid:
Prevents accessing freed `wlr_xdg_client_v6` from `xdg_surface_destroy()` after it has been freed.
Prevents segfault when closing `weston-transformed`.


SirCmpwn:
Thanks!

acrisci:
I think this is a protocol error on the part of the client:

> Destroying a bound xdg_shell object while there are surfaces still alive created by this xdg_shell object instance is illegal and will result in a protocol error.

PR #276, opened 2017-10-16
bug: fix view centering

acrisci:
Make view initialization never fail instead of doing the initialization at some point in the future and flashing partially rendered content:

1. fix xdg-shell so it emits the new event with window geometry (the first commit after ack_configure)
2. pick the center output if no output is found under the cursor as a sensible default

Also, use the effective resolution of the output to determine centering so it works correctly on rotated outputs (this is understandably confusing and probably needs to be made clearer).

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -343,3 +343,20 @@ void wlr_output_layout_add_auto(struct wlr_output_layout *layout,
 	l_output->state->auto_configured = true;
 	wlr_output_layout_reconfigure(layout);
 }
+
+struct wlr_output *wlr_output_layout_get_center_output(
+		struct wlr_output_layout *layout) {
+	if (wl_list_empty(&layout->outputs)) {
+		return NULL;
+	}
+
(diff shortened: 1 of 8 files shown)


types/wlr_output_layout.c
```diff
@@ -343,3 +343,20 @@ void wlr_output_layout_add_auto(struct wlr_output_layout *layout,
...
+
+	struct wlr_box *extents = wlr_output_layout_get_box(layout, NULL);
+	double center_x = extents->width / 2 + extents->x;
+	double center_y = extents->height / 2 + extents->y;
+
+	double dest_x = 0, dest_y = 0;
+	wlr_output_layout_closest_point(layout, NULL, center_x, center_y,
+		&dest_x, &dest_y);
+
+	return wlr_output_layout_output_at(layout, dest_x, dest_y);
```
SirCmpwn: wlr_output_layout is a nice interface

SirCmpwn:
LGTM thanks!

PR #278, opened 2017-10-17
Add CRTC id to atomic DRM request only once

### LINKED ISSUE #204: "Atomic commit failed (modeset): Invalid argument" after switching to Gnome and backward
Reproduced as (assuming Gnome is already running)
1. Start rootston from tty
2. Switch to Gnome
Screen blinks. Looks like Gnome makes clean DRM initialization.
3. Switch back
Gnome interface will be left frozen on screen.

Rootston log says:
```
2017-10-05 07:48:52 - [backend/drm/backend.c:80] DRM fd paused
2017-10-05 07:48:55 - [backend/drm/backend.c:65] DRM fd resumed
2017-10-05 07:48:55 - [backend/drm/atomic.c:57] VGA-1: Atomic commit failed (modeset): Invalid argument
2017-10-05 07:48:55 - [backend/drm/atomic.c:57] HDMI-A-1: Atomic commit failed (modeset): Invalid argument
```

Not reproduced with _weston_ or _xfwm4_. But when switching from Gnome to _weston_ or _xfwm4_ screen also blinks, so maybe they detect such situations and make clean DRM initialization too.

versusvoid:
Fix #204 
Note, CRTC's id added second time in `set_plane_props()`.

Tested with Gnome: changed display resolution, switched back to rootston, mode got successfully.

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -114,7 +114,6 @@ static bool atomic_crtc_pageflip(struct wlr_drm_backend *drm,
 	struct atomic atom;
 
 	atomic_begin(crtc, &atom);
-	atomic_add(&atom, conn->id, conn->props.crtc_id, crtc->id);
 	atomic_add(&atom, crtc->id, crtc->props.mode_id, crtc->mode_id);
 	atomic_add(&atom, crtc->id, crtc->props.active, 1);
 	set_plane_props(&atom, crtc->primary, crtc->id, fb_id, true);


ascent12:
As I said in #204, this is not a valid solution to the issue. It's just being lucky that Gnome picked the same CRTC as us. It does not work in general.
Also, the CRTC ID being set in set_place_props is different. Both connectors and planes have separate properties to set the CRTC they're attached to.

In fact, I'm surprised it's not completely breaking modesetting for you, but that's probably just you being lucky again with the CRTCs happening to already be correct.

PR #279, opened 2017-10-17
bug: detect invalid texture in xdg-shell

acrisci:
Use `surface->texture->valid` instead of `current->buffer != null` to detect whether the surface has a buffer. This fixes configure requests, since now we upload the buffer and set it to NULL immediately, `current->buffer` is always NULL.


SirCmpwn:
Thanks!

PR #280, opened 2017-10-17
xdg-shell: force configure when pending different

acrisci:
This bug exists in Weston as well.

Run rootston (or weston) like this:

```
(sleep 1 && gnome-terminal) &
(sleep 1 && gnome-terminal) &
rootston
```

You will see that both of the windows are shown as activated. This is because the pending surface state is set to `activated` and then fipped back to deactivated before the ack_configure (which is the time at which the `next` state is updated) so the state_compare function says there is no difference between pending/current/next state and the configure request is deemed not necessary.

cc @SardemFF7

### DIFF
--- a/types/wlr_xdg_shell_v6.c
+++ b/types/wlr_xdg_shell_v6.c
@@ -840,6 +840,10 @@ static const struct zxdg_surface_v6_interface zxdg_surface_v6_implementation = {
 static bool wlr_xdg_surface_v6_toplevel_state_compare(
 		struct wlr_xdg_toplevel_v6 *state) {
 	// is pending state different from current state?
+	if (!state->base->configured) {
+		return false;
+	}
+
 	if (state->pending.activated != state->current.activated) {
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


SirCmpwn:
>This bug exists in Weston as well.

Shoot wayland-devel an email about it?

sardemff7:
This commit is actually short-circuiting the whole feature. Also, `state_compare` does not check `next` at all, it’s only about the currently used state (the one on screen) and the pending state (the one we will ask the client for).

I can see two possible fixes that keep the optimization useful:
1. schedule a configure on `committed`: will do nothing if pending is still the same as current, but in the bug case, pending change, we’re good
2. make `state_compare` return false if the configure list is not empty, because it means we sent a different-than-current state already

I think solution 1 can somehow flicker, especially if rendering is slow, since the client will have to render a “wrong” surface before sending the “good” state. While solution 2 may have some corner cases, like, while in state A, sending B then A then B will actually send 3 configure, while sending the first B would have been enough. But good clients should ignore first B and first A anyway, so it should be fine.

I’ve [sent a patch](https://patchwork.freedesktop.org/patch/183214/) for solution 2 for libweston-desktop.

sardemff7:
Just pushed an updated version of the patch:
https://cgit.freedesktop.org/wayland/weston/commit/?id=efade28db954ad02ac283a626238df04bae146da

PR #282, opened 2017-10-18
xdg-shell: send same-as-current configure if needed

acrisci:
Keeping an eye on this:

https://patchwork.freedesktop.org/patch/183214/


acrisci:
New patch:

https://cgit.freedesktop.org/wayland/weston/commit/?id=efade28db954ad02ac283a626238df04bae146da

acrisci:
It should be good now.

SirCmpwn:
Thanks!

PR #286, opened 2017-10-18
Update Connector-CRTC bindings after resume in DRM backend

### LINKED ISSUE #204: "Atomic commit failed (modeset): Invalid argument" after switching to Gnome and backward

versusvoid:
Fix #204

~~Fix `match_obj()` behaviour on `[0, 0b11], [UNMATCHED, 1]`~~ (not required)
- [x] Fix deactivated monitors check

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -62,10 +62,13 @@ static void session_signal(struct wl_listener *listener, void *data) {
 
 	if (session->active) {
 		wlr_log(L_INFO, "DRM fd resumed");
+		wlr_drm_scan_connectors(drm);
 
 		struct wlr_drm_connector *conn;
 		wl_list_for_each(conn, &drm->outputs, link){
-			wlr_drm_connector_start_renderer(conn);
+			if (conn->output.current_mode) {
+				wlr_output_set_mode(&conn->output, conn->output.current_mode);
+			}
 
 			if (!conn->crtc) {
 				continue;
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -314,7 +314,8 @@ static void realloc_crtcs(struct wlr_drm_backend *drm,
 		struct wlr_drm_connector *conn, bool *changed_outputs) {
 	uint32_t crtc[drm->num_crtcs];
 	uint32_t crtc_res[drm->num_crtcs];
-	uint32_t possible_crtc[wl_list_length(&drm->outputs)];
+	ssize_t num_outputs = wl_list_length(&drm->outputs);
+	uint32_t possible_crtc[num_outputs];
 
 	for (size_t i = 0; i < drm->num_crtcs; ++i) {
 		crtc[i] = UNMATCHED;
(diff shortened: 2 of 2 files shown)


ascent12:
I don't like this solution. It only provides a patchwork solution instead of solving the actual core problem.

We shouldn't need to go through the potentially very expensive operation of reconfiguring on VT switch. The old configuration we have should still be good, and we should be able to just keep using it.

I liked the "commit everything at once" solution, but it just needed to be hidden behind the atomic interface properly, and modified to reduce a bunch of code duplication.

versusvoid:
"commit everything at once" means "I don't care what configuration you have, I want you to reconfigure to exact this state"
What is done here is "Get current configuration and try to change it as little as possible".

I see room for improvements in splitting `wlr_drm_connector_set_mode()` and calling `wlr_drm_connector_start_renderer()` only once per connector.

SirCmpwn:
@ascent12 any new comments?

backend/drm/backend.c
```diff
@@ -57,16 +57,69 @@ bool wlr_backend_is_drm(struct wlr_backend *b) {
 	return b->impl == &backend_impl;
 }
 
+static void wlr_drm_connector_update_crtc_binding(struct wlr_drm_connector *conn,
```
ascent12: This type of function belongs in drm.c.

backend/drm/backend.c
```diff
@@ -57,16 +57,69 @@ bool wlr_backend_is_drm(struct wlr_backend *b) {
...
+		struct wlr_drm_backend *drm) {
+	drmModeConnectorPtr drm_conn = drmModeGetConnector(drm->fd, conn->id);
```
ascent12: `drmModeConnector *drm_conn`
I really don't like typedefs on pointers.

backend/drm/backend.c
```diff
@@ -57,16 +57,69 @@ bool wlr_backend_is_drm(struct wlr_backend *b) {
...
+	if (!drm_conn) {
+		wlr_log_errno(L_ERROR, "Failed to get DRM connector %u", conn->id);
+		return;
+	}
+	if (!drm_conn->encoder_id) {
+		conn->crtc = NULL;
+		drmModeFreeCrtc(conn->old_crtc);
+		conn->old_crtc = NULL;
+		conn->state = WLR_DRM_CONN_DISCONNECTED;
```
ascent12: `wlr_drm_connector_cleanup(conn)`

backend/drm/backend.c
```diff
@@ -57,16 +57,69 @@ bool wlr_backend_is_drm(struct wlr_backend *b) {
...
+
+		drmModeFreeConnector(drm_conn);
+		return;
+	}
+
+	drmModeEncoderPtr drm_encoder = drmModeGetEncoder(drm->fd,
```
ascent12: Same as drm_conn

backend/drm/backend.c
```diff
@@ -57,16 +57,69 @@ bool wlr_backend_is_drm(struct wlr_backend *b) {
...
+				conn->crtc = &drm->crtcs[i];
+			}
+		}
+	}
+
+	drmModeFreeEncoder(drm_encoder);
+	drmModeFreeConnector(drm_conn);
+}
+
+static void wlr_drm_backend_update_connector_crtc_bindings(
```
ascent12: I wouldn't bother with a function this small, especially if it's only going to be used in one place.
I would just write the for loop where it's going to be used.

backend/drm/drm.c
```diff
@@ -416,7 +424,8 @@ static bool wlr_drm_connector_set_mode(struct wlr_output *output,
 		goto error_conn;
 	}
 
-	realloc_crtcs(drm, conn);
+	memset(changed_outputs, false, sizeof(bool) * drm->outputs->length);
```
ascent12: `sizeof(changed_outputs)`

ascent12:
This is going to mess with #290, so I'm going to wait for that to be finished and have the list_t related changes in this PR before I would merge it.

backend/drm/backend.c
SirCmpwn: typedefs are not allowed at all in our codebase.

ascent12: It's not our typedef, but they provide both a struct and a pointer to struct typedef in libdrm for some stupid reason.

versusvoid:
`wlr_drm_connector_cleanup()` release surfaces, planes and emits `output_remove` signal. None of it is required here, so I'm not calling it.

versusvoid:
I got rid of new function and set `crtc` field in `wlr_drm_scan_connectors()`. It's effectively the same but with less code.
Current CRTCs are always considered when matching with connectors now. This is workaround _unoptimal_ result of `match_obj()`.

PR #288, opened 2017-10-18
rootston: default meta key

acrisci:


emersion:
Sensible defaults are here:

https://github.com/swaywm/wlroots/blob/master/rootston/config.c#L300

Maybe we want to move that piece there?

acrisci:
> Sensible defaults are here:

Those only get applied if no config file is found.

SirCmpwn:
If you provide a config file it's your responsibility to make sure it's good.

emersion:
On the same note, there's a `config->xwayland = true;` right after, if we don't want a default meta key we may not want to enable xwayland by default too (but enable it if no config file is specified).

SirCmpwn:
I could go either way on that change.

PR #289, opened 2017-10-18
rootston: add a default meta key if no config file is specified

emersion:


SirCmpwn:
Good call, missed this in the last PR.

PR #290, opened 2017-10-18
Use vertex buffer objects

ascent12:
These are just some pretty mundane changes to the GLES renderer. In my extremely limited testing, using VAOs/VBOs instead of uploading the same vertex data over and over lowers the CPU usage _slightly_. There is still a lot more to do, but this is a start.

There are also some minor changes to cleanup the shaders.


SirCmpwn:
I'm not sure these extensions are going to be consistently available. VAOs are a GLES3 feature. I have a laptop that doesn't support GLES3 I'll have to test this with.

SirCmpwn:
That was the main motivator behind switching from gles3 to gles2 by the way.

render/gles2/shaders.c
```diff
@@ -3,100 +3,73 @@
...
-"    vec4 i1 = inMatrix[1];"
-"    vec4 i2 = inMatrix[2];"
-"    vec4 i3 = inMatrix[3];"
-"    mat4 outMatrix = mat4("
-"                 vec4(i0.x, i1.x, i2.x, i3.x),"
-"                 vec4(i0.y, i1.y, i2.y, i3.y),"
-"                 vec4(i0.z, i1.z, i2.z, i3.z),"
-"                 vec4(i0.w, i1.w, i2.w, i3.w)"
-"                 );"
-"    return outMatrix;"
```
SirCmpwn: Where did this logic go?

ascent12:
Mesa supports these on every driver (see https://github.com/mesa3d/mesa/blob/master/docs/features.txt), considering they're a required feature of desktop OpenGL 3.0.
A lot of the GL extensions are ubiquitous,  so I think they're fine to use.

render/gles2/shaders.c
ascent12: https://github.com/swaywm/wlroots/pull/290/commits/7170170a5778a0bbb37827f85882cde4c7762ec5#diff-650ac2c117ab1d2399640eab8ba3ab74R170

SirCmpwn: Hah, derp. Good catch.

SirCmpwn:
I was right - this doesn't work on older hardware. I tested this on a ThinkPad X200 to no avail.

I'm 100% on board with VAOs, but we need to fall back. I think it would be better to implement a gles3 renderer and choose whichever one is most suitable for the hardware (gles3 will fail to initialize on older machines, so we can just ditch it and switch to gles2 and it should Just Werk).

SirCmpwn:
We used to have a gles3 backend, by the way, should be fairly straightforward to bring it back to life.

ascent12:
I'm pretty sure most of the real advantages are coming from using VBOs, and the VAOs only making it nicer to use, so I removed them.
It really is annoying having to limit it to GLES2, but I really can't be bothered writing multiple ways of trying to do the same thing.

SirCmpwn:
That doesn't work either. You have to actually test if it's supported before you use it, or make a separate gles3 renderer.

ascent12:
VBOs are standard GLES2.
I'm not 100% I'm using them "properly", since I've never used them without VAOs before, but the code did work for me.

SirCmpwn:
I will spend some more time with it and see what I can find out.

ascent12:
The non-VBO changes have been integrated into #319

PR #296, opened 2017-10-19
rootston: view set position

### LINKED ISSUE #292: xwayland: configure windows as they change position
Xwayland windows seem to need to know their x/y position in the layout.

This causes a bug where if you move the window around with a compositor action (mod+click and drag), then try to move the window with a window action (drag the title bar), the window move action will cause the window to jump to the place where the last configure event was received (probably when the window was mapped).

I think we need to send configure events to xwayland windows when the compositor moves them around. The view should be extended with a `set_position` function that handles position setting and configures the xwayland window when it moves.

acrisci:
Add `view_set_position()` which configures xwayland windows and fix various window initialization bugs.

fixes #292


SirCmpwn:
Thanks!

PR #297, opened 2017-10-19
wl-output: send layout position

### LINKED ISSUE #284: Pointer events on Xwayland are buggy as hell
I'm still seeing pointer events sent to the wrong clients and in the wrong places all the time. It's especially buggy when you have multiple outputs.

### LINKED ISSUE #291: xrandr doesn't show correct layout info
I think this is causing a bug where xwayland windows can't move beyond a certain point in the layout.

xrandr shows the screens all being stacked on top of each other even though they are arranged next to each other in the layout.

I think I can fix this by setting the correct x/y in `wl_output` interface.

```
Screen 0: minimum 320 x 200, current 1920 x 1920, maximum 8192 x 8192
XWAYLAND0 connected 1080x1920+0+0 left (normal left inverted right x axis y axis) 480mm x 270mm
   1920x1080     59.96*+
XWAYLAND1 connected 1080x1920+0+0 right (normal left inverted right x axis y axis) 480mm x 270mm
   1920x1080     59.96*+
XWAYLAND2 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 480mm x 270mm
   1920x1080     59.96*+
```

acrisci:
fixes #291 
fixes #284

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -56,6 +56,9 @@ struct wlr_output {
 		struct wl_listener surface_destroy;
 	} cursor;
 
+	// the output position in layout space reported to clients
+	int32_t lx, ly;
+
 	void *data;
 };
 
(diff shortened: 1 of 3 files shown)


acrisci:
This is just a temporary solution because I'm trying to get Xwayland to work and this fixes bugs. #299 will change this, but that's a big effort and I don't want to work on it now.

SirCmpwn:
Thanks!

Issue #264, opened 2017-10-12
Remove list_t


ascent12:
After talking about it on IRC, we've decided that we don't really need this any more and should replace it with wl_list.

SirCmpwn:
We refactored a lot of this away, and there's still room to remove it more, but in some places I still think it's the most suitable data structure for the problem. I did rename it to wlr_list, though.

PR #298, opened 2017-10-20
Replace list_t with wl_list

raazvvann:
partial work for #264


backend/multi/backend.c
```diff
@@ -65,8 +65,8 @@ struct wlr_backend *wlr_multi_backend_create(struct wlr_session *session) {
 		return NULL;
 	}
 
-	backend->backends = list_create();
-	if (!backend->backends) {
+	wl_list_init(&backend->backends);
+	if (!&backend->backends) {
```
emersion: I don't think this can fail.

raazvvann: that's right. missed this

raazvvann:
I am still working on replacing list_t with wl_list in wlr_wl_backend now. I'll let you know if I manage to do more on this.

acrisci:
You don't need to free the lists.

SirCmpwn:
Thanks!

Issue #214, opened 2017-10-05
Add startup option to rootston


SirCmpwn:
With a command to run on startup. Not associated with a section.

PR #301, opened 2017-10-20
Add startup command option to rootston

thejan2009:
Allows specifying a command to run at startup.

Closes: #214


SirCmpwn:
Nice, thanks!

PR #303, opened 2017-10-21
Remove wl_list_init when using wl_signal_add

### LINKED ISSUE #283: Unnecessary wl_list_init on wl_listener.link
Cleanup those.

See https://github.com/swaywm/wlroots/pull/258#discussion_r145361921

>According to https://people.freedesktop.org/~whot/wayland-doxygen/wayland/Server/structwl__listener.html, we don't need to do it. Right now we do it in a lot of places, we should clean up the whole codebase.

emersion:
I removed some fields in rootston views because they were not used. Calling `wl_list_remove` on a listener not initialized with `wl_signal_add` segfaults.

Also cleanups rootston:
* Add missing allocation checks
* Remove assertions used to check runtime errors

Fixes #283

### DIFF
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -202,11 +202,18 @@ struct roots_view *view_at(struct roots_desktop *desktop, double lx, double ly,
 
 struct roots_desktop *desktop_create(struct roots_server *server,
 		struct roots_config *config) {
-	struct roots_desktop *desktop = calloc(1, sizeof(struct roots_desktop));
-	assert(desktop);
 	wlr_log(L_DEBUG, "Initializing roots desktop");
 
-	assert(desktop->views = list_create());
+	struct roots_desktop *desktop = calloc(1, sizeof(struct roots_desktop));
+	if (desktop == NULL) {
+		return NULL;
+	}
+
+	desktop->views = list_create();
+	if (desktop->views == NULL) {
+		free(desktop);
+		return NULL;
+	}
 	wl_list_init(&desktop->outputs);
 
 	desktop->output_add.notify = output_add_notify;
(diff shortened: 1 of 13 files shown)


rootston/desktop.c
```diff
@@ -203,16 +203,15 @@ struct roots_view *view_at(struct roots_desktop *desktop, double lx, double ly,
 struct roots_desktop *desktop_create(struct roots_server *server,
 		struct roots_config *config) {
 	struct roots_desktop *desktop = calloc(1, sizeof(struct roots_desktop));
+	assert(desktop);
```
ascent12: You shouldn't use assert for runtime errors.

emersion: Okay. I've seen `assert` used to check allocations elsewhere in rootston's code - I'll remove them.

emersion:
There are still a bunch of `assert` calls to check runtime errors [in `rootston/main.c`](https://github.com/swaywm/wlroots/blob/master/rootston/main.c#L16), should I remove these too?

acrisci:
> There are still a bunch of assert calls

Yeah, if you run out of memory in rootston, you should display a nice message and exit 1 I believe.

rootston/wl_shell.c
```diff
@@ -57,11 +57,8 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
 	struct roots_wl_shell_surface *roots_surface =
 		wl_container_of(listener, roots_surface, destroy);
 	wl_list_remove(&roots_surface->destroy.link);
-	wl_list_remove(&roots_surface->ping_timeout.link);
```
acrisci: Why did these go? You need to remove the listener link to stop listening to the event.

emersion: Because `ping_timeout` and the others are not used, there's no listener. See the header: https://github.com/swaywm/wlroots/pull/303/files#diff-fe0ba5038842e9d235dfd2047ed4d031L28

SirCmpwn:
Thanks!

PR #304, opened 2017-10-21
Moved comments in rootston keybinding example config

inclement:
The comment strings are not stripped so the commands are not currently recognised.


SirCmpwn:
Thanks!

PR #308, opened 2017-10-21
Render fixups

ascent12:
Various small changes to some renderer related things.


render/gles2/renderer.c
```diff
@@ -211,20 +211,9 @@ static bool wlr_gles2_buffer_is_drm(struct wlr_renderer *_renderer,
...
-		uint32_t a = v & 0x000000ff;
-		data[i] = rgb | (a << 24);
-	}
-}
-
 static void wlr_gles2_read_pixels(struct wlr_renderer *renderer, int x, int y,
 		int width, int height, void *out_data) {
-	glReadPixels(x, y, width, height, GL_RGBA, GL_UNSIGNED_BYTE, out_data);
-	rgba_to_argb(out_data, height, width*4);
+	glReadPixels(x, y, width, height, GL_BGRA_EXT, GL_UNSIGNED_BYTE, out_data);
```
emersion: BGRA != ARGB, is this change intended?

ascent12: Yes. OpenGL will actually read the pixels in the reverse order than you expect.
It's what happens in the DRM cursor function as well.

emersion: Not working for me, here's what the screenshooter outputs:

wayland-screenshot

emersion: Though I know that OpenGL reads images upside down, so data must be y-flipped (that happens in `wlr_screenshooter`).

ascent12: I rolled back that commit

PR #309, opened 2017-10-21
Changed broken example rootston keybinding

inclement:
Since `Shift` is part of the `Logo+Shift+e` keybinding, the `e` keysym is never received, only `E`.

Should rootston be detecting the presence of `Shift` and automatically working out that it should switch the expected keysyms to uppercase (I guess by querying the xkb layout)? I looked into it briefly, but couldn't tell if there is any simple way to do this.

### DIFF
--- a/rootston/rootston.ini.example
+++ b/rootston/rootston.ini.example
@@ -40,6 +40,6 @@ meta-key = Logo
 # - "close" to close the current view
 # - "next_window" to cycle through windows
 [bindings]
-Logo+Shift+e = exit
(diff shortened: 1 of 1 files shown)


SirCmpwn:
>Should rootston be detecting the presence of Shift and automatically working out that it should switch the expected keysyms to uppercase

I guess, but I don't really care. Fix it if it pleases you. Rootston isn't designed to be a particularly useful compositor, beyond demonstrating wlroots features I don't care how well it works.

PR #312, opened 2017-10-22
Fix index computation in DRM output scan and CRTC match

versusvoid:


backend/drm/drm.c
```diff
@@ -686,8 +686,7 @@ void wlr_drm_scan_connectors(struct wlr_drm_backend *drm) {
 	}
 
 	size_t seen_len = wl_list_length(&drm->outputs);
-	// +1 so it can never be 0
-	bool seen[seen_len + 1];
+	bool seen[seen_len];
```
ascent12: Trying to create a 0-length VLA is undefined behaviour, which is why that +1 was there. seen_len would be zero the first time this function is called.

PR #315, opened 2017-10-22
[WIP] Redesign keyboard interface

### LINKED ISSUE #287: Redesign keyboard interface
This is a proposal to redesign the keyboard interface to be able to fix #242.

Right now, keyboard events follow this flow:

```
                                    +----> [ rootston ]
                                    |
[ backend ] +------> [ wlroots ] +--+
                                    |
                                    +----> [ client ]
```

So `rootston` has no control over keyboard events. It just selects the client to which events are sent to with `wlr_seat_keyboard_notify_enter` and `wlr_seat_keyboard_clear_focus`.

The idea is to change the flow to something similar to pointer events:

```
[ backend ] +---> [ wlroots ] +---> [ rootston ] +---> [ wlroots ] +---> [ client ]
```

With these functions:

* `wlr_seat_keyboard_send_key`, `wlr_seat_keyboard_send_modifiers` used by backends [same as right now]
* `wlr_seat_keyboard_notify_enter` and `wlr_seat_keyboard_clear_focus` used by rootston [same as right now]
* `wlr_seat_keyboard_notify_key`, `wlr_seat_keyboard_notify_modifiers` used by rootston [new]

Keyboard events won't be forwarded automatically by wlroots anymore, it'll be the compositor's role to listen to them and to call the two new functions.

emersion:
Fixes #287

### DIFF
--- a/include/wlr/types/wlr_keyboard.h
+++ b/include/wlr/types/wlr_keyboard.h
@@ -70,9 +70,13 @@ struct wlr_event_keyboard_key {
 	enum wlr_key_state state;
 };
 
+struct wlr_input_device;
+
+struct wlr_keyboard *wlr_keyboard_create(struct wlr_input_device *dev);
 void wlr_keyboard_set_keymap(struct wlr_keyboard *kb,
 	struct xkb_keymap *keymap);
 void wlr_keyboard_led_update(struct wlr_keyboard *keyboard, uint32_t leds);
 uint32_t wlr_keyboard_get_modifiers(struct wlr_keyboard *keyboard);
+void wlr_keyboard_destroy(struct wlr_keyboard *kb);
 
 #endif
--- a/include/wlr/interfaces/wlr_keyboard.h
+++ b/include/wlr/interfaces/wlr_keyboard.h
@@ -9,11 +9,9 @@ struct wlr_keyboard_impl {
 	void (*led_update)(struct wlr_keyboard *keyboard, uint32_t leds);
 };
 
-void wlr_keyboard_init(struct wlr_keyboard *keyboard, struct wlr_keyboard_impl *impl);
-void wlr_keyboard_destroy(struct wlr_keyboard *keyboard);
-void wlr_keyboard_notify_key(struct wlr_keyboard *keyboard,
+void wlr_keyboard_send_key(struct wlr_keyboard *keyboard,
 		struct wlr_event_keyboard_key *event);
-void wlr_keyboard_notify_modifiers(struct wlr_keyboard *keyboard,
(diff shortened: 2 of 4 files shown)


include/wlr/types/wlr_keyboard.h
```diff
@@ -70,9 +70,13 @@ struct wlr_event_keyboard_key {
 	enum wlr_key_state state;
 };
 
+struct wlr_input_device;
+
+struct wlr_keyboard *wlr_keyboard_create(struct wlr_input_device *dev);
```
emersion: I'm not sure how this could work. Currently, the `wlr_keyboard` is created by the backend, and then the backend calls `wlr_keyboard_send_{key,modifiers}` to sned events to the `wlr_keyboard`.

The `wlr_input_device` contains the `wlr_keyboard_interface` and we need a way to ask the backend to send key events to a `wlr_keyboard`. Maybe `wlr_keyboard_interface.attach(wlr_keyboard *kb)`?

SirCmpwn: Slow down. The user should not be creating wlr keyboards, the backend should still be doing that. The approach here is wrong.

emersion: Okay, I wanted to do this part:

>When a device is added, rootston creates a wlr-keyboard from the device and configures it based on compositor policy.

SirCmpwn: I see. That's probably not the best design. We should still have the backend create a wlr_keyboard and push it up to rootston. The full quote is:

>We should transfer ownership of the wlr-keyboard to rootston. When a device is added, rootston creates a wlr-keyboard from the device and configures it based on compositor policy.

Ownership doesn't speak to who created it (wlr seat isn't creating it now, either), but rather who *manages* it.

acrisci: > Slow down. The user should not be creating wlr keyboards, the backend should still be doing that. The approach here is wrong.

Actually this is exactly what I had in mind. The wlr_keyboard is a wrapper around a device made by the user.

SirCmpwn: I see. -1 to that design.

acrisci: Ok everybody stop working on this and let me do a more specific proposal.

SirCmpwn: I don't think this is really that complicated. Your design was mostly on the money.

acrisci:
Thanks so much for working on this, but it turned out to be a pretty simple fix so I did it in #324.

Issue #55, opened 2017-08-10
Use GL_KHR_debug


ascent12:
Currently we use a macro (GL_CALL) to wrap around each of the OpenGL functions to check glError and log if anything went wrong. It works, but it's used inconsistently (e.g. outside of wlr-render) and is ugly, in my opinion.

With [this OpenGL extension](https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_debug.txt), we can set log handlers that the OpenGL implementation will call when something goes wrong and actually get better quality debug messages than what we have currently.

A disadvantage of this is that we will lose the ability to print line numbers of where the error specifically happens. The extension does have some ways to get some arbitrary data into the messages, so we could probably do as good as file + function.

SirCmpwn:
:-1: 

Line numbers are very important imo. It'd be better to make our use of GL_CALL more consistent.

ascent12:
GL errors (excluding GL_OUT_OF_MEMORY) are basically always application bugs, so these messages are only really needed during 'debug mode' (which is settable at runtime).

I would prefer that we eventually used GL_KHR_no_error for 'release mode' when Mesa finishes it, which will hopefully allow for better performance.
Using these callbacks instead macros allows us to have an easy runtime-switch for error checking that doesn't require a branch and an extra function call for every GL function.

Also, instead of just getting "GL_INVALID_OPERATION" or something, we can actually get a description from the driver for why it's invalid.

I personally think the tradeoff is worth it, but I'm not really pushing this too hard. I've just been looking through the GL and EGL extensions and I thought these sorts of things would be useful.

SirCmpwn:
>Also, instead of just getting "GL_INVALID_OPERATION" or something, we can actually get a description from the driver for why it's invalid.

That's a much more compelling reason. Could we combine the two approaches in debug mode?

ascent12:
I suppose something like
```c
struct line_loc {
    const char *file;
    int line;
};

#define GL_CALL(func) do { \
    glDebugMessageCallback(_gles_error_callback, \
        &(const struct line_loc){ _strip_path(__FILE__), __LINE__ }); \
    func; \
} while (0)
```
could work, so we're constantly setting the user-data pointer to be the current line.
I imagine simply setting a callback is a cheap operation, and it just won't be used if no error happens or KHR_no_error is used.
It doesn't remove the wrapper over the GL calls, though.

I would personally be more of a fan of
```c
void function() {
    // This would probably behind some macro
[...]

Issue #281, opened 2017-10-17
Full hardware cursors support


emersion:
ATM, software cursors are used when using client cursors. This would also solve blinking cursors when entering a surface.

See #257.

Waiting for #319.

PR #319, opened 2017-10-23
Renderer redesign

ascent12:
This is a complete overhaul and redesign of wlr_renderer, to what I believe is a more sane interface and hopefully allow for more optimisations in the renderer.
I'm posting this now to allow for comments on the general design of it before I get too far.

# Interesting differences:
- We actually use vertices to pick a position, instead of trying to cram everything in a matrix.  
A user wouldn't need to pass a matrix at all if they didn't want to.
- We use EGL images for every texture, even if we wouldn't actually need to. I think this will allow us to have fewer special cases for different types of textures.  
Also, using [EGL_MESA_image_dma_buf_export](https://www.khronos.org/registry/EGL/extensions/MESA/EGL_MESA_image_dma_buf_export.txt), it may be possible to push smaller buffers to other GPUs and allow for true Multi-GPU rendering. I need to investigate this more, though.
- Adding data to a texture and allocating a texture is now the same operation. There are no "invalid" textures.
- It uses [GL_KHR_debug](https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_debug.txt), which I mentioned ages ago in #55.  
The error messages are **significantly** better, but we lose the exact line number of which function called it. You can get the function name by commenting the line disabling GL_DEBUG_TYPE_PUSH_GROUP_KHR, but it spams the crap out of the log.
- Removal of global state.

## Just a few notes before you read the code:
- ~~It's not feature complete~~
- ~~The old wlr_renderer has **not** been removed yet.~~  
To prevent name collisions, I've named the new ones wlr_render and wlr_tex. wlr_renderer needs to be detangled from some of the components before I can properly remove it.
- It does not use some kind of wrapper interface to allow for multiple backend types: it's GLES2 only. I would argue that we don't need multiple implementations, but it's just like this because I can't be bothered writing all of the boilerplate. It can be added back later if it's something we want to keep.
- ~~There is no public header. I'll write a proper one later.~~


SirCmpwn:
>It does not use some kind of wrapper interface to allow for multiple backend types: it's GLES2 only. I would argue that we don't need multiple implementations, but it's just like this because I can't be bothered writing all of the boilerplate. It can be added back later if it's something we want to keep.

I think we will want to come back around to this. GLES3 may offer better performance since we can use things like VAOs/VBOs, which I'm convinced don't actually work on my older laptop via those GL extensions. There has also been talk of Vulkan, DirectX, and headless renderers, though of course abstracting these APIs further to support those would be a lot of work. I think we should keep the interface abstract.

acrisci:
> Adding data to a texture and allocating a texture is now the same operation. There are no "invalid" textures.

Right now we abuse `surface->texture->valid` for determining whether or not the texture has a buffer which is required for many of the high level Wayland interfaces. A surface might not have a buffer if it either never committed one, committed a null buffer, or something went wrong with uploading the buffer to the texture. Should we keep track of this buffer state in the surface implementation or in wlr_renderer?

Also, I don't know if this is the responsibility of the DRM backend or the renderer, but we did have great difficulty implementing client cursors and they still don't work correctly. I think the main problem was switching between software rendering and hardware cursors. So maybe we can address that at some point (#281).

Another optimization we are talking about is to not ever wake the renderer at all if nothing has changed on the screen between frames.

ascent12:
>Right now we abuse surface->texture->valid for determining whether or not the texture has a buffer which is required for many of the high level Wayland interfaces

If a texture is not longer valid, destroy it and set it to NULL, then that can be the test. I haven't really looked into how 'valid' was being used, but I figured that would work.

>Another optimization we are talking about is to not ever wake the renderer at all if nothing has changed on the screen between frames.

I plan on eventually getting to that. `wlr_render_subtexture` was for that purpose.

ascent12:
Excluding the examples/rootston, the last components still using the old renderer are wlr_surface and wlr_compositor (but that's just passing it straight to wlr_surface).

I would like to remove the rendering code completely from them, because it's not actually vital to their operation. This issue is how we get the buffers passed from clients to the user.

I'm not very familiar with wlr_subsurface or the wl_surface protocol, so I'll probably need to spend some time to make sure I understand it before I go drastically changing shit.  
If anyone has any opinions on how this should be changed, I would like to hear it.

emersion:
>but we did have great difficulty implementing client cursors and they still don't work correctly.

What we want is to always use hardware cursors when supported. For that, two solutions:
1. Figure out a way to get pixels from an uploaded texture. That may be done by: not releasing the buffer until another buffer is attached, or read pixels from the uploaded texture. I really don't know what are the implications of not releasing a buffer (this would be done for _all_ surfaces, since we cannot tell if a surface will be used for a cursor or for something else).
2. Figure out a way to set hardware cursors directly from a texture. Not sure how this could work. I thought that would be easy, but I tried it and failed.

Does this new design help in any way solving this issue?

backend/drm/drm.c
```diff
@@ -586,24 +573,22 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 
 	wlr_drm_surface_make_current(&plane->surf);
 
-	wlr_texture_upload_pixels(plane->wlr_tex, WL_SHM_FORMAT_ARGB8888,
+	struct wlr_tex *tex = wlr_tex_from_pixels(rend, WL_SHM_FORMAT_ARGB8888,
 		stride, width, height, buf);
```
emersion: It looks pretty easy to pass a `wlr_tex` to `wlr_drm_connector_set_cursor` for the DRM backend. Do you think that would still work when multi-GPU is implemented? In other words: can this `rend` render _any_ `wlr_tex`, no matter where it comes from?

examples/screenshot.c
```diff
@@ -181,8 +171,6 @@ static void write_image(const char *filename, int width, int height) {
 		free(output);
 	}
 
-	argb_to_rgba(data, height, buffer_stride);
```
emersion: What you're doing here is against the spec. The `wl_buffer` is created with the format `WL_SHM_FORMAT_XRGB8888`, so we shouldn't put data in another format in there. That also breaks compatibility with other screenshooters (this protocol is shared with orbital).

types/wlr_output.c
```diff
@@ -377,25 +374,16 @@ void wlr_output_make_current(struct wlr_output *output) {
...
-		glEnable(GL_BLEND);
-		glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-
-		struct wlr_texture *texture = output->cursor.texture;
-		struct wlr_renderer *renderer = output->cursor.renderer;
-		if (output->cursor.surface) {
-			texture = output->cursor.surface->texture;
-			renderer = output->cursor.surface->renderer;
-		}
+		struct wlr_tex *tex = output->cursor.tex;
```
emersion: This texture can be NULL or invalid, if the cursor surface has been destroyed or an error has occurred.

types/wlr_screenshooter.c
```diff
@@ -6,43 +6,30 @@
...
-		state->pixels);
 
-	void *data = wl_shm_buffer_get_data(state->shm_buffer);
-	wl_shm_buffer_begin_access(state->shm_buffer);
-	copy_yflip(data, state->pixels + state->stride * (state->height - 1),
-		state->height, state->stride);
+	wl_shm_buffer_begin_access(shm);
+	wlr_render_read_pixels(rend, wl_shm_buffer_get_format(shm), wl_shm_buffer_get_stride(shm),
+		wl_shm_buffer_get_width(shm), wl_shm_buffer_get_height(shm), 0, 0, 0, 0,
+		wl_shm_buffer_get_data(shm));
```
emersion: +1 for this

examples/screenshot.c
ascent12: orbital-screenshooter is what I tested this with, and it came out fine.
I came to the conclusion that *we* were the ones doing something wrong.
test

emersion: We must figure out why. The screenshooter is requesting data in the format `WL_SHM_FORMAT_XRGB8888`, and then using ImageMagick with `bgra`. There's something wrong somewhere.

emersion: Indeed, it seems that they're not respecting the spec.

* They're shuffling pixel data around: https://github.com/giucam/orbital/blob/master/src/screenshooter/main.cpp#L308
* Solid black is `0x000000FF`, not `0xFF000000`: https://github.com/giucam/orbital/blob/master/src/screenshooter/main.cpp#L280

ascent12: It'll have to do with some endianness-related crap.
I think this comment from weston provides some context: https://github.com/wayland-project/weston/blob/903721a6215f474787b5daf02761fbcb1d3a0bb5/libweston/pixel-formats.h#L56

emersion: I don't think DRM can be used as an excuse to store `WL_SHM_FORMAT_ARGB8888` in little-endian. We should respect the format of the `wl_buffer` (that is, ARGB in sequential order) regardless of what the backend is, just like clients don't store little-endian data in buffers when they want to render something.

ascent12: The wayland formats are compatible with both the DRM and GBM formats.
WL_SHM_FORMAT_ARGB8888 is defined to be little endian as part of the spec. Go read <wayland-server-protocol.h>

emersion: My bad, you're right.

render/render.c
```diff
@@ -0,0 +1,378 @@
...
+		return false;
+	}
+
+	DEBUG_PUSH;
+
+	// Make sure any pending drawing is finished before we try to read it
+	glFinish();
+
+	// Unfortunately GLES2 doesn't support GL_PACK_*, so we have to read
+	// the lines out row by row
```
emersion: Is there a way to speed this up in the special case where `stride == 4*width`?

ascent12: We're actually reading though it backwards as well, so probably not.

render/render.c
```diff
@@ -0,0 +1,378 @@
...
+#include <wlr/types/wlr_output.h>
+#include <wlr/util/log.h>
+#include "render/render.h"
+#include "render/glapi.h"
+
+static const struct format formats[] = {
+	{
+		.wl_fmt = WL_SHM_FORMAT_ARGB8888,
+		.gl_fmt = GL_BGRA_EXT,
+		.gl_type = GL_UNSIGNED_BYTE,
```
emersion: A little comment "WL_SHM_FORMAT_ARGB8888 is in little-endian, GL_BGRA_EXT is in big-endian" might help newcomers. I was rather confused the first time I read this code, and didn't understand it.

render/render.c
emersion: That's true. If we have issues with performance (e.g. trying to get a video stream with the screenshooter interface) we could make only one call to `glReadPixels` and flip it, that may be faster than `height` calls to `glReadPixels`.

ascent12: If this was supposed to be high performance, I would make it hand dma bufs around, instead of raw pixels.

examples/screenshot.c
SirCmpwn: Do we need to file a bug with orbital?

render/render.c
SirCmpwn: This is probably why we're going to want a GLES3 renderer.

include/backend/drm/drm.h
```diff
@@ -30,8 +31,6 @@ struct wlr_drm_plane {
 	struct wlr_drm_surface mgpu_surf;
 
 	// Only used by cursor
```
SirCmpwn: This comment is now redundant

render/render.c
```diff
@@ -0,0 +1,378 @@
...
+		int32_t tex_x1, int32_t tex_y1, int32_t tex_x2, int32_t tex_y2,
+		int32_t pos_x1, int32_t pos_y1, int32_t pos_x2, int32_t pos_y2, int32_t pos_z) {
+	assert(eglGetCurrentContext() == rend->egl->context);
+	DEBUG_PUSH;
+
+	GLuint prog = rend->shaders.extn;
+
+	GLuint proj_loc = glGetUniformLocation(prog, "proj");
+	GLuint pos_loc = glGetAttribLocation(prog, "pos");
+	GLuint texcoord_loc = glGetAttribLocation(prog, "texcoord");
```
SirCmpwn: We can stash these elsewhere instead of doing it for each texture.

SirCmpwn:
Read through this, overall it looks good. lmk when it's ready for a proper review.

SirCmpwn:
If they do want to pass a matrix - what then? Rootston does rotation, for example.

ascent12:
>If they do want to pass a matrix - what then? Rootston does rotation, for example.

I plan to add a "render with matrix" function, but I haven't got around to it yet.

types/wlr_output.c
```diff
@@ -159,29 +160,26 @@ static bool set_cursor(struct wlr_output *output, const uint8_t *buf,
...
-		}
-	}
-
-	if (!output->cursor.texture) {
-		output->cursor.texture =
-			wlr_render_texture_create(output->cursor.renderer);
-		if (!output->cursor.texture) {
-			return false;
-		}
+	output->cursor.tex = wlr_tex_from_pixels(rend, WL_SHM_FORMAT_ARGB8888,
```
emersion: Need to cleanup the old `output->cursor.tex` here.

acrisci:
Do you need some help with the implementation in the surface? We would need a way to upload/update pixels to a texture.

ascent12:
Originally, I wanted to remove the renderer entirely from wlr_surface, and have it happen in the 'commit' callbacks (in rootston) for the various shell types (wl/xdg/xwayland). I still want to do that, but it's a fair bit of work, and this PR is already pretty big, so it can be done separately.

I've already replaced the rendering bits in wlr_surface with the new ones. I was going to try and fix some shit related to cursors before I pushed it here, but it seems there has already been work towards that elsewhere.

rootston/output.c
```diff
@@ -19,61 +18,67 @@ static inline int64_t timespec_to_msec(const struct timespec *a) {
...
-			wlr_matrix_mul(&transform, &translate_center, &transform);
-			wlr_surface_get_matrix(surface, &matrix,
-					&wlr_output->transform_matrix, &transform);
-			wlr_render_with_matrix(desktop->server->renderer,
-					surface->texture, &matrix);
+	if (wlr_output_layout_intersects(desktop->layout, wlr_output,
+			lx, ly, lx + width, ly + height)) {
+		wlr_render_texture(desktop->server->render, surface->tex,
+			ox, oy, ox + width, oy + height, 0);
+#if 0
```
emersion: Is this a TODO?

ascent12:
Ok, I feel this is ready for review.

There are a still a few things that can be done, but they're not core to this PR and can be done later:
- Remove wlr_render from wlr_surface. They are causing wlr_render to be a non-optional type.
- Add [buffer age](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_buffer_age.txt) to EGL, which is necessary for writing a much more efficient renderer in rootston
- Add the "shim" layer that could allow for different renderer types. Maybe this can be added when/if someone actually tries to add another renderer. I've tried to make it so it would be easy to add it with the current interface without requiring a bunch of breaking changes.
- A `wlr_tex_export_dmabuf` function, which may be useful for future multi-GPU work.

Also, I currently haven't added a 'render with matrix' function yet (which breaks rootston rotation). I can't decide what the exact semantics of such a function would be, i.e. at what stage is the user matrix applied.  
- output coordinates, output top left origin
- output coordinates, texture top left origin
- gl coordinates, output top left origin
- gl coordinates, texture middle origin
- something else?

### Some other interesting things I added:
- [EGL_KHR_debug](https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_debug.txt)  
Just like the OpenGL one: better error messages, at the cost of line numbers.
- `wlr_backend_get_render`  
This is the place where a user will actually get their renderer, mandating that only one may exist per backend.

rootston/output.c
ascent12: Yes.

ascent12:
# Design Rationale
There seem to be some disagreements/misunderstandings about the nature of the renderer, so I'm going to lay out why this is designed like this.

## It's intentionally very basic and focused
This was only designed for pretty basic 2D compositing, about to the requirements of what Sway or something like Openbox would need. You're given a surface, and you can draw some textures to it, and read them back. There is not much more than that.
I would consider it comparable to SDL_Renderer.

We're writing a wayland-compositor library, not a fully-fledged rendering library.

I don't want to add a huge amount of bloat to the API for something that is almost certainly going to half-assed, and something that most people aren't even going to use.
If you have some requirements even slightly beyond 2D compositing, roll your own renderer. You certainly know what you want a lot more then we do.
This is also why I want wlr_render to be a strictly optional type.

I acknowledge that it's a lot more work for them to write their own renderer, but that's the price to pay
for customisability.

## It's not a 3D renderer
Designing a proper 3D render that can handle off of the fancy shit that some compositor would want to do (wobbly windows, desktop cubes, etc.) is certainly no easy feat. If you're not going to have it be stupidly limiting, you're going to end up design basically an entire graphics interface, made especially hard if you don't want to leak the backend-specific details to the user.

Simply allowing for 4x4 matrices (like the old design) does not make it a good 3D renderer. You **could** do 3D rendering with it, but it'll be extremely unergonomic, and you're still missing most of the things that people want (e.g. rendering to textures, custom shaders). If you try to implement these sorts of extra features, you're just going to end up with a really crappy wrapper over OpenGL.

Also, as before: we're designing a wayland-compositor library, not a rendering library.

## User-defined matrices are not supported

I was originally going to add a function for this, but I honestly cannot think of a good use-case.
Why on earth would you need to rotate/shear/do weird shit in your 2D compositor?
They may make for flashy demos in things like weston, but they're really just a complete gimmick.
We don't need them for scaling, because that's already support that in in other ways.

This goes back to the design goal: this is a **simple** 2D compositor. We don't really need this.

I don't feel that strongly about this, but I still want to keep the API as small as possible.

# TL;DR
- wlr_render is basic
- Writing a fully-featured renderer is too much work for very little gain
- Write your own renderer if you want to fancy shit

SirCmpwn:
I gave this some more thought. I'm okay with everything here *except* for the removal of matricies. You can change them to 2D matricies if you want but it's easy for us to add and adds a lot of potential features. Things like rotation *are* useful - imagine you're doing some drawing in Krita, and maybe you want to rotate the window a bit?

render/render.c
```diff
@@ -194,6 +197,22 @@ void wlr_render_subtexture(struct wlr_render *rend, struct wlr_tex *tex,
...
+		int32_t tex_x1, int32_t tex_y1, int32_t tex_x2, int32_t tex_y2,
+		int32_t pos_x1, int32_t pos_y1, int32_t pos_x2, int32_t pos_y2, int32_t pos_z) {
+	assert(eglGetCurrentContext() == rend->egl->context);
+	DEBUG_PUSH;
+
+	float verts[] = {
+		pos_x1, pos_y1, pos_z,
+		pos_x2, pos_y1, pos_z,
+		pos_x1, pos_y2, pos_z,
+		pos_x2, pos_y2, pos_z,
```
SirCmpwn: What's pos_z for? Both here and elsewhere.

SirCmpwn: We could change the vertex layout to be two dimensional if this isn't important.

rootston/output.c
```diff
@@ -37,76 +52,56 @@ static void rotate_child_position(double *sx, double *sy, double sw, double sh,
...
+		double c = cos(rotation);
+		double s = sin(rotation);
+
+		// Odd transforms are rotations
+		double r = wlr_output->transform % 2 == 0 ? 1 : -1;
+
+		float mat[9] = {
+			+r * (sw * c)/ow, +r * (sh * -s)/ow, r * ((+2.0f * ox + sw) / ow - 1.0f),
+			-r * (sw * s)/oh, -r * (sh * +c)/oh, r * ((-2.0f * oy - sh) / oh + 1.0f),
+			0.0f, 0.0f, 1.0f,
```
SirCmpwn: I preferred this done more explicitly. This isn't really maintainable if we want to change how these matricies are calculated in the future. If it's a performance concern, we can stow the matricies in the view's structure and only recalculate them when the inputs are invalidated.

SirCmpwn:
Thanks for the follow-up! This approach looks great. I only have some minor concerns; if you have time to address them I'd appreciate it very much.

render/render.c
ascent12: I had an idea of using a z dimension to handle laying 2D surfaces on top of each other (so you don't have to draw them in order), but nothing actually uses it yet, and I'm not even 100% sure if it'll work like I expect it to.

SirCmpwn: Hmm, interesting idea, but I'm not sure how well it'd work over just rendering the views in order. It wouldn't work with semitransparent views, right? Those would have to be rendered in a separate pass anyway afaik.

ascent12: Semi-transparent surfaces were actually what I had in mind.
Draw the transparent surface at z == 2 **then** draw an opaque surface at z == 1, and then I think the GPU will just handle blending them together properly.
This type of "optimisation" basically kills the chances for this to be implemented with a software renderer, though.

rootston/output.c
```diff
@@ -68,22 +81,41 @@ static void render_surface(struct wlr_surface *surface,
...
+			sin(rotation), +cos(rotation), 0.0f,
+			0.0f, 0.0f, 1.0f,
+		};
+
+		// Scale matrix to window size
+		float scale[9] = {
+			sw,   0.0f, 0.0f,
+			0.0f, sh,   0.0f,
+			0.0f, 0.0f, 1.0f,
+		};
```
SirCmpwn: I liked the conveinence functions for creating rotation/scale/translation matricies, too. Was there a particular reason you left them out?

render/render.c
ascent12: I don't think gles2 actually has orthographic projection, so I don't think this trick will work, so I removed it.

SirCmpwn: Projection is done by the fragment shader, isn't it? gles2 doesn't have any functions for generating or manipulating matricies, it just passes them along to to the shader.

rootston/output.c
ascent12: Basically, they were unneeded. I'm deliberately trying to keep the API small.
It was also kind of an inefficient API.

SirCmpwn: I think they were pretty small to begin with, and pretty convenient. It's a lot better design-wise to be able to explicitly say `get_rotation_matrix(float radians)` or something to that effect, as opposed to having to remember how to construct a rotation matrix (I'd have to look it up).

So far as effeciency is concerned, wouldn't the approach of only recalculating the matricies when views are invalidated address that issue?

ascent12: It's really something that I don't think adds a lot of value. It's something which isn't core to wlroots AT ALL (as in, it makes no use of any library internals), and you can find dozens of more efficient single header C "libraries" which can do matrix manipulation extremely easily.

Maybe the functions could live in rootston, but I really don't think we should be bloating the wlroots API with that kind of stuff.

SirCmpwn: I am very opposed to the idea of "single header C 'libraries'". I think they're an awful idea and they subvert a lot of good C programming principles. They might be "more effecient", but our functions are *well* within our performance budget. Even on the highest end of displays our bottleneck is the display's framerate. I'm sure we could compute our matricies tens of thousands of times between frames.

When it comes to API bloat, I don't think matricies really count. The functions are small in number and size. Matrix math is also not a controversial API - it's simple mathematics. We aren't making any design decisions on the user's behalf by providing such an API. We already have precedent for this sort of thing, too - the convenience functions we offer for wlr_box are something we could easily avoid exporting, but they're simple to implement and something many compositors will need.

ascent12: Having some util code that we use internally (like wlr_box) which we export is different. It's going to be there anyway, so you may as well make it part of the API.

However, matricies are not used anywhere internally at all, and I don't think it will ever be needed (obviously excluding the output rotations internal to just the renderer).
I consider this kind of code to be completely superfluous. It basically ties into the same reason I hate wlr_list so much. Sure, lists are useful and all, but I don't think it's our job to be providing them over our API.
If we start providing excess util code, where does it end? Should we provide a hashmap too? Those are pretty useful.

I don't even consider matricies to be common use case. They only there because having support for window rotation was insisted, but it's really just for gimmicky demo features.  I think most realistic uses of wlroots would't even need this.

SirCmpwn: Many of the things we provide are for convenience. If we didn't do that, we'd just be left with unopinionated state machines that implement Wayland interfaces. Tools like wlr_output_layout, wlr_cursor, and wlr_xcursor would not exist. You could even make an argument that our backends wouldn't exist, too. We provide tools that we think compositors will need, but don't couple them so tightly that they can't go without. Matricies are one such tool.

Rotation might seem like a gimmick to you, but only on compositors like Sway. Remember that not everyone prefers a minimalistic experience. Many other kinds of compositors exist, and it's our job to design APIs that suit the broadest audience of compositors. Imagine all of the more novel matrix transformations that would be applied on a less gimmicky compositor that uses animations? Do you use android? Open the app switcher and pull it down beyond the maximum extent. Note the shearing and use of perspective here - is this a gimmick? Mobile compositors are something wlroots is intended to support in addition to desktop compositors. Check out some of GNOME or KDE's animations, too - not all of them can be accomplished by simply mapping the surface to a rectangle on-screen.

I don't like wlr_list either, but it's there because it's the most suited data structure to some of our API and we need to expose its functions so those parts of our API can be interacted with effectively. We could probably make a separate issue about making most of its functions internal, though.

ascent12: To go back to what I said earlier
>This was only designed for pretty basic 2D compositing, about to the requirements of what Sway or something like Openbox would need

I'm honestly not interested in supporting anything beyond that with wlr_render, for the reasons I described in one of my previous posts. "render_texture_with_matrix" just exists to subvert the interface.

>Matrix math is also not a controversial API - it's simple mathematics

That's part of the reason I'm against it. Since it's so simple, why even bother having it?
The user can just throw together 30 lines of code with their own matrix stuff.
There aren't 500 lines of code which interacts with other systems and would be considered essential for traditional window managers.

minus7: I'd add neither, the matrix interface to rendering nor the helper functions, because I do not see a practical use case. But if you're adding a matrix interface, helper functions to build matrices in the right format should be there as well.

If you were to use them for animations, wouldn't a simple matrix-based transformation be too simple to get anything worthwhile anyway?

ascent12: >If you were to use them for animations, wouldn't a simple matrix-based transformation be too simple to get anything worthwhile anyway?

Yes, that's basically my original point from a couple of weeks ago. This renderer is (intentionally) far too limited to do any of the special use-cases being discussed.

SirCmpwn: Alright, whatever. You can revert the matrix addition and remove the rotation code from rootston.

ascent12: It's already there. It shares like 90% of the code with the "normal" interface, so it's not adding a huge amount of code.
I don't really think it needs to be there, but it's not something that I feel **that** strongly about.

SirCmpwn: Either you want matricies or you don't commit to one and fucking do it right.

ascent12:
I've relented and re-added the matrix functions. I've changed the interface to be what I think is more sane.

However, this isn't quite ready for merging yet. I tried testing this on an intel iGPU laptop, and it's giving me problems.
Apparently the radeon drivers are more lenient than intel when it comes to glEGLImageTargetTexture2DOES. It works perfectly find on my development computer, so I must have been making some non-portable assumptions.
It's been a bit harder than I anticipated to fix it.

I'll also fix these new merge conflicts and maybe clean up the history a little bit.

ascent12:
Ok, I think this is ready for final review. I ironed out a few bugs related to the intel drivers being annoyingly picky and ruining what I hoped would be a cleaner design.
I was originally going to rebase this, try get rid of some of the pointless merge commits, and squash some related things together. After a couple of minutes, I gave up: there are far too many non-trivial conflicts for it to be worth it.

There are a few parts of it I haven't really tested extensively, particularly things which landed in wlroots in the last couple of weeks and required me to do some kind of conflict resolution. I tried to give things a quick go, but there are probably a bunch of edge cases I missed.

include/wlr/types/wlr_surface.h
```diff
@@ -86,9 +87,10 @@ struct wlr_surface {
 	void *data;
 };
 
-struct wlr_renderer;
 struct wlr_surface *wlr_surface_create(struct wl_resource *res,
-		struct wlr_renderer *renderer);
+	struct wlr_render *rend);
+
+#if 0
```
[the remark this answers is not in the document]
SirCmpwn: Yes

acrisci:
@ascent12 can I continue this work? What still needs to be done?

PR #320, opened 2017-10-23
abstract buffer exists

acrisci:
Add `wlr_surface_has_buffer` and use it instead of `surface->texture->valid` in dependent interfaces. See #319 .


SirCmpwn:
:+1: good call

PR #322, opened 2017-10-23
Fixes #321, not using result of wl_container_of

### LINKED ISSUE #321: Examples not working in DRM
The examples (all non-rootson ones) are no longer working on my machine (tested on master). I did a bisect, and found the culprit is one of these commits:

```
1d716241afbc721c3ae65d4a2bb0e25866fe081c
c03e7746367194b859125624b059c1725eb7ae30
e1f196a3e9477385ca30180686cda82e3ab735ce
```

It's hard to say because they don't actually build.

The one that _doesn't_ work is `e1f196a` and the last one that _does_ is `bde25fe`

Here's a diff from `e1f196a` to `bde25fe`: https://pastebin.com/jfTyrvyd

EDIT: Sorry hit "enter" before it completed. Stupid vim plugin

EDIT X2: From the logs looks like this was when `wl_list` was used instead in the DRM stuff. Probably an issue somewhere where it's mis-used. Probably shouldn't be hard to find.

Timidger:
Did a grep through the codebase and this is the only place where we don't use the result of `wl_container_of`, so this should be the only place where we misused it.

### DIFF
--- a/examples/shared.c
+++ b/examples/shared.c
@@ -414,8 +414,8 @@ static void output_add_notify(struct wl_listener *listener, void *data) {
 	wlr_log(L_DEBUG, "%s %s %"PRId32"mm x %"PRId32"mm", output->make, output->model,
 		output->phys_width, output->phys_height);
 	if (wl_list_length(&output->modes) > 0) {
-		struct wlr_output_mode *mode = NULL;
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

PR #324, opened 2017-10-23
wlr-seat: keyboard redesign

### LINKED ISSUE #242: Don't send keys to clients when they match a keybinding in rootston
This will be annoying but we'll have to figure it out anyway for sway.

### LINKED ISSUE #287: Redesign keyboard interface

acrisci:
fixes #287 
fixes #242

I haven't tested it with multiple keyboards because I only have one right now, but that should work correctly.

Tests:

* run a binding and see that the keyboard key event does not go to the window
* unplug the keyboard and plug it back in and see that it works
* plug in two keyboards and alternate typing in a window with them and notice that works
* the above but each keyboard has a different keymap
* press a modifier (shift) on one keyboard and then type on the other and notice the modifier is not considered pressed while you type on the other.

### DIFF
--- a/rootston/keyboard.c
+++ b/rootston/keyboard.c
@@ -53,7 +53,13 @@ static void keyboard_binding_execute(struct roots_keyboard *keyboard,
 	}
 }
 
-static void keyboard_keysym_press(struct roots_keyboard *keyboard,
+/**
+ * Process a keypress from the keyboard.
+ *
+ * Returns true if the keysym was handled by a binding and false if the event
+ * should be propagated to clients.
+ */
+static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
 		xkb_keysym_t keysym) {
 	ssize_t i = keyboard_pressed_keysym_index(keyboard, keysym);
 	if (i < 0) {
@@ -74,7 +80,7 @@ static void keyboard_keysym_press(struct roots_keyboard *keyboard,
 				wlr_session_change_vt(session, vt);
 			}
 		}
-		return;
+		return true;
 	}
 
 	uint32_t modifiers = wlr_keyboard_get_modifiers(keyboard->device->keyboard);
@@ -96,8 +102,11 @@ static void keyboard_keysym_press(struct roots_keyboard *keyboard,
 
 		if (ok) {
 			keyboard_binding_execute(keyboard, bc->command);
+			return true;
 		}
 	}
+
+	return false;
 }
 
 static void keyboard_keysym_release(struct roots_keyboard *keyboard,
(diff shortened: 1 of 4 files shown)


versusvoid:
Works with two keyboards (laptop's built-in and normal).

rootston/keyboard.c
```diff
@@ -116,13 +125,36 @@ static void keyboard_key_notify(struct wl_listener *listener, void *data) {
...
-			keyboard_keysym_press(keyboard, syms[i]);
+			bool keysym_handled = keyboard_keysym_press(keyboard, syms[i]);
+			handled = handled || keysym_handled;
 		} else { // WLR_KEY_RELEASED
 			keyboard_keysym_release(keyboard, syms[i]);
 		}
 	}
+
+	if (!handled) {
+		wlr_seat_set_keyboard(keyboard->input->wl_seat, keyboard->device);
```
acrisci: We could get fancy here and have the seat read the keyboard event before rootston and automatically set the keyboard on the seat to the one with the last event then lazily send the keymap on the next seat key notify.

emersion: Yeah, that would be better, the compositor should always do it anyway.

SirCmpwn:
Is this good to go?

acrisci:
I think it should be tested with the new keyboard config which I can do tomorrow.

SirCmpwn:
Tested this on the new keyboard config with two different keyboard layouts, worked as expected.

PR #325, opened 2017-10-23
Revert "wlr-surface: destroy texture on null buffer commit"

acrisci:
This reverts commit 0e7d13fab7f23658a85df58ad26c6c77c9638bff.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -406,8 +406,7 @@ static void wlr_surface_commit_pending(struct wlr_surface *surface) {
 	wlr_surface_move_state(surface, surface->pending, surface->current);
 
 	if (null_buffer_commit) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Context?

acrisci:
I committed some code that causes some segfaults. The reason for this is change was that @ascent12 told me to destroy the texture when no longer being used, but I think that caused some bugs in my surface code and possibly the texture code which assumed it was never destroyed.

I don't know if it's worth it to dig into the texture stuff right now since it is changing a lot in #319, so I'm afraid working on it will cause more problems with his effort than it will solve. It's probably better to wait until he has the interface laid out a bit more stable and let him tell me what the new way of finding out how to handle invalidating a texture and detecting a valid texture.

PR #328, opened 2017-10-23
Send output current mode when changed

### LINKED ISSUE #229: Send updated modes when necessary
Ref #228

emersion:
Fixes #229

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -33,8 +33,8 @@ static void wl_output_send_to_resource(struct wl_resource *resource) {
 			if (output->current_mode == mode) {
 				flags |= WL_OUTPUT_MODE_CURRENT;
 			}
-			wl_output_send_mode(resource, flags,
-				mode->width, mode->height, mode->refresh);
+			wl_output_send_mode(resource, flags, mode->width, mode->height,
+				mode->refresh);
 		}
 
 		if (wl_list_length(&output->modes) == 0) {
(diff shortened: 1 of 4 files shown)


types/wlr_output.c
```diff
@@ -51,6 +51,25 @@ static void wl_output_send_to_resource(struct wl_resource *resource) {
...
+	struct wlr_output *output = wl_resource_get_user_data(resource);
+	assert(output);
+	const uint32_t version = wl_resource_get_version(resource);
+	if (version < WL_OUTPUT_MODE_SINCE_VERSION) {
+		return;
+	}
+	if (output->current_mode != NULL) {
+		struct wlr_output_mode *mode = output->current_mode;
+		wl_output_send_mode(resource, mode->flags | WL_OUTPUT_MODE_CURRENT,
+			mode->width, mode->height, mode->refresh);
```
SirCmpwn: Do you have to destroy the old mode we sent? Maybe there's a way to modify it?

emersion: I think `output.current_mode` is just a pointer to an element of `output->modes`, so it doesn't matter.

emersion: (Ah sorry, misunderstood, was thinking about memory, not protocol)

https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output-event-mode

>The event is sent when binding to the output object and there will always be one mode, the current mode. The event is sent again if an output changes mode, for the mode that is now current. In other words, the current mode is always the last mode that was received with the current flag set.

SirCmpwn:
So should we not be sending the full list of modes in the first place?

emersion:
We already do that in [`wl_output_send_to_resource`](https://github.com/swaywm/wlroots/blob/master/types/wlr_output.c#L28).

SirCmpwn:
I mean, is that wrong?

emersion:
Oh, not at all: when a client first connects, we send all modes, and if the current mode changes, we only resend that mode. Only sending the current mode when a client connects would make the `WL_OUTPUT_MODE_CURRENT` flag useless. As the spec says:

>The mode event describes an available mode for the output.

So if a mode is _available_, we should send it.

SirCmpwn:
Aight

PR #329, opened 2017-10-24
HiDPI

SirCmpwn:
* [x] Accept output scale in the config
* [x] Send output enter/leave events
* [x] Render surfaces scaled up by their scale factor
* [x] Consider the new surface size accurately with respect to pointer events
* [x] Scale client cursors
* [ ] Scale subsurfaces
* [x] Correctly handle surfaces whose scale factor does not match the output
* [ ] Scale rootston cursor
* [ ] Report bug to GTK: moving window from scale=1 -> scale=2 causes incorrect cursor scale
* [ ] Input for rotated surfaces
* [x] Increase input fidelity


include/wlr/types/wlr_surface.h
```diff
@@ -135,4 +135,8 @@ struct wlr_surface *wlr_surface_get_main_surface(struct wlr_surface *surface);
  */
 struct wlr_subsurface *wlr_surface_subsurface_at(struct wlr_surface *surface,
 		double sx, double sy, double *sub_x, double *sub_y);
+
+void wlr_surface_send_enter(struct wlr_surface *surface,
+		struct wlr_output *output);
```
emersion: We should send `leave` too.

rootston/desktop.c
```diff
@@ -46,14 +46,35 @@ void view_get_size(struct roots_view *view, struct wlr_box *box) {
...
+			continue;
+		}
+		if (output == NULL
+				|| output->wlr_output->scale < _output->wlr_output->scale) {
+			output = _output;
+		}
+	}
+	if (output && output != view->output) {
+		view->output = output;
+		wlr_surface_send_enter(view->wlr_surface, output->wlr_output);
```
emersion: Shouldn't we send one `enter` event for each output which intersects with the view?

types/wlr_output.c
```diff
@@ -28,8 +28,7 @@ static void wl_output_send_to_resource(struct wl_resource *resource) {
 	if (version >= WL_OUTPUT_MODE_SINCE_VERSION) {
 		struct wlr_output_mode *mode;
 		wl_list_for_each(mode, &output->modes, link) {
-			// TODO: mode->flags should just be preferred
-			uint32_t flags = mode->flags;
+			uint32_t flags = mode->flags & ~WL_OUTPUT_MODE_PREFERRED;
```
emersion: Does this flag break things?

SirCmpwn: I did this backwards.

emersion: Yeah, so:
* `WL_OUTPUT_MODE_PREFERRED` has only one bit on
* `~WL_OUTPUT_MODE_PREFERRED` has only one bit off, the preferred bit (all others on)
* `mode->flags & ~WL_OUTPUT_MODE_PREFERRED` is `mode->flags` with the preferred bit off

But the comment says the contrary: `mode->flags` should only be able to change `WL_OUTPUT_MODE_PREFERRED` (ie. should not be able to set `WL_OUTPUT_MODE_CURRENT`).

SirCmpwn: I already said I did it backwards -_-

rootston/desktop.c
emersion: Spec: https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_surface-event-enter

>Note that a surface may be overlapping with zero or more outputs.

So yeah, we should send `enter` to all outputs that intersects with the view.

types/wlr_output.c
emersion: Oh, sorry. I though you were speaking about the `~`, saying it inverts bits so it would be fine.

SirCmpwn:
I'm done for today, but if anyone is reviewing my commits and sees the reason behind this bug, I would appreciate your insight:

https://sr.ht/_HSY.png

gnome-calculator should not show up at all on the right output - but it does.

SirCmpwn:
I think I figured out an approach that will address that problem, writing it down here for poserity -

We need to treat HiDPI displays in the *output layout* as scaled up - that is, my 4K display should be arranged as if it were 1920x1080, and its x/y position relative to this. Then, at the renderer, we can scale it up or down depending on the scale of the surface and the genuine size of the output.

include/rootston/view.h
```diff
@@ -67,14 +67,14 @@ struct roots_view {
...
-	void (*get_size)(struct roots_view *view, struct wlr_box *box);
+	void (*get_size)(const struct roots_view *view, struct wlr_box *box);
 	void (*activate)(struct roots_view *view, bool active);
 	void (*resize)(struct roots_view *view, uint32_t width, uint32_t height);
 	void (*set_position)(struct roots_view *view, double x, double y);
 	void (*close)(struct roots_view *view);
 };
 
-void view_get_size(struct roots_view *view, struct wlr_box *box);
+void view_get_size(const struct roots_view *view, struct wlr_box *box);
```
emersion: Maybe rename this to `get_box` now that this also returns the position?

SirCmpwn:
Woo-hoo, got it working

rootston/desktop.c
```diff
@@ -97,20 +126,24 @@ bool view_center(struct roots_view *view) {
...
 	return true;
 }
 
 void view_setup(struct roots_view *view) {
-	view_center(view);
-
 	struct roots_input *input = view->desktop->server->input;
+	view_center(view);
 	set_view_focus(input, view->desktop, view);
+	wlr_seat_keyboard_notify_enter(input->wl_seat, view->wlr_surface);
```
emersion: This is already done in `set_view_focus`: https://github.com/swaywm/wlroots/blob/975b9dc365d5a7bec531522320a1506323575525/rootston/cursor.c#L209

types/wlr_output.c
```diff
@@ -450,6 +451,8 @@ void wlr_output_cursor_set_surface(struct wlr_output_cursor *cursor,
 }
 
 bool wlr_output_cursor_move(struct wlr_output_cursor *cursor, int x, int y) {
+	x *= cursor->output->scale;
+	y *= cursor->output->scale;
```
emersion: Because of this, maybe we want to accept `double`s, multiply them, and cast them to `int`s right away.

emersion:
We could add a config field for XWayland DPI:

```
Xwayland
-dpi int               screen resolution in dots per inch
```

SirCmpwn:
Are you sure that's the right bugzilla link?

types/wlr_output.c
SirCmpwn: Yes, this is in my todo list

emersion:
Whoops, here's the link: https://bugs.freedesktop.org/show_bug.cgi?id=93315

An interesting but complicated idea if we want to really support HiDPI in XWayland is this comment:

>Could we run two xwayland instances? first display is lodpi, and second is hidpi?

SirCmpwn:
>Could we run two xwayland instances? first display is lodpi, and second is hidpi?

We wouldn't be able to move windows between them. We should run an Xwayland server at the highest DPI among our monitors and scale down views on some displays. But it will be very messy.

SirCmpwn:
I have a proposal for how to deal with cursor scale here. Basically, we're going to need to load xcursor themes at multiple scales for them to look good on each output in a mixed-DPI setup. I want to change wlr_cursor to distinguish the surface cursor from the fallback cursor. We'll change the set pixels guy to `wlr_cursor_set_fallback_image` and stash the data away on the cursor somewhere, and add a scale parameter. wlr_cursor will pick the most appropriate scaled image for the output it's showing on. Then when rootston sets an outputs scale factor, it makes sure that the cursor has the necessary images.

SirCmpwn:
Looking for feedback on this now. I'm going to spin out the remaining work into separate tickets.

types/wlr_output.c
```diff
@@ -231,14 +231,15 @@ void wlr_output_destroy(struct wlr_output *output) {
 
 void wlr_output_effective_resolution(struct wlr_output *output,
 		int *width, int *height) {
```
emersion: Maybe it's worth it to have `double`s here too?

SirCmpwn: Hmm, I'm not sure we get any benefit from that. This isn't a place where loss of precision impacts anything.

emersion: Well, we use it below for the cursor code:

https://github.com/swaywm/wlroots/pull/329/files#diff-8c8783fe608815c9626588d85b00f2f4R273

What we basically do is: `(int)(width / scale) * scale`.

SirCmpwn: Yeah, but there are no displays (or if there are, I've never heard of them) with odd dimensions so you are always going to be dealing with integers here anyway.

emersion: Okay, so we should be fine for now. We won't be anymore when we'll introduce fractional scaling.

emersion:
* New views are no longer centered on screen
* `weston-terminal` is blurry
* Subsurfaces seem to work properly
* (Those two TODOs from the PR description)

SirCmpwn:
>weston-terminal is blurry (same for all weston demos)

I'm not convinced that weston supports hidpi

>New views are no longer centered on screen

Genuine bug

Issue #300, opened 2017-10-20
Firefox URL dropdown is positioned incorrectly


SirCmpwn:
Tested on the wayland backend.

SirCmpwn:
This is just conjecture but I bet this works fine on the DRM backend.

SirCmpwn:
Nope, does not work on DRM either.

acrisci:
There's nothing in the xwayland code that handles the position of transient windows.

emersion:
Also happens with `gnome-calculator`'s combobox.

acrisci:
I'll have a fix for this today.

emersion:
Can we close this one?

PR #331, opened 2017-10-24
xwayland window manager fixes

acrisci:
There are lots of problems with the current x window manager implementation. This is an attempt to get it to work correctly.

List of bugs:

* [x] dragging the window near the border and moving quickly can end the drag early if the surface doesn't keep up with the mouse position (support `_NET_WM_MOVERESIZE`).
* [x] firefox context menus
* [ ] <s>junk around the borders of firefox popups and menus and no support for transparency (firefox chooses wrong visual depth).</s> #348 
* [x] <s>user has the ability to size windows unusably tiny and to negative width/height (probably need to honor some minimum size).</s> #346
* [x] firefox crashes when you double click on the titlebar (probably needs `_NET_WM_MOVERESIZE`).

Expectations:

Support a reasonably ok experience with Firefox and Chromium.


SirCmpwn:
Does this address any specific tickets?

acrisci:
This mainly addresses #300, but I'm just going to fix things as I see them and probably do a big cleanup at the end. This won't actually add any new features, just fix existing bugs.

SirCmpwn:
Alright, cool. I would appreciate a run-down of the bugs this fixes, to make it easier to test.

include/wlr/xwayland.h
```diff
@@ -66,14 +66,23 @@ struct wlr_xwayland_surface_size_hints {
...
 struct wlr_xwayland_surface {
 	xcb_window_t window_id;
+	struct wlr_xwm *xwm;
 	uint32_t surface_id;
-	struct wl_list link;
+
+	struct wl_list displayable_link;
+	// XXX: I think this is just a list of all the surfaces
+	struct wl_list new_link;
+	struct wl_list unpaired_link;
```
martinetd: The way you handle these three lists seems a bit odd, you never remove anything from the new_link or unpaired_link?
The way the original link worked there was only one link for multiple lists on purpose as it is a state machine. I've already forgotten my old code but basically we're waiting for two events to display the window, something from the X11 side and something from the wayland side (Xwayland is a wayland client to us) ; if one comes first the surface is put in new_surfaces, if the other comes first it's put in unpaired_surfaces, and when you have both whichever came first it would go to displayable_surfaces. The surface cannot be in multiple lists at once.

I don't really care if you want the increased complexity of handling three links + associated bools but at the very least never cleaning the other two lists is going to make things slower than it needs to be when there are many windows around...

acrisci: Oh thanks for clearing that up. I just assumed whoever wrote the code didn't know how `wl_list` worked. I will definitely make this much simpler in the course of my rewrite.

acrisci:
Ok, #300 is much better and ready to test but there is still some work to do handling transient state.

Make sure the menu bar popups work in gnome-terminal (start with `gnome-terminal --display=:1` or whatever your xwayland DISPLAY is).

Make sure `_NET_ACTIVE_WINDOW` points to the right window id when it is focused (check with `xprop -root`).

In the code, we now no longer destroy the surface when the x11 window is unmapped. Rootston gets an unmap notify in this case when the xwayland surface is destroyed and it removes the roots surface from the view list. This is because xwayland destroys the wl_surface for the view and the view needs to unset its surface. We will probably need to handle unmapped views in rootston at some point (such as when the view surface has no buffer), but this is good enough for now.

The other things are just clearing out cruft. I redid all the weird list data structures. There are now just two lists: a list of all the xwayland surfaces and a list of xwayland surfaces that have not yet been paired with wl_surfaces.

xwayland/xwm.c
```diff
@@ -540,26 +598,27 @@ static void handle_client_message(struct wlr_xwm *xwm,
 	wlr_log(L_DEBUG, "XCB_CLIENT_MESSAGE (%u)", ev->window);
 
 	if (ev->type == xwm->atoms[WL_SURFACE_ID]) {
-		struct wlr_xwayland_surface *surface = lookup_surface(
-			&xwm->new_surfaces, ev->window);
+		struct wlr_xwayland_surface *surface = lookup_surface(xwm, ev->window);
```
martinetd: Well, my naming sense sucks but the way the code was layout with three lists really had a purpose :)

So now I have the code under my nose, what the old code did was (I want a blackboard, parallelism sucks to explain in text):
 * wlr_xwayland_surface_create is called for the X11 side (XCB_CREATE_NOTIFY event), and creates the surface / puts it in new_surfaces
 * At some point later, X11 sends XCB_CLIENT_MESSAGE with type WL_SURFACE_ID. This is meant to let us map the surface that was created through wayland with that ID and the X11 window with that other ID: the only possible candidates are windows which got created but never got this message yet, hence a list just for that for quick lookup.
At this point either the wayland surface exist and we map it, or it's not there yet and we put the wlr_xwayland_surface in the unpaired list to remember the surface_id and look at it when the wayland surface is created.
 * At some point in parallel (before or after any of the other two bullets), wayland creates the surface. Either it's in the unpaired list (both other events were first) and we can map it, or we're first and the WL_SURFACE_ID message will come later and find this surface through wl_client_get_object so we have nothing to do.

To sum up, during the life of a wlr_xwayland_surface it would either be put in new_surfaces then displayable_surfaces if the wayland event was first;  or new_surfaces then unpaired_surfaces then displayable_surfaces if the xcb event was first.

Now, this change will technically work but this bit bothers me slightly. This lookup here will go through all the X11 windows, when there really only are very few candidate (this list will usually only have one element, so that particular lookup was almost free and is now going through everything). I don't suppose most people will have thousands of X11 clients around so the performance impact is probably fine, but this risks remapping windows that already had been mapped if X11 goes crazy -- I'm hardly convinced this code will work in that case and I do not trust Xwayland (I have no clue if and X11 client could make it send arbitrary messages) so this is a DoS or worse waiting to happen.
I guess the code is simpler that way and it might be worth it, but we need to make sure things work in formerly-unexpected cases.

(I definitely agree with getting away from the displayable_windows list that was public -- I was using that list in the original compositor because there was nothing handling the mapping events, but compositors will be managing this side on their own. Unfortunately we still need a list/hash/whatever with all the windows because xcb talks to us with only windows ID and we need to find which structure corresponds to that)

SirCmpwn:
Firefox context menus don't work with this.

SirCmpwn:
gnome-terminal is fine.

xwayland/xwm.c
acrisci: > but this risks remapping windows that already had been mapped if X11 goes crazy

Can you go through this and add assertions so we crash gracefully and get good bug reports?

emersion:
Dragging `gnome-calculator` is sometimes very laggy and the window stops moving during a drag operation.

emersion:
Update: this happens when dragging a window from near the border. There's some lag so doing it quickly makes the cursor leave the window surface. That's when the window stops moving.

xwayland/xwm.c
```diff
@@ -540,26 +598,27 @@ static void handle_client_message(struct wlr_xwm *xwm,
...
-		surface->surface_id = ev->data.data32[0];
 		/* Check if we got notified after wayland surface create event */
-		struct wl_resource *resource = wl_client_get_object(
-			xwm->xwayland->client, surface->surface_id);
+		uint32_t id = ev->data.data32[0];
+		struct wl_resource *resource =
+			wl_client_get_object(xwm->xwayland->client, id);
 		if (resource) {
-			map_shell_surface(xwm, surface, wl_resource_get_user_data(resource));
+			surface->surface_id = 0;
```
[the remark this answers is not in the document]
acrisci: The assumption is surface_id is not zero if and only if it's in the unpaired_surfaces list. Please check to make sure that always holds.

acrisci:
Added a list of todos on the pr for issues I can reproduce.

SirCmpwn:
osu seems to work for me on this branch btw

acrisci:
What does `xcb_composite_redirect_subwindows()` do for us?

SirCmpwn:
I am encountering an issue with firefox dropdowns that are semi-transaprent: https://sr.ht/kgk7.png

Timidger:
Resizing a window (e.g firefox) to be small enough to be non-existant causes both it to crash and the chrome instance I had running (`weston-terminal` survived, making me think the xwayland server did a hard reset)

Timidger:
Keeping your cursor in an xwayland window and then switching to a separate tty and back causes the rootston cursor to stay in the top left but the xwayland cursor to still function inside the window. Moving outside of the window fixes it

martinetd:
Played a bit as well:
 * probably not only an X11 problem, but setting the cursor right on the edge of a window makes the cursor jump. I actually get different behavior with e.g. urxvt (the cursor jumps and comes back right away) and firefox (the cursor stays a couple of hundreds pixels down until I move it again, at which point it came back)
I think I get small jumps with e.g. weston-shell as well, but it's way smaller, not sure why.
 * double-clicking the title bar in firefox makes it crash, as it's got a crash handler I have no clue how to look at it but this looks relevant: `Gdk:ERROR:gdkwindow-x11.c:5165:create_moveresize_window: assertion failed: (mv_resize->moveresize_emulation_window == NULL)`

Looking quite good tbh.

xwayland/xwm.c
martinetd: No clue how to add commits to an existing PR (I could do a PR to your branch on your repo but what a mess that is to track), I'll wait till this gets merged and have another look at this.

xwayland/xwm.c
```diff
@@ -187,44 +300,6 @@ static void read_surface_parent(struct wlr_xwm *xwm,
 	wl_signal_emit(&surface->events.set_parent, surface);
 }
 
-static void handle_surface_state(struct wlr_xwm *xwm,
```
emersion: Why is this broken?

acrisci: Now that I'm looking at it again, it looks ok. Something was setting a bunch of junk in the window property that was making windows think they are maximized, so I just blamed this code and removed it.

There are several things wrong with this approach in general though.

1. It emits events for state changes we don't care about. Many of these don't translate to wayland well and require a level of knowledge of X11 to interpret I wouldn't expect library users to have. It's best to keep this interface slim or else it will become hard to maintain. This is fine because clients are supposed to read `_NET_SUPPORTED` from the root window and not send us events that aren't in there, but some toolkits (GTK) ignore this property and send us the events anyway.
2. It's inconsistent with the other shell apis and will be a bit cumbersome to abstract into a view properly.

acrisci:
Ok, the Firefox transparency issue is hard so I've put it in #348 and I'll look into it next week.

I think I've fixed all the difficult outstanding issues and have the shell interface worked out, so I think this is ready to merge. If anyone finds more issues, post them here.

xwayland/xwm.c
```diff
@@ -492,84 +614,282 @@ static void handle_configure_request(struct wlr_xwm *xwm,
...
+	case _NET_WM_MOVERESIZE_SIZE_BOTTOMRIGHT:
+	case _NET_WM_MOVERESIZE_SIZE_BOTTOM:
+	case _NET_WM_MOVERESIZE_SIZE_BOTTOMLEFT:
+	case _NET_WM_MOVERESIZE_SIZE_LEFT:
+		resize_event.surface = xsurface;
+		resize_event.edges = map[detail];
+		wl_signal_emit(&xsurface->events.request_resize, &resize_event);
+		break;
+	case _NET_WM_MOVERESIZE_CANCEL:
+		break;
```
emersion: Is this a TODO?

xwayland/xwm.c
```diff
@@ -492,84 +614,282 @@ static void handle_configure_request(struct wlr_xwm *xwm,
...
+	int maximized = xsurface_is_maximized(xsurface);
+
+	uint32_t action = client_message->data.data32[0];
+	uint32_t property = client_message->data.data32[1];
+
+	if (property == xwm->atoms[_NET_WM_STATE_FULLSCREEN] &&
+			update_state(action, &xsurface->fullscreen)) {
+		xsurface_set_net_wm_state(xsurface);
+
+		if (xsurface->fullscreen) {
```
emersion: Should we save the size if the window is not yet fullscreen, rather than already fullscreen?

xwayland/xwm.c
```diff
@@ -492,84 +614,282 @@ static void handle_configure_request(struct wlr_xwm *xwm,
...
+			xsurface_set_net_wm_state(xsurface);
+		}
+
+		if (property == xwm->atoms[_NET_WM_STATE_MAXIMIZED_HORZ] &&
+				update_state(action, &xsurface->maximized_horz)) {
+			xsurface_set_net_wm_state(xsurface);
+		}
+
+		if (maximized != xsurface_is_maximized(xsurface)) {
+			if (xsurface_is_maximized(xsurface)) {
```
emersion: Same here?

xwayland/xwm.c
acrisci: I don't want to handle it without a specific example of something that depends on it to function correctly.

SirCmpwn: Can you add a comment to that effect?

xwayland/xwm.c
acrisci: This is a little confusing. The `update_state()` call returns true if the property changes. Thus when `xsurface->fullscreen` is set here, that means we are transitioning from a non-fullscreen state to a fullscreen state and in this case save the width/height for when it goes back. So I think this is correct how it is.

xwayland/xwm.c
acrisci: I think this is correct as well for the same reasons as above.

rootston/xwayland.c
```diff
@@ -63,6 +61,85 @@ static void handle_request_configure(struct wl_listener *listener, void *data) {
 		xwayland_surface, event->x, event->y, event->width, event->height);
 }
 
+// XXX Needs deep refactoring to get this better. We need to select the correct
+// seat based on seat pointer focus, but interactive moving and resizing is not
+// yet seat aware. Even then, we can only guess because X11 events don't give us
+// enough wayland info to know for sure.
```
SirCmpwn: Make sure this is captured in a ticket

xwayland/xwm.c
```diff
@@ -65,31 +63,154 @@ static struct wlr_xwayland_surface *wlr_xwayland_surface_create(
...
+
+		uint32_t values[1];
+		values[0] = XCB_STACK_MODE_ABOVE;
+		xcb_configure_window_checked(xwm->xcb_conn, surface->window_id,
+			XCB_CONFIG_WINDOW_STACK_MODE, values);
+	} else {
+		xcb_set_input_focus_checked(xwm->xcb_conn,
+			XCB_INPUT_FOCUS_POINTER_ROOT,
+			XCB_NONE, XCB_CURRENT_TIME);
+	}
```
SirCmpwn: Minor issue - this would read better if you rearranged it a bit:

```c
if (!surface) {
	xcb_set_input_focus_checked(xwm->xcb_conn,
		XCB_INPUT_FOCUS_POINTER_ROOT,
		XCB_NONE, XCB_CURRENT_TIME);
	return;
}
if (surface->override_redirect) {
	return;
}
// ...
```

xwayland/xwm.c
```diff
@@ -65,31 +63,154 @@ static struct wlr_xwayland_surface *wlr_xwayland_surface_create(
...
+}
+
+
+void xwm_surface_activate(struct wlr_xwm *xwm,
+		struct wlr_xwayland_surface *xsurface) {
+	if (xwm->focus_surface == xsurface ||
+			(xsurface && xsurface->override_redirect)) {
```
SirCmpwn: This is unnecessary, && has a higher precedence than ||

xwayland/xwm.c
```diff
@@ -492,84 +614,282 @@ static void handle_configure_request(struct wlr_xwm *xwm,
...
+static void handle_focus_in(struct wlr_xwm *xwm,
+		xcb_focus_in_event_t *ev) {
+	// Do not interfere with grabs
+	if (ev->mode == XCB_NOTIFY_MODE_GRAB ||
+			ev->mode == XCB_NOTIFY_MODE_UNGRAB) {
+		return;
+	}
+
+	// Do not let X clients change the focus behind the compositor's
+	// back. Reset the focus to the old one if it changed.
```
SirCmpwn: FWIW this may not be desirable. Things like dmenu don't work well with this logic. It's a tricky question.

SirCmpwn:
LGTM aside from style issue. Will be running rootston with this patch today to test.

xwayland/xwm.c
emersion: Adding explicit parenthesis here makes it more obvious what the condition is though.

SirCmpwn: Nah, people should learn the little things imo

xwayland/xwm.c
emersion: Yeah, you're right.

rootston/xwayland.c
acrisci: This is a general multi-seat issue which I will make a ticket for probably today.

xwayland/xwm.c
acrisci: Do you have a specific issue?

SirCmpwn: No, just something to think about.

SirCmpwn:
Testing is good. Will merge when outstanding issues are addressed.

xwayland/xwm.c
acrisci: If I take the parens out it gives me a compiler warning.

acrisci:
Ok should be good to go.

PR #333, opened 2017-10-24
Add server-decoration protocol

emersion:
* [x] Set and communicate server preference
* [x] Bubble up what the client picked
* [x] Rootston implementation

Test plan:
1. Launch a server-decoration-aware client (such as KDE or GTK tip)
2. Switch server preference in rootston
3. See that if preference is server, the client respects that (or not)

### DIFF
--- a/include/rootston/desktop.h
+++ b/include/rootston/desktop.h
@@ -37,11 +37,13 @@ struct roots_desktop {
 	struct wlr_xdg_shell_v6 *xdg_shell_v6;
 	struct wlr_gamma_control_manager *gamma_control_manager;
 	struct wlr_screenshooter *screenshooter;
+	struct wlr_server_decoration_manager *server_decoration_manager;
 
 	struct wl_listener output_add;
 	struct wl_listener output_remove;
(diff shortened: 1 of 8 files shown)


SirCmpwn:
Eeek, wrong button

SirCmpwn:
What are your plans wrt rootston implementation here?

emersion:
I'm thinking of simplifying the interface by automatically sending a `mode` event to the client. The spec says:

>The client can request a different mode through the decoration request.
        The server will acknowledge this by another event with the same mode. So
        even if a server prefers server-side decoration it's possible to force a
        client-side decoration.

So the server cannot choose to acknowledge the request with another value than the one requested.

Maybe we could completely remove `wlr_server_decoration` from the API and replace that with an event `wlr_server_decoration_mode_event` containing the surface and the mode.

SirCmpwn:
Looks good to me.

emersion:
Okay, I think this is ready to go. Tested with GTK master.

SirCmpwn:
Sweet, LGTM. Thanks!

Issue #232, opened 2017-10-07
Closing xwayland windows in rootston segfaults


SirCmpwn:

emersion:
```
xterm: fatal IO error 11 (Resource temporarily unavailable) or KillClient on X server ":5"

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff70bacfa in __strcmp_sse2_unaligned () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff70bacfa in __strcmp_sse2_unaligned () at /usr/lib/libc.so.6
#1  0x00007ffff797b325 in  () at /usr/lib/libwayland-server.so.0
#2  0x00007ffff79761c5 in wl_resource_instance_of () at /usr/lib/libwayland-server.so.0
#3  0x00007ffff7977f9d in wl_shm_buffer_get () at /usr/lib/libwayland-server.so.0
#4  0x00007ffff7bb4f79 in wlr_surface_flush_damage (surface=0x5555560afbe0) at ../types/wlr_surface.c:456
#5  0x000055555555c4c8 in render_surface (surface=0x5555560afbe0, desktop=0x555555f619d0, wlr_output=0x555555f70970, when=0x7fffffffe068, lx=180, ly=175, rotation=0)
    at ../rootston/output.c:22
#6  0x000055555555c45c in render_view (view=0x5555560a6720, desktop=0x555555f619d0, wlr_output=0x555555f70970, when=0x7fffffffe068) at ../rootston/output.c:83
#7  0x000055555555c281 in output_frame_notify (listener=0x555556045100, data=0x555555f70970) at ../rootston/output.c:101
#8  0x00007ffff7ba3f0b in wl_signal_emit (signal=0x555555f70a50, data=0x555555f70970) at /usr/include/wayland-server-core.h:388
#9  0x00007ffff7ba4af3 in surface_frame_callback (data=0x555555f70970, cb=0x555556045670, time=34947392) at ../backend/wayland/output.c:23
#10 0x00007ffff4ae11c8 in ffi_call_unix64 () at /usr/lib/libffi.so.6
#11 0x00007ffff4ae0c2a in ffi_call () at /usr/lib/libffi.so.6
#12 0x00007ffff6e25bad in  () at /usr/lib/libwayland-client.so.0
#13 0x00007ffff6e22679 in  () at /usr/lib/libwayland-client.so.0
[...]

PR #334, opened 2017-10-24
Fix order of WM_HINTS and WM_PROTOCOLS in xwm atom_map

inclement:
This seems to fix https://github.com/swaywm/wlroots/issues/232 .

The issue there seems to arise from the fact that the xwayland protocols for the surface are not set up correctly, because the order of items in `atom_map` does not match the order of `atom_name` xwm.h.

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -16,8 +16,8 @@
 const char *atom_map[ATOM_LAST] = {
 	"WL_SURFACE_ID",
 	"WM_DELETE_WINDOW",
-	"WM_HINTS",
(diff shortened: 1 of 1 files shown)


acrisci:
Thanks so much 👍 .

I'm doing almost a complete rewrite of this in #331 so if you're interested in the xwayland code, some help with general approach and testing would be appreciated.

inclement:
@acrisci I really don't know much about (x)wayland, I just wanted to learn a little about how everything fits together and this was a bug I could understand! I can try to test your changes though.

PR #335, opened 2017-10-25
Focus last view on close in rootston

emersion:
I renamed `view-initialize` to `view_setup` because I want to keep `initialize` when initializing memory of a struct (like `wlr_keyboard_init`). I added the counterpart `view_teardown`.

Test plan: open a window, open another window, close the active window. Check that the previously unfocused window gets active.

### DIFF
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -107,12 +107,24 @@ bool view_center(struct roots_view *view) {
 	return true;
 }
 
-void view_initialize(struct roots_view *view) {
+void view_setup(struct roots_view *view) {
 	view_center(view);
-	struct roots_input *input = view->desktop->server->input;
 
+	struct roots_input *input = view->desktop->server->input;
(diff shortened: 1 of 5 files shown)


rootston/desktop.c
```diff
@@ -107,14 +107,26 @@ bool view_center(struct roots_view *view) {
...
+void view_setup(struct roots_view *view) {
 	view_center(view);
-	struct roots_input *input = view->desktop->server->input;
 
+	struct roots_input *input = view->desktop->server->input;
 	set_view_focus(input, view->desktop, view);
 	wlr_seat_keyboard_notify_enter(input->wl_seat, view->wlr_surface);
 }
 
+void view_teardown(struct roots_view *view) {
```
acrisci: What if the view being torn down is not the focused view?

emersion: That's taken care of with `views->items[views->length-1] != view`.

acrisci:
👍

PR #340, opened 2017-10-26
Do not set hw cursor if disabled when switching VT

emersion:
Fixes the bug reported by @Timidger in this comment: https://github.com/swaywm/wlroots/pull/331#issuecomment-339531064

Test plan:
1. Open `qutebrowser` in rootston, move the mouse in the window
2. Switch VT
3. Switch back to rootston
4. There's only one cursor: the software cursor in the `qutebrowser` window. There's no hardware cursor in the top-left corner.

### DIFF
--- a/include/backend/drm/drm.h
+++ b/include/backend/drm/drm.h
@@ -33,6 +33,7 @@ struct wlr_drm_plane {
 	float matrix[16];
 	struct wlr_texture *wlr_tex;
 	struct gbm_bo *cursor_bo;
+	bool cursor_enabled;
 
 	union wlr_drm_plane_props props;
 };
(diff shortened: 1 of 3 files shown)


include/backend/drm/drm.h
```diff
@@ -33,6 +33,7 @@ struct wlr_drm_plane {
 	float matrix[16];
 	struct wlr_texture *wlr_tex;
 	struct gbm_bo *cursor_bo;
+	bool cursor_enabled;
```
ascent12: There already is a bool in wlr_output which checks if it's a hardware or software cursor.

emersion: This one has a different purpose: it checks whether or not the hardware cursor is enabled. A hardware cursor can be disabled even if software cursors are not used (e.g. when playing a fullscreen video, the cursor is hidden).

PR #341, opened 2017-10-26
Per-keyboard configuration

### LINKED ISSUE #326: Rootston: configure keymap per keyboard
The rootston config should be able to configure a keymap for each keyboard in the keyboard config separately.

emersion:
* We may want to use `config.devices` instead of `config.keyboard`?

Test plan:
1. Set nothing, no meta key should be set
2. Set `meta-key` in the `keyboard:*` section, meta key should work
3. Set `layout` in the `keyboard:*` section, should change the layout
4. Set another `layout` or `meta-key` in a `keyboard:<device name>` section, should change it for only this keyboard
5. `XKB_DEFAULT_LAYOUT` should still work

Fixes #326

### DIFF
--- a/rootston/rootston.ini.example
+++ b/rootston/rootston.ini.example
@@ -29,7 +29,7 @@ map-to-output = VGA-1
 # Restrict cursor movements for this mouse to concrete rectangle
 geometry = 2500x800
 
-[keyboard]
+[keyboard:*]
 meta-key = Logo
 
 # Keybindings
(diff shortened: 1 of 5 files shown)


rootston/rootston.ini.example
```diff
@@ -29,7 +29,7 @@ map-to-output = VGA-1
 # Restrict cursor movements for this mouse to concrete rectangle
 geometry = 2500x800
 
-[keyboard]
+[keyboard:*]
```
bew: I would prefer to use `keyboard:default` instead of `keyboard:*`, because the latter feels like I'm setting sth for all keyboard, not just that this will be the default configuration (and all keyboard that don't have a specific config will fallback to the default).

SirCmpwn: I would just call this section `[keyboard]`

Issue #342, opened 2017-10-27
Make rootston resize aware of surface edge


acrisci:
Right now we always resize from the south east.

If you start the resize from for instance the north west corner of the surface, you should resize by changing the x/y coordinates and adjust width/height. Similarly for the other three corners.

Also display a resizing cursor for the appropriate edge.

emersion:
We should also display the appropriate cursor when moving and rotating views.

PR #344, opened 2017-10-27
Set compositor cursor for move, resize and rotate

emersion:
Test plan:
- Move a window with its border or with the meta key, check that the cursor is a hand
- Resize a window from each edge, check cursor
- Rotate a window, check cursor
- Make sure the surface cursor is used when move/resize/rotate ends

Updates #342

### DIFF
--- a/include/rootston/input.h
+++ b/include/rootston/input.h
@@ -88,8 +88,7 @@ struct roots_input {
 
 	// TODO: multiseat, multicursor
 	struct wlr_cursor *cursor;
-	struct wlr_xcursor_theme *theme;
-	struct wlr_xcursor *xcursor;
+	struct wlr_xcursor_theme *xcursor_theme;
 	struct wlr_seat *wl_seat;
 	struct wl_list drag_icons;
 	struct wl_client *cursor_client;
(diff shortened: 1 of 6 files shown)


SirCmpwn:
Can you establish rootston/xcursor to encapsulate all of the xcursor logic? I have more stuff coming for the HiDPI branch that would benefit from this, too.

SirCmpwn:
Move a window with its border or with the meta key, check that the cursor is a hand: **PASS**

Resize a window from each edge, check cursor **FAIL**

Rotate a window, check cursor  **PASS** assuming the cursor is supposed to be the grabby hand

Make sure the surface cursor is used when move/resize/rotate ends **PASS**

emersion:
>Resize a window from each edge, check cursor: FAIL

I cannot reproduce. Tested on both `weston-terminal` and `nautilus`. What client are you testing this with?

SirCmpwn:
gnome-terminal. Do you mean a shell-induced resize? I'm resizing with mod+rmouse somewhere in the middle of the view.

emersion:
Okay, so this should always resize with the bottom-right edge, with the appropriate cursor. This PR doesn't solve #342, it just solves the cursor part of it. Resizing with the meta key from all edges is somewhat unrelated and will be done in another PR.

SirCmpwn:
Aight

PR #347, opened 2017-10-27
Force resize for all edges in rootston

### LINKED ISSUE #342: Make rootston resize aware of surface edge

emersion:
Test plan: try to force resize from the four quaters of a window. Should resize from the corresponding edge.

Fixes #342


SirCmpwn:
LGTM, thanks!

PR #349, opened 2017-10-28
Enforce resize bounds in rootston

### LINKED ISSUE #346: Honor surface minimum size in rootston
Most of the shells have some way of determining a minimum size. This should be honored for cursor driven resizes. Otherwise, you can make the windows have zero or negative area which messes things up.

emersion:
`wl_shell` doesn't have resize bounds.

Test plan: try to resize a xdg shell surface or a xwayland surface, check it doesn't get smaller than the minimum size.

Fixes #346

### DIFF
--- a/rootston/xdg_shell_v6.c
+++ b/rootston/xdg_shell_v6.c
@@ -29,6 +29,21 @@ static void resize(struct roots_view *view, uint32_t width, uint32_t height) {
 	assert(view->type == ROOTS_XDG_SHELL_V6_VIEW);
 	struct wlr_xdg_surface_v6 *surf = view->xdg_surface_v6;
 	if (surf->role == WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL) {
+		struct wlr_xdg_toplevel_v6_state *state =
+			&surf->toplevel_state->current;
+		if (width < state->min_width) {
+			width = state->min_width;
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

rootston/xdg_shell_v6.c
```diff
@@ -29,6 +29,21 @@ static void resize(struct roots_view *view, uint32_t width, uint32_t height) {
 	assert(view->type == ROOTS_XDG_SHELL_V6_VIEW);
 	struct wlr_xdg_surface_v6 *surf = view->xdg_surface_v6;
 	if (surf->role == WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL) {
+		struct wlr_xdg_toplevel_v6_state *state =
+			&surf->toplevel_state->current;
+		if (width < state->min_width) {
+			width = state->min_width;
+		} else if (state->max_width > 0 &&
+				width < state->max_width) {
```
vially: Shouldn't this be `width > state->max_width`? Same thing on line 43: `height > state->max_height`

emersion: You're right, please send a patch!

vially: Sure. Created #356 to address it

PR #352, opened 2017-10-29
Add wlr_output_cursor

### LINKED ISSUE #351: Draw multiple cursors per output
Multiseat (#350) requires to draw potentially multiple cursors on the same output. The current API isn't designed for this, as an output can only draw zero or one cursor. That's because of hardware cursors: no more than one hardware cursor can be set per output (if hardware cursors are supported).

The functions affected are:
* `wlr_output_set_cursor` and `wlr_output_set_cursor_surface`
* `wlr_output_move_cursor`

A first question is: do we want to support drawing multiple cursors in `wlr_output` or in `wlr_cursor`? Currently it's possible to draw and move a cursor without `wlr_cursor` (see `examples/pointer.c`). Also, currently the `wlr_cursor` is responsible for moving the cursor on the output.

We could either:
1. Add support for multiple cursors per output in `wlr_output`. We'll need a `wlr_output_cursor_create(struct wlr_output *)` function that returns a new cursor for that output, and `wlr_output_cursor_{move,set_image}` functions to move and set its image.
2. Add support for multiple cursors per output in `wlr_cursor`. We'll just need to add a `wlr_cursor_set_image` function and `wlr_output` will be responsible for moving _and_ drawing the cursor.

Note that we can add `wlr_cursor_set_image` even with solution 1 to make things more easy for compositors. Also, solution 2 dangerously ties `wlr_output` to `wlr_cursor`: it won't be possible to use `wlr_output` without `wlr_cursor` anymore.

emersion:
Adds `wlr_output_cursor` to add support for multiple cursors displayed on one output. Also fixes hotspot with wayland backend.

Note: for now, compositors must call `wlr_cursor_set_image` when a new output is added to upload the cursor pixels to the new output. This won't be necessary after the renderer redesign, as the texture is cached in `wlr_output_cursor` and will be reusable (currently we cannot read pixels from a texture, and the output interface wants pixels).

Test plan: launch `examples/multi-cursor`, try to hotplug multiple pointers, move them around, unplug them.

Fixes #351

### DIFF
--- a/examples/multi-pointer.c
+++ b/examples/multi-pointer.c
@@ -155,17 +155,21 @@ static void handle_input_add(struct compositor_state *state,
 	wl_list_insert(&sample->cursors, &cursor->link);
 }
 
+static void cursor_destroy(struct sample_cursor *cursor) {
+	wl_list_remove(&cursor->link);
+	wl_list_remove(&cursor->cursor_motion.link);
+	wl_list_remove(&cursor->cursor_motion_absolute.link);
+	wlr_cursor_destroy(cursor->cursor);
+	free(cursor);
+}
+
 static void handle_input_remove(struct compositor_state *state,
 		struct wlr_input_device *device) {
 	struct sample_state *sample = state->data;
 	struct sample_cursor *cursor;
 	wl_list_for_each(cursor, &sample->cursors, link) {
 		if (cursor->device == device) {
-			wl_list_remove(&cursor->link);
-			wl_list_remove(&cursor->cursor_motion.link);
-			wl_list_remove(&cursor->cursor_motion_absolute.link);
-			wlr_cursor_destroy(cursor->cursor);
-			free(cursor);
+			cursor_destroy(cursor);
 			break;
 		}
 	}
@@ -212,9 +216,9 @@ int main(int argc, char *argv[]) {
 	wl_display_run(compositor.display);
 	compositor_fini(&compositor);
 
-	struct sample_cursor *cursor;
-	wl_list_for_each(cursor, &state.cursors, link) {
-		// TODO
+	struct sample_cursor *cursor, *tmp_cursor;
+	wl_list_for_each_safe(cursor, tmp_cursor, &state.cursors, link) {
+		cursor_destroy(cursor);
 	}
 
 	wlr_xcursor_theme_destroy(theme);
--- a/examples/multi-pointer.c
+++ b/examples/multi-pointer.c
@@ -134,6 +134,7 @@ static void handle_input_add(struct compositor_state *state,
 
 	struct sample_cursor *cursor = calloc(1, sizeof(struct sample_cursor));
 	cursor->state = sample;
+	cursor->device = device;
 
 	cursor->cursor = wlr_cursor_create();
 	wlr_cursor_attach_output_layout(cursor->cursor, sample->layout);
@@ -156,6 +157,22 @@ static void handle_input_add(struct compositor_state *state,
 	wl_list_insert(&sample->cursors, &cursor->link);
 }
 
+static void handle_input_remove(struct compositor_state *state,
+		struct wlr_input_device *device) {
+	struct sample_state *sample = state->data;
+	struct sample_cursor *cursor;
+	wl_list_for_each(cursor, &sample->cursors, link) {
+		if (cursor->device == device) {
+			wl_list_remove(&cursor->link);
+			wl_list_remove(&cursor->cursor_motion.link);
+			wl_list_remove(&cursor->cursor_motion_absolute.link);
+			wlr_cursor_destroy(cursor->cursor);
+			free(cursor);
+			break;
+		}
+	}
+}
+
 int main(int argc, char *argv[]) {
 	struct sample_state state = {
 		.default_color = { 0.25f, 0.25f, 0.25f, 1 },
@@ -173,6 +190,7 @@ int main(int argc, char *argv[]) {
 	compositor.output_remove_cb = handle_output_remove;
 	compositor.output_frame_cb = handle_output_frame;
 	compositor.input_add_cb = handle_input_add;
+	compositor.input_remove_cb = handle_input_remove;
 
 	state.compositor = &compositor;
 
--- a/examples/multi-pointer.c
+++ b/examples/multi-pointer.c
@@ -1,0 +1,202 @@
+#define _POSIX_C_SOURCE 199309L
+#define _XOPEN_SOURCE 500
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+#include <string.h>
+#include <unistd.h>
+#include <math.h>
+#include <assert.h>
+#include <wayland-server.h>
+#include <wayland-server-protocol.h>
+#include <xkbcommon/xkbcommon.h>
+#include <GLES2/gl2.h>
+#include <wlr/render/matrix.h>
+#include <wlr/render/gles2.h>
+#include <wlr/render.h>
+#include <wlr/backend.h>
+#include <wlr/backend/session.h>
+#include <wlr/types/wlr_keyboard.h>
+#include <wlr/types/wlr_output_layout.h>
+#include <wlr/xcursor.h>
+#include <wlr/types/wlr_cursor.h>
+#include <wlr/util/log.h>
+#include <wlr/types/wlr_list.h>
+#include "shared.h"
+#include "config.h"
+#include "cat.h"
+
+struct sample_state;
+
+struct sample_cursor {
+	struct sample_state *state;
+	struct wlr_input_device *device;
+	struct wlr_cursor *cursor;
+	struct wl_list link;
+
+	struct wl_listener cursor_motion;
+	struct wl_listener cursor_motion_absolute;
+	struct wl_listener cursor_button;
+	struct wl_listener cursor_axis;
+};
+
+struct sample_state {
+	struct compositor_state *compositor;
+	struct example_config *config;
+	struct wlr_xcursor *xcursor;
+	float default_color[4];
+	float clear_color[4];
+	struct wlr_output_layout *layout;
+	struct wl_list cursors; // sample_cursor::link
+};
+
+static void handle_output_frame(struct output_state *output,
+		struct timespec *ts) {
+	struct compositor_state *state = output->compositor;
+	struct sample_state *sample = state->data;
+	struct wlr_output *wlr_output = output->output;
+
+	wlr_output_make_current(wlr_output);
+
+	glClearColor(sample->clear_color[0], sample->clear_color[1],
+		sample->clear_color[2], sample->clear_color[3]);
+	glClear(GL_COLOR_BUFFER_BIT);
+
+	wlr_output_swap_buffers(wlr_output);
+}
+
+static void handle_output_add(struct output_state *ostate) {
+	struct sample_state *sample = ostate->compositor->data;
+
+	struct output_config *o_config =
+		example_config_get_output(sample->config, ostate->output);
+
+	if (o_config) {
+		wlr_output_transform(ostate->output, o_config->transform);
+		wlr_output_layout_add(sample->layout, ostate->output, o_config->x,
+			o_config->y);
+	} else {
+		wlr_output_layout_add_auto(sample->layout, ostate->output);
+	}
+
+	struct sample_cursor *cursor;
+	wl_list_for_each(cursor, &sample->cursors, link) {
+		example_config_configure_cursor(sample->config, cursor->cursor,
+			sample->compositor);
+
+		// TODO the cursor must be set depending on which surface it is
+		// displayed over which should happen in the compositor.
+		struct wlr_xcursor_image *image = sample->xcursor->images[0];
+		wlr_cursor_set_image(cursor->cursor, image->buffer, image->width,
+			image->width, image->height, image->hotspot_x, image->hotspot_y);
+
+		wlr_cursor_warp(cursor->cursor, NULL, cursor->cursor->x,
+			cursor->cursor->y);
+	}
+}
+
+static void handle_output_remove(struct output_state *ostate) {
+	struct sample_state *sample = ostate->compositor->data;
+
+	wlr_output_layout_remove(sample->layout, ostate->output);
+
+	struct sample_cursor *cursor;
+	wl_list_for_each(cursor, &sample->cursors, link) {
+		example_config_configure_cursor(sample->config, cursor->cursor,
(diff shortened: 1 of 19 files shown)


emersion:
This PR is ready for review.

examples/multi-pointer.c
```diff
@@ -0,0 +1,220 @@
...
+		wlr_output_layout_add_auto(sample->layout, ostate->output);
+	}
+
+	struct sample_cursor *cursor;
+	wl_list_for_each(cursor, &sample->cursors, link) {
+		example_config_configure_cursor(sample->config, cursor->cursor,
+			sample->compositor);
+
+		// TODO the cursor must be set depending on which surface it is
+		// displayed over which should happen in the compositor.
```
acrisci: Should we add a way to disable client cursors for compositors that don't want to use them? Could it be possible for compositors to have a single cursor theme?

emersion: Just ignore the `set_cursor` request from clients.

acrisci: Yeah I think this todo is kind of old then before we knew about this stuff.

emersion: Exactly, I should remove it (and the other ones across all examples).

types/wlr_output_layout.c
```diff
@@ -153,6 +156,8 @@ static struct wlr_output_layout_output *wlr_output_layout_output_create(
 	wl_signal_add(&output->events.destroy, &l_output->state->output_destroy);
 	l_output->state->output_destroy.notify = handle_output_destroy;
 
+	wl_signal_emit(&layout->events.add, l_output);
```
acrisci: If you add auto, the layout output won't be configured here. You have to emit all the events after `wlr_output_layout_reconfigure()` which is guaranteed to be called when the layout changes.

types/wlr_output.c
```diff
@@ -188,94 +188,172 @@ void wlr_output_set_position(struct wlr_output *output, int32_t lx,
...
+}
 
-	if (!output->cursor.texture) {
-		output->cursor.texture =
-			wlr_render_texture_create(output->cursor.renderer);
-		if (!output->cursor.texture) {
-			return false;
-		}
+void wlr_output_make_current(struct wlr_output *output) {
+	output->impl->make_current(output);
```
acrisci: Do we check if it has the interface methods? Seems inconsistent with destroy.

types/wlr_output.c
```diff
@@ -188,94 +188,172 @@ void wlr_output_set_position(struct wlr_output *output, int32_t lx,
...
+	if (cursor->surface != NULL) {
+		texture = cursor->surface->texture;
+		renderer = cursor->surface->renderer;
 	}
 
-	return wlr_texture_upload_pixels(output->cursor.texture,
-		WL_SHM_FORMAT_ARGB8888, stride, width, height, buf);
+	// We check texture->valid because some clients set a cursor surface
+	// with a NULL buffer to hide it
+	if (renderer && texture && texture->valid) {
```
acrisci: I think we are going to start using `wlr_surface_has_buffer()` for this which is future proof if texture->valid goes away.

types/wlr_output.c
emersion: This is the old code, unchanged. I can add the check if needed.

types/wlr_output.c
emersion: Good point.

types/wlr_output_layout.c
emersion: In `wlr_output_layout_reconfigure` I don't know which output has been added, which is needed for creating one cursor per output in `wlr_cursor`. I'll instead emit the `add` event in auto.

types/wlr_output.c
emersion: There are a lot of other functions that are mandatory for outputs (`enable`, `transform`, `swap_buffers`...) so I'm going to leave it as is, unless there's a reason we should make all of these optional.

types/wlr_output_layout.c
emersion: In the end, `wlr_output_layout_add_auto` is also calling `wlr_output_layout_output_create` so it's fine. I also checked that `wlr_output_layout_output` is not created somewhere else too.

types/wlr_output.c
acrisci: I kind of wish there was a way to check if the required methods are there before we get to them. Wayland doesn't do this and it led to some crashes early on. I'm thinking of someone who is new to the library (which will be everyone at some point).

emersion: That's true, a segfault should definitely not happen. Maybe make the `create` function fail and log the error?

types/wlr_output_layout.c
acrisci: It's still not right. At this point, the x/y coordinates are always zero. If it gets added auto, those will not be the final coordinates. You should emit the event after the layout is reconfigured.

emersion: Oh, I misunderstood what you pointed out. It doesn't matter here, but it makes sense to emit the event after the output is fully configured.

types/wlr_output.c
```diff
@@ -188,94 +188,174 @@ void wlr_output_set_position(struct wlr_output *output, int32_t lx,
...
-		wl_list_remove(&output->cursor.surface_commit.link);
-		wl_list_remove(&output->cursor.surface_destroy.link);
-		output->cursor.surface = NULL;
+void wlr_output_swap_buffers(struct wlr_output *output) {
+	struct wlr_output_cursor *cursor;
+	wl_list_for_each(cursor, &output->cursors, link) {
+		if (output->hardware_cursor == cursor) {
+			continue;
+		}
+		output_cursor_render(cursor);
```
acrisci: Is it a problem that this rendering happens in swap buffers and not the frame callback like the rest of the rendering?

acrisci: What if the compositor wants to customize the rendering of the cursor or hide it?

emersion: A problem is that the `frame` callback is emitted by the output implementation (so by the backend), so we can't just add a listener in `wlr_output` (otherwise we have no guarantee that the cursor drawn will be on top of everything).

(Also this is how it worked before this PR)

@ascent12 Thoughts?

emersion: >What if the compositor wants to customize the rendering of the cursor or hide it?

The compositor is responsible for creating `wlr_output_cursor`s. So if it doesn't want to render a cursor, it can just do nothing. If it wants to customize how the cursor looks, it can just create a `wlr_output_cursor` with whatever pixels it wants. If it wants to render cursors differently, it can never use `wlr_output_cursor`s and render them itself.

types/wlr_output.c
```diff
@@ -288,170 +368,107 @@ static void handle_cursor_surface_commit(struct wl_listener *listener,
...
 	} else {
-		set_cursor(output, NULL, 0, 0, 0, hotspot_x, hotspot_y);
+		// TODO: if hardware cursor, disable cursor
 	}
 }
 
-bool wlr_output_move_cursor(struct wlr_output *output, int x, int y) {
-	output->cursor.x = x;
-	output->cursor.y = y;
+bool wlr_output_cursor_move(struct wlr_output_cursor *cursor, int x, int y) {
```
acrisci: wlr-cursor coordinates are doubles but these are int so there is some rounding going on here and output-cursor coordinates will usually be different from cursor coordinates.

emersion: Is that a problem?

types/wlr_output.c
```diff
@@ -288,170 +368,107 @@ static void handle_cursor_surface_commit(struct wl_listener *listener,
...
-	wl_list_remove(&output->cursor.surface_commit.link);
-	wl_list_remove(&output->cursor.surface_destroy.link);
-	output->cursor.surface = NULL;
+	struct wlr_output_cursor *cursor = wl_container_of(listener, cursor,
+		surface_destroy);
+	output_cursor_reset(cursor);
 }
 
-void wlr_output_set_cursor_surface(struct wlr_output *output,
+void wlr_output_cursor_set_surface(struct wlr_output_cursor *cursor,
```
acrisci: What if I call this on a cursor that already has an image, or call set_image on a cursor that already has a surface? Is that case handled?

types/wlr_output.c
acrisci: The worst that can happen is that the cursor will be rendered 1px away from where it really is. If we can render the cursor at double coordinates, we absolutely should. If not, that's fine because the wlr-cursor will still have the right coordinates for the next move so the inaccuracies will not accumulate as the mouse moves.

types/wlr_output.c
emersion: Yeah. These things happen right now in rootston because compositor cursors are using `set_image` and client cursors are using `set_surface`.

* If you call `set_surface` on a cursor that has an image, the `wlr_output_cursor.surface` has priority over the image. The `texture` and the `renderer` are not freed because they may be reused by a `set_image` call.
* If you call `set_image` on a cursor that has a surface, the `output_cursor_reset` function cleanups the surface.

types/wlr_output.c
emersion: Hardware cursors only support integers, so I don't think that's possible.

types/wlr_output.c
```diff
@@ -188,94 +188,174 @@ void wlr_output_set_position(struct wlr_output *output, int32_t lx,
...
-	if (!output->cursor.texture) {
-		output->cursor.texture =
-			wlr_render_texture_create(output->cursor.renderer);
-		if (!output->cursor.texture) {
-			return false;
+void wlr_output_make_current(struct wlr_output *output) {
+	output->impl->make_current(output);
+}
+
+static void output_cursor_render(struct wlr_output_cursor *cursor) {
```
acrisci: Will it render if the cursor is not on the output?

types/wlr_cursor.c
```diff
@@ -472,10 +513,50 @@ static void handle_layout_destroy(struct wl_listener *listener, void *data) {
...
+		wlr_log(L_ERROR, "Failed to create wlr_output_cursor");
+		free(output_cursor);
+		return;
+	}
+
+	output_cursor->layout_output_destroy.notify = handle_layout_output_destroy;
+	wl_signal_add(&l_output->events.destroy,
+		&output_cursor->layout_output_destroy);
+
+	wl_list_insert(&state->cursor->state->output_cursors, &output_cursor->link);
```
acrisci: Does it get removed from teh list when `output_cursor` is destroyed? Or does that only happen when the actual output is destroyed?

types/wlr_output.c
emersion: It will render off-screen, so it won't be visible.

acrisci: I think it should only render if it will be visible on the output as an optimization.

types/wlr_cursor.c
emersion: That only happens when the output is destroyed, since there's no `destroy` event in `wlr_output_cursor`. In other words, `wlr_output_cursor`s are destroyed only when the corresponding output is destroyed (or when `wlr_output_cursor_destroy` is called, not relevant here).

acrisci:
The multi-pointer example is showing a dead pointer in the top left of each of my screens.

emersion:
>The multi-pointer example is showing a dead pointer in the top left of each of my screens.

Do you have a ThinkPad trackpoint? If so, this is its cursor. You can move it with the trackpoint.

acrisci:
Gettings segfaults on close

X11:

```
==15313==The signal is caused by a READ memory access.
==15313==Hint: address points to the zero page.
    #0 0x7fe892fc6c39 in wlr_cursor_detach_output_layout ../types/wlr_cursor.c:117
    #1 0x7fe892fc976e in handle_layout_destroy ../types/wlr_cursor.c:513
    #2 0x7fe892fd0980 in wl_signal_emit /usr/include/wayland-server-core.h:388
    #3 0x7fe892fd0c16 in wlr_output_layout_destroy ../types/wlr_output_layout.c:55
    #4 0x404375 in main ../examples/multi-pointer.c:217
    #5 0x7fe89253c889 in __libc_start_main (/lib64/libc.so.6+0x20889)
    #6 0x402669 in _start (/home/tony/projects/wlroots/build/examples/multi-pointer+0x402669)
```

DRM:

```
==15405==ERROR: AddressSanitizer: heap-use-after-free on address 0x61400000f3e0 at pc 0x7f324cbc5a43 bp 0x7ffdb6053d20 sp 0x7ffdb6053d10
READ of size 8 at 0x61400000f3e0 thread T0
    #0 0x7f324cbc5a42 in wlr_drm_backend_destroy ../backend/drm/backend.c:33
    #1 0x7f324cbbf854 in wlr_backend_destroy ../backend/backend.c:36
    #2 0x7f324cbdaa9c in multi_backend_destroy ../backend/multi/backend.c:35
    #3 0x7f324cbbf854 in wlr_backend_destroy ../backend/backend.c:36
    #4 0x407a44 in compositor_fini ../examples/shared.c:497
    #5 0x4042f8 in main ../examples/multi-pointer.c:213
    #6 0x7f324c16e889 in __libc_start_main (/lib64/libc.so.6+0x20889)
    #7 0x402669 in _start (/home/tony/projects/wlroots/build/examples/multi-pointer+0x402669)
```

acrisci:
> Do you have a ThinkPad trackpoint? If so, this is its cursor. You can move it with the trackpoint.

It could definitely be an input device I'm not expecting.

emersion:
>It could definitely be an input device I'm not expecting.

Check the logs: all input devices are logged.

types/wlr_output.c
emersion: Fixed. Keep in mind that:
1. Most of the time, software cursors aren't used because hardware cursors are available.
2. Cursor surfaces are small
3. Usally there's just one cursor. If there's more than one, there's not a lot of them.

types/wlr_output.c
acrisci: Add assertions. It's a library user error that needs to be fixed right away.

acrisci:
Looks good to me 👍 . Will merge tonight unless anyone else finds any issues with it.

types/wlr_output.c
emersion: (At least) the Wayland backend is missing `enable`. I wonder what's the best idea:
1. Implement `enable` for Wayland, `enable(false)` destroys the window
2. Make `enable` optional, or implement a no-op `enable` function in the Wayland backend

SirCmpwn: Make enable optional, don't call it if it's null (no-op).

emersion:
All issues pointed by @acrisci should have been addressed now.

PR #354, opened 2017-10-29
Whitepaper

SirCmpwn:
Started working on documentation. Feedback welcome. Here's a rendered PDF of the current version:

https://sr.ht/w-zI.pdf

### DIFF
--- a/whitepaper/wlroots.tex
+++ b/whitepaper/wlroots.tex
@@ -219,6 +219,44 @@ call to action for other compositors to participate in its development and
 implementation, and an example of the sorts of functionality we're committed to
 providing on Wayland.
 
+\subsection{Security Considerations}
+
+A major concern in implementing many of these features is doing them in a
+secure way. Making these interfaces available to third-party software securely
+is the subject of ongoing research and discussion, but we do not believe that
+the issue is too poorly understood to proceed with implementing these features.
+In wlroots, we have chosen to leave securing these protocols as a matter for the
+compositor to support via wl_display_set_global_filter.
+
+Let's summarize ongoing efforts by compositors in this area. The Sway compositor
+has, as of writing, invested the most effort in making privledged protocols
+available to third parties. This is not without consequences-- Sway's current
+implementation has known flaws. We are comfortable being more liberal in our
+exploration in this area, hoping that we can become more secure over time. Our
+justification for this approach is that, at worst, we begin with a system no
+more secure than X11, but over time we will be able to secure it more as we
+experiment with different approaches.
+
+Sway's approach today is configuration files, owned by root, which contain a
+list of paths to executables and the permissions granted to each. To enforce
+these, we look them up via /proc/[pid]/exe. This has a number of drawbacks we
+are working to address. First, this is not very effective for interpreted
+programs, e.g. /usr/bin/python3. Additionally, it may be possible on some
+platforms and in some conditions to manipulate the value of /proc/[pid]/exe.
+Utilizing procfs is also not portable.
+
+In the future we're exploring the possibility of only permitting secure access
+for clients which were forked from Sway. This approach should solve most of the
+outstanding problems, but other deficiencies remain. Granting permissions to
+previously un-approved clients at runtime (with user content) requires
+additional work (something the Orbital compositor has explored). Offering
+similar permissions to children of those processes when appropriate is also a
+non-trivial problem.
+
+Collaboration with the broader Wayland community in the matter of security is
+highly sought after. We ask you to participate in our discussions on the
+subject; contact information is available at the end of this paper.
+
 \section{Project Status}
 
 Integrations with wlroots are well underway in the Wayland community.\\
--- a/whitepaper/wlroots.tex
+++ b/whitepaper/wlroots.tex
@@ -24,13 +24,13 @@ compositor designs.
 
 Wayland is a protocol used to communicate between user applications and a
 \emph{compositor}. The compositor has ownership over input and output
-resources -- typically displays connected over DVI, HDMI, etc, and keyboards,
+resources-- typically displays connected over DVI, HDMI, etc, and keyboards,
 mice, etc.\\
 
 wlroots (stylized in lowercase, or as "wlr") is a subproject of the Sway
 compositor. Originally, Sway was written on top of a similar library called wlc.
-This library offered abstractions on top of DRM, GBM, GLES, libinput, and so on
--- allowing Sway to focus on building the software specific to the compositor.
+This library offered abstractions on top of DRM, GBM, GLES, libinput, and so
+on-- allowing Sway to focus on building the software specific to the compositor.
 However, with time it became clear that wlc's design is too limiting, and many
 of Sway's long-term goals were difficult to achieve based on it.\\
 
@@ -38,7 +38,7 @@ As a result, we built wlroots. We believe it offers a substantially better
 basis for development of Wayland compositors than any competing approach.
 We offer a powerful library that removes most of the heavy lifting from building
 a Wayland compositor while providing compositor authors as much flexibility in
-their design as they might gain by doing it all themselves -- I've jokingly
+their design as they might gain by doing it all themselves-- I've jokingly
 referred to wlroots as
 "about 30,000 lines of code you were going to write anyway." We hope that
 compositors based on wlroots will be written faster, work better, and will
@@ -66,8 +66,8 @@ Another approach is to simply write your compositor entirely in-house. This is
 the approach favored by \textbf{mutter} and \textbf{kwin} of the GNOME and KDE
 projects respectively. The advantage of this approach is the flexibility to
 design your compositor in any way you please. However, lots of work is
-duplicated between projects with this approach, which can lead to compositors
-which are subtly incompatible with each other.
+duplicated between projects with this approach, which can (and does) lead to
+compositors which are subtly incompatible with each other.
 
 The approach taken by \textbf{wlroots} offers reusability and consistency
 between compositors based on it, and offloads lots of work from the compositor
@@ -135,7 +135,7 @@ Each of these protocols, like the shells and everything else, is opt-in. The
 compositor author chooses if our implementations suit their needs, and if not,
 their own implementation is easily used instead and interops correctly with the
 rest of wlroots. For some important protocols (such as wlr_seat), disentangling
-it from the rest of wlroots can be more difficult -- but in these cases we've
+it from the rest of wlroots can be more difficult-- but in these cases we've
 made our implementation extremely flexible to make it less likely that it will
 not suit your needs.
 
@@ -173,11 +173,11 @@ encouraged to copy and adapt any code they find useful for their own
 compositors. Many of the batteries that are "not included" with wlroots can be
 found in Rootston.
 
-\section{Promoting an Interopable Ecosystem}
+\section{Promoting an Interoperable Ecosystem}
 
 We believe that it is tremendously important for more Wayland compositors to be
 extensible by their users and for these tools to work across the entire Wayland
-desktop ecosystem. No one will switch from X11 to Wayland for security alone --
+desktop ecosystem. No one will switch from X11 to Wayland for security alone--
 we must make compositors that offer compelling features and do not ask the user
 to sacrifice the workflows they've grown used to, insecure as they may have been
 on X11.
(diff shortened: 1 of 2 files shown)


whitepaper/wlroots.tex
```diff
@@ -0,0 +1,440 @@
...
+
+Another approach is to simply write your compositor entirely in-house. This is
+the approach favored by \textbf{mutter} and \textbf{kwin} of the GNOME and KDE
+projects respectively. The advantage of this approach is the flexibility to
+design your compositor in any way you please. However, lots of work is
+duplicated between projects with this approach, which can lead to compositors
+which are subtly incompatible with each other.
+
+The approach taken by \textbf{wlroots} offers reusability and consistency
+between compositors based on it, and offloads lots work from the compositor
```
martinetd: "and offloads lots of work", probably?

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,440 @@
...
+corresponding Linux subsystems; \emph{wayland}, which uses Wayland windows on
+another running Wayland compositor; the similar \emph{x11} backend; \emph{rdp},
+which uses the remote desktop protocol; and so on. An interface is provided so
+that compositor implementations can supply their own backend implementations as
+well.
+
+Also available is the \emph{multi} backend, which acts as a container for many
+backends. This is necessary, for example, to use DRM and libinput together. It
+also allows you to add and remove backends at runtime. An example use-case:
+adding an RDP backend at runtime.
```
martinetd: Consistency here: backends given earlier were not capitalized (and are not in the code, event if the emphasis on lower case x11 looks odd when rendered), but are capitalized here without emphasis.

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,440 @@
...
+\subsection{wlr_matrix}\label{wlr matrix}
+
+TODO
+
+\subsection{wlr_output}\label{wlr output}
+
+\code{wlr/types/wlr_output.h}\\
+
+\textbf{wlr_output} abstracts access to display hardware provided by backends.
+The most important functionality offered by \textbf{wlr_outputs} is the frame
```
martinetd: move the `s` out of the bold?

SirCmpwn: I tried that initially and I don't think it really works.

Timidger: You can reword it so you don't have the s. Eg:  "the most important functionality provided by a wlr_output is... "

SirCmpwn: Nah, this is going to come up all over the place and I'd rather not try to worm our way around it everywhere.

Timidger: Good point, best to just keep it like it is. I don't think anyone will be confused by the s, and it looks better being a part of the formatting.

include/wlr/backend.h
```diff
@@ -11,16 +11,30 @@ struct wlr_backend {
 	const struct wlr_backend_impl *impl;
 
 	struct {
+		/** Raised when a wlr_input_device is added. */
 		struct wl_signal input_add;
```
emersion: We should probably document for all `wl_list` and `wl_signal` which are the datatypes used.

whitepaper/wlroots.tex
```diff
@@ -92,7 +92,7 @@ \subsection{Backends}\label{high level backends}
 Also available is the \emph{multi} backend, which acts as a container for many
 backends. This is necessary, for example, to use DRM and libinput together. It
 also allows you to add and remove backends at runtime. An example use-case:
-adding an RDP backend at runtime.
+adding an rdp backend at runtime.
```
emersion: I think this should be capitalized properly.

SirCmpwn: -_-

whitepaper/wlroots.tex
```diff
@@ -379,62 +412,125 @@ \subsection{wlr_output_layout}\label{wlr output layout}
...
+A \textbf{wlr_texture} stores state related to a texture on the GPU. It also
+provides convenience functions for uploading pixels from pointers, Wayland
+shared memory buffers, or DRM buffers.
 
 \subsection{wlr_touch}\label{wlr touch}
 
-TODO
+\code{wlr/types/wlr_touch.h}\\
+
+\textbf{wlr_tablet_tool} provides an abstraction for backend-specific (multi-)
```
emersion: Tyo here - should be `wlr_touch`.

SirCmpwn: >Tyo

heh

include/wlr/backend.h
```diff
@@ -11,16 +11,30 @@ struct wlr_backend {
...
+		/** Raised when a wlr_input_device is removed. */
 		struct wl_signal input_remove;
+		/** Raised when a wlr_output is added. */
 		struct wl_signal output_add;
+		/** Raised when a wlr_output is removed. */
```
acrisci: I think it should be "emitted" for events.

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,536 @@
...
+\usepackage[english]{babel}
+\newcommand{\code}[1]{\texttt{#1}}
+\begin{document}
+\title{Modular Wayland compositors with wlroots}
+\author{Drew DeVault et al
+    \thanks{Complete list available at \url{https://github.com/swaywm/wlroots/graphs/contributors}}}
+\maketitle
+
+\begin{abstract}
+wlroots is a flexible and modular library upon which Wayland compositors can be
```
acrisci: I like stylized lowercase wlroots, but I wish we would make an exception for capitalization at the beginning of a sentence.

acrisci:
Can I just add my commits to this branch, or do you want to merge and do fresh PRs, or something else?

SirCmpwn:
Feel free to just add commits.

SirCmpwn:
You can also do pull requests against this branch if you want your changes reviewed separately, up to you.

acrisci:
I think discussion of the backends and rendering is difficult to have without having their purpose clear in your head. They also contain the most difficult to grasp concepts, so we probably want to move them a little farther back until we know the reader is engaged.

How much Wayland knowledge are we assuming and who is the target audience? I think it should be a programmer and current Linux desktop user with some experience with X11 window managers and protocols who is interested in Wayland but doesn't know any specifics.

The simplest thing to understand is the client and it's also probably the most important thing since they are the ones using all these interfaces so we should probably be leading with that. Talk about the registry and the shared object model that happens through resources. That leads naturally into a discussion of the compositor. From there go to surfaces (what is a buffer? what is double buffering?) and then go to an overview of shell surfaces.

Once we get there, we can branch out any way we want.

SirCmpwn:
I'm not sure that's the best approach. Explaining how all of Wayland works is out of scope for this document. wlroots is a very low level tool and requires an understanding of the Wayland protocol to use effectively. I think someone was thinking about making something similar to wlc on top of wlroots to bridge the gap - that sounds like a more appropriate place for that sort of documentation.

SirCmpwn:
One problem I have with the top-down approach (shells before backends) is that it's the opposite of how you actually write a compositor with wlroots - you will generally be spinning up the backend long before you're thinking about shells.

acrisci:
> Explaining how all of Wayland works is out of scope for this document.

I'm just talking about a short introduction to the basics of Wayland. Without that, you are going to lose a lot of people on the first difficult paragraph.

The reason I'd like to lead with the highest level is because I believe the most valuable reader is someone interested in writing a window manager and is evaluating whether or not they can do this with wlroots. The topics this person will respond to the most are the outer layers. I think it would be best to give them the details after we've given them a chance to decide if this is a library they want to use.

SirCmpwn:
I suppose we can add a couple of paragraphs to the background section.

whitepaper/wlroots.tex
emersion: -1, you don't write IOS instead of iOS or EBay instead of eBay at the begining of a sentence. See https://english.stackexchange.com/questions/9063/how-do-you-capitalize-a-proper-noun-such-as-iphone

acrisci: It's not fair to use examples where there is a capitalization in beginning of the word. This is a practical matter where it's difficult to scan because you're not expecting the beginning of a sentence. Even Reddit now capitalizes their name in official documents.

emersion: elementary is also using this "always lowercase policy" (see https://elementary.io/brand). It seems that Reddit capitalizes its brand in the middle of sentences (it doesn't use the lowercase form in the middle and the capitalized form at the begining, it sticks to one form per document).

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,557 @@
...
+\newpage
+\section{Backends}\label{backends}
+
+\subsection{Backend initialization}\label{backend init}
+
+There are two steps to setting up a backend -- first they are created, and then
+they are started. During creation, the backend will obtain access to the
+resources it needs, configure events on the event loop, test for required
+features, and so on. During this stage, backends providing outputs are also
+expected to configure their EGL context. Starting a backend will begin
```
bb010g: Wording here seems awkward: "Starting a backend will begin presenting itself…"

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,557 @@
...
+
+\textbf{libweston} is the most notable alternative. It is a subproject of the
+Weston compositor, the reference Wayland compositor. However, it is generally
+poorly suited to novel compositors that don't use the traditional desktop
+paradigm. It is also less modular and difficult to break into several pieces.
+Additionally, as a consequence of being refactored out of an existing
+compositor, it inherits many design decisions of Weston itself, which may not be
+amenable to other compositors.
+
+\textbf{swc} and \textbf{wlc} are similar. They both strike a reasonable balance
```
mmlb: `swc` is first seen here and we do not know what it stands for.

ascent12: It's the name of a library, which is why it's bold.

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,557 @@
...
+
+A \textbf{wlr_texture} stores state related to a texture on the GPU. It also
+provides convenience functions for uploading pixels from pointers, Wayland
+shared memory buffers, or DRM buffers.
+
+\subsection{wlr_touch}\label{wlr touch}
+
+\code{wlr/types/wlr_touch.h}\\
+
+\textbf{wlr_tablet_tool} provides an abstraction for backend-specific (multi-)
```
mmlb: should be `wlr_touch`? Sorry if not, just going from surrounding clues.

whitepaper/wlroots.tex
mmlb: Right, but other libraries are introduced in previous text. `libweston` just before and `wlc` before that. If it's a well known library then maybe it isn't necessary, though I think it should be.

(I'm just reading the text and have been following issues here and in sway, I know close to nothing about graphics stuff...)

acrisci:
I think what's being written here is a user guide and not a whitepaper and it probably belongs in a wiki.

SirCmpwn:
To be honest I'm thinking we should drop the type reference section entirely for that reason. It's marginally useful at best and not really on-topic.

SirCmpwn:
I'm pretty happy with this draft. Let me know what you guys think and I'll merge+disseminate it in a couple of days.

SirCmpwn:
The final draft is available now: https://sr.ht/jAFC.pdf

whitepaper/wlroots.tex
```diff
@@ -0,0 +1,299 @@
...
+is the motivating force behind wlroots, and our port from wlc to wlroots is
+going well. We've found it most effective to set aside our current code and
+start from scratch, importing large parts of the old code as it becomes
+necessary. Porting is no simple process, but it seems that it will be well worth
+it. Dozens of features that our users have been asking for have been blocked
+behind this migration for a long time, and many are already supported in our
+port\footnote{Enumerated here:
+\url{https://github.com/swaywm/sway/issues/1524}}.\\
+
+\textbf{Way-Cooler} is also working on porting their compositor from wlc to
```
Timidger: I generally stylize it as Way Cooler. I'm not too picky about this though, since everyone does seem to hate software names that are more than one word.

SirCmpwn: Apologies, I'll remember that for the future.

PR #356, opened 2017-10-29
Fix resize bounds check in rootson

vially:
Fix small typos (introduced in #349) in the resize bounds checks


acrisci:
👍 looks good thanks

acrisci:
While testing this I also noticed that if I resize the window from the top-right downward past the minimum size, it moves the window around with the mouse. That might be another good bug to look into next.

vially:
Unless it's something visible to the naked eye (without actually running any code) I don't think I'll be able to help with that one (because I'm running nvidia proprietary drivers and I can't try wlroots).

emersion:
@vially You can start wlroots from another compositor (Wayland or X11) if you want to try rootston. ;)

vially:
@emersion Thanks for letting me know. I'll try that.

PR #358, opened 2017-10-30
refactor input time_sec to time_msec

acrisci:
Makes rootston code a bit simpler because now it doesn't have to do the weird casting.

`time_usec` is now unused.

### DIFF
--- a/backend/libinput/keyboard.c
+++ b/backend/libinput/keyboard.c
@@ -54,8 +54,8 @@ void handle_keyboard_key(struct libinput_event *event,
 	struct libinput_event_keyboard *kbevent =
 		libinput_event_get_keyboard_event(event);
 	struct wlr_event_keyboard_key wlr_event = { 0 };
-	wlr_event.time_usec = libinput_event_keyboard_get_time_usec(kbevent);
-	wlr_event.time_msec = (uint32_t)(wlr_event.time_usec / 1000);
+	wlr_event.time_msec =
+		usec_to_msec(libinput_event_keyboard_get_time_usec(kbevent));
 	wlr_event.keycode = libinput_event_keyboard_get_key(kbevent);
 	enum libinput_key_state state = 
 		libinput_event_keyboard_get_key_state(kbevent);
--- a/backend/libinput/keyboard.c
+++ b/backend/libinput/keyboard.c
@@ -54,8 +54,8 @@ void handle_keyboard_key(struct libinput_event *event,
 	struct libinput_event_keyboard *kbevent =
 		libinput_event_get_keyboard_event(event);
 	struct wlr_event_keyboard_key wlr_event = { 0 };
-	wlr_event.time_sec = libinput_event_keyboard_get_time(kbevent);
 	wlr_event.time_usec = libinput_event_keyboard_get_time_usec(kbevent);
+	wlr_event.time_msec = (uint32_t)(wlr_event.time_usec / 1000);
 	wlr_event.keycode = libinput_event_keyboard_get_key(kbevent);
 	enum libinput_key_state state = 
 		libinput_event_keyboard_get_key_state(kbevent);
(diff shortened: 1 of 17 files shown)


emersion:
Do we want to remove `time_usec`? Do we want to keep `time_sec`?

SirCmpwn:
Yeah, remove time_usec

backend/libinput/keyboard.c
```diff
@@ -54,8 +54,8 @@ void handle_keyboard_key(struct libinput_event *event,
 	struct libinput_event_keyboard *kbevent =
 		libinput_event_get_keyboard_event(event);
 	struct wlr_event_keyboard_key wlr_event = { 0 };
-	wlr_event.time_sec = libinput_event_keyboard_get_time(kbevent);
 	wlr_event.time_usec = libinput_event_keyboard_get_time_usec(kbevent);
+	wlr_event.time_msec = (uint32_t)(wlr_event.time_usec / 1000);
```
SirCmpwn: So we don't have to do this everywhere can you write a little helper function to convert?

acrisci: where do you want to put it?

SirCmpwn: Somewhere in the libinput backend, maybe backend.c

include/wlr/types/wlr_keyboard.h
```diff
@@ -63,7 +63,7 @@ enum wlr_key_state {
 };
 
 struct wlr_event_keyboard_key {
-	uint32_t time_sec;
+	uint32_t time_msec;
```
martinetd: Does it have to be a uint32_t ?
2^32 msec is "only" 49-50 days, I could totally see folks (e.g. myself) keep a computer hibernated for that long, and wakeup might feel weird... Well I guess as far as wlroots is concerned it'll just be far over whatever threshold we're considering anyway, but if time_msec replaces both time_sec and time_usec, and is computed from time_usec, I'd keep it as uint64_t
(heck, I'd do everything as usec if it's not too much of a pain, but the design is around msecs so sure might as well go for it now)

martinetd: Does it have to be a uint32_t ?
2^32 msec is "only" 49-50 days, I could totally see folks (e.g. myself) keep a computer suspended/hibernated for that long, and wakeup might feel weird... Well I guess as far as wlroots is concerned it'll just be far over whatever threshold we're considering anyway, but if time_msec replaces both time_sec and time_usec, and is computed from time_usec, I'd keep it as uint64_t
(heck, I'd do everything as usec if it's not too much of a pain, but the design is around msecs so sure might as well go for it now)

martinetd: This comes from libinput (`libinput_event_keyboard_get_time_usec`), and I think the functions we pass msec to are our's (`cursor_update_position`, `wlr_seat_pointer_notify_axis`, `do_cursor_button_press`, ...)?
Have I missed timings from wayland itself?

I guess it's a big change just to refactor our code for that anyway so it's probably fine to leave as is, was just something to eventually keep in mind.

emersion: >and I think the functions we pass msec to are our's

Yeah, but our functions don't really care about the time of these events, it's only used when forwarded to clients (`wlr_seat_pointer_notify_button`, `wlr_seat_pointer_notify_motion` and so on).

martinetd: Ok, as good as it will probably get as is then. Thanks for clearing it up.

acrisci:
Rename it to just `time` then?

SirCmpwn:
Nah, leave the msec on there. I usually don't go for hungarian notation but in this case I think it'd be particularly helpful.

acrisci:
Ok should be good.

PR #361, opened 2017-10-31
Check allocation errors in wlr_output_layout

emersion:


types/wlr_output_layout.c
```diff
@@ -8,14 +8,14 @@
 #include <assert.h>
 
 struct wlr_output_layout_state {
-	struct wlr_box *_box;
```
acrisci: I put an underscore here to emphasize that you should never read this, but only use the getter.

types/wlr_output_layout.c
```diff
@@ -52,25 +57,24 @@ void wlr_output_layout_destroy(struct wlr_output_layout *layout) {
 
 	wl_signal_emit(&layout->events.destroy, layout);
 
-	struct wlr_output_layout_output *_output, *temp = NULL;
-	wl_list_for_each_safe(_output, temp, &layout->outputs, link) {
-		wlr_output_layout_output_destroy(_output);
+	struct wlr_output_layout_output *output, *temp = NULL;
```
acrisci: should be `l_output` for consistency with the rest of the code.

types/wlr_output_layout.c
emersion: Oh, okay, I'll add a comment too then.

acrisci:
👍

PR #364, opened 2017-10-31
Remove transform support from Wayland backend

emersion:
Try this config:

```ini
[output:WL-1]
rotate = 90
```

Since hardware cursors are used, this breaks the desktop: the transformation cannot be applied to the parent compositor's cursor.

This PR makes `transform` optional for output interfaces and removes the Wayland implementation.

### DIFF
--- a/backend/wayland/output.c
+++ b/backend/wayland/output.c
@@ -46,12 +46,6 @@ static void wlr_wl_output_swap_buffers(struct wlr_output *_output) {
 	}
 }
 
-static void wlr_wl_output_transform(struct wlr_output *_output,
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Why can't we apply the transformation to the parent compositor's cursor? Surfaces can be transformed, and the cursor is just a surface, so...

emersion:
So we can transform the cursor's surface, but not the cursor position. See https://github.com/swaywm/wlroots/blob/master/backend/wayland/output.c#L175

SirCmpwn:
I don't have time to think the details through but my gut says that shouldn't be an issue.

Issue #188, opened 2017-10-02
Wrong movement bounds for cursor on rotated output


versusvoid:
When output is rotated by 90 degrees, same transformation is applied to whole cursor texture, which on my system have 256x256 pixels. As result visible cursor movements are  bounded to (-256, 1664) interval on y axis on FullHD monitor. I tested using "pointer" demo.

For me solution was #186, but it isn't universal.

versusvoid:
I've exported cursor texture with the following code:
```patch
diff --git a/backend/drm/drm.c b/backend/drm/drm.c
index 1a5fea9..01d6b1a 100644
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -24,6 +24,8 @@
 #include "backend/drm/iface.h"
 #include "backend/drm/util.h"
 
+#include <sys/stat.h>
+
 bool wlr_drm_check_features(struct wlr_drm_backend *drm) {
 	if (drmSetClientCap(drm->fd, DRM_CLIENT_CAP_UNIVERSAL_PLANES, 1)) {
 		wlr_log(L_ERROR, "DRM universal planes unsupported");

@@ -551,6 +553,21 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 	glReadPixels(0, 0, plane->surf.width, plane->surf.height, GL_BGRA_EXT, GL_UNSIGNED_BYTE, bo_data);
 	glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, 0);
 
+	{
+		struct stat info;
+		memset(&info, 0, sizeof(info));
+		char fn[1024] = "cursor-";
+		sprintf(fn + strlen(fn), "%dx%d-%d.data", bo_width, bo_height, conn->output.transform);
+		wlr_log(L_DEBUG, "Stating file %s", fn);
+		if (stat(fn, &info) == -1) {
+			assert(errno == ENOENT);
+			FILE *out = fopen(fn, "w");
+			fwrite(bo_data, 4*sizeof(uint8_t), bo_width*bo_height, out);
+			fclose(out);
+			wlr_log(L_DEBUG, "Saved cursor");
+		}
+	}
+
 	wlr_drm_surface_swap_buffers(&plane->surf);
 
 	gbm_bo_unmap(bo, bo_data);
```
Imported it with GIMP and got following images for normal output:
[...]

PR #365, opened 2017-10-31
Cursors ✕ Output Transformations fixes

emersion:
* Transforms pointer events under Wayland backend (see #364)
* Fixes broken 180 and flipped transformation under DRM backend (introduced in #188, probably never worked properly)
* Fixes disappearing software cursors when applying a 90 or 270 transformation (regression introduced in #352)
* Replaces mysterious switches with easy-to-understand calls to `wlr_output_transformation_*` helpers

Test plan: try all 8 transformations under Wayland and DRM. Open a view, check that the pointer doesn't jump when entering the view. Check (e.g. with a GTK app with buttons) that the view receives pointer events with the right position.

Currently the hardware cursor isn't transformed under the Wayland backend, waiting for #319 to do this.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -267,8 +267,8 @@ static void output_cursor_render(struct wlr_output_cursor *cursor) {
 
 	struct wlr_box output_box;
 	output_box.x = output_box.y = 0;
-	output_box.width = cursor->output->width;
-	output_box.height = cursor->output->height;
+	wlr_output_effective_resolution(cursor->output, &output_box.width,
+		&output_box.height);
 
 	struct wlr_box cursor_box;
 	output_cursor_get_box(cursor, &cursor_box);
@@ -294,7 +294,7 @@ void wlr_output_swap_buffers(struct wlr_output *output) {
 	struct wlr_output_cursor *cursor;
 	wl_list_for_each(cursor, &output->cursors, link) {
 		if (output->hardware_cursor == cursor) {
-			continue;
+			//continue; // TODO
 		}
 		output_cursor_render(cursor);
 	}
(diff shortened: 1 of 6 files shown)


types/wlr_output.c
```diff
@@ -455,3 +455,57 @@ uint32_t wlr_output_get_gamma_size(struct wlr_output *output) {
 	}
 	return output->impl->get_gamma_size(output);
 }
+
+void wlr_output_transform_apply_to_box(enum wl_output_transform transform,
```
acrisci: Probably goes in `wlr_box.h` as `wlr_box_transform()`

types/wlr_output.c
```diff
@@ -500,3 +500,49 @@ void wlr_output_cursor_destroy(struct wlr_output_cursor *cursor) {
 	wl_list_remove(&cursor->link);
 	free(cursor);
 }
+
+void wlr_output_transform_apply_to_box(enum wl_output_transform transform,
+		struct wlr_box *box, struct wlr_box *dest) {
```
acrisci: I think this belongs in `wlr_box.h` as `wlr_box_transform()`

types/wlr_output.c
emersion: I'm not sure about it. `enum wl_output_transform` is specific to outputs so it makes sense to put it here, and to expose it only to backends (via `wlr/interfaces/wlr_output.h`). Also, I've added another helper `wlr_output_transformation_invert` that wouldn't fit in `wlr_box.h`.

emersion:
This PR is ready for review.

types/wlr_output.c
SirCmpwn: wl_output_transform comes from wayland, not wlroots. We can use it wherever we want.

emersion: Sure. But these functions don't make sense outside backends - in particular, the compositor shouldn't use it.

SirCmpwn: Alright.

SirCmpwn:
LGTM :+1:

PR #377, opened 2017-11-01
Add wlr_output.needs_swap

### LINKED ISSUE #375: Allow compositors to do not call wlr_output_swap_buffers
Compositors right now have to call `wlr_output_swap_buffers` even if the output is not damaged because software cursors are rendered during this step. We should add a way for compositors to detect when they don't need to call `wlr_output_swap_buffers`.

Software cursors need to be rendered again when they move or when their image is updated. Hardware cursors never need to be rendered.

A few possible solutions:
1. Add a `damaged` signal to `wlr_output`, compositors can listen to it and schedule a `swap_buffers` call
2. Add a `bool damaged` field to `wlr_output`, compositors can check it before calling `swap_buffers`
3. Add a `bool wlr_output_render_cursors` function that needs to be called by the compositor to render software cursors, and returns `true` if the output becomes damaged. This has the upside of removing software cursor rendering from `swap_buffers` (and let the compositor do it whenever it wants to), but also requires compositors to do one more thing.

Related to #190.

emersion:
Fixes #375

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -1,9 +1,9 @@
 #ifndef WLR_TYPES_WLR_OUTPUT_H
 #define WLR_TYPES_WLR_OUTPUT_H
 
+#include <stdbool.h>
 #include <wayland-util.h>
 #include <wayland-server.h>
-#include <stdbool.h>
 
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Thanks!

PR #378, opened 2017-11-01
Rootston: flush close event for xwayland

martinetd:
Sorry for bulking in the indentation fixes again, this was ready, is just one line, and I really cba making a separate PR.


SirCmpwn:
Can you explain the first commit a bit? What's it fixing?

martinetd:
Hit the 'close' hotkey on any xwayland window like urxvt, it won't close until an xcb event happens. For me it was closing when I hit enter because PS1 would set title and trigger it, but setting PS1 to "$ " and it would just never close.

With the xcb_flush, the window is closed right away.

SirCmpwn:
Gotcha, thanks.

PR #379, opened 2017-11-01
Shutdown ASAN fixes

martinetd:
A couple more shutdown fixes, and no more obvious leak I could see. Good job!


SirCmpwn:
Thank you very much!

PR #380, opened 2017-11-01
Set xwayland root cursor

### LINKED ISSUE #369: Cross cursors in GTK on XWayland
screenshot from 2017-11-01 17-12-21

Over the textarea, the carret cursor is used correctly.

emersion:
Fixes #369


emersion:
This just sets the root cursor (I believe this is the cursor used by apps by default). XWayland calls `set_cursor` in the seat when apps want to set a pointer's cursor image, so we don't need to support that here, it's handled directly by XWayland.

emersion:
This PR is ready for review.

SirCmpwn:
I'm impressed by how complicated this is. 100 points awarded to X11.

PR #381, opened 2017-11-02
Fix disabled cursors

emersion:
When a client requests to disable a cursor, the last compositor cursor is displayed instead.

This PR also allows compositors to hide cursors by calling `wlr_output_cursor_set_image` with a NULL buffer.

Test plan: open `weston-terminal`, start typing, the cursor should disappear. Move the cursor, it should appear again.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -504,6 +504,12 @@ void wlr_output_cursor_destroy(struct wlr_output_cursor *cursor) {
 		}
 		cursor->output->hardware_cursor = NULL;
 	}
+	if (cursor->texture != NULL) {
+		wlr_texture_destroy(cursor->texture);
+	}
+	if (cursor->renderer != NULL) {
+		wlr_renderer_destroy(cursor->renderer);
+	}
 	wl_list_remove(&cursor->link);
 	free(cursor);
 }
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -253,10 +253,6 @@ static void output_cursor_render(struct wlr_output_cursor *cursor) {
 	struct wlr_texture *texture = cursor->texture;
 	struct wlr_renderer *renderer = cursor->renderer;
 	if (cursor->surface != NULL) {
-		// Some clients commit a cursor surface with a NULL buffer to hide it.
-		if (!wlr_surface_has_buffer(cursor->surface)) {
-			return;
-		}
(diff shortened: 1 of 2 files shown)


types/wlr_output.c
```diff
@@ -454,6 +453,10 @@ void wlr_output_cursor_set_surface(struct wlr_output_cursor *cursor,
 		wl_signal_add(&surface->events.destroy, &cursor->surface_destroy);
```
acrisci: If I set the cursor to a different surface than the current one, do we now have listeners on both the surfaces?

types/wlr_output.c
```diff
@@ -454,6 +453,10 @@ void wlr_output_cursor_set_surface(struct wlr_output_cursor *cursor,
...
 		output_cursor_commit(cursor);
```
acrisci: Probably should check if you set the cursor with a null buffer and handle that situation.

types/wlr_output.c
```diff
@@ -454,6 +453,10 @@ void wlr_output_cursor_set_surface(struct wlr_output_cursor *cursor,
...
 	} else {
+		cursor->enabled = false;
+		cursor->width = 0;
+		cursor->height = 0;
```
acrisci: Why are you resetting the width/height? Are these things ever read if the cursor is not enabled?

acrisci:
Why a separate flag? Can't you check if the cursor is enabled whenever you need to know this (invalid texture or no surface buffer) instead of keeping up with the extra state?

types/wlr_output.c
emersion: `wlr_output_cursor_reset` removes the listeners.

types/wlr_output.c
emersion: That's handled in `output_cursor_commit`.

emersion:
>Can't you check if the cursor is enabled whenever you need to know this (invalid texture or no surface buffer) instead of keeping up with the extra state?

That doesn't work with cursors without a surface. These cursors have a renderer and a texture instead, and we create them only once. Also consider the case of a cursor with a renderer and a texture and a call to `wlr_output_cursor_set_surface` with a NULL surface that disables the cursor.

types/wlr_output.c
emersion: Yeah, that's doesn't matter a lot. It's just to avoid having old state data.

PR #384, opened 2017-11-03
multiseat

### LINKED ISSUE #350: Multiseat support in rootston
The [wl-seat](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_seat) is a global that represents a group of input devices.

> A seat is a group of keyboards, pointer and touch devices. This object is published as a global during start up, or when such a device is hot plugged. A seat typically has a pointer and maintains a keyboard focus and a pointer focus.

There can be more than one seat global published on the server that window clients can bind to to receive events for the seat.

If this sounds appealing to you and you would like to do something with it, I would like to hear you explain your use case in this issue.

Example:

The design metaphor goes like this. Imagine there are two seats in front of your computer with people sitting in them. Each person has their own keyboard and mouse. Multiseat allows both the people to use their devices independently of each other. For example, we can both type into different windows at the same time (multiple keyboard focus) and control two cursors that appear in different locations on the screen and have a different window focused.

Some compositors may choose not to support multiseat because it is a niche use case and simplifying things will make their code more understandable.

Required:

* Associate input devices with seats (by name) in the rootston config
* Refactor all input-related functions (such as interactive resize/move) with seats
* Remove the assumption of a single focused window
* Test all existing wlroots interfaces that use the seat interface to make sure they work with multiseat

acrisci:
fixes #350


acrisci:
A lot of stuff is still broken, but can we merge this anyway? I'm rewriting a lot of stuff and the hand merges are getting really difficult to do.

SirCmpwn:
Multiseat stuff is broken or existing functionality is broken?

acrisci:
All of the xisting functionality should be there, but there were some bugs introduced into master recently that broke some things that should be addressed separately.

SirCmpwn:
Do you have a test plan for this PR? Should it still be automatically closing #350?

acrisci:
Set the seats in your config on the devices for the second seat:

```ini
[device:Logitech M510]
seat=seat1

[device:KB Multimedia keyboard ]
seat=seat1
```

Now observe multiseat things.

You'll have two cursors that are controlled independently of each other. Each one should have its own keyboard and pointer focus.

SirCmpwn:
Thanks. If this is a WIP, then what's done, what remains to be done, known bugs, etc?

emersion:
Force move/resize/rotate using the meta key seems to be broken.

types/wlr_keyboard.c
```diff
@@ -41,8 +41,6 @@ static void keyboard_modifier_update(struct wlr_keyboard *keyboard) {
 	keyboard->modifiers.latched = latched;
 	keyboard->modifiers.locked = locked;
 	keyboard->modifiers.group = group;
-
-	wl_signal_emit(&keyboard->events.modifiers, keyboard);
```
emersion: Why did you move the `emit` call outside this function? The X11 backend is still not calling `wlr_keyboard_notify_modifiers` so emiting the event here is necessary. Plus, the purpose of this function is to only emit the event when needed (when modifiers really changed).

include/wlr/types/wlr_keyboard.h
```diff
@@ -66,12 +66,23 @@ enum wlr_key_state {
 };
 
 struct wlr_event_keyboard_key {
+	struct wlr_input_device *device;
+	struct wlr_keyboard *keyboard;
```
emersion: Can't we get `device` from `keyboard.device`?

backend/libinput/keyboard.c
```diff
@@ -54,6 +54,7 @@ void handle_keyboard_key(struct libinput_event *event,
 	struct libinput_event_keyboard *kbevent =
 		libinput_event_get_keyboard_event(event);
 	struct wlr_event_keyboard_key wlr_event = { 0 };
+	wlr_event.device = wlr_dev;
```
emersion: `wlr_event.keyboard` is undefined.

rootston/keyboard.c
```diff
@@ -237,27 +233,19 @@ static void keyboard_config_merge(struct keyboard_config *config,
...
-	keyboard->modifiers.notify = keyboard_modifiers_notify;
-	wl_signal_add(&device->keyboard->events.modifiers, &keyboard->modifiers);
-
-	wl_list_insert(&input->keyboards, &keyboard->link);
-
-	struct keyboard_config config;
-	memset(&config, 0, sizeof(config));
-	keyboard_config_merge(&config, config_get_keyboard(input->config, device));
-	keyboard_config_merge(&config, config_get_keyboard(input->config, NULL));
+	struct keyboard_config *config = calloc(1, sizeof(struct keyboard_config));
```
emersion: Needs a `NULL` check

types/wlr_keyboard.c
acrisci: I still believe we should have the event, but I'm going to undo the change here to clean up the pr and discuss it as a separate issue.

acrisci:
> Force move/resize/rotate using the meta key seems to be broken.

fixed

acrisci:
*Executive summary*

The roots-seat implementation here comes much closer to the "collection of input devices" that the interface is meant to be, rather than a simple event router like wlr-seat.

roots-seat is a container of devices and keeps them in a list. It takes the raw input device from roots-input, wraps it in a roots struct containing user data based on configuration, hooks up the events, and then passes incoming events to the correct handler (see `cursor.c` and `keyboard.c`).

Most of the input state from roots-input has been moved to roots-seat. roots-input is now more like an input manager. It listens for new devices, and creates a seat from the device seat configuration (default: seat0). roots-seats currently live for the duration of the compositor (but should probably be destroyed once they are empty of devices). Functions that act on the state of all seats in the aggregate belong on roots-input.

Each roots-seat has one cursor. As devices are added to the roots-seat, the devices are added to the cursor. The cursor is wrapped in a roots_cursor containing the state of the cursor. This struct should be private to the cursor and not accessed by the rest of the code (todo). Notice how `cursor.c` only contains handlers meant to be called by the seat. Cursor state should be changed by the seat functions only.

roots-keyboard is analagous to the cursor except that there are zero to many of them. Keyboard handlers live in `keyboard.c` as well as basic setup/destruction of keyboards.

Input events are handled either by doing some compositor action (like bindings or meta-key moves) or routing the event to the wlr-seat to be consumed by windows.

Starting an interactive view/resize/rotate is now a seat function. Multiple seats can have interactive moves for different views at the same time.

Each seat has its own focused view. This view should be presented as activated and have keyboard focus for the seat. Setting a view to be focused is now a seat function.

Each seat has a list of drag icons that can be rendered to the screen during a dnd operation.

SirCmpwn:
Executive summary looks good. +1 to that design

>Thanks. If this is a WIP, then what's done, what remains to be done, known bugs, etc?

SirCmpwn:
The answer is probably "it's not my problem" but I may as well ask: how can we disambiguate between multiple versions of the same input device (same manufacturer and model)?

SirCmpwn:
Test results: mostly worked. The main bug I noticed was that only one view would be activated even if different seats were focused on different things.

acrisci:
> Test results: mostly worked. The main bug I noticed was that only one view would be activated even if different seats were focused on different things.

That's never going to work for xwayland because it can only have one window focused at a time.

I tested it with xdg-shell and it worked correctly.

SirCmpwn:
I was testing with xdg-shell. gnome-terminal.

acrisci:
>Thanks. If this is a WIP, then what's done, what remains to be done, known bugs, etc?

View teardown needs to be redone. Now that we have references to views in multiple places, we should use a signal instead to keep the responsibility of cleaning up the view on the component that uses it (instead of a teardown function that affects the state of the whole compositor).

Libinput keyboard doesn't work and there are drm issues, but not related to this change.

I've gotten windows to crash before by doing things with popups but I don't have detailed STR for that yet.

General QA work needs to be done to find the bugs and regressions.

SirCmpwn:
>Libinput keyboard doesn't work and there are drm issues, but not related to this change.

Can you elaborate? I was using libinput when I tested and it seemed to work alright (save the bugs I already mentioned).

acrisci:
Oh nevermind, this was just some weirdness with how I'm starting rootson where it doesn't give keyboard focus to the first window because of some race condition. That's a subtle bug I need to work out.

SirCmpwn:
Thanks! Please make tickets for the other issues.

PR #385, opened 2017-11-03
Fix moving views when resizing below min size

### LINKED ISSUE #357: Resizing below min size moves window with the mouse
Does not affect the lower-right corner. All shells are affected.

The lower-right corner is the right behavior. It should just stop resizing and not move around.

emersion:
Fixes #357

### DIFF
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -73,15 +73,19 @@ struct roots_view {
 	// elsewhere
 	void (*get_size)(struct roots_view *view, struct wlr_box *box);
 	void (*activate)(struct roots_view *view, bool active);
+	void (*move)(struct roots_view *view, double x, double y);
 	void (*resize)(struct roots_view *view, uint32_t width, uint32_t height);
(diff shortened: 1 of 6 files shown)


emersion:
This PR is ready for review. Note that resizing from the top-left corner is still de-synchronized (first the window moves, then it is resized), I'll fix that in another PR.

SirCmpwn:
Test plan?

emersion:
Oh sorry, forgot to add it.

Test plan: resize a window from the top-left corner below the minimum size (and beyond). It shouldn't move.

`weston-terminal` doesn't set a min size, so that doesn't work. That'll be fixed with the other PR I referred to in my previous comment.

SirCmpwn:
What's a good client to test with (that has a minimum size)?

emersion:
Resizable GTK apps under XWayland and Wayland are good test cases (I used `gnome-dictionary`).

SirCmpwn:
Thanks!

PR #387, opened 2017-11-05
Fix laggy move-resize

### LINKED ISSUE #386: Resizing from the top-left corner is laggy
Try to resize from the top-left corner, the view is moved before being resized by the client.

To fix this in xdg-shell, we should wait for an `ack_configure` event from the client. To do this, we need to save the position we want to move the surface to when sending the `configure` event, and read it when receiving `ack_configure`.

So far, I've tried two ways of doing it:

1. Adding a list of configure request states in rootston. However, configure requests are scheduled, so after calling `wlr_xdg_toplevel_v6_set_size` there's no way of guessing the serial that will be used when sending the configure request. This could be fixed by assigning the serial when calling `wlr_xdg_toplevel_v6_set_size` (if not already assigned).
2. Adding a `void *data` field to `wlr_xdg_toplevel_v6_state`. This is appealing because it simplifies a lot rootston (no need to maintain a list of configure requests since wlroots already does it). However, I'm not sure about memory management. Since the client is required to acknowledge only the last configure request, some can be discarded.

@acrisci Thoughts?

emersion:
~~This is a pretty naive fix that waits for surface commits to move the view. It works with clients which commit a surface after configure requests and are fast enough (commits are sent faster than configure requests). This may be good enough.~~

Do we want to do this for all shells? ~~We could pretty easily because it's not using any shell-specific feature, just the surface commit.~~

- [x] `weston-dnd`: resizing from top-left corner, moving the mouse past the bottom-right border makes the window move
- [x] `gnome-calendar` (and all apps, but it's less noticeable): one invalid frame
  * if I drag top left out to make it bigger, I get a flash of either a greater x or y coordinate for a frame
  * if I drag top left in to make it smaller, I get a flash of either a smaller x or y coordinate for a frame

Fixes #386


acrisci:
It doesn't work with the top-right corner or the bottom-left corner.

emersion:
Yeah, that's a known issue. I'll fix that.

emersion:
I think this is ready for review for the xdg-shell side. I'd like to hear what you think before doing it in other shells.

First, this still expects the surface to react to configure requests quicker than they are sent. If the surface doesn't do it, it won't move until the user stops resizing it. All clients I tested this on had no issues with that. Since making this more reliable will add a lot of complexity to rootston (requiring adding a list of pending configure request per shell), I'd like to try the naive approach and change it if we get bug reports.

This approach saves only the last configure request. I've updated xdg-shell to generate and return the configure serial when it's scheduled. When the client reconfigures the surface, it sends an `ack_configure` before the next `commit`. On commit, the view position is updated if the configure requested that triggered the view resize has been acknowledged. I've added a `wlr_xdg_surface_v6.configure_serial` field that contains the last acknowledged configure request serial. I've removed `wlr_xdg_surface_v6.events.ack_configure` since compositors should use `commit` instead.

acrisci:
You can force qt to use wl_shell like this:

```diff
diff --git a/rootston/desktop.c b/rootston/desktop.c
index b3287ed..c6f9360 100644
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -364,10 +364,12 @@ struct roots_desktop *desktop_create(struct roots_server *server,
        desktop->compositor = wlr_compositor_create(server->wl_display,
                server->renderer);
 
+       /*
        desktop->xdg_shell_v6 = wlr_xdg_shell_v6_create(server->wl_display);
        wl_signal_add(&desktop->xdg_shell_v6->events.new_surface,
                &desktop->xdg_shell_v6_surface);
        desktop->xdg_shell_v6_surface.notify = handle_xdg_shell_v6_surface;
+       */
 
        desktop->wl_shell = wlr_wl_shell_create(server->wl_display);
        wl_signal_add(&desktop->wl_shell->events.new_surface,
```

emersion:
I think this is ready to go.

`qt-creator -platform wayland` with wl-shell has some issues, but I don't think we can do anything about it. When we send two configure requests, we then receive two commits, but we move the view only when the first commit is done (not when the second one happens). We cannot keep track of the number of pending configure requests because of this:

>The client is free to dismiss all but the last configure event it received.

SirCmpwn:
LGTM, thanks!

PR #391, opened 2017-11-05
Send current keyboard state when entering a surface

### LINKED ISSUE #389: Changing focused surface does not update modifiers.
When the focused view (in rootston) is changed while a modifier is held, e.g. close a shell+terminal with `Ctrl+d` and focus changes to previous view, the modifiers are not updated.

The easiest way to fix this is probably to send the current modifiers when the `wlr_seat` focus is changed. Either from the compositor or from inside wlroots itself.

emersion:
- [x] Should we read the current keyboard state like this? The compositor might want to filter some keys, and we might end up in an inconsistent state if the compositor is simulating keypresses. Would it be better to take the state as a parameter to `wlr_seat_keyboard_enter`?
- [ ] ~~Remove list of pressed keysyms in rootston, use list of pressed keycodes in `wlr_keyboard`?~~ Can't fix easily after #394

Test plan: open two terminals. Press two times <kbd>Ctrl</kbd> + <kbd>D</kbd> without releasing <kbd>Ctrl</kbd>. This should close both terminals.

Fixes #389

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -45,9 +45,7 @@ struct wlr_keyboard_grab_interface {
 			struct wlr_surface *surface);
 	void (*key)(struct wlr_seat_keyboard_grab *grab, uint32_t time,
 			uint32_t key, uint32_t state);
-	void (*modifiers)(struct wlr_seat_keyboard_grab *grab,
-			uint32_t mods_depressed, uint32_t mods_latched,
-			uint32_t mods_locked, uint32_t group);
+	void (*modifiers)(struct wlr_seat_keyboard_grab *grab);
 	void (*cancel)(struct wlr_seat_keyboard_grab *grab);
 };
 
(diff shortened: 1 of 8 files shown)


emersion:
This PR is ready for review.

include/wlr/types/wlr_seat.h
```diff
@@ -42,12 +42,14 @@ struct wlr_seat_keyboard_grab;
 
 struct wlr_keyboard_grab_interface {
 	void (*enter)(struct wlr_seat_keyboard_grab *grab,
-			struct wlr_surface *surface);
+		struct wlr_surface *surface, uint32_t *keycodes, size_t keycodes_len,
+		uint32_t mods_depressed, uint32_t mods_latched, uint32_t mods_locked,
+		uint32_t group);
```
acrisci: You shouldn't need these arguments here. They should just be able to read the modifiers off the keyboard from the seat.

emersion: >Should we read the current keyboard state like this? The compositor might want to filter some keys, and we might end up in an inconsistent state if the compositor is simulating keypresses. Would it be better to take the state as a parameter to wlr_seat_keyboard_enter?

emersion: If we read modifiers from the `wlr_keyboard` here, then `wlr_seat_notify_modifiers` should also read modifiers from the `wlr_keyboard`.

acrisci: I don't think the overhead of having rootston reading modifier state anytime it wants to set keyboard focus is worth it. I think we can add that feature in with a special interface once we have a concrete use case for it. It's difficult to design it in the abstract when we will probably find a better solution later.

emersion: So I'll change both functions and remove modifier arguments.

What about the keyboard keys? Do we want the seat to read them directly from the keyboard, like modifiers?

acrisci: Yeah I think it's better to have a single source of truth on the keyboard for keyboard state for now and evaluate any other use case separately.

emersion: Compositors can also implement a custom keyboard interface if they want to simulate keypresses.

emersion:
Should be good to go.

SirCmpwn:
Thanks!

SirCmpwn:
Thanks!

PR #394, opened 2017-11-07
Fix #393 

### LINKED ISSUE #393: rootston: xkb/keyboard handling for keybinds is wonky to broken.
The keyboard handling in rootston is a bit weird.

I added my own debug output which can be seen at [xkb_debug](https://github.com/Ongy/wlroots/tree/xkb_debug).

When I press `[Alt]+[Shift]+2` I get the output:
```
2017-11-07 09:37:35 - [rootston/keyboard.c:94] Tring to find keybind with mods: 8
2017-11-07 09:37:35 - [rootston/keyboard.c:100] With pressed key: Alt_L
2017-11-07 09:37:35 - [rootston/keyboard.c:94] Tring to find keybind with mods: 9
2017-11-07 09:37:35 - [rootston/keyboard.c:100] With pressed key: Alt_L
2017-11-07 09:37:35 - [rootston/keyboard.c:100] With pressed key: Shift_L
2017-11-07 09:37:36 - [rootston/keyboard.c:94] Tring to find keybind with mods: 9
2017-11-07 09:37:36 - [rootston/keyboard.c:100] With pressed key: Alt_L
2017-11-07 09:37:36 - [rootston/keyboard.c:100] With pressed key: Shift_L
2017-11-07 09:37:36 - [rootston/keyboard.c:100] With pressed key: at
```

The mods here are the modifiers used. 8 is Alt, 1 is Shift, and they are xored, so 9 is both.

This prevents binding to the example key combiniation I pressed, since there's never a `2` pressed for the keybinds.
A workaround could be to bind to `[Alt]+[Shift]+[at]`. But from how I understand [xkb docs](https://xkbcommon.org/doc/current/group__state.html#ga30061a16ebaccb7b34bb1ce7d106b001) consumed modifiers should be removed before binding, so we could bind `[Alt]+[at]`.

Another way to go about this, would be to use the modifiers kept in `wlr_keyboard` and the event keycode directly.
The main issue with this I can see is that e.g. the [at] here could not be used (or more serious the SWITCH_VT would have to be hardcoded).

Maybe the best approach would be to go through both approaches to figure out what to match.

**Edit:** I tested this with rootston and waymonad on X backend and waymonad on DRM. My code in waymonad is mostly a translation of the rootston code into Haskell, so I doubt rootston behaves any different on the other backend.

Ongy:
As mentioned in https://github.com/swaywm/wlroots/issues/393 keybinds
did't trigger / were checked with "odd" keys and modifiers.

This commit sends the keycode through two paths, one to get the keycode
and modifiers *after* xkb handles them, a secondary path to get a "raw"
keysym without modifiers and then add the modifiers rootston knows
about.

This will result in the `[Alt]+[Shift]+2` combination I mention earlier
going through the keybind detection twice.
  1) `[Alt]+[at]`
  2) `[Alt]+[Shift]+2`

When either combination is found, the appropriate keybind is executed.
The xkb handled version will be prefered over the "raw" version.

**Edit: oh right, test-plan**
Bind something to `[Alt]+[Shift]+2` and check whether the keybind is found.

### DIFF
--- a/rootston/keyboard.c
+++ b/rootston/keyboard.c
@@ -60,7 +60,7 @@ static void keyboard_binding_execute(struct roots_keyboard *keyboard,
  * should be propagated to clients.
  */
 static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
-		xkb_keysym_t keysym) {
+		xkb_keysym_t keysym, uint32_t modifiers) {
 	ssize_t i = keyboard_pressed_keysym_index(keyboard, keysym);
 	if (i < 0) {
 		i = keyboard_pressed_keysym_index(keyboard, XKB_KEY_NoSymbol);
@@ -88,7 +88,6 @@ static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
 		wlr_seat_keyboard_end_grab(keyboard->input->wl_seat);
 	}
 
-	uint32_t modifiers = wlr_keyboard_get_modifiers(keyboard->device->keyboard);
 	struct wl_list *bindings = &keyboard->input->server->config->bindings;
 	struct binding_config *bc;
 	wl_list_for_each(bc, bindings, link) {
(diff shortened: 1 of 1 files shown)


emersion:
I would prefer leaving `wlr_keyboard` as is and do all changes in rootston.

Ongy:
The only file this changes `rootston/keyboard.c`. There's no need to touch `wlr_keyboard` for this.

Maybe I worded things a bit badly. Both paths go branch off in the key handler in rootston.
It's just that one gets the keysym over the full `xkb_state` while the other goes over `xkb_keymap` (from wlr_keyboard->keymap) and only uses the state to get the `xkb_layout_index_t` from the state.

Btw. IMO this is ready for review.

rootston/keyboard.c
```diff
@@ -122,25 +121,84 @@ static void keyboard_keysym_release(struct roots_keyboard *keyboard,
...
+ * Process keypresses from the keyboard as if modifiers didn't change keysyms.
+ *
+ * This avoids the xkb keysym translation based on modifiers considered pressed
+ * in the state and uses the list of modifiers saved on the rootston side.
+ *
+ * This will trigger the keybind: [Alt]+[Shift]+2
+ */
+static bool keyboard_keysyms_simple(struct roots_keyboard *keyboard,
+		uint32_t keycode, enum wlr_key_state state)
+{
```
SirCmpwn: Style - put this brace on the previous line

rootston/keyboard.c
```diff
@@ -122,25 +121,84 @@ static void keyboard_keysym_release(struct roots_keyboard *keyboard,
...
+ *
+ * This uses the xkb keysyms translation based on pressed modifiers and clears
+ * the consumed modifiers from the list of modifiers passed to keybind
+ * detection.
+ *
+ * (On US layout) this will trigger: [Alt]+[at]
+ */
+static bool keyboard_keysyms_xkb(struct roots_keyboard *keyboard,
+		uint32_t keycode, enum wlr_key_state state)
+{
```
SirCmpwn: Ditto

rootston/keyboard.c
```diff
@@ -122,25 +121,84 @@ static void keyboard_keysym_release(struct roots_keyboard *keyboard,
...
+			handled = handled || keysym_handled;
+		} else { // WLR_KEY_RELEASED
+			keyboard_keysym_release(keyboard, syms[i]);
+		}
+	}
+
+	return handled;
+}
+
+/**
```
SirCmpwn: Don't use `/**` on internal functions

rootston/keyboard.c
```diff
@@ -122,25 +121,84 @@ static void keyboard_keysym_release(struct roots_keyboard *keyboard,
...
+
+static void keyboard_key_notify(struct wl_listener *listener, void *data) {
+	struct wlr_event_keyboard_key *event = data;
+	struct roots_keyboard *keyboard = wl_container_of(listener, keyboard, key);
+
+	uint32_t keycode = event->keycode + 8;
+
+	bool handled = keyboard_keysyms_xkb(keyboard, keycode, event->state);
+
+	if (!handled)
```
SirCmpwn: Add braces

SirCmpwn: Actually why can't this just be moved into the next, identical if statement?

SirCmpwn:
LGTM

SirCmpwn:
Thanks!

PR #396, opened 2017-11-08
Remove unused include to improve C++ interop

emolitor:
It appears that session.h is not intended to be exposed externally. This change removes unused session.h includes from the files below. This also improves interoperability as it avoids exposing the C99 array qualifier in session.h which is not compatible with C++.


emolitor:
Ok, I need to do a bit more refactoring as this just allows naive examples to work.

PR #405, opened 2017-11-10
Fix typos

rgnstf2712:


ascent12:
Thanks!
How on earth did I not notice something like that for so long?

PR #408, opened 2017-11-11
Fix resize issues with some xdg-shell apps.

### LINKED ISSUE #407: Resizing first gives a 0x0 size to some views
Try resizing a rootston window in rootston.

Also seems to happen when running HEAD mpv.

emersion:
Also renames view_get_size to view_get_box to be more consistent with wlroots API and make it more obvious that x,y are set (while in roots_view.get_size this isn't the case).

Fixes #407


SirCmpwn:
Thanks!

PR #412, opened 2017-11-11
Scale cursors on scaled outputs

### LINKED ISSUE #401: Scale rootston cursor on scaled outputs

emersion:
- [x] Keep one xcursor theme per seat
- [x] Integrate `roots_xcursor_theme` into wlroots

Test plan: check that compositor cursors (default, move, resize, rotate) still work with different output scales.

Fixes #401


examples/multi-pointer.c
```diff
@@ -86,7 +86,7 @@ static void handle_output_add(struct output_state *ostate) {
 
 		struct wlr_xcursor_image *image = sample->xcursor->images[0];
 		wlr_cursor_set_image(cursor->cursor, image->buffer, image->width,
-			image->width, image->height, image->hotspot_x, image->hotspot_y);
+			image->width, image->height, image->hotspot_x, image->hotspot_y, 0);
```
SirCmpwn: Scale factor of 0? What?

acrisci:
I want cursor theme per seat. Is that possible?

rootston/xcursor.c
```diff
@@ -1,6 +1,85 @@
-#include <wlr/types/wlr_cursor.h>
+#define _POSIX_C_SOURCE 200809L
```
acrisci: How much of this can you bring into wlroots?

rootston/output.c
```diff
@@ -240,6 +241,11 @@ void output_add_notify(struct wl_listener *listener, void *data) {
 		wlr_output_layout_add_auto(desktop->layout, wlr_output);
 	}
 
+	if (roots_xcursor_theme_load(desktop->xcursor_theme, wlr_output->scale)) {
+		wlr_log(L_ERROR, "Cannot load xcursor theme with scale %d",
```
acrisci: This error message makes it sound like it failed because the scale is invalid while I think it can fail for other reasons.

rootston/desktop.c
```diff
@@ -344,6 +352,12 @@ struct roots_desktop *desktop_create(struct roots_server *server,
 		wl_signal_add(&desktop->xwayland->events.new_surface,
 			&desktop->xwayland_surface);
 		desktop->xwayland_surface.notify = handle_xwayland_surface;
+
+		if (roots_xcursor_theme_load(desktop->xcursor_theme, 1)) {
+			wlr_log(L_ERROR, "Cannot load xwayland xcursor theme");
+		}
+		roots_xcursor_theme_xwayland_set_default(desktop->xcursor_theme,
+			desktop->xwayland);
```
acrisci: Here it does make sense to have a global xcursor theme for the desktop.

rootston/desktop.c
```diff
@@ -323,6 +323,14 @@ struct roots_desktop *desktop_create(struct roots_server *server,
 
 	desktop->server = server;
 	desktop->config = config;
+
+	desktop->xcursor_theme = roots_xcursor_theme_create("default");
+	if (desktop->xcursor_theme == NULL) {
+		wlr_list_free(desktop->views);
+		free(desktop);
+		return NULL;
```
acrisci: Does it make sense for this to be fatal? Could there be an embedded application that does not use xcursor themes at all? For instance, devices without a pointer (like a touch-screen kiosk) will never show a pointer.

I think I'm ok with this assumption since it simplifies the code and rootston is assumed to have desktop features, but we should always make sure xcursor is not required to be used in the library.

examples/multi-pointer.c
emersion: https://github.com/swaywm/wlroots/pull/412/files#diff-69dfd47b5b7a981f4364342fc4a54ed0R67

This basically allows compositors to do not care about cursor scaling if they don't want to.

emersion:
>I want cursor theme per seat. Is that possible?

That would make things more complicated because `roots_xcursor_theme` has to listen for outputs to load scaled xcursor themes. But I guess I can make the desktop loop through all xcursor themes and load new scaled cursors when a new output is plugged in.

I prefer to listen for new outputs instead of reading output scales from the config because:
* It allows to lazy load scaled xcursor themes (if an output is setup with a scale of 2 and is never plugged in, xcursor theme with scale of 2 is never loaded)
* It will allow to autodetect the scale of an output, if not specified in the config file

rootston/desktop.c
emersion: Yeah, wlroots doesn't require the use of wlr_xcursor at all, we should be fine.

rootston/xcursor.c
emersion: This would require to create another data structure alongside `wlr_xcursor_theme` and `wlr_xcursor`.

Do we want this feature in wlroots? I'm still not sure about this. If we do it, I would prefer a separate type (`wlr_xcursor_manager`?) that would bridge the gap between `wlr_xcursor` and `wlr_cursor`, to make sure to keep these two separate. How does that sound?

rootston/desktop.c
SirCmpwn: Rootston isn't a kiosk-ready compositor, it's going to have a cursor (at least for now).

rootston/xcursor.c
SirCmpwn: +1 to put it in wlroots. I like that name, too.

examples/multi-pointer.c
acrisci: Normally this is indicated by -1.

emersion: Since it's an `uint32_t` and since zero hasn't any meaning for a scale, I thought it would be okay. I can change it to an `int32_t` though.

rootston/xcursor.c
acrisci: I think we should try to make the interface to rootston as simple as possible.

examples/multi-pointer.c
SirCmpwn: I don't care either way.

emersion:
I added back `roots_seat.xcursor_theme` and the seat now uses this theme for its cursor. However, there's still a global theme (used for xwayland), and currently the global theme is used in all seats. What it means is that for now it's still not possible to configure per-seat themes, but it's just a matter of setting `roots_seat.xcursor_theme` when initializing the seat.

emersion:
This PR is ready for another review.

The last commit introduces `wlr_xcursor_manager`, which manages multiple XCursor themes loaded with different scale factors. Note that `wlr_xcursor` and `wlr_cursor` can be used without `wlr_xcursor_manager`, and `wlr_xcursor_manager` can be used without `wlr_cursor`.

rootston/desktop.c
```diff
@@ -353,11 +356,17 @@ struct roots_desktop *desktop_create(struct roots_server *server,
...
+		if (wlr_xcursor_manager_load(desktop->xcursor_manager, 1)) {
+			wlr_log(L_ERROR, "Cannot load XWayland XCursor theme");
+		}
+		struct wlr_xcursor *xcursor = wlr_xcursor_manager_get_xcursor(
+			desktop->xcursor_manager, ROOTS_XCURSOR_DEFAULT, 1);
+		if (xcursor != NULL) {
+			struct wlr_xcursor_image *image = xcursor->images[0];
+			wlr_xwayland_set_cursor(desktop->xwayland, image->buffer,
+				image->width, image->width, image->height, image->hotspot_x,
+				image->hotspot_y);
```
SirCmpwn: I wonder if it makes sense to change the xwayland default cursor depending on which output the cursor is on.

SirCmpwn:
I like it. Will merge after testing.

rootston/desktop.c
emersion: Well, there's not always just one cursor - and the XWayland cursor applies to all outputs, so I don't think this could work.

SirCmpwn: Yeah it would only work in a one-cursor context but that's also the most common context we'll be running in

SirCmpwn:
https://travis-ci.org/swaywm/wlroots/jobs/300891992

SirCmpwn:
I don't get any cursor at all on my hidpi output.

emersion:
>I don't get any cursor at all on my hidpi output.

Fixed!

SirCmpwn:
Thanks!

rootston/desktop.c
SirCmpwn: I'd like to do this, by the way - spin it out into a separate ticket please.

PR #414, opened 2017-11-12
Fix rotated views on HiDPI outputs

### LINKED ISSUE #402: Rotated surfaces in rootston don't work correctly on scaled outputs

emersion:
Test plan: set output scale to 2, open `gnome-calculator`, rotate it, check that pointer events work (clicking on buttons, text inputs...). Check that views are rendered at the correct position (open menus), **except** the combobox. The combobox uses a popup and these don't work anymore on non-HiDPI outputs (hence, this is an unrelated regression, #415).

Fixes #402


SirCmpwn:
Thanks!

PR #416, opened 2017-11-12
rootston: set keyboard for seat on keyboard add

acrisci:
Since wlr-seat requires a keyboard to be set to send enter, this fixes a bug where the surface doesn't get keyboard focus if the view appears before any keyboard events have been received.


SirCmpwn:
Thanks!

PR #418, opened 2017-11-13
wlr-seat: touch

acrisci:
Proper touch support in wlroots.

Move touch point struct from rootston into wlr-seat which now keeps track of touch state.

Touch points are created on "down", moved on "motion", and destroyed on "up". They are bound to a single surface for their lifetime and are destroyed when the surface is destroyed.

The decision of what pointer-like features to give to the touch device belongs to rootston. The compositor can choose to give it a cursor, or press buttons on touch down if it so chooses.

Test with `weston-simple-touch`. Test dnd with `weston-dnd` (try moving across two windows).

You should be able to focus and move windows around with the cursor.

TODO:

* [ ] <s>rootston touch cursor features</s> move to another pr
* [ ] <s>xdg|wl-shell touch grab</s> probably not necessary
* [x] put drag icon in the right place on dnd
* [x] send dnd motion across surfaces
* [x] refactor touch grab interface

### DIFF
--- a/include/rootston/seat.h
+++ b/include/rootston/seat.h
@@ -26,6 +26,9 @@ struct roots_seat {
 	struct wl_list link;
 	struct wl_list drag_icons;
 
+	// coordinates of the touch grab if one exists
+	double touch_grab_x, touch_grab_y;
+
 	struct roots_view *focus;
 
 	struct wl_list keyboards;
--- a/include/rootston/seat.h
+++ b/include/rootston/seat.h
@@ -10,7 +10,7 @@ struct roots_drag_icon {
 	bool mapped;
 
 	bool is_pointer;
-	bool touch_id;
+	int32_t touch_id;
 
 	int32_t sx;
 	int32_t sy;
@@ -26,8 +26,9 @@ struct roots_seat {
 	struct wl_list link;
 	struct wl_list drag_icons;
 
-	// coordinates of the touch grab if one exists
-	double touch_grab_x, touch_grab_y;
+	// coordinates of the first touch point if it exists
+	int32_t touch_id;
+	double touch_x, touch_y;
 
 	struct roots_view *focus;
 
--- a/include/rootston/seat.h
+++ b/include/rootston/seat.h
@@ -9,6 +9,9 @@ struct roots_drag_icon {
 	struct wl_list link; // roots_seat::drag_icons
 	bool mapped;
 
+	bool is_pointer;
+	bool touch_id;
+
 	int32_t sx;
 	int32_t sy;
 
--- a/include/rootston/seat.h
+++ b/include/rootston/seat.h
@@ -43,13 +43,6 @@ struct roots_touch {
 	struct wl_list link;
 };
 
-struct roots_touch_point {
-	struct roots_touch *device;
-	int32_t slot;
-	double x, y;
-	struct wl_list link;
-};
-
 struct roots_tablet_tool {
 	struct roots_seat *seat;
 	struct wlr_input_device *device;
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -270,6 +270,11 @@ void roots_cursor_handle_touch_motion(struct roots_cursor *cursor,
 		wlr_seat_touch_point_clear_focus(cursor->seat->seat, event->time_msec,
 			event->slot);
 	}
+
+	if (wlr_seat_touch_has_grab(cursor->seat->seat)) {
+		cursor->seat->touch_grab_x = lx;
+		cursor->seat->touch_grab_y = ly;
+	}
 }
 
 void roots_cursor_handle_tool_axis(struct roots_cursor *cursor,
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -128,13 +128,14 @@ static void roots_cursor_update_position(struct roots_cursor *cursor, uint32_t t
 
 static void roots_cursor_press_button(struct roots_cursor *cursor,
 		struct wlr_input_device *device, uint32_t time, uint32_t button,
-		uint32_t state) {
+		uint32_t state, double lx, double ly) {
 	struct roots_seat *seat = cursor->seat;
 	struct roots_desktop *desktop = seat->input->server->desktop;
+	bool is_touch = device->type == WLR_INPUT_DEVICE_TOUCH;
+
 	struct wlr_surface *surface;
 	double sx, sy;
-	struct roots_view *view = view_at(desktop,
-		cursor->cursor->x, cursor->cursor->y, &surface, &sx, &sy);
+	struct roots_view *view = view_at(desktop, lx, ly, &surface, &sx, &sy);
 
 	if (state == WLR_BUTTON_PRESSED && view && roots_seat_has_meta_pressed(seat)) {
 		roots_seat_focus_view(seat, view);
@@ -165,14 +166,20 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 		return;
 	}
 
-	uint32_t serial =
-		wlr_seat_pointer_notify_button(seat->seat, time, button, state);
+	uint32_t serial;
+	if (is_touch) {
+		serial = wl_display_get_serial(desktop->server->wl_display);
+	} else {
+		serial = wlr_seat_pointer_notify_button(seat->seat, time, button, state);
+	}
 
 	int i;
 	switch (state) {
 	case WLR_BUTTON_RELEASED:
 		seat->cursor->mode = ROOTS_CURSOR_PASSTHROUGH;
-		roots_cursor_update_position(cursor, time);
+		if (!is_touch) {
+			roots_cursor_update_position(cursor, time);
+		}
 		break;
 	case WLR_BUTTON_PRESSED:
 		i = cursor->input_events_idx;
@@ -203,7 +210,7 @@ void roots_cursor_handle_motion_absolute(struct roots_cursor *cursor,
 void roots_cursor_handle_button(struct roots_cursor *cursor,
 		struct wlr_event_pointer_button *event) {
 	roots_cursor_press_button(cursor, event->device, event->time_msec,
-		event->button, event->state);
+		event->button, event->state, cursor->cursor->x, cursor->cursor->y);
 }
 
 void roots_cursor_handle_axis(struct roots_cursor *cursor,
@@ -231,12 +238,29 @@ void roots_cursor_handle_touch_down(struct roots_cursor *cursor,
 		wlr_seat_touch_notify_down(cursor->seat->seat, surface,
 			event->time_msec, event->slot, sx, sy);
 	}
+
+	if (wlr_seat_touch_num_points(cursor->seat->seat) == 1) {
+		cursor->seat->touch_id = event->slot;
+		cursor->seat->touch_x = lx;
+		cursor->seat->touch_y = ly;
+		roots_cursor_press_button(cursor, event->device, event->time_msec,
+			BTN_LEFT, 1, lx, ly);
+	}
 }
 
 void roots_cursor_handle_touch_up(struct roots_cursor *cursor,
 		struct wlr_event_touch_up *event) {
+	struct wlr_touch_point *point = wlr_seat_touch_get_point(cursor->seat->seat, event->slot);
+	if (!point) {
+		return;
+	}
+
+	if (wlr_seat_touch_num_points(cursor->seat->seat) == 1) {
+		roots_cursor_press_button(cursor, event->device, event->time_msec,
+			BTN_LEFT, 0, cursor->seat->touch_x, cursor->seat->touch_y);
+	}
+
 	wlr_seat_touch_notify_up(cursor->seat->seat, event->time_msec, event->slot);
-	//roots_cursor_press_button(cursor, event->device, event->time_msec, BTN_LEFT, 0);
 }
 
 void roots_cursor_handle_touch_motion(struct roots_cursor *cursor,
@@ -271,9 +295,9 @@ void roots_cursor_handle_touch_motion(struct roots_cursor *cursor,
 			event->slot);
 	}
 
-	if (wlr_seat_touch_has_grab(cursor->seat->seat)) {
-		cursor->seat->touch_grab_x = lx;
-		cursor->seat->touch_grab_y = ly;
+	if (event->slot == cursor->seat->touch_id) {
+		cursor->seat->touch_x = lx;
+		cursor->seat->touch_y = ly;
 	}
 }
 
@@ -298,7 +322,8 @@ void roots_cursor_handle_tool_axis(struct roots_cursor *cursor,
 void roots_cursor_handle_tool_tip(struct roots_cursor *cursor,
 		struct wlr_event_tablet_tool_tip *event) {
 	roots_cursor_press_button(cursor, event->device,
-		event->time_msec, BTN_LEFT, event->state);
+		event->time_msec, BTN_LEFT, event->state, cursor->cursor->x,
+		cursor->cursor->y);
 }
 
 void roots_cursor_handle_request_set_cursor(struct roots_cursor *cursor,
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -234,12 +234,13 @@ void roots_cursor_handle_touch_down(struct roots_cursor *cursor,
 	double sx, sy;
 	view_at(desktop, lx, ly, &surface, &sx, &sy);
 
+	uint32_t serial = 0;
 	if (surface) {
-		wlr_seat_touch_notify_down(cursor->seat->seat, surface,
+		serial = wlr_seat_touch_notify_down(cursor->seat->seat, surface,
 			event->time_msec, event->slot, sx, sy);
 	}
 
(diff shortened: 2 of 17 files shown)


acrisci:
Ok I think this is ready for review.

We don't have all the cursor features of the touch device (like moving windows around with the touch screen), but I think I'm going to wait for #387 and discuss this issue with @emersion in another issue once cursor.c settles down a bit.

types/wlr_data_device.c
```diff
@@ -587,13 +664,22 @@ static bool seat_client_start_drag(struct wlr_seat_client *client,
...
+	wlr_seat_keyboard_start_grab(drag->seat, &drag->keyboard_grab);
 
-	wlr_seat_keyboard_start_grab(seat, &drag->keyboard_grab);
-	wlr_seat_pointer_start_grab(seat, &drag->pointer_grab);
+	if (drag->is_pointer_grab) {
+		wlr_seat_pointer_clear_focus(drag->seat);
+		wlr_seat_pointer_start_grab(drag->seat, &drag->pointer_grab);
+	} else {
+		wlr_seat_touch_start_grab(drag->seat, &drag->touch_grab);
+		wlr_drag_set_focus(drag, point->surface, point->sx, point->sy);
```
emersion: `point` can be NULL here. Maybe check for `is_touch_grab`?

types/wlr_data_device.c
```diff
@@ -562,13 +614,38 @@ static void drag_handle_drag_source_destroy(struct wl_listener *listener,
...
+
+	bool is_touch_grab = client->touch &&
+		wl_list_length(&client->seat->touch_state.touch_points) == 1 &&
+		client->seat->touch_state.grab_serial == serial;
+
+	struct wlr_touch_point *point = NULL;
+	if (is_touch_grab) {
+		wl_list_for_each(point, &client->seat->touch_state.touch_points, link) {
+			if (point->surface && point->surface == origin) {
+				is_touch_grab = true;
```
emersion: `is_touch_grab` is already set to `true`

types/wlr_seat.c
```diff
@@ -819,3 +889,259 @@ void wlr_seat_keyboard_notify_key(struct wlr_seat *seat, uint32_t time,
...
+		wl_signal_emit(&wlr_seat->events.touch_grab_end, grab);
+		if (grab->interface->cancel) {
+			grab->interface->cancel(grab);
+		}
+	}
+}
+
+static void touch_point_destroy(struct wlr_touch_point *point) {
+	wl_signal_emit(&point->events.destroy, point);
+	wl_list_remove(&point->surface_destroy.link);
```
emersion: `wl_list_remove(&point->focus_surface_destroy.link);`

emersion: This is not always initialized, so we need to add `wl_list_init(&point->focus_surface_destroy.link)` in `create` and in `clear`.

types/wlr_seat.c
```diff
@@ -819,3 +889,259 @@ void wlr_seat_keyboard_notify_key(struct wlr_seat *seat, uint32_t time,
...
+		touch_point_create(seat, touch_id, surface, sx, sy);
+	if (!point) {
+		wlr_log(L_ERROR, "could not create touch point");
+		return;
+	}
+
+	grab->interface->down(grab, time, point);
+
+	if (wl_list_length(&seat->touch_state.touch_points) == 1) {
+		seat->touch_state.grab_serial = wl_display_get_serial(seat->display);
```
emersion: I've seen you're using `wl_display_get_serial` like this at several places. Is it really okay to do this? I'd prefer `grab->interface->down` to return the serial for instance.

types/wlr_seat.c
```diff
@@ -819,3 +889,259 @@ void wlr_seat_keyboard_notify_key(struct wlr_seat *seat, uint32_t time,
...
+	touch_point_clear_focus(point);
+
+	if (surface && surface->resource) {
+		struct wlr_seat_client *client =
+			wlr_seat_client_for_wl_client(point->client->seat,
+				wl_resource_get_client(surface->resource));
+
+		if (client && client->touch) {
+			wl_signal_add(&surface->events.destroy, &point->focus_surface_destroy);
+			point->focus_surface_destroy.notify = handle_point_focus_destroy;
```
emersion: This could be set in `create`, once.

include/rootston/seat.h
```diff
@@ -23,6 +26,10 @@ struct roots_seat {
 	struct wl_list link;
 	struct wl_list drag_icons;
 
+	// coordinates of the first touch point if it exists
+	int32_t touch_id;
+	double touch_x, touch_y;
```
emersion: What happens if there are multiple touch points? What are the consequences of only considering the first?

SirCmpwn:
1. Open weston-simple-touch
2. Touch two points at once

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7973123 in wl_list_insert () from /usr/lib/libwayland-server.so.0
(gdb) bt
#0  0x00007ffff7973123 in wl_list_insert () at /usr/lib/libwayland-server.so.0
#1  0x00007ffff7bb3158 in wl_signal_add (signal=0x555555a607b0, listener=0x555556205268) at /usr/include/wayland-server-core.h:351
#2  0x00007ffff7bb5916 in touch_point_create (seat=0x555556111760, touch_id=0, surface=0x555555a606f0, sx=127.04004884004883, sy=273.16190476190479)
    at ../types/wlr_seat.c:959
#3  0x00007ffff7bb5a54 in wlr_seat_touch_notify_down (seat=0x555556111760, surface=0x555555a606f0, time=162304, touch_id=0, sx=127.04004884004883, sy=273.16190476190479)
    at ../types/wlr_seat.c:987
#4  0x000055555555a955 in roots_cursor_handle_touch_down (cursor=0x555556113640, event=0x7fffffffdfe0) at ../rootston/cursor.c:238
#5  0x000055555555f985 in handle_touch_down (listener=0x555556113878, data=0x7fffffffdfe0) at ../rootston/seat.c:59
#6  0x00007ffff7bacb3f in wl_signal_emit (signal=0x5555561139d8, data=0x7fffffffdfe0) at /usr/include/wayland-server-core.h:388
#7  0x00007ffff7bad983 in handle_touch_down (listener=0x5555561eb060, data=0x7fffffffdfe0) at ../types/wlr_cursor.c:363
#8  0x00007ffff7b9f201 in wl_signal_emit (signal=0x5555561d2708, data=0x7fffffffdfe0) at /usr/include/wayland-server-core.h:388
#9  0x00007ffff7b9f3f2 in handle_touch_down (event=0x55555620b7f0, libinput_dev=0x5555561053e0) at ../backend/libinput/touch.c:40
#10 0x00007ffff7b9dc30 in wlr_libinput_event (backend=0x5555557806a0, event=0x55555620b7f0) at ../backend/libinput/events.c:216
#11 0x00007ffff7b9cb95 in wlr_libinput_readable (fd=22, mask=1, _backend=0x5555557806a0) at ../backend/libinput/backend.c:34
#12 0x00007ffff79709b2 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#13 0x00007ffff796f14c in wl_display_run () at /usr/lib/libwayland-server.so.0
#14 0x000055555555e1d8 in main (argc=1, argv=0x7fffffffe3b8) at ../rootston/main.c:65
```

include/rootston/seat.h
```diff
@@ -9,6 +9,9 @@ struct roots_drag_icon {
 	struct wl_list link; // roots_seat::drag_icons
 	bool mapped;
 
+	bool is_pointer;
+	int32_t touch_id;
```
SirCmpwn: This should be called slot for consistency.

SirCmpwn:
I'm also able to get a segfault here:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7973653 in wl_list_insert (list=0x555556238ae0, elm=0x5555562379e0) at src/wayland-util.c:49
49              elm->next->prev = elm;
(gdb) bt
#0  0x00007ffff7973653 in wl_list_insert (list=0x555556238ae0, elm=0x5555562379e0) at src/wayland-util.c:49
#1  0x00007ffff7bb3158 in wl_signal_add (signal=0x555556208940, listener=0x5555562379e0) at /usr/include/wayland-server-core.h:428
#2  0x00007ffff7bb5d9f in touch_point_set_focus (point=0x555556237990, surface=0x555556208880, sx=202.7130647130648, sy=247.90586080586081) at ../types/wlr_seat.c:1065
#3  0x00007ffff7bb5eaa in wlr_seat_touch_point_focus (seat=0x555556111df0, surface=0x555556208880, time=1027018, touch_id=0, sx=202.7130647130648, sy=247.90586080586081)
    at ../types/wlr_seat.c:1085
#4  0x000055555555abfb in roots_cursor_handle_touch_motion (cursor=0x555556113cd0, event=0x7fffffffdfb0) at ../rootston/cursor.c:289
#5  0x000055555555f9fd in handle_touch_motion (listener=0x555556113f38, data=0x7fffffffdfb0) at ../rootston/seat.c:73
#6  0x00007ffff7bacb3f in wl_signal_emit (signal=0x555556114078, data=0x7fffffffdfb0) at /usr/include/wayland-server-core.h:388
#7  0x00007ffff7bad9c6 in handle_touch_motion (listener=0x5555561e0f60, data=0x7fffffffdfb0) at ../types/wlr_cursor.c:370
#8  0x00007ffff7b9f201 in wl_signal_emit (signal=0x5555561e8d48, data=0x7fffffffdfb0) at /usr/include/wayland-server-core.h:388
#9  0x00007ffff7b9f62a in handle_touch_motion (event=0x5555561ff850, libinput_dev=0x555556105a20) at ../backend/libinput/touch.c:79
#10 0x00007ffff7b9dc60 in wlr_libinput_event (backend=0x5555557806a0, event=0x5555561ff850) at ../backend/libinput/events.c:222
#11 0x00007ffff7b9cb95 in wlr_libinput_readable (fd=22, mask=1, _backend=0x5555557806a0) at ../backend/libinput/backend.c:34
#12 0x00007ffff796f67a in wl_event_source_fd_dispatch (source=0x5555561e1680, ep=0x7fffffffe0c0) at src/event-loop.c:94
#13 0x00007ffff7970124 in wl_event_loop_dispatch (loop=0x55555576ff10, timeout=-1) at src/event-loop.c:633
#14 0x00007ffff796d769 in wl_display_run (display=0x555555772700) at src/wayland-server.c:1245
#15 0x000055555555e1d8 in main (argc=1, argv=0x7fffffffe3b8) at ../rootston/main.c:65
```

SirCmpwn:
It seems the wl_list inside of the touch point's listeners are getting corrupted, specifically the next pointer.

types/wlr_seat.c
acrisci: Yeah that's probably better.

include/rootston/seat.h
acrisci: The intent is to consider only the first touch point for window manager actions. So moving around a window with touch will only use the first touch point and ignore the rest. It would be pretty cool to move two windows around at the same time with two touch points, but that would require a lot of work.

include/rootston/seat.h
acrisci: I don't think consistency is possible here. Libinput gives us "slots" but the seats work with "ids" (named that in the wl-touch interface) which can be slots or anything else depending on what the backend names it. The point at which the terminology changes I'm not sure, but rootston does work with multiple backends.

It's a little sloppy for sure, so I'll give you guys the final decision now that you know that.

SirCmpwn: Alright, IDs is fine if that's the wl terminology.

emersion: I think the wlroots touch event struct should use `id` too then, since it shouldn't be libinput-dependent.

acrisci: The way to be consistent would be to call them "slots" all the way through (breaking consistency with the wayland interface names though) or change the event struct to call it a touch_id so it is consistent all throughout rootston.

emersion: >change the event struct to call it a touch_id so it is consistent all throughout rootston.

That seems better.

types/wlr_seat.c
acrisci: I've adopted this style from Weston to always set the notify function within one line of the `wl_signal_add()` call and it's used all over the project.

acrisci:
Ok I addressed all of @emersion 's issues. One of these issues did involve not cleaning up some touch point listeners so I think you should retest and see if your issue is still present @SirCmpwn .

SirCmpwn:
Aye, no more crashing. However (possibly a limitation of the seat) I cannot drag and drop two things at once.

acrisci:
> I cannot drag and drop two things at once.

I don't think that's a normal feature. It's probably a limitation of the protocol that it can't be supported.

`wl_data_device::enter` event is like a pointer enter event. It provides the offer and x/y coordinates of the enter event. Then `wl_data_device::motion` is like a pointer motion event that moves the dnd point around the surface. However, there is no parameter to specify what offer is being moved around, so there is no way for the client to distinguish motion between the offers.

Let me know if I'm wrong.

include/rootston/seat.h
emersion: Okay, let's open an issue for this feature then!

rootston/cursor.c
```diff
@@ -165,14 +166,20 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 		return;
 	}
 
-	uint32_t serial =
-		wlr_seat_pointer_notify_button(seat->seat, time, button, state);
+	uint32_t serial;
+	if (is_touch) {
+		serial = wl_display_get_serial(desktop->server->wl_display);
```
emersion: I especially don't like this `wl_display_get_serial` because:

1. It's in rootston, meaning it's not internal stuff
2. We really don't understand where the `serial` is coming from

Is there a way to do something similar to pointers?

rootston/cursor.c
```diff
@@ -214,50 +221,84 @@ void roots_cursor_handle_axis(struct roots_cursor *cursor,
...
 void roots_cursor_handle_touch_up(struct roots_cursor *cursor,
 		struct wlr_event_touch_up *event) {
-	struct roots_touch_point *point;
-	wl_list_for_each(point, &cursor->touch_points, link) {
-		if (point->slot == event->slot) {
-			wl_list_remove(&point->link);
-			free(point);
-			break;
-		}
+	struct wlr_touch_point *point = wlr_seat_touch_get_point(cursor->seat->seat, event->touch_id);
```
emersion: 80 char limit

types/wlr_data_device.c
```diff
@@ -562,13 +615,38 @@ static void drag_handle_drag_source_destroy(struct wl_listener *listener,
...
+
+	if (is_touch_grab) {
+		wl_list_for_each(point, &client->seat->touch_state.touch_points, link) {
+			is_touch_grab = point->surface && point->surface == origin;
+			break;
+		}
+	}
+
+	if (!drag->is_pointer_grab && !is_touch_grab) {
+		return true;
```
emersion: `drag` is leaked here

emersion:
Apart from these two comments, LGTM.

>However, there is no parameter to specify what offer is being moved around, so there is no way for the client to distinguish motion between the offers.

I think you're right, there's at most one drag operation per seat.

rootston/cursor.c
SirCmpwn: -1, this is how you do serials

emersion: >this is how you do serials

You do serials by incrementing them in functions (and let the user guess which function increments the serial) and then read them in a completely separate function (which would fail if not called by a function that in fact incremented the serial), hoping for it not to have been incremented by another function?

SirCmpwn: In wayland you do serials with wl_display_get_serial.

acrisci: I agree that we should be getting the serial directly from the down event. If you look at the touch down handler in rootston, you will see that we are guaranteed to get the right serial because the down notify happens right before the call to press the button and won't run if no down event was actually sent.

I think it's good enough for now. This will get refactored into something nicer when we flesh out all of the compositor touch actions in a future PR.

SirCmpwn: It really, really doesn't matter. It just has to be a unique number and wl_display_get_serial does that.

emersion: Wait, `input_events` doesn't seem used anymore. Should we remove it?

>It really, really doesn't matter. It just has to be a unique number and wl_display_get_serial does that.

It should be the serial of the touch event, not just a unique number. Note that we're using `wl_display_get_serial`, not `wl_display_next_serial`.

acrisci: In this particular case, the serial is used to validate a moveresize request from the shells. The risk is that if you don't get the right serial from the down event, then whatever serial you actually got instead is used for validation which can lead to false positives/negatives. The code how it is will always be correct, but @emersion 's concerns that this code is not robust to changes is valid and should be considered in the upcoming refactor.

SirCmpwn: Oh yeah that sucks, I was mistaking it for next_serial

acrisci:
Ok I think it's good enough for now.

SirCmpwn:
>I think it's good enough for now. This will get refactored into something nicer when we flesh out all of the compositor touch actions in a future PR.
￼ 
File ticket please

PR #419, opened 2017-11-14
Fix popups position for rotated views

### LINKED ISSUE #415: xdg-shell popups don't work anymore on rotated views
1. Open `gnome-calculator`
2. Rotate it
3. Open the combobox
4. It's not at the right position (though, input events work as expected)

This used to work, needs a bisect.

emersion:
Test plan: open `gnome-calculator`, rotate it, click on the combobox. Test with a different screen scale factor too.

Fixes #415

### DIFF
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -17,13 +17,40 @@ static inline int64_t timespec_to_msec(const struct timespec *a) {
 	return (int64_t)a->tv_sec * 1000 + a->tv_nsec / 1000000;
 }
 
+/**
+ * Computes the surface size homogeneous to the global compositor coordinates.
+ */
+static void get_surface_size(struct wlr_surface *surface, double *width,
+		double *height) {
+	double surface_scale = surface->current->scale;
+	*width = (double)surface->current->buffer_width / surface_scale;
+	*height = (double)surface->current->buffer_height / surface_scale;
+}
+
+/**
+ * Rotate a child's position relative to a parent. The parent size is (pw, ph),
+ * the child position is (*sx, *sy) and its size is (sw, sh).
+ */
+static void rotate_child_position(double *sx, double *sy, double sw, double sh,
+		double pw, double ph, float rotation) {
+	if (rotation != 0.0) {
+		// Coordinates relative to the center of the subsurface
+		double ox = *sx - pw/2 + sw/2,
+			oy = *sy - ph/2 + sh/2;
+		// Rotated coordinates
+		double rx = cos(-rotation)*ox - sin(-rotation)*oy,
+			ry = cos(-rotation)*oy + sin(-rotation)*ox;
+		*sx = rx + pw/2 - sw/2;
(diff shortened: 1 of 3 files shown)


rootston/output.c
```diff
@@ -17,13 +17,40 @@ static inline int64_t timespec_to_msec(const struct timespec *a) {
...
 }
 
+/**
+ * Computes the surface size homogeneous to the global compositor coordinates.
+ */
+static void get_surface_size(struct wlr_surface *surface, double *width,
+		double *height) {
+	double surface_scale = surface->current->scale;
+	*width = (double)surface->current->buffer_width / surface_scale;
+	*height = (double)surface->current->buffer_height / surface_scale;
```
acrisci: Isn't this already present in `buffer->current->width|height`?

emersion: You're right, fixed.

rootston/output.c
```diff
@@ -17,13 +17,30 @@ static inline int64_t timespec_to_msec(const struct timespec *a) {
 	return (int64_t)a->tv_sec * 1000 + a->tv_nsec / 1000000;
 }
 
+/**
+ * Rotate a child's position relative to a parent. The parent size is (pw, ph),
+ * the child position is (*sx, *sy) and its size is (sw, sh).
+ */
+static void rotate_child_position(double *sx, double *sy, double sw, double sh,
+		double pw, double ph, float rotation) {
```
acrisci: Out args generally go last.

emersion: So this is in _and_ out. Not sure what to do about it. Maybe make it in-only, and add another pair of out arguments?

rootston/output.c
```diff
@@ -95,36 +103,55 @@ static void render_surface(struct wlr_surface *surface,
...
-			popup_y, rotation);
-		render_xdg_v6_popups(popup, desktop, wlr_output, when, popup_x, popup_y,
-			rotation);
+		double popup_width = popup->surface->current->width;
+		double popup_height = popup->surface->current->height;
+
+		double popup_x = surface->geometry->x + popup->popup_state->geometry.x -
+			popup->geometry->x;
+		double popup_y = surface->geometry->y + popup->popup_state->geometry.y -
+			popup->geometry->y;
```
acrisci: I think this calculation is complicated enough that it needs to be in the wlr-xdg implementation.

rootston/output.c
SirCmpwn: This is an internal rootston function, I don't care

rootston/output.c
emersion: Yeah, will do.

acrisci:
Tested with xdg-popup and subsurfaces and works ok.

Maybe unrelated, but resizing is especially broken with rotated windows.

Will test wl-shell a little later.

acrisci:
Works with wl-shell. Does not work with xwayland.

SirCmpwn:
Thanks!

PR #420, opened 2017-11-15
Call wl_output_send_done when current mode is updated

### LINKED ISSUE #330: XWayland pointer bound to first received mode
1. Launch the Wayland backend
2. Resize the rootston window to make it bigger
3. Launch a X11 app
4. Pointer events work well in the initial window size box, but don't outside

emersion:
After reading xwayland's source code, I realized we aren't sending the `done` signal when the current mode is updated.

The spec says:

>This event is sent after all other properties have been sent after binding to the output object and after any other property changes done after that. This allows changes to the output properties to be seen as atomic, even if they happen via multiple events. 

(https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_output-event-done)

Also fixes window maximization on xwayland.

Test plan: run `xrandr`, update rootston window size, run `xrandr` again.

Fixes #330

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -29,8 +29,7 @@ static void wl_output_send_to_resource(struct wl_resource *resource) {
 	if (version >= WL_OUTPUT_MODE_SINCE_VERSION) {
 		struct wlr_output_mode *mode;
 		wl_list_for_each(mode, &output->modes, link) {
-			// TODO: mode->flags should just be preferred
-			uint32_t flags = mode->flags;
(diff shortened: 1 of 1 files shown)


types/wlr_output.c
```diff
@@ -163,6 +166,9 @@ bool wlr_output_set_mode(struct wlr_output *output,
 
 void wlr_output_update_size(struct wlr_output *output, int32_t width,
 		int32_t height) {
+	if (output->width == width && output->height == height) {
```
emersion: Surprisingly, removing this check makes xwayland crash on launch.

SirCmpwn:
Derp, good call.

PR #421, opened 2017-11-16
Terminate on parent compositor exit

### LINKED ISSUE #397: X11 backend: Doesn't die when X exits.
Testing with the X11 backend I (accidently) killed my X and the waymonad process continued to exist spammin **a lot** of 
`2017-11-08 14:31:54 - [backend/x11/backend.c:354] eglSwapBuffers failed: Bad native window`

emersion:
Test plan: open rooston, and open another rootston inside it (but not from within rootston):

* With the Wayland backend: `env WAYLAND_DISPLAY=wayland-1 build/rootston/rootston`
* With the X11 backend: `env WAYLAND_DISPLAY= DISPLAY=:1 build/rootston/rootston`

Kill the outer rootston. Check that the inner one exits.

Fixes #397

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -41,7 +41,7 @@ static bool wlr_wl_backend_start(struct wlr_backend *_backend) {
 	wlr_log(L_INFO, "Initializating wayland backend");
 
 	wlr_wl_registry_poll(backend);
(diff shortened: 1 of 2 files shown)


acrisci:
👍 looks good

Issue #398, opened 2017-11-08
keyboard_keysym_press called twice


emersion:
`keyboard_keysym_press` is called several times per keypress: one time in `keyboard_keysyms_simple` and one time in `keyboard_keysyms_xkb`. So pressing Alt+Shift+2 will add 2 _and_ @ in `roots_keyboard_pressed_keysyms`.

Introduced in #394.

cc @Ongy

Ongy:
Ah, I had noticed this, but scanning the code the only usage I found was this here https://github.com/swaywm/wlroots/blob/master/rootston/keyboard.c#L93
And since that was for the keybinds, it made sense to me.

Is it an integral feature to check for combinations that are more than one key (key, not modifier)?
Otherwise it would be easier to completly bypass this state for keybind detection imo.

emersion:
>Ah, I had noticed this, but scanning the code the only usage I found was this here

The state is inconsistent, and that could fuck up some keybindings. Imagine I want to bind Ctrl+Shift+@ because on my keyboard I have another key for @. The keybinding would be triggered even if Shift has been consumed to transform the 2 into a @.

>Is it an integral feature to check for combinations that are more than one key (key, not modifier)?

Yes.

Now that we track the currently pressed keys in `wlr_keyboard`, it should be pretty easy to remove `roots_keyboard.keysyms` and get keysyms from all pressed keycodes each time a key is pressed.

SirCmpwn:
Thanks!

PR #422, opened 2017-11-16
Fix a bunch of mistakes detected with scan-build

emersion:
To run the static analyzer:

```
ninja scan-build -C build
```

I didn't fix the keyboard thing because I'll refactor it anyway when fixing #398.


acrisci:
👍 looks good.

PR #423, opened 2017-11-16
Refactor rootston keyboard

### LINKED ISSUE #398: keyboard_keysym_press called twice

emersion:
Test plan: keybinds such as `Ctrl+Shift+E` shouldn't work anymore. Keybinds such as `Ctrl+Shift+e` and `Ctrl+E` should work. Check that switching VTs works.

Fixes #398

### DIFF
--- a/rootston/keyboard.c
+++ b/rootston/keyboard.c
@@ -23,6 +23,16 @@ static ssize_t pressed_keysyms_index(xkb_keysym_t *pressed_keysyms,
 	return -1;
 }
 
+static size_t pressed_keysyms_length(xkb_keysym_t *pressed_keysyms) {
+	size_t n = 0;
+	for (size_t i = 0; i < ROOTS_KEYBOARD_PRESSED_KEYSYMS_CAP; ++i) {
+		if (pressed_keysyms[i] != XKB_KEY_NoSymbol) {
+			++n;
+		}
+	}
+	return n;
+}
+
 static void pressed_keysyms_add(xkb_keysym_t *pressed_keysyms,
 		xkb_keysym_t keysym) {
 	ssize_t i = pressed_keysyms_index(pressed_keysyms, keysym);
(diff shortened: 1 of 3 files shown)


rootston/keyboard.c
```diff
@@ -56,21 +75,13 @@ static void keyboard_binding_execute(struct roots_keyboard *keyboard,
...
 
 /**
- * Process a keypress from the keyboard.
+ * Execute a built-in, hardcoded compositor action when a keysym is pressed.
  *
  * Returns true if the keysym was handled by a binding and false if the event
  * should be propagated to clients.
  */
-static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
+static bool keyboard_press_keysym(struct roots_keyboard *keyboard,
```
acrisci: Probably needs a better name like `keyboard_keysym_compositor_action()`.

acrisci:
The keyboard command "close" should be seat aware. It should close the focused window for the seat of the keyboard device that executed the binding.

acrisci:
We should probably have a focus list per seat so things like "next_window" will work correctly. Would also affect reverting focus when the focused view is closed for a seat. (these are probably separate issues though).

rootston/keyboard.c
```diff
@@ -116,86 +155,64 @@ static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
...
-	bool handled = keyboard_keysyms_xkb(keyboard, keycode, event->state);
+	xkb_keycode_t keycode = event->keycode + 8;
 
+	uint32_t modifiers;
+	const xkb_keysym_t *keysyms;
+	size_t keysyms_len = keyboard_keysyms_translated(keyboard, keycode,
+		&keysyms, &modifiers);
+	bool handled = keyboard_handle_keysyms(keyboard,
+		keyboard->pressed_keysyms_translated, modifiers, keysyms, keysyms_len,
+		event->state);
```
acrisci: There is a subtle bug here with this approach. If I bind to `a` and `b` and then press `a`, press `b`, release `a`, release `b`, then the `a` binding will run twice, once when I press `a` and once when I press `b` when I would expect it to only run each binding once.

rootston/keyboard.c
```diff
@@ -90,7 +101,35 @@ static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
...
+/**
+ * Press or release keysyms.
+ *
+ * Returns true if the keysym was handled by a binding and false if the event
+ * should be propagated to clients.
+ */
+static bool keyboard_handle_keysyms(struct roots_keyboard *keyboard,
+		xkb_keysym_t *pressed_keysyms, uint32_t modifiers,
+		const xkb_keysym_t *keysyms, size_t keysyms_len,
+		enum wlr_key_state state) {
```
acrisci: Could use a better name that indicates that it actually executes bindings like `keyboard_execute_bindings()`.

rootston/keyboard.c
```diff
@@ -116,86 +155,64 @@ static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
...
 /*
- * Process keypresses from the keyboard as xkb sees them.
+ * Get keysyms and modifiers from the keyboard as xkb sees them.
  *
  * This uses the xkb keysyms translation based on pressed modifiers and clears
  * the consumed modifiers from the list of modifiers passed to keybind
  * detection.
  *
- * (On US layout) this will trigger: [Alt]+[at]
+ * On US layout, this will trigger: Alt + @
```
acrisci: I don't understand this comment line.

acrisci: Oh I see, the context for this comment comes from several lines below. This is kind confusing to read and should probably be made self contained for each.

rootston/keyboard.c
acrisci: I think the easy fix to this is to make sure the keysyms pressed are *exactly* the ones specified by the binding.

rootston/keyboard.c
```diff
@@ -116,86 +155,64 @@ static bool keyboard_keysym_press(struct roots_keyboard *keyboard,
...
-
-	bool handled = keyboard_keysyms_xkb(keyboard, keycode, event->state);
+	xkb_keycode_t keycode = event->keycode + 8;
 
+	uint32_t modifiers;
+	const xkb_keysym_t *keysyms;
+	size_t keysyms_len = keyboard_keysyms_translated(keyboard, keycode,
+		&keysyms, &modifiers);
+	bool handled = keyboard_handle_keysyms(keyboard,
+		keyboard->pressed_keysyms_translated, modifiers, keysyms, keysyms_len,
```
acrisci: These variables (`keyboard->pressed_keysyms_*`) are set but never used.

acrisci: Disregard that, I see how they are used now.

acrisci:
Is there a good reason why keysym translation is a rootston responsibility? Why is this state not kept on wlr-keyboard itself?

emersion:
>Is there a good reason why keysym translation is a rootston responsibility? Why is this state not kept on wlr-keyboard itself?

The seat takes keycodes, not keysyms. Also, this whole story about using modifiers or not when getting keysyms from keycodes shows that this compositor should be able to customize how it handles keycodes.

emersion:
Refactored a little bit more, this should address all comments.

acrisci:
👍 looks good.

PR #426, opened 2017-11-16
readme: add basic running doc

acrisci:


README.md
```diff
@@ -33,3 +33,25 @@ Run these commands:
...
+## Running the Reference Compositor
+
+wlroots comes with a reference compositor called rootston that demonstrates the
+features of the library.
+
+After building, run rootston from a terminal or VT with:
+
+    ./build/rootston/rootston
+
+Run windows on the compositor with the display variable set:
```
SirCmpwn: I would rather document keybindings for summoning a terminal or whatever (which will have WAYLAND_DISPLAY set) than explaining this.

README.md
```diff
@@ -33,3 +33,25 @@ Run these commands:
...
+
+```bash
+# run gtk apps like this
+gnome-calculator --display=wayland-0
```
emersion: I think `WAYLAND_DISPLAY` is preferred because it'll work with all Wayland apps.

README.md
acrisci: My intent is to get people up and running as fast as possible with as little overhead as possible to get a basic demo before diving deeper. Can that be done in a small bash script?

SirCmpwn: No one is going to come here and need to get up and running so fast that they don't understand the wayland ecosystem they live in.

PR #427, opened 2017-11-17
rootston: add per-seat views

### LINKED ISSUE #424: make rootston commands seat aware
The "close" command should close the focused window for the seat.

The "next_window" command should focus the next window in the seat's focus stack (not implemented).

emersion:
This also allows to prevent some seats from sending input events to some views.

Test plan: with two seats, check that focus semantics and keybindings operate per-seat. This includes: open three views, focus different ones with the seats, focus the same one with both, close the focused view. Expected: the two previous views get focus.

Fixes #424


rootston/xwayland.c
```diff
@@ -174,29 +173,24 @@ static void handle_request_maximize(struct wl_listener *listener, void *data) {
...
 	struct roots_view *view = roots_surface->view;
-	struct wlr_xwayland_surface *xsurface = view->xwayland_surface;
-	struct roots_desktop *desktop = view->desktop;
+	//struct roots_desktop *desktop = view->desktop;
 
 	view->wlr_surface = xsurface->surface;
 	view->x = (double)xsurface->x;
 	view->y = (double)xsurface->y;
 
-	wlr_list_push(desktop->views, roots_surface->view);
```
emersion: I don't really understand what to do here, the view has already been added here: https://github.com/swaywm/wlroots/pull/427/files#diff-94b621ef8193066933ec606392245a3bL253

acrisci:
It's a very good start, but I think we can improve the data structures a bit.

I don't like it how the seat is now a first-class container of views when all we want to do is maintain a focus stack. I think this is an excellent use case for wlr_list. I'm also not thrilled about wrapping the view in `roots_seat_view` only for the purpose of storing it in a list because it's a conceptually heavy abstraction for that purpose alone.

Here's what I'll suggest:

Keep a wlr_list of views on the seat. When you focus a view for a seat, add it to the front of the list. When the view is destroyed, remove it from the list. The view at the head of the list is the focused view. When the focus of the seat is cleared, set a `has_focus` bit on the seat to false.

Now abstract getting the focus for a seat into a getter. It returns NULL if the list is empty or `has_focus` is false or otherwise the head of the list.

What do you think about that?

emersion:
One drawback is that we loose the `destroy` listener, making it necessary for the desktop to remove views from all seats when these are destroyed.

But I agree that an array of pointers is better than a `wl_list` in this case. I've made the change in the last commit.

acrisci:
> One drawback is that we loose the destroy listener

Each seat should listen to the `destroy` signal of the view itself and remove it from the list when that happens.

rootston/keyboard.c
```diff
@@ -91,16 +91,14 @@ static void keyboard_binding_execute(struct roots_keyboard *keyboard,
 	if (strcmp(command, "exit") == 0) {
 		wl_display_terminate(keyboard->input->server->wl_display);
 	} else if (strcmp(command, "close") == 0) {
-		if (!wl_list_empty(&seat->views)) {
-			struct roots_seat_view *first_seat_view = wl_container_of(
-				seat->views.next, first_seat_view, link);
-			view_close(first_seat_view->view);
+		if (seat->views->length > 0) {
```
acrisci: Should use the getter. What if there is nothing focused?

emersion:
Where do we store the wl_listener then?

rootston/seat.c
```diff
@@ -544,92 +549,70 @@ void roots_seat_focus_view(struct roots_seat *seat, struct roots_view *view) {
...
-	}
-}
-
-static void seat_view_handle_destroy(struct wl_listener *listener, void *data) {
-	struct roots_seat_view *seat_view =
-		wl_container_of(listener, seat_view, destroy);
-	seat_view_destroy(seat_view);
-}
-
 void roots_seat_add_view(struct roots_seat *seat, struct roots_view *view) {
```
acrisci: There's no reason for this. Just add it when it is focused and remove it when it is destroyed.

emersion: We may want to give some views to some seats only.

emersion:
Ready for review!

rootston/keyboard.c
```diff
@@ -87,19 +87,20 @@ static const char *exec_prefix = "exec ";
...
+	struct roots_seat *seat = keyboard->seat;
 	if (strcmp(command, "exit") == 0) {
-		wl_display_terminate(server->wl_display);
+		wl_display_terminate(keyboard->input->server->wl_display);
 	} else if (strcmp(command, "close") == 0) {
-		if (server->desktop->views->length > 0) {
-			struct roots_view *view =
-				server->desktop->views->items[server->desktop->views->length-1];
-			view_close(view);
+		if (!wl_list_empty(&seat->views)) {
```
acrisci: This should act on the focused view. If no view is focused, it should do nothing.

emersion: You're right. Maybe it makes sense to add `roots_seat_focus_next_view` for the other command?

rootston/xwayland.c
acrisci: I think the wayland surface is destroyed on unmap and a new wayland surface is created on map.

include/rootston/seat.h
```diff
@@ -14,14 +14,23 @@ struct roots_seat {
...
 	struct wl_list touch;
 	struct wl_list tablet_tools;
 };
 
+struct roots_seat_view {
+	struct roots_seat *seat;
+	struct roots_view *view;
+	struct wl_list link; // roots_seat::views
+
+	struct wl_listener destroy;
```
acrisci: Should probably be explicit here with `view_destroy`.

include/rootston/seat.h
```diff
@@ -60,6 +69,8 @@ void roots_seat_configure_xcursor(struct roots_seat *seat);
 
 bool roots_seat_has_meta_pressed(struct roots_seat *seat);
 
+struct roots_view *roots_seat_get_focused_view(struct roots_seat *seat);
```
acrisci: Since views are primarily what is focused, I think we can change the name to `roots_seat_get_focused()`.

rootston/seat.c
```diff
@@ -495,32 +549,44 @@ void roots_seat_focus_view(struct roots_seat *seat, struct roots_view *view) {
...
+			if (seat_view == NULL) {
+				wlr_log(L_ERROR, "Allocation failed");
+				return;
+			}
+		}
+	}
+
+	seat->has_focus = false;
 
 	// unfocus the old view if it is not focused by some other seat
```
acrisci: should read `s/unfocus/deactivate`.

emersion:
@acrisci's comments have been addressed.

rootston/keyboard.c
acrisci: I think that makes sense.

rootston/seat.c
```diff
@@ -495,32 +549,44 @@ void roots_seat_focus_view(struct roots_seat *seat, struct roots_view *view) {
...
-		}
-	}
-
 	view_activate(view, true);
-	// TODO: list_swap
-	wlr_list_del(desktop->views, index);
-	wlr_list_add(desktop->views, view);
+	wl_list_remove(&seat_view->view->link);
+	wl_list_insert(&seat->input->server->desktop->views,
+		&seat_view->view->link);
```
acrisci: I think it should be raised on the desktop regardless of whether or not it is currently the focused view.

acrisci:
Focusing a new window for all the seats is going to have some serious usability problems. I wish there was some way to map a new window to a seat.

martinetd:
Would be a bit cumbersome but we might be able to map a window to a seat by looking at its pid/ppid and see if these are related to anything we recognize, and if we find such a view look for seats that has it focused as of the new window spawning?
That'd require keeping tracks of what pid owns every surface though, not sure we want to bother with that -- or actually of how easy that would be. X11 has _NET_WM_PID property, is there something equivalent for wl/xdg shells that'd let us know?

rootston/seat.c
```diff
@@ -590,6 +590,15 @@ void roots_seat_focus_view(struct roots_seat *seat, struct roots_view *view) {
 	wlr_seat_keyboard_notify_enter(seat->seat, view->wlr_surface);
 }
 
+void roots_seat_cycle_focus(struct roots_seat *seat) {
```
acrisci: I think it would be more usable to focus the previously focused view here, which is the second in the list. Take the first view in the list and put it at the back to preserve the order.

acrisci:
> is there something equivalent for wl/xdg shells that'd let us know?

If you can propose *any* way to get this working with xdg-shell without a new protocol we should consider it, but I think it's going to need a new protocol.

martinetd:
> If you can propose any way to get this working with xdg-shell without a new protocol we should consider it, but I think it's going to need a new protocol.
I was looking along for how an xkill replacement would work, but there actually is something standard, found in weston, give or take null checks:
```
pid_t weston_desktop_surface_get_pid(...) {
  pid_t pid;
...
  struct wl_client *wl_client = ...;
  wl_client_get_credentials(wl_client, &pid, NULL, NULL);
...
  return pid;  
}
```
So, *if* we want to, we could track the PID of every surface (through this for native wayland and X11 props for xwayland) -- I'm still dubious if it's a good idea to try to associate focus based on PID or PPID though, it was an idea but it'll definitely need some consideration...

rootston/seat.c
emersion: Yeah. Done!

acrisci:
👍 looks good

PR #428, opened 2017-11-17
Fix segfault when xwayland is disabled

emersion:
Test plan: disable xwayland, launch rootston, should not crash.


SirCmpwn:
Thanks!

PR #429, opened 2017-11-18
Add cursor theme to rootston config

emersion:
Test plan: install a [XCursor theme](https://www.archlinux.org/packages/?sort=&q=xcursor-&maintainer=&last_update=&flagged=&limit=50) and update your config, For instance, install `xcursor-themes` and use this config:

```ini
[cursor]
theme = redglass
```

### DIFF
--- a/include/rootston/config.h
+++ b/include/rootston/config.h
@@ -48,6 +48,7 @@ struct roots_config {
 	struct {
 		char *mapped_output;
 		struct wlr_box *mapped_box;
+		char *theme;
 	} cursor;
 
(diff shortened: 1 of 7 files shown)


SirCmpwn:
Do we want to configure this on a per-seat basis?

emersion:
Implemented per-seat cursor configuration. The syntax is:

```ini
[cursor:seat name]
theme = default
...
```

`[cursor]` will configure the default seat.

SirCmpwn:
Thanks!

PR #431, opened 2017-11-18
Refactor wlr_list

emersion:
Note: the only part of the code using `wlr_list` is the libinput backend.


emersion:
This is ready for review.

PR #437, opened 2017-11-20
Fullscreen support

emersion:
This PR implements basic fullscreen support. This PR doesn't implement hardware fullscreen, this will be done in a separate PR. We need to do software fullscreen anyway in case hardware fullscreen is not supported.

- [x] xdg-shell support
- [x] wl-shell support
- [x] xwayland support
- [x] Render subsurfaces/popups/...
- [ ] ~~Render subsurfaces in `wlr_output`?~~ No compositing in `wlr_output`

Test plan:
- [x] `mpv`
  - [x] xwayland
  - [x] wl-shell
- [x] `evince`
  - [x] xdg-shell
  - [x] xwayland
- [x] `epiphany` (xdg-shell)
- [x] `chromium` (xwayland)
- [x] Test scaled outputs
- [x] Test rotated outputs

### DIFF
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -38,6 +38,7 @@ struct roots_xwayland_surface {
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
 	struct wl_listener request_maximize;
+	struct wl_listener request_fullscreen;
 	struct wl_listener map_notify;
 	struct wl_listener unmap_notify;
 };
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -57,7 +57,7 @@ struct roots_view {
 	float rotation;
 
 	bool maximized;
-	struct wlr_output *fullscreen_output;
+	struct roots_output *fullscreen_output;
 	struct {
 		double x, y;
 		uint32_t width, height;
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -12,8 +12,8 @@ struct roots_wl_shell_surface {
 	struct wl_listener destroy;
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
-	struct wl_listener request_set_maximized;
-	struct wl_listener request_set_fullscreen;
+	struct wl_listener request_maximize;
+	struct wl_listener request_fullscreen;
 	struct wl_listener set_state;
 
 	struct wl_listener surface_commit;
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -13,6 +13,7 @@ struct roots_wl_shell_surface {
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
 	struct wl_listener request_set_maximized;
+	struct wl_listener request_set_fullscreen;
 	struct wl_listener set_state;
 
 	struct wl_listener surface_commit;
@@ -26,6 +27,7 @@ struct roots_xdg_surface_v6 {
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
 	struct wl_listener request_maximize;
+	struct wl_listener request_fullscreen;
 };
 
 struct roots_xwayland_surface {
@@ -54,6 +56,7 @@ struct roots_view {
 	float rotation;
 
 	bool maximized;
+	struct wlr_output *fullscreen_output;
 	struct {
 		double x, y;
 		uint32_t width, height;
@@ -93,6 +96,7 @@ struct roots_view {
 	void (*move_resize)(struct roots_view *view, double x, double y,
 		uint32_t width, uint32_t height);
 	void (*maximize)(struct roots_view *view, bool maximized);
+	void (*set_fullscreen)(struct roots_view *view, bool fullscreen);
 	void (*close)(struct roots_view *view);
 };
 
@@ -103,6 +107,8 @@ void view_resize(struct roots_view *view, uint32_t width, uint32_t height);
 void view_move_resize(struct roots_view *view, double x, double y,
 	uint32_t width, uint32_t height);
 void view_maximize(struct roots_view *view, bool maximized);
+void view_set_fullscreen(struct roots_view *view, bool fullscreen,
+	struct wlr_output *output);
 void view_close(struct roots_view *view);
 bool view_center(struct roots_view *view);
 void view_setup(struct roots_view *view);
--- a/include/wlr/types/wlr_xdg_shell_v6.h
+++ b/include/wlr/types/wlr_xdg_shell_v6.h
@@ -152,6 +152,13 @@ struct wlr_xdg_toplevel_v6_resize_event {
 	uint32_t edges;
 };
 
+struct wlr_xdg_toplevel_v6_set_fullscreen_event {
+	struct wl_client *client;
+	struct wlr_xdg_surface_v6 *surface;
+	bool fullscreen;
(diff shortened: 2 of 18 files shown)


SirCmpwn:
We do have to render subsurfaces.

emersion:
All right, this is ready for review!

SirCmpwn:
Can you add a keybinding to fullscreen arbitrary windows?

include/rootston/view.h
```diff
@@ -13,6 +13,7 @@ struct roots_wl_shell_surface {
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
 	struct wl_listener request_set_maximized;
+	struct wl_listener request_set_fullscreen;
```
acrisci: Naming should be consistent across the views if possible (so `request_fullscreen` and `request_mazimized`).

emersion: I tried to respect each shell's terminology. I'll unify all of them then.

include/wlr/types/wlr_xdg_shell_v6.h
```diff
@@ -152,6 +152,13 @@ struct wlr_xdg_toplevel_v6_resize_event {
 	uint32_t edges;
 };
 
+struct wlr_xdg_toplevel_v6_set_fullscreen_event {
+	struct wl_client *client;
```
acrisci: Why are we putting the client in these events? Shouldn't it be always the same as the client for the shell surface?

emersion: I just copy-pasta-ed the boilerplate from other events. So yeah, I don't think the client is useful in these events. Let's remove it.

types/wlr_surface.c
```diff
@@ -929,3 +929,17 @@ void wlr_surface_send_leave(struct wlr_surface *surface,
 		}
 	}
 }
+
+static inline int64_t timespec_to_msec(const struct timespec *a) {
```
ascent12: We have this functions in a couple of places.
It should probably live in util or something.

rootston/wl_shell.c
```diff
@@ -59,6 +59,17 @@ static void handle_request_set_maximized(struct wl_listener *listener,
...
 }
 
+static void handle_request_set_fullscreen(struct wl_listener *listener,
+		void *data) {
+	struct roots_wl_shell_surface *roots_surface =
+		wl_container_of(listener, roots_surface, request_set_fullscreen);
+	struct roots_view *view = roots_surface->view;
+	struct wlr_wl_shell_surface_set_fullscreen_event *e = data;
+
+	// TODO: support e->method, e->framerate
```
acrisci: I really doubt we'll ever support these things.

acrisci:
Why couldn't we do direct scanout for a surface with subsurfaces?

emersion:
If I've understood correctly, the goal of direct scan-out is avoiding a rendering step: the buffer is directly displayed in an overlay plane (maybe even without a copy, from what I've seen from weston's source). Correct me if I'm wrong, I'm not familiar with this.

SirCmpwn:
Same reason we can't do it for all views at all times. They need to be composited.

types/wlr_xdg_shell_v6.c
```diff
@@ -677,15 +663,38 @@ static void xdg_toplevel_protocol_unset_maximized(struct wl_client *client,
...
-	wl_signal_emit(&surface->events.request_fullscreen, surface);
+
+	struct wlr_xdg_toplevel_v6_set_fullscreen_event event = {
+		.client = client,
+		.surface = surface,
+		.fullscreen = false,
+		.output = NULL,
+	};
+
+	wl_signal_emit(&surface->events.request_fullscreen, &event);
```
acrisci: I think we wait until the shell surface commits the whole state before we emit the event.

emersion: What do you mean? We should wait? Right now `request_fullscreen` isn't triggered on commit.

Note: the state doesn't keep track of the `output` parameter, which would make waiting for a commit more tricky.

types/wlr_output.c
```diff
@@ -252,6 +252,27 @@ void wlr_output_make_current(struct wlr_output *output) {
 	output->impl->make_current(output);
 }
 
+static void output_fullscreen_surface_render(struct wlr_output *output,
+		struct wlr_surface *surface, const struct timespec *when) {
```
acrisci: Is the whole point of handling the rendering here to support direct scanout later?

emersion: Yes. This is a software fallback in case overlay planes and direct scan-out isn't supported.

types/wlr_xdg_shell_v6.c
acrisci: I'm making a note of this now because I'm pretty sure from looking at the code we're going to see rendering artifacts when the fullscreen happens. I'll test it when you make the command.

types/wlr_surface.c
emersion: Actually, there's only one of them in wlroots, which is this one. All others are in `examples/` (in `output-layout.c`, `rotation.c` and `simple.c`), so I wonder if it's worth it.

emersion:
>Can you add a keybinding to fullscreen arbitrary windows?

Added the `fullscreen` command that toggles fullscreen for the focused view. Some views don't expect that (e.g. `gnome-calculator` is buggy, have to click outside the view when fullscreened otherwise it's clipped) but most of them react correctly.

Addressed all other comments.

acrisci:
I'm getting lots of bugs with this.

1. partially rendered content on fullscreen similar to #439 
2. gets the wrong size on my output (drm and x11 backend)
3. pointer is messed up after return to normal state - doesn't respond to click events and pointer is put in the wrong location on the view (probably related to client cursors)

emersion:
Yeah, this is pretty strange. `wlr_output_set_fullscreen_surface` messes up the output, even after there's no fullscreen surface anymore. I'll look into this.

I'd like to address the partially rendered content in another PR, solving the issue at the same time for fullscreen and maximize.

emersion:
@acrisci Issues 2 and 3 should be fixed with the last commit.

emersion:
https://github.com/swaywm/wlroots/pull/437/commits/3262661e1e77db11d7e4939aa377804046df43f5 fixes a bug: `wl_output.enter` wasn't sent to the output at (0, 0) in the layout. This fix reveals a bug: on rotated outputs, `weston-terminal` draws an already rotated surface, and we rotate it a second time. I'll open an issue when this PR gets merged.

acrisci:
I'm still seeing a little jump sometimes on the transition. The window moves around before it goes in and out of fullscreen.

Also, take the window in and out of fullscreen many times and you will see that the size shrinks every time.

emersion:
>I'm still seeing a little jump sometimes on the transition.

As I said, I'll fix this with #439 in another PR since the fixes for these two bugs will be very similar.

>Also, take the window in and out of fullscreen many times and you will see that the size shrinks every time.

Which client? I can't reproduce with `evince` or `chromium`.

acrisci:
I can't reproduce it with those either. It's just gnome-terminal.

emersion:
>I can't reproduce it with those either. It's just gnome-terminal.

I don't think we can do anything about it. I can reproduce in GNOME.

acrisci:
👍

PR #443, opened 2017-11-22
Change glapi.sh to generate both source and header simeltaneously

ascent12:
This is how I originally wanted to write this file, but due to some meson limitations, I couldn't.
It's been fixed in 0.43.0, so now we can do this.


SirCmpwn:
Thanks!

PR #444, opened 2017-11-22
Xwayland signal confusion fix & non-Linux socket path fix

myfreeweb:
I managed to get Xwayland working on FreeBSD, writing this from a GPU accelerated Firefox running under rootston :)

Turns out the problem was literally the same as this from 2014: [Weston confuses SIGUSR1 signals from Xwayland and VT switch](https://phabricator.freedesktop.org/T7080). Xwayland sends SIGUSR1, rootston interprets it as vt switch and freezes the whole desktop. "I'm surprised it hasn't blown up for anyone else yet" indeed. I guess most testing was done with logind.

Also fixed a socket creation error, it was creating at the same path twice (my copy-paste mistake when doing the same changes here as in wlc, forgot the second path thing)

### DIFF
--- a/backend/session/direct-freebsd.c
+++ b/backend/session/direct-freebsd.c
@@ -148,8 +148,8 @@ static bool setup_tty(struct direct_session *session, struct wl_display *display
 
 	struct vt_mode mode = {
 		.mode = VT_PROCESS,
-		.relsig = SIGUSR1,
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Can you address the build error?

myfreeweb:
yeah done

SirCmpwn:
Thanks!

PR #446, opened 2017-11-26
Use anonymous shared memory fds for tmpfiles when available

myfreeweb:
On Linux >= 3.17, memfd is used.
On FreeBSD, shm_open with SHM_ANON is used.

This reduces potential filesystem pollution and avoids potential problems such as usage of posix_fallocate on ZFS filesystems (which was disallowed in FreeBSD 12).

I haven't tested that memfd works, only that it compiles — the only Linux box I have right now is a headless VM.

### DIFF
--- a/backend/wayland/os-compatibility.c
+++ b/backend/wayland/os-compatibility.c
@@ -23,18 +23,26 @@
  * SOFTWARE.
  */
 
-#define _XOPEN_SOURCE 700
+#ifdef __FreeBSD__
+#include <sys/mman.h>
+#define UNUSED_ON_FBSD __attribute__((__unused__))
+#else
+#define UNUSED_ON_FBSD
+#endif
+#ifdef HAS_MEMFD
+#define _GNU_SOURCE
+#include <sys/syscall.h>
+#include <linux/memfd.h>
+#endif
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <errno.h>
-#ifdef __linux__
-#include <sys/epoll.h>
-#endif
 #include <string.h>
 #include <stdlib.h>
 
+#if !defined(HAS_MEMFD) && !defined(__FreeBSD__) && !defined(HAVE_MKOSTEMP)
 int
 os_fd_set_cloexec(int fd)
 {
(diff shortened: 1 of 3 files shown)


SirCmpwn:
This is not portable enough, no thanks.

ascent12:
wlroots is supposed to be as POSIX compliant as we can possibly get it. There needs to be an extremely good reason to break that (e.g. it's unavoidable, like DRM).
Naturally, anything which defines _GNU_SOURCE is going to be rejected.

SirCmpwn:
You should file a bug with FreeBSD. Not supporting posix_fallocate is their bug, not ours.

SirCmpwn:
>posix_fallocate is supported, just not on ZFS.

That's still their goddamn bug

>You have three optional Linux-specific dependencies listed

But not GNU specific.

SirCmpwn:
Also, I looked into the README. None of those are workarounds for ***bugs*** in Linux.

myfreeweb:
This isn't a workaround for a bug either, except as a side effect. My patch includes `memfd` for Linux too.

memfd/shm is just more elegant than creating a file and unlinking it.

SirCmpwn:
Your patch is Linux/BSD specific. Not POSIX. Not portable. No thanks.

main--:
I believe the relevant bug that caused this change is https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=223287. The way I understand it the reasoning here is that `posix_fallocate` doesn't make sense for a copy-on-write filesystem like ZFS since once you actually write something other than zeroes to the file, this would always create a new copy - it's thus impossible to "allocate" the space prior to performing the actual writes in the sense that they would be guaranteed to not run out of space. Since this guarantee is the entire point of `posix_fallocate`, the fact that it used to work is a bug and disallowing it on ZFS is the correct fix.

Since the code already falls back to a simple truncate in some cases, a best-effort approach really seems best here. Which (at least in my opinion) means that the correct solution would be to use appropriate APIs where possible (memfd/shm as suggested here) and fall back to just a truncate otherwise. `posix_fallocate` really is the wrong tool for the job once you consider that its main purpose is this no-ENOSPC guarantee that wlroots apparently doesn't *really* need and that some filesystems just can't provide *by design*.

SirCmpwn:
It's a fallback for a reason - it's the less ideal solution. Your runtime dir wasn't already on a ramdisk in the first place. These files are not meant to be persisted, and depending on the guarantee fallocate makes is a necessary part of the design, despite what some random on GitHub might have to say about it. I'm not interested in this patch and that's final.

Issue #448, opened 2017-11-26
Change backend/wayland/os-compatibility.c


ascent12:
Some functions are being used outside of the wayland backend, so this file should be moved to util and an appropriate header should be written for it.
I'm not sure if we want this to be public or not, though.

~~Also, the file doesn't match our coding style (it's largely unmodified after it was copied from weston), so that should be fixed.~~

SirCmpwn:
It doesn't need to be updated to use our coding style. It's intentional that files originating from external sources use the external coding style.

moritz31:
Did some minor changes, hopefully i understood things right, otherwise please correct me

moritz31:
Issue should be solved, or is there something missing?

SirCmpwn:
In the future you can add `Fixes #448` to your extended commit message and it'll automatically close the issue.

Timidger:
You can also add it to the root comment on the PR, in case you don't have a distinct commit that "fixes" it and you forget to add it before opening the PR

PR #452, opened 2017-11-30
Moved os-compatibility and added header

moritz31:
* Moved os-compatibility.c to util
* Added header under util
* Removed static since it isn't needed (i think so)
* Adjusted meson.build to include lib_wlr 

Linked issue #448

### DIFF
--- a/examples/screenshot.c
+++ b/examples/screenshot.c
@@ -35,7 +35,7 @@
 #include <limits.h>
 #include <sys/param.h>
 #include <screenshooter-client-protocol.h>
-#include "../backend/wayland/os-compatibility.c"
+#include "util/os-compatibility.h"
 
 static struct wl_shm *shm = NULL;
 static struct orbital_screenshooter *screenshooter = NULL;
--- a/include/util/os-compatibility.h
+++ b/include/util/os-compatibility.h
@@ -1,0 +1,12 @@
+#ifndef _WLR_UTIL_OS_COMPATIBILITY_H
+#define _WLR_UTIL_OS_COMPATIBILITY_H
+
+int os_fd_set_cloexec(int fd);
+
+int set_cloexec_or_close(int fd);
+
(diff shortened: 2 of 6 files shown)


examples/screenshot.c
```diff
@@ -35,7 +35,7 @@
 #include <limits.h>
 #include <sys/param.h>
 #include <screenshooter-client-protocol.h>
-#include "../backend/wayland/os-compatibility.c"
+#include "../include/util/os-compatibility.h"
```
SirCmpwn: I think this can just be

```
#include "util/os-compatability.h"
```

include/util/os-compatibility.h
```diff
@@ -0,0 +1,7 @@
+int os_fd_set_cloexec(int fd);
+
+int set_cloexec_or_close(int fd);
+
+int create_tmpfile_cloexec(char *tmpname);
+
+int os_create_anonymous_file(off_t size);
```
SirCmpwn: Add guards to this file

```
#ifndef _WLR_OS_COMPATABILITY_H
#define _WLR_OS_COMPATABILITY_H

// ...

#endif
```

moritz31: shouldn't the guard equals the file name ?

SirCmpwn: Sure, you can call it `_WLR_UTIL_OS_COMPATABILITY_H` if you want.

SirCmpwn:
Thanks. Can you squash your commits together?

moritz31:
Is there an option on github? Never squashed commits

SirCmpwn:
No. Use `git rebase -i` and squash everything into the first commit. Read the instructions git gives you. Then `git push -f`

moritz31:
think i have messed up, have to later look at

moritz31:
should be fine now
something new learned, was quite easy if one knows what to do

SirCmpwn:
Thanks!

PR #453, opened 2017-11-30
Add surface transforms support

### LINKED ISSUE #451: Transformed surfaces support
Transformed surfaces don't render correctly in rootston. If the output transform is a 90 degree rotation and the surface transform is the same, the surface shouldn't be rotated by the compositor (since it's already rotated by the client). Other combinations should be supported too.

`weston-terminal` is a client that transforms its surfaces. It's currently broken on rotated outputs.

I think there are issues with input events too.

See https://github.com/swaywm/wlroots/pull/437#issuecomment-346158774rootston

emersion:
- [x] Fix rendering of transformed surfaces
- [x] `flipped-90` and `flipped-270` are inverted
- [x] Fix input events for transformed surfaces

Test plan:
- All output transforms
- A client that transforms its surfaces like `weston-terminal`, a client that doesn't like `gnome-calculator`

Fixes #451


ascent12:
My renderer PR will completely break this.

emersion:
Yeah, exactly. I was wondering if there's a simple way to do this without matrices.

ascent12:
Maybe this would be justification enough for me to add the matrix functions back.

emersion:
This PR is ready for review.

I didn't find a way to do this without matrices while keeping a readable and easy-to-understand code.

SirCmpwn:
@ascent12 @emersion you two need to sort this out

emersion:
Bump

SirCmpwn:
This is good shit, nice work.

PR #455, opened 2017-12-02
Attempt to fix #454

### LINKED ISSUE #454: wlroots and sway tries to use a virtual terminal when XDG_SEAT != seat0
Hi

It seems that when starting on seat1 (instead of seat0) 
wlroots tries to use a virtual terminal

Virtual Terminals (TTYs) only work on seat0 , (and not on any other seat)

Thanks

n3rdopolis:
With logind, only seat0 can use TTYs

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -353,10 +353,9 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
 		wlr_log(L_ERROR, "Failed to get seat id: %s", strerror(-ret));
 		goto error;
 	}
-
 	snprintf(session->base.seat, sizeof(session->base.seat), "%s", seat);
-	if (seat == "seat0")
-	{
+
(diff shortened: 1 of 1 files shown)


backend/session/logind.c
```diff
@@ -347,19 +347,22 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
 	char *seat;
 	ret = sd_session_get_seat(session->id, &seat);
 	if (ret < 0) {
 		wlr_log(L_ERROR, "Failed to get seat id: %s", strerror(-ret));
 		goto error;
 	}
+
 	snprintf(session->base.seat, sizeof(session->base.seat), "%s", seat);
+	if (seat == "seat0")
+	{
```
SirCmpwn: Style - put this on the same line as the if statement

SirCmpwn: Also this is not how string comparison works in C. `man strcmp`

PR #457, opened 2017-12-03
wl-shell: add commit event

acrisci:
Like xdg-shell commit event, makes sure the user commit handler always runs after the surface commit handlers.


emersion:
Can you update rootston's `handle_surface_commit` to use this?

PR #458, opened 2017-12-03
wrap server decoration modes

acrisci:
So we don't make sway use wayland-scanner to get this header.

### DIFF
--- a/include/wlr/types/wlr_server_decoration.h
+++ b/include/wlr/types/wlr_server_decoration.h
@@ -3,12 +3,35 @@
 
 #include <wayland-server.h>
 
+/**
+ * Possible values to use in request_mode and the event mode. Same as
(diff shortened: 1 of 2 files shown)


include/wlr/types/wlr_server_decoration.h
```diff
@@ -3,12 +3,31 @@
 
 #include <wayland-server.h>
 
+/**
+ * Possible values to use in request_mode and the event mode. Same as
+ * org_kde_kwin_server_decoration_manager_mode.
+ */
+enum wlr_server_decoration_manager_mode {
+	/**
+	 * Undecorated: The surface is not decorated at all, neither server nor client-side. An example is a popup surface which should not be decorated.
```
emersion: 80 char line width

PR #460, opened 2017-12-03
xwayland: xsurface commit

acrisci:
Not strictly necessary (yet) but might be one day, and makes it like the other shells.

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -106,6 +106,7 @@ struct wlr_xwayland_surface {
 	bool has_alpha;
 
 	struct {
+		struct wl_signal commit;
 		struct wl_signal destroy;
 		struct wl_signal request_configure;
 		struct wl_signal request_move;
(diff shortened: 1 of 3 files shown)


emersion:
I'm not sure about this: given how xwayland works, will we ever need it? State isn't double-buffered in X11.

emersion:
After all, you're right. We should have a best-effort approach even if X11 doesn't really double-buffer state.

SirCmpwn:
Until this actually does something meaningful let's not.

acrisci:
This actually does make things a bit easier now that I think about it. The xwayland surface is ephemeral so it means the user doesn't have to unsubscribe/resubscribe to the surface events on map and unmap.

emersion:
>The xwayland surface is ephemeral so it means the user doesn't have to unsubscribe/resubscribe to the surface events on map and unmap.

Yes, but the compositor still needs to add and remove the view from the desktop when it's mapped and unmapped. So I don't think it adds a lot of value (apart from saving one line of code).

In the end, other shells define particular semantics for the surface commit event, but xwayland doesn't. It makes sense not to add it, at least for now.

acrisci:
I think the ultimate solution here is for wlr-surface to have a special function for role surfaces to set that always runs before wlr-surface commit and get rid of all the role surface commit events.

emersion:
Yeah, that makes sense since a surface can only have one role.

PR #468, opened 2017-12-06
Use destroy listeners on wl_display

### LINKED ISSUE #467: Use destroy listeners on the wl_display
We never use `wl_display_add_destroy_listener` anywhere in our codebase.
If we changed our code to use it, it could simplify proper cleanup of wlroots and have it just become `wl_display_terminate(disp)` for the library users.

We'll probably need to investigate how appropriate these destroy listeners are for our uses. I don't know if there are any cases where one type NEEDS be destroyed before another, and whether this may have an effect on that.

This is just a point for discussion, but you could even take it a step further and remove the `wlr_type_destroy' functions from our API.
Maybe there is a use case for destroying some types before the wl_display is terminated, though.

emersion:
Fixes #467


include/wlr/types/wlr_screenshooter.h
```diff
@@ -5,13 +5,21 @@
 struct wlr_screenshooter {
 	struct wl_global *wl_global;
 	struct wlr_renderer *renderer;
+	struct wl_list screenshots; // wlr_screenshot::link
+
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal destroy;
```
emersion: Not sure this is actually needed

emersion:
This PR is ready for review.

I've added display destroy listeners on all types that create globals.

I've noticed that cleanup is not handled correctly in all types, that's another long-term issue that needs to be solved.

ascent12:
This doesn't include the backends.

emersion:
Added backends and xwayland.

The `multi` backend still needs to be free'd manually because it hasn't access to the display. I didn't implemented the display destroy listener in `wlr_egl` because the backend has ownership over it (we should really implement this on globals, stuff that is a child of a global shouldn't do it).

I spent a fair amount of time debugging xwayland, it turns out calling `wl_client_destroy` makes the compositor segfault:

```
#0  0x0000555555a41be0 in  ()
#1  0x00007ffff1bf9813 in  () at /usr/lib/libEGL_mesa.so.0
#2  0x00007ffff7967c12 in  () at /usr/lib/libwayland-server.so.0
#3  0x00007ffff796c619 in  () at /usr/lib/libwayland-server.so.0
#4  0x00007ffff7967cff in wl_client_destroy () at /usr/lib/libwayland-server.so.0
#5  0x00007ffff7b8a0a8 in wlr_xwayland_finish (wlr_xwayland=0x555555f55950) at ../xwayland/xwayland.c:141
#6  0x00007ffff7b8a005 in wlr_xwayland_destroy (wlr_xwayland=0x555555f55950) at ../xwayland/xwayland.c:328
#7  0x00007ffff7b8a9a5 in handle_display_destroy (listener=0x555555f559b0, data=0x555555773700) at ../xwayland/xwayland.c:188
#8  0x00007ffff7967b82 in  () at /usr/lib/libwayland-server.so.0
#9  0x00007ffff7968177 in wl_display_destroy () at /usr/lib/libwayland-server.so.0
#10 0x000055555555f041 in main (argc=1, argv=0x7fffffffe648) at ../rootston/main.c:73
```

Westion doesn't call `wl_client_destroy`. I just commented it for now, I'm out of ideas.

emersion:
I did more research on this. Mutter doesn't destroy the client either. wlc does in `wlc_xwayland_terminate` but this function is called before `wl_display_destroy`.

I believe the segfault happens because the backend is destroying `wlr_egl`. Destroying a client after that segfaults (even if `eglUnbindWaylandDisplayWL` is called).

The issue is that clients are not destroyed by `wl_display_destroy`. Destroying them in a display destroy listener is too late. A patch was proposed to Wayland to destroy clients in `wl_display_destroy` (just before emitting the destroy event), but was rejected because it's an ABI change: https://patchwork.freedesktop.org/patch/128832/

What was proposed instead is to add a new function `wl_display_destroy_all_clients` that should be called by the compositor before `wl_display_destroy`. I've implemented this function, it works pretty well except xwayland is respawned automatically when its client disconnects (https://github.com/swaywm/wlroots/blob/master/xwayland/xwayland.c#L140).

So we have two solutions:
- Do not destroy clients when the compositor exits. Make xwayland leak its client if destroyed in a display destroy listener, because it's too late to destroy clients there. That's what I implemented so far.
- Add `wlr_display_destroy_all_clients` (maybe propose a Wayland patch for `wl_display_destroy_all_clients`) and do not respawn xwayland automatically. This should effectively cleanup everything properly.

emersion:
Also note that `wlr_backend_destroy` should still be called before `wl_display_destroy` because some backends (`multi`) don't have access to the `wl_display`.

I added a display destroy listener for the session.

I think this PR is ready for another review.

emersion:
Bump

types/wlr_screenshooter.c
```diff
@@ -144,6 +160,25 @@ static void screenshooter_bind(struct wl_client *wl_client, void *data,
 		screenshooter, NULL);
 }
 
+void wlr_screenshooter_destroy(struct wlr_screenshooter *screenshooter) {
```
SirCmpwn: This and those like it should be exported so users can do it if they want to

emersion: Already done

```c
void wlr_screenshooter_destroy(struct wlr_screenshooter *screenshooter);
```

emersion: I'm pretty sure all destructors are exported btw

emersion:
For reference, I submitted a patch to Wayland to properly destroy all clients: https://patchwork.freedesktop.org/patch/193088/

Issue #403, opened 2017-11-10
Proposal: configure devices with their name and their connector


emersion:
## Input devices

Currently input devices are configured with their `wlr_input_device.name` - but if you have two identical devices, you can't configure them independently:

```
[backend/libinput/events.c:73] Added Metadot - Das Keyboard Das Keyboard [9456:320]
[backend/libinput/events.c:73] Added Metadot - Das Keyboard Das Keyboard [9456:320]
```

Unfortunately, inputs devices don't seem to have a serial number - only a friendly name, manufacturer and model.

Proposal: being able to configure them by their connector (retrieved from udev). We could add a `char *connector` field to `wlr_input_device` for that.

- [ ] Find a way to get the connector name of an input device and expose it
- [ ] Allow rootston config to use this to configure input devices

## Ouputs

Outputs are always identified by their connector for now: `VGA-1`, `HDMI-1-A`... However, if you have for instance two DisplayPort connectors, it can make more sense to configure outputs by their name (just like input devices right now). Also outputs have a serial number which can be used to uniquely identify them (given a valid EDID).

Proposal: being able to configure them by their name (manufacturer, model and serial). The `wlr_output` currently has a `name` field used for the connector name - we could rename it to `connector` and add a `serial` field (`make` and `model` are already here).

- [x] Add a `serial` field
- [x] ~~Find a better name for the `name` field~~
- [x] Allow rootston config to use `make`, `model` and `serial` to configure outputs

ascent12:
>However, if you have for instance two DisplayPort connectors, it can make more sense to configure outputs by their name

The names already describe a physical connector on a GPU so there is already no ambiguity, but I do see the value in configuring a monitor based on information about the monitor itself, instead of where you plugged it in.  
Maybe you have some external monitor that is only plugged in occasionally and may not use the same connector every time. Also, DP MST can sometimes renumber the DP connectors in different ways if your setup changes, breaking your configs (There is a way to get more stable naming for these, but we currently don't do it).

None of this configuration needs to live in wlroots, though. We already make users configure it from a `wlr_output`, so they can make their decision based on whatever fields they want. Internally, `name` is meaningless.

The only thing that would actually require wlroots changes is adding a serial number, which is a pretty minor change. DRM is the only backend you could get this information. Although, I suppose it's the only backend that would find it useful too.

>currently has a `name` field used for the connector name - we could rename it to `connector`

I don't see the point in doing this in particular.

emersion:
[...]

PR #469, opened 2017-12-07
Configure outputs with make, model, serial in rootston

emersion:
This allows for instance to configure the screen you plug in your HDMI connector at home differently from the one you plug in the same connector at work.

Also added fallbacks in DRM backend in case EDID extension data for model and serial is missing.

Test plan: get your output name in rootston logs:

```
2017-12-07 13:57:23 - [rootston/output.c:293] Output 'LVDS-1' added
2017-12-07 13:57:23 - [rootston/output.c:296] 'Chimei Innolux Corporation 0x1484 0x00000000' 310mm x 174mm
```

Add it to your config:

```ini
[output:Chimei Innolux Corporation 0x1484 0x00000000]
scale = 2
```

Check that the config is applied to the output.

Updates #403

### DIFF
--- a/backend/drm/util.c
+++ b/backend/drm/util.c
@@ -93,6 +93,12 @@ void parse_edid(struct wlr_output *restrict output, size_t len, const uint8_t *d
 	uint16_t id = (data[8] << 8) | data[9];
 	snprintf(output->make, sizeof(output->make), "%s", get_manufacturer(id));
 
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Personally I'd rather use connector names in my config. Can you make this support both forms?

emersion:
Yeah that's still the case!

SirCmpwn:
Sweet.

PR #470, opened 2017-12-07
wlr-edges

acrisci:
Refactor to use wlr-edges to take some code out of rootston.


include/wlr/types/wlr_xcursor_manager.h
```diff
@@ -50,4 +51,10 @@ struct wlr_xcursor *wlr_xcursor_manager_get_xcursor(
 void wlr_xcursor_manager_set_cursor_image(struct wlr_xcursor_manager *manager,
 	const char *name, struct wlr_cursor *cursor);
 
+/**
+ * Get the name of the cursor image for the given edges.
+ */
+const char *wlr_xcursor_manager_get_resize_name(enum wlr_edges edges);
```
emersion: Should this be in `wlr_xcursor` instead?

include/wlr/util/edges.h
```diff
@@ -0,0 +1,12 @@
+#ifndef WLR_UTIL_EDGES_H
+#define WLR_UTIL_EDGES_H
+
+enum wlr_edges {
+	WLR_EDGE_NONE = 0,
+	WLR_EDGE_TOP = 1,
+	WLR_EDGE_BOTTOM = 2,
+	WLR_EDGE_LEFT = 4,
+	WLR_EDGE_RIGHT = 8,
+};
```
SirCmpwn: This should go into xcursor instead of getting its own header.

emersion: We discussed this in IRC and decided that since it's used in xwayland it shouldn't be (only) in xcursor (to keep xwayland and xcursor separate).

include/wlr/types/wlr_xcursor_manager.h
acrisci: I put it here to be as close as possible to the thing that uses it. Nothing in wlr-xcursor uses this. I'll give you the final decision on it.

emersion: Well, nothing in the xcursor manager uses this. The xcursor manager is only an abstraction around `wlr_xcursor_theme` and `wlr_cursor`.

You could use `get_resize_name` to get the name of the cursor and then call `wlr_xcursor_theme_get_cursor`. If a compositor doesn't want to use `wlr_xcursor_manager`, it could want to use `get_resize_name` with `wlr_xcursor_theme`.

include/wlr/util/edges.h
SirCmpwn: Why would we need to keep xwayland and xcursor separate?

emersion: Because we're building a modular Wayland library and `wlr_xcursor` is optional for compositors. If we put this enum in `wlr_xcursor`, compositors not using `wlr_xcursor` will still need to include it to get the enum for xwayland. `wlr_xcursor` and xwayland are unrelated.

SirCmpwn: I don't like either solution. This is modularity to the point of absurdity. I guess we'll just leave it here.

include/wlr/types/wlr_xcursor_manager.h
acrisci: Ok that makes sense.

include/wlr/util/edges.h
acrisci: This is a consequence of not having a single shell abstraction. If we did, I think it would belong there because resize edges seems more like a shell concept to me. The only source of this data structure is a shell event.

PR #473, opened 2017-12-08
Send wl_surface.{enter,leave} to cursor surfaces

### LINKED ISSUE #471: wl_surface.{enter,leave} not sent to cursor surfaces
I believe this is the root cause of https://bugzilla.gnome.org/show_bug.cgi?id=790168

I analyzed WAYLAND_DEBUG logs from gnome-calculator on rootston vs. weston and noticed these events are not sent to the cursor surface.

emersion:
Test plan: setup a scaled output, spawn `gnome-calculator`. Cursor should be scaled too (and not blurry).

`weston-terminal` doesn't send a scaled cursor, so its cursor is just scaled up (same behavior on Weston).

I changed rootston cursor size from 16 to 24 because the default theme doesn't have 16x16 cursors and fallbacks to 24x24 (that caused issues when loading the theme with a scale factor of 2).

Fixes #471

### DIFF
--- a/include/rootston/xcursor.h
+++ b/include/rootston/xcursor.h
@@ -3,7 +3,7 @@
 
 #include <stdint.h>
 
-#define ROOTS_XCURSOR_SIZE 16
+#define ROOTS_XCURSOR_SIZE 24
 
 #define ROOTS_XCURSOR_DEFAULT "left_ptr"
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

PR #475, opened 2017-12-08
Renderer redesign v2

ascent12:
This is the same code as #319, but it has been rebased and cleaned up to make the commits more logical.

Now, I'm probably as sick as the rest of you for the old one not being merged yet, and it's seriously getting in the way of other PRs and porting sway.
There are just so many little changes to a lot of code I don't really understand the requirements for, so it was just taking forever.

So to try and remedy this, here is a slightly different strategy:
This PR adds all of the new renderer stuff, but leaves the old one in, and does not have the changes to rootston and a couple of other areas that were causing problems.
(As a consequence of leaving the old one in, the new matrix functions are called matrix2; they can be renamed later).
Once everything is finally changed to new renderer, then the old one can be deleted.


SirCmpwn:
-10 to this approach. This is a real mess. We should just do it all in one go and if we miss stuff sort out the bugs as they're found.

ascent12:
Yes, it's a mess in the sense that there are two renderers. The commits themselves are actually much cleaner.

Of all of the outstanding issues in the first PR, most of them are rootston related. I hardly know shit about rootston and I don't really have the capacity to dig deeper and fix the deeper layers of rootston right now.
Also, I don't think "real" work (i.e. sway, other wlroots changes) should be blocked on some mere demo code. I agree that demo code is important, just not **blocking**.

If you really want the "all in one go" approach, go with the original PR. The issues aren't horribly breaking, and only really affect some more esoteric uses of rootston (mostly rotated views). It is merging knowingly "broken" code though, which almost anyone would have an issue with.

SirCmpwn:
Ugh, what a pain. I consider it a bug if rootston and wlroots are out of sync just as much as any of the bugs the other PR comes with. This patch has been a real thorn in everyone's sides. Can you please step up and put in the work to integrate it properly?

ascent12:
Yes, I know I'm being slow as fuck, annoying to work with, and probably pissing everybody off, but that's exactly why I made this PR.

Basically, I just want to get the shit that works merged and give other people a chance to design their code against the new interface. I'm going to be somewhat busy this weekend, so the other code is probably not going to get polished in the very near future.
And then a whole bunch of other things are going to get merged (#453 particularly), I'm going to get more and more merge conflicts, and then probably do a shitty job resolving them, not give it the testing it deserves, and introduce new bugs.
I've spend so much fucking time resolving merge conflicts on this PR.

Now as a consequence, since it's not merging the currently broken code, the old renderer has to stay for now, since it's still being used. Obviously I don't think it should stay that way.

Since the old and new renderer share the EGL resources, things aren't broken as you may expect them to be. They can currently live side by side, until rootston and the other small bits of code can be cleaned up.

SirCmpwn:
There isn't a happy medium here. You're giving me two different broken strategies to pick from. I choose neither. I will close both PRs before I merge a broken change.

ascent12:
It's not broken: it's incomplete. The old code **still works** and gets along just fine with the new code.
The reason I'm presenting things like this is because it's **not** broken, at least as far as I'm aware.

If you took out ee9be9cc47f6679490eb5d2bca4d86245628e596 and 160418e09ca6562bec82af828955b32d42f7ac98, you could classify this PR as simply "Add an alternative renderer".
Yes, obviously the goal is to get things on the new renderer, but it doesn't actually need to happen at the same time. Doing this is one monolithic go has just not worked so far.

The longer it takes, the more work needs to be done, making it take longer.

SirCmpwn:
Doing it all in one go could have worked fine! It was *this close*. All that we needed was matrix functions put back in. I would have merged it right there. I'm not merging this. Fix the other one or close it and we'll live with the old renderer.

PR #478, opened 2017-12-10
wlr_seat: send keyboard modifiers to the right client on enter

### LINKED ISSUE #476: modifiers not updated properly when changing focus
Example: in rootston, run mpv --opengl-backend=wayland <some video>, press alt, click on another surface, release alt, click back on mpv window, press space: mpv complains there is no key bound to Alt+SPACE instead of pausing.

This doesn't seem to happen on xwayland window, but as far as I can tell happens on wl_shell as well.
Looking a bit further it seems that we try to work around this in `types/wlr_seat.c:849`: in the code path entering a new surface, we explicitely call `wlr_seat_keyboard_send_modifiers(seat);` to update it... But looking a bit harder it looks like the leaving surface gets that final `wl_keyboard_send_modifiers()` and not the newly focused surface. That is because `seat->keyboard_state.focused_client` is only updated AFTER we call that `wlr_seat_keyboard_send_modifiers()`, which uses `seat->keyboard_state.focused_client` to tell where to send the event.

Proposal: send modifier later, or make a new function that takes client as argument.
This works:
```diff
$ git diff
diff --git a/types/wlr_seat.c b/types/wlr_seat.c
index c64cbef..a126d12 100644
--- a/types/wlr_seat.c
+++ b/types/wlr_seat.c
@@ -846,7 +846,6 @@ void wlr_seat_keyboard_enter(struct wlr_seat *seat,
                        surface->resource, &keys);
                wl_array_release(&keys);
 
-               wlr_seat_keyboard_send_modifiers(seat);
                wlr_seat_client_send_selection(client);
        }
 
@@ -868,6 +867,10 @@ void wlr_seat_keyboard_enter(struct wlr_seat *seat,
 
        seat->keyboard_state.focused_client = client;
        seat->keyboard_state.focused_surface = surface;
+
+       if (client && client->keyboard && seat->keyboard_state.keyboard) {
+               wlr_seat_keyboard_send_modifiers(seat);
+       }
 }
 
 void wlr_seat_keyboard_notify_enter(struct wlr_seat *seat, struct
```

Will submit a PR once one of the solution has been picked.

martinetd:
We were previously sending modifiers to the leaving client instead.
Fixes #476.

As briefly mentioned over here, I might feel more comfortable with less seat abstraction (letting us pick the destination client instead of forcing onto focused one in `wlr_seat_keyboard_send_modifiers`, could make it an optional-ish argument by saying NULL client would be focused one?), but this works without changing the API.

### DIFF
--- a/types/wlr_seat.c
+++ b/types/wlr_seat.c
@@ -860,7 +860,6 @@ void wlr_seat_keyboard_enter(struct wlr_seat *seat,
 			surface->resource, &keys);
 		wl_array_release(&keys);
 
-		wlr_seat_keyboard_send_modifiers(seat);
 		wlr_seat_client_send_selection(client);
 	}
 
@@ -882,6 +881,12 @@ void wlr_seat_keyboard_enter(struct wlr_seat *seat,
 
 	seat->keyboard_state.focused_client = client;
 	seat->keyboard_state.focused_surface = surface;
+
+	if (client && client->keyboard && seat->keyboard_state.keyboard) {
+		// tell new client about any modifier change last,
+		// as it targets seat->keyboard_state.focused_client
+		wlr_seat_keyboard_send_modifiers(seat);
+	}
 }
 
 void wlr_seat_keyboard_notify_enter(struct wlr_seat *seat, struct


emersion:
-1 to having a focused client _and_ being able to override it

PR #480, opened 2017-12-11
Set default cursor image in rootston

emersion:
Test plan: setup two seats, and use this config:

```ini
[cursor:seat1]
default-image = crosshair
```


acrisci:
👍

PR #483, opened 2017-12-12
Update output layout when scale or transform changes

emersion:


types/wlr_output_layout.c
```diff
@@ -134,6 +138,18 @@ static void handle_output_resolution(struct wl_listener *listener, void *data) {
...
+static void handle_output_scale(struct wl_listener *listener, void *data) {
+	struct wlr_output_layout_output_state *state =
+		wl_container_of(listener, state, scale);
+	wlr_output_layout_reconfigure(state->layout);
+}
+
+static void handle_output_transform(struct wl_listener *listener, void *data) {
+	struct wlr_output_layout_output_state *state =
+		wl_container_of(listener, state, transform);
+	wlr_output_layout_reconfigure(state->layout);
```
acrisci: I don't know if it matters too much, but certain transform changes won't require a layout update like normal to flipped.

emersion: This would require to track the previous output transform. Not sure if it's worth it.

acrisci: 👍

PR #485, opened 2017-12-13
Multiple outputs with Wayland backend

emersion:
- [x] Remove output instead of terminating display when a wayland backend view closed
- [x] Fix pointer events

Test plan: `env WLR_WL_OUTPUTS=2 build/rootston/rootston`, check with one output with scale=2 and rotate=90.

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -1,6 +1,7 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <assert.h>
+#include <limits.h>
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
 #include <wayland-server.h>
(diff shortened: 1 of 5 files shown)


emersion:
This PR is ready for review.

backend/wayland/backend.c
```diff
@@ -117,6 +118,38 @@ struct wlr_wl_backend_output *wlr_wl_output_for_surface(
 	return NULL;
 }
 
+void wlr_wl_output_layout_get_box(struct wlr_wl_backend *backend,
+		struct wlr_box *box) {
```
SirCmpwn: What the heck is this?

emersion: Yeah, I needed to re-implement this function. Basically we send absolute events, which means pointer events are mapped to the whole output layout. Each window acts like a tablet tool, so we need to re-map each "tablet tool" to the right region.

SirCmpwn: Aren't pointer events (even absolute ones) relative to the surface the pointer is over?

emersion: There's no "surface hovered by the pointer".

Imagine a two-output setup with one tablet pad. If you click on the left part of the tablet pad, it ends up on the left output - same for the one on the right.

The wayland backend, started with two windows, has two outputs and two tablet pads. But you want to map each tablet pad to an output, not each tablet pad to the whole output layout.

PR #486, opened 2017-12-13
xwm: user custom event handler

acrisci:
Custom user event handler for the wlr-xwayland xwm. A compositor can use this to implement whatever extra xwayland interfaces they want to or override the behavior of existing ones.


SirCmpwn:
Nice, thanks.

Issue #295, opened 2017-10-19
xwayland clipboard sync and dnd


acrisci:
Whenever the wayland clipboard selection changes for a seat (selection-change event), we need to get the data source, read out the data from the fd and set the xwayland clipboard selection to its contents.

Whenever the X11 CLIPBOARD selection changes, we need to create a wayland data source, override the source methods, and set it on the seat.

A similar process will need to be done for the dnd selection (see the Weston xwayland implementation for guidance).

SirCmpwn:
We should be careful not to do this if xwayland is not available. We should also probably only commit the sync when focus switches between wayland and xwayland views.

emersion:
>We should also probably only commit the sync when focus switches between wayland and xwayland views.

`xclip` may be used in a wayland terminal, without switching focus between a wayland view and a xwayland one.

SirCmpwn:
That's a good point, but it still presents a security concern. Let's make it toggleable.

acrisci:
> Let's make it toggleable.

FYI, the wayland spec says to share the clipboard selection with windows only at the time they receive keyboard focus (and when the selection changes while a window has keyboard focus) which is the minimum amount of visibility required for a reasonably usable clipboard. So something like xclip is not possible using only core wayland protocols.

On X11, I think the clipboard is managed through acquiring a selection on the root window, but there would be no way to prove that the requester ever had keyboard focus.

I think we should find or create a standard for people who want something like xclip to work and understand the security implications of this. I find myself typing something like `curl $(xclip -o)` sometimes.

SirCmpwn:
>FYI, the wayland spec says to share the clipboard selection with windows only at the time they receive keyboard focus (and when the selection changes while a window has keyboard focus) which is the minimum amount of visibility required for a reasonably usable clipboard. So something like xclip is not possible using only core wayland protocols.

I think we should do this. Not every tool is going to work straight up, especially if they pose a security threat.

>I think we should find or create a standard for people who want something like xclip to work and understand the security implications of this. I find myself typing something like curl $(xclip -o) sometimes.

Sway, for one, now does this through IPC/swaymsg. I don't think a standard is necessary yet.

acrisci:
Anyone who wants to take this up should look at #440 for a poc

emersion:
Can confirm that apart from rich clipboard and the security feature, everything works fine now.

emersion:
References:
[...]

PR #487, opened 2017-12-14
xwm clipboard sync, 2nd edition

emersion:
- [x] MIME type from xwayland to wayland
- [x] MIME type from wayland to xwayland
- [x] Fix `xclip`
- [x] Don't send selection to clients that don't have focus

Weston doesn't implement selection MIME types, so we can't use it as a starting point.

Test plan:
* Copy text in wayland, paste it in xwayland
* Copy text in xwayland, paste it in wayland
* Copy text in xwayland, paste it in xwayland
* Same with rich text
* Copy something, run `xclip -o -t TARGETS -selection clipboard`, check MIME types
* Copy something, run `xclip -o -selection clipboard`
* Run `echo "poi" | xclip -i -selection clipboard`, paste somewhere

Test clients:
* Wayland: gedit, epiphany, lowriter
* Xwayland: gedit, chromium, lowriter

Updates #295


emersion:
I tried to implement the security switch - but I didn't find a way to do it. Basically, the window that requests the clipboard content isn't the same as the focused window. Tried to get window parents, that didn't work either.

Apart from that, everything is ready to be tested.

SirCmpwn:
>Basically, the window that requests the clipboard content isn't the same as the focused window.

What? How does that work?

emersion:
This is ready for review.

SirCmpwn:
This is super buggy for me. I tried testing a lot of different situations but I couldn't get anything to work consistently. Some time after I copied (and attempted to paste) content into lowriter, it would pop up "Requested clipboard format is not available", and epiphany was logging "error reading selection buffer: operation was cancelled". I got gnome-terminal -> chromium to work once, then never again, and I never got chromium -> gnome-terminal to work. I got rich text from epiphany -> lowriter to work once, but then never again.

emersion:
These bugs are weird. Xwayland -> wayland wasn't working in some cases because of the security feature, so I fixed it, but wayland -> xwyaland should have worked. Can you try again and specify from which shell and to which shell you try to copy-paste? (lowriter and GTK apps are available on both, so it's hard to guess) Also, some errors are logged - are there any `selection.c`-related logs in your rootston output?

What I tried with the last commit (everything working as expected):
* Text and rich text, chromium (xwayland) <-> lowriter (wayland)
* Text, chromium (xwayland) <-> gedit (wayland)
* Text, chromium (xwayland) <-> gedit (xwayland)
* Text and rich text, epiphany (wayland) <-> lowriter (xwayland)
* Text and rich text, chromium (xwayland) <-> lowriter (xwayland)
* Image, chromium (xwayland) -> gimp (xwayland)
* Image, epiphany (wayland) -> gimp (xwayland) (use "Copy image" in the menu, not "Copy")

@acrisci Can you try this PR please?

acrisci:
It's kind of buggy with gnome-terminal running in X11 mode. Copying from gnome-terminal x11 to gnome-calculator wayland doesn't work for me. It might be a client bug, because copying from gnome-calculator with x11 mode works correctly. I think this might only apply to the first gnome-terminal x11 I open because it seems to work when I open a new one.

I don't remember what I did, but I just got gnome-calcuator with x11 to crash by doing random stuff.

```
(gnome-calculator:30077): Gtk-WARNING **: Invalid text buffer iterator: either the iterator is uninitialized, or the characters/pixbufs/widgets in the buffer have been modified since the iterator was created.
You must use marks, character numbers, or line numbers to preserve a position across buffer modifications.
You can apply tags and insert marks without invalidating your iterators,
but any mutation that affects 'indexable' buffer contents (contents that can be referred to by character offset)
will invalidate all outstanding iterators

(gnome-calculator:30077): GtkSourceView-CRITICAL **: gtk_source_buffer_real_insert_text: assertion 'gtk_text_iter_get_buffer (iter) == buffer' failed

(gnome-calculator:30077): Gtk-CRITICAL **: gtk_text_buffer_get_iter_at_mark: assertion 'GTK_IS_TEXT_MARK (mark)' failed

(gnome-calculator:30077): Gtk-CRITICAL **: gtk_text_buffer_delete_mark: assertion 'GTK_IS_TEXT_MARK (mark)' failed

[1]  + 30077 segmentation fault (core dumped)  gnome-calculator --display :1
```

Copying from weston-terminal to gnome-calculator x11 seems to fail like 20% of the time when I focus it with the right mouse button, the paste is grayed out. If I click on the window again, it usually work correctly again.

emersion:
>Copying from gnome-terminal x11 to gnome-calculator wayland doesn't work for me. It might be a client bug, because copying from gnome-calculator with x11 mode works correctly.

I can reproduce. Will investigate.

>Copying from weston-terminal to gnome-calculator x11 seems to fail like 20% of the time when I focus it with the right mouse button, the paste is grayed out. If I click on the window again, it usually work correctly again.

Are you referring to this?

screenshot from 2017-12-17 19-52-13

Notice that other items (e.g. select all, insert emoji) are also greyed out. I think that's because the text input is not focused, so it can't paste/select all/insert emoji. So I don't think it's related to the clipboard.

emersion:
>Copying from gnome-terminal x11 to gnome-calculator wayland doesn't work for me. It might be a client bug, because copying from gnome-calculator with x11 mode works correctly.

I can reproduce this one on GNOME, so this is probably a client bug.

@SirCmpwn, want to give it another try?

SirCmpwn:
1. Run rootston (wayland backend)
2. Start gnome-terminal and copy some text from it
3. Start another gnome-terminal and attempt to paste

fails

SirCmpwn:
gedit->gnome-terminal works, gnome-terminal->gedit does not

SirCmpwn:
plaintext: chromium->gedit works, gedit->chromium works

rich text from chromium to lowriter works

copy image from epiphany (wayland) -> gimp (xwayland) does not work

SirCmpwn:
copy image from epiphany to lowriter works

emersion:
As discussed on IRC, the only remaining remaining bug is `gnome-terminal`.

What happens is that `gnome-terminal` is retarded. It spawns a server _attached to the first compositor it finds_ apparently. So if you copy something in `gnome-terminal` within rootston, you'll be able to paste it in your outer Wayland session, not in rootston. That happens also if there's no `gnome-terminal` running in the outer session, and even if there's just another session on another VT. One weird thing that confirms it: the preferences window pops up on the outer session too.

All of this also happens if `gnome-terminal` is used with xwayland.

emersion:
Note for posterity: a way to make it work correctly is to spawn the server in the right place:

```shell
/usr/lib/gnome-terminal/gnome-terminal-server &
gnome-terminal
```

SirCmpwn:
Thanks!

PR #488, opened 2017-12-15
Add fractional scaling support

emersion:
- [x] Fix wayland backend pointer events

Test plan:

```ini
[output:XXX]
scale = 1.5
```

Try various wayland and xwayland apps. Check rendering and input events. Try fullscreen.

### DIFF
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -59,11 +59,11 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		&box.width, &box.height);
 	box.x = wl_fixed_to_int(surface_x);
 	box.y = wl_fixed_to_int(surface_y);
-	struct wlr_box transformed;
 
+	struct wlr_box transformed;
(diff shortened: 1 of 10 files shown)


SirCmpwn:
I know this says WIP, but I tested it anyway so here are my results.

I set scale=2.5 on a 4K output and got rendering artifacts and gedit crashes:

wayland-screenshot

SirCmpwn:
I got similar results when scale=3 but not when scale=2. I wonder if it's a GTK bug...

SirCmpwn:
Yeah, I can reproduce all of that on master as well.

emersion:
Okay, so this isn't related to fractional scaling right? I only tested with scale=1.5. Does that work for you?

SirCmpwn:
Yeah, 1.5 works. Sorry for the noise, I can reproduce the GTK issues on Weston too so I'll file a bug with them.

emersion:
This PR is ready for review.

SirCmpwn:
https://bugzilla.gnome.org/show_bug.cgi?id=791656

SirCmpwn:
Thanks!

PR #492, opened 2017-12-16
Support multiple resources per seat client

### LINKED ISSUE #270: Support winit
Support [winit](https://github.com/tomaka/winit) gui toolkit and do QA so that it is reasonably well supported on wlroots. This should make [alacritty](https://github.com/jwilm/alacritty) terminal emulator work.

Currently doesn't work on weston because of xdg-shell protocol violations [(issue)](https://github.com/tomaka/winit/issues/306). Currently doesn't work on wlroots because of rendering issues.

cc @Timidger

emersion:
Alacritty still doesn't receive keyboard events, but is built with an older winit. winit examples, in other hand, don't receive keyboard/pointer events on master but receive them in this branch.

A PR updates Alacritty with the latest winit: https://github.com/jwilm/alacritty/pull/811. Everything works as expected with this branch and a newer winit.

Test plan: check that pointer/keyboard/touch events still works, check that wayland-to-wayland keyboard/drag'n'drop still works.

Fixes #270

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -16,9 +16,11 @@ struct wlr_seat_client {
 	struct wl_client *client;
 	struct wlr_seat *seat;
 
-	struct wl_resource *pointer;
-	struct wl_resource *keyboard;
-	struct wl_resource *touch;
+	// lists of wl_resource
+	struct wl_list pointers;
+	struct wl_list keyboards;
+	struct wl_list touches;
+
 	struct wl_resource *data_device;
 
 	struct {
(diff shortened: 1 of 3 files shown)


SirCmpwn:
This works, but alacritty is kind of wonky. Ends up looking like this:

wayland-screenshot

cc @vberger, this smells like an alacritty/winit bug to me.

wayland-to-wayland dnd is broken for me on the wayland backend with this patch.

emersion:
>this smells like an alacritty/winit bug to me.

Yeah, this is a bug specific to the Alacritty PR (see the PR's comments).

PR #496, opened 2017-12-17
Headless backend

### LINKED ISSUE #495: Headless backend design proposal
Very straightforward backend:

```c
struct wlr_backend *wlr_headless_backend_create();

struct wlr_output *wlr_headless_add_output(unsigned int width, unsigned int height);

struct wlr_input_device *wlr_headless_add_input(enum wlr_input_device_type type);
```

If you want to simulate events on it, just dereference the pointers and raise signals yourself. Override the functions on the type impl too if you want to e.g. handle calls to `wlr_keyboard_impl.led_update`. If you want to take screenshots of the output, use wlr_render_read_pixels or just wire up screenshooter to it.

emersion:
Test plan: launch rootston, open `gnome-calculator`, take a screenshot.

Fixes #495

### DIFF
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -26,6 +26,9 @@ static void backend_destroy(struct wlr_backend *wlr_backend) {
 	}
 
 	wl_list_remove(&backend->display_destroy.link);
+
+	// TODO: destroy outputs
+
 	wlr_egl_finish(&backend->egl);
 	free(backend);
 }
@@ -61,6 +64,12 @@ static bool egl_get_config(EGLDisplay disp, EGLConfig *out) {
 
 	static const EGLint attribs[] = {
 		EGL_SURFACE_TYPE, EGL_PBUFFER_BIT,
+		EGL_BUFFER_SIZE, 32,
+		EGL_ALPHA_SIZE, 0,
+		EGL_BLUE_SIZE, 8,
+		EGL_GREEN_SIZE, 8,
+		EGL_RED_SIZE, 8,
+		EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
 		EGL_NONE,
 	};
 
@@ -146,6 +155,7 @@ struct wlr_backend *wlr_headless_backend_create(struct wl_display *display) {
 	}
 	wlr_backend_init(&backend->backend, &backend_impl);
 	backend->display = display;
+	wl_list_init(&backend->outputs);
 
 	egl_init(&backend->egl);
 
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -1,0 +1,156 @@
+#include <stdlib.h>
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#include <wlr/util/log.h>
+#include "backend/headless.h"
+#include "glapi.h"
+
+static bool backend_start(struct wlr_backend *wlr_backend) {
+	struct wlr_headless_backend *backend =
+		(struct wlr_headless_backend *)wlr_backend;
+	wlr_log(L_INFO, "Initializating headless backend");
+
+	// TODO
+	for (size_t i = 0; i < 1; ++i) {
+		wlr_headless_add_output(&backend->backend, 1280, 720);
+	}
+
+	return true;
+}
+
+static void backend_destroy(struct wlr_backend *wlr_backend) {
+	struct wlr_headless_backend *backend =
+		(struct wlr_headless_backend *)wlr_backend;
+	if (!wlr_backend) {
+		return;
+	}
+
+	wl_list_remove(&backend->display_destroy.link);
+	wlr_egl_finish(&backend->egl);
+	free(backend);
+}
+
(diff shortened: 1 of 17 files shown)


backend/headless/backend.c
```diff
@@ -0,0 +1,166 @@
...
+#include <wlr/util/log.h>
+#include "backend/headless.h"
+#include "glapi.h"
+
+static bool backend_start(struct wlr_backend *wlr_backend) {
+	struct wlr_headless_backend *backend =
+		(struct wlr_headless_backend *)wlr_backend;
+	wlr_log(L_INFO, "Initializating headless backend");
+
+	// TODO
```
SirCmpwn: ?

emersion: This `i < 1` and `1280, 720` need to be replaced. Maybe I'll just refuse to create outputs before backend startup, and do not create any output by default.

SirCmpwn: Should probably cache output create requests like Wayland does I suppose.

emersion: An issue with this is that NULL will be returned by `wlr_headless_add_output` instead of a `wlr_output`.

SirCmpwn: Nah. It's not like backend_start is really doing anything here. You already have an EGL context ready once init is called.

emersion: Okay, so let's make `wlr_headless_add_output` create outputs regardless of whether or not the backend is started, and start output timers on backend startup.

Should I use one timer for all outputs, or one timer per output like it's done currently?

SirCmpwn: One timer per output and make the framerate configurable. Let's also make the resolution reconfigurable at runtime.

emersion:
This PR is ready for review. Since backends are not configurable in rootston, there are some changes in `main.c` that'll need to be reverted before merge.

SirCmpwn:
Tested, works great. Great work!

SirCmpwn:
Code looks good too. :+1:

backend/headless/output.c
```diff
@@ -0,0 +1,146 @@
...
+		(struct wlr_headless_output *)wlr_output;
+	if (!eglMakeCurrent(output->backend->egl.display,
+		output->egl_surface, output->egl_surface,
+		output->backend->egl.context)) {
+		wlr_log(L_ERROR, "eglMakeCurrent failed: %s", egl_error());
+	}
+}
+
+static void output_swap_buffers(struct wlr_output *wlr_output) {
+	// No-op
```
ascent12: Are you sure you don't need to call eglSwapBuffers here?

emersion: From https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglSwapBuffers.xhtml

>If surface is a pixel buffer or a pixmap, eglSwapBuffers has no effect, and no error is generated.

ascent12:
There is an awful lot of duplication of the EGL code.
I feel the code in render/egl.c could be made more generally useful instead of just copying it.

SirCmpwn:
I thought the same thing, but the changes touch a few places and the code isn't that long so I didn't object. I wouldn't mind seeing it refactored, though.

ascent12:
Would calling `wlr_egl_init(&backend->egl, EGL_PLATFORM_SURFACELESS_MESA)` not work for this?
It would require some changes to how EGLConfigs are handled, though. We'd need some way to get custom attributes to it, and possibly separate it out from `wlr_egl_init` entirely, and make it its own API function.

SirCmpwn:
That should basically be it, yes.

emersion:
I didn't want to split `wlr_egl_init` into two functions and allow users to half-initialize an `wlr_egl`, so I added a `config_attribs` parameter.

`eglGetPlatformDisplayEXT` isn't working with `EGL_PLATFORM_SURFACELESS_MESA` (even if the docs say it should, tried `eglGetPlatformDisplayEXT(EGL_PLATFORM_SURFACELESS_MESA, EGL_DEFAULT_DISPLAY, NULL)`), so I added a special case for this one.

SirCmpwn:
Thanks!

rootston/main.c
```diff
@@ -29,7 +30,11 @@ int main(int argc, char **argv) {
 	assert(server.wl_display = wl_display_create());
 	assert(server.wl_event_loop = wl_display_get_event_loop(server.wl_display));
 
-	assert(server.backend = wlr_backend_autocreate(server.wl_display));
+	//assert(server.backend = wlr_backend_autocreate(server.wl_display));
+	assert(server.backend = wlr_headless_backend_create(server.wl_display));
+	wlr_headless_add_output(server.backend, 1280, 720);
+	wlr_headless_add_input_device(server.backend, WLR_INPUT_DEVICE_KEYBOARD);
+	wlr_headless_add_input_device(server.backend, WLR_INPUT_DEVICE_POINTER);
```
martinetd: Shouldn't this part be commented out/reverted for now?
Not sure if it does any harm but it might confuse output lists/things like that.

PR #498, opened 2017-12-18
rootston: revert to auto backend

emersion:
It's my fault not making it more obvious that this needed to be reverted before merging #496.

>Since backends are not configurable in rootston, there are some changes in main.c that'll need to be reverted before merge.


emersion:
I should really stop cleaning up branches in my fork.

SirCmpwn:
derp

PR #501, opened 2017-12-18
misc keyboard fixes

acrisci:


SirCmpwn:
What remains to be done here? Which problems does this fix?

acrisci:
It fixes all of the xkb memory leaks.

There is also a global destroy event on wlr-output that @emersion forgot to remove that's fixed here that fixes a segfault on exit.

SirCmpwn:
derp, missed @emersion's comment

acrisci:
Oh how did that not cause problems for me...

emersion:
Probably because the free'd block still contained the old data by chance.

PR #504, opened 2017-12-19
Destroy multi backend on display destroy

emersion:
That allows to remove the `wlr_backend_destroy` call in `main.c`, because everything is cleaned up by the display destroy.


SirCmpwn:
Thanks!

Issue #497, opened 2017-12-18
libinput configuration


martinetd:
One of the things I'm missing right now when using rootson (/ need to test recent wlroots branch of sway), is libinput configuration.

I can use a script that runs some xinput commands at startup, but that wouldn't work without xwayland and I think it's rather kludgy compared to how far our XKB support has gotten.

My problem is that the libinput API is quite complex... https://wayland.freedesktop.org/libinput/doc/latest/group__config.html
There are plenty of function calls I do not really care about/need (calibration matrixes), some that I don't care about but should probably be done anyway (left handed etc), and in general the API is so large that I feel it's going to be pretty close to config parsing ultimately: I can't see how to abstract this easily.

I see rootston hooks on input_add.notify/input_remove.notify so should we just try to check if input is of proper type (e.g. pointer + libinput backend?) and leave it to rootston to try to set things then?
I'm not even sure how to check if the input is a libinput input or not at this point, but if that's fine I suggest we do a couple of arbitrary input settings in rootston just to illustrate and then do as many as we're comfortable having in sway?
Is there a better way I didn't think of?

(I'll have time to work on this starting next week, but might as well get some design plan first)

emersion:
>I see rootston hooks on input_add.notify/input_remove.notify so should we just try to check if input is of proper type (e.g. pointer + libinput backend?) and leave it to rootston to try to set things then?

That's the plan. This will be done in sway after https://github.com/swaywm/sway/pull/1505 is merged. See https://github.com/swaywm/wlroots/issues/482

martinetd:
Ok, had indeed missed #482.
I'm not sure why we need to wait for swaywm/sway#1505 though (except to not disperse too many efforts all around) so unless you've started I'll take a look at both soonish™

acrisci:
Let's keep the discussion about this at the original issue. I'll write up our current thinking of how this should work over there.

@martinetd read the requirements at the [sway-input man page](https://github.com/swaywm/sway/blob/master/sway/sway-input.5.txt) for the input pr and write an example config file for rootston. Any further suggestions for how this works is appreciated.

PR #505, opened 2017-12-19
Add wlr_input_device_is_foo and wlr_output_is_bar functions

### LINKED ISSUE #482: Add wlr_input_device_is_libinput
And corresponding functions for wlr_output, etc

martinetd:
This adds all the variants of wlr_input_device and wlr_output that I could find **that have and impl**.

This last point is important, because X11 has wlr_input_devices with a NULL implem, so we can't use that to check as we have done with backends.
I've considered creating a dummy impl initialized to 0 to use that just for check, should I?

Closes #482.
Reopens #497. (I'm not sure that's legit, but these really are two separate issues. I doubt I'll forget but feels cleaner with the issue open)

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -658,6 +658,10 @@ static struct wlr_output_impl output_impl = {
 	.get_gamma_size = wlr_drm_connector_get_gamma_size,
 };
 
+bool wlr_output_is_drm(struct wlr_output *output) {
+	return output->impl == &output_impl;
+}
+
 static int retry_pageflip(void *data) {
 	struct wlr_drm_connector *conn = data;
 	wlr_log(L_INFO, "%s: Retrying pageflip", conn->output.name);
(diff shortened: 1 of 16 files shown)


martinetd:
> Reopens #497. (I'm not sure that's legit, but these really are two separate issues. I doubt I'll forget but feels cleaner with the issue open)

Answering myself, it's not (keyword isn't underlined like Closes). I'll leave a comment to reopen it there when this gets merged if nobody did.

>  I've considered creating a dummy impl initialized to 0 to use that just for check, should I?

I've done that, and tested (started rootston on x11)

emersion:
>Reopens #497.

How so?

martinetd:
>> Reopens #497.
> How so?

Because it was closed to "keep the discussion about this at the original issue", but I think it was what this PR is about (and thus closing) and not what libinput config is about (which is still missing, and thus needs tracking).
I don't really care either way, will continue working on it as intented whatever happens to these, it's just messy :)

emersion:
Oh, okay. libinput config is more of a rootston concern now. Also it's not really that this PR reopens the libinput config issue, it's more that the libinput config issue was never fixed and depended on this PR.

Just submit a PR with rootston libinput config ;)

Issue #503, opened 2017-12-19
wlr-backend: return multi backend from autocreate by default


acrisci:
This way we can use it with the sway multibackend features by using the autocreate function, and then always being able to add more backends later if needed instead of writing all of the autocreate functionality into sway.

PR #506, opened 2017-12-19
[wip] multibackend fixes

acrisci:
Preparation for the multibackend features of sway outlined in [sway#1529](https://github.com/swaywm/sway/issues/1529).

Changes:

* [x] backend destroy event
* [x] remove assumption that the multibackend must have a session (drm backend can be destroyed/recreated and the multibackend lives on or multibackend on X11+Wayland only)
* [x] #503 - return multibackend from autocreate in a useful way
* [x] `wlr_multi_backend_remove()`
* [x] backend_add and backend_remove events

Open issues:

* Does the multibackend own the wlr-backend children? That is, do we really want to destroy the wlr-backends when the parent multi is destroyed?
* Current implementation will return an empty (non-NULL but no subbackends) wlr-multi-backend instead of NULL when no backends could be created
* When you add a backend to the multi, will it reemit all of the outputs and inputs on added multibackend for the newly added backend? (current solution: no, use the backend_add and backend_remove events to do this)

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -18,6 +18,7 @@ void wlr_backend_init(struct wlr_backend *backend,
 		const struct wlr_backend_impl *impl) {
 	assert(backend);
 	backend->impl = impl;
+	wl_signal_init(&backend->events.destroy);
 	wl_signal_init(&backend->events.input_add);
 	wl_signal_init(&backend->events.input_remove);
 	wl_signal_init(&backend->events.output_add);
@@ -32,6 +33,7 @@ bool wlr_backend_start(struct wlr_backend *backend) {
 }
 
 void wlr_backend_destroy(struct wlr_backend *backend) {
+	wl_signal_emit(&backend->events.destroy, backend);
 	if (backend->impl && backend->impl->destroy) {
 		backend->impl->destroy(backend);
 	} else {
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -72,30 +72,31 @@ static struct wlr_backend *attempt_wl_backend(struct wl_display *display) {
 }
 
 struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
-	struct wlr_backend *backend;
+	struct wlr_backend *backend = wlr_multi_backend_create(display);
+	if (!backend) {
+		wlr_log(L_ERROR, "could not allocate multibackend");
+		return NULL;
+	}
+
 	if (getenv("WAYLAND_DISPLAY") || getenv("_WAYLAND_DISPLAY")) {
-		backend = attempt_wl_backend(display);
-		if (backend) {
-			return backend;
+		struct wlr_backend *wl_backend = attempt_wl_backend(display);
+		if (wl_backend) {
+			wlr_multi_backend_add(backend, wl_backend);
 		}
 	}
 
 	const char *x11_display = getenv("DISPLAY");
 	if (x11_display) {
-		return wlr_x11_backend_create(display, x11_display);
+		struct wlr_backend *x11_backend =
+			wlr_x11_backend_create(display, x11_display);
+		wlr_multi_backend_add(backend, x11_backend);
 	}
 
 	// Attempt DRM+libinput
-
 	struct wlr_session *session = wlr_session_create(display);
 	if (!session) {
 		wlr_log(L_ERROR, "Failed to start a DRM session");
-		return NULL;
-	}
-
-	backend = wlr_multi_backend_create(display);
-	if (!backend) {
-		goto error_session;
+		return backend;
 	}
 
 	struct wlr_backend *libinput = wlr_libinput_backend_create(display, session);
@@ -134,9 +135,8 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
 
 error_multi:
 	wlr_backend_destroy(backend);
-error_session:
 	wlr_session_destroy(session);
-	return NULL;
+	return backend;
 }
 
 uint32_t usec_to_msec(uint64_t usec) {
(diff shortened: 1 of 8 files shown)


emersion:
>Does the multibackend own the wlr-backend children? That is, do we really want to destroy the wlr-backends when the parent multi is destroyed?

I think so. One could always remove a sub-backend before destroying the multi-backend if he wants to.

>Current implementation will return an empty (non-NULL but no subbackends) wlr-multi-backend instead of NULL when no backends could be created

I'm not sure if it's helpful to return an empty multi-backend. If we want to do that, maybe add a `wlr_multi_backend_empty` function that checks if a multi-backend doesn't have sub-backends?

>When you add a backend to the multi, will it reemit all of the outputs and inputs on added multibackend for the newly added backend? (current solution: no, use the backend_add and backend_remove events to do this)

But I don't think there's a way to list all outputs and input devices from a backend?

acrisci:
The multibackend is so different from the others, I'm wondering if we should consider if it would fit better as its own type. It's not like nested multibackends make sense.

SirCmpwn:
>The multibackend is so different from the others, I'm wondering if we should consider if it would fit better as its own type. It's not like nested multibackends make sense.

I think that conceptually it works as an implementation of a backend, but I'm okay with `assert`ing that they can't be nested.

SirCmpwn:
>Current implementation will return an empty (non-NULL but no subbackends) wlr-multi-backend instead of NULL when no backends could be created

Return NULL

emersion:
>When you add a backend to the multi, will it reemit all of the outputs and inputs on added multibackend for the newly added backend? (current solution: no, use the backend_add and backend_remove events to do this)

I think hotplugging shouldn't work differently as implemented right now. This means, when you want to hotplug a backend:
1. Create it
2. Add it to the multi-backend
3. Start it

This way, outputs and input devices will be advertized through signals just like right now, and we won't need to add functions to list outputs and input devices for a backend.

backend/multi/backend.c
```diff
@@ -28,13 +30,22 @@ static bool multi_backend_start(struct wlr_backend *wlr_backend) {
...
+	wl_list_remove(&sub->link);
+	free(sub);
+}
+
 static void multi_backend_destroy(struct wlr_backend *wlr_backend) {
 	struct wlr_multi_backend *backend = (struct wlr_multi_backend *)wlr_backend;
-	wl_list_remove(&backend->display_destroy.link);
 	struct subbackend_state *sub, *next;
 	wl_list_for_each_safe(sub, next, &backend->backends, link) {
+		// XXX do we really want to take ownership over added backends?
```
SirCmpwn: Yes

emersion:
What are the remaining TODOs here?

acrisci:
> I think hotplugging shouldn't work differently as implemented right now. This means, when you want to hotplug a backend:

I guess the use case I was thinking of was taking an already running backend and adding it to an existing multibackend, but now I don't know if that's necessary.

> What are the remaining TODOs here?

Figure out if this will work for the sway feature. There's probably not much more to do.

backend/multi/backend.c
acrisci: Ok, well the implication is that backends cannot belong to multiple multibackends which needs to be made clear to people. Dunno why anyone would want more than one multibackend though.

SirCmpwn:
>Figure out if this will work for the sway feature. There's probably not much more to do.

Probably shouldn't block this patch on that.

backend/multi/backend.c
```diff
@@ -57,27 +68,19 @@ struct wlr_backend_impl backend_impl = {
...
 		return NULL;
 	}
 
-	backend->session = session;
 	wl_list_init(&backend->backends);
 	wlr_backend_init(&backend->backend, &backend_impl);
 
-	session->display_destroy.notify = handle_display_destroy;
-	wl_display_add_destroy_listener(display, &session->display_destroy);
+	wl_signal_init(&backend->events.backend_add);
```
emersion: Why don't you listen to display destroy anymore? We need to free all backends, including the multi-backend, when the display is destroyed.

acrisci: My intention is that the drm backend should be able to added and removed at runtime, and I thought session destroy would emit the destroy event on the drm backend and have it removed at that time.

emersion: But you still have to free the multi-backend

backend/backend.c
```diff
@@ -70,30 +72,31 @@ static struct wlr_backend *attempt_wl_backend(struct wl_display *display) {
...
+			wlr_multi_backend_add(backend, wl_backend);
 		}
 	}
 
 	const char *x11_display = getenv("DISPLAY");
 	if (x11_display) {
-		return wlr_x11_backend_create(display, x11_display);
+		struct wlr_backend *x11_backend =
+			wlr_x11_backend_create(display, x11_display);
+		wlr_multi_backend_add(backend, x11_backend);
```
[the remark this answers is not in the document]
acrisci: Yeah, the way this works is if you do `DISPLAY=:1 WAYLAND_DISPLAY=wayland-1 rootston` on a tty, it will open all three backends at the same time. Do you not like that?

acrisci:
Ok I think this is good.

backend/backend.c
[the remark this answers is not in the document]
emersion: I don't want the X11 backend if I'm running Wayland (in GNOME, for instance).

rootston/main.c
```diff
@@ -32,6 +33,12 @@ int main(int argc, char **argv) {
 
 	assert(server.backend = wlr_backend_autocreate(server.wl_display));
 
+	if (wlr_multi_is_empty(server.backend)) {
```
emersion: This check is no longer needed.

SirCmpwn:
In the future please don't give me a heart attack when I see [wip] patches getting merged in my inbox

PR #508, opened 2017-12-22
Fix rootston seat button count

### LINKED ISSUE #493: Moving windows with rootston keybindings breaks dnd
button_count becomes negative (or MAX, rather) which fucks shit up real good.

emersion:
When we're moving a window, this can be:
* Because the client has sent a move request, in which case the client got the button press, and we need to send the button released event to the seat
* Because of a keybinding, in which case we don't send the button pressed event to the seat, and we have not to send the button released event to the seat

Test plan: open `weston-dnd`, move the window with the titlebar and with rootston keybindings, try to drag some flowers.

Fixes #493

### DIFF
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -166,6 +166,13 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 		}
 		return;
 	}
+	if (state == WLR_BUTTON_RELEASED &&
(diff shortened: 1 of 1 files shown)


SirCmpwn:
@mikkeloscar take a look at this Travis failure

mikkeloscar:
@SirCmpwn upstream issue: https://github.com/mikkeloscar/arch-travis/issues/33

emersion:
Restarted the build

PR #509, opened 2017-12-22
Add gtk-primary-selection protocol

emersion:
Test plan: open two different GNOME apps, select some text on one of them, focus the other, and press the middle mouse button. The text should be pasted.

### DIFF
--- a/types/wlr_primary_selection.c
+++ b/types/wlr_primary_selection.c
@@ -7,9 +7,6 @@
 #include <wlr/types/wlr_seat.h>
 #include <wlr/util/log.h>
 
-static void client_source_accept(struct wlr_primary_selection_source *source,
-		uint32_t serial, const char *mime_type) {}
-
 static void client_source_send(struct wlr_primary_selection_source *source,
 		const char *mime_type, int32_t fd) {
 	gtk_primary_selection_source_send_send(source->resource, mime_type, fd);
@@ -24,7 +21,14 @@ static void client_source_cancel(
 
 static void offer_handle_receive(struct wl_client *client,
 		struct wl_resource *resource, const char *mime_type, int32_t fd) {
-	// TODO
+	struct wlr_primary_selection_offer *offer =
+		wl_resource_get_user_data(resource);
+
+	if (offer->source && offer == offer->source->offer) {
+		offer->source->send(offer->source, mime_type, fd);
+	} else {
+		close(fd);
+	}
 }
 
 static void offer_handle_destroy(struct wl_client *client,
@@ -52,7 +56,7 @@ static void offer_resource_handle_destroy(struct wl_resource *resource) {
 	}
 
 	if (offer->source->resource) {
-		wl_data_source_send_cancelled(offer->source->resource);
+		gtk_primary_selection_source_send_cancelled(offer->source->resource);
 	}
 
 	offer->source->offer = NULL;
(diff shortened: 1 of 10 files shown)


SirCmpwn:
This should be synchronized with the xwayland primary selection when it's ready. Also could use a test plan.

emersion:
I don't feel like making xwayland support in this PR, I'll do that in another one. I'll add a test plan when ready.

SirCmpwn:
:+1:

emersion:
This PR is ready for review.

types/wlr_primary_selection.c
```diff
@@ -1,10 +1,118 @@
...
 #include <stdlib.h>
+#include <unistd.h>
 #include <string.h>
 #include <gtk-primary-selection-protocol.h>
 #include <wlr/types/wlr_primary_selection.h>
+#include <wlr/types/wlr_seat.h>
 #include <wlr/util/log.h>
 
+static void client_source_accept(struct wlr_primary_selection_source *source,
+		uint32_t serial, const char *mime_type) {}
```
SirCmpwn: Code style here is pretty bad

SirCmpwn: Prefer:

```
... {
    // This space intentionally left blank
}
```

SirCmpwn:
Was looking at an old diff. This LGTM.

types/wlr_primary_selection.c
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	manager->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &manager->display_destroy);
+
+	return manager;
+}
+
+void wlr_primary_selection_device_manager_destroy(
+		struct wlr_primary_selection_device_manager *manager) {
```
acrisci: need to clean up `display_destroy` listener.

types/wlr_primary_selection.c
```diff
@@ -0,0 +1,366 @@
...
+		uint32_t serial) {
+	struct wlr_primary_selection_source *source = NULL;
+	if (source_resource != NULL) {
+		source = wl_resource_get_user_data(source_resource);
+	}
+
+	struct wlr_seat_client *seat_client =
+		wl_resource_get_user_data(resource);
+
+	// TODO: store serial and check against incoming serial here
```
acrisci: How could the serial be validated? Serial validation is currently thought to be a responsibility of rootston. Should we add the ability for the seat itself to validate serials and use this approach in rootston as well?

types/wlr_primary_selection.c
```diff
@@ -0,0 +1,366 @@
...
+
+static const struct gtk_primary_selection_source_interface source_impl = {
+	.offer = source_handle_offer,
+	.destroy = source_handle_destroy,
+};
+
+static void source_resource_handle_destroy( struct wl_resource *resource) {
+	struct wlr_primary_selection_source *source =
+		wl_resource_get_user_data(resource);
+	wl_signal_emit(&source->events.destroy, source);
```
acrisci: Need to free the mime types here?

types/wlr_primary_selection.c
acrisci: I think #511 is the right way to do this.

emersion:
Fixed @acrisci's comments

acrisci:
👍

PR #512, opened 2017-12-23
[wip] Headless fixes

acrisci:
Some fixes for the headless backend needed for sway multibackend.

### DIFF
--- a/backend/headless/output.c
+++ b/backend/headless/output.c
@@ -94,7 +94,7 @@ static int signal_frame(void *data) {
 }
 
 struct wlr_output *wlr_headless_add_output(struct wlr_backend *wlr_backend,
-		unsigned int width, unsigned int height) {
+		unsigned int width, unsigned int height, char *name) {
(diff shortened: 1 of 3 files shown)


backend/headless/output.c
```diff
@@ -93,8 +93,10 @@ static int signal_frame(void *data) {
 	return 0;
 }
 
+static int current_output_id = 0;
```
emersion: You can use `wl_list_length` instead of a global.

acrisci: I'm trying to avoid duplicate names because it might be a hard case to handle for some uses. For instance, if I add two backends, delete the first one, then add another one, I have two HEADLESS-1 outputs.

SirCmpwn:
What is the status of this?

acrisci:
This work can't really be started until we support multiple renderers. There's not much here so I'll close it.

SirCmpwn:
I don't think headless and multibackend are necessarily related.

PR #516, opened 2017-12-24
enabled libcap by default on arch buildpkg (arch default)

cawilliamson:
It looks like libcap needs to be enabled by default on ArchLinux since Arch pushes systemd and logind which require libcap to be enabled.

I may be missing something here though so happy to be corrected.


SirCmpwn:
I'm getting rid of the PKGBUILD altogether instead of either of your patches.

PR #518, opened 2017-12-24
rootston: add arbitrary libinput config 'tap_enabled'

martinetd:
This ought to be enough to serve as an example for other compsitors to use.

I've already done the sway config as well, so not sure this is actually strictly neeeded, but I guess it can't hurt?

One last reference to #497 to properly close it in my head :)

### DIFF
--- a/include/rootston/config.h
+++ b/include/rootston/config.h
@@ -21,6 +21,7 @@ struct roots_device_config {
 	char *name;
 	char *seat;
 	char *mapped_output;
+	bool tap_enabled;
 	struct wlr_box *mapped_box;
 	struct wl_list link;
 };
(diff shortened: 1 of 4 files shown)


emersion:
I'd like to rename `wlr_libinput_get_device_handle(device)` to `wlr_input_device_get_libinput(device)` or something like that.

martinetd:
If we rename this, let's do it before the sway side gets merged. I have no preference here. Others?

SirCmpwn:
>I'd like to rename wlr_libinput_get_device_handle(device) to wlr_input_device_get_libinput(device) or something like that.

-1. We prefix function names with their subsystem, not half-prefix half-postfix.

SirCmpwn:
>I'd like to rename wlr_libinput_get_device_handle(device) to wlr_input_device_get_libinput(device) or something like that.

This is actually backwards, what you're proposing is what's already the case and the opposite of your proposal is a good idea.

SirCmpwn:
Sorry, I'm still misunderstanding you.

`wlr_input_device_is_libinput` is correct. `wlr_libinput_get_device_handle` is correct.

I see where you're coming from but let's not change it.

PR #520, opened 2017-12-25
adds a configure_notify event to xwayland

Ongy:
To know when a client reacts to a resize or other event we can listen to
the configure event on xdg_shell.
For xwayland the required event isn't exported so far.
This adds a signal to wlr_xwayland_surface that's fired whenever it
changes position or size.

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -102,6 +102,7 @@ static struct wlr_xwayland_surface *wlr_xwayland_surface_create(
 	wl_signal_init(&surface->events.request_fullscreen);
 	wl_signal_init(&surface->events.map_notify);
 	wl_signal_init(&surface->events.unmap_notify);
+	wl_signal_init(&surface->events.configure_notify);
 	wl_signal_init(&surface->events.set_class);
 	wl_signal_init(&surface->events.set_title);
 	wl_signal_init(&surface->events.set_parent);
@@ -632,6 +633,22 @@ static void xwm_handle_configure_notify(struct wlr_xwm *xwm,
 	xsurface->y = ev->y;
 	xsurface->width = ev->width;
 	xsurface->height = ev->height;
+
+	struct wlr_xwayland_surface_configure_event *wlr_event =
+		calloc(1, sizeof(struct wlr_xwayland_surface_configure_event));
+	if (wlr_event == NULL) {
+		return;
+	}
+
+	wlr_event->surface = xsurface;
+	wlr_event->x = ev->x;
+	wlr_event->y = ev->y;
+	wlr_event->width = ev->width;
+	wlr_event->height = ev->height;
+
+	wl_signal_emit(&xsurface->events.request_configure, wlr_event);
+
+	free(wlr_event);
 }
 
 #define ICCCM_WITHDRAWN_STATE	0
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Paging @martinetd for a look, too.

xwayland/xwm.c
```diff
@@ -632,6 +633,22 @@ static void xwm_handle_configure_notify(struct wlr_xwm *xwm,
...
+		return;
+	}
+
+	wlr_event->surface = xsurface;
+	wlr_event->x = ev->x;
+	wlr_event->y = ev->y;
+	wlr_event->width = ev->width;
+	wlr_event->height = ev->height;
+
+	wl_signal_emit(&xsurface->events.request_configure, wlr_event);
```
martinetd: You added a new configure_notify but you use request_configure?

xwayland/xwm.c
```diff
@@ -632,6 +633,22 @@ static void xwm_handle_configure_notify(struct wlr_xwm *xwm,
 	xsurface->y = ev->y;
```
martinetd: More a nitpick than anything, this function doesn't have a log message unlike handle_configure_request. I think it could make sense to add one if we find xsurface that logs new x/y/width/height at debug level.
Not overly fussy about this

xwayland/xwm.c
Ongy: thx. Right, I had simply copy pasted the code and seems I missed that rather important thing.

Ongy:
Yes, I was refering to the `events.commit` on `xdg_shell`, which can be used to react to size changes (unless I misunderstood the protocol).
From how I understand X this isn't exactly the same, since the commit is triggered every time the surface provides a new image, while the `configure_notify` should be strictly when the size changes.

That's why I went for a different name.

martinetd:
>  From how I understand X this isn't exactly the same, since the commit is triggered every time the surface provides a new image, while the configure_notify should be strictly when the size changes.
> That's why I went for a different name.

This can't be everytime there's a new image, e.g. I get no event when I just type in my xwayland terminal in rootston.
I only seem to get events on things like resize like configure commit in xdg/wl_shells ; I'd think we should use that here as well unless this is really different. Maybe @emersion or @acrisci know if there's such a fondamental difference.

emersion:
In all shells, you should listen to the surface commit to know when the client has updated the surface size.

In xdg-shell, the client sends a configure_ack and then commits when it wants to resize after a configure request. When the configure_ack is received, you know that the next commit will be in response to the configure request. But when the configure_ack is received, the client still hasn't resized its surface: that's why there's no configure_ack event.

In xdg-shell, some operations on the xdg-surface are performed on wl-surface commit and then the xdg-surface commit event is emitted. If you want to make sure the xdg-surface state is consistent, listen to xdg-surface's commit. Note: we've been discussing with @acrisci about removing the xdg-surface commit while still ensuring the xdg-surface state is updated when the wl-surface commit is emitted.

In wl-shell, there's no configure_ack. You cannot do better than assuming the commit right after a configure request has been triggered by the configure request.

In xwayland, there's a configure_notify, but you don't know for which configure request the configure_notify is for. The configure_notify still runs before the surface commit, so updating the xwayland surface size there makes sense. But just like in xdg-shell, you shouldn't listen to this event since the wl-surface still has the old size. You should listen to the wl-surface commit event instead.

You can just look how rootston does it.

acrisci:
This is correct.

The surface commit event is the authority on the state of the surface. All the things that happen before that is protocol negotiation which is our job to abstract away from you.

PR #521, opened 2017-12-26
add nop command

acrisci:
The `nop` command can be used by developers to test things at runtime. Arbitrary code can be put in the nop command handler and shared with diffs to demonstrate issues in bug reports.

### DIFF
--- a/rootston/keyboard.c
+++ b/rootston/keyboard.c
@@ -117,6 +117,8 @@ static void keyboard_binding_execute(struct roots_keyboard *keyboard,
 		if (focus != NULL) {
 			view_maximize(focus, !focus->maximized);
 		}
+	} else if (strcmp(command, "nop") == 0) {
+		wlr_log(L_DEBUG, "nop command");
 	} else {
 		wlr_log(L_ERROR, "unknown binding command: %s", command);
 	}


emersion:
It wouldn't be difficult to add a `strcasecmp` instead of modifying the `nop` handler, so I'm not sure there's a real value for this use case. Though, in sway were you can reconfigure stuff at runtime, I do see value in adding a `nop` command to "unbind" a keybinding.

acrisci:
If reconfiguring things at runtime is a library feature of a wlroots component, then it should be demonstrated in rootston for reference.

I'm having problems with multibackend stuff and I need to demonstrate the issue in a issue somehow. I *should* implement the full dynamic backend feature into rootston, but that would be a lot of work because we would need an ipc for that. But maybe that's actually the right way to go.

emersion:
I think rootston not having dynamic configuration is a separate issue. Right now rootston doesn't have dynamic configuration so we should decide whether to accept this patch or not with static configuration in mind.

I don't agree with your first point about putting arbitrary stuff in the `nop` command (since it would be easier to just add a new command instead), and using `nop` to unbind keybindings is only useful in a dynamic configuration scenario.

Though, I just thought of another use-case: `nop` could be used as a placeholder in the configuration file (just like `rotation=normal` that doesn't change the rotation). Also, when submitting patches for keyboard-related stuff, `nop` could be used as a harmless command to check that keybindings work.

acrisci:
Yeah just a harmless check that might be useful for development purposes (I'll use it).

SirCmpwn:
>If reconfiguring things at runtime is a library feature of a wlroots component, then it should be demonstrated in rootston for reference.
>
>I'm having problems with multibackend stuff and I need to demonstrate the issue somehow. I should implement the full dynamic backend feature into rootston, but that would be a lot of work because we would need an ipc for that. But maybe that's actually the right way to go.

I don't think this is necessarily worth the effort. We need to balance not wasting time on a useless compositor with adequately demonstrating the features of wlroots, and I think this would kind of be pushing it.

PR #522, opened 2017-12-26
Fix cursor on transformed output with DRM backend

### LINKED ISSUE #517: Wrong cursor position on rotated outputs with DRM backend
1. `rotate = 90`
2. Launch rooston with DRM backend

emersion:
This PR aims to fix issues with cursor on transformed outputs with the DRM backend. For some output transforms, the hardware cursor was not transformed or not at the correct position. I think these issues have been introduced when fixing flipped-90 and flipped-270 transforms in the surface transforms PR.

This PR aims to fix output transforms and check for any corner case issue.

Test plan:
* For the Wayland and DRM backends
* Configure the output transform to normal, 90, 180, 270, flipped, flipped-90, flipped-180, flipped-270 (90 and flipped-90 are the more important ones, because they're tricky)
* Spawn `weston-terminal` (has a surface transform) and `gnome-calculator` (doesn't have a surface transform), check that both have the same orientation
* Move the mouse over these two windows, check that the hardware cursor position agrees with the software cursor position (by checking that the cursor doesn't jump when entering the window)
* Try to move a window and check that the window moves in the same way as the cursor

Fixes #517

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -118,6 +118,9 @@ bool wlr_output_cursor_move(struct wlr_output_cursor *cursor,
 	double x, double y);
 void wlr_output_cursor_destroy(struct wlr_output_cursor *cursor);
 
-enum wl_output_transform wlr_output_transform_invert(enum wl_output_transform);
+enum wl_output_transform wlr_output_transform_invert(
+	enum wl_output_transform tr);
+enum wl_output_transform wlr_output_transform_compose(
+	enum wl_output_transform tr_a, enum wl_output_transform tr_b);
 
 #endif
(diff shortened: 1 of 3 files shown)


include/wlr/types/wlr_output.h
```diff
@@ -118,6 +118,9 @@ bool wlr_output_cursor_move(struct wlr_output_cursor *cursor,
 	double x, double y);
 void wlr_output_cursor_destroy(struct wlr_output_cursor *cursor);
 
-enum wl_output_transform wlr_output_transform_invert(enum wl_output_transform);
+enum wl_output_transform wlr_output_transform_invert(
```
acrisci: Can you add some documentation for the new functions? It's not obvious to me what they do from reading the names.

acrisci:
👍

PR #523, opened 2017-12-26
config: Put all defines into config.h

besser82:
This makes sense, given that there are installed header-files, which rely on some of those defines.


emersion:
Thanks!

PR #524, opened 2017-12-27
[wip] Role committed

acrisci:
Add a private role-committed func to wlr-surface to ensure the role commit handler always runs before all the other handlers. This simplifies the api by removing the need for a commit event on the role types.

Role call:

* [x] xdg-shell
* [x] wl-shell
* [x] wl_pointer-cursor (no committed func)
* [x] wl_data_device-icon
* [x] wl_subsurface (no committed func)
* [x] xwayland

### DIFF
--- a/include/wlr/types/wlr_data_device.h
+++ b/include/wlr/types/wlr_data_device.h
@@ -72,7 +72,6 @@ struct wlr_drag_icon {
 	} events;
 
 	struct wl_listener surface_destroy;
-	struct wl_listener surface_commit;
 	struct wl_listener seat_client_destroy;
 };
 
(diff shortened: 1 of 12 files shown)


emersion:
👍

emersion:
Ow, was that still WIP? Missed the last title change.

acrisci:
No I think it was good, just waiting for the build to pass. I think it's flaky.

SirCmpwn:
Hmm, I'm not sure I like this. I know you're pushing to unify the shell interface more, @acrisci, but this is an example where it might not make sense (i.e. abstracting features that exist in a subset of, but not all, shells). Implementation-wise the `role_data` abstraction is also iffy, it's not very type-safe and feels like a bit of a hack imo.

SirCmpwn:
This should also almost certainly use a wl_listener rather than a plain function pointer. I would probably be happier about `role_data` if it wasn't passed along in parameters but rather was just expected to be taken out of the wlr_surface.

PR #525, opened 2017-12-27
Serial validation for grab requests

### LINKED ISSUE #511: wlr-seat: validate serials
Serial validation should be moved into wlroots for library user convenience, and to avoid exposing certain library functions for the sole purpose of validation by the compositor library user.

If custom serial validation is required, like a noop validator for some reason, the interfaces that require validation should expose hooks (I wouldn't call this a priority right now though).

If some event does require validation and validation fails, the compositor should never know about the event.

The seat should store a list of recent events to validate against. Events that can validate an event include a button event or touch down. This should probably be done as a ring buffer of events.

The interface should be as simple as `bool wlr_seat_validate_serial(seat, serial)`

acrisci:
Add serial validation for grab requests.

Grab requests handled:

* xdg-shell move,resize
* wl-shell move,resize

Xwayland shell events cannot be validated because there is no serial information.

We thought serial validation was necessary for setting selections, but the the protocol doesn't specify how to validate these serials and Weston does no serial validation, so we won't do any validation of those after all.

fixes #511

### DIFF
--- a/include/rootston/cursor.h
+++ b/include/rootston/cursor.h
@@ -10,12 +10,6 @@ enum roots_cursor_mode {
 	ROOTS_CURSOR_ROTATE = 3,
 };
 
-struct roots_input_event {
-	uint32_t serial;
-	struct wlr_cursor *cursor;
-	struct wlr_input_device *device;
-};
-
 struct roots_cursor {
 	struct roots_seat *seat;
 	struct wlr_cursor *cursor;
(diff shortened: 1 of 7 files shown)


acrisci:
This does slightly change the behavior of grabs in a way that is more restrictive, but more in line with the text of the protocol.

You cannot start an interactive drag when more than one button is pressed on the pointer or there is more than one touch point down on the touch device.

SirCmpwn:
Thanks!

PR #526, opened 2017-12-27
Further improvements to header files and shared lib version

besser82:


meson.build
```diff
@@ -150,8 +156,9 @@ subdir('examples')
...
 	libraries: lib_wlr,
-	version: '0.0.1',
-	filebase: 'wlroots',
-	name: 'wlroots',
+	subdirs: '@0@-@1@'.format(meson.project_name(), meson.project_version()),
+	version: meson.project_version(),
+	filebase: meson.project_name(),
+	name: meson.project_name(),
 	description: 'Wayland compositor library',
 )
```
SirCmpwn: -1 to this commit. I'd rather not try to support several wlroots versions installed side-by-side.

SirCmpwn:
Looks good aside from that issue.

meson.build
besser82: Well, it's less about supporting parallel installs of different versions.  The sole purpose of that part is just to have the installed include-tree in an isolated directory, so it won't (possibly) clash with the include-tree in the source-dir.

SirCmpwn: Still not convinced. I don't like it.

besser82: So how should that be changed?  Just keeping `meson.project_name()` for the subdir?

SirCmpwn: Aye.

besser82: Fixed.

SirCmpwn:
LGTM, thanks! Will merge on build success.

besser82:
Yay!  =)

besser82:
@SirCmpwn Ready to merge, now  =)

SirCmpwn:
Thanks!

besser82:
You're welcome!

PR #530, opened 2017-12-27
meson: Install built rootston binary

besser82:


martinetd:
This is just my opinion, but I don't think we want to install rootston, at least not part of the default install target (not sure how meson handle multiple install_foo)

SirCmpwn:
By design.

besser82:
Okie…

PR #531, opened 2017-12-27
wlr-keyboard: keep track of number of keycodes pressed

acrisci:
Makes this useful for sway bindcode.

### DIFF
--- a/types/wlr_seat.c
+++ b/types/wlr_seat.c
@@ -874,11 +874,9 @@ void wlr_seat_keyboard_enter(struct wlr_seat *seat,
 
 		struct wl_array keys;
 		wl_array_init(&keys);
-		for (size_t i = 0; i < WLR_KEYBOARD_KEYS_CAP; ++i) {
-			if (keyboard->keycodes[i] != 0) {
-				uint32_t *p = wl_array_add(&keys, sizeof(uint32_t));
-				*p = keyboard->keycodes[i];
-			}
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Thanks!

types/wlr_seat.c
```diff
@@ -874,11 +874,9 @@ void wlr_seat_keyboard_enter(struct wlr_seat *seat,
 
 		struct wl_array keys;
 		wl_array_init(&keys);
-		for (size_t i = 0; i < WLR_KEYBOARD_KEYS_CAP; ++i) {
-			if (keyboard->keycodes[i] != 0) {
-				uint32_t *p = wl_array_add(&keys, sizeof(uint32_t));
-				*p = keyboard->keycodes[i];
-			}
+		for (size_t i = 0; i < keyboard->num_keycodes; ++i) {
+			uint32_t *p = wl_array_add(&keys, sizeof(uint32_t));
```
martinetd: wl_array_add can return null

emersion: Can you grep for wl_array_add and make sure that all return values are checked?

emersion: Can you grep for wl_array_add and make sure that all return values are checked?

martinetd: see top comment, about half are. I'll submit a PR to fix all that don't I guess.

PR #533, opened 2017-12-28
ENOMEM checks: consistently check wl_array_add return

martinetd:
Not quite perfect, but arguably better than nothing.

### DIFF
--- a/types/wlr_data_device.c
+++ b/types/wlr_data_device.c
@@ -907,6 +907,9 @@ static void data_source_offer(struct wl_client *client,
 		*p = strdup(mime_type);
 	}
 	if (!p || !*p){
+		if (p) {
+			source->mime_types.size -= sizeof *p;
+		}
 		wl_resource_post_no_memory(resource);
 	}
 }
--- a/types/wlr_data_device.c
+++ b/types/wlr_data_device.c
@@ -901,14 +901,14 @@ static void data_source_offer(struct wl_client *client,
 		wl_resource_get_user_data(resource);
 	char **p;
 
-	p = wl_array_add(&source->mime_types, sizeof *p);
+	p = wl_array_add(&source->mime_types, sizeof(*p));
 
 	if (p) {
 		*p = strdup(mime_type);
 	}
 	if (!p || !*p){
 		if (p) {
-			source->mime_types.size -= sizeof *p;
+			source->mime_types.size -= sizeof(*p);
 		}
 		wl_resource_post_no_memory(resource);
 	}
--- a/types/wlr_xdg_shell_v6.c
+++ b/types/wlr_xdg_shell_v6.c
@@ -922,19 +922,35 @@ static void wlr_xdg_toplevel_v6_send_configure(
 	wl_array_init(&states);
 	if (surface->toplevel_state->pending.maximized) {
 		s = wl_array_add(&states, sizeof(uint32_t));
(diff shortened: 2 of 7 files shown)


martinetd:
Sorry, doesn't build, will force push something in a min.

xwayland/selection.c
```diff
@@ -55,6 +55,15 @@ static int xwm_read_data_source(int fd, uint32_t mask, void *data) {
 	int current = selection->source_data.size;
 	if (selection->source_data.size < incr_chunk_size) {
 		p = wl_array_add(&selection->source_data, incr_chunk_size);
+		if (!p){
```
martinetd: Now I understand this slightly better, I guess it's probably safer to do the same as read fail (clear x11 selection, remove source and close fd); will do that instead.

types/wlr_xdg_shell_v6.c
```diff
@@ -922,19 +922,35 @@ static void wlr_xdg_toplevel_v6_send_configure(
 	wl_array_init(&states);
 	if (surface->toplevel_state->pending.maximized) {
 		s = wl_array_add(&states, sizeof(uint32_t));
-		*s = ZXDG_TOPLEVEL_V6_STATE_MAXIMIZED;
+		if (!s) {
+			wlr_log(L_ERROR, "Could not allocate state for maximized xdg_toplevel");
+		} else {
+			*s = ZXDG_TOPLEVEL_V6_STATE_MAXIMIZED;
```
emersion: Is it possible to abort if no memory is available instead of sending an invalid state to the client?

martinetd: Sure, done.

types/wlr_data_device.c
```diff
@@ -907,6 +907,9 @@ static void data_source_offer(struct wl_client *client,
 		*p = strdup(mime_type);
 	}
 	if (!p || !*p){
+		if (p) {
+			source->mime_types.size -= sizeof *p;
```
SirCmpwn: `sizeof(*p)`

types/wlr_xdg_shell_v6.c
```diff
@@ -948,6 +964,7 @@ static void wlr_xdg_toplevel_v6_send_configure(
 	zxdg_toplevel_v6_send_configure(surface->toplevel_state->resource, width,
 		height, &states);
 
+out:
 	wl_array_release(&states);
```
SirCmpwn: We should probably accompany these cases by posting OOM to the client

types/wlr_data_device.c
martinetd: I've sticked to the style of the file, 5 lines above. I guess I'll change that one as well then.

types/wlr_xdg_shell_v6.c
martinetd: I've seen `wl_resource_post_no_memory` around, is that what you are referring to? What does that actually does?

SirCmpwn: Yes, that's what I'm talking about. It tells the client that we were unable to fulfill its request due for OOMy reasons.

martinetd: Ok, I believe something like that might work. Not sure how to test in practice as this can't actually happen on linux/glibc... Hope for the best?

SirCmpwn:
Thanks!

PR #536, opened 2017-12-28
Send mode info on DRM refresh only once

Timidger:
When DRM refreshed, `wlr_output_set_mode` is called. It would then call
the DRM `set_mode` callback which sends the updated matrix and mode info.

However once that call completed it would then immediately send the
information again. This is handled poorly by xwayland, causing it to
scale up the clients twice.

I'm only half sure my reasoning is correct, so if someone with more knowledge on DRM / intricacies of the output part of the Wayland protocol wants to chime in that would be great.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -163,16 +163,7 @@ bool wlr_output_set_mode(struct wlr_output *output,
 	if (!output->impl || !output->impl->set_mode) {
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


emersion:
I think your fix is correct. Backends' `set_mode` will call `wlr_output_update_mode` that will broadcast the new mode.

Can you do the same for `wlr_output_set_custom_mode`?

SirCmpwn:
Thanks!

PR #541, opened 2017-12-29
Fix xwayland restart... and add wlr_seat destroy listener.

martinetd:
This started as a "simple" kludge to restore rootston hooks when xwayland restarts, ~~I'm still not quite happy with that part so comments more than welcome: basically, I'm just copying over listeners as a list + seat from old struct to new struct and it's not very pretty and quite fragile.
I think there needs to be some work to have a wlr_xwayland_reinit() function instead of using init() that doesn't memset(0) the whole struct as blindly as I used to do. Happy to do that either in this PR or in another one.~~ Done here.

The rest is quite good though:
 - some small xwm leak (there's tons left when I switch VT, but it's hard to tell where it comes from)
 - on restart, the shutdown happens in a different order to shutdown was completly broken, so added an event for when wlr_seat destroys itself that others can listen on.
 - Rootston apparently needed that event too because I kept crashing on shutdown after adding that event.

At this point, this works with no crash for me, so test plan:
 - start some Xwaylands stuff / wayland stuff
 - pkill Xwayland, Xwayland windows die (make sure there's been at least 5s since start or previous pkill)
 - start some new Xwayland window, they should appear (didn't previously, the last time this worked was on example compositor pre-rootston)
 - shutdown rootston, this shouldn't crash either.

### DIFF
--- a/rootston/seat.c
+++ b/rootston/seat.c
@@ -220,6 +220,27 @@ static void roots_seat_init_cursor(struct roots_seat *seat) {
 	seat->cursor->request_set_cursor.notify = handle_request_set_cursor;
 }
 
+static void seat_view_destroy(struct roots_seat_view *seat_view);
+
+void roots_seat_destroy(struct roots_seat *seat) {
+	struct roots_seat_view *view, *nview;
+
+	// TODO: probably more to be freed here
+
+	wl_list_for_each_safe(view, nview, &seat->views, link) {
+		seat_view_destroy(view);
+	}
+	// Can be called from seat_destroy handler, so don't destroy seat
+}
+
+static void roots_seat_handle_seat_destroy(struct wl_listener *listener,
+		void *data) {
+	struct roots_seat *seat =
+		wl_container_of(listener, seat, seat_destroy);
+
+	roots_seat_destroy(seat);
+}
+
 struct roots_seat *roots_seat_create(struct roots_input *input, char *name) {
 	struct roots_seat *seat = calloc(1, sizeof(struct roots_seat));
 	if (!seat) {
@@ -254,11 +275,10 @@ struct roots_seat *roots_seat_create(struct roots_input *input, char *name) {
 
 	wl_list_insert(&input->seats, &seat->link);
 
-	return seat;
-}
+	seat->seat_destroy.notify = roots_seat_handle_seat_destroy;
+	wl_signal_add(&seat->seat->events.destroy, &seat->seat_destroy);
 
-void roots_seat_destroy(struct roots_seat *seat) {
-	// TODO
+	return seat;
 }
 
 static void seat_add_keyboard(struct roots_seat *seat,
--- a/rootston/seat.c
+++ b/rootston/seat.c
@@ -222,23 +222,23 @@ static void roots_seat_init_cursor(struct roots_seat *seat) {
 
 static void seat_view_destroy(struct roots_seat_view *seat_view);
 
-void roots_seat_destroy(struct roots_seat *seat) {
-	struct roots_seat_view *view, *nview;
-
-	// TODO: probably more to be freed here
-
-	wl_list_for_each_safe(view, nview, &seat->views, link) {
-		seat_view_destroy(view);
-	}
-	// Can be called from seat_destroy handler, so don't destroy seat
-}
-
 static void roots_seat_handle_seat_destroy(struct wl_listener *listener,
 		void *data) {
 	struct roots_seat *seat =
 		wl_container_of(listener, seat, seat_destroy);
 
-	roots_seat_destroy(seat);
+	// TODO: probably more to be freed here
+	wl_list_remove(&seat->seat_destroy.link);
+
+	struct roots_seat_view *view, *nview;
+	wl_list_for_each_safe(view, nview, &seat->views, link) {
+		seat_view_destroy(view);
+	}
+}
+
+void roots_seat_destroy(struct roots_seat *seat) {
+	roots_seat_handle_seat_destroy(&seat->seat_destroy, seat->seat);
+	wlr_seat_destroy(seat->seat);
 }
 
 struct roots_seat *roots_seat_create(struct roots_input *input, char *name) {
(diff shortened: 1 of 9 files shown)


rootston/seat.c
```diff
@@ -220,6 +220,27 @@ static void roots_seat_init_cursor(struct roots_seat *seat) {
...
+
+void roots_seat_destroy(struct roots_seat *seat) {
+	struct roots_seat_view *view, *nview;
+
+	// TODO: probably more to be freed here
+
+	wl_list_for_each_safe(view, nview, &seat->views, link) {
+		seat_view_destroy(view);
+	}
+	// Can be called from seat_destroy handler, so don't destroy seat
```
emersion: We should destroy the seat if necessary. One way to do this is setting the seat to NULL in the destroy handler, and destroying it here only if it's non-NULL.

rootston/seat.c
```diff
@@ -220,6 +220,27 @@ static void roots_seat_init_cursor(struct roots_seat *seat) {
 	seat->cursor->request_set_cursor.notify = handle_request_set_cursor;
 }
 
+static void seat_view_destroy(struct roots_seat_view *seat_view);
+
+void roots_seat_destroy(struct roots_seat *seat) {
+	struct roots_seat_view *view, *nview;
+
+	// TODO: probably more to be freed here
```
emersion: Need to remove the seat destroy listener here.

types/wlr_seat.c
```diff
@@ -348,6 +348,8 @@ void wlr_seat_destroy(struct wlr_seat *wlr_seat) {
 		return;
 	}
 
+	wl_signal_emit(&wlr_seat->events.destroy, NULL);
```
emersion: We generally emit the signal with the object itself as data, here `wlr_seat`.

rootston/seat.c
martinetd: This what I first tried, but doesn't work because almost everything in here depends on seat->seat being set during tear-down as well....
I think in practice this is fine, seat listens to display destroy anyway so for rootston it doesn't matter too much.

xwayland/xwayland.c
```diff
@@ -169,6 +170,14 @@ static void handle_client_destroy(struct wl_listener *listener, void *data) {
...
 		wl_container_of(listener, wlr_xwayland, client_destroy);
 
+	struct wlr_seat *seat = wlr_xwayland->seat;
+	struct wl_list new_surface_signals;
+	struct wl_listener *new_surface_listener, *next;
+	wl_list_init(&new_surface_signals);
+	wl_list_for_each_safe(new_surface_listener, next, &wlr_xwayland->events.new_surface.listener_list, link) {
+		wl_list_remove(&new_surface_listener->link);
+		wl_list_insert(&new_surface_signals, &new_surface_listener->link);
+	}
```
emersion: You can do this more simply:

```c
struct wl_list new_surface_signals;
wl_list_init(&new_surface_signals);
wl_list_insert_list(&new_surface_signals, &wlr_xwayland->events.new_surface.listener_list);
wl_list_init(&wlr_xwayland->events.new_surface.listener_list);
```

Same for a few lines below.

I don't really like this, maybe it's possible to add a new function that doesn't reset these fields? (`wlr_xwayland_start`?)

xwayland/selection.c
```diff
@@ -826,6 +826,30 @@ void xwm_selection_init(struct wlr_xwm *xwm) {
...
+			wlr_seat_set_selection(xwm->seat, NULL,
+					wl_display_next_serial(xwm->xwayland->wl_display));
+		}
+		if (xwm->seat->primary_selection_source &&
+				xwm->seat->primary_selection_source->cancel == primary_selection_source_cancel) {
+			wlr_seat_set_primary_selection(xwm->seat, NULL,
+					wl_display_next_serial(xwm->xwayland->wl_display));
+		}
+		wl_list_remove(&xwm->seat_selection.link);
+		wl_list_remove(&xwm->seat_primary_selection.link);
```
emersion: `wlr_xwayland_set_seat(xwayland, NULL);`?

xwayland/xwayland.c
martinetd: I don't like this _at all_ :)
See PR summary comment, I was suggesting something like that yes.

xwayland/xwm.c
```diff
@@ -1270,10 +1268,12 @@ static void xwm_get_render_format(struct wlr_xwm *xwm) {
 
 	if (format == NULL) {
 		wlr_log(L_DEBUG, "No 32 bit render format");
+		free(reply);
```
emersion: Nitpick: add a NULL check for `reply` right after `xcb_render_query_pict_formats_reply`

rootston/seat.c
emersion: Then make this function static, so we don't call it from outside the destroy handler.

martinetd: I can put the whole implementation in the destroy handler, and move this function back for public use that just manually calls the handler + destroys the seat?
I'm not comfortable removing a function that's been declared forever, even if it's empty and never called...

xwayland/selection.c
martinetd: I'm not sure this is the same, but back when I wrote this I hadn't realized selection was the only use of the seat in here.
I think it's still separate conceptually, we could want to remove selections without detaching the seat? It's not set during xwm_selection_init at least.

martinetd: Actually we should even if it's not, current code can call these wl_list_remove twice and that won't work well. Will do.

xwayland/xwayland.c
martinetd: Done something better. I think I've addressed all your points with that, feel free to have another look.

xwayland/xwm.c
```diff
@@ -1256,6 +1254,10 @@ static void xwm_get_render_format(struct wlr_xwm *xwm) {
 		xcb_render_query_pict_formats(xwm->xcb_conn);
 	xcb_render_query_pict_formats_reply_t *reply =
 		xcb_render_query_pict_formats_reply(xwm->xcb_conn, cookie, NULL);
+	if (!reply) {
+		wlr_log(L_ERROR, "Did not get any reply from xcb_rrender_query_pict_formats");
```
emersion: Nitpick: typo in error message

emersion:
LGTM and works, thanks!

PR #542, opened 2017-12-29
Abstract data sources

### LINKED ISSUE #528: wlr_data_source: use an interface
Right now `wlr_data_source` has some callback fields (e.g. `send`) mixed with a bunch of other data fields. This makes it complicated to know which fields need to be populated, and this doesn't follow the interface pattern of other types. We could add a `wlr_data_source_interface` with these callbacks instead.

Also, the `resource` field is only used when the source comes from a client. We could remove that field from the struct, and add a `wlr_client_data_source` struct instead for instance.

emersion:
- [x] Abstract `wlr_data_source`
- [x] Abstract `wlr_primary_selection_source`

This refactors `wlr_data_source` and adds a new `struct client_data_source`. The `resource` field, only populated when the source comes from a Wayland client, is removed from the abstracted struct. Some new drag'n'drop-related callbacks are added, they'll be useful for the xwayland sync implementation.

Test plan:
* Check that clipboard still works: wayland <-> wayland, xwayland <-> wayland, xwayland <-> xwayland
* Same for primary selection
* Check that drag'n'drop still works (wayland <-> wayland only): `weston-flower`, dragging text

Fixes #528

### DIFF
--- a/include/wlr/types/wlr_data_device.h
+++ b/include/wlr/types/wlr_data_device.h
@@ -23,33 +23,39 @@ struct wlr_data_offer {
 	struct wl_resource *resource;
 	struct wlr_data_source *source;
 
-	uint32_t dnd_actions;
-	enum wl_data_device_manager_dnd_action preferred_dnd_action;
+	uint32_t actions;
+	enum wl_data_device_manager_dnd_action preferred_action;
 	bool in_ask;
 
 	struct wl_listener source_destroy;
 };
 
 struct wlr_data_source {
-	struct wl_resource *resource;
+	// source metadata
+	struct wl_array mime_types;
+	int32_t actions;
+
+	// source implementation
+	void (*send)(struct wlr_data_source *source, const char *mime_type,
+		int32_t fd);
+	void (*accept)(struct wlr_data_source *source, uint32_t serial,
+		const char *mime_type);
+	void (*cancel)(struct wlr_data_source *source);
+
+	// drag'n'drop implementation
+	void (*dnd_drop)(struct wlr_data_source *source);
+	void (*dnd_finish)(struct wlr_data_source *source);
+	void (*dnd_action)(struct wlr_data_source *source,
+		enum wl_data_device_manager_dnd_action action);
+
+	// source status
+	bool accepted;
 	struct wlr_data_offer *offer;
 	struct wlr_seat_client *seat_client;
 
-	struct wl_array mime_types;
-
-	bool accepted;
-
-	// drag and drop
+	// drag'n'drop status
 	enum wl_data_device_manager_dnd_action current_dnd_action;
-	uint32_t dnd_actions;
 	uint32_t compositor_action;
-	bool actions_set;
-
-	void (*accept)(struct wlr_data_source *source, uint32_t serial,
-			const char *mime_type);
-	void (*send)(struct wlr_data_source *source, const char *mime_type,
-			int32_t fd);
-	void (*cancel)(struct wlr_data_source *source);
 
 	struct {
 		struct wl_signal destroy;
--- a/types/wlr_data_device.c
+++ b/types/wlr_data_device.c
@@ -173,6 +173,8 @@ static void data_offer_resource_destroy(struct wl_resource *resource) {
 		goto out;
 	}
 
+	offer->source->offer = NULL;
+
 	// If the drag destination has version < 3, wl_data_offer.finish
 	// won't be called, so do this here as a safety net, because
 	// we still want the version >= 3 drag source to be happy.
@@ -183,7 +185,6 @@ static void data_offer_resource_destroy(struct wl_resource *resource) {
 		offer->source->cancel(offer->source);
 	}
 
-	offer->source->offer = NULL;
 out:
 	free(offer);
 }
--- a/types/wlr_data_device.c
+++ b/types/wlr_data_device.c
@@ -78,7 +78,6 @@ static void data_offer_update_action(struct wlr_data_offer *offer) {
 struct client_data_source {
 	struct wlr_data_source source;
 	struct wl_resource *resource;
-	bool actions_set;
 };
 
 static void client_data_source_accept(struct wlr_data_source *wlr_source,
@@ -380,7 +379,7 @@ void wlr_seat_set_selection(struct wlr_seat *seat,
 static void data_device_set_selection(struct wl_client *client,
 		struct wl_resource *dd_resource, struct wl_resource *source_resource,
 		uint32_t serial) {
-	struct wlr_data_source *source = NULL;
+	struct client_data_source *source = NULL;
 	if (source_resource != NULL) {
 		source = wl_resource_get_user_data(source_resource);
 	}
@@ -389,7 +388,8 @@ static void data_device_set_selection(struct wl_client *client,
 		wl_resource_get_user_data(dd_resource);
 
 	// TODO: store serial and check against incoming serial here
-	wlr_seat_set_selection(seat_client->seat, source, serial);
+	struct wlr_data_source *wlr_source = (struct wlr_data_source *)source;
+	wlr_seat_set_selection(seat_client->seat, wlr_source, serial);
 }
 
 static void data_device_release(struct wl_client *client,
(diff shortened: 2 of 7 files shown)


emersion:
This PR is ready for review.

include/wlr/types/wlr_data_device.h
```diff
@@ -23,33 +23,39 @@ struct wlr_data_offer {
...
+	// source metadata
 	struct wl_array mime_types;
+	int32_t actions;
+
+	// source implementation
+	void (*send)(struct wlr_data_source *source, const char *mime_type,
+		int32_t fd);
+	void (*accept)(struct wlr_data_source *source, uint32_t serial,
+		const char *mime_type);
+	void (*cancel)(struct wlr_data_source *source);
```
SirCmpwn: Isn't this better done with the standard impl pattern?

SirCmpwn:
LGTM overall

include/wlr/types/wlr_data_device.h
emersion: Yeah, I was hesitating switching to that. Putting mixed metadata and callbacks is still bothering me because the metadata fields need to be initialized and free'd so we'll end up with `wlr_data_source_impl_init` and `wlr_data_source_impl_finish`...

SirCmpwn: We already have things of that sort going on, nbd

martinetd:
LGTM as well, and tested a bit of everything except rich text copy (no clue how to use it but I don't really want to know either)

types/wlr_primary_selection.c
```diff
@@ -55,8 +44,8 @@ static void offer_resource_handle_destroy(struct wl_resource *resource) {
 		goto out;
 	}
 
-	if (offer->source->resource) {
-		gtk_primary_selection_source_send_cancelled(offer->source->resource);
+	if (offer->source->cancel) {
+		offer->source->cancel(offer->source);
 	}
 
 	offer->source->offer = NULL;
```
martinetd: Actually don't think this is a new problem, but when I go around killing windows at random (rootston close) with weird states I got a use-after-free here. offer->source->cancel() freed offer->source so we can't set offer->source->offer to NULL anymore at this point? Or if it shouldn't free it, need to remove the free in primary_selection_source_cancel in xwayland/selection.c

emersion: Thanks for spotting this! When destroying an offer, the source can still be used by another offer, so it doesn't make sense to cancel it. Removed that call (it was there because needed by previous version of the `wl_data_source` protocol).

Cancelling an offer is spec'd in the protocol as marking it as no longer valid. So we should be pretty careful not to use the offer anymore after a cancel. I've checked other calls to cancel, they seem fine.

include/wlr/types/wlr_data_device.h
emersion: Hmm, this is more tricky than I expected. If we split that into two structs, then the impl side won't be able to inspect the state side:

```c
struct wlr_data_source_state {
	bool accepted;
	struct wlr_data_offer *offer;
	struct wlr_seat_client *seat_client;

	enum wl_data_device_manager_dnd_action current_dnd_action;
	uint32_t compositor_action;
};

struct wlr_data_source {
	// source metadata
	struct wl_array mime_types;
	int32_t actions;

	// source implementation
	void (*send)(struct wlr_data_source *source, const char *mime_type,
		int32_t fd);
	void (*accept)(struct wlr_data_source *source, uint32_t serial,
		const char *mime_type);
	void (*cancel)(struct wlr_data_source *source);

	// drag'n'drop implementation
	void (*dnd_drop)(struct wlr_data_source *source);
	void (*dnd_finish)(struct wlr_data_source *source);
	void (*dnd_action)(struct wlr_data_source *source,
		enum wl_data_device_manager_dnd_action action);

	struct {
		struct wl_signal destroy;
	} events;
};
```

But I think we still want the impl to be able to inspect it (to implement security features depending on the client requesting the data for instance, or inspecting/changing the drag'n'drop action). So we could:

* Call all callbacks with the state, kinda breaks the impl pattern
* Keep the state in `wlr_data_source`, but in a child struct

martinetd: This feels a bit weird compared to usual? We normally have the state split the other way around, e.g. state is in wlr_data_source and there's another struct with implem in a pointer there e.g. something like that?
```
struct wlr_data_source;

struct wlr_data_source_impl {
	void (*send)(struct wlr_data_source *source, const char *mime_type,
		int32_t fd);
	void (*accept)(struct wlr_data_source *source, uint32_t serial,
		const char *mime_type);
	void (*cancel)(struct wlr_data_source *source);

	// drag'n'drop implementation
	void (*dnd_drop)(struct wlr_data_source *source);
	void (*dnd_finish)(struct wlr_data_source *source);
	void (*dnd_action)(struct wlr_data_source *source,
		enum wl_data_device_manager_dnd_action action);
};

struct wlr_data_source {
	// source metadata
	struct wl_array mime_types;
	int32_t actions;

	// source state
	bool accepted;
	struct wlr_data_offer *offer;
	struct wlr_seat_client *seat_client;

	enum wl_data_device_manager_dnd_action current_dnd_action;
	uint32_t compositor_action;

	struct wlr_data_source_impl *impl;
	struct {
		struct wl_signal destroy;
	} events;
};
```

(I might just as well be wrong :p)

emersion: Yeah we generally do that, but we can't here because `mime_types` and `actions` must be populated by `wlr_seat_set_selection`'s caller.

martinetd: I'm not sure what just putting the functions in an impl vector changes for that, can't we keep doing what we do right now? The user is then responsible of setting both `impl` and the metadata (`mime_types` and `actions`); and also its own responsibility to listen to the destroy event signal if they need to free mime_types. (actually since we have an impl vector it might be more fitting to add a destroy handler function instead of the destroy signal even if that's less generic, unsure if we need multiple listeners right now)
If you want to be more generic, you can replace these two fields by an anonymous `void *data` and fit that in a substruct, impl will be free to use that as they see fit? But that'd require more changes as we're currently aware of the fields in wlr_data_device itself. I think I'm missing the point though...

types/wlr_data_device.c
```diff
@@ -309,25 +285,30 @@ static void seat_client_selection_data_source_destroy(
...
 
 	wl_signal_emit(&seat->events.selection, seat);
 }
 
 void wlr_seat_set_selection(struct wlr_seat *seat,
 		struct wlr_data_source *source, uint32_t serial) {
-	if (seat->selection_source &&
+	if (source) {
+		assert(source->send);
+		assert(source->cancel);
```
acrisci: We need to be careful about what is required of the data source and fail as fast as possible when it is not present.

types/wlr_data_device.c
```diff
@@ -359,7 +340,8 @@ static void data_device_set_selection(struct wl_client *client,
 		wl_resource_get_user_data(dd_resource);
 
 	// TODO: store serial and check against incoming serial here
```
acrisci: I think we decided that serial validation is not necessary here so we can remove these todos.

types/wlr_data_device.c
```diff
@@ -433,7 +415,7 @@ static void wlr_drag_set_focus(struct wlr_drag *drag,
 		if (wl_resource_get_version(offer->resource) >=
 				WL_DATA_OFFER_SOURCE_ACTIONS_SINCE_VERSION) {
```
acrisci: Seems like you got rid of the version check everywhere but here. Will it crash a client if we send actions and they don't support it?

acrisci: Oh I see where the checks are now. This seems to be for the offer.

types/wlr_data_device.c
```diff
@@ -866,10 +866,10 @@ static void data_source_destroy(struct wl_client *client,
 
 static void data_source_set_actions(struct wl_client *client,
 		struct wl_resource *resource, uint32_t dnd_actions) {
-	struct wlr_data_source *source =
+	struct client_data_source *source =
 		wl_resource_get_user_data(resource);
 
-	if (source->actions_set) {
+	if (source->source.actions >= 0) {
```
acrisci: Can this miss the error if actions are "set" to zero?

emersion: No. The default value is -1 (it's set in `wlr_data_source_init`).

types/wlr_data_device.c
emersion: Yeah, that's the intent of these checks.

acrisci:
👍

Issue #168, opened 2017-09-28
Rootston todo


SirCmpwn:
Ordered loosely by priority

* [x] Subsurfaces (#160)
* [x] Moving windows around
  * [x] xdg shell
  * [x] wl shell
  * [x] Force any window to move with logo+lclick+drag
* [x] Resizing windows
  * [x] xdg shell
  * [x] wl shell
  * [x] xwayland
  * [x] Force any window to resize with logo+rclick+drag
* [x]  z ordering
* [x] Simple keybindings to launch programs with
* [x] support qt
* [x] xdg-popup

(simple daily driver becomes possible here)

* [x] Use client provided cursors
* [x] Fullscreen windows
* [x] Rotating windows with logo+mclick+drag
* [x] Touch
* [x] Tablet
* [x] Elegantly handle output hotplugging

emersion:
>Elegantly handle output hotplugging

What's that? BTW you can check "Fullscreen windows".

SirCmpwn:
Move all windows that were on the removed output to a remaining output, until all outputs are removed.

PR #544, opened 2017-12-31
Fix views outside output layout

emersion:
Test plan:
* Open some views on two outputs
* Remove one output (close output window or unplug it)
* There should be no views outside the output layout

Updates #168


SirCmpwn:
Thanks!

PR #546, opened 2017-12-31
Allow configurable verbosity

4e554c4c:


util/log.c
```diff
@@ -9,6 +9,7 @@
 #include <wlr/util/log.h>
 
 static bool colored = true;
+static log_importance_t v = L_SILENT;
```
martinetd: Would suggest a longer variable name for a global value. Like log_level or log_importance or whatever that won't be as annoying as 'v' in gdb :)

Also thinking default to silent is weird, but I don't really care.

4e554c4c: It's what we did in sway, but I think I'll change it since wlr should at least show error messages by default.

PR #548, opened 2018-01-02
Fix build without systemd

PoroCYon:
`WLR_HAS_ELOGIND` isn't a valid preprocessor expression, so the build fails (but only if `WLR_HAS_SYSTEMD` isn't defined).


PoroCYon:
Can someone tell me what the Travis error is? The site is blank even after enabling all JS/3rd party requests, tested both in Firefox and Qutebrowser...

emersion:
The Travis CI failed because of https://github.com/mikkeloscar/arch-travis/issues/33, it's not your fault.

Thanks for your contribution!

PR #549, opened 2018-01-04
Add wlr_output::enabled

emersion:
This un-exports `wlr_output_{create,destroy}_global`, since this is a protocol matter. This instead adds a `wlr_output_update_enabled` for backends.
  
This fixes disabled outputs with the DRM backend.

- [x] Fix cursor at (0, 0) when re-enabling an output. This happens because the output layout moves the cursor to (0, 0) when the output is disabled.
- [x] Fix failed pageflip. This happens when re-enabling the output, no idea how to fix it.

Test plan:

```ini
[bindings]
Logo+t = toggle_outputs
```

See @acrisci's comment for edge cases.

Related Weston patch: https://patchwork.freedesktop.org/patch/194357/

### DIFF
--- a/include/wlr/interfaces/wlr_output.h
+++ b/include/wlr/interfaces/wlr_output.h
@@ -26,14 +26,11 @@ struct wlr_output_impl {
 };
 
 void wlr_output_init(struct wlr_output *output, struct wlr_backend *backend,
-	const struct wlr_output_impl *impl);
-void wlr_output_free(struct wlr_output *output);
+	const struct wlr_output_impl *impl, struct wl_display *display);
 void wlr_output_update_mode(struct wlr_output *output,
 	struct wlr_output_mode *mode);
 void wlr_output_update_custom_mode(struct wlr_output *output, int32_t width,
 	int32_t height, int32_t refresh);
-struct wl_global *wlr_output_create_global(struct wlr_output *wlr_output,
-	struct wl_display *display);
-void wlr_output_destroy_global(struct wlr_output *wlr_output);
+void wlr_output_update_enabled(struct wlr_output *output, bool enabled);
 
 #endif
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -33,7 +33,6 @@ static bool atomic_end(int drm_fd, struct atomic *atom) {
 	}
 
 	uint32_t flags = DRM_MODE_ATOMIC_TEST_ONLY | DRM_MODE_ATOMIC_NONBLOCK;
-
 	if (drmModeAtomicCommit(drm_fd, atom->req, flags, NULL)) {
 		wlr_log_errno(L_ERROR, "Atomic test failed");
 		drmModeAtomicSetCursor(atom->req, atom->cursor);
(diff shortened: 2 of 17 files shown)


include/wlr/interfaces/wlr_output.h
```diff
@@ -26,14 +26,11 @@ struct wlr_output_impl {
 };
 
 void wlr_output_init(struct wlr_output *output, struct wlr_backend *backend,
-	const struct wlr_output_impl *impl);
-void wlr_output_free(struct wlr_output *output);
```
martinetd: No wlr_output_destroy or similar to replace that?

include/wlr/interfaces/wlr_output.h
```diff
@@ -26,14 +26,11 @@ struct wlr_output_impl {
...
+	const struct wlr_output_impl *impl, struct wl_display *display);
 void wlr_output_update_mode(struct wlr_output *output,
 	struct wlr_output_mode *mode);
 void wlr_output_update_custom_mode(struct wlr_output *output, int32_t width,
 	int32_t height, int32_t refresh);
-struct wl_global *wlr_output_create_global(struct wlr_output *wlr_output,
-	struct wl_display *display);
-void wlr_output_destroy_global(struct wlr_output *wlr_output);
+void wlr_output_update_enabled(struct wlr_output *output, bool enabled);
```
martinetd: I'm saying that without thinking, would it make sense for someone to disable the output, and re-enable it later with a different display than the one it was init with? I guess it's better for them to create a different output in that case, but I'm not really sure what display represents so just asking.

include/wlr/interfaces/wlr_output.h
emersion: There's `wlr_output_destroy` in `include/wlr/types/wlr_output.h`. `wlr_output_free` was not implemented.

acrisci:
Can you add some configuration for this in rootston to make it easier to demonstrate the feature?

There should be a config option to start with the output disabled, and maybe a command to enable/disable outputs at runtime.

emersion:
Disabling outputs with the DRM backend fails with `Atomic test failed: Invalid argument`. I've try to debug that, but I don't understand why it fails (cc @ascent12).

emersion:
Also, how should the output layout react to disabled outputs? Allocate a 0x0 area?

acrisci:
Edge cases I can think of without testing:

* We need to automatically update the output layout when the output is enabled/disabled (make sure the cursor does not go on the disabled output).
* Make sure the output frame callback doesn't do anything
* Move all the windows from the disabled output to an enabled one in rootston
* Make sure the enabled state persists when switching VTs

acrisci:
> Also, how should the output layout react to disabled outputs?

I think we should just treat it like it's not in the layout at all when it concerns any coordinate queries. The auto layout should reflow to reflect the output not being there. But leave it in there and make it come back when it gets enabled again.

emersion:
I think this is ready for a first review. Testers: there are two known issues detailed in the first comment.

emersion:
Concerning the output layout, as discussed on IRC, it makes more sense not to do anything special for disabled outputs. A compositor might want to remove an output from the output layout when disabled, but if a screen is turned off because the user is idle the compositor might not want the windows and the cursor to move.

emersion:
Bump

acrisci:
Did you fix the failed page flip issue?

emersion:
>Did you fix the failed page flip issue?

Fixed!

PR #551, opened 2018-01-04
meson: fix include path in generated .pc

martinetd:
The generated .pc still contained -I${prefix}/wlroots, this is no longer
needed at all as the prefix is now 'wlr' and all includes are 'wlr/foo'

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -159,7 +159,6 @@ subdir('examples')
 pkgconfig = import('pkgconfig')
 pkgconfig.generate(
 	libraries: lib_wlr,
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Good catch.

martinetd:
repushed because my commit message stripped ${prefix} as usual, I need to be more careful with setting message from shell with -m...

And, yeah, I kind of need this to work here :) It was still working for a while because I had both headers coexisting as I didn't clean my install directory.

SirCmpwn:
Thanks!

Issue #489, opened 2017-12-15
wlr_keyboard_set_keymap releases all modifiers


acrisci:
If the keymap is set in the modifiers handler, the xkb keyboard state will not reflect the state of the keys physically pressed on the keyboard. Upon the next keypress when modifiers are updated with `wlr_keyboard.c::keyboard_modifier_update()`, xkb always reports the modifiers are not pressed.

Setting the modifiers while keys are pressed is necessary to support proposed features in sway. A keyboard may be attached to different seats at the same time with different layouts per seat. For instance, attaching a keyboard to seat0 with us layout and attaching a keyboard to seat1 with a de layout will create two seat events for the same keyboard keypress event with each different layouts being reported to seat clients.

Modifiers are broken in rootston with this diff:

```diff
diff --git a/rootston/keyboard.c b/rootston/keyboard.c
index 4aaf2d4..b35f2fb 100644
--- a/rootston/keyboard.c
+++ b/rootston/keyboard.c
@@ -276,6 +276,7 @@ void roots_keyboard_handle_key(struct roots_keyboard *keyboard,
 
 void roots_keyboard_handle_modifiers(struct roots_keyboard *r_keyboard) {
 	struct wlr_seat *seat = r_keyboard->seat->seat;
+	wlr_keyboard_set_keymap(r_keyboard->device->keyboard, r_keyboard->device->keyboard->keymap);
 	wlr_seat_set_keyboard(seat, r_keyboard->device);
 	wlr_seat_keyboard_notify_modifiers(seat);
 }
```

acrisci:
If keyboard layout is now a per-seat-keyboard concept rather than a per-keyboard concept, then maybe all layout handling and xkb logic should be moved to the seat. To support this feature, there is no longer one xkb state per keyboard, but rather a number of xkb states, one for each seat.

Replace `wlr_keyboard_set_keymap()` with `wlr_seat_keyboard_set_keymap()` and move all state associated with xkb to the seat. This will allow the seat keyboard to keep a persistent keyboard layout rather than reporting extra layout switches to every seat client.

This has the effect of making wlr-keyboard extremely thin and almost stateless. It will become a simple router from input device events to keyboard events. We might want to consider removing this interface from wlroots and simply listening to keypress events on the input device itself since almost all of the responsibilities will be moved to the seat.

SirCmpwn:
That seems like a reasonable design to me, but I don't think we should remove wlr-keyboard entirely.

[...]

PR #553, opened 2018-01-05
compositor modifier hooks

acrisci:
Add the ability to change modifiers in flight with `wlr_seat_keyboard_notify_modifiers()` by giving the compositor control over what modifiers are sent to clients. Further decouples the device from the seat and makes it more consistent with the rest of the interface.

### DIFF
--- a/include/wlr/types/wlr_keyboard.h
+++ b/include/wlr/types/wlr_keyboard.h
@@ -52,7 +52,7 @@ struct wlr_keyboard {
 
 	uint32_t keycodes[WLR_KEYBOARD_KEYS_CAP];
 	size_t num_keycodes;
-	struct wlr_keyboard_modifiers *modifiers;
+	struct wlr_keyboard_modifiers modifiers;
 
 	struct {
 		int32_t rate;
--- a/include/wlr/types/wlr_keyboard.h
+++ b/include/wlr/types/wlr_keyboard.h
@@ -32,6 +32,13 @@ enum wlr_keyboard_modifier {
 
 struct wlr_keyboard_impl;
 
+struct wlr_keyboard_modifiers {
+	xkb_mod_mask_t depressed;
+	xkb_mod_mask_t latched;
+	xkb_mod_mask_t locked;
+	xkb_mod_mask_t group;
+};
+
 struct wlr_keyboard {
 	struct wlr_keyboard_impl *impl;
 	// TODO: Should this store key repeat info too?
@@ -45,12 +52,7 @@ struct wlr_keyboard {
 
 	uint32_t keycodes[WLR_KEYBOARD_KEYS_CAP];
 	size_t num_keycodes;
-	struct {
-		xkb_mod_mask_t depressed;
-		xkb_mod_mask_t latched;
-		xkb_mod_mask_t locked;
-		xkb_mod_mask_t group;
-	} modifiers;
+	struct wlr_keyboard_modifiers *modifiers;
 
 	struct {
 		int32_t rate;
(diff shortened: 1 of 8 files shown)


include/wlr/types/wlr_keyboard.h
```diff
@@ -45,12 +52,7 @@ struct wlr_keyboard {
 
 	uint32_t keycodes[WLR_KEYBOARD_KEYS_CAP];
 	size_t num_keycodes;
-	struct {
-		xkb_mod_mask_t depressed;
-		xkb_mod_mask_t latched;
-		xkb_mod_mask_t locked;
-		xkb_mod_mask_t group;
-	} modifiers;
+	struct wlr_keyboard_modifiers *modifiers;
```
emersion: No need to use a pointer here.

emersion:
To completely separate the seat from the `wlr_keyboard`, I'd like to change `wlr_seat_keyboard_notify_enter` and `wlr_seat_keyboard_enter` to take the currently pressed modifiers and keys.

include/wlr/types/wlr_keyboard.h
acrisci: Maybe you're right, but why not use a pointer?

acrisci:
> To completely separate the seat from the wlr_keyboard

There's still the layout.

include/wlr/types/wlr_keyboard.h
emersion: Because you have to allocate another data structure, check if it fails, free it. When accessing fields there's an extra pointer indirection (and maybe a cache miss). Here we don't really care about perf, but it's just more readable, easier and simpler.

emersion:
Right. "To allow the compositor to completely change modifiers and keys" is a better goal for this change.

acrisci:
Maybe the seat should have a "default keyboard" implementation of wlr-keyboard so there is always the keyboard capability of the seat.

emersion:
>Maybe the seat should have a "default keyboard" implementation of wlr-keyboard so there is always the keyboard capability of the seat.

Why? Some setups might not have a keyboard (like big touch screens).

acrisci:
The alternative is to be able to create a "dummy keyboard" which is not tied to a device and allow them to set that for the seat keyboard. That allows to configure xkb in the normal way without adding a bunch of stupid functions like `wlr_seat_set_default_keyboard_layout()` or something.

SirCmpwn:
>Maybe the seat should have a "default keyboard" implementation of wlr-keyboard so there is always the keyboard capability of the seat.

This sounds wrong and the justification doesn't make sense to me.

emersion:
What remains to be done here?

acrisci:
I think it's ready.

types/wlr_keyboard.c
```diff
@@ -109,14 +125,20 @@ void wlr_keyboard_notify_key(struct wlr_keyboard *keyboard,
...
+	}
+
 	keyboard_key_update(keyboard, event);
 	wl_signal_emit(&keyboard->events.key, event);
 }
 
 void wlr_keyboard_init(struct wlr_keyboard *kb,
 		struct wlr_keyboard_impl *impl) {
 	kb->impl = impl;
+	kb->modifiers = calloc(1, sizeof(struct wlr_keyboard_modifiers));
```
emersion: If there's a good enough reason to keep a pointer, this is missing a NULL check.

acrisci: I'll just get rid of the pointer.

emersion:
I think this fixes #489.

PR #554, opened 2018-01-05
sends the modifiers on wlr_seat_set_keyboard

Ongy:
Without this, a client will lose modifiers for one keyboard, when a key
is pressed on the other.
With this the client will always use the modifiers tate of the keyboard
the key was pressed on.

Testplan:
Connect 2 keyboard to the same seat, focus a window (I tested with weston-terminal).
Hold shift on one keyboard and press a character key, this should input capital letters.
Keep shift pressed, press a key on the other keyboard.
Press the key on the first keyboard again.
With this patch it should insert capital letters, without lower case.


SirCmpwn:
Nice, thanks!

PR #556, opened 2018-01-07
Idle protocol

raazvvann:
Add idle protocol

My question is if it's ok how I get notified about input events, because on rootstoon mouse motion on an empty desktop will lead to idle state bacause in this case rootston doesn't call wlr_seat_pointer_notify_motion
  
I am heghe on IRC

### DIFF
--- a/examples/idle.c
+++ b/examples/idle.c
@@ -1,0 +1,189 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <getopt.h>
+#include <pthread.h>
+#include <wayland-client.h>
+#include <wayland-client-protocol.h>
+#include <idle-client-protocol.h>
+
+static struct org_kde_kwin_idle *idle_manager = NULL;
+static struct wl_seat *seat = NULL;
+static uint32_t timeout = 0, simulate_activity_timeout = 0, close_timeout = 0;
+static int run = 1;
+
+struct thread_args {
+	struct wl_display *display;
+	struct org_kde_kwin_idle_timeout *timer;
+};
+
+static void handle_global(void *data, struct wl_registry *registry,
+		uint32_t name, const char *interface, uint32_t version) {
+	fprintf(stdout, "interfaces found: %s\n", interface);
+	if (strcmp(interface, "org_kde_kwin_idle") == 0) {
+		idle_manager = wl_registry_bind(registry, name, &org_kde_kwin_idle_interface, 1);
+	}
+	else if (strcmp(interface, "wl_seat") == 0) {
+		seat = wl_registry_bind(registry, name, &wl_seat_interface, 1);
+	}
+}
+
+static void handle_global_remove(void *data, struct wl_registry *registry, uint32_t name) {
+	//TODO
+}
+
+static const struct wl_registry_listener registry_listener = {
+	.global = handle_global,
+	.global_remove = handle_global_remove,
+};
+
+static void handle_idle(void* data, struct org_kde_kwin_idle_timeout *timer) {
+	fprintf(stdout, "idle state\n");
+}
+
+static void handle_resume(void* data, struct org_kde_kwin_idle_timeout *timer) {
+	fprintf(stdout, "active state\n");
+}
+
+static const struct org_kde_kwin_idle_timeout_listener idle_timer_listener = {
+	.idle = handle_idle,
+	.resumed = handle_resume,
+};
+
+int parse_args(int argc, char *argv[]) {
+	int c;
+	while ((c = getopt(argc, argv, "c:hs:t:")) != -1) {
+		switch(c)
+		{
+			case 'c':
+				close_timeout = strtoul(optarg, NULL, 10);
+				break;
+			case 's':
+				simulate_activity_timeout = strtoul(optarg, NULL, 10);
+				break;
+			case 't':
+				timeout = strtoul(optarg, NULL, 10);
+				break;
+			case 'h':
+			case '?':
+				printf("Usage: %s [OPTIONS]\n", argv[0]);
+				printf("  -t seconds\t\t\tidle timeout in seconds\n");
+				printf("  -s seconds optional\t\tsimulate user activity after x seconds\n");
+				printf("  -c seconds optional\t\tclose program after x seconds\n");
+				printf("  -h\t\t\t\tthis help menu\n");
+				return 1;
+			default:
+				return 1;
+		}
+	}
+	return 0;
+}
+
+void *simulate_activity(void *data) {
+	sleep(simulate_activity_timeout);
+	fprintf(stdout, "simulate user activity\n");
+	struct thread_args *arg = data;
+	org_kde_kwin_idle_timeout_simulate_user_activity(arg->timer);
+	wl_display_roundtrip(arg->display);
+	return NULL;
+}
+
+void *close_program(void *data) {
+	sleep(close_timeout);
+	struct thread_args *arg = data;
+	org_kde_kwin_idle_timeout_release(arg->timer);
+	wl_display_roundtrip(arg->display);
+	fprintf(stdout, "close program\n");
+	run = 0;
+	return NULL;
+}
+
+void *main_loop(void *data) {
+	struct wl_display *display = data;
+	while (wl_display_dispatch(display) != -1) {
+		;
+	}
+	return NULL;
+}
+
+int main(int argc, char *argv[]) {
+
+	if (parse_args(argc, argv) != 0) {
+		return -1;
+	}
+	if (timeout == 0) {
+		printf("idle timeout 0 is invalid\n");
+		return -1;
+	}
+
+	struct wl_display *display = wl_display_connect(NULL);
+	if (display == NULL) {
+		fprintf(stderr, "failed to create display\n");
+		return -1;
+	}
+
+	struct wl_registry *registry = wl_display_get_registry(display);
+	wl_registry_add_listener(registry, &registry_listener, NULL);
+	wl_display_dispatch(display);
+	wl_display_roundtrip(display);
+
+	if (idle_manager == NULL) {
+		fprintf(stderr, "display doesn't support idle protocol\n");
+		return -1;
+	}
+	if (seat== NULL) {
+		fprintf(stderr, "seat error\n");
+		return -1;
+	}
+	struct org_kde_kwin_idle_timeout *timer =
+		org_kde_kwin_idle_get_idle_timeout(idle_manager, seat, timeout * 1000);
+
+	if (timer == NULL) {
+		fprintf(stderr, "Could not create idle_timeout\n");
+		return -1;
+	}
+
+	pthread_t t1, t2, t3;
+	struct thread_args arg = {
+		.timer = timer,
+		.display = display,
+	};
+
+	if (simulate_activity_timeout != 0 && simulate_activity_timeout < close_timeout) {
+		if (pthread_create(&t1, NULL, &simulate_activity, (void *)&arg) != 0) {
+			return -1;
+		}
+	}
+	if (close_timeout != 0) {
+		if (pthread_create(&t2, NULL, &close_program, (void *)&arg) != 0) {
+			if (simulate_activity_timeout != 0) {
+				pthread_cancel(t1);
+			}
+			return -1;
+		}
+	}
+
+	org_kde_kwin_idle_timeout_add_listener(timer, &idle_timer_listener, timer);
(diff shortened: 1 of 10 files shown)


emersion:
Maybe `wlr_idle_notify_activity(idle, seat)` is better?

raazvvann:
@emersion You say create to have `wlr_idle_notify_activity` which will be called from compositor when there are input events? I wanted to stay in wlroots with input events notification, but what you say, if I understood correctly, seems nicer.

emersion:
Some events are not sent to the seat (e.g. pointer outside a view as you said, or compositor keybindings) so I'm not sure if it's possible to only listen to the `wlr_seat`.

raazvvann:
Ok. I get it. Thanks for the info. I'll finish this tomorrow.

types/wlr_idle.c
```diff
@@ -0,0 +1,159 @@
...
+#include <string.h>
+#include <wayland-server.h>
+#include <wlr/types/wlr_idle.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/util/log.h>
+#include "idle-protocol.h"
+
+static void idle_timeout_destroy(struct wlr_idle_timeout *timer) {
+	wl_list_remove(&timer->input_listener.link);
+	wl_event_source_timer_update(timer->idle_source, 0);
```
acrisci: need to call wl_event_source_remove?

acrisci:
The protocol seems to intend to specifically tie this to the activity of a seat and not the input devices themselves.

include/wlr/types/wlr_idle.h
```diff
@@ -0,0 +1,32 @@
+#ifndef WLR_TYPES_WLR_IDLE_H
+#define WLR_TYPES_WLR_IDLE_H
+
+#include <wayland-server.h>
+
```
acrisci: Can you add a bit of documentation to the header to explain what all this does?

acrisci:
Are there any applications that use this in the wild that we can test?

emersion:
>The protocol seems to intend to specifically tie this to the activity of a seat and not the input devices themselves.

Yeah, but in cases mentioned above, the seat activity isn't forwarded to the `wlr_seat`.

acrisci:
But what is "seat activity"? There are multiple ways to interpret this.

1. If you see the seat as a "collection of input devices", you could say that "seat activity" is any activity of an input device associated with the seat.

2. With a stricter reading of the protocol, you could interpret "seat activity" as only when the seat sends input events.

The example use case of the protocol given seems to me to work best with 1). In that case (since the seat doesn't know what devices are attached to it in our input model) would it be better to leave the input notifications entirely up to rootston since it is in charge of associating input devices with the seat, instead of this way which leaves the user in charge of input notification only some of the time?

types/wlr_idle.c
```diff
@@ -0,0 +1,159 @@
...
+	struct wlr_idle_timeout *timer = data;
+	timer->idle_state = true;
+	org_kde_kwin_idle_timeout_send_idle(timer->resource);
+	return 1;
+}
+
+static void handle_activity(struct wlr_idle_timeout *timer) {
+	// rearm the timer
+	wl_event_source_timer_update(timer->idle_source, timer->timeout);
+	// in case the previouse state was sleeping send a resume event and switch state
```
acrisci: typo here

emersion:
>leave the input notifications entirely up to rootston

That's exactly what's planned with `wlr_idle_notify_activity(idle, seat)`.

raazvvann:
I implemented `wlr_idle_notify_activity(seat)` but I don't send the wlr_idle, because I didn't want to add wlr_idle all over input events, so I use only the wlr_seat to pass on the notification. With this changes the compositor is responsible to send an activity notification when there is input events.

@acrisci AFAIK the kde plasma is using this protocol. But I implemented a simple idle example in which I used all the features of the protocol.

types/wlr_idle.c
```diff
@@ -0,0 +1,162 @@
...
+	idle->wl_global = wl_global_create(display, &org_kde_kwin_idle_interface, 1, idle, idle_bind);
+	if (idle->wl_global == NULL){
+		free(idle);
+		return NULL;
+	}
+	wlr_log(L_DEBUG, "idle manager created");
+	return idle;
+}
+
+void wlr_idle_notify_activity(struct wlr_seat *seat) {
```
emersion: This should be in the seat, and used by rootston.

rootston/seat.c
```diff
@@ -14,6 +15,7 @@
 static void handle_keyboard_key(struct wl_listener *listener, void *data) {
 	struct roots_keyboard *keyboard =
 		wl_container_of(listener, keyboard, keyboard_key);
+	wl_signal_emit(&keyboard->seat->seat->events.activity_notify, NULL);
```
acrisci: I don't like emitting the raw event on the seat signal from rootston. Having the signal on the seat just for the one protocol also doesn't seem right to me. What about notifying the wlr-idle itself from here? Sort of like `wlr_idle_notify_activity(seat)` which then handles the event on its own.

acrisci:
If a seat is destroyed, all of its associated timers should probably be destroyed as well.

rootston/seat.c
emersion: +1. `wlr_idle_notify_activity(idle, seat)` could iterate over all clients and update them accordingly.

raazvvann: that means that I should have wlr_idle accessible here in rootston seat? this will lead to removing wlr_seat form idle, but that's not a problem.

emersion: You can access idle from rootston via `desktop->idle`.

raazvvann:
The example it's a bit slow because of that `sleep(1)` in the main loop. I could use `wl_display_dispatch`, but then I couldn't close the program when the thread that should close the program if example it's started with the additional -c argument. If it's not ok, I could spent more time on this, but I think it's good enough.

rootston/seat.c
```diff
@@ -14,6 +15,8 @@
 static void handle_keyboard_key(struct wl_listener *listener, void *data) {
 	struct roots_keyboard *keyboard =
 		wl_container_of(listener, keyboard, keyboard_key);
+	struct roots_desktop *desktop = keyboard->input->server->desktop;
+	wl_signal_emit(&desktop->idle->activity_notify, NULL);
```
emersion: Why don't you call `wlr_idle_notify_activity` in rootston?

include/wlr/types/wlr_idle.h
```diff
@@ -0,0 +1,48 @@
...
+struct wlr_idle *wlr_idle_create(struct wl_display *display);
+
+void wlr_idle_destroy(struct wlr_idle *idle);
+
+/**
+ * Send notification to restart all timers. Called by compositor when
+ * there is an user activity event.
+ */
+void wlr_idle_notify_activity(struct wlr_idle *idle);
```
emersion: The seat on which the activity needs to be notified is missing.

rootston/seat.c
raazvvann: where else in rootston I could call this?

include/wlr/types/wlr_idle.h
raazvvann: I no longer need the seat now, because wlr_idle holds the wl_signal to notify the timers on user activity

rootston/seat.c
emersion: No, the place from which you notify activity (ie. adding a line here) is fine. But you're triggering the event manually. Can you call `wlr_idle_notify_activity` instead? (`wlr_idle_notify_activity` is unused otherwise)

include/wlr/types/wlr_idle.h
emersion: You need both the `wlr_idle` and the `wlr_seat`. When a client wants to be notified when a seat is idle, it doesn't care if another seat is not idle. In other words: when I move my mouse, the mouse is tied to a seat, and that should reset the idle timer for this seat only, not for all seats.

rootston/seat.c
raazvvann: You are totally right. My bad. Sorry about this.

types/wlr_idle.c
```diff
@@ -0,0 +1,177 @@
...
+			handle_timer_resource_destroy);
+	wl_list_insert(&idle->idle_timers, &timer->link);
+
+	timer->seat_destroy.notify = handle_seat_destroy;
+	wl_signal_add(&timer->seat->events.destroy, &timer->seat_destroy);
+
+	timer->input_listener.notify = handle_input_notification;
+	wl_signal_add(&idle->activity_notify, &timer->input_listener);
+	// create the timer
+	timer->idle_source = wl_event_loop_add_timer(idle->event_loop, idle_notify, timer);
```
emersion: Nitpick: add a NULL check here.

types/wlr_idle.c
```diff
@@ -0,0 +1,177 @@
...
+	idle->event_loop = wl_display_get_event_loop(display);
+	if (idle->event_loop == NULL) {
+		free(idle);
+		return NULL;
+	}
+
+	idle->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &idle->display_destroy);
+
+	idle->wl_global = wl_global_create(display, &org_kde_kwin_idle_interface, 1, idle, idle_bind);
```
emersion: Nitpick: 80 line length limit

examples/idle.c
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	pthread_t t1, t2;
+	struct thread_args arg = {
+		.timer = timer,
+		.display = display,
+	};
+
+	if (simulate_activity_timeout != 0 && simulate_activity_timeout < close_timeout) {
+		if(pthread_create(&t1, NULL, &simulate_activity, (void *)&arg) != 0)
```
acrisci: need a space after `if` (here and above) and always use braces.

examples/idle.c
```diff
@@ -0,0 +1,171 @@
...
+			}
+			return -1;
+		}
+	}
+
+	org_kde_kwin_idle_timeout_add_listener(timer, &idle_timer_listener, timer);
+	fprintf(stdout, "waiting\n");
+	while (run) {
+		wl_display_roundtrip(display);
+		sleep(1);
```
acrisci: This isn't the way to run the main loop. I think you want to use `wl_display_run()`.

examples/idle.c
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if (simulate_activity_timeout != 0)
+		pthread_join(t1, NULL);
```
acrisci: braces for all conditional blocks.

examples/meson.build
```diff
@@ -32,3 +32,10 @@ executable(
 	dependencies: [wayland_client, wlr_protos, wlroots],
 	link_with: lib_shared,
 )
+
+executable(
+	'idle',
+	'idle.c',
+	dependencies: [wayland_client, wlr_protos, wlroots, dependency('threads')],
```
acrisci: the 'threads' dependencies should be declared with the other dependencies.

types/wlr_idle.c
```diff
@@ -0,0 +1,188 @@
...
+	wl_signal_add(&idle->activity_notify, &timer->input_listener);
+	// create the timer
+	timer->idle_source =
+		wl_event_loop_add_timer(idle->event_loop, idle_notify, timer);
+	if (timer->idle_source == NULL) {
+		wl_list_remove(&timer->link);
+		wl_list_remove(&timer->input_listener.link);
+		wl_list_remove(&timer->seat_destroy.link);
+		wl_resource_set_user_data(timer->resource, NULL);
+		free(timer);
```
acrisci: post no memory to the client.

types/wlr_idle.c
```diff
@@ -0,0 +1,188 @@
...
+		return NULL;
+	}
+
+	idle->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &idle->display_destroy);
+
+	idle->wl_global = wl_global_create(display, &org_kde_kwin_idle_interface,
+				1, idle, idle_bind);
+	if (idle->wl_global == NULL){
+		free(idle);
```
acrisci: need to do the full cleanup here.

examples/idle.c
raazvvann: `wl_display_run` is part of the wayland server API. This example is suppose to be a wayland client. I'll rework the example to use `wl_display_dispatch`  and get rid of the `sleep`

emersion:
@acrisci It seems your comments have been addressed, are you okay to merge this?

acrisci:
👍

PR #559, opened 2018-01-13
adds remote argument to wayland backend create

Ongy:
Add a remote display name argument to wlr_wl_backend_create.
If NULL is passed to the wayland backend at all times, creating a
wayland backend *after* the compositor was started up, would require
changing the WAYLAND_DISPLAY environment variable.

### DIFF
--- a/include/wlr/backend/wayland.h
+++ b/include/wlr/backend/wayland.h
@@ -12,7 +12,7 @@
  * Creates a new wlr_wl_backend. This backend will be created with no outputs;
  * you must use wlr_wl_output_create to add them.
  */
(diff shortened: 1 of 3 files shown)


include/wlr/backend/wayland.h
```diff
@@ -12,7 +12,7 @@
  * Creates a new wlr_wl_backend. This backend will be created with no outputs;
  * you must use wlr_wl_output_create to add them.
  */
-struct wlr_backend *wlr_wl_backend_create(struct wl_display *display);
+struct wlr_backend *wlr_wl_backend_create(struct wl_display *display, const char *remote);
```
SirCmpwn: Can you update the doc comment here

PR #567, opened 2018-01-16
Fixed logging for examples

Timidger:
Recent change means you need to call `wlr_log_init`, which wasn't being done in the examples.


SirCmpwn:
Thanks!

PR #569, opened 2018-01-16
rootston: hide cursor if seat has only keyboards

emersion:
Also set the real seat capabilities instead of always all of them.

Test plan: create a seat with a mouse and a keyboard. Unpluging the mouse should hide the pointer.

Updates https://github.com/swaywm/sway/issues/1568

### DIFF
--- a/rootston/seat.c
+++ b/rootston/seat.c
@@ -313,6 +313,14 @@ static void seat_update_capabilities(struct roots_seat *seat) {
 		caps |= WL_SEAT_CAPABILITY_TOUCH;
 	}
 	wlr_seat_set_capabilities(seat->seat, caps);
+
+	// Hide cursor if seat doesn't have pointer capability
+	if ((caps & WL_SEAT_CAPABILITY_POINTER) == 0) {
+		wlr_cursor_set_image(seat->cursor->cursor, NULL, 0, 0, 0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


rootston/seat.c
```diff
@@ -306,6 +301,28 @@ struct roots_seat *roots_seat_create(struct roots_input *input, char *name) {
...
+	if (!wl_list_empty(&seat->pointers) || !wl_list_empty(&seat->tablet_tools)) {
+		caps |= WL_SEAT_CAPABILITY_POINTER;
+	}
+	if (!wl_list_empty(&seat->touch)) {
+		caps |= WL_SEAT_CAPABILITY_TOUCH;
+	}
+	wlr_seat_set_capabilities(seat->seat, caps);
+
+	// Hide cursor if seat doesn't have pointer or touch capability
+	if ((caps & (WL_SEAT_CAPABILITY_POINTER | WL_SEAT_CAPABILITY_TOUCH)) == 0) {
```
acrisci: I think you can hide the cursor if the seat has the touch capability but no pointer right? I think touch points in the current implementation never move the cursor, but I'd have to recheck.

emersion: Oh, you're right. I checked that but incorrectly read `roots_cursor_press_button`.

PR #570, opened 2018-01-17
Decorations

acrisci:
Add decorations to rootston xwayland views.

Feedback on the approach would be great.

Tests:

* mouse over decoration borders and the correct cursor should show up
* start an interactive move by dragging the titlebar
* start an interactive resize by dragging the borders
* try again with the view rotated


types/wlr_output.c
```diff
@@ -751,3 +751,46 @@ enum wl_output_transform wlr_output_transform_compose(
 	}
 	return flipped | rotated;
 }
+
+void wlr_output_get_box_matrix(struct wlr_output *output, int ox, int oy,
+		int width, int height, enum wl_output_transform transform,
+		float rotation, float (*mat)[16]) {
+		float translate_center[16];
```
martinetd: identation here looks wrong, continues with two tabs within function.

martinetd:
Also, rotating just a window does rotate the decoration but the position is subtly off (probably need to move rotation center) and the edges/titlebar positions for move aren't rotated so are pretty hard to find (might be a problem in the coordinates seat_view_deco_motion receives that aren't in the view's rotated world).

SirCmpwn:
On the wayland backend at least, interactive move+resize does not work for me.

rootston/cursor.c
```diff
@@ -30,19 +30,81 @@ void roots_cursor_destroy(struct roots_cursor *cursor) {
...
+static void seat_view_deco_button(struct roots_seat_view *view, double vx,
+		double vy, uint32_t button, uint32_t state) {
+	if (button == BTN_LEFT && state == WLR_BUTTON_PRESSED) {
+		view->has_button_grab = true;
+		view->grab_vx = vx;
+		view->grab_vy = vy;
+	} else {
+		view->has_button_grab = false;
+	}
+}
```
SirCmpwn: I would prefer to establish a decorations file and move this (and other relevant logic) there.

rootston/output.c
```diff
@@ -53,46 +53,8 @@ static void render_surface(struct wlr_surface *surface,
...
-				&transform);
-			wlr_matrix_mul(&transform, &surface_transform, &transform);
-			wlr_matrix_mul(&transform, &surface_translate_origin,
-				&transform);
-		}
-
-		wlr_matrix_mul(&wlr_output->transform_matrix, &transform, &matrix);
-
+		wlr_output_get_box_matrix(wlr_output, ox, oy, render_width,
+			render_height, surface->current->transform, rotation, &matrix);
```
SirCmpwn: :+1:

types/wlr_output.c
SirCmpwn: bump

SirCmpwn:
I like it.

acrisci:
Should the view `x` and `y` struct members refer to the coordinates of the window or the top left corner of the decorations?

SirCmpwn:
The former

rootston/output.c
acrisci: @emersion do you like this abstraction?

emersion: * Make it take a `wlr_box` instead of four parameters
* Coordinates are in output-local coordinates (multiplied by `output.scale`), if it's an output function I'd expect coordinates to be in surface-local coordinates and the function to take care of multiplying (as it already takes care of `output->transform`)
* This function has nothing to do with the output, apart from multiplying the result to the output matrix at the end. I'd prefer this to be a `wlr_matrix` function.
* Rotation is opiniated and rootston-specific

SirCmpwn: >Make it take a wlr_box instead of four parameters

:+1: 

>Coordinates are in output-local coordinates (multiplied by output.scale), if it's an output function I'd expect coordinates to be in surface-local coordinates and the function to take care of multiplying (as it already takes care of output->transform)

:-1: 

>Rotation is opiniated and rootston-specific

Not really. This function appears to me to be box-oriented rather than surface-oriented and rotating boxes isn't particularly opinionated.

SirCmpwn: >This function has nothing to do with the output, apart from multiplying the result to the output matrix at the end. I'd prefer this to be a wlr_matrix function.

I'd need to see the proposed function signature of it as it appears in wlr_matrix.

emersion: >I'd need to see the proposed function signature of it as it appears in wlr_matrix.

```c
void wlr_matrix_box_transform(float (*mat)[16], struct wlr_box *box,
	enum wl_output_transform transform, float rotation);
```

Still searching for a better name.

SirCmpwn: Maybe it'd be better on wlr_box rather than wlr_matrix? `wlr_box_transform`

emersion: This name seems like it transforms a box to another box. I'd like to have `wlr_matrix`, `box` and `transform` in the name. Maybe `wlr_matrix_transform_box`? `wlr_matrix_from_transformed_box`?

SirCmpwn: `wlr_box_projection`?

emersion: That sounds better. But this function really belongs to `render/matrix.c` IMHO. `wlr_matrix_project_box`? `wlr_matrix_projection`?

acrisci: Doesn't this matrix have an official name? Is this a "model" matrix or a "view" matrix?

SirCmpwn: I'm not sure that the lingo maps that precisely to our circumstances.

acrisci: I don't think we can use wlr_box for this as-is because it has `x` and `y` in integers but we need it to be a double or a float to match the output coordinates.

emersion: That's not true. Coordinates are output-locale, which means using a double will allow rendering a view between two pixels. We had blurry rendering because of this in the past.

acrisci:
All of @martinetd 's issues are fixed and this is ready for another round of testing.

emersion:
Resize/move via decorations doesn't work for me (Wayland backend, tried Chromium/Firefox/gnome-calculator under xwayland).

Also we may want not to draw decorations when the client does it (Chromium, gnome-calculator under xwayland).

acrisci:
yeah that's what the motif hints are for.

acrisci:
Fixed the outstanding issues, but still getting the old "cursor bounce" issue when setting the cursor for the decorations on DRM backend. Not really sure how to solve it.

acrisci:
@emersion do you see the cursor bounce issues on the DRM backend? I remember you fixed them one time. What am I doing wrong?

emersion:
@acrisci Are you referring to the cursor have an incorrect hotspot for a short period of time? What I fixed is the cursor appearing at its previous location for a short period of time. This seems like another issue. Can you fill a bug?

PR #578, opened 2018-01-22
cursor: fix segfault when destroying a wlr_output_layout with a wlr_cursor attached

emersion:
cc @Timidger


Timidger:
Awesome, thanks!

Issue #568, opened 2018-01-16
Decouple DPMS from wl_output globals


Ongy:
Sorry, missed it before, but https://github.com/swaywm/wlroots/pull/549 does a something I feel like it shouldn't.

So this PR couples the `wl_outupt` globals to the enabled/disabled state of the output. This state also determines the DPMS state on DRM backend.
IMO this isn't the best idea. The general hierarchy of output states I can think of is roughly:
* Enabled (DPMS/Renderloop on, in layout, wl_output exists)
* Sleep (DPMS/Renderloop off, in layout, wl_output exists)
     This is for powersaving

     Things like the desktop background, notification-daemon etc. shouldn't react to this. IMO clients shouldn't get to know at all (outside done callbacks)
* Disabled (DPMS/Renderloop off, wl_output *may* exist, **not** in layout)
    Plugged in, but not configured, should exist in whatever is going to replace xrandr.

    Desktop background/notifaction-daemon etc. should consider this as off, don't try to appear on it.
    **Problem:** I can't see an enabled flag in `wl_output`, so this may require the split protocols KDE has current
* Unplugged -- Doesn't exist, doesn't show up for `xrandr`-ish application.

IIRC the KDE protocol for setting this up had a custom information protocol attached as well, which could be used for the disabled state.
Then the `wl_output` could be bound to the `wlr_output_layout` and have the disabled state controlled by the compositor removing the output from the layout.

emersion:
I don't see why clients should be aware of sleeping outputs. That would break the screenshooter. Also, the `wlr_output` is not aware weather it's sleeping or disabled.

KDE has another interface ([`outputdevice`](https://github.com/KDE/kwayland/blob/master/src/client/protocols/outputdevice.xml)) for outputs not exported as `wl_output`s.

Ongy:
I'd consider the screenshooter more of an edgecase here. I don't know the protocol, is there no way to tell the application it cant get this output?
I'd prefer (and expect) other things that interact with `wl_output` not to react to powersave features.
And there are more things that interact with `wl_output`
e.g. xdg-shell fullscreen, I would expect `surface_layers`
and probably others. (pointer constraints? Mapping drawing tablets to areas?)

And sure, I would bind the sleep/disabled states here (I just made the names up) to whether the output is in the `wlr_output_layout` or not.
If that doesn't make sense/wouldn't work with wlroots, that's nothing important, just my current idea.

Ongy:
[...]

PR #579, opened 2018-01-23
decouples outputenable state and wl_output global

Ongy:
This decouples wlr_output_enable and the wl_global.
The previously internal functions wlr_output_(destroy/create)_global are
exposed and used automatically in the wlr_output_layout to create/tear
down the global.
The compositor can handle them itself if it wants to, but I think this
is the right moment to create/destroy the wl_output when the
wlr_output_layout is used.

Examples are not updates as they either use `wlr_output_layout` or won't make use of the `wl_output` either way.

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -89,8 +89,8 @@ struct wlr_output {
 struct wlr_surface;
 
 void wlr_output_enable(struct wlr_output *output, bool enable);
-void wlr_output_create_global();
-void wlr_output_destroy_global();
+void wlr_output_create_global(struct wlr_output *output);
+void wlr_output_destroy_global(struct wlr_output *output);
 bool wlr_output_set_mode(struct wlr_output *output,
 	struct wlr_output_mode *mode);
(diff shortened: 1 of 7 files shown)


Ongy:
Forgot to mention, this is for https://github.com/swaywm/wlroots/issues/568

The one thing I'd consider as well is to set outputs into a disabled state by default. But since being disabled is currently a drm specific thing, I think it's better to start in the mode all backends handle well.

SirCmpwn:
>The one thing I'd consider as well is to set outputs into a disabled state by default.

Hmm, probably not. But I don't feel strongly about it.

include/wlr/types/wlr_output.h
```diff
@@ -89,6 +89,8 @@ struct wlr_output {
 struct wlr_surface;
 
 void wlr_output_enable(struct wlr_output *output, bool enable);
+void wlr_output_create_global();
+void wlr_output_destroy_global();
```
emersion: This is missing the `output` parameter.

emersion:
Thanks!

PR #580, opened 2018-01-23
Backport screenshooter fixes from the renderer redesign v1

emersion:
This backports some changes to #319 to fix the screenshooter data format. This also adds `wlr_backend_get_renderer` which will be useful to support multiple renderers.

### DIFF
--- a/backend/multi/backend.c
+++ b/backend/multi/backend.c
@@ -62,10 +62,24 @@ static struct wlr_egl *multi_backend_get_egl(struct wlr_backend *wlr_backend) {
 	return NULL;
 }
 
+static struct wlr_renderer *multi_backend_get_renderer(
+		struct wlr_backend *backend) {
+	struct wlr_multi_backend *multi = (struct wlr_multi_backend *)backend;
+	struct subbackend_state *sub;
+	wl_list_for_each(sub, &multi->backends, link) {
+		struct wlr_renderer *rend = wlr_backend_get_renderer(sub->backend);
+		if (rend != NULL) {
+			return rend;
+		}
+	}
+	return NULL;
+}
+
 struct wlr_backend_impl backend_impl = {
 	.start = multi_backend_start,
 	.destroy = multi_backend_destroy,
 	.get_egl = multi_backend_get_egl,
+	.get_renderer = multi_backend_get_renderer,
 };
 
 static void handle_display_destroy(struct wl_listener *listener, void *data) {
--- a/examples/screenshot.c
+++ b/examples/screenshot.c
@@ -74,7 +74,7 @@ static void output_handle_mode(void *data, struct wl_output *wl_output,
 }
 
 static void output_handle_done(void *data, struct wl_output *wl_output) {
-
+	// No-op
 }
 
 static const struct wl_output_listener output_listener = {
@@ -148,16 +148,6 @@ static struct wl_buffer *create_shm_buffer(int width, int height,
 	return buffer;
 }
 
-static void argb_to_rgba(uint32_t *data, size_t height, size_t stride) {
-	size_t n = height*stride/4;
-	for (size_t i = 0; i < n; ++i) {
-		uint32_t v = data[i];
-		uint32_t rgb = v & 0x00ffffff;
-		uint32_t a = (v & 0xff000000) >> 24;
-		data[i] = (rgb << 8) | a;
-	}
-}
-
 static void write_image(const char *filename, int width, int height) {
 	int buffer_stride = width * 4;
 
@@ -181,8 +171,6 @@ static void write_image(const char *filename, int width, int height) {
 
 		free(output);
 	}
-
-	argb_to_rgba(data, height, buffer_stride);
 
 	char size[10 + 1 + 10 + 2 + 1]; // int32_t are max 10 digits
 	sprintf(size, "%dx%d+0", width, height);
@@ -213,7 +201,9 @@ static void write_image(const char *filename, int width, int height) {
 			exit(EXIT_FAILURE);
 		}
 		close(fd[0]);
-		execlp("convert", "convert", "-depth", "8", "-size", size, "rgba:-",
+		// We requested WL_SHM_FORMAT_XRGB8888 in little endian, so that's BGRA
+		// in big endian.
+		execlp("convert", "convert", "-depth", "8", "-size", size, "bgra:-",
 			"-alpha", "opaque", filename, NULL);
 		fprintf(stderr, "cannot execute convert\n");
 		exit(EXIT_FAILURE);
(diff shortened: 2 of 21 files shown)


backend/multi/backend.c
```diff
@@ -62,10 +62,24 @@ static struct wlr_egl *multi_backend_get_egl(struct wlr_backend *wlr_backend) {
 	return NULL;
 }
 
+static struct wlr_renderer *multi_backend_get_renderer(
+		struct wlr_backend *backend) {
```
acrisci: It's a separate issue, but I believe this operation should not be supported for the multi-backend because it could have multiple renderers.

examples/screenshot.c
```diff
@@ -213,7 +201,9 @@ static void write_image(const char *filename, int width, int height) {
 			exit(EXIT_FAILURE);
 		}
 		close(fd[0]);
-		execlp("convert", "convert", "-depth", "8", "-size", size, "rgba:-",
+		// We requested WL_SHM_FORMAT_XRGB8888 in little endian, so that's BGRA
+		// in big endian.
+		execlp("convert", "convert", "-depth", "8", "-size", size, "bgra:-",
```
acrisci: Doesn't this come from ImageMagick? Does that need to be added to the dependencies for the library?

render/gles2/renderer.c
```diff
@@ -211,20 +211,33 @@ static bool wlr_gles2_buffer_is_drm(struct wlr_renderer *_renderer,
...
+
+	// Make sure any pending drawing is finished before we try to read it
+	glFinish();
+
+	// Unfortunately GLES2 doesn't support GL_PACK_*, so we have to read
+	// the lines out row by row
+	unsigned char *p = data + dst_y * stride;
+	for (size_t i = src_y; i < src_y + height; ++i) {
+		glReadPixels(src_x, src_y + height - i - 1, width, 1, fmt->gl_format,
+			fmt->gl_type, p + i * stride + dst_x * 4);
```
acrisci: Is the `* 4` in these calculations format dependent or is 4 the correct value for all formats?

types/wlr_screenshooter.c
```diff
@@ -41,27 +30,32 @@ static void handle_screenshot_resource_destroy(
 	}
 }
 
-static void output_frame_notify(struct wl_listener *listener, void *_data) {
+static void output_handle_frame(struct wl_listener *listener, void *_data) {
```
acrisci: I think we've settled on something like `handle_output_frame` for callbacks but there's no accepted standard yet.

types/wlr_screenshooter.c
```diff
@@ -41,27 +30,32 @@ static void handle_screenshot_resource_destroy(
...
-	wl_shm_buffer_end_access(state->shm_buffer);
-
-	free(state->pixels);
-	wl_list_remove(&listener->link);
+	enum wl_shm_format format = wl_shm_buffer_get_format(shm_buffer);
+	int32_t width = wl_shm_buffer_get_width(shm_buffer);
+	int32_t height = wl_shm_buffer_get_height(shm_buffer);
+	int32_t stride = wl_shm_buffer_get_stride(shm_buffer);
+	void *data = wl_shm_buffer_get_data(shm_buffer);
+	wl_shm_buffer_begin_access(shm_buffer);
```
acrisci: probably needs to be swapped with the line above (begin access before reading the buffer data).

acrisci:
This is not orbital-screenshooter. This is weston-screenshooter!

https://github.com/wayland-project/weston/blob/master/protocol/weston-screenshooter.xml

https://github.com/giucam/orbital/blob/master/protocol/screenshooter.xml

types/wlr_screenshooter.c
```diff
@@ -41,27 +30,32 @@ static void handle_screenshot_resource_destroy(
...
+	bool ok = wlr_renderer_read_pixels(renderer, format, stride, width, height,
+		0, 0, 0, 0, data);
+	wl_shm_buffer_end_access(shm_buffer);
+
+	if (!ok) {
+		wlr_log(L_ERROR, "Cannot read pixels");
+		goto cleanup;
```
acrisci: is it a problem that the `done` event never gets emitted in this case? The actual orbital screenshooter has a `fail` event for this case.

types/wlr_screenshooter.c
```diff
@@ -127,17 +120,12 @@ static void screenshooter_shoot(struct wl_client *client,
 	if (!state) {
 		wl_resource_destroy(screenshot->resource);
 		free(screenshot);
-		free(pixels);
 		wl_resource_post_no_memory(screenshooter_resource);
```
acrisci: Doesn't posting no memory to a client cause it to be disconnected? In that case you don't need to destroy the resource above if it belongs to that client because it will get cleaned up anyway (I may be wrong).

acrisci:
Screenshots do not follow the output layout.

acrisci:
It takes a screenshot on the next frame, but is the frame callback guaranteed to be called quickly when the new output damage pr is merged?

emersion:
>This is not orbital-screenshooter. This is weston-screenshooter!

What do you mean? We use Orbital screenshooter, not Weston's.

>Screenshots do not follow the output layout.

Works for me. More details please?

>It takes a screenshot on the next frame, but is the frame callback guaranteed to be called quickly when the new output damage pr is merged?

Handled in the PR.

backend/multi/backend.c
emersion: We do need it for the multi libinput+drm, just like EGL.

examples/screenshot.c
emersion: No, it is a runtime dependency. It's also a dependency specific to this example.

types/wlr_screenshooter.c
emersion: I prefer `<type>_handle_<event>`.

types/wlr_screenshooter.c
emersion: Nah, we're just getting a pointer so it doesn't matter. I just copied the Wayland docs example btw.

types/wlr_screenshooter.c
emersion: It seems we have an older version of the protocol.

types/wlr_screenshooter.c
emersion: Meh, the code is more readable that way.

acrisci:
> What do you mean? We use Orbital screenshooter, not Weston's.

Actually the protocol exactly matches Weston's protocol. I think if you change the name, then any Weston screenshot program will work. Weston has `weston-screenshooter`. It will however not work with the latest orbital protocol. It looks like they made breaking changes without bumping the version.

emersion:
>Actually the protocol exactly matches Weston's protocol.

Nah, it has a `screenshot` object that Weston doesn't have. I think upgrading the protocol is a different task, that could be filled as a separate issue (I also don't like Orbital's changes to the protocol).

acrisci:
Ok let's just keep an eye on those issues. I'll make real issues when I get around to it. Otherwise it looks good. 👍 .

PR #582, opened 2018-01-24
[old/reference] Roots decoration manager

acrisci:
Implement the decoration manager into rootston.

To enable ssd:

```ini
[core]
decoration-mode=server
#decoration-mode=client
#decoration-mode=none
```

TODO:

* [ ] find some program that implements the protocol correctly for testing

Support status:

Surprisingly not supported by qt apps on my system (tried konsole, qgit, qtcreator, kcalc on fedora 27).

gnome-terminal supports it but support is not correct. It never actually creates the decoration protcol object. It checks the default mode and uses that but never tells us what mode it's using.

gnome-calendar gives us a decoration protocol object but then tells us it wants client decorations.

### DIFF
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -179,7 +179,13 @@ static void render_decorations(struct roots_view *view,
 	float matrix[16];
 	wlr_matrix_project_box(&matrix, &project_box, WL_OUTPUT_TRANSFORM_NORMAL,
 		view->rotation, &output->transform_matrix);
-	float color[4] = { 0.2, 0.2, 0.2, 1 };
+	float color[4]	= { 0.2, 0.2, 0.2, 1 };
+	if (view_get_activated(view)) {
+		color[0] = 0.09804;
+		color[1] = 0.09804;
+		color[2] = 0.43921;
+	}
+
 	wlr_render_colored_quad(desktop->server->renderer, &color, &matrix);
 }
 
--- a/include/rootston/config.h
+++ b/include/rootston/config.h
@@ -2,6 +2,7 @@
 #define _ROOTSTON_CONFIG_H
 #include <wlr/types/wlr_output_layout.h>
 #include <wlr/types/wlr_input_device.h>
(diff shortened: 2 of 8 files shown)


SirCmpwn:
>GTK supports it but support is not correct. It never actually creates the decoration protcol object. It checks the default mode and uses that but never tells us what mode it's using.

It only announces it if wants to use client side decorations, I think. This may be the incorrect behavior.

SirCmpwn:
The patch

https://git.gnome.org/browse/gtk+/commit/?h=gtk-3-22&id=f2adaba237519642b88ccb42b47a95cf539a2a12

emersion:
Have you tried some KDE app?

acrisci:
> Have you tried some KDE app?

Updated support status. I cannot find a program that uses the protocol correctly and chooses ssd.

acrisci:
I believe what I have here is the complete protocol implemented in rootston. Now we just need to find out how to test it to confirm it works.

rootston/output.c
```diff
@@ -179,7 +179,13 @@ static void render_decorations(struct roots_view *view,
 	float matrix[16];
 	wlr_matrix_project_box(&matrix, &project_box, WL_OUTPUT_TRANSFORM_NORMAL,
 		view->rotation, &output->transform_matrix);
-	float color[4] = { 0.2, 0.2, 0.2, 1 };
+	float color[4]	= { 0.2, 0.2, 0.2, 1 };
+	if (view_get_activated(view)) {
+		color[0] = 0.09804;
+		color[1] = 0.09804;
+		color[2] = 0.43921;
+	}
```
acrisci: I'm not good at picking colors, if anyone else has a preference.

acrisci:
> It only announces it if wants to use client side decorations, I think. This may be the incorrect behavior.

Do you want to fix it or should I?

SirCmpwn:
>Do you want to fix it or should I?

I'm not picky, I'll try to get to it this weekend if you don't.

PR #585, opened 2018-01-25
meson: add tags and ctags targets

agx:
to create tag files. Idea taken from systemd.

Signed-off-by: Guido Günther <agx@sigxcpu.org>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -180,3 +180,31 @@ pkgconfig.generate(
 	name: meson.project_name(),
 	description: 'Wayland compositor library',
 )
+
+git = find_program('git', required: false)
+if git.found()
+	all_files = run_command(
+		git,
+		['--git-dir=@0@/.git'.format(meson.source_root()),
+		 'ls-files',
+		 ':/*.[ch]'])
+	all_files = files(all_files.stdout().split())
+
+	etags = find_program('etags', required: false)
+	if etags.found() and all_files.length() > 0
+		custom_target('etags',
+		   build_by_default: true,
+		   input: all_files,
+		   output: 'TAGS',
+		   command: [etags.path(), '-o', 'TAGS'] + all_files)
+	endif
+
+	ctags = find_program('ctags', required: false)
+	if ctags.found() and all_files.length() > 0
+		custom_target('ctags',
+		   build_by_default: true,
+		   input: all_files,
+		   output: 'tags',
+		   command: [ctags.path(), '-o', 'tags'] + all_files)
+	endif
+endif


meson.build
```diff
@@ -180,3 +180,18 @@ pkgconfig.generate(
 	name: meson.project_name(),
 	description: 'Wayland compositor library',
 )
+
+git = find_program('git', required : false)
```
SirCmpwn: This should also find_program etags and skip it if not installed.

meson.build
```diff
@@ -180,3 +180,18 @@ pkgconfig.generate(
...
+if git.found()
+        all_files = run_command(
+                git,
+                ['--git-dir=@0@/.git'.format(meson.source_root()),
+                 'ls-files',
+                 ':/*.[ch]'])
```
acrisci: Should check if this fails, for instance if the `.git` directory isn't there.

acrisci:
Tried it and it didn't work. No `tags` file was generated on `ninja -C build`.

ascent12:
It's a run target, so you would need to call it as `ninja -C build ctags`

SirCmpwn:
It should probably just happen every time if we can make that work.

martinetd:
It's modifying stuff in the source tree, not the build dir, so I'd rather not build this everytime. Having a build system that works perfectly fine out of tree and adding stuff that want to write in source tree is a pain (yes I'm the only one who put my source in a read-only mount, but that happens); also given how this works I don't think it can use meson dependencies to regenerate on source code change so it's a bit ad-hoc to tell when to rerun.
Should also add /TAGS and /tags to .gitignore

martinetd:
FWIW, if you want to rerun this every time, as things stand you just need to run `ninja -C build all tags` so it's not so bad.

SirCmpwn:
We could make it output tags to the build dir. Vim lets you specify the path to the tags file.

martinetd:
I'm fine with adding it to default target if it goes to build dir -- happy either way.

agx:
Updated. Changes

- name etags target etags not tags for consistency
- check for etags/ctags as well
- don't fail on empty list (e.g. when not in git)
- .gitignore tag files

SirCmpwn:
Would still like to see this generating on every build. Doing two commands is annoying.

agx:
Updated to create tags files in build directory and on every build (with changes)

meson.build
```diff
@@ -180,3 +180,31 @@ pkgconfig.generate(
...
+		 ':/*.[ch]'])
+	all_files = files(all_files.stdout().split())
+
+	etags = find_program('etags', required : false)
+	if etags.found() and all_files.length() > 0
+		custom_target('etags',
+		   build_by_default : true,
+		   input : all_files,
+		   output : 'TAGS',
+		   command : ['env', 'etags', '-o', 'TAGS'] + all_files)
```
SirCmpwn: Use etags.path() instead of env

agx:
Yeah, that's better. using .path() now.

meson.build
```diff
@@ -180,3 +180,31 @@ pkgconfig.generate(
...
+		git,
+		['--git-dir=@0@/.git'.format(meson.source_root()),
+		 'ls-files',
+		 ':/*.[ch]'])
+	all_files = files(all_files.stdout().split())
+
+	etags = find_program('etags', required : false)
+	if etags.found() and all_files.length() > 0
+		custom_target('etags',
+		   build_by_default : true,
```
emersion: Nitpick: no space before `:`

agx:
Spaces removed now.

emersion:
Thanks!

PR #587, opened 2018-01-26
Ignore emacs backup files

agx:


SirCmpwn:
https://www.emacswiki.org/emacs/BackupDirectory

agx:
@SirCmpwn I know but vim swap files are being ignored as well so it makes things simpler for folks that haven't tweaked things (or that clone/build wlroots on a system with not much emacs setup)

SirCmpwn:
https://github.com/swaywm/wlroots/commit/ab738256ac3cdaadb33dcd6d59accc876c45a51a

Timidger:
You could also just add it to your global git ignore file, like what I do.

PR #588, opened 2018-01-26
Fix warnings on arm-linux

agx:


backend/drm/drm.c
```diff
@@ -459,7 +459,7 @@ static bool wlr_drm_connector_set_mode(struct wlr_output *output,
 	if (!crtc) {
 		return false;
 	}
-	wlr_log(L_DEBUG, "%s: crtc=%ju ovr=%jd pri=%jd cur=%jd", conn->output.name,
+	wlr_log(L_DEBUG, "%s: crtc=%zd ovr=%zd pri=%zd cur=%zd", conn->output.name,
```
ascent12: `%zd` refers to a ssize_t, which definitely isn't the right type.
Since it's subtracting pointers, it's a ptrdiff_t, so the correct format would be `%td`.

agx:
@ascent12 thanks, forgot about ptrdiff_t. Fixed now.

PR #590, opened 2018-01-28
Fix building rootston without XWayland

pks-t:
Just recently I tried compiling wlroots without XWayland support, which is obviously broken due to some missing #ifdefs. This problem is easy to fix by itself, but I went a little bit further to try and avoid such problems in the future by adding a second CI job to Travis. While the default job will continue to default to all build options being set to `true`, the second job will set everything to `false`. This should in theory allow wlroots to detect such issues much earlier in the future.

I don't know whether that change is desired by the project, but I think it should be worthwhile. Building wlroots is not really a long-running task, so having the CI build the project twice shouldn't take too long while the benefit is kind of obvious to me.

Anyway. If you feel like the second CI job is not welcome, you can just drop the last commit :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,12 @@ compiler:
   - gcc
   - clang
 
+# Settings to try
+env:
+ matrix:
+  - OPTIONS="-Denable_libcap=true  -Denable_systemd=true  -Denable_elogind=false -Denable_xwayland=true"
+  - OPTIONS="-Denable_libcap=false -Denable_systemd=false -Denable_elogind=false -Denable_xwayland=false"
+
 arch:
   packages:
     - meson
@@ -19,7 +25,7 @@ arch:
     - xcb-util-image
     - libcap
   script:
-    - "meson build"
+    - "meson build $OPTIONS"
     - "ninja -C build"
 
 script:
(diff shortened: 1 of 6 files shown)


martinetd:
I like the idea of the extra travis build, but might need to fix the build without systemd first as well :)
Also, elogind is not found atm (and there's no arch package for it that I can see easily), so it's not built in main variant either; from my poor understanding it's also mutually exclusive with systemd anyway so we'd need 3 builds, one with systemd, one with elogind and one with neither (or if neither just isn't supported, make meson fail earlier in that case)

FWIW this isn't your fault but the report message with what's enabled or not at the end of meson.build doesn't take into account the dependencies having been found or not, so it's misleadingly saying `elogind: true` when it's not actually built.

pks-t:
True about elogind. I'd just set it to `false` in both lines and ignore it for now. Building the elogind package would be too much of a hassle, and it might even conflict with systemd on Ubuntu.

Funny that Travis fails without systemd, though, as I verified first that it _does_ in fact build on my system. Well, different compiler versions, I guess. I'll take a look

pks-t:
Found the issue. The systemd/elogind backends are always being included when the libraries are found on the system, disregarding whether their options are set or not. Fixed that and switched Travis to build both jobs with elogind disabled.

SirCmpwn:
Thanks!

PR #597, opened 2018-01-30
Fix some typos

agx:
Prefer initialize over initialise since used more often in the sources.

Debian's lintian tripped on these.


emersion:
Thanks!

Issue #589, opened 2018-01-26
Wrong output orientation on screenshot


fiete201:
Hi, 

my netbook has a display that was taken from a tablet and hence is portrait from the controller logic so needs to be rotated 90 degrees clockwise to be landscape and usable.

When I now connect my landscape monitor in my office and take a screenshot the internal display is rendered as portrait.

But other applications just recognize top and bottom quite well.
![wayland-screenshot](https://user-images.githubusercontent.com/1582793/35464339-a94dfa22-02f5-11e8-9a60-f487e114ab70.png)

Please just tell me if there are some details missing.

Best regards
Fritz

fiete201:
I found an easier way to "reproduce" just did a screenshot without external display connected.
![wayland-screenshot](https://user-images.githubusercontent.com/1582793/35612763-42ec0f94-066a-11e8-82e4-6f4d2883f364.png)

PR #598, opened 2018-01-31
Screenshot transform

Ongy:
Sets up correct handling for each possible output transformation in the screenshot example.

Fixes https://github.com/swaywm/wlroots/issues/589


emersion:
I get a shift between rotated outputs and normal ones:

wayland-screenshot

Ongy:
Good catch. Looking at the code there's probably a horizontal shift in other modes, since my inverse loop iterations are off by one.

PR #599, opened 2018-02-01
output: add destroy signal for wlr_output_cursor

### LINKED ISSUE #595: Freeing wlr_output_cursor when wlr_output is freed
Right now it's up to the user of wlroots to keep track of when a `wlr_output` is freed and to destroy the corresponding `wlr_output_cursor` associated with it.

This has me running into issues on [my rust bindings](https://github.com/swaywm/wlroots-rs) because it's not safe to "own" a `wlr_output_cursor` because at any time its state could become invalid. See [this gist](https://gist.github.com/Timidger/1d4f6c2e27bd0d0bb9a0cb53e7adfcbc) of me trying to capture the memory model and failing to make `OutputCursor` safe to use. Instead I opted to use lifetimes which are difficult to deal with for the user.

I have a dedicated pattern for dealing with things like this (e.g using `Rc` and `Weak` to make a structure invalid when wlroots frees something, which is how I let users own an `Output` via an `OutputHandle`), but maybe it would be useful to others if `wlr_output_cursor` state was `free`d when the `wlr_output` goes out of scope.

It comes down to how much does the compositor writer "own" `wlr_output_cursor`? Are they totally in charge of its memory even though part of it depends on something it can't control (e.g an output always being there)?

emersion:
Fixes #595

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -29,6 +29,10 @@ struct wlr_output_cursor {
 	struct wlr_surface *surface;
 	struct wl_listener surface_commit;
 	struct wl_listener surface_destroy;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
 };
 
 struct wlr_output_impl;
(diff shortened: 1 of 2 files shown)


SirCmpwn:
@Timidger thoughts please

Timidger:
Looks good to me, thanks!

PR #602, opened 2018-02-03
Refactor include/

emersion:
Move some files into subdirectories to match the source code organization.


SirCmpwn:
These are different by design.

emersion:
This is very confusing when editing a source code file and searching for its header. You search in the same directory, but you don't find anything.

SirCmpwn:
That doesn't bother me.

SirCmpwn:
The ergonomics of the pubic API are more important than the ergonomics of the internal development workflow.

emersion:
I think having both a folder and a file with the same name is bad ergonomics for the public API.

SirCmpwn:
I disagree. You want to interact with the backend?

```c
#include <wlr/backend.h>
```

You want to do stuff with a specific backend?

```c
#include <wlr/backend/wayland.h>
```

The latter case is less common so we save you the keystrokes in the former case. I agree that wlr_matrix is poorly placed, though, we should move it to types.

PR #611, opened 2018-02-07
Make pointer button release outside window still count down the button_count

DonOregano:
I am just starting contributions to wlroots, so I may have misunderstood everything. But hunting for bugs is always a good way of getting into a codebase.

When double-clicking a maximized window title, so that the windows size is restored and the mouse pointer ends up _outside_ the window it becomes impossible to move windows.
The reason is that the button_count variable is not counted down if the mouse button is released outside the window, so the button_count remains incremented even after the button is released.
This patch adds a call to wlr_seat_pointer_notify_button if the mouse button is released outside the window.

Is this fix the right way to fix the problem? Are there any undesirable side-effects? I don't quite understand the view and surface concepts.

### DIFF
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -261,7 +261,9 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 		}
 	}
 
-	if (view && surface) {
+	if ((view && surface) ||
+		(state == WLR_BUTTON_RELEASED &&
+		 seat->seat->pointer_state.button_count != 0)) {
 		if (!is_touch) {
 			wlr_seat_pointer_notify_button(seat->seat, time, button, state);
 		}


VincentVanlaer:
This is already fixed in #609, the fix I had is quite similar to this one.

acrisci:
I accept these fixes as a short term solution, but I wish we could do better with a stronger solution in the long term. I don't like being in a place where the seat is a state machine for the pointer. I wish calling the pointer notify functions was completely optional.

SirCmpwn:
Do you want to draw up a spec for improving this in a separate ticket?

acrisci:
> Do you want to draw up a spec for improving this in a separate ticket?

Yeah I have a feeling we're on track to find something nice in #613

PR #622, opened 2018-02-10
rootston: allow scale < 1 as well

martinetd:
Since we support fractional scale, having scale=0.5 can make sense
(and, as far as I can tell, works fine)


emersion:
Thanks!

PR #623, opened 2018-02-10
Meson option enhancements

martinetd:
The two commits are fairly independant, but under the same theme:
 - first commit makes the message printed at meson time to reflect what is actually built, rather than the option default values (been annoying me for a while to have elogind = true in that message when I don't have elogind...)
 - second commit implements some sort of auto/true/false mechanism, that by default tries to build whatever we have, but if someone explicitely sets meson -Denable_foo=true will fail when foo is not found.
This should not break either default (true was technically auto before) nor explicit configs as -Denable_foo=false is the same syntax.
Happy to remove this commit if it is not welcome but this usually makes sense to do.

### DIFF
--- a/backend/meson.build
+++ b/backend/meson.build
@@ -27,6 +27,17 @@ backend_files = files(
 	'x11/backend.c',
 )
 
+backend_deps = [
+	wayland_server,
+	egl,
+	gbm,
+	libinput,
+	wlr_render,
+	wlr_protos,
+	drm,
+	pixman,
+]
+
 if host_machine.system().startswith('freebsd')
 	backend_files += files('session/direct-freebsd.c')
 else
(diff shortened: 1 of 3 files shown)


ascent12:
Looks good to me. I was actually thinking about adding 'auto' ages ago, but never got around to doing it.

On a slightly related note, this will still link optional libraries, even if they were explicitly disabled. Meson 0.44.0 adds something called a disabler, which would make dealing with that easier.
https://github.com/mesonbuild/meson/blob/master/docs/markdown/Disabler.md

martinetd:
Hmm good point, I hadn't thought of that.
I don't think we need a disabler, just need to add the dependency to wlr_deps only within the switch -- I'll push a third patch for that after testing a bit.

martinetd:
I don't have any way to test elogind right now, but it wasn't in wlr_deps previously, only in the backend deps. This should be the same as before as far as linking goes.

PR #627, opened 2018-02-12
Remove wlr_backend.events.{output_remove,device_remove} 

### LINKED ISSUE #605: Remove input_remove and output_remove from wlr_backend
We can just use `wlr_output.destroy` and `wlr_input_device.destroy` instead.

One issue is that `wl_signal_emit` is not safe if you remove the next signal handler. I've submitted a patch to wayland (https://lists.freedesktop.org/archives/wayland-devel/2018-January/036711.html) but it got ignored. Maybe we could implement a safe `wlr_signal_emit` in util?

emersion:
Fixes #605


util/signal.c
```diff
@@ -0,0 +1,34 @@
+#include <wlr/util/signal.h>
```
SirCmpwn: Do you think this should be an internal header?

backend/drm/backend.c
```diff
@@ -11,7 +11,7 @@
 #include <wlr/interfaces/wlr_output.h>
 #include <wlr/types/wlr_list.h>
 #include <wlr/util/log.h>
-#include <wlr/util/signal.h>
+#include <util/signal.h>
```
SirCmpwn: single quotes for internal headers

SirCmpwn: Also, I'm kind of alphebetizing these lately

emersion: Okay, maybe we should unify headers. What about sorting but keeping external and internal includes groupped?

SirCmpwn: Yeah, that.

emersion:
I'll do that include style change in another PR, since it's somewhat unrelated to this one.

backend/drm/drm.c
```diff
@@ -18,6 +18,7 @@
 #include <wlr/backend/interface.h>
 #include <wlr/interfaces/wlr_output.h>
 #include <wlr/util/log.h>
+#include "util/signal.h"
```
SirCmpwn: Well you still need to group this with the other internal includes. Group quotes separate from <>

emersion: Double newline to separate include groups?

SirCmpwn: No no no no no please no

emersion: One last question: do I group stdlib/external/public/private, or external/public/private?

SirCmpwn: <> in one group, "" in another. Use <> for stuff that is (or will be, e.g. public wlr headers) installed in /usr/include

Ongy:
LGTM.

The one thing I am wondering about is the performance inpact.
Do we want to replace every emit? Or should e.g. cursor movement emits be on the old style?
It would imply other semantics on what the handler is allowed to do, but if the signal usually emits to a low number of listeners very often (cursor movements, graphic tablets pen movement) this may be interesting from a performance perspective.

SirCmpwn:
I don't think this presents significantly different performance compared to wl_signal_emit. They're almost the same process.

emersion:
Yeah, the only difference is that it's updating a few pointers. Also libwayland's private, internal `wl_priv_signal_emit` has similar overhead and is used everywhere in libwayland's code.

PR #629, opened 2018-02-13
Wrap wl_resource_get_user_data into safer helper functions

### LINKED ISSUE #628: Wrap wl_resource_get_user_data
Currently the whole codebase contains a lot of `wl_resource_get_user_data`. This is error-prone since type safety isn't ensured.

We could instead wrap `wl_resource_get_user_data` with type-safe helper functions (e.g. `pixman_region32_t *wlr_resource_get_region(struct wl_resource *resource)` for `wlr_region`), and possibly assert `wl_resource_instance_of` in the helper.

emersion:
This ensures we're not incorrectly casting a resource.

Fixes #628

### DIFF
--- a/types/wlr_compositor.c
+++ b/types/wlr_compositor.c
@@ -7,13 +7,21 @@
 #include <wlr/util/log.h>
 #include "util/signal.h"
 
+static const struct wl_compositor_interface wl_compositor_impl;
+
+static struct wlr_compositor *compositor_from_resource(struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &wl_compositor_interface,
+		&wl_compositor_impl));
+	return wl_resource_get_user_data(resource);
+}
+
 static void destroy_surface_listener(struct wl_listener *listener, void *data) {
 	wl_list_remove(wl_resource_get_link(data));
 }
 
 static void wl_compositor_create_surface(struct wl_client *client,
 		struct wl_resource *resource, uint32_t id) {
-	struct wlr_compositor *compositor = wl_resource_get_user_data(resource);
+	struct wlr_compositor *compositor = compositor_from_resource(resource);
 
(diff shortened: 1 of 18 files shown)


types/wlr_compositor.c
```diff
@@ -44,13 +52,13 @@ static void wl_compositor_create_region(struct wl_client *client,
...
 
-struct wl_compositor_interface wl_compositor_impl = {
+static const struct wl_compositor_interface wl_compositor_impl = {
 	.create_surface = wl_compositor_create_surface,
 	.create_region = wl_compositor_create_region
 };
 
 static void wl_compositor_destroy(struct wl_resource *resource) {
-	struct wlr_compositor *compositor = wl_resource_get_user_data(resource);
+	struct wlr_compositor *compositor = compositor_from_resource(resource);
```
SirCmpwn: Let's be consistent: wlr_compositor_from_resource.

emersion: Static fonctions drop the wlr_ prefix.

SirCmpwn: -1

SirCmpwn: fwiw I see an argument for exporting these functions to the public API

emersion: >-1

You mean static functions should have the wlr_ prefix? Not dropping the prefix for this particular function *is* inconsistent.

>fwiw I see an argument for exporting these functions to the public API

A few are exported (wlr_output, wlr_region, wlr_surface). We could export others, but not sure it makes sense (wlr_screenshooter, wlr_decoration_manager...).

SirCmpwn: >You mean static functions should have the wlr_ prefix? Not dropping the prefix for this particular function is inconsistent.

I mean *these* static functions should have the wlr_ prefix. But maybe not. Let me think about it.

PR #630, opened 2018-02-14
xdg-shell stable implementation

### LINKED ISSUE #557: Add stable xdg-shell support
Source: https://github.com/wayland-project/wayland-protocols/blob/master/stable/xdg-shell/xdg-shell.xml

Changelog: https://cgit.freedesktop.org/wayland/wayland-protocols/log/stable/xdg-shell/xdg-shell.xml

We'll probably have to copy the whole code. If the changes are minor maybe we can share some code?

It's still a bit early, we may want to wait a little.

Test clients:

* https://github.com/glfw/glfw/pull/1152
* https://github.com/Enlightenment/efl/commit/ffa276fce3b812372a063431903fa0569d71fcb2

No idea if GTK+ implements this already. Mutter does.

martinetd:
Roadmap:
* [x] Implement xdg-shell stable
* ~~share as much code as possible with xdg-shell-v6~~

Test plan:
 * Run some glfw examples, check with WAYLAND_DEBUG that they use the new interface (xdg_wm_base etc) and that things appear to work™
 * Would like to test popups as well somehow, not aware of any way to test right now
 * Check I didn't break xdg-shell-v6 as well

Fixes #557


types/wlr_xdg_shell.c
```diff
@@ -0,0 +1,1457 @@
...
+	positioner->anchor_rect.width = width;
+	positioner->anchor_rect.height = height;
+}
+
+static void xdg_positioner_protocol_set_anchor(struct wl_client *client,
+		struct wl_resource *resource, uint32_t anchor) {
+	struct wlr_xdg_positioner *positioner =
+		wl_resource_get_user_data(resource);
+
+	/* post error if anchor > XDG_POSITIONER_ANCHOR_BOTTOM_RIGHT ? */
```
martinetd: Should we check invalid value here?
Given it's an enum without a last enum value to use as marker, what's the style-wise correct way of checking bound?

emersion: Yes, checking would be cool (please backport to xdg-shell unstable v6 as well).

We generally define a mask of valid bits (see https://github.com/swaywm/wlroots/blob/master/types/wlr_data_device.c#L13).

emersion: Ah, if it isn't a bitmask anymore, it makes more sense to check if `anchor > XDG_POSITIONER_ANCHOR_BOTTOM_RIGHT`.

emersion:
>I assume it's ok to break the xdg-shell-v6 interface at this point?

What do you mean?

>can we share the same types between xdg-shell-v6 and xdg-shell?

I'm not sure we want to share anything. That would makes the code a lot more difficult to read, and dropping xdg-shell unstable v6 will be harder. We could maybe just copy-paste and make sure fixes in one shell are backported in the other while xdg-shell v6 is still alive.

>FWIW, the only differences I have seen in the interface are:

You can have a changelog here: https://cgit.freedesktop.org/wayland/wayland-protocols/log/stable/xdg-shell/xdg-shell.xml

types/wlr_xdg_shell.c
martinetd: xdg-shell unstable-v6 had the check, but the logic changed so I couldn't keep it as is.

I don't 100% like the `anchor > XDG_POSITIONER_ANCHOR_BOTTOM_RIGHT` construct for enum because if an extra enum value is added at the end it is hard to locate these in the code (that's why my enums tend to have a final _COUNT or similar at the end), but I guess that's a non-concern for a stable xdg-shell interface.
I'll add the check as is.

SirCmpwn:
>I assume it's ok to break the xdg-shell-v6 interface at this point?

No

>can we share the same types between xdg-shell-v6 and xdg-shell?

Probably not?

>I'd like to share many static functions as well, so my current thought is to plainly merge types/wlr_xdg_shell_v6.c into types/wlr_xdg_shell.c and have a single wlr type handle both. Any qualm over that?

Depends on how it shakes out

martinetd:
(how do you get github to quote a previous message to reply? I should start replying in my mail client...)

> breaking xdg-shell-v6 interface

I was thinking of removing the _v6 types to be able to share the code.
The structs are 100% identical except for their names, but it would break code using it to rename them

> re-using functions from wlr_xdg_shell.c in wlr_xdg_shell_v6.c

Well, that's pointless if we can't share the same types, so let's stay as things are. I can agree keeping things separate will be easier to remove xdg_shell_v6 at some point, even if it duplicates a lot of code -- we'll just need to be careful to backport all the fixes to both codebases.

That means this PR is ready for review/testing as well, I'll just need to add the check mentionned in the other comment for anchor/gravity.

SirCmpwn:
>(how do you get github to quote a previous message to reply? I should start replying in my mail client...)

I just copypasta and do it manually, using the primary selection

>I was thinking of removing the _v6 types to be able to share the code.
>The structs are 100% identical except for their names, but it would break code using it to rename them

I'm of the opinon that merging the shells is probably a bad idea, but I'm open to taking a look at the other approach. Eventually we are likely to remove support for v6, so it'd be easier to remove it wholesale if there aren't too many interactions with the stable shell code. But I can see the argument for the other approach, if you want to try that let's see how it works out.

martinetd:
> I'm of the opinon that merging the shells is probably a bad idea, but I'm open to taking a look at the other approach. Eventually we are likely to remove support for v6, so it'd be easier to remove it wholesale if there aren't too many interactions with the stable shell code. But I can see the argument for the other approach, if you want to try that let's see how it works out.

I've had another look, I don't think it's impossible but it will definitely not be pretty - there are enums/defines from the protocol sprinkled all around. Let's keep them separate.

protocol/meson.build
```diff
@@ -31,6 +32,7 @@ protocols = [
 
 client_protocols = [
 	[wl_protocol_dir, 'unstable/xdg-shell/xdg-shell-unstable-v6.xml'],
+	[wl_protocol_dir, 'stable/xdg-shell/xdg-shell.xml'],
```
martinetd: I've added it as client_protocols as well because I just copied unstable-v6 everywhere, but this isn't used -- should I remove that for now?

martinetd: No answer, so remove this and other unused client protocols. These aren't installed either so cleaned up protocols included with <> incorrectly.

martinetd:
Just proof-read everything and retested, I believe this is ready.

protocol/meson.build
emersion: Yeah, that's a good idea.

types/wlr_xdg_shell.c
```diff
@@ -252,10 +252,10 @@ static void xdg_positioner_protocol_set_anchor_rect(struct wl_client *client,
 	struct wlr_xdg_positioner *positioner =
 		wl_resource_get_user_data(resource);
 
-	if (width < 1 || height < 1) {
+	if (width < 0 || height < 0) {
 		wl_resource_post_error(resource,
 			XDG_POSITIONER_ERROR_INVALID_INPUT,
-			"width and height must be positives and non-zero");
+			"width and height must be positives");
```
emersion: Typo: "must be positive"

martinetd: fixed

PR #632, opened 2018-02-14
add wlr_output_layout_adjacent_output

acrisci:
This function is for determining the adjacent output to another output, like we have to do with sway in some places.

The output extents must be entirely within the direction from the corresponding border of the reference output to be considered. This function then picks the closest output to the reference point.

TODO:

* [x] test the function


acrisci:
Here is an important edge case:

```
+-----------------+
|                 |
|                 |
|       a         |
|                 |
|                 |
+-----------------+

            +-------------------+
            |                   |
            |                   |
            |        b          |
            |                   |
            |                   |
            +-------------------+
```

In this case, `a` is *not* left of `b` because it contains a region that is not to the left of the left border of `b`.

SirCmpwn:
Also test the case of mirrored and partially mirrored outputs

include/wlr/util/direction.h
```diff
@@ -0,0 +1,11 @@
+#ifndef WLR_UTIL_DIRECTION_H
+#define WLR_UTIL_DIRECTION_H
+
+enum wlr_direction {
+	WLR_DIRECTION_UP = 0,
+	WLR_DIRECTION_DOWN = 1,
+	WLR_DIRECTION_LEFT = 2,
+	WLR_DIRECTION_RIGHT = 4,
+};
```
SirCmpwn: I don't think this merits its own header

types/wlr_output_layout.c
```diff
@@ -421,3 +421,53 @@ struct wlr_output *wlr_output_layout_get_center_output(
 
 	return wlr_output_layout_output_at(layout, dest_x, dest_y);
 }
+
+
+struct wlr_output *wlr_output_layout_adjacent_output(
+		struct wlr_output_layout *layout, enum wlr_direction direction,
+		struct wlr_output *reference, double ref_x, double ref_y) {
+	// XXX should we allow reference to be NULL?
```
SirCmpwn: Why?

include/wlr/util/direction.h
acrisci: Where do you want it?

types/wlr_output_layout.c
acrisci: The only reason I can think of is to make it consistent with the rest of the api. The rule generally is that if reference is not passed, then it acts on the whole layout, but there's no concrete use case for this yet.

include/wlr/util/direction.h
SirCmpwn: Can just live in wlr_output_layout.h imo

types/wlr_output_layout.c
SirCmpwn: I don't know how that could work, probably best not to.

acrisci:
This is now used in [sway#1585](https://github.com/swaywm/sway/pull/1585).

types/wlr_output_layout.c
SirCmpwn: Should drop the comment, leave the assert

emersion: Can't we automatically get the `reference` output from `ref_x` and `ref_y`?

acrisci: We could, but we need to start being more careful about supporting the overlapping output case. There are a few functions that currently don't support this well that need to be revisited.

Issue #624, opened 2018-02-10
Implement xwayland-keyboard-grab-unstable-v1


Ongy:
This protocol allows things like `dmenu_run` to work properly (i.e. steal keybaord focus) while other windows (e.g. dunst) that don't require keyboard focus don't steal it.

https://github.com/wayland-project/wayland-protocols/blob/master/unstable/xwayland-keyboard-grab/xwayland-keyboard-grab-unstable-v1.xml

emersion:
We found out this is unrelated to `dmenu` and we've chosen not to provide an implementation for it.

ammen99:
I think it may be time to reconsider implementing this. My tests with GNOME on Wayland show that this protocol IS used when for ex. VirtualBox is running.

subdiff:
Indeed VirtualBox is one of the interesting use cases for this extension. But it is also needed for certain Wayland native clients (for example KWin's nested sessions or qemu VMs). There is also the keyboard_shortcuts_inhibit_unstable_v1 protocol, which does apparently the same for arbitrary Wayland clients, but is not supported by XWayland. I'm not sure why there are two different protocols, one for XWayland alone.

Ongy:
Do you have a source on this being used in VMs native?
Iirc this specifically says it should only be exposed to xwayland and other clients shouldn't get access to the protocol.

Thus getting them together isn't exactly anything to plan for. One is legacy support, the other a more generic method going forward

Ongy:
I should get into the habit of reading a full comment again..
The shortcut-inhibit and input-inhibitor are both for native clients.
Shortcut inhibit+pointer constraints might be better for VMs than input-inhibitor

emersion:
input-inhibitor is a privileged protocol for lockscreens.

helloworld1:
I found it definitely very annoying without any keyboard grab support. I mainly uses tigervnc's vncviewer. But I also tried other vnc clients. Non can grab keyboard in sway. I am unable to run i3 or sway remote desktops because local will take the input.

helloworld1:
Is there any follow-up on this issue? It has been 5 months without progress.

PR #635, opened 2018-02-15
[WiP] Xwayland keyboard grab

Ongy:
This is an implementation for the protocol mentioned in https://github.com/swaywm/wlroots/issues/624
While implementing it, I noticed I had a wrong idea of the exact scope. This does not work for dmenu, but `pinentry-gtk-2` uses it for secure input.

Things still left to be done:
- [ ] Cleanup the name of things/squash commits
- [x] Restrict to xwayland? (rootston doesn't restrict access to protocols at all)

Testplan:
Get an xwayland build from git (at least arch package is too old).
Start up rootston with this patch, and open `pinentry-gtk-2` inside it.
**EDIT**: Type `GETPIN` into the pinentry shell, to get the window that uses passive grabs.
Try to move focus to another window to input there.


include/wlr/types/wlr_passive_grab.h
```diff
@@ -0,0 +1,37 @@
...
+	struct wl_list clients; // wlr_passive_grab_manager::link;
+	
+	struct {
+		struct wl_signal new_grab;
+	} events;
+
+	struct wl_listener display_destroy;
+};
+
+struct wlr_passive_grab_v1 *wlr_passive_grab_create_v1(struct wl_display *display);
```
emersion: I think we want this to be named `wlr_passive_grab_v1_create`.

Ongy:
I implemented the restriction to xwayland in the `_bind` function of the protocol.

The names should make sense now. I would like to squash and clean up the commits before a merge though (they don't fit the guidelines :))

rootston/desktop.c
```diff
@@ -609,6 +613,51 @@ static void handle_layout_change(struct wl_listener *listener, void *data) {
...
+			calloc(1, sizeof(struct roots_xwayland_keyboard_grab_v1));
+		if (!roots_grab) {
+			return;
+		}
+
+		roots_grab->grab = grab;
+		wl_signal_add(&grab->events.destroy, &roots_grab->destroy_listener);
+		roots_grab->destroy_listener.notify = handle_xwayland_keyboard_grab_destroy;
+
+		wl_list_insert(&seat->xwayland_keyboard_grabs, &roots_grab->link);
```
emersion: We probably don't want compositors to change the list manually. I'm also not sure if adding fields like `xwayland_keyboard_grabs` to the seat for extensions is fine.

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
...
+	.destroy = xwayland_keyboard_grab_grab_destroy,
+};
+
+static void xwayland_keyboard_grab_grab_keyboard(struct wl_client *client,
+		struct wl_resource *resource, uint32_t id,
+		struct wl_resource *surface, struct wl_resource *seat) {
+	assert(client);
+	assert(resource);
+	assert(surface);
+	assert(seat);
```
emersion: All these `assert` are useless, Wayland ensures these aren't NULL (there's a protocol property to allow null values).

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
...
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+};
+
+static void xwayland_keyboard_grab_grab_handle_destroy(struct wl_resource *resource) {
+	assert(resource);
+	struct wlr_xwayland_keyboard_grab_v1_grab *grab =
+		wl_resource_get_user_data(resource);
```
emersion: We now use `wl_resource_get_user_data` wrappers to ensure type safety. Have a look at other protocol implementations to see how to do it.

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
...
+		wl_client_post_no_memory(client);
+		free(grab);
+		return;
+	}
+	grab->resource = grab_resource;
+	wl_signal_init(&grab->events.destroy);
+
+	struct wlr_seat_client *seat_client = wl_resource_get_user_data(seat);
+	grab->seat = seat_client->seat;
+	wlr_log(L_DEBUG, "Seat class: %s", wl_resource_get_class(seat));
```
emersion: This debug log can be removed.

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
...
+	xwayland_keyboard_grab->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &xwayland_keyboard_grab->display_destroy);
+	xwayland_keyboard_grab->xwayland = xwayland;
+
+	return xwayland_keyboard_grab;
+}
+
+void wlr_xwayland_keyboard_grab_v1_destroy(struct wlr_xwayland_keyboard_grab_v1 *xwayland_keyboard_grab) {
+	struct wlr_xwayland_keyboard_grab_v1_manager *manager;
+	struct wlr_xwayland_keyboard_grab_v1_manager *tmp;
```
emersion: Style nitpick: both variables can be defined at the same time.

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
...
+static void xwayland_keyboard_grab_v1_bind(struct wl_client *client, void *data,
+		uint32_t version, uint32_t id) {
+	assert(data);
+	assert(client);
+	struct wlr_xwayland_keyboard_grab_v1 *xwayland_keyboard_grab = data;
+
+	if (client != xwayland_keyboard_grab->xwayland->client) {
+		wlr_log(L_DEBUG, "Denying xwayland_keyboard_grab access for client");
+		// TODO: Can we post a better error here?
+		wl_client_post_no_memory(client);
```
emersion: It's probably better not to post an ENOMEM error and just do nothing.

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
...
+#include <assert.h>
+#include <wayland-server.h>
+#include <wayland-util.h>
+#include <wlr/util/log.h>
+#include <wlr/types/wlr_surface.h>
+#include <wlr/types/wlr_xwayland_keyboard_grab_v1.h>
+
+#include <xwayland-keyboard-grab-unstable-v1-protocol.h>
+
+struct wlr_xwayland_keyboard_grab_v1_manager {
```
emersion: Why isn't this struct in the header?

SirCmpwn:
We're not gonna do this one.

types/wlr_xwayland_keyboard_grab_v1.c
```diff
@@ -0,0 +1,200 @@
+#include <stdlib.h>
+#include <assert.h>
+#include <wayland-server.h>
+#include <wayland-util.h>
+#include <wlr/util/log.h>
+#include <wlr/types/wlr_surface.h>
+#include <wlr/types/wlr_xwayland_keyboard_grab_v1.h>
+
+#include <xwayland-keyboard-grab-unstable-v1-protocol.h>
```
emersion: Protocol header files are local includes, so they should use `"`. Also please don't add a newline between includes, and sort external and internal includes.

include/wlr/types/wlr_xwayland_keyboard_grab_v1.h
```diff
@@ -0,0 +1,41 @@
...
+	
+	struct {
+		struct wl_signal new_grab;
+	} events;
+
+	struct wl_listener display_destroy;
+};
+
+struct wlr_xwayland_keyboard_grab_v1 *wlr_xwayland_keyboard_grab_v1_create(
+	struct wl_display *display, struct wlr_xwayland *xwayland);
```
emersion: This is weird. I'd expect `wlr_xwayland_keyboard_grab_manager_v1_create` instead.

include/wlr/types/wlr_xwayland_keyboard_grab_v1.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_surface.h>
+#include <wayland-util.h>
+#include <wayland-server.h>
+
+#include <wlr/xwayland.h>
+
+
+struct wlr_xwayland_keyboard_grab_v1_grab {
```
emersion: This should be `wlr_xwayland_keyboard_grab_v1` instead.

PR #638, opened 2018-02-16
[DO NOT MERGE] Add xdg-toplevel-decoration support

emersion:
This adds support for xdg-toplevel-decoration in rootston. The preferred mode is "client". rootston will accept the client's mode requests.

A new example client is added. When launched, it will be yellow if using server-side decorations and green if using client-side decorations. The client's initial preferred mode can be given as an argument (either "client" or "server"). The client's preferred mode can be toggled by clicking on its surface.

Test plan: try the example


BurningTreeC:
I like copy-pasta!

emersion:
Just look at the last 3 commits f you don't want to review the xdg-shell changes.

emersion:
This PR is ready for review.

examples/toplevel-decoration.c
```diff
@@ -0,0 +1,189 @@
...
+	wl_display_roundtrip(display);
+
+	draw();
+
+	while (wl_display_dispatch(display) != -1) {
+		// No-op
+	}
+
+	return EXIT_SUCCESS;
+}
```
SirCmpwn: Would be nice if the example could be made to toggle CSD at runtime

emersion: Just click on the window, it will switch the mode.

Sunderland93:
This is a replace of KDE's Server-Decoration protocol, right?

SirCmpwn:
It is, yes, developed in cooperation with KDE and Mir.

Sunderland93:
@SirCmpwn will it be added to wayland-protocols?

SirCmpwn:
We have sent it in for inclusion, yeah.

emersion:
Superseded by #1053

PR #640, opened 2018-02-16
Remove WL_HIDE_DEPRECATED

emersion:
From the wayland-devel mailing list: _[RFC wayland] wayland-server: Finally remove deprecated struct wl_buffer definition_

We don't use it anyway, so we should remove WL_HIDE_DEPRECATED from the build options.


emersion:
Damn, I misunderstood what this WL_HIDE_DEPRECATED is for. It's actually hiding struct definitions, i was thinking it was exposing us more structs.

PR #641, opened 2018-02-17
Add idle-inhibit implementation

Ongy:
This adds an implementation of the `idle-inhibit-v1` protocol and an example client.

The client is very simple and skips a little bit of error checking (whether the seat actually has a pointer), but works to trigger the code paths in the server.

The implementation in `rootston` does nothing but create the manager.
Since rootston doesn't have any kind of idling action, I'm not sure what else we could do here, ideas?

### DIFF
--- a/examples/idle-inhibit.c
+++ b/examples/idle-inhibit.c
@@ -96,8 +96,15 @@ static void xdg_toplevel_handle_configure(void *data,
 	height = h;
 }
 
+static void xdg_toplevel_handle_close(void *data,
+		struct xdg_toplevel *xdg_toplevel) {
+	exit(EXIT_SUCCESS);
+}
+
+
 static const struct xdg_toplevel_listener xdg_toplevel_listener = {
 	.configure = xdg_toplevel_handle_configure,
+	.close = xdg_toplevel_handle_close,
 };
 
 // static const struct zxdg_toplevel_decoration_v1_listener decoration_listener = {
--- a/examples/idle-inhibit.c
+++ b/examples/idle-inhibit.c
@@ -1,0 +1,203 @@
+#include <GLES2/gl2.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wayland-client.h>
+#include <wayland-egl.h>
+#include <wlr/render/egl.h>
+#include "xdg-shell-client-protocol.h"
+#include "idle-inhibit-unstable-v1-client-protocol.h"
+
+#include <linux/input-event-codes.h>
+
+/**
+ * Usage: toplevel-decoration [mode]
+ * Creates a xdg-toplevel supporting decoration negotiation. If `mode` is
+ * specified, the client will prefer this decoration mode.
+ */
+
+static int width = 500, height = 300;
+
+static struct wl_compositor *compositor = NULL;
+static struct wl_seat *seat = NULL;
+static struct xdg_wm_base *wm_base = NULL;
+static struct zwp_idle_inhibit_manager_v1 *idle_inhibit_manager = NULL;
+static struct zwp_idle_inhibitor_v1 *idle_inhibitor = NULL;
+
+struct wlr_egl egl;
+struct wl_egl_window *egl_window;
+struct wlr_egl_surface *egl_surface;
+
+static void draw(void);
+
+static void pointer_handle_button(void *data, struct wl_pointer *pointer, uint32_t serial,
+		uint32_t time, uint32_t button, uint32_t state_w) {
+	struct wl_surface *surface = data;
+
+	if (button == BTN_LEFT && state_w == WL_POINTER_BUTTON_STATE_PRESSED) {
+		if (idle_inhibitor) {
+			zwp_idle_inhibitor_v1_destroy(idle_inhibitor);
+			idle_inhibitor = NULL;
+		} else {
+			idle_inhibitor =
+				zwp_idle_inhibit_manager_v1_create_inhibitor(
+					idle_inhibit_manager,
+					surface);
+		}
+	}
+
+	draw();
+}
+
+static void noop () {}
+
+static const struct wl_pointer_listener pointer_listener = {
+	.enter = noop,
+	.leave = noop,
+	.motion = noop,
+	.button = pointer_handle_button,
+	.axis = noop,
+	.frame = noop,
+	.axis_source = noop,
+	.axis_stop = noop,
+	.axis_discrete = noop,
+};
+
+static void draw(void) {
+	eglMakeCurrent(egl.display, egl_surface, egl_surface, egl.context);
+
+	float color[] = {1.0, 1.0, 0.0, 1.0};
+	if (idle_inhibitor) {
+		color[0] = 0.0;
+	}
+
+	glViewport(0, 0, width, height);
+	glClearColor(color[0], color[1], color[2], 1.0);
+	glClear(GL_COLOR_BUFFER_BIT);
+
+	eglSwapBuffers(egl.display, egl_surface);
+}
+
+static void xdg_surface_handle_configure(void *data,
+		struct xdg_surface *xdg_surface, uint32_t serial) {
+	xdg_surface_ack_configure(xdg_surface, serial);
+	wl_egl_window_resize(egl_window, width, height, 0, 0);
+	draw();
+}
+
+static const struct xdg_surface_listener xdg_surface_listener = {
+	.configure = xdg_surface_handle_configure,
+};
+
+static void xdg_toplevel_handle_configure(void *data,
+		struct xdg_toplevel *xdg_toplevel, int32_t w, int32_t h,
+		struct wl_array *states) {
+	width = w;
+	height = h;
+}
+
+static const struct xdg_toplevel_listener xdg_toplevel_listener = {
+	.configure = xdg_toplevel_handle_configure,
+};
+
+// static const struct zxdg_toplevel_decoration_v1_listener decoration_listener = {
+// 	.preferred_mode = decoration_handle_preferred_mode,
+// 	.configure = decoration_handle_configure,
+// };
+
+static void handle_global(void *data, struct wl_registry *registry,
+		uint32_t name, const char *interface, uint32_t version) {
+	if (strcmp(interface, "wl_compositor") == 0) {
+		compositor = wl_registry_bind(registry, name, &wl_compositor_interface,
+			1);
+	} else if (strcmp(interface, xdg_wm_base_interface.name) == 0) {
+		wm_base = wl_registry_bind(registry, name, &xdg_wm_base_interface, 1);
+	} else if (strcmp(interface, zwp_idle_inhibit_manager_v1_interface.name) == 0) {
+		idle_inhibit_manager = wl_registry_bind(registry, name,
+			&zwp_idle_inhibit_manager_v1_interface, 1);
+	} else if (strcmp(interface, wl_seat_interface.name) == 0) {
+		seat = wl_registry_bind(registry, name, &wl_seat_interface, version);
+	}
+}
+
+static void handle_global_remove(void *data, struct wl_registry *registry,
+		uint32_t name) {
+	// TODO
+}
+
+static const struct wl_registry_listener registry_listener = {
+	.global = handle_global,
+	.global_remove = handle_global_remove,
+};
+
+int main(int argc, char **argv) {
+//	if (argc == 2) {
+//		char *mode = argv[1];
+//		if (strcmp(mode, "client") == 0) {
+//			decoration_mode = ZXDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT;
+//		} else if (strcmp(mode, "server") == 0) {
+//			decoration_mode = ZXDG_TOPLEVEL_DECORATION_V1_MODE_SERVER;
+//		} else {
+//			fprintf(stderr, "Invalid decoration mode\n");
+//			return EXIT_FAILURE;
+//		}
+//	}
+
+	struct wl_display *display = wl_display_connect(NULL);
+	if (display == NULL) {
+		fprintf(stderr, "Failed to create display\n");
+		return EXIT_FAILURE;
+	}
+
+	struct wl_registry *registry = wl_display_get_registry(display);
+	wl_registry_add_listener(registry, &registry_listener, NULL);
+	wl_display_dispatch(display);
+	wl_display_roundtrip(display);
+
+	if (compositor == NULL) {
+		fprintf(stderr, "wl-compositor not available\n");
+		return EXIT_FAILURE;
+	}
+	if (wm_base == NULL) {
+		fprintf(stderr, "xdg-shell not available\n");
+		return EXIT_FAILURE;
+	}
+	if (idle_inhibit_manager == NULL) {
+		fprintf(stderr, "idle-inhibit not available\n");
+		return EXIT_FAILURE;
+	}
+
+	wlr_egl_init(&egl, EGL_PLATFORM_WAYLAND_EXT, display, NULL,
+		WL_SHM_FORMAT_ARGB8888);
+
+	struct wl_surface *surface = wl_compositor_create_surface(compositor);
+	struct xdg_surface *xdg_surface =
+		xdg_wm_base_get_xdg_surface(wm_base, surface);
+	struct xdg_toplevel *xdg_toplevel = xdg_surface_get_toplevel(xdg_surface);
+
+	idle_inhibitor =
+		zwp_idle_inhibit_manager_v1_create_inhibitor(idle_inhibit_manager,
+		surface);
+
+	struct wl_pointer *pointer = wl_seat_get_pointer(seat);
+	wl_pointer_add_listener(pointer, &pointer_listener, surface);
+
+
+	xdg_surface_add_listener(xdg_surface, &xdg_surface_listener, NULL);
+	xdg_toplevel_add_listener(xdg_toplevel, &xdg_toplevel_listener, NULL);
+
+	wl_surface_commit(surface);
+
+	egl_window = wl_egl_window_create(surface, width, height);
+	egl_surface = wlr_egl_create_surface(&egl, egl_window);
+
+	wl_display_roundtrip(display);
+
+	draw();
+
+	while (wl_display_dispatch(display) != -1) {
+		// No-op
+	}
+
+	return EXIT_SUCCESS;
+}
--- a/examples/idle-inhibit.c
+++ b/examples/idle-inhibit.c
@@ -68,41 +68,41 @@ static void pointer_handle_button(void *data, struct wl_pointer *pointer, uint32
 static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 		uint32_t serial, struct wl_surface *surface,
 		wl_fixed_t surface_x, wl_fixed_t surface_y) {
-	/* NOOP: ignore event */
+	// This space intentionally left blank
 }
 
 static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 		uint32_t serial, struct wl_surface *surface) {
-	/* NOOP: ignore event */
+	// This space intentionally left blank
 }
 
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
-	/* NOOP: ignore event */
+	// This space intentionally left blank
 }
 
 static void pointer_handle_axis(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, uint32_t axis, wl_fixed_t value) {
-	/* NOOP: ignore event */
+	// This space intentionally left blank
 }
 
 static void pointer_handle_frame(void *data, struct wl_pointer *wl_pointer) {
-	/* NOOP: ignore event */
+	// This space intentionally left blank
 }
 
 static void pointer_handle_axis_source(void *data,
 		struct wl_pointer *wl_pointer, uint32_t axis_source) {
-	/* NOOP: ignore event */
+	// This space intentionally left blank
 }
 
 static void pointer_handle_axis_stop(void *data,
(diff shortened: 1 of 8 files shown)


agx:
Maybe turning off all displays after n minutes when no idle inhibitor is present?

raazvvann:
It will be nice to have integration with wlr_idle to inhibit idle events while there exist at least an inhibitor for that seat, but this will gonna be tricky.
And I get a SIGABRT for the client when closing it.

Ongy:
@raazvvann inhibitors are bound to surfaces.
The idle protocol is bound to seats. Interaction between those is a bit wonky.

It's a bit of a scope/style decision on wlroots for me.
In a "real" compositor I would break spec slightly and block idle when at least one idle-inhibitor inhibits, maybe filtered to there being a single seat.
On a compositor that's intended to be a reference, I'd prefer to stick closer to actual specs

raazvvann:
The last commit fixed the crash.

About the interaction between idle and idle-inhibit, I disagree that might be out of wlroots scope, because in a "real" compositor, based on wlroots, this will be imposible without some modification in wlr_idle and wlr_idle_inhibit to support this. So it's better to be done in wlroots, rather that to be done in compositors.
Anyway this can be consider out of this PR scope, and be done later.

Ongy:
I'd say that should be fixed purely in the idle implementation.
If the event is routed through the compositor, it should be easy enough to intercept it, if an idle inhibitor exists.

I don't think the logic of when and how the inhibitor interacts with the idle protocol should be in wlroots.

examples/idle-inhibit.c
```diff
@@ -0,0 +1,210 @@
...
+#include "xdg-shell-client-protocol.h"
+#include "idle-inhibit-unstable-v1-client-protocol.h"
+
+#include <linux/input-event-codes.h>
+
+/**
+ * Usage: toplevel-decoration [mode]
+ * Creates a xdg-toplevel supporting decoration negotiation. If `mode` is
+ * specified, the client will prefer this decoration mode.
+ */
```
emersion: You may want to change this comment.

examples/idle-inhibit.c
```diff
@@ -0,0 +1,210 @@
...
+				zwp_idle_inhibit_manager_v1_create_inhibitor(
+					idle_inhibit_manager,
+					surface);
+		}
+	}
+
+	draw();
+}
+
+static void noop () {}
```
emersion: Nitpick: extra space between the name and the opening parenthesis.

examples/idle-inhibit.c
```diff
@@ -0,0 +1,210 @@
...
+
+static const struct xdg_toplevel_listener xdg_toplevel_listener = {
+	.configure = xdg_toplevel_handle_configure,
+	.close = xdg_toplevel_handle_close,
+};
+
+// static const struct zxdg_toplevel_decoration_v1_listener decoration_listener = {
+// 	.preferred_mode = decoration_handle_preferred_mode,
+// 	.configure = decoration_handle_configure,
+// };
```
emersion: This can be cleaned up.

examples/idle-inhibit.c
```diff
@@ -0,0 +1,210 @@
...
+//		char *mode = argv[1];
+//		if (strcmp(mode, "client") == 0) {
+//			decoration_mode = ZXDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT;
+//		} else if (strcmp(mode, "server") == 0) {
+//			decoration_mode = ZXDG_TOPLEVEL_DECORATION_V1_MODE_SERVER;
+//		} else {
+//			fprintf(stderr, "Invalid decoration mode\n");
+//			return EXIT_FAILURE;
+//		}
+//	}
```
emersion: Same here.

include/wlr/types/wlr_idle_inhibit_v1.h
```diff
@@ -0,0 +1,27 @@
+#ifndef WLR_TYPES_WLR_IDLE_INHIBIT_V1_H
+#define WLR_TYPES_WLR_IDLE_INHIBIT_V1_H
+
+#include <wayland-server.h>
+
+struct wlr_idle_inhibit_v1 {
```
emersion: This represents a idle-inhibit-manager global, so it should be named `wlr_idle_inhibit_manager_v1`.

include/wlr/types/wlr_idle_inhibit_v1.h
```diff
@@ -0,0 +1,27 @@
...
+	struct wl_list clients;
+	struct wl_global *global;
+	
+	struct wl_listener display_destroy;
+	struct wl_signal new_inhibitor;
```
emersion: Signals should be in a `struct { … } events;`.

include/wlr/types/wlr_idle_inhibit_v1.h
```diff
@@ -0,0 +1,27 @@
...
+};
+
+struct wlr_idle_inhibit_inhibitor_v1 {
+	struct wlr_surface *surface;
+	struct wl_resource *resource;
+	struct wl_listener surface_destroy;
+
+	struct wl_list link; // wlr_idle_inhibit_manager::inhibitors;
+
+	struct wl_signal destroy;
```
emersion: Same here.

protocol/meson.build
```diff
@@ -32,8 +33,13 @@ protocols = [
 
 client_protocols = [
 	[wl_protocol_dir, 'unstable/xdg-shell/xdg-shell-unstable-v6.xml'],
+	[wl_protocol_dir, 'stable/xdg-shell/xdg-shell.xml'],
+	[wl_protocol_dir, 'unstable/idle-inhibit/idle-inhibit-unstable-v1.xml'],
+	'gamma-control.xml',
+	'gtk-primary-selection.xml',
 	'idle.xml',
 	'screenshooter.xml',
+	'server-decoration.xml',
```
emersion: No need to generate client protocols if they're not used.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,191 @@
...
+	struct wl_list link; // wlr_idle_inhibit_v1::clients
+
+	struct wl_list inhibitors;
+};
+
+static void idle_inhibit_inhibitor_destroy(struct wl_resource *resource) {
+	struct wlr_idle_inhibit_inhibitor_v1 *inhibitor = wl_resource_get_user_data(resource);
+	assert(inhibitor);
+
+	wl_signal_emit(&inhibitor->destroy, inhibitor->surface);
```
emersion: Use `wlr_signal_emit_safe`

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,191 @@
...
+	struct wlr_idle_inhibit_v1 *wlr_idle_inhibit;
+	struct wl_resource *resource;
+
+	struct wl_list link; // wlr_idle_inhibit_v1::clients
+
+	struct wl_list inhibitors;
+};
+
+static void idle_inhibit_inhibitor_destroy(struct wl_resource *resource) {
+	struct wlr_idle_inhibit_inhibitor_v1 *inhibitor = wl_resource_get_user_data(resource);
```
emersion: Wrap `wl_resource_get_user_data` with a helper function that checks the type of the resource.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,191 @@
...
+	wl_signal_emit(&inhibitor->destroy, inhibitor->surface);
+
+	wl_list_remove(&inhibitor->link);
+	wl_list_remove(&inhibitor->surface_destroy.link);
+	free(inhibitor);
+}
+
+static void idle_inhibit_inhibitor_handle_surface_destroy(struct wl_listener *listener,
+		void *data) {
+	assert(listener);
```
emersion: No need for this assertion.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,191 @@
...
+		struct wl_resource *manager_resource) {
+	wl_resource_destroy(manager_resource);
+}
+
+static struct zwp_idle_inhibit_manager_v1_interface idle_inhibit_impl = {
+	.destroy = idle_inhibit_manager_handle_destroy,
+	.create_inhibitor = wlr_create_inhibitor,
+};
+
+void wlr_idle_inhibit_v1_destroy(struct wlr_idle_inhibit_v1 *idle_inhibit);
```
emersion: No need for this forward declaration.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,191 @@
...
+
+	wlr_idle_inhibit_v1_destroy(idle_inhibit);
+}
+
+static void idle_inhibit_bind(struct wl_client *wl_client, void *data,
+		uint32_t version, uint32_t id) {
+	struct wlr_idle_inhibit_v1 *idle_inhibit = data;
+	assert(wl_client && idle_inhibit);
+
+	struct wlr_idle_inhibit_manager *manager = calloc(1, sizeof(struct wlr_idle_inhibit_manager));
```
emersion: Nitpick: 80 char limit

emersion:
Integration with `wlr_idle`, if there is one, should be purely optional. Since wlroots doesn't know where a surface is displayed (ie. on which output), an integration would be tricky. Unless you can come up with a good design, we should probably leave that to the compositor.

@SirCmpwn Do we want a rootston implementation of this?

SirCmpwn:
>Do we want a rootston implementation of this?

Probably, yeah

examples/idle-inhibit.c
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	draw();
+}
+
+/* Function that just does nothing.
+ * When it is noop(void) (like draw) the compiler complains about type
+ * mismatches in the listener struct.
+ * Without any arguments, it can be implicitly casted
+ */
```
emersion: I don't think this comment is necessary. I assume people should just know about that.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,188 @@
...
+	.destroy = idle_inhibit_inhibitor_v1_handle_destroy,
+};
+
+static void wlr_create_inhibitor(struct wl_client *client,
+		struct wl_resource *resource, uint32_t id,
+		struct wl_resource *surface_resource) {
+	struct wlr_surface *surface = wlr_surface_from_resource(surface_resource);
+	struct wlr_idle_inhibit_manager_v1 *manager =
+		wlr_idle_inhibit_manager_v1_from_resource(resource);
+	assert(surface && manager);
```
emersion: This assertion can be removed.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,188 @@
...
+	struct wlr_idle_inhibit_manager_v1 *idle_inhibit =
+		wl_container_of(listener, idle_inhibit, display_destroy);
+
+	wlr_idle_inhibit_v1_destroy(idle_inhibit);
+}
+
+static void idle_inhibit_bind(struct wl_client *wl_client, void *data,
+		uint32_t version, uint32_t id) {
+	struct wlr_idle_inhibit_manager_v1 *idle_inhibit = data;
+	assert(wl_client && idle_inhibit);
```
emersion: Same here.

protocol/meson.build
```diff
@@ -32,6 +33,10 @@ protocols = [
 
 client_protocols = [
 	[wl_protocol_dir, 'unstable/xdg-shell/xdg-shell-unstable-v6.xml'],
+	[wl_protocol_dir, 'stable/xdg-shell/xdg-shell.xml'],
+	[wl_protocol_dir, 'unstable/idle-inhibit/idle-inhibit-unstable-v1.xml'],
+	'gamma-control.xml',
+	'gtk-primary-selection.xml',
```
emersion: gamma-control and gtk-primary-selection are unused

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,188 @@
...
+wlr_idle_inhibit_inhibitor_v1_from_resource(struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &zwp_idle_inhibitor_v1_interface,
+		&idle_inhibitor_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void idle_inhibit_inhibitor_destroy(struct wl_resource *resource) {
+	struct wlr_idle_inhibit_inhibitor_v1 *inhibitor =
+		wlr_idle_inhibit_inhibitor_v1_from_resource(resource);
+	assert(inhibitor);
```
emersion: Same

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,188 @@
...
+
+void wlr_idle_inhibit_v1_destroy(struct wlr_idle_inhibit_manager_v1 *idle_inhibit) {
+	if (!idle_inhibit) {
+		return;
+	}
+
+	wl_list_remove(&idle_inhibit->display_destroy.link);
+
+	struct wlr_idle_inhibit_inhibitor_v1 *inhibitor;
+	struct wlr_idle_inhibit_inhibitor_v1 *tmp;
```
emersion: This can be on the same line as the other declaration

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,188 @@
...
+	wl_list_for_each_safe(inhibitor, tmp, &idle_inhibit->inhibitors, link) {
+		wl_resource_destroy(inhibitor->resource);
+	}
+
+	struct wl_resource *resource;
+	struct wl_resource *tmp_resource;
```
emersion: Same

include/wlr/types/wlr_idle_inhibit_v1.h
```diff
@@ -0,0 +1,33 @@
...
+	struct wl_global *global;
+	
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal new_inhibitor;
+	} events;
+};
+
+struct wlr_idle_inhibit_inhibitor_v1 {
```
emersion: The interface name is:

```xml
<interface name="zwp_idle_inhibitor_v1" version="1">
```

So this should be named `wlr_idle_inhibitor_v1`.

emersion:
Apart from minor comments I just posted, code LGTM.

So the next step is to implement it in rootston. We'll probably want a config option for the idle timeout + integration with wlr_idle. We'll see if we can bring part of this integration into wlroots, but I don't think it'll be possible.

One tricky thing is that the inhibitor is bound to a surface and we have no way of getting directly a surface position in rootston.

examples/idle-inhibit.c
SirCmpwn: Still, make it conform to the style guidelines

```c
static void noop() {
    /* This space intentionally left blank */
}
```

Also, don't these functions take arguments? You need to specify them or you might FUBAR the stack (depending on the target's ABI).

examples/idle-inhibit.c
```diff
@@ -0,0 +1,201 @@
...
+	if (idle_inhibitor) {
+		color[0] = 0.0;
+	}
+
+	glViewport(0, 0, width, height);
+	glClearColor(color[0], color[1], color[2], 1.0);
+	glClear(GL_COLOR_BUFFER_BIT);
+
+	eglSwapBuffers(egl.display, egl_surface);
+}
```
SirCmpwn: Just put this at the top instead of forward declaring it?

examples/idle-inhibit.c
```diff
@@ -0,0 +1,201 @@
...
+static const struct xdg_toplevel_listener xdg_toplevel_listener = {
+	.configure = xdg_toplevel_handle_configure,
+	.close = xdg_toplevel_handle_close,
+};
+
+static void handle_global(void *data, struct wl_registry *registry,
+		uint32_t name, const char *interface, uint32_t version) {
+	if (strcmp(interface, "wl_compositor") == 0) {
+		compositor = wl_registry_bind(registry, name, &wl_compositor_interface,
+			1);
```
SirCmpwn: Split it at `&wl_compositor_interface`, just "1" all alone here looks sad

examples/idle-inhibit.c
```diff
@@ -0,0 +1,201 @@
...
+		idle_inhibit_manager = wl_registry_bind(registry, name,
+			&zwp_idle_inhibit_manager_v1_interface, 1);
+	} else if (strcmp(interface, wl_seat_interface.name) == 0) {
+		seat = wl_registry_bind(registry, name, &wl_seat_interface, version);
+	}
+}
+
+static void handle_global_remove(void *data, struct wl_registry *registry,
+		uint32_t name) {
+	// TODO
```
SirCmpwn: I would just replace this with `// Who cares`

SirCmpwn:
LGTM overall

>So the next step is to implement it in rootston. We'll probably want a config option for the idle timeout + integration with wlr_idle. We'll see if we can bring part of this integration into wlroots, but I don't think it'll be possible.

Let's do this on a separate PR

protocol/meson.build
Ongy: Did I add them? Then a rebase in between must have gone horribly wrong, since I certainly did not do this on purpose...

include/wlr/types/wlr_idle_inhibit_v1.h
Ongy: The guidelines say to use `wlr_submodule_<name>` for typenames.
This is the `idle_inhibit_v1` submodle. So we we not use the same naming conventions for protocol implementations? If that's the case, the proper naming scheme should probably be added to the guidelines.

types/wlr_idle_inhibit_v1.c
Ongy: Unless that's officially part of the coding guidelines I would rather not, since I consider it horrible style.

And I would also like to point towards the kernel coding guidelines (which are referenced as base in the wlroots guidelines) and quote:

> It's also important to comment data, whether they are basic types or derived
> types.  To this end, use just one data declaration per line (no commas for
> multiple data declarations).  This leaves you room for a small comment on each
> item, explaining its use.

examples/idle-inhibit.c
Ongy: I would asume the compiler implements it correctly, when it does not specify arguments, but I would not want to be quoted on that.
Let me check if I can find something, declaring multiple `noop` functions just feels silly to me.

I can't find anything in the wlroots nor the kernel coding style guids about the comment for empty blocks. Could we have that added for future reference?

include/wlr/types/wlr_idle_inhibit_v1.h
```diff
@@ -0,0 +1,33 @@
...
+
+	struct wl_list link; // wlr_idle_inhibit_manager_v1::inhibitors;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+};
+
+struct wlr_idle_inhibit_manager_v1 *wlr_idle_inhibit_v1_create(struct wl_display *display);
+void wlr_idle_inhibit_v1_destroy(struct wlr_idle_inhibit_manager_v1 *idle_inhibit);
```
acrisci: Would like to see some docs in the header to motivate the usage of this protocol.

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,185 @@
...
+#include <wlr/types/wlr_idle_inhibit_v1.h>
+#include "wayland-util.h"
+#include "wayland-server.h"
+#include "idle-inhibit-unstable-v1-protocol.h"
+
+static struct zwp_idle_inhibit_manager_v1_interface idle_inhibit_impl;
+
+static struct zwp_idle_inhibitor_v1_interface idle_inhibitor_impl;
+
+struct wlr_idle_inhibit_manager_v1 *
```
acrisci: Static?

types/wlr_idle_inhibit_v1.c
```diff
@@ -0,0 +1,185 @@
...
+wlr_idle_inhibit_manager_v1_from_resource(struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &zwp_idle_inhibit_manager_v1_interface,
+		&idle_inhibit_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+struct wlr_idle_inhibitor_v1 *
+wlr_idle_inhibitor_v1_from_resource(struct wl_resource *resource) {
```
acrisci: Static?

types/wlr_idle_inhibit_v1.c
Ongy: Or exported... But should be either one of those.

Do we have a policy on when to export the `wl_resource_get_user_data` wrappers?

Ongy:
I don't get that last question.
We don't have any idle handling in rootston yet, and the last word on that was to do it in another PR.

But even for that, I don't see how i3lock would interact with this. (Outside being opaque over the entire output, thus hiding all inhibitors).

examples/idle-inhibit.c
```diff
@@ -52,40 +65,58 @@ static void pointer_handle_button(void *data, struct wl_pointer *pointer, uint32
...
+static void pointer_handle_axis(void *data, struct wl_pointer *wl_pointer,
+		uint32_t time, uint32_t axis, wl_fixed_t value) {
+	/* NOOP: ignore event */
+}
 
-	glViewport(0, 0, width, height);
-	glClearColor(color[0], color[1], color[2], 1.0);
-	glClear(GL_COLOR_BUFFER_BIT);
+static void pointer_handle_frame(void *data, struct wl_pointer *wl_pointer) {
+	/* NOOP: ignore event */
```
SirCmpwn: We use

    // This space intentionally left blank

emersion:
>What if the idle action is a user action like running i3lock?

This doesn't change anything, the compositor should prevent i3lock from being spawned.

types/wlr_idle_inhibit_v1.c
emersion: The policy so far has been "export it if we need it elsewhere". Even if this is reasonable for protocols like this one, we can't guess what users will need. So maybe we should export all of them.

PR #645, opened 2018-02-18
x11 backend: set window title

agx:
This makes windows identifiable in the window list

### DIFF
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -221,10 +221,20 @@ static bool wlr_x11_backend_start(struct wlr_backend *backend) {
 
 	init_atom(x11, &x11->atoms.wm_protocols, 1, "WM_PROTOCOLS");
 	init_atom(x11, &x11->atoms.wm_delete_window, 0, "WM_DELETE_WINDOW");
+	init_atom(x11, &x11->atoms.net_wm_name, 1, "_NET_WM_NAME");
+	init_atom(x11, &x11->atoms.utf8_string, 0, "UTF8_STRING");
 
 	xcb_change_property(x11->xcb_conn, XCB_PROP_MODE_REPLACE, output->win,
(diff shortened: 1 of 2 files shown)


include/backend/x11.h
```diff
@@ -46,6 +46,8 @@ struct wlr_x11_backend {
 	struct {
 		struct wlr_x11_atom wm_protocols;
 		struct wlr_x11_atom wm_delete_window;
+	        struct wlr_x11_atom net_wm_name;
```
SirCmpwn: Weird indent here

backend/x11/backend.c
```diff
@@ -221,11 +221,20 @@ static bool wlr_x11_backend_start(struct wlr_backend *backend) {
...
 	init_atom(x11, &x11->atoms.wm_delete_window, 0, "WM_DELETE_WINDOW");
+	init_atom(x11, &x11->atoms.net_wm_name, 1, "_NET_WM_NAME");
+	init_atom(x11, &x11->atoms.utf8_string, 0, "UTF8_STRING");
 
 	xcb_change_property(x11->xcb_conn, XCB_PROP_MODE_REPLACE, output->win,
 		x11->atoms.wm_protocols.reply->atom, XCB_ATOM_ATOM, 32, 1,
 		&x11->atoms.wm_delete_window.reply->atom);
 
+	char title[32];
+	if (snprintf(title, 32, "wlroots - %s", output->wlr_output.name))
```
SirCmpwn: Add `{}` to this if statement please

emersion: Also, can you use `sizeof(title)` so that there's no bad surprise if we change it?

agx: All updated. Thanks for the review.

SirCmpwn:
Thanks!

PR #654, opened 2018-02-20
rootston: send frame done events unconditionally

### LINKED ISSUE #643: mpv's video output freezes after seeking in rootston
Using master for both mpv and wlroots.

According to @atomnuker this is due to mishandling of frame callbacks

### LINKED ISSUE #644: Send frame events to surfaces unconditionally
Right now we don't send frame events to surfaces when we're skipping the renderer. We should send those unconditionally, so clients can sync to the vblank cycle even if they don't always damage their content.

emersion:
Fixes #644

### DIFF
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -535,6 +535,11 @@ damage_finish:
 
 	// Send frame done events to all surfaces
 	if (output->fullscreen_view != NULL) {
+		if (wlr_output->fullscreen_surface == output->fullscreen_view->wlr_surface) {
+			// The surface is managed by the wlr_output
+			return;
+		}
+
 		view_for_each_surface(view, surface_send_frame_done, &data);
 
 #ifdef WLR_HAS_XWAYLAND
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -182,6 +182,21 @@ static void xwayland_children_for_each_surface(
 }
 #endif
 
+static void drag_icons_for_each_surface(struct roots_input *input,
+		surface_iterator_func_t iterator, void *user_data) {
+	struct roots_seat *seat;
+	wl_list_for_each(seat, &input->seats, link) {
+		struct roots_drag_icon *drag_icon;
+		wl_list_for_each(drag_icon, &seat->drag_icons, link) {
+			if (!drag_icon->wlr_drag_icon->mapped) {
+				continue;
+			}
+			surface_for_each_surface(drag_icon->wlr_drag_icon->surface,
+				drag_icon->x, drag_icon->y, 0, iterator, user_data);
+		}
+	}
+}
+
 
 struct render_data {
 	struct roots_output *output;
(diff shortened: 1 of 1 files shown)


rootston/output.c
```diff
@@ -512,6 +518,25 @@ static void render_output(struct roots_output *output) {
...
+#ifdef WLR_HAS_XWAYLAND
+		if (view->type == ROOTS_XWAYLAND_VIEW) {
+			xwayland_children_for_each_surface(view->xwayland_surface,
+				surface_send_frame_done, &now);
+		}
+#endif
+	} else {
+		struct roots_view *view;
+		wl_list_for_each_reverse(view, &desktop->views, link) {
+			view_for_each_surface(view, surface_send_frame_done, &now);
```
Ongy: This should check that the surface actually intersects the output.

Ongy: This should check that the surface actually intersects the output.

emersion: Ah, you're right - for some reason I was believing `view_for_each_surface` filtered them.

rootston/output.c
```diff
@@ -512,6 +532,26 @@ static void render_output(struct roots_output *output) {
 
 damage_finish:
 	pixman_region32_fini(&damage);
+
+	// Send frame done events to all surfaces
+	if (output->fullscreen_view != NULL) {
+		view_for_each_surface(view, surface_send_frame_done, &data);
```
martinetd: There should be a view = output->fullscreen_view before that?

Also, is that correct if the fullscreen view isn't fully opaque? How do we handle semi-transparent fullscreen stuff?

Ongy: The output->fullscreen_view effectivly disables compositing.

If you check the render code, it skips rendering anything else as well.

SirCmpwn: It does? That seems wrong, what if the fullscreen view is semitransparent?

emersion: The spec says fullscreen views are rendered on a black background.

SirCmpwn: The spec can get stuffed, then.

rootston/output.c
```diff
@@ -512,6 +532,31 @@ static void render_output(struct roots_output *output) {
...
+
+#ifdef WLR_HAS_XWAYLAND
+		if (view->type == ROOTS_XWAYLAND_VIEW) {
+			xwayland_children_for_each_surface(view->xwayland_surface,
+				surface_send_frame_done, &data);
+		}
+#endif
+	} else {
+		struct roots_view *view;
```
martinetd: (nitpick) there already is a struct roots_view *view 40 lines earlier, although that's in the middle of the code, so not quite sure how fair it is to reuse it.
Conversely, that's what let the earlier re-use of view work; might make sense to delimit such 'local' variables without flow in an artificial block?

rootston/output.c
```diff
@@ -512,6 +532,31 @@ static void render_output(struct roots_output *output) {
...
 	pixman_region32_fini(&damage);
+
+	// Send frame done events to all surfaces
+	if (output->fullscreen_view != NULL) {
+		if (wlr_output->fullscreen_surface == output->fullscreen_view->wlr_surface) {
+			// The surface is managed by the wlr_output
+			return;
+		}
+
+		view_for_each_surface(view, surface_send_frame_done, &data);
```
martinetd: I should have posted two separate comments, I think you missed my second point: `view` here is some random view (likely last of `desktop->views`), should it be assigned `output->fullscren_view` before using it here and below?

emersion: Ah, you're right, good catch.

rootston/output.c
emersion: @martinetd Also note that I originally didn't paid attention to send frame done events to surfaces only once, because doing so flushes the clients callbacks. But yeah, it's better to do it right.

rootston/output.c
```diff
@@ -384,6 +400,20 @@ static bool has_standalone_surface(struct roots_view *view) {
 	return true;
 }
 
+static void surface_send_frame_done(struct wlr_surface *surface, double lx,
+		double ly, float rotation, void *_data) {
```
acrisci: Why isn't `void *_data` typed?

occivink:
I can't see any difference, the same freezes happen.

Ongy:
Code looks good to me. But can confirm, it does  not fix #643

I remember some talk about removing the callbacks at some points where we should not, has that addressed?
Otherwise we should get that one done and test with both patches.
If that's still buggy, yell at atomnuker some more.

rootston/output.c
emersion: Because it's a callback for the `*_for_each_surface` functions.

emersion:
So in the end this isn't required to fix #643. But we should still do that, as it's a protocol violation not to do so.

SirCmpwn:
@Ongy @emersion if your comments translate into PR approval, indicate that in the UI and merge it.

PR #658, opened 2018-02-21
x11: parse vendor and model out of xcb setup information

agx:


backend/x11/backend.c
```diff
@@ -188,6 +188,23 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
 	atom->reply = xcb_intern_atom_reply(x11->xcb_conn, atom->cookie, NULL);
 }
 
+
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn)
```
acrisci: Brace should be on the same line as the function.

backend/x11/backend.c
```diff
@@ -188,6 +188,23 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
...
+	unsigned int length;
+
+	xcb_setup = xcb_get_setup (xcb_conn);
+	length = xcb_setup_vendor_length (xcb_setup);
+	if (length > sizeof(output->make)-1)
+		length = sizeof(output->make)-1;
+	memcpy (output->make, xcb_setup_vendor (xcb_setup), length);
+	output->make[length-1] = '\0';
+	sprintf(output->model, "%d.%d", xcb_setup->protocol_major_version,
+			xcb_setup->protocol_minor_version);
```
acrisci: use `snprintf` here.

backend/x11/backend.c
```diff
@@ -188,6 +188,23 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
...
 
+
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn)
+{
+	const xcb_setup_t *xcb_setup;
+	unsigned int length;
+
+	xcb_setup = xcb_get_setup (xcb_conn);
+	length = xcb_setup_vendor_length (xcb_setup);
+	if (length > sizeof(output->make)-1)
```
acrisci: always use braces on if statements.

backend/x11/backend.c
```diff
@@ -188,6 +188,23 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
 	atom->reply = xcb_intern_atom_reply(x11->xcb_conn, atom->cookie, NULL);
 }
 
+
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn)
+{
+	const xcb_setup_t *xcb_setup;
+	unsigned int length;
+
+	xcb_setup = xcb_get_setup (xcb_conn);
```
acrisci: Don't put spaces before function parens.

agx:
@acrisci thanks for review, all cleaned up now.

backend/x11/backend.c
```diff
@@ -188,6 +188,24 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
...
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn) {
+	const xcb_setup_t *xcb_setup;
+	unsigned int length;
+
+	xcb_setup = xcb_get_setup(xcb_conn);
+	length = xcb_setup_vendor_length (xcb_setup);
+	if (length > sizeof(output->make)-1) {
+		length = sizeof(output->make)-1;
+	}
+	memcpy (output->make, xcb_setup_vendor (xcb_setup), length);
```
ascent12: I'd prefer that snprintf is used for this as well, which would actually remove the need for a lot of the preceding code:
```c
snprintf(output->make, sizeof(output->make), "%.*s",
    xcb_setup_vendor_length(xcb_setup),
    xcb_setup_vendor(xcb_setup));
```

backend/x11/backend.c
```diff
@@ -188,6 +188,24 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
 	atom->reply = xcb_intern_atom_reply(x11->xcb_conn, atom->cookie, NULL);
 }
 
+
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn) {
+	const xcb_setup_t *xcb_setup;
+	unsigned int length;
+
+	xcb_setup = xcb_get_setup(xcb_conn);
+	length = xcb_setup_vendor_length (xcb_setup);
```
ascent12: No space before the `(`

backend/x11/backend.c
```diff
@@ -188,6 +188,24 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
 	atom->reply = xcb_intern_atom_reply(x11->xcb_conn, atom->cookie, NULL);
 }
 
+
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn) {
+	const xcb_setup_t *xcb_setup;
+	unsigned int length;
```
ascent12: We're not using C89, so putting your declarations at the front isn't necessary. You can just declare them you use use them.

backend/x11/backend.c
Ongy: Why not strncpy?

agx: xcb_setup_vendor is not null necessarily null terminated.

martinetd: That's what the %.*s of printf is about, it's safe :)

agx: @martinetd but it could still outgrow output->make

agx: @martinetd ...no it can't sorry for the noise.

Ongy: That's why it's `s*n*printf` the output->make size is the second argument to the function.

The maximum string length is then given as format option.

agx:
All addressed now. Thanks for the snprintf( , "%.*s", ) hint.

backend/x11/backend.c
```diff
@@ -188,6 +188,18 @@ static void init_atom(struct wlr_x11_backend *x11, struct wlr_x11_atom *atom,
...
+static void parse_xcb_setup(struct wlr_output *output, xcb_connection_t *xcb_conn) {
+	const xcb_setup_t *xcb_setup = xcb_get_setup(xcb_conn);
+
+	snprintf(output->make, sizeof(output->make), "%.*s",
+			xcb_setup_vendor_length(xcb_setup),
+			xcb_setup_vendor(xcb_setup));
+	snprintf(output->model, sizeof(output->model), "%"PRIu16".%"PRIu16,
+			xcb_setup->protocol_major_version,
+			xcb_setup->protocol_minor_version);
+
```
SirCmpwn: style: drop this empty line

agx:
Empty line removed.

PR #659, opened 2018-02-21
Make wlr_render_with_matrix use alpha

agx:
I wonder what's the best way forward here?

### DIFF
--- a/examples/rotation.c
+++ b/examples/rotation.c
@@ -52,7 +52,7 @@ static void handle_output_frame(struct output_state *output, struct timespec *ts
 			wlr_texture_get_matrix(sample->cat_texture, &matrix,
 				&wlr_output->transform_matrix, x, y);
 			wlr_render_with_matrix(sample->renderer,
-					sample->cat_texture, &matrix);
+					sample->cat_texture, &matrix, 1.0f);
 		}
 	}
 
--- a/include/wlr/render.h
+++ b/include/wlr/render.h
@@ -33,12 +33,13 @@ struct wlr_texture *wlr_render_texture_create(struct wlr_renderer *r);
  * 	float projection[16];
  * 	float matrix[16];
  * 	wlr_texture_get_matrix(texture, &matrix, &projection, 123, 321);
- * 	wlr_render_with_matrix(renderer, texture, &matrix);
+ * 	wlr_render_with_matrix(renderer, texture, &matrix, 0.5f);
  *
- * This will render the texture at <123, 321>.
+ * This will render the texture at <123, 321> with an alpha channel of 0.5.
  */
 bool wlr_render_with_matrix(struct wlr_renderer *r,
-	struct wlr_texture *texture, const float (*matrix)[16]);
+	struct wlr_texture *texture, const float (*matrix)[16], float alpha);
+
 /**
  * Renders a solid quad in the specified color.
  */
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -36,8 +36,8 @@ struct wlr_texture *wlr_render_texture_create(struct wlr_renderer *r) {
 }
 
 bool wlr_render_with_matrix(struct wlr_renderer *r,
-		struct wlr_texture *texture, const float (*matrix)[16]) {
-	return r->impl->render_with_matrix(r, texture, matrix);
+		struct wlr_texture *texture, const float (*matrix)[16], float alpha) {
+	return r->impl->render_with_matrix(r, texture, matrix, alpha);
 }
 
 void wlr_render_colored_quad(struct wlr_renderer *r,
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -25,6 +25,10 @@
 
 struct roots_view *view_create() {
 	struct roots_view *view = calloc(1, sizeof(struct roots_view));
+	if (!view) {
+		return NULL;
+	}
+	view->alpha = 1.0f;
 	return view;
 }
 
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -278,6 +278,15 @@ void view_rotate(struct roots_view *view, float rotation) {
 	view_damage_whole(view);
 }
 
+void view_cycle_alpha(struct roots_view *view) {
+	view->alpha -= 0.05;
+	/* Don't go completely transparent */
+	if (view->alpha < 0.1) {
+		view->alpha = 1.0;
+	}
+	view_damage_whole(view);
+}
+
 void view_close(struct roots_view *view) {
 	if (view->close) {
 		view->close(view);
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -22,6 +22,12 @@
 #include "rootston/view.h"
 #include "rootston/xcursor.h"
 
+
+struct roots_view *view_create() {
+	struct roots_view *view = calloc(1, sizeof(struct roots_view));
+	return view;
+}
+
 void view_get_box(const struct roots_view *view, struct wlr_box *box) {
 	box->x = view->x;
 	box->y = view->y;
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -356,7 +356,7 @@ static void render_decorations(struct roots_view *view,
 	float matrix[16];
 	wlr_matrix_project_box(&matrix, &box, WL_OUTPUT_TRANSFORM_NORMAL,
 		view->rotation, &output->wlr_output->transform_matrix);
-	float color[] = { 0.2, 0.2, 0.2, 1 };
+	float color[] = { 0.2, 0.2, 0.2, view->alpha };
 
 	int nrects;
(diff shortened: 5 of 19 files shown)


acrisci:
I think this general approach is ok.

I would not overload the method though. Just have the one render function with the alpha and add it to the other `render_*` methods too.

agx:
@acrisci render_quad and render_ellipse get a color argument which includes alpha, is there anything else missing?

examples/rotation.c
```diff
@@ -47,13 +47,17 @@ static void handle_output_frame(struct output_state *output, struct timespec *ts
...
 		for (int x = -128 + (int)odata->x_offs; x < width; x += 128) {
 			wlr_texture_get_matrix(sample->cat_texture, &matrix,
 				&wlr_output->transform_matrix, x, y);
 			wlr_render_with_matrix(sample->renderer,
-					sample->cat_texture, &matrix);
+				sample->cat_texture, &matrix, alpha);
 		}
+		alpha -= 0.1;
+		if (alpha < 0.0)
+			alpha = 1.0;
```
SirCmpwn: style - use braces here

SirCmpwn: Also, this seems like a weird place to test alpha. Can we add a command to rootston to set the alpha of a window?

include/wlr/render.h
```diff
@@ -32,13 +32,15 @@ struct wlr_texture *wlr_render_texture_create(struct wlr_renderer *r);
  * 	struct wlr_texture *texture;
  * 	float projection[16];
  * 	float matrix[16];
+ *  float alpha;
```
emersion: s/spaces/tab/

render/wlr_renderer.c
```diff
@@ -36,8 +36,8 @@ struct wlr_texture *wlr_render_texture_create(struct wlr_renderer *r) {
 }
 
 bool wlr_render_with_matrix(struct wlr_renderer *r,
-		struct wlr_texture *texture, const float (*matrix)[16]) {
-	return r->impl->render_with_matrix(r, texture, matrix);
+	        struct wlr_texture *texture, const float (*matrix)[16], float alpha) {
+        return r->impl->render_with_matrix(r, texture, matrix, alpha);
```
emersion: s/spaces/tabs/

agx:
@SirCmpwn  more like this?

rootston/desktop.c
```diff
@@ -22,6 +22,18 @@
 #include "rootston/view.h"
 #include "rootston/xcursor.h"
 
+
+struct roots_view *view_create()
+{
```
SirCmpwn: style - { on the same line

rootston/desktop.c
```diff
@@ -22,6 +22,18 @@
...
+	struct roots_view *view;
+
+	view = calloc(1, sizeof(struct roots_view));
+	if (!view)
+		return NULL;
```
SirCmpwn: style - always use braces on your if statements

agx:
Sorry about the lack of braces, still have to get used to that. Updated.

emersion:
Please rebase your PR!

agx:
Done.

include/wlr/render.h
```diff
@@ -32,13 +32,15 @@ struct wlr_texture *wlr_render_texture_create(struct wlr_renderer *r);
  * 	struct wlr_texture *texture;
  * 	float projection[16];
  * 	float matrix[16];
+ * 	float alpha;
```
Ongy: `alpha` isn't used.

include/wlr/render.h
```diff
@@ -32,13 +32,15 @@ struct wlr_texture *wlr_render_texture_create(struct wlr_renderer *r);
...
  * 	wlr_texture_get_matrix(texture, &matrix, &projection, 123, 321);
- * 	wlr_render_with_matrix(renderer, texture, &matrix);
+ * 	wlr_render_with_matrix(renderer, texture, &matrix, 0.5f);
  *
- * This will render the texture at <123, 321>.
+ * This will render the texture at <123, 321> with an alpha channel of 0.5.
```
Ongy: What's this exactly?
It sounds to me like every alpha value in the texture would be set to 0.5, but judging from changes to the code, I would asume it's actuall half of the original value in the texture.

rootston/output.c
```diff
@@ -37,9 +37,9 @@ static void rotate_child_position(double *sx, double *sy, double sw, double sh,
 }
 
 static void surface_for_each_surface(struct wlr_surface *surface, double lx,
-		double ly, float rotation, surface_iterator_func_t iterator,
+		double ly, float rotation, float alpha, surface_iterator_func_t iterator,
```
Ongy: Since #654 we have anoter user of the for_each functions, which currently does not require alpha.

I feel this should be in the user_data struct.

include/wlr/render.h
agx: What do you mean by not used? It's in the wlr_render_with_matrix.

include/wlr/render.h
agx: 1.0f means texture RGB as is while 0.0 means background shines perfectly through. Try CTRL-shift-a on any window in rootston.

SirCmpwn: Wayland surfaces can be RGBA, not just RGB. Your wording here implies the alpha channel is overwritten when it is in fact a multiplier.

rootston/output.c
agx: I thought the same first but once I started to shift things around it looked more readable since we don't need to deref *data all the time.

include/wlr/render.h
Ongy: This small example defines a few things used in the calls bellow.

I can see a literal `0.5f` at a place a variable named alpha would make sense though.
All other literals are inline, so I don't think it would make sense to declare and set alpha here.

rootston/output.c
Ongy: I can't see any point where `alpha` is used passed without also passing `user_data`.

And it is an unused argument to all iterator functions but `render_surface`.

agx:
Moved alpha into render_data

rootston/config.c
```diff
@@ -428,6 +428,7 @@ struct roots_config *roots_config_create_from_args(int argc, char *argv[]) {
 		add_binding_config(&config->bindings, "Logo+Shift+E", "exit");
 		add_binding_config(&config->bindings, "Ctrl+q", "close");
 		add_binding_config(&config->bindings, "Alt+Tab", "next_window");
+		add_binding_config(&config->bindings, "Ctrl+Shift+a", "alpha");
```
emersion: Please don't add this as a default binding.

rootston/desktop.c
```diff
@@ -22,6 +22,18 @@
 #include "rootston/view.h"
 #include "rootston/xcursor.h"
 
+
+struct roots_view *view_create() {
+	struct roots_view *view;
+
+	view = calloc(1, sizeof(struct roots_view));
```
emersion: Declaration and assignment can be on the same line.

rootston/output.c
```diff
@@ -516,6 +517,7 @@ static void render_output(struct roots_output *output) {
 	// Render all views
 	struct roots_view *view;
 	wl_list_for_each_reverse(view, &desktop->views, link) {
+		data.alpha = view->alpha;
```
emersion: This should be done in `render_view`.

agx:
Dropped default key binding and moved settings alpha into render_view.

Ongy:
The server side decorations should use the alpha.
We can't handle them any different in the CSD case, so for consistency the SSD should follow.

There's also two different things, but they should probably be follow up issues:
SSDs are a single rectangle and drawn under the entire surface. This has been the case before, just wasn't visible so far.

XWayland children are drawn twice, once in the renderer for XWayland, and once becaus they have their own view. With this there's a fully opaque window above the transparent one, which is again inconsistent to wayland clients.

agx:
@Ongy SSD honor alpha now.

rootston/desktop.c
```diff
@@ -233,6 +243,7 @@ void view_set_fullscreen(struct roots_view *view, bool fullscreen,
 		view->saved.x = view->x;
 		view->saved.y = view->y;
 		view->saved.rotation = view->rotation;
+		view->saved.alpha = view->alpha;
```
Ongy: Is that every read?

It see a reason to set to full opacity for fullscreen, but that doesn't seem to be implemented.

agx:
@Ongy moving to fullscreen drops transparency for me and moving out of fullscreen brings it back, tested with eog.

agx:
@Ongy I've removed that alpha saving for fullscreen since it does not seem to be necessary (although I can't see why atm).

agx:
Ah...output_fullscreen_surface_render renders with alpha = 1.0f so saving is not necessary. Sorry for the noise.

agx:
Thanks for pulling this in!

PR #661, opened 2018-02-22
move activity_notify into events in wlr_idle

Ongy:
This moves the `struct wl_signal activity_notify` in `struct wlr_idle`
into a local `struct {} events` to keep consistency with other modules
in the library.

Pointing it out on my PR but merging one that doesn't do it >:[
Just noticed that, checking the naming for protocol interface implementations.


emersion:
Thanks!

PR #665, opened 2018-02-23
Add tags to anonymous structs

Armael:
At present the wlroots interface includes several anonymous structs, which can make it difficult for some tools to create bindings to wlroots. I stumbled on this issue while writing OCaml FFI bindings for wlroots.

## The problem

Currently, a number of struct types have neither a tag name nor a typedef alias. For example, struct `wlr_backend` contains a [struct type without a name](https://github.com/swaywm/wlroots/blob/master/include/wlr/backend.h#L13):

```c
    struct {
        struct wl_signal destroy;
        struct wl_signal new_input;
        struct wl_signal new_output;
    } events;
```

Anonymous struct types like the type above make it difficult or impossible for certain tools to create foreign language bindings to a library. For example, one binding strategy (used, for instance, in the [OCaml ctypes library](https://github.com/ocamllabs/ocaml-ctypes)) is to generate C code that determines the layout of C objects, then use the layout information to access those objects directly from the foreign language. The generated C code uses type names to determine layout, like this:

```c
    … sizeof(struct foo) …
    … offsetof (struct foo, field_x) …
    … alignmentof (struct foo) …
```

Clearly this approach won't work if the types involved don't have names.

## The fix

The patch adds tags to all the anonymous types that form part of the public interface. The tag names are derived from the enclosing type and the name of the member — for example, the type of the `events` member in `wlr_backend` becomes struct `wlr_backend_events`.

-------------

Note that fully anonymous structs and unions with no type and no member name are fine. [For example](https://github.com/swaywm/wlroots/blob/master/include/backend/drm/drm.h#L49):
```c
    union {
        struct {
            struct wlr_drm_plane *overlay;
            struct wlr_drm_plane *primary;
            struct wlr_drm_plane *cursor;
        };
        struct wlr_drm_plane *planes[3];
    };
```

 Since the intermediate anonymous structs and unions cannot be named in C code, simply inferring the layout of their members is enough.

### DIFF
--- a/include/backend/multi.h
+++ b/include/backend/multi.h
@@ -13,7 +13,7 @@ struct wlr_multi_backend {
 
 	struct wl_listener display_destroy;
 
-	struct {
+	struct wlr_multi_backend_events {
 		struct wl_signal backend_add;
 		struct wl_signal backend_remove;
 	} events;
--- a/include/backend/wayland.h
+++ b/include/backend/wayland.h
@@ -45,7 +45,7 @@ struct wlr_wl_backend_output {
 	struct wl_egl_window *egl_window;
 	struct wl_callback *frame_callback;
 
-	struct {
+	struct wlr_wl_backend_cursor {
 		struct wl_shm_pool *pool;
 		void *buffer; // actually a (client-side) struct wl_buffer*
 		uint32_t buf_size;
--- a/include/backend/x11.h
+++ b/include/backend/x11.h
@@ -43,7 +43,7 @@ struct wlr_x11_backend {
 	struct wl_event_source *event_source;
 	struct wl_event_source *frame_timer;
 
-	struct {
+	struct wlr_x11_backend_events {
 		struct wlr_x11_atom wm_protocols;
 		struct wlr_x11_atom wm_delete_window;
 		struct wlr_x11_atom net_wm_name;
--- a/include/rootston/config.h
+++ b/include/rootston/config.h
@@ -13,7 +13,7 @@ struct roots_output_config {
 	int x, y;
 	float scale;
 	struct wl_list link;
-	struct {
+	struct roots_output_config_mode {
 		int width, height;
 		float refresh_rate;
 	} mode;
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -90,13 +90,13 @@ struct roots_view {
 
 	bool maximized;
 	struct roots_output *fullscreen_output;
-	struct {
+	struct roots_view_saved {
 		double x, y;
 		uint32_t width, height;
 		float rotation;
 	} saved;
 
-	struct {
+	struct roots_view_pending_move_resize {
 		bool update_x, update_y;
 		double x, y;
 		uint32_t width, height;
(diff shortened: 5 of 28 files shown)


emersion:
Can't you define these structs in your code?

Armael:
What do you mean? I could probably copy-paste wlroots' headers files, modify them, and ship those with the ocaml library, but that doesn't seem ideal from a maintainability perspective.

Ongy:
Can't find it right now, but I'm pretty sure @SirCmpwn stated before he doesn't want to change valid C for interop with C++ bindings. I wouldn't bet on Ocaml being different.

I feel like this should be a bug report to your tooling. If it claims to support C FFIs, it certainly has a bug you discovered.

(Side note: works for me TM. I have Haskell bindings)

Armael:
I see, thank you for your answer. To give a bit more context about FFI bindings and OCaml; there exists currently two main approaches for writing C bindings:

- The [traditional approach](http://caml.inria.fr/pub/docs/manual-ocaml/intfc.html) requires writing some additional C code that acts as a wrapper, and explains how to translate between values of the C library and OCaml values. This approach would perfectly work with wlroots as it is now. Its main drawback though is that it is inherently "unsafe": a mistake in the wrapper code (in its view of the world on the C side or the OCaml side) usually results in a crash at runtime.


- The other approach is by using the [OCaml ctypes](https://github.com/ocamllabs/ocaml-ctypes) library, which doesn't require the user to write any C code. Instead the user writes a description of the C API in OCaml, leveraging the OCaml typechecker, and the library does some magic to generate C stubs that correspond to this description. This approach is much safer, and a mistake in describing the C API results in an error at compile-time. It is also in general quite convenient and automates a lot of grunt work compared to the first approach. However, as I detailed a bit above, at some point in this process, the library needs to determine the layout of the C structures involved. It does so by producing C code containing calls to `sizeof`, `offsetof`,..., hence the need for the additional names and the present PR. 

In short, I could probably instead write my wlroots bindings using the first approach, but it would be super nice if the second, more convenient approach was also supported :-).

SirCmpwn:
@Ongy is correct, you should either improve OCaml's FFI or you should copypasta the structures into your code. Once the API becomes stable the copypasta approach will be less of an issue.

martinetd:
I also don't see how this is a problem, at least offsetof works fine. I just wrote this very simple test:
```c
#include <stddef.h>
#include <stdio.h>

struct testme {
	struct {
		char foo;
		char moo;
		char *bar;
	} events;
	char foo;
};

int main() {
	printf("events.foo: %zd\n", offsetof(struct testme, events.foo));
	printf("events.moo: %zd\n", offsetof(struct testme, events.moo));
	printf("events.bar: %zd\n", offsetof(struct testme, events.bar));
	printf("foo: %zd\n", offsetof(struct testme, foo));
	return 0;
}
```

and it prints as I'd expect:
```
events.foo: 0
events.moo: 1
events.bar: 8
foo: 16
```

You just need to find a way to print use `.` in your field name

Armael:
Hmm, indeed, it seems I can use e.g. `events.foo` as a "field" name in ctypes, which then computes the offset as expected. Nice trick, thanks! 

Now, I won't be able to get a pointer to the `event` member structure itself -- but maybe the fact that its type has no name is to indicate that one shouldn't do that anyway?

SirCmpwn:
Correct, that's discouraged. But the pointer to the first struct member is also a pointer to that struct.

PR #667, opened 2018-02-24
prevent screenshot from resuing template array

Ongy:
The template array given to mkstemp was declared static. This reused the
memory, which caused mkstemp to fail if backingfile is run more than
once, because the array no longer contained the template syntax
(which is forced to end in XXXXXX) but the previous file name.


SirCmpwn:
tyvm

PR #668, opened 2018-02-25
Fix null deref in wlr_libinput_backend_destroy

dlrobertson:
If input_event is null (e.g. if backend_start has not been called yet)
`wl_event_source_remove` will result in a null deref.

Hit this when running `rootston` at [rootston/main.c:51](https://github.com/swaywm/wlroots/blob/master/rootston/main.c#L51)


SirCmpwn:
Thanks!

Issue #676, opened 2018-02-25
Alt-tabbing out of a fullscreen client doesn’t display the next client


linkmauve:
It gives the focus to the next client, but the buffer displayed is still the one of the fullscreen client.

acrisci:
Can confirm this bug is still present.

PR #680, opened 2018-02-26
unset fullscreen when surface is brought to front

Ongy:
The output fullscreen surfaces are drawn in front of everything, without
consideration for view z-order.
If a view is brought to front, unset any fullscreen view that would
cover this view to make sure the view is visible.

Testplan: Fullscreen a view, then `[Alt]+[Tab]` to a surface that intersects the relevant output.

Fixes https://github.com/swaywm/wlroots/issues/676
wl_shell mpv misbehaves in two ways.
* It doesn't notice it gets unfullscreened (from a quick check a protocol issue)
* It is placed at the wrong coordinates.

### DIFF
--- a/rootston/seat.c
+++ b/rootston/seat.c
@@ -715,16 +715,6 @@ struct roots_seat_view *roots_seat_view_from_view(
 	return seat_view;
 }
 
-static void release_fullscreen(struct roots_output *output) {
-	if (output->fullscreen_view) {
-		if (output->fullscreen_view->set_fullscreen) {
-			output->fullscreen_view->set_fullscreen(
-				output->fullscreen_view, false);
-		}
-		view_set_fullscreen(output->fullscreen_view, false, output->wlr_output);
-	}
-}
-
 void roots_seat_set_focus(struct roots_seat *seat, struct roots_view *view) {
 	// Make sure the view will be rendered on top of others, even if it's
 	// already focused in this seat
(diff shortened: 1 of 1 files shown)


Ongy:
The xdg_shell version of mpv (i.e. newer mpv) works fine with regards to both problems mentioned.

rootston/seat.c
```diff
@@ -706,6 +706,16 @@ struct roots_seat_view *roots_seat_view_from_view(
 	return seat_view;
 }
 
+static void release_fullscreen(struct roots_output *output) {
+	if (output->fullscreen_view) {
+		if (output->fullscreen_view->set_fullscreen) {
+			output->fullscreen_view->set_fullscreen(
+				output->fullscreen_view, false);
```
emersion: This is already done by `view_set_fullscreen`

rootston/seat.c
```diff
@@ -706,6 +706,16 @@ struct roots_seat_view *roots_seat_view_from_view(
...
+		}
+		view_set_fullscreen(output->fullscreen_view, false, output->wlr_output);
```
emersion: The `output` argument makes no sense here? (Just pass NULL?)

rootston/seat.c
```diff
@@ -714,6 +724,17 @@ void roots_seat_set_focus(struct roots_seat *seat, struct roots_view *view) {
 		wl_list_insert(&seat->input->server->desktop->views, &view->link);
 	}
 
+	struct roots_desktop *desktop = view->desktop;
+	struct roots_output *output;
+	struct wlr_box box;
+	view_get_box(view, &box);
+	wl_list_for_each(output, &desktop->outputs, link) {
+		if (wlr_output_layout_intersects(desktop->layout,
+				output->wlr_output, &box)) {
```
emersion: I don't think this works with xwayland popups for fullscreened xwayland views.

Ongy: Works better than firefox going completly black for me on the current master if it's right clicked while fullscreen...

Got a likely fix, but can't exactly test with this behaviour

Ongy:
I don't remember, why was this closed? It was intended to fix https://github.com/swaywm/wlroots/issues/676

emersion:
Err, I don't remember either. :\

Ongy:
Rebased and still works.

Issue #674, opened 2018-02-25
mpv’s initial fullscreen is ignored


linkmauve:
Here is the full `WAYLAND_DEBUG` log: [mpv.txt](https://github.com/swaywm/wlroots/files/1756658/mpv.txt)

As we can see, mpv is asking to set fullscreen on line 110, yet on line 125 the configure array is empty.

On line 333 I then pressed f, so on line 334 it asks for fullscreen again, and then on line 342 we can see the value 2 being passed to the configure event.

The only difference between these is that the initial fullscreen request requests a specific output, while the latter doesn’t, but it may be unrelated and a simple race somewhere.

Ongy:
This happens before the first commit.
So we don't have the window yet on the rootston side, but only in the `wlr_xdg_shell_v6` stuff.

The compositor should probably check the toplevel state for the fullscreen state.
To make this work properly we will likely have to move the output into that as well.

linkmauve:
Of course this happens before the first commit, xdg-shell is designed to allow clients to do all of their configuration before the first configure, in order to avoid any race that could happen there.

(Sadly not all of them, scale factor isn’t specified before the surface is mapped, at which point the client has already drawn its first frame.)

PR #681, opened 2018-02-26
wlroots handle protocol request pre first commit

Ongy:
Clients are supposed to set themselves up completly before their first
commit, so no crud gets rendered.
The wlr_shell(_v6) implementations emit the events, but they don't reach
any listener and effectivly get dropped.

With this commit, wlroots will handle the protocol internally (just
allowing clients to do as they please) if the surface hasn't been passed
to the compositor yet.
This requires to keep track of the output a surface is fullscreened on.

The compositor is responsible for checking the state and ensuring it
sets up internal state on new_surface event.

Fixes https://github.com/swaywm/wlroots/issues/674
Testplan: start mpv with --fs argument and watch it appear fullscreened.

### DIFF
--- a/types/wlr_xdg_shell.c
+++ b/types/wlr_xdg_shell.c
@@ -760,6 +760,7 @@ static void xdg_toplevel_protocol_set_fullscreen(struct wl_client *client,
 	}
 
 	surface->toplevel_state->next.fullscreen = true;
+	surface->toplevel_state->next.fullscreen_output = output;
 
 	struct wlr_xdg_toplevel_set_fullscreen_event event = {
 		.surface = surface,
@@ -767,7 +768,11 @@ static void xdg_toplevel_protocol_set_fullscreen(struct wl_client *client,
 		.output = output,
 	};
 
-	wlr_signal_emit_safe(&surface->events.request_fullscreen, &event);
+	if (surface->toplevel_state->added) {
+		wlr_signal_emit_safe(&surface->events.request_fullscreen, &event);
+	} else {
+		wlr_xdg_toplevel_set_fullscreen(surface, true, output);
+	}
 }
 
 static void xdg_toplevel_protocol_unset_fullscreen(struct wl_client *client,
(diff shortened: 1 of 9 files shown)


types/wlr_xdg_shell.c
```diff
@@ -760,14 +760,19 @@ static void xdg_toplevel_protocol_set_fullscreen(struct wl_client *client,
...
 		.surface = surface,
 		.fullscreen = true,
 		.output = output,
 	};
 
-	wlr_signal_emit_safe(&surface->events.request_fullscreen, &event);
+	if (surface->toplevel_state->added) {
+		wlr_signal_emit_safe(&surface->events.request_fullscreen, &event);
+	} else {
+		wlr_xdg_toplevel_set_fullscreen(surface, true, output);
```
emersion: I don't agree with this line. The client sends a _request_ to enter fullscreen, the compositor should be able to deny it. `next` contains the state requested by the client, `current` contains the state applied by the server. Instead of adding this line the compositor could check `next` when a surface appears and call `wlr_xdg_toplevel_set_fullscreen` if it wants to.

Ongy: This event never makes it to the compositor in the current state, which is the core problem.

The client sets itself up to the output size and with fullscreen before it does it's first commit (see the referenced issue).
The compositor currently **can't** attach a listener to this event before the commit and therefore can't grant or deny the request.

This just acks everything until the compositor can intercept, and gives the compositor the state the clients wants to have during the first commit.
Should the compositor decide that the client should not have this state to begin with, it can send the appropriate events (size+unset fullscreen) on the first commit.

Ongy: Erm, forgot the important part:
next is cleared at various points because the `wlr_xdg_shell(_v6)` implementation sets it up to the state in `ack_configure`, so we don't get it on first commit if it's just placed on the next state.

emersion: >This event never makes it to the compositor in the current state, which is the core problem.

No need for an event. Just check `next` in rootston when a surface appears, and send a configure event from there. Don't send configure events without the compositor's consent.

SirCmpwn:
Closing in favor of #711

PR #689, opened 2018-02-27
set fullscreen alpha to 1

Ongy:
Fixes https://github.com/swaywm/wlroots/pull/680#discussion_r170856102

Currently the alpha defaults to 0, which causes fullscreen surfaces to be completly transparent when they have to be rendered by the rootston code and aren't managed by `wlr_output`.

### DIFF
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -523,6 +523,7 @@ static void render_output(struct roots_output *output) {
 	}
 
 	// Render drag icons
+	data.alpha = 1.0;
 	drag_icons_for_each_surface(server->input, render_surface, &data);
 
 renderer_end:
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -470,6 +470,7 @@ static void render_output(struct roots_output *output) {
 		.output = output,
 		.when = &now,
 		.damage = &damage,
+		.alpha = 1.0,
 	};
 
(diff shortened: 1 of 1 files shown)


agx:
Isn't that ensured by https://github.com/swaywm/wlroots/blob/master/types/wlr_output.c#L394 ?

Ongy:
Only if the fullscreen view has a single surface. This isn't the case when any popup exists.

The branch in question is: https://github.com/swaywm/wlroots/blob/master/rootston/output.c#L450 and https://github.com/swaywm/wlroots/blob/master/rootston/output.c#L498

agx:
Ahh..I see, it's one of the side effects of reading this from render_data. Should we do the same after 
    https://github.com/swaywm/wlroots/blob/master/rootston/output.c#L521
since render_view might change it so drag_icons_for_each_surface sees a 1.0f alpha as well?

agx:
LGTM.

rootston/output.c
```diff
@@ -522,6 +523,7 @@ static void render_output(struct roots_output *output) {
 	}
 
 	// Render drag icons
+	data.alpha = 1.0,
```
emersion: `,` → `;`

emersion:
Thanks!

PR #690, opened 2018-02-28
wlr_egl_create_image: return NULL when function is missing

agx:
This matches the return value of elgCreateImage in case of error. Stumbled upon while working on linux-dmabuf.


emersion:
Thanks!

PR #693, opened 2018-03-01
Fix clients binding multiple times to the same output

### LINKED ISSUE #620: Clients can bind to the same wl-output multiple times
We should check that we handle this correctly. One place I'm sure we don't is `wlr_surface_send_done`.

emersion:
Also removes some extraneous assertions.

Fixes #620


agx:
@emersion thanks. I'll try to lurk on IRC more often.

PR #696, opened 2018-03-01
output-layout: fix output global not destroyed

### LINKED ISSUE #695: Unplugging an output and plugging another on the same port reuses the exact same object
If I unplug an output and plug in another one, I get a new geometry/mode/done suite of events, but there has been no notification to the client that a disconnect has happened.  This may cause issues when an application is expecting to be noticed, for example to destroy a surface associated to this now-removed output.

emersion:
Fixes #695


SirCmpwn:
@linkmauve please test

PR #697, opened 2018-03-01
xwm: fix call of xcb_change_window_attributes for old gcc versions

martinetd:
gcc 6.3.0 (at least) complains about &values:
```
expected ‘const uint32_t * {aka const unsigned int *}’ but argument
is of type ‘uint32_t (*)[1] {aka unsigned int (*)[1]}’
```

Reported by thorwil on irc


thorwil:
This solved it.

emersion:
Works for me, thanks!

PR #698, opened 2018-03-01
Add initial linux_dmabuf protocol support

agx:
Tested with ./weston-simple-dmabuf-drm
  
Todo  
- [x]  fix FIXMEs
- [x] style cleanups
- [x] check build without drm headers
- [x] more tests
- [x] drop renderer indirection
- [x] check error handling (logged errors vs. errors reported to clients)
 -later:  look at YUV
Initial feedback would be welcome.


types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+#include <unistd.h>
+#include <wayland-server.h>
+#include <wlr/render.h>
+#include <wlr/types/wlr_linux_dmabuf.h>
+#include <wlr/util/log.h>
+#include "linux-dmabuf-unstable-v1-protocol.h"
+
+static void wl_buffer_destroy(struct wl_client *client,
+	struct wl_resource *resource)
+{
```
emersion: Style nitpick: no newlines between `)` and `{`

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+	if (!resource) {
+		return NULL;
+	}
+
+	if (!wl_resource_instance_of(resource, &wl_buffer_interface,
+			&wl_buffer_impl)) {
+		return NULL;
+	}
+
+	buffer = wl_resource_get_user_data(resource);
```
emersion: * We generally create `<type>_from_resource` helpers to provide better type safety (for instance, https://github.com/swaywm/wlroots/blob/master/types/wlr_gamma_control.c#L28). You'll be able to re-use this helper on line 64.
* Style nitpick: declare variables when they're first used

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+	assert(buffer);
```
emersion: This is not necessary.

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+	for (int i = 0; i < buffer->attributes.n_planes; i++) {
+		close(buffer->attributes.fd[i]);
+		buffer->attributes.fd[i] = -1;
+	}
+	buffer->attributes.n_planes = 0;
+	free(buffer);
+}
+
+static void
+params_destroy(struct wl_client *client, struct wl_resource *resource) {
```
emersion: Style nitpick: no newline between `static void` and the rest of the line

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+	struct wlr_dmabuf_buffer *buffer = wl_resource_get_user_data(params_resource);
+	if (!buffer) {
+		wl_resource_post_error(params_resource,
+			ZWP_LINUX_BUFFER_PARAMS_V1_ERROR_ALREADY_USED,
+			"params was already used to create a wl_buffer");
+		return;
+	}
+
+	assert(buffer->params_resource == params_resource);
+	assert(!buffer->buffer_resource);
```
emersion: These assertions could be moved in the `*_from_resource` helper

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+
+	/* reject unknown flags */
+	if (buffer->attributes.flags & ~ZWP_LINUX_BUFFER_PARAMS_V1_FLAGS_Y_INVERT) {
+		wlr_log(L_ERROR, "Unknown dmabuf flags %"PRIu32, buffer->attributes.flags);
+		goto err_out;
+	}
+
+	/* initial buffer import */
+	if (!wlr_renderer_import_dmabuf(buffer->renderer, buffer))
+		goto err_failed;
```
emersion: Style nitpick: always use brackets

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,431 @@
...
+	/* send 'created' event when the request is not for an immediate
+	 * import, that is buffer_id is zero */
+	if (buffer_id == 0) {
+		zwp_linux_buffer_params_v1_send_created(params_resource,
+			buffer->buffer_resource);
+	}
+	return;
+
+err_failed:
+	if (buffer_id == 0)
```
emersion: Style nitpick: use brackets

render/wlr_renderer.c
```diff
@@ -72,3 +72,17 @@ bool wlr_renderer_format_supported(struct wlr_renderer *r,
 		enum wl_shm_format fmt) {
 	return r->impl->format_supported(r, fmt);
 }
+
+bool wlr_renderer_import_dmabuf(struct wlr_renderer *r,
+	  struct wlr_dmabuf_buffer *dmabuf) {
+	return r->impl->import_dmabuf(r, dmabuf);
```
emersion: Check if this function is NULL, return false if so?

include/wlr/types/wlr_linux_dmabuf.h
```diff
@@ -0,0 +1,42 @@
...
+
+#define MAX_LINUX_BUFFER_PLANES 4
+
+struct wlr_dmabuf_buffer_attribs {
+	/* set via params_add */
+	int n_planes;
+	uint32_t offset[MAX_LINUX_BUFFER_PLANES];
+	uint32_t stride[MAX_LINUX_BUFFER_PLANES];
+	uint64_t modifier[MAX_LINUX_BUFFER_PLANES];
+	int fd[MAX_LINUX_BUFFER_PLANES];
```
emersion: You could maybe instead use a struct for `offset`, `stride`, `modifier` and `fd`?

SirCmpwn:
Looking good so far. Take care to review the style guide, please.

include/wlr/types/wlr_linux_dmabuf.h
ascent12: It's much more consistent for other dmabuf APIs if it stays this way.

ascent12:
I would prefer if this was in terms of wlr_egl instead of wlr_renderer.
All of the gles2 dmabuf functions are just calling EGL functions. It's already capable of importing EGLImages, so we can just leave it at that. wlr_renderer is "supposed" (I don't think it actually is right now) to be an optional type, so I don't want it tangled with any of our other interfaces.

My work on Vulkan will certainly rethinking part of this interface, and its relationship between the renderer, but that work probably won't land any time soon, so it shouldn't be considered a blocker.

agx:
On Fri, Mar 02, 2018 at 11:26:03AM -0800, Scott Anderson wrote:
> I would prefer if this was in terms of wlr_egl instead of
> wlr_renderer.  All of the gles2 dmabuf functions are just calling EGL
> functions. It's already capable of importing EGLImages, so we can just
> leave it at that. wlr_renderer is "supposed" (I don't think it
> actually is right now) to be an optional type, so I don't want it
> tangled with any of our other interfaces.

I only took the detour via wlr_renderer since no other code in types/
used wlr_egl_* directly but wlr_render{,er}_*. If it's o.k. to do so I'm
happy to drop the indirection.

ascent12:
Yeah, the relationship between wlr_egl and wlr_renderer isn't very obvious and not spelled out anywhere.
wlr_egl is geared towards integrating with the platform/backends, while wlr_renderer is just drawing stuff.
Importing dmabufs fits into 'integrating with the platform'.

The idea was that a library user could use wlr_egl with their own renderer code if they wanted to do something more advanced.

agx:
@ascent12 Thanks for spelling this out so clearly. I'll drop the renderer indirection.

rootston/desktop.c
```diff
@@ -729,9 +730,9 @@ struct roots_desktop *desktop_create(struct roots_server *server,
 		wlr_primary_selection_device_manager_create(server->wl_display);
 	desktop->idle = wlr_idle_create(server->wl_display);
 	desktop->idle_inhibit = wlr_idle_inhibit_v1_create(server->wl_display);
-	desktop->linux_dmabuf = wlr_linux_dmabuf_create(
-		server->wl_display, server->renderer);
 
+	struct wlr_gles2_renderer *r = (struct wlr_gles2_renderer *)server->renderer;
```
emersion: You can use `wlr_backend_get_egl` instead of casting the renderer.

include/wlr/render/matrix.h
```diff
@@ -8,6 +8,7 @@ void wlr_matrix_identity(float (*output)[16]);
 void wlr_matrix_translate(float (*output)[16], float x, float y, float z);
 void wlr_matrix_scale(float (*output)[16], float x, float y, float z);
 void wlr_matrix_rotate(float (*output)[16], float radians);
+void wlr_matrix_invert_y(float (*output)[16]);
```
emersion: Seems that this one isn't relevant anymore.

agx:
On Wed, Mar 07, 2018 at 08:50:55AM -0800, emersion wrote:
> emersion commented on this pull request.
> 
> 
> 
> > @@ -8,6 +8,7 @@ void wlr_matrix_identity(float (*output)[16]);
>  void wlr_matrix_translate(float (*output)[16], float x, float y, float z);
>  void wlr_matrix_scale(float (*output)[16], float x, float y, float z);
>  void wlr_matrix_rotate(float (*output)[16], float radians);
> +void wlr_matrix_invert_y(float (*output)[16]);
> 
> Seems that this one isn't relevant anymore.

The commit message wasn't nice either. Thanks for having a look!

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+}
+
+static void linux_dmabuf_create_params(struct wl_client *client,
+		struct wl_resource *linux_dmabuf_resource,
+		uint32_t params_id) {
+	struct wlr_linux_dmabuf *linux_dmabuf = wl_resource_get_user_data(linux_dmabuf_resource);
+
+	uint32_t version = wl_resource_get_version(linux_dmabuf_resource);
+	struct wlr_dmabuf_buffer *buffer = calloc(1, sizeof *buffer);
+	if (!buffer)
```
SirCmpwn: Braces mandatory

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	if ((uint64_t) buffer->attributes.offset[0] + buffer->attributes.stride[0] > UINT32_MAX) {
+		wl_resource_post_error(params_resource,
+			ZWP_LINUX_BUFFER_PARAMS_V1_ERROR_OUT_OF_BOUNDS,
+			"size overflow for plane");
+		goto err_out;
+	}
+
+	if ((uint64_t) buffer->attributes.offset[0] +
```
SirCmpwn: No space after cast

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+	free(linux_dmabuf);
+}
+
+static void handle_display_destroy(struct wl_listener *listener, void *data) {
+	struct wlr_linux_dmabuf *linux_dmabuf = wl_container_of(listener, linux_dmabuf, display_destroy);
+	wlr_linux_dmabuf_destroy(linux_dmabuf);
+}
+
+struct wlr_linux_dmabuf *wlr_linux_dmabuf_create(struct wl_display *display,
+												 struct wlr_egl *egl) {
```
SirCmpwn: Not sure what happened here

SirCmpwn:
Style issues aside this LGTM

agx:
@SirCmpwn thanks. I've cleaned up these style issues plus some more and shuffeled the patches around grouped by functionality.

include/wlr/render.h
```diff
@@ -72,6 +73,14 @@ bool wlr_renderer_read_pixels(struct wlr_renderer *r, enum wl_shm_format fmt,
  */
 bool wlr_renderer_format_supported(struct wlr_renderer *r,
 	enum wl_shm_format fmt);
+
+bool wlr_renderer_import_dmabuf(struct wlr_renderer *r,
+	struct wlr_dmabuf_buffer *dmabuf);
+
+int wlr_renderer_get_dmabuf_formats(struct wlr_renderer *r, int **formats);
+int wlr_renderer_get_dmabuf_modifiers(struct wlr_renderer *r, int format,
+	uint64_t **modifiers);
```
emersion: These three functions don't exist anymore.

include/wlr/types/wlr_linux_dmabuf.h
```diff
@@ -0,0 +1,73 @@
+#ifndef WLR_TYPES_WLR_SIMPLE_DMABUF_H
+#define WLR_TYPES_WLR_SIMPLE_DMABUF_H
+
+#define MAX_LINUX_BUFFER_PLANES 4
```
emersion: This should be prefixed by `WLR_LINUX_DMABUF` since this is exported.

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+	buffer->attributes.offset[plane_idx] = offset;
+	buffer->attributes.stride[plane_idx] = stride;
+	buffer->attributes.modifier[plane_idx] = ((uint64_t)modifier_hi << 32) |
+		modifier_lo;
+	buffer->attributes.n_planes++;
+}
+
+static void
+handle_buffer_destroy(struct wl_resource *buffer_resource)
+{
```
emersion: Style nit: should be all three on the same line

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+
+static const struct zwp_linux_buffer_params_v1_interface linux_buffer_params_impl = {
+	params_destroy,
+	params_add,
+	params_create,
+	params_create_immed,
+};
+
+struct wlr_dmabuf_buffer *wlr_dmabuf_buffer_from_params_resource(
+	struct wl_resource *params_resource) {
```
emersion: Style nit: missing an extra tab here, so we see the difference between the function signature and the function body.

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+	assert(!buffer->buffer_resource);
+	assert(buffer->params_resource == params_resource);
+
+	return buffer;
+}
+
+static void handle_params_destroy(struct wl_resource *params_resource) {
+	struct wlr_dmabuf_buffer *buffer = wl_resource_get_user_data(params_resource);
+	/* Check for NULL since wlr_dmabuf_buffer_from_params_resource will choke */
+	if (!buffer) {
```
emersion: I'd prefer not casting here but rather write:

```c
if (!wl_resource_get_user_data(params_resource)) {
```

So we can't by mistake use the unchecked `struct wlr_dmabuf_buffer *`

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	buffer = wlr_dmabuf_buffer_from_params_resource(params_resource);
+	linux_dmabuf_buffer_destroy(buffer);
+}
+
+static void linux_dmabuf_create_params(struct wl_client *client,
+		struct wl_resource *linux_dmabuf_resource,
+		uint32_t params_id) {
+	struct wlr_linux_dmabuf *linux_dmabuf = wl_resource_get_user_data(linux_dmabuf_resource);
```
emersion: There should be a `wlr_linux_dmabuf_from_resource`

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+
+	uint32_t version = wl_resource_get_version(linux_dmabuf_resource);
+	struct wlr_dmabuf_buffer *buffer = calloc(1, sizeof *buffer);
+	if (!buffer) {
+		goto err;
+	}
+
+	for (int i = 0; i < MAX_LINUX_BUFFER_PLANES; i++)
```
emersion: Style nit: brackets wanted!

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+		buffer->attributes.fd[i] = -1;
+
+	buffer->egl = linux_dmabuf->egl;
+	buffer->params_resource = wl_resource_create(client,
+		&zwp_linux_buffer_params_v1_interface,
+		version, params_id);
+	if (!buffer->params_resource)
```
emersion: Style nit: ditto

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+	wl_resource_post_no_memory(linux_dmabuf_resource);
+}
+
+static void linux_dmabuf_destroy(struct wl_client *client, struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static const struct zwp_linux_dmabuf_v1_interface linux_dmabuf_impl = {
+        linux_dmabuf_destroy,
+        linux_dmabuf_create_params
```
emersion: Style nit: this isn't a tab

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+			modifiers = &modifier_invalid;
+		}
+		for (int j = 0; j < num_modifiers; j++) {
+			uint32_t modifier_lo = modifiers[j] & 0xFFFFFFFF;
+			uint32_t modifier_hi = modifiers[j] >> 32;
+			zwp_linux_dmabuf_v1_send_modifier(resource, formats[i],
+				modifier_hi,
+				modifier_lo);
+		}
+		if (modifiers != &modifier_invalid)
```
emersion: Style nit: brackets

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,435 @@
...
+		  version, id);
+
+	if (resource == NULL) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	wl_resource_set_implementation(resource, &linux_dmabuf_impl,
+		linux_dmabuf, NULL);
+	if (version < ZWP_LINUX_DMABUF_V1_MODIFIER_SINCE_VERSION)
```
emersion: Style nit: brackets

agx:
Besides the style fixes (@emersion thanks for review) I've moved the check for YUV surfaces from gles2_texture_upload_dmabuf to wlr_egl_check_import_dmabuf so the client knows that we can't render the texture instead of just having a warning in the logs

include/wlr/render/egl.h
```diff
@@ -61,6 +64,31 @@ EGLSurface wlr_egl_create_surface(struct wlr_egl *egl, void *window);
 EGLImageKHR wlr_egl_create_image(struct wlr_egl *egl,
 		EGLenum target, EGLClientBuffer buffer, const EGLint *attribs);
 
+/**
+ * Creates an egl image from the given dmabuf attributes. Check usability
+ * of the dmabuf with egl_check_import_dmabuf once first.
```
emersion: Nitpick: should be `with wlr_egl_check_import_dmabuf`

include/wlr/render/egl.h
```diff
@@ -61,6 +64,31 @@ EGLSurface wlr_egl_create_surface(struct wlr_egl *egl, void *window);
...
+ * Try to import the given dmabuf. On success return true false otherwise.
+ * If this succeeds the dmabuf can be used for rendering on a texture
+ */
+bool wlr_egl_check_import_dmabuf(struct wlr_egl *egl,
+		struct wlr_dmabuf_buffer *dmabuf);
+
+/**
+ * Get the available dmabuf formats
+ */
+int wlr_egl_get_dmabuf_formats(struct wlr_egl *egl,	int **formats);
```
emersion: Nitpick: here's an extra tab here

include/wlr/types/wlr_linux_dmabuf.h
```diff
@@ -0,0 +1,80 @@
+#ifndef WLR_TYPES_WLR_SIMPLE_DMABUF_H
+#define WLR_TYPES_WLR_SIMPLE_DMABUF_H
```
emersion: Nitpick: should be `WLR_TYPES_WLR_LINUX_DMABUF_H`

include/wlr/types/wlr_linux_dmabuf.h
```diff
@@ -0,0 +1,80 @@
...
+	struct wl_global *wl_global;
+	struct wl_listener display_destroy;
+	struct wlr_egl *egl;
+};
+
+/**
+ * Create linux-dmabuf interface
+ */
+struct wlr_linux_dmabuf *wlr_linux_dmabuf_create(struct wl_display *display,
+												 struct wlr_egl *egl);
```
emersion: Nitpick: no need to indent the second line with so many tabs/spaces, one is enough.

types/wlr_linux_dmabuf.c
```diff
@@ -0,0 +1,449 @@
...
+static void params_create_common(struct wl_client *client,
+		struct wl_resource *params_resource, uint32_t buffer_id, int32_t width,
+		int32_t height, uint32_t format, uint32_t flags) {
+	struct wlr_dmabuf_buffer *buffer = wlr_dmabuf_buffer_from_params_resource(
+		params_resource);
+	if (!buffer) {
+		wl_resource_post_error(params_resource,
+			ZWP_LINUX_BUFFER_PARAMS_V1_ERROR_ALREADY_USED,
+			"params was already used to create a wl_buffer");
+		return;
```
emersion: This branch is never reached, because `wlr_dmabuf_buffer_from_params_resource` asserts that the buffer isn't NULL.

A fix would be to replace the condition by `if (!wl_resource_get_user_data(params_resource)) {` and move the `wlr_dmabuf_buffer_from_params_resource` call below.

agx:
@emersion thanks for the review. I've fixed that and also plugged a resoure leak in wlr_egl_check_import_dmabuf where we I did not free the eglimage in case of a YUV format, dropped the unused user_data and fixed typo comments.

agx:
I have an additional patch that adds support for multiple planes:

     https://github.com/agx/wlroots/commit/a279202e648588bbe5fb23cc6b88a2b66d3beb24

shall I put that on on top or use a different PR? I can't really test this since I dont have NV12 support but it gets as far as with weston so it shouldn't make things worse.

types/wlr_surface.c
```diff
@@ -325,6 +325,10 @@ static void wlr_surface_apply_damage(struct wlr_surface *surface,
 					surface->current->buffer)) {
 			wlr_texture_upload_drm(surface->texture, surface->current->buffer);
 			goto release;
+		} else if (wlr_dmabuf_buffer_from_buffer_resource(
+					   surface->current->buffer)) {
```
emersion: `wlr_dmabuf_buffer_from_buffer_resource` aborts if the resource isn't a dmabuf buffer or if the buffer is NULL.

You seem to assume here (and in other places too, grep for `wlr_dmabuf_buffer_from_buffer_resource`) that it returns NULL instead of aborting. Maybe a `wlr_dmabuf_resource_is_buffer` could be used instead?

agx:
@emersion a leftover from when the ``wlr_dmabuf_buffer_from_buffer_resource`` didn't have the asserts. ``wlr_dmabuf_resource_is_buffer`` sounds good so I added it.

emersion:
Thanks!

>I have an additional patch that adds support for multiple planes

Let's do this in a new PR, make sure to provide a test plan!

SirCmpwn:
Thanks!

agx:
That was https://github.com/swaywm/wlroots/pull/698/commits/2d0db169422b74e5ef7c07965c376273c7bb2961 in particular. Thanks for fixing this!

PR #699, opened 2018-03-03
Print xcb errors

### LINKED ISSUE #637: check xcb error again
88316e7921942e186df0c6f41fb405fe85bb83f9 (xwm: get rid of xcb call macro) removed most of the _checked() calls, so when there is an error we now get this awesome debug message:

```
2018-02-16 06:11:41 - [xwayland/xwm.c:1015] X11 event: 0
```

[xcb-requests.3](https://www.x.org/archive//X11R7.7/doc/man/man3/xcb-requests.3.xhtml) says event type of 0 means error... So we can cast the event to `xcb_generic_error_t` and get an error code instead, like `3`, which can translate to `window` with a bit of clockwork grease, as well as the opcode of the failed request (as major/minor) which can tell us what failed, like `25`/`0`, which can translate `send_event`, and is the call I messed up to test.

(There's a handy [xcb util-errors](https://cgit.freedesktop.org/xcb/util-errors/) that parses xcb headers at compile time and could help translate these, but it doesn't seem packaged so we probably shouldn't rely on that)

Now, I can't find any way to tell which send_event failed as we use quite a few, but I suppose that'd be a good start for most reports with debug logs -- the question is how to make that decently. As far as I can tell xcb util-errors exists because there is no nice array in xcb headers we could use to programatically tell `send_event` from `25`, we need to manually match against each opcode define like `XCB_SEND_EVENT` until we find one -- there's 120 of these right now even if we don't use all.

What I can do is, add some generic error catching to our event loop, so we get better than what we have; and either print raw values and let folks figure it out or handle one event arbitrarily and print raw values for others and say "write code here if you see that".
Which is better? Any better idea?

martinetd:
There are two possible messages, with the optional dependency it will look like:
`[xwayland/xwm.c:991] xcb error: op ChangeProperty (no minor), code Window (no extension), value 6291465`

Without it:
`[xwayland/xwm.c:999] xcb error: op 18:0, code 3, sequence 103, value 6291465`

The value in case of Window is the window id, so we can tell what function applied on which window which is a good start.
The sequence ought to be able to tell us more precisely which invocation it was, but we never log it when calling functions so is useless in practice and no longer logged.

This message appears naturally when closing urxvt here, so it's easy to test without messing with the code to generate fake errors... We must be calling `xcb_change_property` on the window after it got closed?

Fixes #637.

### DIFF
--- a/include/xwayland/xwm.h
+++ b/include/xwayland/xwm.h
@@ -105,6 +105,9 @@ struct wlr_xwm {
 	struct wl_list unpaired_surfaces; // wlr_xwayland_surface::unpaired_link
 
 	const xcb_query_extension_reply_t *xfixes;
+#ifdef WLR_HAS_XCB_ERRORS
+	xcb_errors_context_t *errors_context;
+#endif
 
 	struct wl_listener compositor_new_surface;
 	struct wl_listener compositor_destroy;
--- a/include/xwayland/xwm.h
+++ b/include/xwayland/xwm.h
@@ -5,6 +5,13 @@
 #include <wlr/xwayland.h>
 #include <xcb/render.h>
 
+#ifdef WLR_HAS_XCB_ICCCM
+	#include <xcb/xcb_icccm.h>
+#endif
+#ifdef WLR_HAS_XCB_ERRORS
+	#include <xcb/xcb_errors.h>
+#endif
+
 enum atom_name {
 	WL_SURFACE_ID,
 	WM_DELETE_WINDOW,
--- a/include/xwayland/xwm.h
+++ b/include/xwayland/xwm.h
@@ -1,5 +1,5 @@
-#ifndef WLR_XWM_H
-#define WLR_XWM_H
+#ifndef XWAYLAND_XWM_H
+#define XWAYLAND_XWM_H
 
 #include <wayland-server-core.h>
 #include <wlr/xwayland.h>
@@ -129,7 +129,7 @@ void xwm_selection_finish(struct wlr_xwm *xwm);
 
 void xwm_set_seat(struct wlr_xwm *xwm, struct wlr_seat *seat);
 
-bool wlr_xwm_atoms_contains(struct wlr_xwm *xwm, xcb_atom_t *atoms,
+bool xwm_atoms_contains(struct wlr_xwm *xwm, xcb_atom_t *atoms,
 		size_t num_atoms, enum atom_name needle);
 
 #endif
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -986,10 +986,10 @@ static void xwm_handle_xcb_error(struct wlr_xwm *xwm, xcb_value_error_t *ev) {
 		goto log_raw;
 	}
 
-	wlr_log(L_ERROR, "xcb error: op %s (%s), code %s (%s), value %"PRIu32,
+	wlr_log(L_ERROR, "xcb error: op %s (%s), code %s (%s), sequence %"PRIu16", value %"PRIu32,
 		major_name, minor_name ? minor_name : "no minor",
 		error_name, extension ? extension : "no extension",
-		ev->bad_value);
+		ev->sequence, ev->bad_value);
 
 	xcb_errors_context_free(err_ctx);
 	return;
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -19,6 +19,9 @@
 #ifdef WLR_HAS_XCB_ICCCM
 	#include <xcb/xcb_icccm.h>
 #endif
+#ifdef WLR_HAS_XCB_ERRORS
+	#include <xcb/xcb_errors.h>
+#endif
 
 const char *atom_map[ATOM_LAST] = {
 	"WL_SURFACE_ID",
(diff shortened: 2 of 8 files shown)


xwayland/xwm.c
```diff
@@ -954,6 +957,50 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
...
+static void xwm_handle_xcb_error(struct wlr_xwm *xwm, xcb_value_error_t *ev) {
+#ifdef WLR_HAS_XCB_ERRORS
+	xcb_errors_context_t *err_ctx;
+	if (xcb_errors_context_new(xwm->xcb_conn, &err_ctx)) {
+		wlr_log(L_DEBUG, "xcb error happened, but could not allocate error context");
+		goto log_raw;
+	}
+
+	const char *major_name;
+	major_name = xcb_errors_get_name_for_major_code(err_ctx, ev->major_opcode);
```
emersion: Style: we generally prefer:

```c
const char *major_name =
    xcb_errors_get_name_for_major_code(err_ctx, ev->major_opcode);
```

martinetd:
I changed my mind about sequence number, it is useful to debug which exact call it is as adding a few print statements is easier than guessing.

The last commit is just here because I could™, we do have quite a few unhandled events and I couldn't be bothered checking individually which number is what, so that might help figuring if we are missing important ones.
There are quite a few calls, I wonder if it'd be worth initializing the `xcb_errors_context_t` once in xwm and reusing it instead of doing alloc/free all the time?

martinetd:
>  There are quite a few calls, I wonder if it'd be worth initializing the xcb_errors_context_t once in xwm and reusing it instead of doing alloc/free all the time?

I've done that, the `xcb_errors_context_new()` function is more than just a few mallocs but also does xcb calls.
Also let me notice we weren't clean with our xwayland headers (WLR_HAS_XCB_ICCCM used in public headers is bad), so this was a good thing.

I'm done messing around, will need a new review when you can please @emersion.

include/xwayland/xwm.h
```diff
@@ -5,6 +5,13 @@
 #include <wlr/xwayland.h>
 #include <xcb/render.h>
 
+#ifdef WLR_HAS_XCB_ICCCM
```
emersion: Can you change `#ifndef WLR_XWM_H` to be `#ifndef XWAYLAND_XWM_H`?

emersion: Also, `wlr_xwm_atoms_contains` should probably be renamed to `xwm_atoms_contains` for consistency (if it is present in the symbols file, it should be removed from there too).

I don't remember why we put `wlr_xwm` in the public headers, it seems to make sense to keep it private. @acrisci comments?

martinetd: Done and done. This was one of the "auto symbol" so will remove itself with name change on rebuild.
I believe the header just got moved together with `xwayland.h` because I had it incorrectly in the source tree at start and @acrisci moved both away, please say if you see a reason for it to be public, I can do some more cleanup.

Ongy: xwm_atoms_contains was intended to be somewhat public.
It's currently a helper exclusivly used for the `is_unmanaged` function, but it could also be used to e.g. detect dzen2 (which will have `NET_WM_WINDOW_TYPE` as `NET_WM_WINDOW_TYPE_DOCK` but I think no more overrdie redirect if called with `-dock` argument). 

It's declared in  `include/wlr/xwm.h` this should change if it's non-public.
Since the headers under `include/wlr` are installed with the normal install, all functions declared there should be public, otherwise it's just confusing.

martinetd: Yes, I moved xwm.h to `include/xwayland/xwm.h` which is private - if this function was meant to be public I'll move it to `xwayland.h` instead of renaming, will amend the last commit.

Ongy: Sounds good. A public function should probably return the `xcb_atom_t` in question, so the value can be read, instead of just being bool.
But this can be done, when we have a usecase.

martinetd: Ok, let's keep this as it is now and either of us can make it public properly later, it's a bit of a bigger change than just moving a line.

SirCmpwn: File a ticket for follow-up please

xwayland/xwm.c
```diff
@@ -1002,6 +1002,33 @@ static void xwm_handle_xcb_error(struct wlr_xwm *xwm, xcb_value_error_t *ev) {
...
+	const char *event_name =
+		xcb_errors_get_name_for_xcb_event(err_ctx, ev, &extension);
+	if (!event_name) {
+		wlr_log(L_DEBUG, "no name for unhandled event: %u",
+			ev->response_type);
+		xcb_errors_context_free(err_ctx);
+		return;
+	}
+
+	wlr_log(L_DEBUG, "unhandled X11 event: %s (%u)", event_name, ev->response_type);
```
psychon: Why not also print the name of the extension that generated the event (`extension`)? (if one is provided, I'd guess this is `NULL` for core X11 events)

martinetd: I just missed it, feel free to add it -- my guess is extension will be null here 99% of the time but will be glad to test any PR you send us.

PR #704, opened 2018-03-05
rootston: Update cursors on set focus

### LINKED ISSUE #678: Alt-tabbing between two clients should draw the cursor of the next one
It currently requires the user to move the pointer to let it get drawn by the next client.  Before doing so, the last cursor from the previous client continues to be displayed at its last location, even if it should otherwise be hidden.

martinetd:
Setting focus can change the views ordering so all cursors need to
be updated to reflect such changes e.g. edges are different.

This is a bigger issue than I thought it'd be.
I've just hooked in `roots_seat_set_focus` and call `roots_cursor_update_position` at the end for all seats, but that's just shoving the dust under the rug - multiseat and cursor as we do them is difficult.
For example what if a seat user moves a window just under another cursor?

As a bonus question, I don't like calling `clock_gettime` and `timespec_to_msec`. Technically I think we could get a time from the event most of the time, e.g. for next_window we have the keyboard event that triggered the key binding all the way up the stack, should I bother and bubble it down?
The problem with this approach is there are two places calling `roots_seat_set_focus` without an `event->time_msec` handy - `view_initial_focus` and `seat_view_destroy`, which makes two places calling `clock_gettime` instead of one...

test plan:
* [x] Go to edge of a view with another view below and alt-tab, check cursor changes both times
* [x] Place an edge somewhere in the middle where a new view would pop and check the new view correctly updates cursor when it appears
* [x] Check fullscreening a view and unfullscreening updates cursor (just noticed it doesn't work right now)
* [ ] Check 2nd seat's cursor gets updated when resizing a window under it
* [ ] . . .but not if the resized view is below another

Fixes #678.

### DIFF
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -98,7 +98,7 @@ static void seat_view_deco_button(struct roots_seat_view *view, double sx,
 	}
 }
 
-static void roots_cursor_update_position(struct roots_cursor *cursor,
+void roots_cursor_update_position(struct roots_cursor *cursor,
 		uint32_t time) {
 	struct roots_desktop *desktop = cursor->seat->input->server->desktop;
 	struct roots_seat *seat = cursor->seat;
@@ -134,7 +134,9 @@ static void roots_cursor_update_position(struct roots_cursor *cursor,
 		} if (view && surface) {
 			// motion over a view surface
 			wlr_seat_pointer_notify_enter(seat->seat, surface, sx, sy);
-			wlr_seat_pointer_notify_motion(seat->seat, time, sx, sy);
+			if (time) {
+				wlr_seat_pointer_notify_motion(seat->seat, time, sx, sy);
+			}
 		} else {
 			wlr_seat_pointer_clear_focus(seat->seat);
 		}
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -120,6 +120,32 @@ static void view_update_output(const struct roots_view *view,
 	}
 }
 
+void view_update_cursors(const struct roots_view *view,
+		const struct wlr_box *before) {
+	struct roots_desktop *desktop = view->desktop;
+
+	if (view->wlr_surface == NULL) {
+		return;
+	}
+
+	struct wlr_box box;
+	view_get_deco_box(view, &box);
+
+	struct roots_seat *seat;
+	wl_list_for_each(seat, &desktop->server->input->seats, link) {
+		if (seat->cursor->mode != ROOTS_CURSOR_PASSTHROUGH) {
+			continue;
+		}
+		bool in_before = before != NULL && wlr_box_contains_point(before,
+			seat->cursor->cursor->x, seat->cursor->cursor->y);
+		bool in_now = wlr_box_contains_point(&box,
+			seat->cursor->cursor->x, seat->cursor->cursor->y);
+		if (in_before != in_now) {
+			roots_cursor_update_position(seat->cursor, 0);
+		}
+	}
+}
+
 void view_move(struct roots_view *view, double x, double y) {
 	if (view->x == x && view->y == y) {
 		return;
@@ -133,6 +159,7 @@ void view_move(struct roots_view *view, double x, double y) {
 		view_update_position(view, x, y);
 	}
 	view_update_output(view, &before);
+	view_update_cursors(view, &before);
 }
 
 void view_activate(struct roots_view *view, bool activate) {
@@ -148,6 +175,7 @@ void view_resize(struct roots_view *view, uint32_t width, uint32_t height) {
 		view->resize(view, width, height);
 	}
 	view_update_output(view, &before);
+	view_update_cursors(view, &before);
 }
 
 void view_move_resize(struct roots_view *view, double x, double y,
@@ -499,6 +527,7 @@ void view_setup(struct roots_view *view) {
 	}
 
 	view_update_output(view, NULL);
+	view_update_cursors(view, NULL);
 }
 
 void view_apply_damage(struct roots_view *view) {
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -151,15 +151,11 @@ void view_move(struct roots_view *view, double x, double y) {
 		return;
 	}
 
-	struct wlr_box before;
-	view_get_box(view, &before);
 	if (view->move) {
 		view->move(view, x, y);
 	} else {
 		view_update_position(view, x, y);
 	}
-	view_update_output(view, &before);
-	view_update_cursors(view, &before);
 }
 
 void view_activate(struct roots_view *view, bool activate) {
@@ -169,13 +165,9 @@ void view_activate(struct roots_view *view, bool activate) {
 }
 
 void view_resize(struct roots_view *view, uint32_t width, uint32_t height) {
-	struct wlr_box before;
-	view_get_box(view, &before);
 	if (view->resize) {
 		view->resize(view, width, height);
 	}
-	view_update_output(view, &before);
-	view_update_cursors(view, &before);
 }
 
 void view_move_resize(struct roots_view *view, double x, double y,
@@ -549,10 +541,14 @@ void view_update_position(struct roots_view *view, double x, double y) {
 		return;
 	}
 
+	struct wlr_box before;
+	view_get_box(view, &before);
 	view_damage_whole(view);
 	view->x = x;
 	view->y = y;
 	view_damage_whole(view);
+	view_update_output(view, &before);
+	view_update_cursors(view, &before);
 }
 
 void view_update_size(struct roots_view *view, uint32_t width, uint32_t height) {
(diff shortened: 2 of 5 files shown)


martinetd:
Hmm that build failure doesn't really look like it's my fault, but I'm not sure why that worked when @SirCmpwn changed it. I wonder how he clones the repo, need to fetch refs/pull/x/head?
Anyway, this PR is half WIP, can give him some breathing room :)

rootston/seat.c
```diff
@@ -762,6 +768,14 @@ void roots_seat_set_focus(struct roots_seat *seat, struct roots_view *view) {
 		wlr_seat_keyboard_notify_enter(seat->seat, view->wlr_surface,
 			NULL, 0, NULL);
 	}
+
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	struct roots_seat *all_seat;
+	wl_list_for_each(all_seat, &seat->input->seats, link) {
```
Ongy: This feels a little out of place here.
It's fitting to call it during `roots_seat_set_focus`, but it's logically part of the first few lines in the function (bring to top) and should probably be closer to that (maybe in a dedicated function?)

martinetd: I added it at the end because we do the same with keyboard enter just above, but happy to move it up - it only needs to be below `wl_list_insert(&seat->input->server->desktop->views, &view->link)`

Actually now that I look at it again it needs to be just after that and before the next prev = view check, because this can bubble a view that was already focused up.
I'll put this in its own function and move it there after work, I'm laaaate :D

martinetd:
I moved things quite a bit, changing how I looked at that: basically z-axis change is when we touch at desktop->views, so made helpers to manipulate this list and refresh cursors.

It works for alt-tab, but doesn't work reliably for new/removed views, I'm not 100% sure why.
I think we might add new views before they have a size/position, so new view can't be handled at this time and I don't really want to recheck all cursors everytime a view moves (so on that end I'm tempted to say that's as far as rootston will get from my efforts), but I don't see why removing view isn't reliable so I'd like to dig a bit more there. Will continue tomorrow when I find time.

rootston/wl_shell.c
```diff
@@ -243,7 +243,7 @@ void handle_wl_shell_surface(struct wl_listener *listener, void *data) {
 	view->close = close;
 	roots_surface->view = view;
 	view_init(view, desktop);
-	wl_list_insert(&desktop->views, &view->link);
+	desktop_add_view(desktop, view);
```
Ongy:
```
desktop_add_view
view_setup
```

I would look at this first to get the insertion right.

martinetd: Hmm, good point, hadn't noticed `view_setup` because I was mostly looking at removal but that will explain what I said about insertion. I notice this is calling `view_update_output`, that's also called on view move, if I can do something a bit less naive I'll change `desktop_add_view` to `view_update_cursors`.

Obviously I can't reproduce the problem I seemed to have with removing a view yesterday, so I guess I'll start with that...
For the record, we also get "fun" behaviour if a resize is active (logo+right click or edge drag) and alt-tab happens. It's not as bad as I had expected but I'm not sure how we'll want to handle that, fun stuff.

SirCmpwn:
What's the status of this?

martinetd:
I need to do fix it but got caught up with other things, should have some time to finish over next weekend. Happy to temporarily close or set title to WIP if that helps

rootston/cursor.c
```diff
@@ -134,7 +134,9 @@ static void roots_cursor_update_position(struct roots_cursor *cursor,
 		} if (view && surface) {
 			// motion over a view surface
 			wlr_seat_pointer_notify_enter(seat->seat, surface, sx, sy);
-			wlr_seat_pointer_notify_motion(seat->seat, time, sx, sy);
+			if (time) {
```
martinetd: I don't really like this, but it feels better than asking for time and notifying for motion that hasn't really happened anyway (we're moving a window, not the pointer).
Might be able to rip part of this function off to share code instead of hijacking it.

martinetd:
hmm. Fullscreening doesn't work because `view_update_cursors` called in resize on fullscreening is still considering the old x/y/width/height (the fullscreened values are in `pending_move_resize`).
I just monkeyed `view_update_output` but would need to call it in view_update_size/position instead?

For output and fullscreening it doesn't matter (it's still within the same output), but if you manage to resize in one big mouse event you would also miss the output update, so I think that'd need to move both to `view_update_size` and `view_update_position` that get the correct x/y? Or make that a responsibility of each shell handlers (`handle_surface_commit`) so it's only called once per commit instead?

martinetd:
>  I just monkeyed view_update_output but would need to call it in view_update_size/position instead?

I did that, works better for fullscreening, and still works for the rest I tested.
Still can't test multi-pointer unfortunately, will do for sure over the weekend.

rootston/desktop.c
```diff
@@ -120,19 +120,48 @@ static void view_update_output(const struct roots_view *view,
...
+	wl_list_for_each(seat, &desktop->server->input->seats, link) {
+		if (seat->cursor->mode != ROOTS_CURSOR_PASSTHROUGH) {
+			continue;
+		}
+		bool in_before = before != NULL && wlr_box_contains_point(before,
+			seat->cursor->cursor->x, seat->cursor->cursor->y);
+		bool in_now = wlr_box_contains_point(&box,
+			seat->cursor->cursor->x, seat->cursor->cursor->y);
+		if (before) {
+                       wlr_log(L_ERROR, "in_before: %d, %fx%f ∈ %dx%d+%dx%d\n"
```
acrisci: Need tabs instead of spaces here and 2 lines below.

martinetd: Bleh, didn't mean to push that at all, sorry. will force push in a min.

acrisci:
> For example what if a seat user moves a window just under another cursor?

Does the PR intend to handle this? Because it doesn't currently.

acrisci:
> Go to edge of a view with another view below and alt-tab, check cursor changes both times

This doesn't work correctly in this PR for me on the DRM output (tested single seat and multiseat).

acrisci:
Another bug:

Grab the window titlebar (interactive move) and it has the hand-grab icon. Close the window with the button pressed and no window behind it. The cursor should revert to the compositor cursor, but it remains as the hand-grab.

acrisci:
> Check fullscreening a view and unfullscreening updates cursor (just noticed it doesn't work right now)

This is really flaky. It seems to work for some parts of the window with gnome-terminal and not work for others. It works for the i-beam but not for the resize handles.

martinetd:
>> Check fullscreening a view and unfullscreening updates cursor
>
> This is really flaky. It seems to work for some parts of the window with gnome-terminal and not work for others. It works for the i-beam but not for the resize handles.

Hm this is tricky - it forks fine on fullscreening, but on unfullscreening (once I found some part of the window that isn't moving compared to where it was pre-fullscreening) it indeed doesn't restore the resize "grab"... The cursor is out of the surface so I do call `roots_cursor_update_position` in that case, but I don't see why it's not reset to resize cursor -- heck, I don't see why it's set in the first place so that'll be fun to debug...

>> Go to edge of a view with another view below and alt-tab, check cursor changes both times
>
> This doesn't work correctly in this PR for me on the DRM backend with gnome-terminal (tested single seat and multiseat). It works on the first alt-tab but not the second one.

This is the exact same problem as unfullscreening, it works for me on top/left sides but not bottom/right sides.
I need to understand how these 5px below the window are supposed to be a resize area first, `desktop_view_at`/`view_at` should fall through as there is no surface there...

> Grab the window titlebar (interactive move) and it has the hand-grab icon. Close the window with the button pressed and no window behind it. The cursor should revert to the compositor cursor, but it remains as the hand-grab.

I decided to ignore change of focuses (alt tabbing for example) while resizing. It'll resize the new window instead with some weird side effects but looking for which cursor might currently be resizing the previous view seemed too much work for how often that'll happen.
I'm not sure about closing though, that might be easier to fix, will have a look.

>> For example what if a seat user moves a window just under another cursor?
>
> Does the PR intend to handle this? Because it doesn't currently.
> 
> It seems to work somewhat when the cursor is already under the view, but not if it is not under the view before the move.

I think it checks what needs to be checked, but can't test until this weekend. I don't understand what you mean by "cursor is under the view", a cursor is always on top? Do you mean the view it's trying to resize, so a difference between newly resized window changing focus order at the start of the interactive resize and resized window already being on top?

rootston/cursor.c
emersion: Yeah, what we're trying to do here is faking input events, and it doesn't work well because we don't know the time reference. I'm still thinking we should send pointer motion events, but I'm not sure how to fix this. We could try to get the time from the seat that triggered the update, but nothing guarantees that all input devices are from the same backend and have the same time reference.

martinetd: I also originally tried that but we don't always have an input event related to the call. Sometimes it's a key event (e.g. alt-tab) and we have one, but when a new window pops in we don't have anything :/

I haven't totally given up on this PR btw but really won't have time for another while, sorry

martinetd:
Closing for now, I had totally forgotten about it and there were too many flaky edge cases ; might or might not take it up again at some point but it's easy enough to just reopen when/if I do.

PR #707, opened 2018-03-06
Xcb fixes

### LINKED ISSUE #706: x11 backend: Broken window close handling
I am not quite sure what the intention behind the following code is, but I am sure that it does not work:

https://github.com/swaywm/wlroots/blob/9716aa9b9247e7af08f4887fec494dcb9a918d9d/backend/x11/backend.c#L147-L150

`XCB_GLX_DELETE_QUERIES_ARB` is the opcode for the request that `xcb_glx_delete_queries_arb()` sends/generates. It is not an opcode for an event.

Since this define has value 161, which is in the space for extension events, this is just a time bomb: If the x11 backend ever starts using some x11 extension that generates events, eventually someone will run an x11 server where this extension's events ends up being assigned number 161, which means that wlroots now handles whatever this event is by quitting.

Since this code comes from commit 6bf508df816750f41b9b0c362f5bc6f64f737a70, let me CC @ascent12 and quote ICCCM §4.2.8.1.

> Clients, usually those with multiple top-level windows, whose server connection must survive the deletion of some of their top-level windows, should include the atom WM_DELETE_WINDOW in the WM_PROTOCOLS property on each such window. They will receive a ClientMessage event as described above whose data[0] field is WM_DELETE_WINDOW.

So, the code for quitting should be under `case XCB_CLIENT_MESSAGE`, it should cast the event to `xcb_client_message_event_t *` and check if the event has `->data.data32[0] == x11->atoms.wm_delete_window.reply->atom`.

(To whoever writes a patch for this: I suggest a `puts("This works");` in the newly-added `case` for testing)

(Also, please get rid of `#include <xcb/glx.h>` while you are at it)

(Oh and: Why does `struct wlr_x11_backend` contain an `xcb_intern_atom_cookie_t` and an `xcb_intern_atom_reply_t` when all that you want is the `xcb_atom_t` that is saved as `->reply->atom`? Plus, the content of the reply field is just plain leaked instead of being `free()`d on shutdown)

(The symptom of what the code currently does should be that closing the window just does not work at all... either I am wrong here or how come no one noticed?)

Edit: Okay, the current code actually does work correctly... kind of. What I forgot is that the highest bit of an event indicates that it was generated by `xcb_send_event`. Now, `XCB_CLIENT_MESSAGE` is 33 and 33+128 is 161, so we end up with the value of `XCB_GLX_DELETE_QUERIES_ARB`. However, all of this is just accidential and not really the right way to handle client messages.

So: Instead of `switch (event->response_type)` you need `switch (event->response_type & 0x7f)`, or for readability `switch (XCB_EVENT_RESPONSE_TYPE(event))` (this macro is from `xcb_event.h`).

ascent12:
Fixes #706

### DIFF
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -26,9 +26,9 @@
 
 #define XCB_EVENT_RESPONSE_TYPE_MASK 0x7f
 
-static struct wlr_backend_impl backend_impl;
-static struct wlr_output_impl output_impl;
-static struct wlr_input_device_impl input_device_impl = { 0 };
+static const struct wlr_backend_impl backend_impl;
+static const struct wlr_output_impl output_impl;
+static const struct wlr_input_device_impl input_device_impl = { 0 };
 
 static uint32_t xcb_button_to_wl(uint32_t button) {
 	switch (button) {
@@ -337,7 +337,7 @@ static struct wlr_renderer *wlr_x11_backend_get_renderer(
 	return x11->renderer;
 }
 
-static struct wlr_backend_impl backend_impl = {
+static const struct wlr_backend_impl backend_impl = {
 	.start = wlr_x11_backend_start,
 	.destroy = wlr_x11_backend_destroy,
 	.get_egl = wlr_x11_backend_get_egl,
@@ -465,7 +465,7 @@ static bool output_swap_buffers(struct wlr_output *wlr_output,
 	return wlr_egl_swap_buffers(&x11->egl, output->surf, damage);
 }
 
-static struct wlr_output_impl output_impl = {
+static const struct wlr_output_impl output_impl = {
 	.set_custom_mode = output_set_custom_mode,
 	.transform = output_transform,
 	.destroy = output_destroy,
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -15,7 +15,6 @@
 #include <wlr/render/gles2.h>
 #include <wlr/util/log.h>
 #include <X11/Xlib-xcb.h>
-#include <xcb/glx.h>
 #include <xcb/xcb.h>
 #ifdef __linux__
 #include <linux/input-event-codes.h>
@@ -24,6 +23,8 @@
 #endif
 #include "backend/x11.h"
 #include "util/signal.h"
+
+#define XCB_EVENT_RESPONSE_TYPE_MASK 0x7f
 
 static struct wlr_backend_impl backend_impl;
 static struct wlr_output_impl output_impl;
@@ -44,7 +45,7 @@ static uint32_t xcb_button_to_wl(uint32_t button) {
 static bool handle_x11_event(struct wlr_x11_backend *x11, xcb_generic_event_t *event) {
 	struct wlr_x11_output *output = &x11->output;
 
-	switch (event->response_type) {
+	switch (event->response_type & XCB_EVENT_RESPONSE_TYPE_MASK) {
 	case XCB_EXPOSE: {
 		wlr_output_send_frame(&output->wlr_output);
 		break;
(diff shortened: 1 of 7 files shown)


backend/x11/backend.c
```diff
@@ -15,8 +15,8 @@
 #include <wlr/render/gles2.h>
 #include <wlr/util/log.h>
 #include <X11/Xlib-xcb.h>
-#include <xcb/glx.h>
 #include <xcb/xcb.h>
+#include <xcb/xcb_event.h>
```
psychon: I've been reading some of your code and found the following in `xwayland/xwm.c`:
```
/* This is in xcb/xcb_event.h, but pulling xcb-util just for a constant
 * others redefine anyway is meh
 */
#define XCB_EVENT_RESPONSE_TYPE_MASK (0x7f)
```

backend/x11/backend.c
```diff
@@ -231,20 +229,58 @@ static bool wlr_x11_backend_start(struct wlr_backend *backend) {
...
+	}
+
+	for (size_t i = 0; i < sizeof(atom) / sizeof(atom[0]); ++i) {
+		xcb_intern_atom_reply_t *reply = xcb_intern_atom_reply(
+			x11->xcb_conn, atom[i].cookie, NULL);
+
+		if (reply) {
+			*atom[i].atom = reply->atom;
+			free(reply);
+		}
```
psychon: I think this should have an `else *atom[i].atom = XCB_ATOM_NONE` (or does something else initialise these variables to 0?)

backend/x11/backend.c
```diff
@@ -231,20 +229,58 @@ static bool wlr_x11_backend_start(struct wlr_backend *backend) {
...
+			.name = "_NET_WM_NAME",
+			.only_if_exists = true,
+			.atom = &x11->atoms.net_wm_name,
+		},
+		{
+			.name = "UTF8_STRING",
+			.only_if_exists = false,
+			.atom = &x11->atoms.utf8_string,
+		},
+	};
```
psychon: Unrelated to this commit, but: Why is `only_if_exists` not just always set to `true`?

backend/x11/backend.c
```diff
@@ -44,7 +44,7 @@ static uint32_t xcb_button_to_wl(uint32_t button) {
 static bool handle_x11_event(struct wlr_x11_backend *x11, xcb_generic_event_t *event) {
 	struct wlr_x11_output *output = &x11->output;
 
-	switch (event->response_type) {
+	switch (XCB_EVENT_RESPONSE_TYPE(event)) {
```
psychon: Not my call (I have nothing to do with wlroots), but I'd say that this change does not belong into this commit. The commit message is about atoms.

psychon:
Oh and: Thanks for handling this so quickly.

backend/x11/backend.c
ascent12: For some reason, I thought we were pulling in xcb-util, but I guess I was mistaken.
Yeah, it's really not worth it just for that.

backend/x11/backend.c
ascent12: We allocate most things with calloc, but I suppose that would be more obvious for people reading this code.

backend/x11/backend.c
martinetd: Right, I had forgotten I did that myself, we need to either add the dependency to meson or do the same

backend/x11/backend.c
martinetd: This will be set to 0 by the way the struct is filled, non-specified fields in { .foo = bar, etc } are initialized to 0

backend/x11/backend.c
ascent12: Most of this code was copied from some pretty poor examples. It was probably there originally as some portability to some ancient X server that nobody uses any more.

backend/x11/backend.c
martinetd: Sorry for double-replies, for some reason github didn't update with your comments when I logged in here.

Issue #705, opened 2018-03-05
Rootston: Gksudo


rolfen:
Launching `gksudo` or `gksu` will kill rootston.

emersion:
Can you dump the stack trace?

rolfen:
How do I do that, please?

emersion:
1. `gdb build/rootston/rootston`
2. `run`
3. Make rootston crash
4. `bt`

rolfen:
Thank you for your instructions.
`Program received signal SIGSEGV, Segmentation fault.`
```
#0  0x00007ffff795d327 in wl_list_remove () at /usr/lib/x86_64-linux-gnu/libwayland-server.so.0
#1  0x0000555555564ec9 in roots_seat_set_focus (seat=0x555555e2c980, view=0x5555561035c0) at ../rootston/seat.c:713
#2  0x0000555555564d02 in seat_view_destroy (seat_view=0x55555609cf60) at ../rootston/seat.c:656
#3  0x0000555555564d32 in seat_view_handle_destroy (listener=0x55555609cf98, data=0x5555561114d0) at ../rootston/seat.c:663
#4  0x000055555555c646 in wl_signal_emit (signal=0x5555561115a8, data=0x5555561114d0) at /usr/include/wayland-server-core.h:388
#5  0x000055555555d79f in view_finish (view=0x5555561114d0) at ../rootston/desktop.c:407
#6  0x00005555555688a2 in handle_destroy (listener=0x555556097898, data=0x5555561266a0) at ../rootston/xwayland.c:124
#7  0x00007ffff7bb7083 in wlr_signal_emit_safe (signal=0x555556126780, data=0x5555561266a0) at ../util/signal.c:29
#8  0x00007ffff7b7f315 in wlr_xwayland_surface_destroy (xsurface=0x5555561266a0) at ../xwayland/xwm.c:216
#9  0x00007ffff7b8015a in xwm_handle_destroy_notify (xwm=0x555556099ee0, ev=0x5555561045b0) at ../xwayland/xwm.c:601
#10 0x00007ffff7b80e8b in x11_event_handler (fd=12, mask=1, data=0x555556099ee0) at ../xwayland/xwm.c:1036
#11 0x00007ffff795ac12 in wl_event_loop_dispatch () at /usr/lib/x86_64-linux-gnu/libwayland-server.so.0
[...]

Issue #679, opened 2018-02-25
Segfault or damage tracking issue when displaying a tooltip in Firefox and alt-tabbing to another client


linkmauve:
How to reproduce:
- Go to a page where a tooltip can be displayed by Firefox.
- Put your pointer on the element that will generate the tooltip.
- Alt-Tab to another client before the tooltip is displayed.

In some cases the tooltip will keep being drawn on the other client, until the cursor moves on top of it; in other cases it will vanish properly; in yet other cases rootston will segfault (but I don’t have a gdb trace yet).

This may be related to #678.

Ongy:
I'm more inclined to connect it to https://github.com/swaywm/wlroots/issues/671 and https://github.com/swaywm/wlroots/issues/650 in terms of segfault.

The tooltip issue is something I have seen on waymonad when pointer focus went from firefox to a native wayland client. I had blamed it on Xwayland keeping the pointer where it is, once it lost focus.

How does weston behave in that case?

linkmauve:
AFAIK Weston still displays the tooltip on top of the next client, but removes it on the first interaction with it.

Feel free to report that issue to xorg-xserver instead.

Issue #671, opened 2018-02-25
rootston crash on focus change with xwayland popup


Ongy:
Reproduce (may be overly specific):
* Open firefox and termite in rootston
* ~~Place firefox and termite in a way, such that right click menu of firefox is over termite~~ Not required
* Right click on firefox
* Hover the menu and [Alt]+[Tab] to change focus to termite
* [Alt]+[Tab] back to firefox

```
           PID: 4591 (rootston)
           UID: 1000 (ongy)
           GID: 1000 (ongy)
        Signal: 11 (SEGV)
     Timestamp: Sun 2018-02-25 10:08:21 CET (30s ago)
  Command Line: /home/ongy/Source/wlroots/build/rootston/rootston -E termite
    Executable: /home/ongy/Source/wlroots/build/rootston/rootston
 Control Group: /user.slice/user-1000.slice/session-c1.scope
          Unit: session-c1.scope
         Slice: user-1000.slice
       Session: c1
     Owner UID: 1000 (ongy)
       Boot ID: 0146d6ac264047c6ad34c6f2a2211f28
    Machine ID: e47085c8d34048d7b5485047dcf6fdeb
      Hostname: link
       Storage: /var/lib/systemd/coredump/core.rootston.1000.0146d6ac264047c6ad34c6f2a2211f28.4591.1519549701000000.lz4
       Message: Process 4591 (rootston) of user 1000 dumped core.
                
                Stack trace of thread 4591:
                #0  0x00007f5f60343137 wl_list_remove (libwayland-server.so.0)
                #1  0x00005570d8e1e4fa n/a (/home/ongy/Source/wlroots/build/rootston/rootston)
                #2  0x00005570d8e1e76d n/a (/home/ongy/Source/wlroots/build/rootston/rootston)
                #3  0x00005570d8e1898a n/a (/home/ongy/Source/wlroots/build/rootston/rootston)
                #4  0x00005570d8e18dbc n/a (/home/ongy/Source/wlroots/build/rootston/rootston)
                #5  0x00005570d8e18f96 n/a (/home/ongy/Source/wlroots/build/rootston/rootston)
                #6  0x00005570d8e1c5db n/a (/home/ongy/Source/wlroots/build/rootston/rootston)
                #7  0x00007f5f6059eb27 n/a (/home/ongy/Source/wlroots/build/libwlroots.so.0.0.0)
[...]

Issue #650, opened 2018-02-19
rootston: Segfault on closing some XWayland windows with opened drop down menu


chebykinn:
__Commit:__ 868ad5af69583646ba71d6e73e362818c7416941
__Reproduction:__ Open some X11 application, for instance vlc, and exit via drop down menu.
__Result:__
```
Thread 1 "rootston" hit Breakpoint 1, roots_seat_set_focus (seat=0x561f7c9d7d00, view=0x561f7cd2fe00) at ../rootston/seat.c:713
713                     wl_list_remove(&view->link);
1: view = (struct roots_view *) 0x561f7cd2fe00
2: view->link = {prev = 0x0, next = 0x0}
(gdb) c
Continuing.
Thread 1 "rootston" received signal SIGSEGV, Segmentation fault.
0x00007f9d7e607137 in wl_list_remove () from /usr/lib/libwayland-server.so.0
(gdb) bt
#0  0x00007f9d7e607137 in wl_list_remove () at /usr/lib/libwayland-server.so.0
#1  0x0000561f7b178265 in roots_seat_set_focus (seat=0x561f7c9d7d00, view=0x561f7cd2fe00) at ../rootston/seat.c:713
#2  0x0000561f7b17809e in seat_view_destroy (seat_view=0x561f7ccc0ce0) at ../rootston/seat.c:656
#3  0x0000561f7b1780ce in seat_view_handle_destroy (listener=0x561f7ccc0d18, data=0x561f7ccedf60) at ../rootston/seat.c:663
#4  0x0000561f7b16f819 in wl_signal_emit (signal=0x561f7ccee038, data=0x561f7ccedf60) at /usr/include/wayland-server-core.h:388
#5  0x0000561f7b1709ba in view_finish (view=0x561f7ccedf60) at ../rootston/desktop.c:387
#6  0x0000561f7b17bdca in handle_destroy (listener=0x561f7cd48328, data=0x561f7cb3a7a0) at ../rootston/xwayland.c:124
#7  0x00007f9d7e86278f in wlr_signal_emit_safe (signal=0x561f7cb3a880, data=0x561f7cb3a7a0) at ../util/signal.c:29
#8  0x00007f9d7e829bb8 in wlr_xwayland_surface_destroy (xsurface=0x561f7cb3a7a0) at ../xwayland/xwm.c:215
[...]

PR #708, opened 2018-03-06
emit view->events.destroy on xwayland unmap

Ongy:
Emits the view destroy signal when an xwayland view is unmaped.
This makes sure, no references to it are held, and fixes a few
situations where inconsistent view state could lead to segfault.

This fixes the segfaults, but I am honestly not sure, if it could introduce other issues.
Quick testing doesn't expose anything obvious, but there may be more complicated issues.

Relevant issues:
https://github.com/swaywm/wlroots/issues/705
https://github.com/swaywm/wlroots/issues/679
https://github.com/swaywm/wlroots/issues/671
https://github.com/swaywm/wlroots/issues/650

The `view_finish` in `desktop.c` uses the same libwayland signal emit.
Should we change that (in another PR)?

### DIFF
--- a/rootston/xwayland.c
+++ b/rootston/xwayland.c
@@ -276,6 +276,8 @@ static void handle_unmap_notify(struct wl_listener *listener, void *data) {
 	view->wlr_surface = NULL;
 	view->width = view->height = 0;
 	wl_list_remove(&view->link);
+
+	wl_signal_emit(&view->events.destroy, view);
 }
 
 void handle_xwayland_surface(struct wl_listener *listener, void *data) {


emersion:
>This fixes the segfaults, but I am honestly not sure, if it could introduce other issues.

I think `handle_map_notify` needs a call to `view_initial_focus`, otherwise `seat_add_view` is not called and the view is not inserted in `seat->views`.

>The view_finish in desktop.c uses the same libwayland signal emit.

If we don't get issues with destruct listeners removing the next listener, I don't think this is necessary. Weston uses `wl_signal_emit` without issues.

acrisci:
My first impression is that this is an abuse of the destroy signal because it gets emitted without the actual surface being destroyed.

Ongy:
This was pretty much a bandaid level PR to reduce the crashyness, and is discontinued in favor of https://github.com/swaywm/wlroots/pull/711#issuecomment-372485427
I should probably just close it.

That PR gives us explicit map/unmap to be handled by the compositor, so this little hack isn't required.

Issue #581, opened 2018-01-23
Cursor jumps  on decoration border enter in DRM backend


acrisci:
Introduced in #570 with decoration support.

To reproduce:

Enter a decoration border. In the first few pixels of entry, the cursor resize edge xcursor will appear and upon the next move, it will "jump" to (I think) what is the correct position.

Issue is not present on the X11 or Wayland backend.

emersion:
Can reproduce:
* Enter a surface with CSD
* Click on the titlebar without moving, the cursor changes to be a hardware one but has an invalid position (because of the hotspot)
* Move a little, the cursor jumps and now has the correct position

This happens because `crtc_move_cursor` needs to be called in `wlr_drm_connector_set_cursor` when the hotspot changes.

PR #710, opened 2018-03-07
output: don't move hidden cursors

emersion:
When using the DRM backend and multiple outputs, that reduces the
number of commits.

Test plan: check that cursors aren't broken on a multi-monitor setup

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -544,10 +544,10 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 	if (!crtc) {
 		return false;
 	}
-	struct wlr_drm_plane *plane = crtc->cursor;
 
-	// We don't have a real cursor plane, so we make a fake one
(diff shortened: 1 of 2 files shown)


Ongy:
On startup rootston sets the cursor on the top left corner of the output.
With this, that only changes once the pointer enters the relevant output, which makes this behaviour look rather weird.
It should probably set up the cursor in the virtual layout properly at startup.


It also updates all outputs when a client cursor is set.
Can we just upload the new texture without updating the output?

emersion:
Fixed the trailing cursor bug on startup.

>It also updates all outputs when a client cursor is set.
Can we just upload the new texture without updating the output?

I'll try to fix this in an upcoming PR taking care of #581, because this needs to change how the cursor position is cached in the DRM backend.

emersion:
Oops, missed that @acrisci was assigned. Let me know if you want to revert. Sorry about that.

PR #711, opened 2018-03-09
Add map/unmap to xdg-shell

### LINKED ISSUE #650: rootston: Segfault on closing some XWayland windows with opened drop down menu

### LINKED ISSUE #671: rootston crash on focus change with xwayland popup

### LINKED ISSUE #674: mpv’s initial fullscreen is ignored

### LINKED ISSUE #679: Segfault or damage tracking issue when displaying a tooltip in Firefox and alt-tabbing to another client

### LINKED ISSUE #688: Compositor cannot choose xdg-surface state before first frame
From IRC:

```
23:45 <emersion> in short:
23:45 <emersion> when a new xdg-surface appears:
23:45 <emersion> - it can request to be fullscreen/maximized/whatever
23:46 <emersion> - we have to send a configure event with the xdg-surface state (accepting requests, or not)
23:46 <emersion> - then the client knows how to draw itself and can commit a buffer
23:46 <emersion> the issue is:
23:47 <emersion> the compositor doesn't know about the xdg-surface before the first non-NULL commit
23:48 <emersion> ie. before it's mapped
23:48 <emersion> so there are two solutions:
23:48 <emersion> 1. accept all requests before the first non-NULL commit directly in wlroots
23:49 <emersion>    (if the compositors doesn't want this, the first frame will be not drawn in the correct state)
23:50 <emersion> 2. emit new_surface for unmapped xdg-surfaces, to let the compositor choose the xdg-surface state
23:50 <emersion> acrisci: what do you think
23:51 <acrisci> let me understand the problem first
23:51 <acrisci> so the window can ask to be fullscreen or something before it is mapped?
23:51 <emersion> yeah, for instance mpv
23:52 <emersion> there's a flag to launch it fullscreen
23:52 <acrisci> so if the compositor doesn't know about it, then wlroots makes the decision
23:52 <emersion> yeah
23:53 <emersion> let me give you another example
23:53 <acrisci> but the compositor should make the decision of course
23:53 <emersion> in sway apps are in containers
23:53 <emersion> containers have a size
23:54 <emersion> when a client creates an xdg-surface it doesn't know how to choose its size
23:54 <emersion> so before the first frame, the compositor should send a configure event with the desired size
23:55 <emersion> but right now the compositor doesn't know about the xdg-surface before the first frame because new_surface hasn't been emitted
23:55 <emersion> so in sway, the first frame has a size choosen by the client, then the compositor sends a request to resize the client
23:56 <emersion> this is bad
23:58 <acrisci> yep that's a problem
00:04 <acrisci> I'm afraid of leaking details on what constitutes a mapped/unmapped xdg-surface
00:05 <acrisci> I want to keep compositor commit handlers as small as possible
00:09 <emersion> hmmm
00:09 <emersion> let's create an issue about it
00:10 <acrisci> I wonder if we should do map/unmap events
00:10 <emersion> yeah
```

Also see #681

emersion:
- [x] Add unmap support to zxdg-shell v6
- [x] Read client requested state in `new_surface` in zxdg-shell-v6
- [x] Add map/unmap support to xdg-shell stable

Test plan:
* Make sure xdg-shell still works
* Try `mpv --fs` or https://gist.github.com/emersion/34ed72adecd78bdfac9208a97ffb46a3
* Try to reproduce crashes described in issues below

Fixes #688
Fixes #674
Fixes #679
Fixes #671
Fixes #650
Closes #681
Closes #708

### DIFF
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -54,7 +54,8 @@ void view_get_deco_box(const struct roots_view *view, struct wlr_box *box) {
 	box->height += (view->border_width * 2 + view->titlebar_height);
 }
 
-enum roots_deco_part view_get_deco_part(struct roots_view *view, double sx, double sy) {
+enum roots_deco_part view_get_deco_part(struct roots_view *view, double sx,
+		double sy) {
 	if (!view->decorated) {
 		return ROOTS_DECO_PART_NONE;
 	}
@@ -94,9 +95,15 @@ enum roots_deco_part view_get_deco_part(struct roots_view *view, double sx, doub
 static void view_update_output(const struct roots_view *view,
 		const struct wlr_box *before) {
 	struct roots_desktop *desktop = view->desktop;
-	struct roots_output *output;
+
+	if (view->wlr_surface == NULL) {
+		return;
+	}
+
 	struct wlr_box box;
 	view_get_box(view, &box);
+
+	struct roots_output *output;
 	wl_list_for_each(output, &desktop->outputs, link) {
 		bool intersected = before != NULL && wlr_output_layout_intersects(
 			desktop->layout, output->wlr_output, before);
@@ -479,7 +486,10 @@ void view_initial_focus(struct roots_view *view) {
 void view_setup(struct roots_view *view) {
 	view_initial_focus(view);
 
-	view_center(view);
+	if (view->fullscreen_output == NULL) {
+		view_center(view);
+	}
+
 	view_update_output(view, NULL);
 }
 
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -417,11 +417,11 @@ void view_destroy(struct roots_view *view) {
 		return;
 	}
 
+	wl_signal_emit(&view->events.destroy, view);
+
 	if (view->wlr_surface != NULL) {
 		view_unmap(view);
 	}
-
-	wl_signal_emit(&view->events.destroy, view);
 
 	if (view->destroy) {
 		view->destroy(view);
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -30,6 +30,7 @@ struct roots_view *view_create(struct roots_desktop *desktop) {
 	}
 	view->desktop = desktop;
 	view->alpha = 1.0f;
+	wl_signal_init(&view->events.unmap);
 	wl_signal_init(&view->events.destroy);
 	wl_list_init(&view->children);
 	return view;
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -459,6 +459,8 @@ void view_map(struct roots_view *view, struct wlr_surface *surface) {
 void view_unmap(struct roots_view *view) {
 	assert(view->wlr_surface != NULL);
 
+	wl_signal_emit(&view->events.unmap, view);
+
 	view_damage_whole(view);
 	wl_list_remove(&view->link);
 
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -422,6 +422,10 @@ void view_destroy(struct roots_view *view) {
 
 	wl_signal_emit(&view->events.destroy, view);
 
+	if (view->destroy) {
+		view->destroy(view);
+	}
+
 	free(view);
 }
 
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -491,7 +491,7 @@ void view_initial_focus(struct roots_view *view) {
 void view_setup(struct roots_view *view) {
 	view_initial_focus(view);
 
-	if (view->fullscreen_output == NULL) {
+	if (view->fullscreen_output == NULL && !view->maximized) {
 		view_center(view);
 	}
 
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -23,13 +23,15 @@
 #include "rootston/view.h"
 #include "rootston/xcursor.h"
 
-
-struct roots_view *view_create() {
+struct roots_view *view_create(struct roots_desktop *desktop) {
 	struct roots_view *view = calloc(1, sizeof(struct roots_view));
 	if (!view) {
 		return NULL;
 	}
+	view->desktop = desktop;
 	view->alpha = 1.0f;
+	wl_signal_init(&view->events.destroy);
+	wl_list_init(&view->children);
 	return view;
 }
 
@@ -402,20 +404,18 @@ struct roots_subsurface *subsurface_create(struct roots_view *view,
 	return subsurface;
 }
 
-void view_finish(struct roots_view *view) {
-	view_damage_whole(view);
+void view_destroy(struct roots_view *view) {
+	if (view == NULL) {
+		return;
+	}
+
+	if (view->wlr_surface != NULL) {
+		view_unmap(view);
+	}
+
 	wl_signal_emit(&view->events.destroy, view);
 
-	wl_list_remove(&view->new_subsurface.link);
-
-	struct roots_view_child *child, *tmp;
-	wl_list_for_each_safe(child, tmp, &view->children, link) {
-		child->destroy(child);
-	}
-
-	if (view->fullscreen_output) {
-		view->fullscreen_output->fullscreen_view = NULL;
-	}
+	free(view);
 }
 
 static void view_handle_new_subsurface(struct wl_listener *listener,
@@ -425,12 +425,10 @@ static void view_handle_new_subsurface(struct wl_listener *listener,
 	subsurface_create(view, wlr_subsurface);
 }
 
-void view_init(struct roots_view *view, struct roots_desktop *desktop) {
-	assert(view->wlr_surface);
+void view_map(struct roots_view *view, struct wlr_surface *surface) {
+	assert(view->wlr_surface == NULL);
 
-	view->desktop = desktop;
-	wl_signal_init(&view->events.destroy);
-	wl_list_init(&view->children);
+	view->wlr_surface = surface;
 
 	struct wlr_subsurface *subsurface;
 	wl_list_for_each(subsurface, &view->wlr_surface->subsurface_list,
@@ -442,7 +440,31 @@ void view_init(struct roots_view *view, struct roots_desktop *desktop) {
 	wl_signal_add(&view->wlr_surface->events.new_subsurface,
 		&view->new_subsurface);
 
+	wl_list_insert(&view->desktop->views, &view->link);
 	view_damage_whole(view);
+}
+
+void view_unmap(struct roots_view *view) {
+	assert(view->wlr_surface != NULL);
+
+	view_damage_whole(view);
+	wl_list_remove(&view->link);
+
+	wl_list_remove(&view->new_subsurface.link);
+
+	struct roots_view_child *child, *tmp;
+	wl_list_for_each_safe(child, tmp, &view->children, link) {
+		child->destroy(child);
+	}
+
+	if (view->fullscreen_output != NULL) {
+		output_damage_whole(view->fullscreen_output);
+		view->fullscreen_output->fullscreen_view = NULL;
+		view->fullscreen_output = NULL;
+	}
+
+	view->wlr_surface = NULL;
+	view->width = view->height = 0;
 }
 
 void view_initial_focus(struct roots_view *view) {
--- a/rootston/xdg_shell_v6.c
+++ b/rootston/xdg_shell_v6.c
@@ -63,9 +63,11 @@ static void get_size(const struct roots_view *view, struct wlr_box *box) {
 	if (surface->geometry->width > 0 && surface->geometry->height > 0) {
 		box->width = surface->geometry->width;
 		box->height = surface->geometry->height;
-	} else {
+	} else if (view->wlr_surface != NULL) {
 		box->width = view->wlr_surface->current->width;
 		box->height = view->wlr_surface->current->height;
+	} else {
+		box->width = box->height = 0;
 	}
 }
 
--- a/rootston/xdg_shell_v6.c
+++ b/rootston/xdg_shell_v6.c
@@ -309,6 +309,8 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
 	wl_list_remove(&roots_xdg_surface->surface_commit.link);
 	wl_list_remove(&roots_xdg_surface->destroy.link);
 	wl_list_remove(&roots_xdg_surface->new_popup.link);
+	wl_list_remove(&roots_xdg_surface->map.link);
+	wl_list_remove(&roots_xdg_surface->unmap.link);
 	wl_list_remove(&roots_xdg_surface->request_move.link);
 	wl_list_remove(&roots_xdg_surface->request_resize.link);
 	wl_list_remove(&roots_xdg_surface->request_maximize.link);
--- a/rootston/xdg_shell_v6.c
+++ b/rootston/xdg_shell_v6.c
@@ -180,6 +180,21 @@ static void close(struct roots_view *view) {
 	}
 }
 
+static void destroy(struct roots_view *view) {
+	assert(view->type == ROOTS_XDG_SHELL_V6_VIEW);
+	struct roots_xdg_surface_v6 *roots_xdg_surface = view->roots_xdg_surface_v6;
+	wl_list_remove(&roots_xdg_surface->surface_commit.link);
+	wl_list_remove(&roots_xdg_surface->destroy.link);
+	wl_list_remove(&roots_xdg_surface->new_popup.link);
+	wl_list_remove(&roots_xdg_surface->map.link);
+	wl_list_remove(&roots_xdg_surface->unmap.link);
+	wl_list_remove(&roots_xdg_surface->request_move.link);
+	wl_list_remove(&roots_xdg_surface->request_resize.link);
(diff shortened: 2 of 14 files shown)


rootston/desktop.c
```diff
@@ -402,20 +411,18 @@ struct roots_subsurface *subsurface_create(struct roots_view *view,
...
+		return;
 	}
 
-	if (view->fullscreen_output) {
-		view->fullscreen_output->fullscreen_view = NULL;
+	if (view->wlr_surface != NULL) {
+		view_unmap(view);
 	}
+
+	wl_signal_emit(&view->events.destroy, view);
```
acrisci: destroy event normally goes first before all the extra cleanup is done.

rootston/desktop.c
```diff
@@ -425,12 +432,10 @@ static void view_handle_new_subsurface(struct wl_listener *listener,
 	subsurface_create(view, wlr_subsurface);
 }
 
-void view_init(struct roots_view *view, struct roots_desktop *desktop) {
-	assert(view->wlr_surface);
+void view_map(struct roots_view *view, struct wlr_surface *surface) {
```
acrisci: The interface seems a little bit misleading since you aren't able to unmap a view, then map it with an arbitrary different surface. The view might already be set up with a different type of surface and they won't match.

rootston/xdg_shell_v6.c
```diff
@@ -287,9 +313,7 @@ static void handle_destroy(struct wl_listener *listener, void *data) {
 	wl_list_remove(&roots_xdg_surface->request_resize.link);
 	wl_list_remove(&roots_xdg_surface->request_maximize.link);
 	wl_list_remove(&roots_xdg_surface->request_fullscreen.link);
-	wl_list_remove(&roots_xdg_surface->view->link);
-	view_finish(roots_xdg_surface->view);
-	free(roots_xdg_surface->view);
+	view_destroy(roots_xdg_surface->view);
```
acrisci: Would it fit better if we added a `destroy()` method to the view interface which is called by `view_destroy()`? That seems to be how all the other interfaces work.

acrisci: The problem is that if `view_destroy()` emits the `destroy` event, then that has to happen before all the extra cleanup to avoid people accidently using the freed things in their destroy handler.

acrisci:
Looks like the map/unmap listeners are not removed on view destroy.

rootston/desktop.c
acrisci: Would it be better to not have the `surface` as a function parameter and just get the surface from the shell surface implementation?

acrisci:
Approach looks good. Will do the more in depth testing once WIP status is removed.

rootston/desktop.c
emersion: I don't want to call `view_unmap` after emitting the destroy event because the destroy event may have free'd stuff used by `view_unmap`.

rootston/desktop.c
emersion: >The interface seems a little bit misleading since you aren't able to unmap a view, then map it with an arbitrary different surface.

The map/unmap API is private to shells. Only the shell can decide which wl_surface needs to be mapped.

>Would it be better to not have the surface as a function parameter and just get the surface from the shell surface implementation?

That would mean adding `view->wl_surface = surface` before each `view_map`. Not sure it's better.

emersion:
Added unmap support to zxdg-shell-v6.

Is a in-depth review possible right now? The only thing missing is xdg-shell stable support, since it's difficult to test I'll backport it from zxdg-shell-v6 when the zxdg-shell-v6 implementation is approved.

rootston/desktop.c
acrisci: That shouldn't be an issue (and if it is, we have bigger problems).

According to our patterns, the view should have ownership over all its own properties and should be responsible for freeing them. If another part of the program frees the properties, then we'll get double frees when the view is destroyed by the `view_destroy` method. This is an application bug.

In the destroy event, the wrappers of the view should *only* be responsible for cleanup up their own wrappers and not touch the views own properties at all. Users of the view as a property should simply set it to NULL. That should be the extent of the cleanup done in those destroy handlers (and anything more is an application bug).

acrisci:
How do you reproduce the mpv bug?

emersion:
But you need mpv-git to be able to reproduce right?

acrisci:
On the DRM backend, it's rendering a bunch of junk and I get this:

```
Playing: /home/tony/Downloads/Kyary Pamyu Pamyu - PONPONPON[720p][MV].mp4
 (+) Video --vid=1 (*) (h264 1280x720 23.976fps)
 (+) Audio --aid=1 --alang=und (*) (aac 2ch 44100Hz)
[vo/opengl] retrieving framebuffer depth: OpenGL error INVALID_OPERATION.
AO: [pulse] 44100Hz stereo 2ch float
VO: [opengl] 1280x720 yuv420p
AV: 00:00:00 / 00:04:13 (0%) A-V:  0.000
[vo/opengl/wayland] error occurred on the display fd: closing file descriptor
[vo/opengl/wayland] error occurred on the display fd: closing file descriptor
```

Ongy:
mpv 0.28 is required for xdg-shell support. But it shouldn't be garbage on wl-shell either.

Is the fullscreen handling implemented? I tried on drm backend and mpv 0.28 (mpv-git) and it still looks similar to before.

emersion:
You're right, I'm missing the code reading the next state on `new_surface`.

acrisci:
Installed development version of mpv and it doesn't fullscreen with the `--fs` switch. Nice to see the video without tearing though.

emersion:
Okay, finished all the work for zxdg-shell-v6. Can you try again?

Added a test client to the test plan.

acrisci:
The test works now.

emersion:
I changed a lot of stuff in zxdg-shell-v6 configure, ack_configure and commit. Do you agree with these changes?

types/wlr_xdg_shell_v6.c
```diff
@@ -868,15 +868,19 @@ static void wlr_xdg_toplevel_v6_ack_configure(
...
+		configure->toplevel_state->fullscreen;
 	surface->toplevel_state->current.resizing =
-		configure->toplevel_state.resizing;
+		configure->toplevel_state->resizing;
 	surface->toplevel_state->current.activated =
-		configure->toplevel_state.activated;
+		configure->toplevel_state->activated;
+
+	free(configure->toplevel_state);
+	configure->toplevel_state = NULL;
```
acrisci: What if I ack_configure twice?

types/wlr_xdg_shell_v6.c
```diff
@@ -1025,7 +1029,12 @@ static void wlr_xdg_toplevel_v6_send_configure(
 	uint32_t *s;
 	struct wl_array states;
 
-	configure->toplevel_state = surface->toplevel_state->pending;
+	configure->toplevel_state = malloc(sizeof(*configure->toplevel_state));
+	if (configure->toplevel_state == NULL) {
+		wlr_log(L_ERROR, "Allocation failed");
```
acrisci: post no memory to the client.

acrisci:
You'll leak the toplevel_state if it never gets acked (~line 900)

acrisci:
I think `wlr_xdg_surface_v6_configure` needs a dedicated free func.

acrisci:
You made the toplevel state of the configure nullable, but I don't see anywhere you actually use this fact. What is the reason for the change?

emersion:
>I think wlr_xdg_surface_v6_configure needs a dedicated free func.

Good idea, added that.

>You made the toplevel state of the configure nullable, but I don't see anywhere you actually use this fact. What is the reason for the change?

It's because configure events are not just for toplevels, they're also for popups. I just wanted to make it clear that the toplevel part of the configure struct doesn't make sense and shouldn't be used when the surface is a popup.

acrisci:
ok good so far :+1:

types/wlr_xdg_shell_v6.c
```diff
@@ -159,6 +159,13 @@ static struct wlr_xdg_popup_grab_v6 *xdg_shell_popup_grab_from_seat(
 }
 
 
+static void xdg_surface_configure_destroy(
+		struct wlr_xdg_surface_v6_configure *configure) {
```
acrisci: need a NULL guard here.

emersion:
All right, this is ready for final review.

rootston/desktop.c
```diff
@@ -442,9 +452,33 @@ void view_init(struct roots_view *view, struct roots_desktop *desktop) {
 	wl_signal_add(&view->wlr_surface->events.new_subsurface,
 		&view->new_subsurface);
 
+	wl_list_insert(&view->desktop->views, &view->link);
 	view_damage_whole(view);
 }
 
+void view_unmap(struct roots_view *view) {
+	assert(view->wlr_surface != NULL);
+
```
Ongy: Shouldn't there be a `wlr_signal_emit_safe(&view->events.unmap, view);` here?
Quick text-search on the diff page doesn't bring up a signal_emit for unmap on the view, only for surfaces.

rootston/desktop.c
```diff
@@ -402,20 +412,22 @@ struct roots_subsurface *subsurface_create(struct roots_view *view,
...
+	wl_signal_emit(&view->events.destroy, view);
 
-	struct roots_view_child *child, *tmp;
-	wl_list_for_each_safe(child, tmp, &view->children, link) {
-		child->destroy(child);
+	if (view->wlr_surface != NULL) {
+		view_unmap(view);
 	}
 
-	if (view->fullscreen_output) {
```
Ongy: Where is this handled now? I expected to see the output use the unmap signal, but (I may be blind) I'm missing that code.

emersion: It's done in `view_unmap`.

rootston/desktop.c
emersion: You're right!

Ongy:
Fixes crashes in:
* https://github.com/swaywm/wlroots/issues/679
* https://github.com/swaywm/wlroots/issues/671
* https://github.com/swaywm/wlroots/issues/650

Someone else should probably test as well before closing.

PR #716, opened 2018-03-11
Fix cursor hotspot update

### LINKED ISSUE #581: Cursor jumps  on decoration border enter in DRM backend

emersion:
Fixes #581
~~Depends on #710~~

Tried to do not update the output when a new cursor image is set and the cursor isn't visible, but in the end IMHO it's not worth it: it complicates the DRM code and the benefits are too low.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -602,12 +602,25 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 		wlr_output_transform_invert(output->transform);
 	wlr_box_transform(&hotspot, transform,
 		plane->surf.width, plane->surf.height, &hotspot);
-	plane->cursor_hotspot_x = hotspot.x;
-	plane->cursor_hotspot_y = hotspot.y;
+
+	if (plane->cursor_hotspot_x != hotspot.x ||
+			plane->cursor_hotspot_y != hotspot.y) {
+		// Update cursor hotspot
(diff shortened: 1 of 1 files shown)


emersion:
This PR is ready for review.

backend/drm/drm.c
```diff
@@ -602,12 +602,26 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
...
+			plane->cursor_hotspot_y != hotspot.y) {
+		// Update cursor hotspot
+		conn->cursor_x -= hotspot.x - plane->cursor_hotspot_x;
+		conn->cursor_y -= hotspot.y - plane->cursor_hotspot_y;
+		plane->cursor_hotspot_x = hotspot.x;
+		plane->cursor_hotspot_y = hotspot.y;
+
+		bool ok = drm->iface->crtc_move_cursor(drm, conn->crtc, conn->cursor_x,
+			conn->cursor_y);
+		if (!ok) {
```
SirCmpwn: I don't think this ok variable needs to exist

emersion: Fixed

PR #720, opened 2018-03-13
xdg-positioner

### LINKED ISSUE #231: Popups render off screen and across outputs
To reproduce, move a window adjacent to an output boundary and open some menu. You will notice that the menu can be drawn off the screen. Also, the menu should not be drawn across outputs.

The xdg-positioner has rules for how the popup should be adjusted when the popup is constrained like this, but those rules are currently not used. The popup might ask to "slide over" or maybe to have the popup change its size.

The implementation might be complicated because now the shell will need to find some way to interact with the wlr-output-layout to find the screen boundaries given its layout coordinates.

This feature seems implemented in X11-based window managers, but not in Weston. So maybe we need to try to patch Weston for this first to get their feedback on the implementation.

### LINKED ISSUE #684: xdg positioner

acrisci:
Let's try to make the xdg-positioner work properly. This specifies what to do when a popup is constrained. A popup is constrained when part of it is not visible such as when it intersects the edge of an output.

TODO:
* [x] store positioner properties on the popup state
* [x] popup constraint detection
* [x] find test clients
* [x] flip_x, flip_y
    - the popup flips to the other side of its parent
* [x] slide_x, slide_y
    - the popup moves over until it is not constrained
* [x] resize_x, resize_y
    - the popup resizes so it fits the screen bounds
* [ ] port to xdg-shell stable
* [x] wlroots library abstractions

xdg-positioner can specify to do all three of these (in the order presented above) until it is no longer constrained (which should always happen at or before the resize step.

The actual moving/resizing must happen in rootston because only it knows when the popup is constrained.

Fixes #684
Fixes #231

Test plan:

* test flip: open a popup so that it would be constrained on the side of the screen, but filpping it (reversing appropriate x/y direction and gravity) would have the popup unconstrained, then the popup should do the flip. Do this for each edge of the output and also for nested popups.
* test slide: open a popup so that it would be constrained on the side of the screen, but flipping it will not unconstrain the popup (e.g., if the anchor point is very close to the edge), then it should slide along the axis and still be visible on the screen. (test for each edge of the output and for nested popups).

There should never be a case where a popup with every flag has some portion of the popup not visible on the screen.


Ongy:
What are your current plans to get the box to layout in?
An explicit "layout_popup" event? Would it make sense to have an approach like https://github.com/swaywm/wlroots/pull/711 for popups as well?

acrisci:
> What are your current plans to get the box to layout in?

The box of the output the parent is on. I can't think of anything else that would constrain a popup but there might be other things.

> An explicit "layout_popup" event? Would it make sense to have an approach like #711 for popups as well?

I think we have an event for that (`new_popup` or something). On that event, rootston should adjust the popup geometry so it is not constrained.

emersion:
>The box of the output the parent is on. I can't think of anything else that would constrain a popup but there might be other things.

There might be panels or widgets (surface-layers) that the compositor doesn't want to occlude. Maybe a pixman region would be more flexible?

Ongy:
wlroots doesn't have information about where on the output the surface is on.
Or which output the popup should be on, when a surface is displayed on multiple outputs, so using wlr_output_layout won't be enough to make this work properly

emersion:
>wlroots doesn't have information about where on the output the surface is on.

Just translate the box/region so that the surface corner is at (0, 0)?

acrisci:
> Maybe a pixman region would be more flexible?

My plan is to have a helper function that takes some sort of geometry (like box or pixman region) and the positioner info and update the popup geometry so library users don't have to do the math.

include/wlr/types/wlr_xdg_shell_v6.h
```diff
@@ -98,9 +113,16 @@ struct wlr_xdg_popup_v6 {
 	struct wlr_seat *seat;
 	struct wlr_box geometry;
 
-	enum wlr_positioner_v6_anchor anchor;
-	enum wlr_positioner_v6_gravity gravity;
-	enum wlr_positioner_v6_constraint_adjustment constraint_adjustment;
+	struct wlr_xdg_positioner_v6_attributes positioner;
+	struct wlr_box anchor_rect;
```
Ongy: Is there a reason we have this twice? Once in the attributes, once on the wlr_xdg_popup itself.

Same for size and offset

types/wlr_xdg_shell_v6.c
```diff
@@ -18,22 +18,9 @@ static const char *wlr_desktop_xdg_popup_role = "xdg_popup_v6";
...
-	enum zxdg_positioner_v6_constraint_adjustment constraint_adjustment;
-
-	struct {
-		int32_t width, height;
-	} size;
-
-	struct {
-		int32_t x, y;
-	} offset;
+	struct wlr_xdg_positioner_v6_attributes *attrs;
```
Ongy: Is there a reason to have this a pointer?
The wlr_xdg_popup_v6 has it inline, and it's allocated and free()d together with the positioner struct

include/wlr/types/wlr_xdg_shell_v6.h
```diff
@@ -32,6 +32,77 @@ struct wlr_xdg_client_v6 {
...
+	WLR_POSITIONER_V6_ANCHOR_BOTTOM = 2,
+	/**
+	 * the left edge of the anchor rectangle
+	 */
+	WLR_POSITIONER_V6_ANCHOR_LEFT = 4,
+	/**
+	 * the right edge of the anchor rectangle
+	 */
+	WLR_POSITIONER_V6_ANCHOR_RIGHT = 8,
+};
```
SirCmpwn: -1, can't we just use the protocol enums?

rootston/xdg_shell_v6.c
```diff
@@ -38,6 +38,186 @@ static void popup_handle_new_popup(struct wl_listener *listener, void *data) {
...
+static void popup_get_coords(struct wlr_xdg_popup_v6 *popup,
+		double *sx, double *sy) {
+	struct wlr_xdg_surface_v6 *parent = popup->parent;
+	double popup_sx = popup->geometry.x;
+	double popup_sy = popup->geometry.y;
+	while (parent->role != WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL) {
+		popup_sx += parent->popup->geometry.x;
+		popup_sy += parent->popup->geometry.y;
+		parent = parent->popup->parent;
+	}
```
SirCmpwn: This would make for a nice helper function in wlroots

SirCmpwn: Meaning this while loop, but also the whole thing

include/wlr/types/wlr_xdg_shell_v6.h
SirCmpwn: bump

rootston/xdg_shell_v6.c
```diff
@@ -50,6 +50,48 @@ static void popup_handle_new_popup(struct wl_listener *listener, void *data) {
...
+	anchor_lx += popup_lx;
+	anchor_ly += popup_ly;
+
+	double dest_x = 0, dest_y = 0;
+	wlr_output_layout_closest_point(layout, NULL, anchor_lx, anchor_ly,
+		&dest_x, &dest_y);
+
+	struct wlr_output *output =
+		wlr_output_layout_output_at(layout, dest_x, dest_y);
+	// XXX: handle empty output layout
```
SirCmpwn: Can this happen?

rootston/xdg_shell_v6.c
```diff
@@ -50,6 +50,48 @@ static void popup_handle_new_popup(struct wl_listener *listener, void *data) {
...
+	struct wlr_box toplevel_box = {
+		.x = output->lx - view->x,
+		.y = output->ly - view->y,
+		.width = width,
+		.height = height
+	};
+
+	wlr_xdg_popup_v6_unconstrain_from_box(popup->wlr_popup, &toplevel_box);
+}
+
```
SirCmpwn: I really like this function

types/wlr_output_layout.c
```diff
@@ -339,8 +339,12 @@ void wlr_output_layout_closest_point(struct wlr_output_layout *layout,
...
 	}
 
-	*dest_x = min_x;
-	*dest_y = min_y;
+	if (dest_x) {
+		*dest_x = min_x;
+	}
+	if (dest_y) {
+		*dest_y = min_y;
+	}
```
SirCmpwn: We can return early from this function without doing anything if these are both NULL, right?

types/wlr_xdg_shell_v6.c
SirCmpwn: bump

types/wlr_xdg_shell_v6.c
```diff
@@ -423,6 +411,10 @@ static void xdg_shell_handle_create_positioner(struct wl_client *wl_client,
 		return;
 	}
 
+	// TODO: allocate the positioner attrs?
```
SirCmpwn: ?

SirCmpwn:
I like it. Needs minor changes but it looks great, nice work.

rootston/xdg_shell_v6.c
acrisci: I think you just return early in this case because there is no way to unconstrain it.

types/wlr_xdg_shell_v6.c
acrisci: Going to make this not allocated.

types/wlr_output_layout.c
acrisci: Fixed. We should probably do more NULL checks like this for out variables.

include/wlr/types/wlr_xdg_shell_v6.h
emersion: We should be consistent. We already redefine enums for other protocols.

types/wlr_xdg_shell_v6.c
```diff
@@ -1639,3 +1627,243 @@ struct wlr_xdg_surface_v6 *wlr_xdg_surface_v6_popup_at(
...
+		sy = rect.y;
+	} else if (anchor == (WLR_POSITIONER_V6_ANCHOR_BOTTOM |
+				WLR_POSITIONER_V6_ANCHOR_LEFT)) {
+		sx = rect.x;
+		sy = rect.y + rect.height;
+	} else if (anchor == (WLR_POSITIONER_V6_ANCHOR_BOTTOM |
+				WLR_POSITIONER_V6_ANCHOR_RIGHT)) {
+		sx = rect.x + rect.width;
+		sy = rect.y + rect.height;
+	}
```
emersion: This can be factored, for instance:

```c
if (anchor & WLR_POSITIONER_V6_ANCHOR_TOP) {
    sy = rect.y;
}
```

types/wlr_xdg_shell_v6.c
```diff
@@ -1639,3 +1627,243 @@ struct wlr_xdg_surface_v6 *wlr_xdg_surface_v6_popup_at(
...
+
+	*root_sx = sx;
+	*root_sy = sy;
+}
+
+void wlr_xdg_popup_v6_get_toplevel_coords(struct wlr_xdg_popup_v6 *popup,
+		int popup_sx, int popup_sy, int *toplevel_sx, int *toplevel_sy) {
+	struct wlr_xdg_surface_v6 *parent = popup->parent;
+	while (parent && parent->role != WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL) {
```
emersion: Since we're reading the `parent->popup` field, I'd prefer the condition to be `parent->role == WLR_XDG_SURFACE_V6_ROLE_POPUP`.

include/wlr/types/wlr_xdg_shell_v6.h
acrisci: I'll give an opinion, but these fields are never required to be read by rootston so the stakes aren't very high here.

This serves several functions. We probably won't include the generated header with the library, so the wrapper type ensures we have a type. Types are better than just using uint32_t because they serve as documentation without people needing to look up what the values mean in the protocol, or be required to use wayland-scanner themselves to generate the header. This makes the library more accessible and user friendly. It could end up saving our users hours of time.

Defining the enums adds a trivial amount of complexity to the library. There are no moving parts, and these values should never change (they will bump the version if they make breaking changes). So if you do the cost/benefit analysis, it seems to me that defining the enums is a win by adding value to the user.

This can be seen as the responsibility of the library to do because really at the end of the day, isn't everything we have done here just a collection of wrapper types just like this?

SirCmpwn: I agree, but let's not merge more the wrong way

SirCmpwn: @acrisci 

I don't want to duplicate code. There is a single source of truth for these values, and it is the protocol extensions. Getting wayland-scanner working is the *least* of your worries when you're making a Wayland compositor based on wlroots.

untitled

emersion: sava

rootston/xdg_shell_v6.c
```diff
@@ -50,6 +50,50 @@ static void popup_handle_new_popup(struct wl_listener *listener, void *data) {
...
+	int width = 0, height = 0;
+	wlr_output_effective_resolution(output, &width, &height);
+
+	struct wlr_box toplevel_box = {
+		.x = output->lx - view->x,
+		.y = output->ly - view->y,
+		.width = width,
+		.height = height
+	};
```
emersion: I don't quite understand this part. This isn't the _toplevel box_ per se, because it has the width/height of the output right?

acrisci: Maybe it's not named correctly. What I'm trying to express is that it is the box of the output in toplevel coordinates (relative to the upper left corner of the toplevel surface corner). Do you have a suggestion for a better name?

include/wlr/types/wlr_xdg_shell_v6.h
```diff
@@ -229,4 +247,46 @@ struct wlr_xdg_surface_v6 *wlr_xdg_surface_v6_popup_at(
...
+void wlr_xdg_popup_v6_get_toplevel_coords(struct wlr_xdg_popup_v6 *popup,
+		int popup_sx, int popup_sy, int *toplevel_sx, int *toplevel_sy);
+
+/**
+ * Set the geometry of this popup to unconstrain it according to its
+ * xdg-positioner rules. The box should be in the popup's toplevel
+ * parent surface coordinate system.
+ */
+void wlr_xdg_popup_v6_unconstrain_from_box(struct wlr_xdg_popup_v6 *popup,
+		struct wlr_box *toplevel_box);
```
emersion: Can you rename and describe what the box is here?

acrisci: what about `toplevel_sx_box`?

SirCmpwn:
:tada:

PR #721, opened 2018-03-14
examples: simplify screenshot client

### LINKED ISSUE #604: Screenshooter issues with output scales
The screenshooter example doesn't work with multiple outputs with a different scale each.

Pixel buffers are just copied to the output image without scaling. This means views will look bigger on HiDPI screens in the resulting image, and more importantly some outputs can be be copied on top of each other (for instance try a scale=2 output with a scale=1 output on the right).

Fractional scaling makes this even more difficult, because the screenshooter has no way to know the real output scale. That can be solved by implementing xdg-output which provides outputs' logical position and size.

From here we have to make a decision:
* Support output scales, we'll probably need to use a proper image library to scale buffers
* Only allow to screenshot one output with the screenshooter

emersion:
Fixes #604

### DIFF
--- a/examples/screenshot.c
+++ b/examples/screenshot.c
@@ -26,6 +26,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


SirCmpwn:
I liked the fancier approach. But seeing as I don't want to maintain it any more than you do, this is probably better.

Ongy:
Why remove the transform handling? We had an issue with the position of outputs, but transform worked fine afaik.

emersion:
>Why remove the transform handling? We had an issue with the position of outputs, but transform worked fine afaik.

It just seemed weird to support only half of the output properties. Since this is just an example that isn't really usable anymore (will probably just be used for testing), I thought it would be simpler if we removed the transforms handling. Also this part of the code wasn't really pretty nor efficient (because we aren't using a proper image library). I can add it back if you think it's useful.

Ongy:
mhh, readding the in-line solution probably doesn't make sense. Since we have one file per output now this can be done with imagemagick/use-site.

It could be interesting to use the `convert` arguments (-flip/-transpose/-transverse) for writing the image, but I don't think it's important.

acrisci:
:+1:

PR #722, opened 2018-03-15
[WIP] Matrix redesign

emersion:
This ports the matrix part of #319.


emersion:
This PR is ready for review.

emersion:
@ascent12 Actually it would make more sense to rename `wlr_matrix_texture` to something like `wlr_matrix_projection`. What do you think?

ascent12:
That's fine. I think it was originally called that, because it was the only matrix function, but now the name is too generic.

While you're doing this, you can easily remove the matrix transpose stuff from the shaders.
You just need to change GL_FALSE to GL_TRUE in https://github.com/swaywm/wlroots/blob/master/render/gles2/renderer.c#L181 and other glUniformMatrix4fv functions.

emersion:
>While you're doing this, you can easily remove the matrix transpose stuff from the shaders.

Nice, thanks for the tip. I was wondering how you removed that in your renderer redesign PR.

martinetd:
I know I come in a bit late, but there's some weird behaviour with view rotation (meta-key + middle click) - it rotates in the wrong direction and the decoration moves weirdly. Happy to have a shot at fixing this but will probably be faster for @emersion

emersion:
@martinetd Can you open an issue for this? I'll have a look tomorrow.

PR #723, opened 2018-03-15
Fix y-inverted rendering for xwayland surfaces

emersion:
#698 introduces a regression for xwayland surfaces, they appear inverted wrt the y axis.


SirCmpwn:
Nice

PR #724, opened 2018-03-15
Linux-dmabuf with n_planes > 1

agx:
This aims to add support for formats with more than one plane like NV12.

Test plan
- [x] ./weston-simple-dmabuf-drm --import-format=NV12 (new functionality)
- [X] ./weston-simple-dmabuf-drm --import-format=XRGB (make sure old code still works)

The drivers I have don't support NV12 but maybe someone's does. Therefore marking as WiP since I can't really test it atm but I'll see if can come up with s.th. else that has more than one plane.

According to https://cgit.freedesktop.org/wayland/weston/commit/?id=ee58911912d88caacf340a5cb6a9e25fcba24996 freedreno supports NV12 if somebody running rootston on freedreno could test it that would be great.

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -325,15 +325,6 @@ bool wlr_egl_swap_buffers(struct wlr_egl *egl, EGLSurface surface,
 
 EGLImage wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
 		struct wlr_dmabuf_buffer_attribs *attributes) {
-	int atti = 0;
-	EGLint attribs[20];
-	attribs[atti++] = EGL_WIDTH;
-	attribs[atti++] = attributes->width;
-	attribs[atti++] = EGL_HEIGHT;
-	attribs[atti++] = attributes->height;
-	attribs[atti++] = EGL_LINUX_DRM_FOURCC_EXT;
-	attribs[atti++] = attributes->format;
-
 	bool has_modifier = false;
 	if (attributes->modifier[0] != DRM_FORMAT_MOD_INVALID) {
 		if (!egl->egl_exts.dmabuf_import_modifiers) {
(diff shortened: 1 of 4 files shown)


render/egl.c
```diff
@@ -343,6 +337,52 @@ EGLImage wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
 		attribs[atti++] = EGL_DMA_BUF_PLANE0_MODIFIER_HI_EXT;
 		attribs[atti++] = attributes->modifier[0] >> 32;
 	}
+
+	if (attributes->n_planes > 0) {
```
SirCmpwn: Would prefer to do this in some kind of loop rather than as 3 if-statements with copypasta'd bodies

agx: I thought the same but since the constants vary (they have the plane number included) I went that way. Otherwise we'd have to setup all 5 constants upfront with 5 arrays:

pitch_ex_consts[] = [ EGL_DMABUF_PLANE0_PITCH_EX, EGL_DMABUF_PLANE1_EX, ...]

(same for fd, offset, modifier_lo, modifier_hi) and loop over these too or use some macro magic so I went for the open coded version.

SirCmpwn: I'd rather set up the array like you suppose.

agx:
It turns out my intel card supports NV12 (https://github.com/swaywm/wlroots/pull/726) but not the buffer modifier used by weston-simple-dmabuf-drm (MOD_SAMSUNG_64_32_TILE). If I use another buffer modifier I get something rendered so it doesn't look too bad.

emersion:
Nice! Can you provide a weston-simple-dmabuf-drm patch so we can try it?

agx:
It really only disables the  modifier atm:

https://github.com/agx/weston/commit/4f788cff433a6e381554c013681f636fc2865bff

so it displays only garbage but we can see that the multiple planes get added at least.

agx:
@emersion I have a more complete NV12 test client now

https://github.com/agx/weston/tree/nv12

it renders correctly on weston but rootston seems to pick up the first plane as RGB, I'll have a look

agx:
This renders correctly on my intel card. Besides @SirCmpwn comment I need to cleanup the last patch that handles GL_TEXTURE_EXTERNAL_OES.

render/egl.c
```diff
@@ -365,6 +359,52 @@ EGLImage wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
```
ascent12: You're adding plane 0 to the array twice.

agx:
On Mon, Mar 19, 2018 at 01:37:01PM -0700, Scott Anderson wrote:
> ascent12 commented on this pull request.
> 
> 
> 
> > @@ -365,6 +359,52 @@ EGLImage wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
>  		attribs[atti++] = EGL_DMA_BUF_PLANE0_MODIFIER_HI_EXT;
>  		attribs[atti++] = attributes->modifier[0] >> 32;
>  	}
> +
> +	if (attributes->n_planes > 0) {
> 
> You're adding plane 0 to the array twice.

Thanks for catching that!

agx:
Depends on https://github.com/swaywm/wlroots/pull/736 which should go in first.

render/egl.c
```diff
@@ -365,6 +359,53 @@ EGLImage wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
+		if (has_modifier) {
+			attribs[atti++] = EGL_DMA_BUF_PLANE3_MODIFIER_LO_EXT;
+			attribs[atti++] = attributes->modifier[3] & 0xFFFFFFFF;
+			attribs[atti++] = EGL_DMA_BUF_PLANE3_MODIFIER_HI_EXT;
+			attribs[atti++] = attributes->modifier[3] >> 32;
+		}
+	}
+
+
 	attribs[atti++] = EGL_NONE;
```
emersion: Maybe adding an assertion here would be safer:

```c
assert(atti < sizeof(attribs)/sizeof(attribs[0]));
```

emersion:
Is it expected that XRGB and NV12 don't render the same image?

render/egl.c
agx: Since the size of attribs is fixed we're safe but in case we change the number of planes we might forget to increase it so I added the assert.

agx:
@emersion where do we render a different image?

emersion:
I mean, `weston-simple-dmabuf-drm --import-format=NV12` and `weston-simple-dmabuf-drm --import-format=XRGB` don't show the same image.

agx:
@SirCmpwn is using a loop for the attributes a must have? I do like the open coded version better since it makes it obvious what's happening. Once we support > 4 planes I would agree to not open code it.

agx:
@emersion ahh...thought so. Yes, I didn't bother to reproduce it exactly for the time being, it will make it simpler to find out which method is actually used.

ascent12:
I don't believe any format exists that has more than 4 planes. In fact, I don't know of any formats with more than 3, although I guess the 4th plane could be for an alpha channel or something.

I'd much prefer a loop as well. Unrolling is fine for maybe a few lines of code, but what's going on there is too much.
It wouldn't be hard to do it with an unnamed struct and an array.

agx:
@ascent12 I don't think we'll have more than 4 planes anytime soon either (in fact NV12 only uses two and for YUV we have more work ahead). 

I'll use a loop then.

agx:
Removed the unrolled loop now

render/egl.c
```diff
@@ -347,25 +338,52 @@ EGLImage wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
-		attribs[atti++] = attributes->modifier[0] & 0xFFFFFFFF;
-		attribs[atti++] = EGL_DMA_BUF_PLANE0_MODIFIER_HI_EXT;
-		attribs[atti++] = attributes->modifier[0] >> 32;
+#define ATTR_NAMES(N) { \
+	EGL_DMA_BUF_PLANE ## N ## _FD_EXT,          \
+	EGL_DMA_BUF_PLANE ## N ## _OFFSET_EXT,	    \
+	EGL_DMA_BUF_PLANE ## N ## _PITCH_EXT,       \
+	EGL_DMA_BUF_PLANE ## N ## _MODIFIER_LO_EXT, \
+	EGL_DMA_BUF_PLANE ## N ## _MODIFIER_HI_EXT  \
+}
```
SirCmpwn: We don't use macros, find another way

SirCmpwn:
LGTM aside from the macro

emersion:
>Droped the macro in favour or spelled out constants.

It seems the macro is still here.

agx:
On Wed, Mar 21, 2018 at 05:42:02PM +0000, emersion wrote:
> >Droped the macro in favour or spelled out constants.
> 
> It seems the macro is still here.

Oh my, fixed now.

agx:
Rebased against recent gles2 changes.

agx:
On Fri, Mar 23, 2018 at 01:41:41PM +0000, Drew DeVault wrote:
> Merged #724.

Thanks for pulling this in.

SirCmpwn:
Thanks for the hard work!

PR #726, opened 2018-03-16
egl: print supported dmabuf formats

agx:
Useful for debugging dmabuf related problems e.g. when a client requests a certain and we fail to support it.

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -1,4 +1,5 @@
 #include <assert.h>
+#include <stdio.h>
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
 #include <GLES2/gl2.h>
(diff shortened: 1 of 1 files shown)


render/egl.c
```diff
@@ -101,6 +102,28 @@ static bool check_egl_ext(const char *egl_exts, const char *ext) {
 	return false;
 }
 
+static void print_dmabuf_formats(struct wlr_egl *egl)
+{
```
emersion: Style nit: these two lines should be on the same one (I understand it's difficult to switch between wlroots' style and weston's :P ).

render/egl.c
```diff
@@ -101,6 +102,28 @@ static bool check_egl_ext(const char *egl_exts, const char *ext) {
...
+		return;
+	}
+
+	int *formats;
+	int num = wlr_egl_get_dmabuf_formats(egl, &formats);
+	if (num < 0) {
+		return;
+	}
+
+	char *str_formats = malloc(num * 5 + 1);
```
emersion: I'd prefer to use VLAs here:

```c
char str_formats[num * 5 + 1];
```

render/egl.c
```diff
@@ -101,6 +102,28 @@ static bool check_egl_ext(const char *egl_exts, const char *ext) {
...
+	for (int i = 0; i < num; i++) {
+		sprintf(&str_formats[i*5], "%.4s ", (char*)&formats[i]);
```
emersion: Use `snprintf` just in case.

agx:
@emersion if that's the only style nit then I think I'm improving. Fixed.

emersion:
Thanks!

PR #727, opened 2018-03-17
Address sanitizer fixes

martinetd:
- Some memory leaks in xdg shells, rootston output and x11 backend
- A use after free in wayland backend.

I'm not 100% happy with my use-after-free fix, if someone has a better idea to access the wlr_wl_pointer from the output destroy handler. At first I wanted to add another output event handler listener to the pointer itself but since the wayland backend can have many outputs and a single pointer (at least with the current design, how does it handle being started from a rootston with multiple seats?) it was more straight-forward to cleanup from output, but I'm sure this can be done better.

### DIFF
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -26,6 +26,11 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 		// GNOME sends a pointer enter when the surface is being destroyed
 		return;
 	}
+	if (wlr_wl_pointer->current_output) {
+		wl_list_remove(&wlr_wl_pointer->output_destroy_listener.link);
+	}
+	wl_signal_add(&output->wlr_output.events.destroy,
+		&wlr_wl_pointer->output_destroy_listener);
 	wlr_wl_pointer->current_output = output;
 	output->enter_serial = serial;
 	wlr_wl_output_update_cursor(output);
@@ -49,7 +54,7 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 	struct wlr_wl_pointer *wlr_wl_pointer =
 		(struct wlr_wl_pointer *)dev->pointer;
 	if (!wlr_wl_pointer->current_output) {
-		wlr_log(L_ERROR, "pointer motion event without current output");
+		wlr_log(L_DEBUG, "pointer motion event without current output");
 		return;
 	}
 
@@ -231,6 +236,14 @@ static struct wlr_input_device *allocate_device(struct wlr_wl_backend *backend,
 	return wlr_device;
 }
 
+static void wlr_wl_pointer_handle_output_destroy(struct wl_listener *listener,
+		void *data) {
+	struct wlr_wl_pointer *wlr_wl_pointer =
+		wl_container_of(listener, wlr_wl_pointer, output_destroy_listener);
+	wlr_wl_pointer->current_output = NULL;
+	wl_list_remove(&wlr_wl_pointer->output_destroy_listener.link);
+}
+
 static void seat_handle_capabilities(void *data, struct wl_seat *wl_seat,
 		enum wl_seat_capability caps) {
 	struct wlr_wl_backend *backend = data;
@@ -243,6 +256,8 @@ static void seat_handle_capabilities(void *data, struct wl_seat *wl_seat,
 			wlr_log(L_ERROR, "Unable to allocate wlr_wl_pointer");
 			return;
 		}
+		wlr_wl_pointer->output_destroy_listener.notify =
+			wlr_wl_pointer_handle_output_destroy;
 
 		struct wlr_input_device *wlr_device;
 		if (!(wlr_device = allocate_device(backend, WLR_INPUT_DEVICE_POINTER))) {
--- a/include/backend/wayland.h
+++ b/include/backend/wayland.h
@@ -71,6 +71,7 @@ struct wlr_wl_pointer {
 	struct wlr_pointer wlr_pointer;
 	enum wlr_axis_source axis_source;
 	struct wlr_wl_backend_output *current_output;
+	struct wl_listener output_destroy_listener;
 };
 
 void wlr_wl_registry_poll(struct wlr_wl_backend *backend);
--- a/types/wlr_xdg_shell.c
+++ b/types/wlr_xdg_shell.c
@@ -131,6 +131,24 @@ static const struct wlr_keyboard_grab_interface xdg_keyboard_grab_impl = {
 	.cancel = xdg_keyboard_grab_cancel,
 };
 
+static void xdg_surface_destroy(struct wlr_xdg_surface *surface);
+
+static void wlr_xdg_popup_grab_handle_seat_destroy(
+		struct wl_listener *listener, void *data) {
+	struct wlr_xdg_popup_grab *xdg_grab =
+		wl_container_of(listener, xdg_grab, seat_destroy);
+
+	wl_list_remove(&xdg_grab->seat_destroy.link);
+
+	struct wlr_xdg_popup *popup, *next;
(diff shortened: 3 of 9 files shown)


backend/wayland/output.c
```diff
@@ -186,6 +186,11 @@ static void wlr_wl_output_destroy(struct wlr_output *wlr_output) {
 		wl_callback_destroy(output->frame_callback);
 	}
 
+	struct wlr_wl_pointer *pointer = output->backend->wlr_wl_pointer;
+	if (pointer && pointer->current_output == output) {
+		pointer->current_output = NULL;
+	}
```
SirCmpwn: Yeah I agree that this isn't great but I don't have any better ideas

types/wlr_xdg_shell.c
```diff
@@ -131,6 +131,23 @@ static const struct wlr_keyboard_grab_interface xdg_keyboard_grab_impl = {
 	.cancel = xdg_keyboard_grab_cancel,
 };
 
+static void xdg_surface_destroy(struct wlr_xdg_surface *surface);
```
SirCmpwn: Add a newline after this

types/wlr_xdg_shell_v6.c
```diff
@@ -131,6 +131,23 @@ static const struct wlr_keyboard_grab_interface xdg_keyboard_grab_impl = {
 	.cancel = xdg_keyboard_grab_cancel,
 };
 
+static void xdg_surface_destroy(struct wlr_xdg_surface_v6 *surface);
```
SirCmpwn: Here too

acrisci:
Did you find the leaks with an automated tool? If so, what command did you use so I can verify?

acrisci:
What is up with your system clock that these commits report being from December 2017?

martinetd:
> Did you find the leaks with an automated tool? If so, what command did you use so I can verify?

I run with address sanitizer, which can be enabled with `meson -db_sanitize=address`, then just running the program as usual. The leaks will be displayed on stderr so it might be a good idea to redirect that to a file.

> What is up with your system clock that these commits report being from December 2017?

My system clock works fine, thank you very much. The first commit was really authored on December 30 but only rebased and pushed recently with the other. The other commits are more recent. You can see the commit date with `git log --format=fuller`, blame github for being silly with what it displays.

backend/wayland/output.c
acrisci: The pattern we normally use here is for the `wlr_wl_pointer` to have a listener for the output destroy signal and run its own cleanup code when that event is emitted.

types/wlr_xdg_shell.c
```diff
@@ -131,6 +131,24 @@ static const struct wlr_keyboard_grab_interface xdg_keyboard_grab_impl = {
...
+
+	wl_list_remove(&xdg_grab->seat_destroy.link);
+
+	struct wlr_xdg_popup *popup, *next;
+	wl_list_for_each_safe(popup, next, &xdg_grab->popups, grab_link) {
+		xdg_surface_destroy(popup->base);
+	}
+
+	wl_list_remove(&xdg_grab->link);
+	free(xdg_grab);
```
acrisci: This event is the only place I see where `xdg_grab` is freed, but there are other places where the grab could be released. I think we are leaking grabs here.

The grab can be released when the grab ends. There is a seat event for this that you can try to hook into.

This should be refactored into a destroy func that runs whenever the grab can be released.

include/backend/wayland.h
```diff
@@ -32,6 +32,7 @@ struct wlr_wl_backend {
 	struct wl_shm *shm;
 	struct wl_seat *seat;
 	struct wl_pointer *pointer;
+	struct wlr_wl_pointer *wlr_wl_pointer;
```
acrisci: Why do we store both a `wlr_wl_pointer` and a `wl_pointer`?

backend/wayland/output.c
martinetd: That is what I first wanted to do, but there can be multiple outputs - how can you make a listener for an arbitrary number of outputs? Have a listener on output add and allocate destroy ones whenever an output is added seems overkill?

include/backend/wayland.h
martinetd: I couldn't figure how to access the wlr_wl_pointer from the wl_pointer.
The opposite is certainly possible, I didn't spend the time to look, happy to refactor all the rest around wlr_wl_pointer if you think it makes a difference.

backend/wayland/output.c
acrisci: You only need a listener for whatever output `current_output` points to.

include/backend/wayland.h
acrisci: If it only had one of these, it should make the code simpler because you don't have to worry about keeping them in sync, or people needing to choose which one to access depending on the situation which would be valuable.

types/wlr_xdg_shell.c
martinetd: I've had a fresh look, I don't think we ever remove the popup from `shell->popup_grabs` (the pendant of `xdg_grab->link`), so we never make the popup unreachable.

Basically, when a grab ends we keep the item in the list and reuse it next time someone wants a grab from the seat, so it's only ever required to free it when the seat is destroyed. If we destroy it all the time we'll need to reallocate it all the time, which I assume is why it was designed that way.

backend/wayland/output.c
martinetd: So change that listener everytime current output changes? I'm not sure I like evolving listeners much, but if you insist I will...

acrisci: Just make sure there is a setter and do all the listener logic in the setter. That is how we do it in many other places in the project.

types/wlr_xdg_shell.c
acrisci: Oh I remember that now. I think what you say is right.

backend/wayland/output.c
martinetd: It's possible, but I don't see anywhere else where that would be done. Could you point me at a place? (I don't need an example, I want a precedent because that's the argument you push forward. Give another argument like "do it because I said so" and I will)
I did this search: `git grep -C 4 wl_signal_add | grep -C 5 wl_signal_remove
`

acrisci: Another example is `display_destroy`.

martinetd: Done in a force-push. No longer need to remove wl_pointer as well, since I don't need to add wlr_wl_pointer to the backend. Tested and it seems to work, I dropped the level of "pointer motion event without current output" because we seem to be missing an enter event sometimes (rootston-in-rootston, so probably a bug in the parent rootston that I couldn't reproduce reliably, set `WLR_WL_OUTPUTS` to something >1 and close outputs to test)
<bike-shedding>I went over the 107 uses of `display_destroy` without seeing anywhere that seems to do that, but meh, I don't care anymore.</bike-shedding>

backend/wayland/wl_seat.c
```diff
@@ -49,7 +54,7 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 	struct wlr_wl_pointer *wlr_wl_pointer =
 		(struct wlr_wl_pointer *)dev->pointer;
 	if (!wlr_wl_pointer->current_output) {
-		wlr_log(L_ERROR, "pointer motion event without current output");
+		wlr_log(L_DEBUG, "pointer motion event without current output");
```
acrisci: You don't have to do it now, but I'd like to see all the pointer motion stuff have `L_TRACE`.

martinetd: I'd really rather not wait for `L_TRACE` to come in, as there is not even a PR open with it. If that's the only thing left I can either revert this or keep as is but let's please merge this?

acrisci: Yeah it's fine. Sometimes I just make these notes for other readers or myself.

acrisci:
:+1:

PR #730, opened 2018-03-17
Rootston: fix rotation heading (was: wlr_matrix: fix matrix_rotate)

### LINKED ISSUE #729: View rotation issues since matrix redesign
Since #722, rotating a view with meta-key + middle click has two problems:
 - the rotation is the opposite size from what it used to be/I'd expect, e.g. if I do a circle clockwise the view will rotate counter-clockwise.
 - The decoration doesn't follow rotation properly, the coordinate system must have changed a bit (an upside-down view is correct, but a view at 45° will touch the outside of the decoration)

martinetd:
Fixes #729.

Change rootston's `view->rotation` to be computed with normal expectation.

Test plan:
 * [x] Start normally and rotate a window
 * [x] Start rotated and rotate a window
 * [x] Start flipped/rotated and rotate a window

~~For reference, the old 3d code was this, that seems to agree:~~ The old behaviour was against usual conventions:
```c
        wlr_matrix_identity(mat);
        float _cos = cosf(radians);
        float _sin = sinf(radians);
        mat[mind(1, 1)] = _cos;
        mat[mind(1, 2)] = _sin;
        mat[mind(2, 1)] = -_sin;
        mat[mind(2, 2)] = _cos;
```

### DIFF
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -200,7 +200,7 @@ static void roots_cursor_update_position(struct roots_cursor *cursor,
 				uy = cursor->offs_y - oy;
 			int vx = cursor->cursor->x - ox,
 				vy = cursor->cursor->y - oy;
-			float angle = atan2(vx*uy - vy*ux, vx*ux + vy*uy);
+			float angle = atan2(ux*vy - uy*vx, vx*ux + vy*uy);
 			int steps = 12;
 			angle = round(angle/M_PI*steps) / (steps/M_PI);
 			view_rotate(view, cursor->view_rotation + angle);
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -557,8 +557,8 @@ static bool view_at(struct roots_view *view, double lx, double ly,
 		double ox = view_sx - (double)box.width/2,
 			oy = view_sy - (double)box.height/2;
 		// Rotated coordinates
-		double rx = cos(view->rotation)*ox - sin(view->rotation)*oy,
-			ry = cos(view->rotation)*oy + sin(view->rotation)*ox;
+		double rx = cos(view->rotation)*ox + sin(view->rotation)*oy,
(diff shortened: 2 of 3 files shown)


emersion:
Actually, the convention for rotation matrices is not the old code - it's the new one: https://en.wikipedia.org/wiki/Rotation_matrix

Would it be possible to instead search for all `wlr_matrix_rotate` calls (there shouldn't be that many) and put a minus in front of the angle?

martinetd:
I'm not comfortable having a minus pop out of nowhere at some point, I'd rather change how we compute view->rotation instead so we have `view->rotation` -> `wlr_matrix_project_box` -> `wlr_matrix_rotate` all use the same comvention

emersion:
Yeah, I know that view->rotation is negative when it should be positive (there are already minus signs in rootston).

martinetd:
> there are already minus signs in rootston

Meh.
One I see is `wlr_box_rotated_eounds(&box, -rotation, &rotated)`, but that uses fabs(sin(rotation)) that should be sign-agnostic (`sin(-x) = -sin(x)` and `cos(-x) = cos(x)`, right?)

We also have `rx = cos(-rotation)*ox - sin(-rotation)*oy` in rootston/output.c and `rx = cos(view->rotation)*ox - sin(view->rotation)*oy` in rootston/desktop.c.... Who bets `view_at` doesn't work right know :/

Spring cleaning coming up, I'll repost once I'm done and ask for testing.

martinetd:
All done, updated top post with test plan, seems to work for me.

emersion:
It seems that input events are messed up for rotated views.

martinetd:
Uh, so `view_at` was correctly at opposite sign, I guess the difference is that it computes pointer position inside the view and not the views' position inside the output.

This all made me notice damage on rotated views is also messed up, but it was before, will have a look separately.

@emersion: yeah noticed and edited previous comment, sorry for your time, it's fixed now.

Issue #685, opened 2018-02-26
surface layers


SirCmpwn:
Protocol: https://github.com/swaywm/sway-protocols/pull/7

PR #731, opened 2018-03-18
Implement wlr-layer-shell protocol & rendering

SirCmpwn:
Partially implements the protocol extension described in https://github.com/swaywm/wlr-protocols/pull/7

This PR is limited to rendering support

- [x] render frames
- [x] different layers
- [x] anchors
- [x] exclusive zones
- [x] margins

Updates #685


include/wlr/types/wlr_layer_shell.h
```diff
@@ -0,0 +1,102 @@
...
+	struct wl_resource *resource;
+	struct wl_client *client;
+	struct wl_list surfaces;
+
+	struct wl_list link; // wlr_layer_shell::clients
+};
+
+struct wlr_layer_surface_state {
+	// Client
+	uint32_t anchor;
```
acrisci: I think we should wrap the enums for this.

include/wlr/types/wlr_layer_shell.h
```diff
@@ -0,0 +1,102 @@
...
+	enum zwlr_layer_shell_v1_layer layer;
+
+	bool added, configured, mapped;
+	uint32_t configure_serial;
+	struct wl_event_source *configure_idle;
+	uint32_t configure_next_serial;
+	struct wl_list configure_list;
+
+	struct wlr_layer_surface_state next; // client protocol requests
+	struct wlr_layer_surface_state pending; // our configure requests
```
acrisci: Can we think of better names for next/pending that reflects these comments?

include/wlr/types/wlr_layer_shell.h
```diff
@@ -0,0 +1,102 @@
...
+};
+
+struct wlr_layer_shell *wlr_layer_shell_create(struct wl_display *display);
+void wlr_layer_shell_destroy(struct wlr_layer_shell *layer_shell);
+
+/**
+ * Notifies the layer surface to configure itself with this width/height. The
+ * layer_surface will signal its map event when the surface is ready to assume
+ * this size.
+ */
```
acrisci: Couldn't this just happen automatically? The protocol is pretty explicit about what width/height the layer surface should have and all the info to determine that can be contained within the shell with listeners on the output. That would take some of the math burden off of rootston.

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+        must also be treated as errors.
+
+        Clients can specify a namespace that defines the purpose of the layer
+        surface.
+      </description>
+      <arg name="id" type="new_id" interface="zwlr_layer_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+      <arg name="output" type="object" interface="wl_output"/>
+      <arg name="layer" type="uint" enum="layer" summary="layer to add this surface to"/>
+      <arg name="namespace" type="string" summary="namespace for the layer surface"/>
```
acrisci: `namespace` is not specified here. What are the plans for specifying namespaces in the future? Would we be better served with an enum of the possible namespaces here instead to make the specification easier? Or will this always be left unspecified and compositors can interpret it however they want?

include/wlr/types/wlr_layer_shell.h
SirCmpwn: We shouldn't store bitfields in an enum value type

include/wlr/types/wlr_layer_shell.h
SirCmpwn: Yes

protocol/wlr-layer-shell-unstable-v1.xml
SirCmpwn: No, that defeats the purpose of namespace. The goal is to make the protocol arbitrarily extensible in the future by letting the client specify what kind of tool it is and letting the user specify rules for e.g. resolving z-ordering issues

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+    </request>
+
+    <enum name="error">
+      <entry name="role" value="0" summary="wl_surface has another role"/>
+      <entry name="invalid_layer" value="1" summary="layer value is invalid"/>
+      <entry name="already_constructed" value="2" summary="wl_surface has a buffer attached or committed"/>
+      <entry name="request_denied" value="4" summary="the compositor has denied your request to use this layer"/>
```
acrisci: "4" looks weird here because it's not a bitfield. What about "3"?

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+  </interface>
+
+  <interface name="zwlr_layer_surface_v1" version="1">
+    <description summary="layer metadata interface">
+      An interface that may be implemented by a wl_surface, for surfaces that
+      are designed to be rendered as a layer of a stacked desktop-like
+      environment.
+
+      Layer surface state (anchor, exclusive zone, margin) is double-buffered.
+      Protocol requests modify the pending state, as opposed to the current
```
acrisci: This second sentence of this paragraph seems redundant to me. I think this paragraph could be reworded for clarity.

protocol/wlr-layer-shell-unstable-v1.xml
acrisci: Is the compositor allowed to violate other parts of the protocol when interpreting the namespace?

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+      new pending state is as documented for each related request.
+    </description>
+
+    <request name="set_events">
+      <description summary="set input events to receive">
+        Requests that the seat send input events for the specified input devices
+        to this layer surface.
+
+        Positional events (pointer and touch) will only be sent if the layer
+        surface is the top-most interactive surface, and only when the position
```
acrisci: What does "only when the position of these events are relative to the surface" mean?

SirCmpwn: The phrasing here is kind of awkward, I should change it. It's just reminding the reader that wl_pointer events are always relative to a surface, not in absolute coordinates.

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+
+    <enum name="anchor" bitfield="true">
+      <entry name="top" value="1" summary="the top edge of the anchor rectangle"/>
+      <entry name="bottom" value="2" summary="the bottom edge of the anchor rectangle"/>
+      <entry name="left" value="4" summary="the left edge of the anchor rectangle"/>
+      <entry name="right" value="8" summary="the right edge of the anchor rectangle"/>
+    </enum>
+  </interface>
+
+  <interface name="zwlr_layer_input_v1" version="1">
```
acrisci: What is this meant to accomplish that cannot be done with the surface input region and how does this interact with surface input region?

types/wlr_layer_shell.c
```diff
@@ -0,0 +1,384 @@
...
+static void layer_surface_handle_ack_configure(struct wl_client *client,
+		struct wl_resource *resource, uint32_t serial) {
+	struct wlr_layer_surface *surface = layer_surface_from_resource(resource);
+
+	bool found = false;
+	struct wlr_layer_surface_configure *configure, *tmp;
+	wl_list_for_each_safe(configure, tmp, &surface->configure_list, link) {
+		if (configure->serial < serial) {
+			layer_surface_configure_destroy(configure);
+		} else if (configure->serial == serial) {
```
acrisci: It looks to me like the way this is written, the comparison will happen for exactly one configure serial when it should consider all of them in the list.

types/wlr_layer_shell.c
```diff
@@ -0,0 +1,384 @@
...
+
+static void layer_surface_handle_set_exclusive_zone(struct wl_client *client,
+		struct wl_resource *resource, uint32_t zone) {
+	struct wlr_layer_surface *surface = layer_surface_from_resource(resource);
+	surface->next.exclusive_zone = zone;
+}
+
+static void layer_surface_handle_set_margin(struct wl_client *client,
+		struct wl_resource *resource, uint32_t top,
+		uint32_t right, uint32_t bottom, uint32_t left) {
```
acrisci: Should be `int32_t` for negatives.

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+    <request name="set_exclusive_zone">
+      <description summary="configures the exclusive geometry of this surface">
+        Requests that the compositor avoids occluding an area of the surface
+        with other surfaces. The compositor's use of this information is
+        implementation-dependent - do not assume that this region will not
+        actually be occluded.
+
+        This value is only meaningful if the surface is anchored to an edge,
+        rather than a corner. The zone is the number of pixels from the edge
+        that are considered exclusive.
```
acrisci: How do exclusive zones interact with the margin? Does "the edge" start at the margin?

include/wlr/types/wlr_layer_shell.h
emersion: I don't think this should happen automatically. If the compositor always draws a bar at the top on its own or implements another similar protocol, it may want to choose a different size given other constraints. We can provide default helpers though.

protocol/wlr-layer-shell-unstable-v1.xml
SirCmpwn: This is unrelated to the input region. The input region will affect which parts of the surface are elegible to receive pointer/touch events. Strictly speaking you can set the input region to 0x0@0,0 to avoid getting pointer/touch events but this is a fat hack. The layer input device lets you be more explicit about it, and also specifies keyboard and tablet semantics.

protocol/wlr-layer-shell-unstable-v1.xml
emersion: The protocol wording for this is usually "in surface-local coordinates".

types/wlr_layer_shell.c
SirCmpwn: Margins can't be negative

protocol/wlr-layer-shell-unstable-v1.xml
SirCmpwn: Yeah. Should specify this explicitly.

types/wlr_layer_shell.c
emersion: That's how it should be done. The client is free to ignore all configure requests but the last one.

protocol/wlr-layer-shell-unstable-v1.xml
emersion: Note that a region is not a rectangle. There are empty regions that contain no rectangles (and they don't need to contain a 0x0@0,0 rectangle), so it's not a hack, it's legit.

types/wlr_layer_shell.c
emersion: We already discussed this in the previous iteration and decided negative margins could be useful for animations.

protocol/wlr-layer-shell-unstable-v1.xml
SirCmpwn: Alright, fair enough. However, it's still less explicit and doesn't specify keyboard or tablet behavior.

protocol/wlr-layer-shell-unstable-v1.xml
SirCmpwn: No?

protocol/wlr-layer-shell-unstable-v1.xml
emersion: IIRC tablet behavior is defined to respect the wl_surface input region in the tablet protocol. But yeah, you're right, the keyboard is missing.

types/wlr_layer_shell.c
acrisci: The use case is for guake-style terminals that can use this to animate downward from the top of the screen.

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,279 @@
...
+        Requests that the seat send input events for the specified input devices
+        to this layer surface.
+
+        Positional events (pointer and touch) will only be sent if the layer
+        surface is the top-most interactive surface, and only when the position
+        of these events are relative to the surface. Enter and leave events will
+        be signalled normally in these cases.
+
+        Keyboard events will treat the layer surface as the only focused surface
+        on the seat. Upon requesting keyboard events, the layer surface will
```
acrisci: Does this mean the layer surface will receive all keyboard input for the seat all the time? Will this make rootston need to handle multiple windows with keyboard focus?

emersion: No, when the layer surface has keyboard focus, other views don't have it. No fallthrough.

include/wlr/types/wlr_layer_shell.h
acrisci: If you want people to read the field, it should have a wrapped type or else people have to make their own (or use literals).

protocol/wlr-layer-shell-unstable-v1.xml
acrisci: I fear this will be used for evil purposes.

types/wlr_layer_shell.c
acrisci: What if they are ack_configuring not the last one, but one further diwn in the list? They will get the error then.

emersion: Why? The configure will be in the list, `found` will be set to `true`… All previous configures will be destroyed, but the next ones will remain, so the client will be able to ack_configure them in the future.

rootston/layer_shell.c
```diff
@@ -0,0 +1,181 @@
...
+			| ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT;
+		if ((state->anchor & both_horiz) && box.width == -1) {
+			box.x = 0;
+			box.width = output_area.width;
+		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_LEFT)) {
+			box.x = output_area.x;
+		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_RIGHT)) {
+			box.x = output_area.width - box.width;
+		} else {
+			box.x = (output_area.width / 2) - (box.width / 2);
```
SirCmpwn: If box.width is -1 here it's a protocol error but not sure how best to express that. Probably should have wlr_layer_shell enforce this on commit but it's annoying

rootston/layer_shell.c
```diff
@@ -0,0 +1,181 @@
...
+		} else {
+			box.y = (output_area.height / 2) - (box.height / 2);
+		}
+		wlr_log(L_DEBUG, "arranged layer at %dx%d@%d,%d",
+				box.width, box.height, box.x, box.y);
+		roots_surface->geo = box;
+		apply_exclusive(&output_area, state->anchor, state->exclusive_zone);
+		if (box.width != (int)state->width
+				|| box.height != (int)state->height
+				|| !roots_surface->configured) {
```
SirCmpwn: Note to self to make this if statement prettier

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -0,0 +1,253 @@
...
+
+        If you pass -1 for either value, the compositor will assign it and
+        inform you of the assignment in the configure event. You must set your
+        anchor to opposite edges in the dimensions you omit; not doing so is a
+        protocol error. Both values are -1 by default.
+
+        Size is double-buffered, see wl_surface.commit.
+      </description>
+      <arg name="width" type="int"/>
+      <arg name="height" type="int"/>
```
emersion: For consistency with xdg-shell, these should be `uint` and the default value should be 0.

SirCmpwn:
Weird problem: on the DRM backend, specifying the output to `./examples/layer-shell` doesn't work. Specifying an output index that is greater than the number of outputs breaks as expected though. On the Wayland backend I can specify the output and it works correctly.

rootston/layer_shell.c
```diff
@@ -114,26 +114,58 @@ static void arrange_layers(struct wlr_output *_output) {
 	}
 }
 
-static void handle_destroy(struct wl_listener *listener, void *data) {
-	// TODO
+static void handle_output_destroy(struct wl_listener *listener, void *data) {
+	struct roots_layer_surface *layer =
+		wl_container_of(listener, layer, output_destroy);
+	layer->layer_surface->output = NULL;
+	wl_resource_destroy(layer->layer_surface->resource);
```
emersion: This is dangerous, the next request on this object will disconnect the client.

SirCmpwn: yeah TODO add close event to protocol

protocol/wlr-layer-shell-unstable-v1.xml
acrisci: The xdg-positioner stable has width/height as int and "If a negative size is set the invalid_input error is raised" for `set_anchor_rect` and `set_size`.

emersion: `set_max_size` and `set_min_size` say:

>If never set, or a value of zero in the request, means that the client has no expected maximum size in the given dimension.

And:

> The width and height must be greater than or equal to zero.

Also see `xdg_toplevel.configure`:

>If the width or height arguments are zero, it means the client should decide its own window dimension.

All of these are `int` though. Not sure why.

rootston/layer_shell.c
SirCmpwn: This has been replaced with a close event

SirCmpwn:
The rendering portion of this pull request is done. I'd like to see it reviewed as-is and split the rest of the work into additional patches.

rootston/output.c
```diff
@@ -493,6 +522,11 @@ static void render_output(struct roots_output *output) {
 		wlr_renderer_clear(renderer, clear_color);
 	}
 
+	render_layer(output, output_box, &data,
```
emersion: We don't want to render layers in fullscreen mode.

SirCmpwn: We want to render some layers in fullscreen mode. But yeah

examples/layer-shell.c
```diff
@@ -197,6 +200,22 @@ int main(int argc, char **argv) {
 				fprintf(stderr, "invalid anchor %s\n", optarg);
 				return 1;
 			}
+			break;
+		}
+		case 't':
+			alpha = atof(optarg);
+			break;
+		case 'm': {
+			char *endptr = optarg;
```
ascent12: sscanf

SirCmpwn: nah

include/rootston/layers.h
```diff
@@ -0,0 +1,24 @@
...
+struct roots_layer_surface {
+	struct wlr_layer_surface *layer_surface;
+	struct wl_list link;
+
+	struct wl_listener destroy;
+	struct wl_listener map;
+	struct wl_listener unmap;
+	struct wl_listener surface_commit;
+	struct wl_listener output_destroy;
+	struct wl_listener output_mode;
```
emersion: Need to listen for output transform too.

include/wlr/types/wlr_layer_shell.h
```diff
@@ -0,0 +1,108 @@
...
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal new_surface;
+	} events;
+
+	void *data;
+};
+
+struct wlr_layer_client {
```
emersion: Why not use the `wlr_layer_shell` struct (with `struct wl_list resources` + `struct wl_list surfaces` fields)? This introduces complexity for little benefit.

rootston/layer_shell.c
```diff
@@ -0,0 +1,324 @@
...
+				layer->geo.x, layer->geo.y, 0);
+		arrange_layers(wlr_output);
+		output_damage_from_local_surface(output, layer_surface->surface,
+				layer->geo.x, layer->geo.y, 0);
+	}
+}
+
+static void unmap(struct wlr_layer_surface *layer_surface) {
+	struct roots_layer_surface *layer = layer_surface->data;
+	if (layer->link.prev) {
```
emersion: meh

rootston/output.c
```diff
@@ -695,6 +733,14 @@ static void damage_from_surface(struct wlr_surface *surface,
 	}
 }
 
+void output_damage_from_local_surface(struct roots_output *output,
+		struct wlr_surface *surface, double ox, double oy, float rotation) {
+	struct wlr_output_layout_output *layout = wlr_output_layout_get(
+		output->desktop->layout, output->wlr_output);
+	damage_from_surface(surface, ox + layout->x, oy + layout->y,
+			rotation, output);
```
emersion: Should use `surface_for_each_surface` to apply damage from subsurfaces too.

rootston/layer_shell.c
```diff
@@ -0,0 +1,324 @@
...
+	roots_surface->output_mode.notify = handle_output_mode;
+	wl_signal_add(&layer_surface->output->events.mode,
+		&roots_surface->output_mode);
+
+	roots_surface->destroy.notify = handle_destroy;
+	wl_signal_add(&layer_surface->events.destroy, &roots_surface->destroy);
+	roots_surface->map.notify = handle_map;
+	wl_signal_add(&layer_surface->events.map, &roots_surface->map);
+	roots_surface->unmap.notify = handle_unmap;
+	wl_signal_add(&layer_surface->events.unmap, &roots_surface->unmap);
```
emersion: Need to listen for new subsurfaces, create `view_child` (might want to rename that), listen for commit and apply damage. See how other shells do this.

rootston/layer_shell.c
```diff
@@ -0,0 +1,324 @@
...
+	struct roots_view *view;
+	wl_list_for_each(view, &output->desktop->views, link) {
+		if (view->maximized) {
+			view_arrange_maximized(view);
+		}
+	}
+
+	// Arrange non-exlusive surfaces from top->bottom
+	wlr_output_effective_resolution(output->wlr_output,
+			&usable_area.width, &usable_area.height);
```
emersion: Reset `x` and `y` to 0?

rootston/layer_shell.c
```diff
@@ -0,0 +1,324 @@
...
+static void arrange_layer(struct wlr_output *output, struct wl_list *list,
+		struct wlr_box *usable_area, bool exclusive) {
+	struct roots_layer_surface *roots_surface;
+	struct wlr_box full_area = { 0 };
+	wlr_output_effective_resolution(output,
+			&full_area.width, &full_area.height);
+	wl_list_for_each(roots_surface, list, link) {
+		struct wlr_layer_surface *layer = roots_surface->layer_surface;
+		struct wlr_layer_surface_state *state = &layer->current;
+		if (exclusive) {
```
emersion:
```c
if (exclusive == (state->exclusive_zone > 0))
```

rootston/layer_shell.c
```diff
@@ -0,0 +1,324 @@
...
+				*edges[i].positive_axis += exclusive + edges[i].margin;
+			}
+			if (edges[i].negative_axis) {
+				*edges[i].negative_axis -= exclusive + edges[i].margin;
+			}
+		}
+	}
+}
+
+static void arrange_layer(struct wlr_output *output, struct wl_list *list,
```
emersion: I want to bring this into wlroots somehow.

types/wlr_layer_shell.c
```diff
@@ -0,0 +1,438 @@
...
+	surface->client_pending.margin.right = right;
+	surface->client_pending.margin.bottom = bottom;
+	surface->client_pending.margin.left = left;
+}
+
+static void layer_surface_handle_set_keyboard_interactivity(
+		struct wl_client *client, struct wl_resource *resource,
+		uint32_t interactive) {
+	struct wlr_layer_surface *surface = layer_surface_from_resource(resource);
+	surface->client_pending.keyboard_interactive = interactive == 1;
```
emersion: I'd use `!!interactive` here, requiring strict equality is a little weird.

types/wlr_layer_shell.c
```diff
@@ -0,0 +1,438 @@
...
+static void layer_surface_unmap(struct wlr_layer_surface *surface) {
+	// TODO: probably need to ungrab before this event
+	wlr_signal_emit_safe(&surface->events.unmap, surface);
+
+	struct wlr_layer_surface_configure *configure, *tmp;
+	wl_list_for_each_safe(configure, tmp, &surface->configure_list, link) {
+		layer_surface_configure_destroy(configure);
+	}
+
+	surface->added = surface->configured = surface->mapped = false;
```
emersion: `added` shouldn't be reset when unmapped

types/wlr_layer_shell.c
```diff
@@ -0,0 +1,438 @@
...
+	configure->serial = surface->configure_next_serial;
+	configure->state.actual_width = surface->server_pending.actual_width;
+	configure->state.actual_height = surface->server_pending.actual_height;
+
+	zwlr_layer_surface_v1_send_configure(surface->resource,
+			configure->serial, configure->state.actual_width,
+			configure->state.actual_height);
+}
+
+static uint32_t wlr_layer_surface_schedule_configure(
```
emersion: The reason we have this in xdg-shell is that there are multiple configure events. The compositor sends the role-specific configure event (and maybe others because of extensions like the decoration protocol) and then ends the configure sequence with a `xdg_surface.configure`. I don't think we need to schedule configure events for layer-shell, we can directly send it since there's only one.

types/wlr_layer_shell.c
```diff
@@ -0,0 +1,438 @@
...
+}
+
+void wlr_layer_surface_configure(struct wlr_layer_surface *surface,
+		uint32_t width, uint32_t height) {
+	surface->server_pending.actual_width = width;
+	surface->server_pending.actual_height = height;
+	wlr_layer_surface_schedule_configure(surface);
+}
+
+void wlr_layer_surface_close(struct wlr_layer_surface *surface) {
```
emersion: Don't close if already closed

types/wlr_seat.c
```diff
@@ -1257,8 +1257,9 @@ bool wlr_seat_touch_has_grab(struct wlr_seat *seat) {
 }
 
 bool wlr_seat_validate_grab_serial(struct wlr_seat *seat, uint32_t serial) {
-	return serial == seat->pointer_state.grab_serial ||
-		serial == seat->touch_state.grab_serial;
+	return true;
```
emersion: TODO: fix this, probably in another PR

rootston/layer_shell.c
SirCmpwn: Hm, probably not.

rootston/layer_shell.c
```diff
@@ -0,0 +1,324 @@
...
+		if ((state->anchor & both_vert) == both_vert) {
+			box.y += state->margin.top;
+			box.height -= state->margin.top + state->margin.bottom;
+		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_TOP)) {
+			box.y += state->margin.top;
+		} else if ((state->anchor & ZWLR_LAYER_SURFACE_V1_ANCHOR_BOTTOM)) {
+			box.y -= state->margin.bottom;
+		}
+		// Apply
+		roots_surface->geo = box;
```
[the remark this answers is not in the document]
SirCmpwn: They're unsigned

rootston/layer_shell.c
SirCmpwn: ack

include/wlr/types/wlr_layer_shell.h
SirCmpwn: Sounds like a good plan

types/wlr_layer_shell.c
SirCmpwn: aight can do

rootston/layer_shell.c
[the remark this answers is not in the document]
emersion: Need to add overflow checks

rootston/layer_shell.c
emersion: Why not? We could just provide helpers, something similar to the xdg-positionner stuff.

SirCmpwn: I don't see a good way to break this down without removing the ability for compositors to customize how these are laid out.

SirCmpwn:
I think this is ready for review.

emersion:
I just understood why this fucking bug we've been trying to figure out for an hour or so is happening.

In the client, if animate is set to true, you update the margin and commit. This is a commit with an empty damage region. In the compositor, on commit you call `output_damage_from_local_surface` before arranging the layer surface. But this function takes damage from the surface and applies it. When the margin is updated (ie. the surface moves) we don't care about the damage, we just want to damage the whole damn surface.

Commenting `wl_surface_commit(wl_surface);` right after `zwlr_layer_surface_v1_set_margin` in the client makes the client work, because `eglSwapBuffers` will damage the whole surface and commit.

The best way to fix this bug is to keep a copy of the old layer geo before arranging layers, and if it has changed, damage the whole surface (@ old geo and new geo). If it hasn't changed, just take the surface damage (no need to take it twice).

SirCmpwn:
Hack rolled back and proper fix added. This is ready for review.

include/wlr/types/wlr_output_damage.h
```diff
@@ -10,7 +10,7 @@
  * damage tracking to work with triple buffering, a history of two frames is
  * required.
  */
-#define WLR_OUTPUT_DAMAGE_PREVIOUS_LEN 2
+#define WLR_OUTPUT_DAMAGE_PREVIOUS_LEN 3
```
emersion: Please rollback

SirCmpwn:
Render fullscreen shell surfaces between top and overlay

SirCmpwn:
If a surface is moved due to another surface, it needs to be damaged

include/wlr/types/wlr_output_damage.h
SirCmpwn: done

rootston/layer_shell.c
SirCmpwn: meh

rootston/output.c
SirCmpwn: Doing subsurfaces later, added a todo

include/rootston/layers.h
emersion: Bump

rootston/desktop.c
```diff
@@ -187,6 +188,32 @@ static struct wlr_output *view_get_output(struct roots_view *view) {
...
+void view_arrange_maximized(struct roots_view *view) {
+	struct wlr_box view_box;
+	view_get_box(view, &view_box);
+
+	view->maximized = true;
+	view->saved.x = view->x;
+	view->saved.y = view->y;
+	view->saved.rotation = view->rotation;
+	view->saved.width = view_box.width;
+	view->saved.height = view_box.height;
```
emersion: We only want to save the position + size + rotation of the view if it isn't yet maximized. This function can be called when the view is already maximized from the layer surface code.

SirCmpwn: Oh that explains the problem I was hitting earlier

rootston/layer_shell.c
```diff
@@ -0,0 +1,338 @@
...
+
+static void handle_destroy(struct wl_listener *listener, void *data) {
+	struct roots_layer_surface *layer = wl_container_of(
+			listener, layer, destroy);
+	if (layer->layer_surface->mapped) {
+		unmap(layer->layer_surface);
+	}
+	wl_list_remove(&layer->output_destroy.link);
+	wl_list_remove(&layer->output_mode.link);
+	wl_list_remove(&layer->output_transform.link);
```
emersion: We need to remove other listeners too, like the surface commit one.

rootston/layer_shell.c
emersion: Try to remove this check. If it doesn't work, this should:

```c
wl_list_remove(&layer->link);
wl_list_init(&layer->link);
```

acrisci:
:+1:

PR #733, opened 2018-03-18
Don't use OpenGL matrix transposition

emersion:
See https://github.com/swaywm/wlroots/commit/f1cbbac636d59d22373883f6bc85f8863e01c8c5#r28140751

We have two potential solutions:
- Transpose right before binding, this introduces a little overhead
- Use column-major matrices, this makes matrix functions a little confusing because the indices are not in the order we'd expect

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -178,7 +178,8 @@ static void free_eglimage(struct gbm_bo *bo, void *data) {
 	free(tex);
 }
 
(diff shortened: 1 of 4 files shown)


emersion:
This PR is ready for review.

I've decided to transpose explicitly because the overhead for a 3x3 matrix is negligible and GL stuff is already complicated enough, no need to use column-major matrices to make the code not only difficult to understand but also unreadable.

Issue #558, opened 2018-01-12
Renderer redesign v4


SirCmpwn:
Goals:

- [x] 2D matrices
- [x] Opacity uniform
- [ ] Multiple renderers at the same time
- [x] Simple interface for common tasks

acrisci:
My goals:

- Simple interface for common tasks
- Multiple renderers at the same time
- Partial screen updates (damage tracking)

ascent12:
Regarding multiple renderers at the same time, you'll certainly want to be using dma-bufs. It's a zero-copy way of sharing buffers between processes or devices.
EGL has a couple of relevant extensions: [EGL_EXT_image_dma_buf_import_modifiers](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt) and [EGL_MESA_image_dma_buf_export](https://www.khronos.org/registry/EGL/extensions/MESA/EGL_MESA_image_dma_buf_export.txt).
We're already using dma-bufs for multi-GPU, but it should work more generally.

My original renderer redesign already could do some of these things, but it wasn't written to use multi-planar formats (i.e. YUV) properly.

emersion:
#571 is a first step towards multiple renderers at the same time: it always uses the output's renderer (instead of the global `server->renderer`). When using multiple renderers, far less GL errors are triggered. The only place `server->renderer` is still used is to upload textures in `wlr_surface`. Uploading a texture using one renderer and drawing it using another one probably doesn't work.

emersion:
Bits that still need to be ported from the old renderer redesign PR (#319):
- [x] Screenshooter
- [x] Matrix
- [ ] Renderer (GL_KHR_debug, remove global state)
- [ ] Texture (immutable, always use EGL images)
- [ ] `wlr_render_subtexture` and use that for damage tracking
- [ ] Check nobody's using raw GL calls anymore
- [ ] [EGL_KHR_debug](https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_debug.txt)

Things that could be done after that:
- Remove wlr_render from wlr_surface. They are causing wlr_render to be a non-optional type.
- A `wlr_tex_export_dmabuf` function, which may be useful for future multi-GPU work.

ascent12:
>always use EGL images

Unfortunately, some drivers are really picky (intel) and don't allow using EGLimages created from a "local" texture, so I had to remove that in my renderer PR.

emersion:
[...]

PR #738, opened 2018-03-20
Redesign GLES2 renderer

### LINKED ISSUE #55: Use GL_KHR_debug

emersion:
This ports the GLES2 `wlr_renderer` part of #319. This PR tries not to touch too much to the texture part, which will be redesigned in another PR.

- [x] Use [GL_KHR_debug](https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_debug.txt)
- [x] Remove global state
- [x] ~~Investigate if it's possible to remove some shaders (is the RGBX shader required at all?)~~
- [x] ~~Save output projection in `wlr_renderer_begin`, remove it from `wlr_render_texture`. Should we apply it automatically when the user provides a matrix?~~
- [x] Remove raw GL calls outside the renderer

Fixes #55
Updates #558

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -582,11 +582,8 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 			return false;
 		}
 
-		// OpenGL will read the pixels out upside down,
-		// so we need to flip the image vertically
-		enum wl_output_transform transform = wlr_output_transform_compose(
-			wlr_output_transform_invert(output->transform),
-			WL_OUTPUT_TRANSFORM_FLIPPED_180);
+		enum wl_output_transform transform =
+			wlr_output_transform_invert(output->transform);
 		wlr_matrix_projection(plane->matrix, plane->surf.width,
 			plane->surf.height, transform);
 
@@ -649,11 +646,8 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 		wlr_render_texture(rend, plane->wlr_tex, plane->matrix, 0, 0, 1.0f);
 		wlr_renderer_end(rend);
 
-		// TODO: remove these raw GL calls
-		glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, bo_stride);
-		glReadPixels(0, 0, plane->surf.width, plane->surf.height, GL_BGRA_EXT,
-			GL_UNSIGNED_BYTE, bo_data);
-		glPixelStorei(GL_UNPACK_ROW_LENGTH_EXT, 0);
+		wlr_renderer_read_pixels(rend, WL_SHM_FORMAT_ARGB8888, bo_stride,
+			plane->surf.width, plane->surf.height, 0, 0, 0, 0, bo_data);
 
 		wlr_drm_surface_swap_buffers(&plane->surf, NULL);
 
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -643,14 +643,13 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
 		wlr_texture_upload_pixels(plane->wlr_tex, WL_SHM_FORMAT_ARGB8888,
 			stride, width, height, buf);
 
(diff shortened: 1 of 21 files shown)


render/wlr_renderer.c
```diff
@@ -17,8 +17,13 @@ void wlr_renderer_destroy(struct wlr_renderer *r) {
...
 
-void wlr_renderer_begin(struct wlr_renderer *r, struct wlr_output *o) {
-	r->impl->begin(r, o);
+void wlr_renderer_begin_raw(struct wlr_renderer *r, uint32_t width,
+		uint32_t height) {
+	r->impl->begin(r, width, height);
+}
+
+void wlr_renderer_begin(struct wlr_renderer *r, struct wlr_output *output) {
+	wlr_renderer_begin_raw(r, output->width, output->height);
```
SirCmpwn: I would sooner just make it take a width/height always instead of an output

emersion:
>I would sooner just make it take a width/height always instead of an output

Right, I've done that.

This PR is now ready for review. Let me know if you think something else needs a redesign in `wlr_renderer`.

wlroots.syms
```diff
@@ -39,6 +39,7 @@ WLROOTS_0_0_0 {
         wlr_libinput_tablet_tool_create;
         wlr_libinput_touch_create;
         wlr_log_stderr;
+        wlr_renderer_begin_raw;
```
SirCmpwn: remove

backend/drm/drm.c
```diff
@@ -643,14 +643,13 @@ static bool wlr_drm_connector_set_cursor(struct wlr_output *output,
...
+		wlr_renderer_begin(rend, plane->surf.width, plane->surf.height);
+		wlr_renderer_clear(rend, (float[]){ 0.0, 0.0, 0.0, 0.0 });
+		wlr_render_texture(rend, plane->wlr_tex, plane->matrix, 0, 0, 1.0f);
+		wlr_renderer_end(rend);
 
-		wlr_render_texture(plane->surf.renderer->wlr_rend, plane->wlr_tex,
-			plane->matrix, 0, 0, 1.0f);
-
-		glFinish();
+		// TODO: remove these raw GL calls
```
ascent12: Didn't I originally write a function for this?

emersion: Yeah, but it's reading pixel lines one by one in a loop. I wonder if I:
* Should use it anyway
* Should add a `bool invert_y` parameter to be able to use only one GL call
* Should keep it as is until there's a better solution

emersion:
>This LGTM, I ran some tests with dmabuf and egl and didn't see any regressions.

Nice, thanks for testing!

backend/drm/drm.c
ascent12: There is a way it can be done in one call (GL_PACK_ROW_LENGTH), but as usual, it requires a version later than GLES2 (and is not available via extensions), so I was stuck doing it the stupid, inefficient way.

emersion: Yeah. But you would still need to invert the y axis after that right? I didn't find a way to do this with GLES3 (it's possible to do so when rendering, not when reading pixels). Maybe we should make `wlr_renderer_read_pixels` always read pixels with an inverted y axis?

Anyway, I've pushed a commit to use the inefficient way.

SirCmpwn:
Tested, works, please fix conflicts

emersion:
Done!

SirCmpwn:
Thanks!

PR #744, opened 2018-03-24
Redesign wlr_texture

emersion:
This PR ports the texture part of #319.

- Textures are now immutable (apart from those created from raw
  pixels), no more invalid textures
- Move all wl_drm stuff in wlr_renderer
- Most wlr_texture fields are now private
- Remove some duplicated DMA-BUF code in the DRM backend
- Add more assertions
- Stride is now always given in bytes rather than in pixels
- Drop wl_shm functions

Test plan: try `weston-simple-shm`, `weston-simple-egl` and `weston-simple-dmabuf-drm` (with `--import-format` set to XRGB and then set to NV12). Try GTK clients, `weston-terminal` and Xwayland clients.

Updates #558

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -14,6 +14,10 @@
 #include "backend/drm/drm.h"
 #include "glapi.h"
 
+#ifndef DRM_FORMAT_MOD_LINEAR
+#define DRM_FORMAT_MOD_LINEAR 0
+#endif
+
 bool wlr_drm_renderer_init(struct wlr_drm_backend *drm,
 		struct wlr_drm_renderer *renderer) {
 	renderer->gbm = gbm_create_device(drm->fd);
@@ -195,7 +199,7 @@ static struct wlr_texture *get_tex_for_bo(struct wlr_drm_renderer *renderer,
 	};
 	attribs.offset[0] = 0;
 	attribs.stride[0] = gbm_bo_get_stride_for_plane(bo, 0);
-	attribs.modifier[0] = 0;
+	attribs.modifier[0] = DRM_FORMAT_MOD_LINEAR;
 	attribs.fd[0] = gbm_bo_get_fd(bo);
 
 	tex->tex = wlr_texture_from_dmabuf(renderer->wlr_rend, &attribs);
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -178,47 +178,33 @@ static void free_eglimage(struct gbm_bo *bo, void *data) {
 static struct wlr_texture *get_tex_for_bo(struct wlr_drm_renderer *renderer,
 		struct gbm_bo *bo) {
 	struct tex *tex = gbm_bo_get_user_data(bo);
-	if (tex) {
+	if (tex != NULL) {
 		return tex->tex;
 	}
 
-	// TODO: use wlr_texture_upload_dmabuf instead
-
-	tex = malloc(sizeof(*tex));
-	if (!tex) {
-		wlr_log_errno(L_ERROR, "Allocation failed");
+	tex = calloc(1, sizeof(struct tex));
+	if (tex == NULL) {
 		return NULL;
 	}
 
-	tex->egl = &renderer->egl;
-
-	int dmabuf_fd = gbm_bo_get_fd(bo);
-	uint32_t width = gbm_bo_get_width(bo);
-	uint32_t height = gbm_bo_get_height(bo);
-
-	EGLint attribs[] = {
-		EGL_WIDTH, width,
-		EGL_HEIGHT, height,
-		EGL_LINUX_DRM_FOURCC_EXT, gbm_bo_get_format(bo),
-		EGL_DMA_BUF_PLANE0_FD_EXT, dmabuf_fd,
-		EGL_DMA_BUF_PLANE0_OFFSET_EXT, gbm_bo_get_offset(bo, 0),
-		EGL_DMA_BUF_PLANE0_PITCH_EXT, gbm_bo_get_stride_for_plane(bo, 0),
-		EGL_IMAGE_PRESERVED_KHR, EGL_FALSE,
-		EGL_NONE,
+	struct wlr_dmabuf_buffer_attribs attribs = {
+		.n_planes = 1,
+		.width = gbm_bo_get_width(bo),
+		.height = gbm_bo_get_height(bo),
+		.format = gbm_bo_get_format(bo),
 	};
+	attribs.offset[0] = 0;
+	attribs.stride[0] = gbm_bo_get_stride_for_plane(bo, 0);
+	attribs.modifier[0] = 0;
+	attribs.fd[0] = gbm_bo_get_fd(bo);
 
-	tex->img = eglCreateImageKHR(renderer->egl.display, EGL_NO_CONTEXT,
-		EGL_LINUX_DMA_BUF_EXT, NULL, attribs);
-	if (!tex->img) {
-		wlr_log(L_ERROR, "Failed to create EGL image");
-		abort();
+	tex->tex = wlr_texture_from_dmabuf(renderer->wlr_rend, &attribs);
+	if (tex->tex == NULL) {
+		free(tex);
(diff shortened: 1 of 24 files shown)


backend/drm/renderer.c
```diff
@@ -178,47 +178,35 @@ static void free_eglimage(struct gbm_bo *bo, void *data) {
...
+	attribs.modifier[0] = 0;
+	attribs.fd[0] = gbm_bo_get_fd(bo);
 
-	tex->img = eglCreateImageKHR(renderer->egl.display, EGL_NO_CONTEXT,
-		EGL_LINUX_DMA_BUF_EXT, NULL, attribs);
-	if (!tex->img) {
-		wlr_log(L_ERROR, "Failed to create EGL image");
-		abort();
-	}
+	// TODO: add EGL_IMAGE_PRESERVED_KHR, EGL_FALSE to attribs?
```
ascent12: It's a very minor optimisation thing. But even then, I think it defaults to false anyway.
I think I originally put it there, because I was experimenting what different it made setting it to true.

backend/drm/renderer.c
```diff
@@ -178,47 +178,35 @@ static void free_eglimage(struct gbm_bo *bo, void *data) {
...
+	struct wlr_dmabuf_buffer_attribs attribs = {
+		.n_planes = 1,
+		.width = gbm_bo_get_width(bo),
+		.height = gbm_bo_get_height(bo),
+		.format = gbm_bo_get_format(bo),
 	};
+	attribs.offset[0] = 0;
+	attribs.stride[0] = gbm_bo_get_stride_for_plane(bo, 0);
+	attribs.modifier[0] = 0;
+	attribs.fd[0] = gbm_bo_get_fd(bo);
```
ascent12: These lines can be made part of the initialisation above. You can mix in the fields and array indices exactly how you think you would work.
It would be possible to add multi-planar support here, but there is so much else that would need to be done in the DRM backend for that to even be useful, so I'm fine with just putting that off until later.

examples/touch.c
```diff
@@ -110,13 +113,13 @@ int main(int argc, char *argv[]) {
 		wlr_log(L_ERROR, "Could not start compositor, OOM");
 		exit(EXIT_FAILURE);
 	}
-	state.cat_texture = wlr_render_texture_create(state.renderer);
+	state.cat_texture = wlr_texture_from_pixels(state.renderer,
+		WL_SHM_FORMAT_ARGB8888, cat_tex.width, cat_tex.width, cat_tex.height,
```
SirCmpwn: Stride is used inconsistently, is it necessary here? I don't even know how stride works in this library, what are pixels and how do they work

backend/drm/renderer.c
SirCmpwn: Yeah let's not

render/wlr_renderer.c
```diff
@@ -6,6 +7,15 @@
...
 		const struct wlr_renderer_impl *impl) {
+	assert(impl->begin);
+	assert(impl->clear);
+	assert(impl->scissor);
+	assert(impl->render_texture_with_matrix);
+	assert(impl->render_quad);
+	assert(impl->render_ellipse);
+	assert(impl->formats);
+	assert(impl->format_supported);
+	assert(impl->texture_from_pixels);
```
SirCmpwn: :+1:, but let's not make a habit of this - some types are allowed to implement a subset of the interface.

emersion: Actually, I made some of them optional - mainly the `wl_drm` stuff.

include/wlr/render/egl.h
```diff
@@ -46,37 +46,25 @@ void wlr_egl_finish(struct wlr_egl *egl);
...
-EGLImageKHR wlr_egl_create_image(struct wlr_egl *egl,
-		EGLenum target, EGLClientBuffer buffer, const EGLint *attribs);
-
 /**
  * Creates an egl image from the given dmabuf attributes. Check usability
  * of the dmabuf with wlr_egl_check_import_dmabuf once first.
  */
 EGLImageKHR wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
-		struct wlr_dmabuf_buffer_attribs *attributes);
+	struct wlr_dmabuf_buffer_attribs *attributes);
```
emersion: I wonder if we should move this into `gles2/texture.c`, just like the wl-drm one.

SirCmpwn: I don't feel strongly either way.

emersion: I'm anticipating a bit wrt. Vulkan and WSI here - we can just let it as is for now and see what happens with `wlr_egl` in the future.

Your idea about merging `wlr_egl` and `wlr_renderer` could solve it too.

SirCmpwn: I retract my idea of merging wlr_egl and wlr_renderer.

backend/drm/renderer.c
```diff
@@ -178,47 +178,33 @@ static void free_eglimage(struct gbm_bo *bo, void *data) {
...
-	if (!tex->img) {
-		wlr_log(L_ERROR, "Failed to create EGL image");
-		abort();
+	attribs.offset[0] = 0;
+	attribs.stride[0] = gbm_bo_get_stride_for_plane(bo, 0);
+	attribs.modifier[0] = 0;
+	attribs.fd[0] = gbm_bo_get_fd(bo);
+
+	tex->tex = wlr_texture_from_dmabuf(renderer->wlr_rend, &attribs);
+	if (tex->tex == NULL) {
```
agx: I thought `if (!tex->tex)` is the preferred style? (same for other occurances)

backend/drm/renderer.c
```diff
@@ -178,47 +178,33 @@ static void free_eglimage(struct gbm_bo *bo, void *data) {
...
 	};
-
-	tex->img = eglCreateImageKHR(renderer->egl.display, EGL_NO_CONTEXT,
-		EGL_LINUX_DMA_BUF_EXT, NULL, attribs);
-	if (!tex->img) {
-		wlr_log(L_ERROR, "Failed to create EGL image");
-		abort();
+	attribs.offset[0] = 0;
+	attribs.stride[0] = gbm_bo_get_stride_for_plane(bo, 0);
+	attribs.modifier[0] = 0;
```
agx: DRM_FORMAT_MOD_LINEAR is the right modifier here I think.

include/render/gles2.h
```diff
@@ -38,21 +38,49 @@ struct wlr_gles2_renderer {
...
 struct wlr_gles2_texture {
 	struct wlr_texture wlr_texture;
 
-	struct wlr_egl *egl;
-	GLuint tex_id;
-	const struct gles2_pixel_format *pixel_format;
+	struct wlr_gles2_renderer *renderer;
+	enum wlr_gles2_texture_type type;
+	int width, height;
+	bool has_alpha;
```
agx: Is this (and others here) a propery of GLES2 textures and not rather textures in general (wlr_texture)?

agx:
Did some tests with X11 backend and DRM backend on etnaviv and it looked good.

backend/drm/renderer.c
SirCmpwn: I use `if (!foobar)` but I don't care if other people do it explicitly, w/e

include/wlr/render/wlr_texture.h
```diff
@@ -5,62 +5,49 @@
...
- * returns. Under some circumstances, this function may re-upload the entire
- * buffer - therefore, the entire buffer must be valid.
+ * Create a new texture from a wayland DRM resource. The returned texture is
+ * immutable.
  */
-bool wlr_texture_update_pixels(struct wlr_texture *surf,
-	enum wl_shm_format format, int stride, int x, int y,
-	int width, int height, const unsigned char *pixels);
+struct wlr_texture *wlr_texture_from_wl_drm(struct wlr_renderer *renderer,
+	struct wl_resource *data);
```
acrisci: call this `buffer_resource`.

acrisci:
:+1:

PR #750, opened 2018-03-26
render: add simple functions to render rectangles and ellipses

emersion:
This PR ports the simple renderer interface from #319.

Updates #558

### DIFF
--- a/include/wlr/render/wlr_renderer.h
+++ b/include/wlr/render/wlr_renderer.h
@@ -36,14 +36,24 @@ bool wlr_render_texture(struct wlr_renderer *r, struct wlr_texture *texture,
 bool wlr_render_texture_with_matrix(struct wlr_renderer *r,
 	struct wlr_texture *texture, const float matrix[static 9], float alpha);
 /**
- * Renders a solid quad in the specified color.
+ * Renders a solid rectangle in the specified color.
  */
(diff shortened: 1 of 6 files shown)


include/wlr/render/wlr_renderer.h
```diff
@@ -36,14 +36,24 @@ bool wlr_render_texture(struct wlr_renderer *r, struct wlr_texture *texture,
 bool wlr_render_texture_with_matrix(struct wlr_renderer *r,
 	struct wlr_texture *texture, const float matrix[static 9], float alpha);
 /**
- * Renders a solid quad in the specified color.
+ * Renders a solid rectangle in the specified color.
  */
-void wlr_render_colored_quad(struct wlr_renderer *r,
+void wlr_render_rect(struct wlr_renderer *r, const struct wlr_box *box,
```
SirCmpwn: I prefer the name quad, after all the matrix could deform it into a non-rectangular shape

emersion: Then we should rename `ellipse` to something else.

SirCmpwn: nah cause no matter how deformed it gets it'll still be an ellipse

emersion: Our ellipse is defined by a rectangle (and just a rectangle), so it's the same issue.

PR #751, opened 2018-03-26
Fix damage tracking for rotated surfaces

### LINKED ISSUE #714: Surface damage is buggy on rotated windows
Running the `weston-simple-damage` demo in rootston works, unless you rotate the window...
Green ball tearing
My guess is the damaged region's coordinates are not rotated along with the window, or not properly at least, causing tearing or the ball flickering between 2 frames, depending on the rotation. Only 0° rotation works properly.
May be linked to #558

emersion:
It was broken because the damage extents were rotated about its
own center, not about the center of the surface.

This adds a new wlr_region_rotated_bounds that rotates regions.
This allows us to have only one code path (for both non-rotated
views and rotated views) and optimizes rendering for rotated
views.

Fixes #714

### DIFF
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -486,6 +486,9 @@ static void render_output(struct roots_output *output) {
 		goto renderer_end;
 	}
 
+	// Uncomment this line to debug damage tracking
+	//wlr_renderer_clear(renderer, (float[]){1, 1, 0, 0});
+
 	int nrects;
 	pixman_box32_t *rects = pixman_region32_rectangles(&damage, &nrects);
 	for (int i = 0; i < nrects; ++i) {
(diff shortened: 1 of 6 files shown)


rootston/output.c
```diff
@@ -486,6 +486,9 @@ static void render_output(struct roots_output *output) {
 		goto renderer_end;
 	}
 
+	// Uncomment this line to debug damage tracking
+	//wlr_renderer_clear(renderer, (float[]){1, 1, 0, 0});
```
SirCmpwn: I would rather see a startup flag to give to rootston for this purpose

SirCmpwn:
tyvm

PR #753, opened 2018-03-26
xdg-shell-v6: don't destroy role resources on unmap

### LINKED ISSUE #747: Segfault in wlr_surface after null commit
See https://github.com/swaywm/wlroots/pull/741#issuecomment-375483525

emersion:
The motivation for this is:
- `get_popup` and `get_toplevel` allocate role-specific resources.
- On the first non-null commit, the surface gets mapped.
- On a null commit, the surface gets unmapped. It can be mapped
  again with a non-null commit.
- When the role object (xdg-toplevel or xdg-popup) is
  destroyed, the surface is unmapped and role-specific resources
  are destroyed. The client can call `get_popup` or `get_toplevel`
  again on that surface.
- When the xdg-surface object is destroyed, the surface is
  unmapped, role-specific resources are destroyed and the surface
  itself is destroyed.

TODO:
- [x] zxdg-shell-v6
- [x] xdg-shell stable

Fixes #747

### DIFF
--- a/include/wlr/types/wlr_xdg_shell_v6.h
+++ b/include/wlr/types/wlr_xdg_shell_v6.h
@@ -74,9 +74,22 @@ struct wlr_xdg_toplevel_v6 {
 	struct wlr_xdg_surface_v6 *base;
 	struct wlr_xdg_surface_v6 *parent;
 	bool added;
+
 	struct wlr_xdg_toplevel_v6_state next; // client protocol requests
 	struct wlr_xdg_toplevel_v6_state pending; // user configure requests
 	struct wlr_xdg_toplevel_v6_state current;
+
+	char *title;
+	char *app_id;
+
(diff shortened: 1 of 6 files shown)


include/wlr/types/wlr_xdg_shell_v6.h
```diff
@@ -106,6 +106,7 @@ struct wlr_xdg_surface_v6 {
 	uint32_t configure_next_serial;
 	struct wl_list configure_list;
 
+	// Only for toplevel
```
SirCmpwn: We should probably move it to the toplevel struct then

SirCmpwn:
And also xdg shell stable changes, ofc

acrisci:
:+1: so far so good

emersion:
This PR is ready for review.

PR #754, opened 2018-03-26
[WIP] xdg-shell: damage view when popup is mapped/unmapped

### LINKED ISSUE #752: damage tracking on xdg-popup only partial
As the title says, the damage tracking for xdg-popups seems to only partially cover the popup.

It looks like the damage is clipped to the xdg-toplevel's surface (not geometry)

emersion:
Test plan: spawn an xdg-shell client that has menus with xdg-popups. Resize the window so that it becomes small. Open a menu and then open a submenu so that the submenu si displayed outside the parent toplevel. Close the submenu. There are damage artifacts on master, there shouldn't be any with this PR.

TODO:
- [x] zxdg-shell-v6
- [x] xdg-shell stable

Fixes #752

### DIFF
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -186,6 +186,8 @@ struct roots_xdg_popup {
 	struct roots_view_child view_child;
 	struct wlr_xdg_popup *wlr_popup;
 	struct wl_listener destroy;
+	struct wl_listener map;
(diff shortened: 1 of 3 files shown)


emersion:
This PR is ready for review.

acrisci:
:+1:

PR #776, opened 2018-03-29
WIP: Add partial support for zwp_text_input_unstable_v3

dcz-purism:
Present features: preediting string, sending enter/leave events and handling basic events.

There is a rootston version taking advantage of this change to paste some text into text fields [here](https://code.puri.sm/dorota.czaplejewicz/wlroots/src/text_input_test). GTK version updated to support this input method is [here](https://code.puri.sm/dorota.czaplejewicz/gtk).

### DIFF
--- a/include/wlr/types/wlr_text_input.h
+++ b/include/wlr/types/wlr_text_input.h
@@ -1,0 +1,69 @@
+#ifndef WLR_TYPES_WLR_TEXT_INPUT_H
+#define WLR_TYPES_WLR_TEXT_INPUT_H
+
+#include <wayland-server.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_surface.h>
+
+struct wlr_text_input_properties {
+	struct {
+		char *text; // NULL is allowed and equivalent to empty string
+		uint32_t cursor;
+		uint32_t anchor;
+	} surrounding;
+
+	struct {
+		uint32_t hint;
+		uint32_t purpose;
+	} content_type;
+
+	struct {
+		int32_t x;
+		int32_t y;
+		int32_t width;
+		int32_t height;
+	} cursor_rectangle;
+};
+
+struct wlr_text_input {
+	struct wlr_seat *seat; // used only for unlinking seat->text_input; perhaps better to use the destroy signal
+	struct wl_resource *resource;
+	struct wlr_surface *focused_surface;
+	struct wlr_text_input_properties pending;
+	struct wlr_text_input_properties current;
+	struct wl_listener seat_destroy_listener;
+
+	struct {
+		struct wl_signal enable; // (uint32_t show_input_panel)
+		struct wl_signal commit; // (wlr_text_input*)
+		struct wl_signal disable; // (void*)
+		struct wl_signal destroy; // (wlr_seat* seat)
+	} events;
+};
+
+struct wlr_text_input_manager {
+	struct wl_global *wl_global;
+	struct wl_list text_inputs; // wlr_text_input::resource::link
+
+	struct {
+		struct wl_signal text_input; // (wlr_text_input*)
+	} events;
+};
+
+struct wlr_text_input_manager *wlr_text_input_manager_create(
+	struct wl_display *wl_display);
+void wlr_text_input_manager_destroy(
+	struct wlr_text_input_manager *manager);
+void wlr_text_input_send_enter(struct wlr_text_input *text_input,
+	struct wlr_surface *wlr_surface);
+void wlr_text_input_send_leave(struct wlr_text_input *text_input,
+	struct wlr_surface *wlr_surface);
+void wlr_text_input_send_preedit_string(struct wlr_text_input *text_input,
+	const char *text, uint32_t cursor);
+void wlr_text_input_send_commit_string(struct wlr_text_input *text_input,
+	const char *text);
+void wlr_text_input_send_delete_surrounding_text(
+	struct wlr_text_input *text_input, uint32_t before_length,
+	uint32_t after_length);
+
+#endif
--- a/protocol/text-input-unstable-v3.xml
+++ b/protocol/text-input-unstable-v3.xml
@@ -1,0 +1,308 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<protocol name="text_input_unstable_v3">
+  <copyright>
+    Copyright © 2012, 2013 Intel Corporation
+    Copyright © 2015, 2016 Jan Arne Petersen
+    Copyright © 2017, 2018 Red Hat, Inc.
+    Copyright © 2018 Purism SPC
+
+    Permission to use, copy, modify, distribute, and sell this
+    software and its documentation for any purpose is hereby granted
+    without fee, provided that the above copyright notice appear in
+    all copies and that both that copyright notice and this permission
+    notice appear in supporting documentation, and that the name of
+    the copyright holders not be used in advertising or publicity
+    pertaining to distribution of the software without specific,
+    written prior permission.  The copyright holders make no
+    representations about the suitability of this software for any
+    purpose.  It is provided "as is" without express or implied
+    warranty.
+
+    THE COPYRIGHT HOLDERS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwp_text_input_v3" version="1">
+    <description summary="text input">
+      The zwp_text_input_v3 interface represents text input and input methods
+      associated with a seat. It provides enter/leave events to follow the
+      text input focus for a seat.
+
+      Requests are used to enable/disable the text-input object and set
+      state information like surrounding and selected text or the content type.
+      The information about the entered text is sent to the text-input object
+      via the pre-edit and commit_string events.
+
+      Text is valid UTF-8 encoded, indices and lengths are in bytes. Indices
+      have to always point to the first byte of an UTF-8 encoded code point.
+      Lengths are not allowed to contain just a part of an UTF-8 encoded code
+      point.
+
+      Focus moving throughout surfaces will result in the emission of
+      zwp_text_input_v3.enter and zwp_text_input_v3.leave events. The focused
+      surface must perform zwp_text_input_v3.enable and
+      zwp_text_input_v3.disable requests as the keyboard focus moves across
+      editable and non-editable elements of the UI. Those two requests are not
+      expected to be paired with each other, the compositor must be able to
+      handle consecutive series of the same request.
+
+      State is sent by the state requests (set_surrounding_text,
+      set_content_type and set_cursor_rectangle) and a commit request.
+      After an enter event or disable request all state information is
+      invalidated and needs to be resent by the client.
+
+      This protocol defines requests and events necessary for regular clients
+      to communicate with an input method. The zwp_input_method protocol
+      defines the interfaces necessary to implement standalone input methods.
+      If a compositor implements both interfaces, it will be the arbiter of the
+      communication between both.
+
+      Warning! The protocol described in this file is experimental and
+      backward incompatible changes may be made. Backward compatible changes
+      may be added together with the corresponding interface version bump.
+      Backward incompatible changes are done by bumping the version number in
+      the protocol and interface names and resetting the interface version.
+      Once the protocol is to be declared stable, the 'z' prefix and the
+      version number in the protocol and interface names are removed and the
+      interface version number is reset.
+    </description>
+
+    <request name="destroy" type="destructor">
+      <description summary="Destroy the wp_text_input">
+       Destroy the wp_text_input object. Also disables all surfaces enabled
+       through this wp_text_input object.
+      </description>
+    </request>
+
+    <enum name="enable_flags" bitfield="true">
+      <description summary="enable flags">
+       Enable flags is a bitmask to allow to modify the behavior of the text
+       input.
+      </description>
+      <entry name="none" value="0x0" summary="no special behavior"/>
+      <entry name="can_show_preedit" value="0x1" summary="hints that the UI is capable of showing pre-edit text"/>
+      <entry name="toggle_input_panel" value="0x2" summary="requests toggling input panel (eg. on-screen keyboard)"/>
+    </enum>
+
+    <request name="enable">
+      <description summary="Request text input to be enabled">
+        Requests text input on a surface.
+      </description>
+      <arg name="flags" type="uint" enum="enable_flags" summary="details of the enable request"/>
+    </request>
+
+    <request name="disable">
+      <description summary="Disable text input on a surface">
+        Explicitly disable text input in a surface (typically when there is no
+        focus on any text entry inside the surface).
+      </description>
+    </request>
+
+    <request name="set_surrounding_text">
+      <description summary="sets the surrounding text">
+       Sets the surrounding plain text around the input position. Text is
+       UTF-8 encoded. Cursor is the byte offset within the surrounding text.
+       Anchor is the byte offset of the selection anchor within the
+       surrounding text. If there is no selected text, anchor is the same as
+       cursor.
+
+       Make sure to always send some text before and after the cursor
+       except when the cursor is at the beginning or end of text.
+
+       There is a maximum length of wayland messages so text can not be
+       longer than 4000 bytes.
+       
+       Values set with this request are double-buffered. They will get applied
+       on the next zwp_text_input_v3.commit request.
+
+       The initial value for text is an empty string, and the initial values for 
+       cursor and anchor are 0.
+      </description>
+      <arg name="text" type="string"/>
+      <arg name="cursor" type="int"/>
+      <arg name="anchor" type="int"/>
+    </request>
+
+    <enum name="content_hint" bitfield="true">
+      <description summary="content hint">
+       Content hint is a bitmask to allow to modify the behavior of the text
+       input.
+      </description>
+      <entry name="none" value="0x0" summary="no special behavior"/>
+      <entry name="completion" value="0x1" summary="suggest word completions"/>
+      <entry name="spellcheck" value="0x2" summary="suggest word corrections"/>
+      <entry name="auto_capitalization" value="0x4" summary="switch to uppercase letters at the start of a sentence"/>
+      <entry name="lowercase" value="0x8" summary="prefer lowercase letters"/>
+      <entry name="uppercase" value="0x10" summary="prefer uppercase letters"/>
+      <entry name="titlecase" value="0x20" summary="prefer casing for titles and headings (can be language dependent)"/>
+      <entry name="hidden_text" value="0x40" summary="characters should be hidden"/>
+      <entry name="sensitive_data" value="0x80" summary="typed text should not be stored"/>
+      <entry name="latin" value="0x100" summary="just Latin characters should be entered"/>
+      <entry name="multiline" value="0x200" summary="the text input is multiline"/>
+    </enum>
+
+    <enum name="content_purpose">
+      <description summary="content purpose">
+       The content purpose allows to specify the primary purpose of a text
+       input.
+
+       This allows an input method to show special purpose input panels with
+       extra characters or to disallow some characters.
+      </description>
+      <entry name="normal" value="0" summary="default input, allowing all characters"/>
+      <entry name="alpha" value="1" summary="allow only alphabetic characters"/>
+      <entry name="digits" value="2" summary="allow only digits"/>
+      <entry name="number" value="3" summary="input a number (including decimal separator and sign)"/>
+      <entry name="phone" value="4" summary="input a phone number"/>
+      <entry name="url" value="5" summary="input an URL"/>
+      <entry name="email" value="6" summary="input an email address"/>
+      <entry name="name" value="7" summary="input a name of a person"/>
+      <entry name="password" value="8" summary="input a password (combine with sensitive_data hint)"/>
+      <entry name="pin" value="9" summary="input is a numeric password (combine with sensitive_data hint)"/>
+      <entry name="date" value="10" summary="input a date"/>
+      <entry name="time" value="11" summary="input a time"/>
+      <entry name="datetime" value="12" summary="input a date and time"/>
+      <entry name="terminal" value="13" summary="input for a terminal"/>
+    </enum>
+
+    <request name="set_content_type">
+      <description summary="set content purpose and hint">
+       Sets the content purpose and content hint. While the purpose is the
+       basic purpose of an input field, the hint flags allow to modify some
+       of the behavior.
+       
+       Values set with this request are double-buffered. They will get applied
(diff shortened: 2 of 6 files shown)


include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (C) 2018 Purism SPC
+ *
+ * SPDX-License-Identifier: MIT
+ */
```
SirCmpwn: hm

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef WLR_TYPES_WLR_GTK_TEXT_INPUT_H
+#define WLR_TYPES_WLR_GTK_TEXT_INPUT_H
```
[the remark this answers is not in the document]
SirCmpwn: Also this all should be s/gtk/zwp/g right

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,41 @@
...
+	struct wlr_text_input_manager *manager);
+void wlr_text_input_send_enter(struct wlr_text_input *text_input,
+	struct wlr_surface *wlr_surface);
+void wlr_text_input_send_leave(struct wlr_text_input *text_input,
+	struct wlr_surface *wlr_surface);
+void wlr_text_input_send_preedit_string(struct wlr_text_input *text_input,
+	const char *text, uint32_t cursor);
+void wlr_text_input_send_commit_string(struct wlr_text_input *text_input,
+	const char *text);
+#endif
```
SirCmpwn: Add newline before endif

protocol/text-input-unstable-v3.xml
```diff
@@ -0,0 +1,293 @@
...
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="zwp_text_input_v3" version="1">
+    <description summary="text input">
+      The zwp_text_input_v3 interface represents text input and input methods
```
SirCmpwn: I thought this was on v2?

include/wlr/types/wlr_text_input.h
dcz-purism: Please correct me if the license is wrong - I tried to match the identifier to the LICENSE file.

SirCmpwn:
I like where this is going

dcz-purism:
Changes implemented in this revision:

- removed header comments
- removed `manager::text_input`
- sorted `include`s
- fixed name of include guard to reflect file name
- added newline before `endif`

Not changed yet:

- serial number checking, since its purpose is unclear (waiting for feedback)

SirCmpwn:
>serial number checking, since its purpose is unclear (waiting for feedback)

We can probably just yank this and mention it when we send the protocol around for review.

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,35 @@
+#ifndef WLR_TYPES_WLR_TEXT_INPUT_H
+#define WLR_TYPES_WLR_TEXT_INPUT_H
+
+#include <wayland-server.h>
+#include <wlr/types/wlr_surface.h>
+
+
```
SirCmpwn: Spacing nits: no newline between `#define` and `#include`s, and only one space between `#include`s and declarations

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,35 @@
...
+struct wlr_text_input {
+	struct wl_resource *resource;
+	struct wlr_surface *focused_surface;
+	uint32_t next_serial;
+	struct {
+		struct wl_signal enable_notify;
+		struct wl_signal disable_notify;
```
SirCmpwn: `s/_notify//`

types/wlr_text_input.c
```diff
@@ -0,0 +1,182 @@
...
+#include <stdlib.h>
+#include <stddef.h>
+
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_text_input.h>
+
+#include <wlr/util/log.h>
+
+#include "text-input-unstable-v3-protocol.h"
+#include "util/signal.h"
```
SirCmpwn: More spacing nit: remove all of the empty lines above here and one of the empty lines below here

types/wlr_text_input.c
```diff
@@ -0,0 +1,182 @@
...
+}
+
+void wlr_text_input_send_commit_string(struct wlr_text_input *text_input,
+		const char *text) {
+	zwp_text_input_v3_send_commit_string(text_input->resource, text);
+}
+
+static void text_input_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	wlr_log(L_INFO, "STUB: text_input_destroy");
```
SirCmpwn: Can you implement this

types/wlr_text_input.c
```diff
@@ -0,0 +1,198 @@
...
+}
+
+static void text_input_enable(struct wl_client *client,
+		struct wl_resource *resource, uint32_t serial,
+		uint32_t show_input_panel) {
+	struct wlr_text_input *text_input = text_input_from_resource(resource);
+	if (serial >= text_input->next_serial) {
+		wlr_log(L_INFO,
+			"text_input_enable called with a serial that wasn't sent yet: %d",
+			serial);
```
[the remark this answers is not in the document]
SirCmpwn: bump

types/wlr_text_input.c
```diff
@@ -0,0 +1,182 @@
...
+static void text_input_set_cursor_rectangle(struct wl_client *client,
+		struct wl_resource *resource, int32_t x, int32_t y, int32_t width,
+		int32_t height) {
+	wlr_log(L_INFO, "STUB: set_cursor_rectangle(%d, %d, %d, %d)", x, y, width,
+		height);
+}
+
+static void text_input_commit(struct wl_client *client,
+		struct wl_resource *resource) {
+	wlr_log(L_INFO, "STUB: text_input_commit()");
```
SirCmpwn: Please implement

types/wlr_text_input.c
```diff
@@ -0,0 +1,182 @@
...
+static void text_input_set_content_type(struct wl_client *client,
+		struct wl_resource *resource, uint32_t hint, uint32_t purpose) {
+	wlr_log(L_INFO, "STUB: set_content_type(%d, %d)", hint, purpose);
+}
+
+static void text_input_set_cursor_rectangle(struct wl_client *client,
+		struct wl_resource *resource, int32_t x, int32_t y, int32_t width,
+		int32_t height) {
+	wlr_log(L_INFO, "STUB: set_cursor_rectangle(%d, %d, %d, %d)", x, y, width,
+		height);
```
SirCmpwn: This stuff too, you can just stash it and not use it on rootston but definitely implement the state transitions

types/wlr_text_input.c
```diff
@@ -0,0 +1,182 @@
...
+	.set_content_type = text_input_set_content_type,
+	.set_cursor_rectangle = text_input_set_cursor_rectangle,
+	.commit = text_input_commit,
+};
+
+static const struct zwp_text_input_manager_v3_interface text_input_manager_impl;
+
+static void text_input_manager_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	wlr_log(L_INFO, "STUB: text_input_manager_destroy");
```
SirCmpwn: Implement this dude too

Basically this file shouldn't have stubs, even if the rootston implementation doesn't take advantage of everything we can still provide a complete text_input state machine implementation reasonably easily here

dcz-purism:
Changes:
- yanked serial and replaced it with `0` in calls
- implemented stubs by exposing them as signals
- whitespace fixes
- removed `_notify` from signal names

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,57 @@
...
+	struct {
+		int32_t x;
+		int32_t y;
+		int32_t width;
+		int32_t height;
+	} cursor_rectangle;
+};
+
+struct wlr_text_input {
+	struct wlr_seat *seat; // used only for unlinking seat->text_input; perhaps better to use the destroy signal
```
SirCmpwn: This sort of comment is better put into the commit message or here on GitHub

SirCmpwn:
>yanked serial and replaced it with 0 in calls

I was expecting you to remove it from the protocol

>implemented stubs by exposing them as signals

Hmm, this doesn't look like I expected. It looks like you're throwing away the state when it's committed? That doesn't make sense. You should check out the commit flow that the shells use, the purpose is to double buffer the state. Should also update the xml descriptions to use the usual language around double buffering that we see in other protocols.

emersion:
>I was expecting you to remove it from the protocol

Ugh. What about posting in wayland-devel and keeping it while waiting for a protocol v2?

include/wlr/types/wlr_text_input.h
dcz-purism: Sorry, forgot to move it before committing. What do you think about the idea in the comment anyway?

SirCmpwn:
No, this is part of the process. No one is going to push this protocol forward for us. Take the initiative to sponsor v3 by cleaning it up, iterating it through us and other stakeholders (KDE, GTK) for feedback, then posting your version to wayland-devel for consideration. As part of that process it's useful to have an example implementation of the protocol to show off - that's where this PR comes in.

dcz-purism:
@emersion: my email to wayland-devel is stuck in moderation...

include/wlr/types/wlr_text_input.h
SirCmpwn: Probably better to use the destroy signal, yeah.

emersion: I don't think so. That would require adding text_input code in the seat.

SirCmpwn: Ah, good point. This patch already touches the seat, should probably be cleaned out.

dcz-purism:
About throwing away state: initially I implemented it with double-buffering, before realizing that it doesn't make sense to keep `set_surrounding_text` state between calls. No one wants to paste the same text after the content type has changed. I'm under the impression that the `commit` is only meant to convey what was changed from the most recent `commit` call. Perhaps `set_content_type` and `set_cursor_rectangle` might be kept between calls, but I also don't necessarily see them as state.

SirCmpwn:
>Perhaps set_content_type and set_cursor_rectangle might be kept between calls, but I also don't necessarily see them as state.

I mean, they are

include/wlr/types/wlr_text_input.h
dcz-purism: Seat code is touched because there needs to be some way to get notified of new `text_input` instances. Since only one per seat is allowed, I thought seat is [the best place](https://github.com/swaywm/wlroots/pull/776/files#diff-6eb821df1a7fff48e265e52e0a742741R215). Should it be placed somewhere else instead?

SirCmpwn: Probably better to construct a text_input with a seat reference and have it manage it. Or let the compositor glue stuff together.

dcz-purism:
Changes:
- removed new code from `wlr_seat`
- removed serial from protocol
- implemented double buffering of state
- abandoned clearing of state between commits; this means client needs to keep track of what's in `set_surrounding_text`
- moved create signal to manager

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,67 @@
...
+#include <wayland-server.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_surface.h>
+
+struct wlr_text_input_atomic_properties {
+	struct {
+		char *text;
+		uint32_t cursor;
+		uint32_t anchor;
+	} surrounding;
```
SirCmpwn: Nitpick, add a newline after this

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,67 @@
+#ifndef WLR_TYPES_WLR_TEXT_INPUT_H
+#define WLR_TYPES_WLR_TEXT_INPUT_H
+
+#include <wayland-server.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_surface.h>
+
+struct wlr_text_input_atomic_properties {
```
SirCmpwn: Nitpick, just call this wlr_text_input_properties

include/wlr/types/wlr_text_input.h
```diff
@@ -0,0 +1,67 @@
...
+		int32_t height;
+	} cursor_rectangle;
+};
+
+struct wlr_text_input {
+	struct wlr_seat *seat; // used only for unlinking seat->text_input; perhaps better to use the destroy signal
+	struct wl_resource *resource;
+	struct wlr_surface *focused_surface;
+	struct wlr_text_input_atomic_properties pending_properties;
+	struct wlr_text_input_atomic_properties properties;
```
SirCmpwn: We use `pending` and `current` as field names for this sort of thing.

SirCmpwn:
This is getting pretty close, I like it.

1. The XML descriptions should use the conventional language for describing double buffered state (see e.g. wl_surface, xdg_surface, etc)
2. Is there a test client we can use?

SirCmpwn:
>abandoned clearing of state between commits; this means client needs to keep track of what's in set_surrounding_text

Can you elaborate on this?

dcz-purism:
Updated noticed nitpicks.

dcz-purism:
1. I'll try to do that.
2. There is a test client at https://code.puri.sm/dorota.czaplejewicz/wlroots/src/text_input_test , although it needs to be updated after seat code was cleaned (the newest update was not tested outside of compilation). It tests only the basic preedit/commit_string functionality. See my original comment on this merge request.

I'm not 100% sure how `set_surrounding_text` is meant to work, but assuming the client almost never wants to set the same non-null text twice, it can no longer assume that the text will be empty after each commit.

SirCmpwn:
Well the text wouldn't be empty after commit, would it be? Why would it be?

SirCmpwn:
>needs to be updated

Please do that and give us a test plan so we can make sure this works

dcz-purism:
What form would you like the test plan to take?

I'm not sure what other implementations will do when a commit is called; if they come to the conclusion (as I did) that surrounding is invalid between contexts, and thus commit calls, they could clear it.

Either way, it's not cleared now.

SirCmpwn:
>What form would you like the test plan to take?

Just a list of steps we can perform to exercise this code and verify it works as promised

>I'm not sure what other implementations will do when a commit is called; if they come to the conclusion (as I did) that surrounding text is not valid across contexts, and thus commit calls, they could clear it.

If you cleared it out after commit, then when is the server ever going to be able to display it?

dcz-purism:
The scenario in my mind is this:

- text field A gets focus
- client sends content_purpose = email
- client sends set_surrounding_text = "@foobar.com"
- client commits
- field B gets focused
- client uses the same text_input to set content_purpose = normal
- client commits

If commit doesn't clean up set_surrounding_text, then the client must ensure it's correct for the given "context". In this case, field B would get "@foobar.com" as the prompt.

If the compositor is made under the assumption that set_surrounding_text is ephemeral and clear by default, clients will not get accidental leftovers when switching fields.

Perhaps this is underspecified; or perhaps this should be prevented by replacing the separate calls with a single compound one to reduce the potential for such bugs. I don't know how much benefit there is from keeping those calls separate though.

SirCmpwn:
I see the sense in your argument, but it works a bit differently from how most wayland protocols are assumed to work. I'm okay with your approach if you document it clearly in the XML.

dcz-purism:
I can think of a couple of approaches. I would try to solve it first by changing the protocol in such a way that making an error is not possible first. In order of preference:
- turn 3 calls and commit into a single call: the code generator/compiler enforces compliance
- clear state on each focus change (`enable` request or `enter` event): this is closest to my intuitive understanding of when state changes, aligning behaviour with expectation
- clear state on each commit: this will force clients to re-request all state each time even within a single field, in the likely event of forgetting this will result in bugs (resetting field to default)
- do not clear state on each commit: similar as above, any bugs will be more confusing because they will come from old values
- only reset surrounding text: splitting state into permanent and volatile is even more surprising IMO

If there's a way to enforce compliance automatically and keep separate calls for each property, I'm all ears.

SirCmpwn:
I don't want to merge requests, I think that the several requests plus double buffering approach works well (and is better for future extensibility, we can add more double buffered requests later more easily).

>clear state on each commit: this will force clients to re-request all state each time even within a single field, in the likely event of forgetting this will result in bugs (resetting field to default)

I like this. We can make not setting the text before commit a protocol error.

dcz-purism:
> We can make not setting the text before commit a protocol error.
Singling out just the text property for special treatment is a bug invitation IMO, and future implementations would break compatibility if they tried to introduce another such property.
If state should change on commit and not focus switch, I'd make it reset to a set of predefined safe defaults.

SirCmpwn:
>Singling out just the text property for special treatment is a bug invitation

I don't think this would lead to bugs

>future implementations would break compatibility if they tried to introduce another such property

This would naturally necessitate a version bump and the compositor can handle it accordingly, wayland is designed to handle such cases.

dcz-purism:
What criterion would you propose to separate properties resetting on commit and those which should stay the same?

dcz-purism:
Updated:
- double-buffering language in protocol description
- described initial values
- minor bugs

SirCmpwn:
>What criterion would you propose to separate properties resetting on commit and those which should stay the same?

Discussion on GitHub/MLs?

protocol/text-input-unstable-v3.xml
```diff
@@ -0,0 +1,314 @@
...
+      <description summary="set content purpose and hint">
+       Sets the content purpose and content hint. While the purpose is the
+       basic purpose of an input field, the hint flags allow to modify some
+       of the behavior.
+       
+       Content type is double-buffered state, see zwp_text_input_v3.commit.
+
+       zwp_text_input_v3.commit copies the pending content type to the
+       current content type. Otherwise the pending and current content types
+       never change.
```
SirCmpwn: This language still doesn't quite match. Try:

>Values set in this way are double-buffered. They will get applied on the next commit.

Also, don't duplicate docs on what commit does in every request.

dcz-purism: Updated - I'm not sure if the update is what you're looking for; the text was copied almost verbose from wayland.xml.

SirCmpwn: Ah, I was looking at xdg-shell. I guess the "standard langauge" is a bit nebulous. xdg-shell's language is a big more standard now because I copied it for layer-shell and now you've copied it for text input :tada:

SirCmpwn:
Can you send the current iteration of the protocol around to the main stakeholders for comments?

Also still looking for a way to test this, but there's no rush, take your time.

dcz-purism:
Current iteration of the protocol (can't be attached here)

https://paste.debian.net/1018117/

dcz-purism:
WIP testing procedure: use `text_input_test` branch of rootston from [here](https://code.puri.sm/dorota.czaplejewicz/wlroots/src/text_input_test).

Only 1 client at a time supported. Works with gtk 3.22's wayland IM module from https://code.puri.sm/dorota.czaplejewicz/gtk .

To test with older GTK installed, build the altered version and make use only of the loadable input method module with a program of your choice (`yad` here):

```
$ chrpath -d .../gtk-install/lib/gtk-3.0/3.0.0/immodules/im-wayland.so
$ rootston
$ WAYLAND_DEBUG=1 GDK_BACKEND=wayland GTK_IM_MODULE=wayland GTK_IM_MODULE_FILE=.../gtk-install/lib/gtk-3.0/3.0.0/immodules.cache yad --entry
```

The procedure is unlikely to change with subsequent updates.

SirCmpwn:
I was able to get this sort of... working? I'm not really sure. What is the expected behavior? Can you elaborate on your test plan?

dcz-purism:
Sorry - I should have added a bit.

The demo works by sending the `enter`, `leave`, `preedit_string` and `commit_string` requests. It will handle the creation of a new text input. The `enter` and `leave` are sent automatically, the `preedit_string` and `commit_string` requests are sent when a field is in focus and the mouse is left-clicked. The first 2 events are `preedit_string`, and the last one is `commit_string`, all with predefined strings "str2", "string1", "textfinal". After `commit_string` is submitted, the instance of `text_input` will not send any more string events.

Events attached to the `text_input` instance are not handled. The request `delete_surrounding_text` is not sent.

I'm not sure how the GTK side should be interpreting this - as far as the protocol goes, the messages are being exchanged correctly, but `commit_string` doesn't seem to be clearing the previously received preedit strings, which may be a bug with original GTK implementation.

SirCmpwn:
I'm still not entirely clear on what buttons I need to press in which order and what I should see when I know it's working.

dcz-purism:
1. Run a GTK program,
2. Get a text field focused
3. Left-click anywhere, observe text suggestion
4. Left-click anywhere, observe different text suggestion
5. Left-click anywhere, observe committed text
6. Back to normal

SirCmpwn:
Okay, I was able to confirm that works.

SirCmpwn:
I think this is only a small piece of a rather broad pie, though. I'm not sure what the future direction is and how we should incoroproate all of it (and in what degrees at what times) into wlroots.

dcz-purism:
The protocol support is rather self-contained.
There needs to be something to handle it on the compositor side, of course. There are two basic possibilities are to make an input method part of the compositor or not.

In the former case, this protocol already gives enough support on the wlroots side to make this a reality. For interfacing with a separate input method, there's input-method-unstable-v1, which can also make its way into wlroots. That was in fact my next planned step after this.

To summarize, I think the text-input protocol on its own is already useful, and it's going to be even more useful together with input-method while not interfering with anything else in wlroots if compositor authors decide not to make use of them.

SirCmpwn:
Alright, thanks for clarifying.

Shugyousha:
This WIP looks very interesting! The protocol suggestion (with links to the implementation) should be sent to the wayland mailing list soon.

Currently there was a different version 3 of the text-input protocol posted there: https://patchwork.freedesktop.org/patch/209838/ some of the comments garnered there could be applied here as well.

dcz-purism:
I've updated the XML file with fixes indicated in the wayland-devel thread.

SirCmpwn:
Can you also reply to the wayland-devel thread updating them about the work we've been doing over here? Be sure to reply-all so the Gnome guys gets it in his inbox.

dcz-purism:
Sure. In the meantime, I debugged multi-client support in the demo and pushed it [here](https://code.puri.sm/dorota.czaplejewicz/wlroots/src/text_input_test). Hopefully it doesn't crash too much.

SirCmpwn:
I would love to see your Rootston changes added to this patch, and perhaps a test client in examples/ if you have time.

dcz-purism:
Are you sure about Rootston changes? The "click to have arbitrary text appear" behaviour is going to make rootston rather useless for clients supporting this protocol.
I could see "noop" support added to rootston: all the structs and signals in the right places.

Anyway, I'm beginning work on `zwp_input_method`, which I think should give enough basis for rootston support without reimplementing a virtual keyboard inside rootston.

SirCmpwn:
Oh, alright then.

dcz-purism:
Some highlights from the IRC discussion about text input.

The current protocol does not support highlighting a part of preedit text in order to indicate that it's focus (important for Japanese, Korean, Chinese input methods). Ibus protocol solves this by passing around styling information between the input method and the client. Possible appraoaches:

1. Communicate information about what is active and leave styling to application. Possibly also about what is wrong, or what's in rubies etc. Con: limited possibilities.
2. Communicate styling directly via some form of rich text. Con: style like colors will probably clash with the application theme.
3. Leave styling to input method using an additional surface. Con: can never match up styling with the application.

Option 3 is not mutually exclusive with the others.

The consensus seemed to be to choose a version of option 1, possibly limiting it to the bare minimum: 1 possible currently active range. This could be related to the cursor position which is already present in `set_preedit_string`.

PR #779, opened 2018-03-30
xwayland: rename map_notify to map for consistency

emersion:
`notify` is never used in event names, and `map` is used in other shells.

Will send a Sway PR when this one gets approved.


acrisci:
:+1:

PR #780, opened 2018-03-30
output-damage: fix missing wlr_box include

emersion:


SirCmpwn:
Feel free to merge this when build is successful

PR #783, opened 2018-03-30
Add pointer input to layer surfaces

SirCmpwn:
This glues things together in rootston and extends the layer shell example to visually demonstrate pointer events.

Test plan:

1. Run rootston
2. Run layer shell example with various configurations
3. Test that pointer appears in correct place and clicking changes the color of the surface
4. Test that surface gets a client side cursor
5. Bring a shell surface to the party and make sure nothing breaks
6. Test z-ordering works correctly with several layer surfaces

### DIFF
--- a/examples/layer-shell.c
+++ b/examples/layer-shell.c
@@ -7,12 +7,17 @@
 #include <time.h>
 #include <unistd.h>
 #include <wayland-client.h>
+#include <wayland-cursor.h>
 #include <wayland-egl.h>
 #include <wlr/render/egl.h>
 #include <wlr/util/log.h>
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 
 static struct wl_compositor *compositor = NULL;
+static struct wl_seat *seat = NULL;
+static struct wl_shm *shm = NULL;
+static struct wl_pointer *pointer = NULL;
+//static struct wl_keyboard *keyboard = NULL;
 static struct zwlr_layer_shell_v1 *layer_shell = NULL;
 struct zwlr_layer_surface_v1 *layer_surface;
 static struct wl_output *wl_output = NULL;
@@ -32,6 +37,12 @@ static double alpha = 1.0;
 static bool run_display = true;
 static bool animate = false;
 static double frame = 0;
+static int cur_x = -1, cur_y = -1;
+static int buttons = 0;
+
+struct wl_cursor_theme *cursor_theme;
+struct wl_cursor_image *cursor_image;
+struct wl_surface *cursor_surface;
 
 static struct {
 	struct timespec last_frame;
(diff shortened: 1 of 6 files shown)


SirCmpwn:
This is ready for review

examples/layer-shell.c
```diff
@@ -119,18 +144,113 @@ struct zwlr_layer_surface_v1_listener layer_surface_listener = {
...
+static void wl_pointer_leave(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface) {
+	cur_x = cur_y = -1;
+	buttons = 0;
+}
+
+static void wl_pointer_motion(void *data, struct wl_pointer *wl_pointer,
+		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
+	cur_x = (int)wl_fixed_to_double(surface_x);
+	cur_y = (int)wl_fixed_to_double(surface_y);
```
acrisci: `wl_fixed_to_int()` is a thing

examples/layer-shell.c
```diff
@@ -119,18 +144,113 @@ struct zwlr_layer_surface_v1_listener layer_surface_listener = {
...
 		if (output == 0 && !wl_output) {
 			wl_output = wl_registry_bind(registry, name,
 					&wl_output_interface, 1);
 		} else {
 			output--;
 		}
+	} else if (strcmp(interface, wl_seat_interface.name) == 0) {
+		seat = wl_registry_bind(registry, name,
+				&wl_seat_interface, 1);
+		wl_seat_add_listener(seat, &seat_listener, NULL);
```
acrisci: multiseat?

SirCmpwn: Should Just Werk(TM) ish but I think it's outside of the scope of the layer shell example client to deal with

examples/layer-shell.c
```diff
@@ -119,18 +144,113 @@ struct zwlr_layer_surface_v1_listener layer_surface_listener = {
 	.closed = layer_surface_closed,
 };
 
+static void wl_pointer_enter(void *data, struct wl_pointer *wl_pointer,
+		uint32_t serial, struct wl_surface *surface,
+		wl_fixed_t surface_x, wl_fixed_t surface_y) {
+	wl_surface_attach(cursor_surface,
+			wl_cursor_image_get_buffer(cursor_image), 0, 0);
```
acrisci: do you need `wl_surface_damage()` here?

SirCmpwn: Not by my tests, since the surface never changes it's not important

rootston/cursor.c
```diff
@@ -98,51 +99,56 @@ static void seat_view_deco_button(struct roots_seat_view *view, double sx,
 	}
 }
 
-static void roots_cursor_update_position(struct roots_cursor *cursor,
+static void roots_passthrough_cursor(struct roots_cursor *cursor,
 		uint32_t time) {
-	struct roots_desktop *desktop = cursor->seat->input->server->desktop;
-	struct roots_seat *seat = cursor->seat;
 	struct roots_view *view;
```
emersion:
```c
struct roots_view *view = NULL;
```

SirCmpwn: nack, it's set by desktop_surface_at

emersion: It's not obvious that it's set to NULL if there's no view.

rootston/cursor.c
```diff
@@ -98,51 +99,56 @@ static void seat_view_deco_button(struct roots_seat_view *view, double sx,
...
+			cursor->pointer_view = seat_view;
+			seat_view_deco_motion(seat_view, sx, sy);
 		}
+	}
 
-		struct roots_drag_icon *drag_icon;
-		wl_list_for_each(drag_icon, &seat->drag_icons, link) {
-			roots_drag_icon_update_position(drag_icon);
-		}
+	if (surface) {
```
emersion: Should set the default cursor if `cursor->cursor_client != wl_resource_get_client(surface->resource)`

PR #786, opened 2018-03-31
backend/x11: add support for multiple outputs

### LINKED ISSUE #768: X11 backend does not support multiple outputs
Would be neato if it did

emersion:
I'm not satisfied with the X11 event handling, but it's not like there's a better way to do this.

Test plan:
* Config: `[output:X11-1] rotate = 90 scale=1.5`
* `env WLR_X11_OUTPUTS=2 build/rootston/rootston`
* Open a window, check that input events are correct

Fixes #768

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -61,7 +61,7 @@ struct wlr_renderer *wlr_backend_get_renderer(struct wlr_backend *backend) {
 	return NULL;
 }
 
-static int parse_outputs_env(const char *name) {
+static size_t parse_outputs_env(const char *name) {
 	const char *outputs_str = getenv(name);
 	if (outputs_str == NULL) {
 		return 1;
@@ -83,8 +83,8 @@ static struct wlr_backend *attempt_wl_backend(struct wl_display *display) {
 		return NULL;
 	}
 
-	int outputs = parse_outputs_env("WLR_WL_OUTPUTS");
-	for (int i = 0; i < outputs; ++i) {
+	size_t outputs = parse_outputs_env("WLR_WL_OUTPUTS");
+	for (size_t i = 0; i < outputs; ++i) {
 		wlr_wl_output_create(backend);
 	}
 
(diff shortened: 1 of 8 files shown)


emersion:
This PR is ready for review.

SirCmpwn:
This opens n+1 outputs!

martinetd:
Also the layout is a bit weird, this opens X11-2 to the left of X11-1 (wayland opens them on the right) and if you use different scale you get the same area displayed multiple time with the default layouting, might be due to outputs being right-to-left as well.

backend/backend.c
```diff
@@ -61,26 +61,48 @@ struct wlr_renderer *wlr_backend_get_renderer(struct wlr_backend *backend) {
...
+
+	char *end;
+	int outputs = (int)strtol(outputs_str, &end, 10);
+	if (*end || outputs < 0) {
+		wlr_log(L_ERROR, "%s specified with invalid integer, ignoring", name);
+		return 1;
+	}
+
+	return outputs;
+}
```
SirCmpwn: -1

SirCmpwn: I retract my -1 as I notice that this is a static function

backend/backend.c
```diff
@@ -61,26 +61,48 @@ struct wlr_renderer *wlr_backend_get_renderer(struct wlr_backend *backend) {
...
-		}
-		while (outputs--) {
-			wlr_wl_output_create(backend);
-		}
+	if (backend == NULL) {
+		return NULL;
+	}
+
+	int outputs = parse_outputs_env("WLR_WL_OUTPUTS");
+	for (int i = 0; i < outputs; ++i) {
```
SirCmpwn: size_t

backend/wayland/wl_seat.c
```diff
@@ -60,37 +60,28 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
...
 	double oy = wlr_output->ly / (double)layout_box.height;
 
-	wlr_event.x = tx * ((double)owidth / layout_box.width) + ox;
-	wlr_event.y = ty * ((double)oheight / layout_box.height) + oy;
+	struct wlr_event_pointer_motion_absolute wlr_event = {
+		.device = dev,
+		.time_msec = time,
+		.x = box.x / (double)layout_box.width + ox,
+		.y = box.y / (double)layout_box.height + oy,
+	};
```
SirCmpwn: Thank you for fixing my disgusting code

PR #787, opened 2018-03-31
Add wlr_surface_point_accepts_input

SirCmpwn:
Ref https://github.com/swaywm/sway/pull/1674


emersion:
Can you move `wlr_box_contains_point(&box, view_sx, view_sy)` inside?

SirCmpwn:
>Can you move wlr_box_contains_point(&box, view_sx, view_sy) inside?

No, that's in output coords

emersion:
Need to use this function in `layer_surface_at` too.

>No, that's in output coords

I mean, the only thing this function does is checking if the output-local surface box contains the output-local point. You can just add a `sx => 0 && sx <= surface->current->width` (ditto for `sy`) in the new function and remove these calls to `wlr_box_contains_point` (and the code building that box).

SirCmpwn:
Good call.

PR #791, opened 2018-03-31
fix debug log color

acrisci:
According to [this](https://misc.flogisoft.com/bash/tip_colors_and_formatting), `\x1B[1;30m` is a black color. My vim uses this color for the background. Change it to `\x1B[1;37m` which is light gray according to the thing.


emersion:
Wikipedia ref: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors

emersion:
Not sure it's a good idea. 37 is "White" which will be unreadable on terminals with a white background.

SirCmpwn:
How about we just make people configure their terminal color schemes correctly? Your black color should not be the same as your background color.

Issue #527, opened 2017-12-27
#524 follow-up


SirCmpwn:
Opening a ticket so it doesn't get forgotten.

acrisci:
In terms of the client api, this is what we want and the change accomplishes those goals: a single commit event for the surface where the role commit is guaranteed to have run first.

This approach mirrors Weston's, which is my usual starting point.

The alternative to this is to have two separate commit events for the surface: `surface::commit` and `surface::role_commit` where `role_commit` emits first.

SirCmpwn:
I think my major beef with this is the roll-your-own callback logic rather than wl_signal. I can live with the rest of the changes.

emersion:
The intent of the callback is to make sure there's only one listener (because a surface has only one role). This doesn't change much about type safety. I guess a wl_signal would be OK too.

SirCmpwn:
Maybe it would be better to build out a `wlr_surface_role` abstraction that uses the same interface pattern we've done for i.e. wlr_output, wlr_pointer, etc.

emersion:
An issue with this proposal is that role types (`wlr_output_cursor`, `wlr_wl_shell_surface`...) don't wrap a `wlr_surface`, so you can't get back the role type struct from the generic `wlr_surface`.

SirCmpwn:
I don't understand how that follows.

emersion:
Let's say you want to use a role interface and implement it for `wlr_wl_shell_surface`. The interface will look like:

```c
struct wlr_surface_role {
    const char *name;
    void (*commit)(struct wlr_surface *surface);
};
```

In `wlr_wl_shell.c` you'll write something like:

```c
static void wl_shell_surface_commit(struct wlr_surface *surface) {
    // Here I have a `wlr_surface`, but I want a `wlr_wl_shell_surface`
}

static const struct wlr_surface_role wl_shell_surface_role = {
    .name = "wl-shell-surface",
    .commit = wl_shell_surface_commit,
};
```

You'll need the `wlr_wl_shell_surface` in `wl_shell_surface_commit`.

SirCmpwn:
We already encounter this in other places.

```c
struct wlr_wl_shell_surface {
    struct wlr_surface wlr_surface;
    // ...wl_shell state...
};

struct wlr_surface *wlr_surface = (struct wlr_surface*)wlr_wl_shell_surface;
[...]

PR #804, opened 2018-04-02
Add keyboard input to layer surfaces

SirCmpwn:
Per the spec, the topmost layer surface above the shell surface layer always has keyboard focus. Beneath the shell surface layer, layer surfaces will receive keyboard focus normally (when clicked). Note that layer surfaces beneath the shell surface layer receive a sort of temporary focus from the seat - they do not participate in the usual focus list for alt-tabbing and such. Layer surfaces that do not want keyboard input continue to not receive keyboard enter events or take focus away from other surfaces when clicked.

To request keyboard events, pass `-k` to the example layer shell and look for key events in the log.

### DIFF
--- a/include/wlr/types/wlr_wl_shell.h
+++ b/include/wlr/types/wlr_wl_shell.h
@@ -150,4 +150,9 @@ struct wlr_wl_shell_surface *wlr_wl_shell_surface_popup_at(
 		struct wlr_wl_shell_surface *surface, double sx, double sy,
 		double *popup_sx, double *popup_sy);
 
+bool wlr_surface_is_wl_shell_surface(struct wlr_surface *surface);
+
+struct wlr_wl_surface *wlr_wl_shell_surface_from_wlr_surface(
+		struct wlr_surface *surface);
+
 #endif
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -271,6 +271,12 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 			break;
 		case WLR_BUTTON_PRESSED:
 			roots_seat_set_focus(seat, view);
+			if (surface && !view) {
+				struct wlr_layer_surface *layer = surface->role_data;
+				if (layer->current.keyboard_interactive) {
+					roots_seat_set_focus_layer(seat, layer);
+				}
+			}
 			break;
 		}
 	}
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -271,8 +271,9 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 			break;
 		case WLR_BUTTON_PRESSED:
 			roots_seat_set_focus(seat, view);
-			if (surface && !view) {
-				struct wlr_layer_surface *layer = surface->role_data;
+			if (wlr_surface_is_layer_surface(surface)) {
+				struct wlr_layer_surface *layer =
+					wlr_layer_surface_from_wlr_surface(surface);
 				if (layer->current.keyboard_interactive) {
 					roots_seat_set_focus_layer(seat, layer);
 				}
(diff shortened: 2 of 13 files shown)


rootston/cursor.c
```diff
@@ -271,6 +271,12 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 			break;
 		case WLR_BUTTON_PRESSED:
 			roots_seat_set_focus(seat, view);
+			if (surface && !view) {
+				struct wlr_layer_surface *layer = surface->role_data;
```
emersion: This is very meh™. We want to get rid of `role_data`, and it's an internal field.

SirCmpwn: We need some means of getting a wlr_layer_surface/wlr_xdg_surface/etc from a wlr_surface, be it this or some other way

rootston/layer_shell.c
```diff
@@ -192,6 +192,35 @@ void arrange_layers(struct roots_output *output) {
...
 			&usable_area, false);
+
+	// Find topmost keyboard interactive layer, if such a layer exists
+	uint32_t layers_above_shell[] = {
+		ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY,
+		ZWLR_LAYER_SHELL_V1_LAYER_TOP,
+	};
+	struct roots_layer_surface *layer, *topmost = NULL;
+	for (size_t i = 0;
+			i < sizeof(layers_above_shell) / sizeof(layers_above_shell[0]);
```
emersion: Create a temporary variable for the size?

rootston/cursor.c
emersion: Just iterate over the layer surfaces list.

SirCmpwn: That is Very Meh™.

emersion: https://github.com/swaywm/wlroots/issues/527

Find another way. Set `wlr_surface.data` for instance.

SirCmpwn: Meh.

rootston/seat.c
```diff
@@ -800,6 +804,37 @@ void roots_seat_set_focus(struct roots_seat *seat, struct roots_view *view) {
...
+ * You also cannot alt-tab between layer surfaces and shell surfaces.
+ */
+void roots_seat_set_focus_layer(struct roots_seat *seat,
+		struct wlr_layer_surface *layer) {
+	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat->seat);
+	if (!layer) {
+		seat->focused_layer = NULL;
+		return;
+	}
+	if (seat->has_focus) {
```
emersion: What happens if there's a fullscreen view? What happens if the view is focused by another client?

Calling `roots_seat_set_focus(seat, NULL)` could work better.

SirCmpwn: So what if there's a fullscreen view? You can't click layers beneath the shell layer if there is. Layers above the shell layer behave as expected.

rootston/cursor.c
SirCmpwn: We need two functions:

`wlr_layer_surface_from_wlr_surface`

`wlr_surface_is_layer_surface`

Same for xwayland, xdg-shell, etc. This is out of scope for this PR, come back and fix this later.

include/wlr/types/wlr_wl_shell.h
```diff
@@ -150,4 +150,9 @@ struct wlr_wl_shell_surface *wlr_wl_shell_surface_popup_at(
 		struct wlr_wl_shell_surface *surface, double sx, double sy,
 		double *popup_sx, double *popup_sy);
 
+bool wlr_surface_is_wl_surface(struct wlr_surface *surface);
```
emersion: wlr_surface_is_wl_shell_surface

SirCmpwn: derp

PR #805, opened 2018-04-02
Change meson option names to use hypen

DanySpin97:
Replace underscore with hypen in meson option names, following style
recommendations.

http://mesonbuild.com/Style-guide.html#naming-options


SirCmpwn:
Thanks!

PR #806, opened 2018-04-03
Install include directory into includedir

DanySpin97:
Follow includedir parameter when installing include directory.


emersion:
Thanks!

PR #818, opened 2018-04-04
egl: silence dmabuf error when extension is not present

agx:
This makes it match 4bf936360d42fb5b96a44fd17028ae66fc462362.


emersion:
Thanks!

PR #825, opened 2018-04-05
Add surface iterators

### LINKED ISSUE #822: Move surface_for_each_surface into wlr_surface
We already have helpers for input (`wlr_surface_subsurface_at` plus shell-specific functions), but nothing for rendering.

(Note to myself: `wlr_surface_subsurface_at` is broken, plsu we might want to refactor it into `wlr_surface_surface_at`).

emersion:
This PR adds surface iterators to `wlr_surface` and all shells.

Test plan: check that rendering of surfaces, subsurfaces and popups works.

Fixes #822

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -93,8 +93,12 @@ struct wlr_surface {
 
 	// wlr_subsurface::parent_pending_link
 	struct wl_list subsurface_pending_list;
+
 	void *data;
 };
+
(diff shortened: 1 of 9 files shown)


SirCmpwn:
Would be nice to see a corresponding sway patch

emersion:
I'm not sure if the last commit is a good idea. It removes the rootston-specific `surface_iterator_func_t`. Let me know what you think.

>Would be nice to see a corresponding sway patch

This only adds APIs, and Sway doesn't have surface iterators (yet). I'll implement damage tracking tomorrow and start using wlroots iterators directly.

PR #830, opened 2018-04-05
Add wlr_xdg_output_manager

### LINKED ISSUE #826: Implement xdg-output

SirCmpwn:
Fixes #826

### DIFF
--- a/include/wlr/types/wlr_xdg_output.h
+++ b/include/wlr/types/wlr_xdg_output.h
@@ -1,0 +1,32 @@
+#ifndef WLR_TYPES_WLR_XDG_OUTPUT_H
+#define WLR_TYPES_WLR_XDG_OUTPUT_H
+#include <wayland-server.h>
+#include <wlr/types/wlr_output_layout.h>
+
+struct wlr_xdg_output {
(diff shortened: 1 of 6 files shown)


include/wlr/types/wlr_xdg_output_manager.h
```diff
@@ -0,0 +1,32 @@
+#ifndef WLR_TYPES_WLR_XDG_OUTPUT_MANAGER_H
```
emersion: Since the protocol is named `xdg-output`, I think it would make sense to name this `wlr_xdg_output.c`.

SirCmpwn: I thought about that but I supposed that it makes more sense to name it after the primary exported interface

include/wlr/types/wlr_xdg_output.h
```diff
@@ -0,0 +1,32 @@
+#ifndef WLR_TYPES_WLR_XDG_OUTPUT_MANAGER_H
```
emersion: This `ifndef` should be updated too

PR #861, opened 2018-04-12
Fix wlr_output_transform_compose

### LINKED ISSUE #796: Damage tracking is fucked up with flipped-270 output transform

emersion:
It was like this probably because we inverted flipped-90 and flipped-270 in the past.

Test plan: check that damage tracking works with all transforms.

Fixes #796


SirCmpwn:
danke

PR #865, opened 2018-04-13
Fix gcc string truncation warnings

martinetd:
Or if you don't like these let's add -Wno-format-truncation -Wno-stringop-truncation (the later only happens in -O2, but I'm not sure what enables both as I only get them on fedora)

likewise, I have the same for sway, just waiting to see how this is received to push.

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -55,7 +55,7 @@ struct wlr_output {
 	struct wl_global *wl_global;
 	struct wl_list wl_resources;
 
-	char name[16];
+	char name[24];
 	char make[48];
 	char model[16];
 	char serial[16];
--- a/examples/support/config.c
+++ b/examples/support/config.c
@@ -202,7 +202,10 @@ struct example_config *parse_args(int argc, char *argv[]) {
 		char cwd[MAXPATHLEN];
 		if (getcwd(cwd, sizeof(cwd)) != NULL) {
 			char buf[MAXPATHLEN];
-			snprintf(buf, MAXPATHLEN, "%s/%s", cwd, "wlr-example.ini");
+			if (snprintf(buf, MAXPATHLEN, "%s/%s", cwd, "wlr-example.ini") >= MAXPATHLEN) {
+				wlr_log(L_ERROR, "config path too long");
+				exit(1);
+			}
 			config->config_path = strdup(buf);
 		} else {
 			wlr_log(L_ERROR, "could not get cwd");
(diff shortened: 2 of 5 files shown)


include/wlr/types/wlr_output.h
```diff
@@ -55,7 +55,7 @@ struct wlr_output {
 	struct wl_global *wl_global;
 	struct wl_list wl_resources;
 
-	char name[16];
+	char name[24];
```
martinetd: That one is because of HEADLESS-<maxint> being too large, could also check for the snprintf return value instead.

```
../backend/headless/output.c: In function ‘wlr_headless_add_output’:
../backend/headless/output.c:124:65: error: ‘%d’ directive output may be truncated writing between 1 and 11 bytes into a region of size 7 [-Werror=format-truncation=]
  snprintf(wlr_output->name, sizeof(wlr_output->name), "HEADLESS-%d",
                                                                 ^~
../backend/headless/output.c:124:55: note: directive argument in the range [-2147483647, 2147483647]
  snprintf(wlr_output->name, sizeof(wlr_output->name), "HEADLESS-%d",
                                                       ^~~~~~~~~~~~~
../backend/headless/output.c:124:2: note: ‘snprintf’ output between 11 and 21 bytes into a destination of size 16
  snprintf(wlr_output->name, sizeof(wlr_output->name), "HEADLESS-%d",
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   wl_list_length(&backend->outputs) + 1);
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

emersion: Note: this 16 comes from the EDID spec.

martinetd: I'll revert it and check the return value then. I don't see much use of having > 999999 HEADLESS outputs...

SirCmpwn: On the contrary

SirCmpwn: wl_outputs have nothing to do with EDID

martinetd: Fair enough, less work if I don't have to do anything :)

emersion: Then we shouldn't have fixed-size fields?

SirCmpwn: nah it' sk

emersion: My bad, this is the only fixed-sized field that isn't part of EDID.

ascent12: I'm pretty sure I set it to 16 because it was big enough to fit all of the DRM names with a reasonable number of outputs (i.e. not having 4 billion) and it was a nice round number.

PR #867, opened 2018-04-14
[WIP] Generalize xdg-shell popups and add to layer-shell

SirCmpwn:
* [x] Server implementation
* [x] Rendering
  * [ ] At the correct depth
  * [x] At the correct position
* [ ] Damage
* [ ] Grabs
* [x] Example client

### DIFF
--- a/types/wlr_layer_shell.c
+++ b/types/wlr_layer_shell.c
@@ -137,7 +137,7 @@ static void layer_surface_handle_get_popup(struct wl_client *client,
 	struct wlr_layer_surface *parent =
 		layer_surface_from_resource(layer_resource);
 	struct wlr_xdg_surface *popup_surface =
-		wlr_xdg_surface_from_resource(popup_resource);
+		wlr_xdg_surface_from_popup_resource(popup_resource);
 
 	assert(popup_surface->role == WLR_XDG_SURFACE_ROLE_POPUP);
 	struct wlr_xdg_popup *popup = popup_surface->popup;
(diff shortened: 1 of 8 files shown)


agx:
The popup parts that are outside the layer-shell in examples/layer-shell are not damaged properly but that's I guess that's expected since the todo list lacks a tick at "damage". Looks good otherwise!

types/wlr_layer_shell.c
```diff
@@ -435,3 +448,74 @@ void wlr_layer_shell_destroy(struct wlr_layer_shell *layer_shell) {
...
+
+static void layer_surface_iterator(struct wlr_surface *surface,
+		int sx, int sy, void *data) {
+	struct layer_surface_iterator_data *iter_data = data;
+	iter_data->user_iterator(surface, iter_data->x + sx, iter_data->y + sy,
+		iter_data->user_data);
+}
+
+static void xdg_surface_for_each_surface(struct wlr_xdg_surface *surface,
+		int x, int y, wlr_surface_iterator_func_t iterator, void *user_data) {
```
emersion: You can use `wlr_xdg_surface_for_each_surface` instead of copy-pasting it here.

SirCmpwn: I had issues with that

emersion: It shouldn't be hard, you can use the `data` struct you defined earlier

SirCmpwn: That wasn't the issue. It was related to double-rendering popups

SirCmpwn:
Closing in favor of https://github.com/swaywm/wlroots/pull/877

PR #869, opened 2018-04-15
Some changes to the x11 backend

psychon:
The first commit is a refactoring with (I think) obviously no change in behaviour.
The second commit has a harmless, internal-only behaviour change that should not make any difference.
The last two commits in this deal with resizing. They can be tested by starting rootston with the X11 backend, (making it floating if on a tiling WM,) and resizing it with the mouse a lot.

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -114,9 +114,8 @@ struct wlr_output *wlr_x11_output_create(struct wlr_backend *backend) {
 		wl_list_length(&x11->outputs) + 1);
 	parse_xcb_setup(wlr_output, x11->xcb_conn);
 
-	uint32_t mask = XCB_CW_BACK_PIXEL | XCB_CW_EVENT_MASK;
-	uint32_t values[2] = {
-		x11->screen->white_pixel,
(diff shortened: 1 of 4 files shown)


backend/x11/output.c
```diff
@@ -114,9 +114,8 @@ struct wlr_output *wlr_x11_output_create(struct wlr_backend *backend) {
 		wl_list_length(&x11->outputs) + 1);
 	parse_xcb_setup(wlr_output, x11->xcb_conn);
 
-	uint32_t mask = XCB_CW_BACK_PIXEL | XCB_CW_EVENT_MASK;
+	uint32_t mask = XCB_CW_EVENT_MASK;
 	uint32_t values[2] = {
```
ascent12: Can this be changed to a single uint32_t now?

emersion: Err, we shouldn't specify the size of these arrays explicitly.

psychon: Good catch. I changed into a variable length array instead.

emersion:
Thanks!

PR #873, opened 2018-04-17
output: don't emit the mode event if it hasn't changed

### LINKED ISSUE #535: xwayland clients apply scaling twice when switching between TTYs
When switching between TTYs in rootston, xwayland apps double in scaling size. This is a continuation of [this sway bug report](https://github.com/swaywm/sway/issues/1532).

The commit that introduced this regression is `bf1f461eba320cdeded1d5d8ae80caaa9d4cef98`.

The problem line is [this one in that diff](https://github.com/swaywm/wlroots/blob/bf1f461eba320cdeded1d5d8ae80caaa9d4cef98/types/wlr_output.c#L73). We send the "done" event twice and xwayland handles that poorly (e.g it applies scaling twice). This path is only executed when it's re-sent when you switch between TTYs.

This bug is also present in wlc and was probably copied over from there.

Commenting out done signal send works, but I haven't gone through the logic on master to see if it's actually necessary here.

emersion:
This also fixes #535.

@Timidger, can you confirm?

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -496,11 +496,15 @@ static bool wlr_drm_connector_set_mode(struct wlr_output *output,
 	conn->state = WLR_DRM_CONN_CONNECTED;
 	wlr_output_update_mode(&conn->output, mode);
 
+	// When switching VTs, the mode is not updated but the buffers become
+	// invalid, so we need to manually damage the output here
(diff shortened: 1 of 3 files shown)


Timidger:
Yes this fixed the issue. Though there is also a pending patch to xwayland to fix this fundementally, it should be fine to merge this too.

PR #876, opened 2018-04-18
DPMS handling for wlr_output

snaggen:
This pull request adds DPMS handling to wlr_output.
This is needed for https://github.com/swaywm/sway/pull/1824
That fixes  https://github.com/swaywm/sway/issues/541

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -461,6 +461,17 @@ error_conn:
 	return 0;
 }
 
+static void wlr_drm_connector_set_dpms(struct wlr_output *output, enum dpms_enum level) {
+	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
+	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)output->backend;
+
+    if (drmModeObjectSetProperty (drm->fd, conn->id,
+                DRM_MODE_OBJECT_CONNECTOR,
+                conn->props.dpms, level) < 0)
+        wlr_log(L_ERROR, "Failed to set dpms state for %s", conn->output.name);
+
+}
+
 static bool wlr_drm_connector_set_mode(struct wlr_output *output,
 		struct wlr_output_mode *mode) {
 	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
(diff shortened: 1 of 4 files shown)


backend/drm/drm.c
```diff
@@ -461,6 +461,17 @@ static uint32_t get_possible_crtcs(int fd, uint32_t conn_id) {
 	return 0;
 }
 
+static void wlr_drm_connector_set_dpms(struct wlr_output *output, enum dpms_enum level) {
+	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
+	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)output->backend;
+
+    if (drmModeObjectSetProperty (drm->fd, conn->id,
```
emersion: This is using the legacy DRM interface. DPMS levels have been removed from the atomic interface IIRC. Should we handle DPMS levels at all?

snaggen: Ok, must admit I'm no drm expert, so I looked at weston and mutter for reference and they both had levels. But if we shouldn't use levels, would wlr_drm_connector_enable fit my usecase then?
In that case I can throw away some code...

emersion: Yeah, the initial plan was to use `wlr_output_enable`. I'm not a DRM expert either though, so I'm not sure it's the right thing to do™. I know that Mutter doesn't support atomic modesetting and that Weston only recently gained support for it, so they might not be the best examples to look at wrt. the atomic interface.

emersion: cc @ascent12

ascent12: `wlr_drm_connector_enable` is already the DPMS function.

snaggen: Just tested wlr_output_enable, but that seemed to turn off the screen so it didn't wake up again.

snaggen: Must have been some other glitch... tested again and it works! 
So, that means this pull request is void. And can be discarded...

snaggen:
Have rewritten sway part to use wlr_output_enable instead, since levels shouldn't be used.

ascent12:
I know this is closed and everything, but i'll just point out that DPMS "levels" are completely useless these days. It used to be related to some shit CRT monitors used to do, but means nothing to LEDs.
These days, all of the levels that aren't "on" are just converted to "off".

emersion:
Thanks for the info, I've been wondering about that!

agx:
See e.g. https://lwn.net/Articles/653466/ "the atomic helpers only have on and off."

PR #877, opened 2018-04-19
Layer popups: work on top of PR 867

agx:
- Extend the example client to allow to close the popup and handle mouse input
- Send frame_done events so surface updates work
- Damage popups

TODO (in followups): 
- [ ] handle popup with popups using surface_iterators
- [ ] handle subsurfaces
- [ ] handle grabs (including dismissing popups with grabs when losing focus, see xdg_popup.grab)

### DIFF
--- a/examples/layer-shell.c
+++ b/examples/layer-shell.c
@@ -165,14 +165,19 @@ static void create_popup() {
 	}
 	struct wl_surface *surface = wl_compositor_create_surface(compositor);
 	assert(xdg_wm_base && surface);
-	struct xdg_surface *xdg_surface = xdg_wm_base_get_xdg_surface(
-			xdg_wm_base, surface);
-	struct xdg_positioner *xdg_positioner = xdg_wm_base_create_positioner(
-			xdg_wm_base);
+	struct xdg_surface *xdg_surface =
+		xdg_wm_base_get_xdg_surface(xdg_wm_base, surface);
+	struct xdg_positioner *xdg_positioner =
+		xdg_wm_base_create_positioner(xdg_wm_base);
 	assert(xdg_surface && xdg_positioner);
 
 	xdg_positioner_set_size(xdg_positioner, 256, 256);
-	xdg_positioner_set_anchor_rect(xdg_positioner, 0, 0, width, height);
+	xdg_positioner_set_offset(xdg_positioner, 0, 0);
+	xdg_positioner_set_anchor_rect(xdg_positioner, cur_x, cur_y, 1, 1);
+	xdg_positioner_set_anchor(xdg_positioner, XDG_POSITIONER_ANCHOR_TOP_LEFT);
+	xdg_positioner_set_gravity(xdg_positioner, XDG_POSITIONER_GRAVITY_TOP_LEFT);
+	xdg_positioner_set_constraint_adjustment(xdg_positioner,
+			XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_NONE);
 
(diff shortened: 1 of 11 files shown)


agx:
Pointer input works, destroying popups as well.

agx:
I have a branch of my phone shelk PoC here that uses layer shell instead of a private protocol to test things (besides the layer-shell in examples/): https://code.puri.sm/guido.gunther/phosh/commits/WiP/layer-surfaces

SirCmpwn:
Found bug, possibly with phosh rather than rootston: cannot dismiss the popups by clicking outside of them

agx:
@SirCmpwn yeah, that is a bug in phosh atm (see commit message of e83f1b5f5861c8b640911e14f458124e43e55430). 

 examples/layer-shell now closes the popup when right clicking in the parent.

SirCmpwn:
This might be born from ignorance of how xdg popups actually work but isn't the compositor meant to close them when you click about elsewhere?

SirCmpwn:
Just checked, we send popup_done to xdg surfaces when we focus elsewhere (but not when we click on something else within the parent surface). We should be doing the same thing for layer surfaces.

agx:
O.k., I can add that

agx:
From what I read in xdg_shell the popup is only dismissed when a pointer grab ends. Since we're not doing grabs yet the current behaviour is IMHO correct. Grabs are TODO though (and then the popup needs to be dismissed when clicked outside)

SirCmpwn:
I see, thanks for clarifying.

SirCmpwn:
Can you enumerate the TODOs for a follow-up ticket

examples/layer-shell.c
```diff
@@ -188,12 +188,20 @@ static void xdg_popup_configure(void *data, struct xdg_popup *xdg_popup,
 	}
 }
 
-static void xdg_popup_done(void *data, struct xdg_popup *xdg_popup) {
+static void popup_destroy ()
```
SirCmpwn: style errors here: no space before `()` and put the brace on the same line

examples/layer-shell.c
```diff
@@ -300,7 +308,11 @@ static void wl_pointer_button(void *data, struct wl_pointer *wl_pointer,
 	if (input_surface == wl_surface) {
 		if (state == WL_POINTER_BUTTON_STATE_PRESSED) {
 			if (button == BTN_RIGHT) {
-				create_popup();
+				if (popup_wl_surface) {
+					popup_destroy ();
```
SirCmpwn: style issue here as well

examples/layer-shell.c
```diff
@@ -188,12 +188,20 @@ static void xdg_popup_configure(void *data, struct xdg_popup *xdg_popup,
...
 	xdg_popup_destroy(popup);
+	wl_surface_destroy(popup_wl_surface);
+	popup_wl_surface = NULL;
 	popup = NULL;
+	popup_egl_window = NULL;
+}
+
+static void xdg_popup_done(void *data, struct xdg_popup *xdg_popup) {
+	wlr_log (L_DEBUG, "Popup done");
+	popup_destroy ();
```
SirCmpwn: And these two lines

SirCmpwn:
LGTM aside from minor issues

agx:
Style nits fixed. I've added the TODOs to the PR description and will file a ticket once this is merged.

I'll have a look at grabs next.

SirCmpwn:
Thank you very much!

PR #879, opened 2018-04-20
[WIP] Add wlr_idle_listen, needed by sway idle_v2

snaggen:
Adding wlr_idle_listen to use idle timers without going through the idle protocol, but using the timers directly. Needed by sway idle_v2 https://github.com/swaywm/sway/pull/1831


emersion:
Two general comments:
1. About `wlr_idle_listen`: I don't think we want to expose the timers in `wlr_idle`. Better to use timers in the compositor, it's really not hard.
2. The session interface should be generic. That means "releasing the lock" isn't going to be always "closing a file". You need to abstract this. Also, the callback should be replaced with a signal.

backend/session/logind.c
```diff
@@ -409,10 +416,132 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
 
+int aquire_sleep_lock(struct wlr_session *base) {
+	struct logind_session *session = wl_container_of(base, session, base);
+	int fd = -1;
+	sd_bus_message *msg = NULL;
+	sd_bus_error error = SD_BUS_ERROR_NULL;
+
+	int ret = sd_bus_call_method(session->bus, "org.freedesktop.login1",
+			"/org/freedesktop/login1", "org.freedesktop.login1.Manager", "Inhibit",
+			&error, &msg, "ssss", "sleep", "sway-idle", "Setup Up Lock Screen", "delay");
```
emersion: This is wlroots, not sway, so this shouldn't mention sway

backend/session/session.c
```diff
@@ -350,3 +350,11 @@ size_t wlr_session_find_gpus(struct wlr_session *session,
 	return i;
 #endif
 }
+
+void wlr_session_prepare_for_sleep_listen(struct wlr_session *session, wlr_session_sleep_listener callback, wlr_session_sleep_lock_condition release_condition, void *data) {
+	if (!session || !session->impl || !session->impl->prepare_sleep_listen) {
```
emersion: This function shouldn't silently do nothing if `session` is NULL, this will make bugs hard to find.

include/wlr/backend/session.h
```diff
@@ -83,4 +85,6 @@ bool wlr_session_change_vt(struct wlr_session *session, unsigned vt);
 size_t wlr_session_find_gpus(struct wlr_session *session,
 	size_t ret_len, int *ret);
 
+void wlr_session_prepare_for_sleep_listen(struct wlr_session *session, wlr_session_sleep_listener callback, wlr_session_sleep_lock_condition release_condition, void *data);
```
emersion: These callbacks should be signals.

include/wlr/types/wlr_idle.h
```diff
@@ -4,6 +4,25 @@
...
+
+	/**
+	 * Triggered on the first user activity after an idle event
+	 *
+	 *
+	 */
+	wl_event_loop_timer_func_t resumed;
+	//    void (*resumed)(void *data,
+	//            struct org_kde_kwin_idle_timeout *org_kde_kwin_idle_timeout);
+};
```
emersion: No need for this anymore

backend/session/logind.c
```diff
@@ -409,10 +416,132 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
+		release_lock(session);
+	}
+
+	if (session->prepare_sleep_callback) {
+		wlr_log(L_INFO, "Calling sleep callback, holding lock %d", session->ongoing_lock);
+		session->prepare_sleep_callback(base, session->prepare_sleep_callback_data);
+	}
+
+	if (session->ongoing_lock >= 0) {
+		struct wl_event_source *source = wl_event_loop_add_timer(session->event_loop, test_release_condition, session);
```
emersion: 80 char limit

backend/session/logind.c
```diff
@@ -409,10 +416,132 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
+		return 0;
+	}
+
+	session->inhibit_cnt++;
+	if(session->inhibit_cnt > 10) {
+		wlr_log(L_DEBUG, "Reached inhibit timeout, releasing lock");
+		release_lock(session);
+		return 0;
+	}
+	struct wl_event_source *source = wl_event_loop_add_timer(session->event_loop, test_release_condition, session);
```
emersion: 80 char limit

backend/session/logind.c
```diff
@@ -409,10 +416,132 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
+	}
+
+	if (session->prepare_sleep_callback) {
+		wlr_log(L_INFO, "Calling sleep callback, holding lock %d", session->ongoing_lock);
+		session->prepare_sleep_callback(base, session->prepare_sleep_callback_data);
+	}
+
+	if (session->ongoing_lock >= 0) {
+		struct wl_event_source *source = wl_event_loop_add_timer(session->event_loop, test_release_condition, session);
+		wl_event_source_timer_update(source, 100);
```
emersion: This shouldn't use a timer that tries, each 100ms, to release the lock.

PR #882, opened 2018-04-21
Remove wlr_ prefix from local symbols

emersion:
This removes the long list of prefixed-but-local symbols.


ascent12:
The `libinput_` and `drm_` prefixes are encroaching on the "namespace" of their respective libraries. It could lead to a lot of confusion when debugging, and I've even run into a problem where I accidentally used one of our constants instead of a libdrm constant before.
They probably should be renamed to something else.

emersion:
>They probably should be renamed to something else.

The only reasonable prefix I can think of is `backend_{drm,libinput}`. This isn't consistent with global functions names (prefixed with `wlr_{drm,libinput}`).

SirCmpwn:
`backend_{drm,libinput}` is meh but I don't have a better suggestion

emersion:
We could also assume the DRM module is self-contained and drop the `drm_` prefix.

ascent12:
I'd be fine with that.

emersion:
>We could also assume the DRM module is self-contained and drop the drm_ prefix.

Hmm, it seems it's not possible to apply the symbols file to archives. That means all symbols are global to wlroots.

emersion:
>but would it matter? You basically want to have them as local symbols in the generated library/shared object so programs can't link against them.

It would have been cool to be able to define the same symbol in the libinput backend and in the DRM backend (e.g. `handle_event`).

Now that I think of it, maybe not, because it makes debugging more difficult (even if you generally have the name of the file defining the function).

emersion:
This (boring) PR is ready for another review.

agx:
Looks great!

PR #883, opened 2018-04-21
Do not advertize *RGB8888 as supported shm formats

atomnuker:
They're always available and supported, and are added automatically.
Adding them again just duplicates them.


render/wlr_renderer.c
```diff
@@ -168,6 +168,9 @@ void wlr_renderer_init_wl_shm(struct wlr_renderer *r,
 	}
 
 	for (size_t i = 0; i < len; ++i) {
-		wl_display_add_shm_format(display, formats[i]);
+		if (formats[i] != WL_SHM_FORMAT_ARGB8888 &&
+			formats[i] != WL_SHM_FORMAT_XRGB8888) {
```
emersion: Can you increase the indentation of one level here please?

atomnuker: Done

PR #886, opened 2018-04-21
xwayland: preserve the PATH variable

n3rdopolis:
Some systems don't have xwayland in /usr/bin, but in other paths. 
wlroots was unable to find Xwayland binaries outside of standard locations


emersion:
Thanks!

PR #887, opened 2018-04-22
Update layer shell with optional wl_output

SirCmpwn:


SirCmpwn:
Note, this was already added to the protocol in wlr-protocols, I just didn't bother to implement it in wlroots at the time.

include/wlr/types/wlr_layer_shell.h
```diff
@@ -27,6 +27,9 @@ struct wlr_layer_shell {
 	struct wl_listener display_destroy;
 
 	struct {
+		 // struct wlr_layer_surface *
+		 // Note: the output may be NULL. In this case, it is your
+		 // responsibility to assign an output before returning.
```
emersion: You could `assert(layer_surface->output != NULL)` after emitting the signal.

PR #890, opened 2018-04-23
layer-shell: Fix crash when cursor is intially outside any output

agx:
On the X11 backend the cursor position might be outside the output
window so no output is returned leading to the assert to trigger. Use
sane fallback instead of crashing.

### DIFF
--- a/rootston/layer_shell.c
+++ b/rootston/layer_shell.c
@@ -395,8 +395,18 @@ void handle_layer_shell_surface(struct wl_listener *listener, void *data) {
 			wlr_output_layout_output_at(desktop->layout,
 					seat->cursor->cursor->x,
 					seat->cursor->cursor->y);
-		assert(output); // And this one
-		layer_surface->output = output;
+		if (!output) {
+			wlr_log(L_ERROR, "Couldn't find output at (%.0f,%.0f)",
+				seat->cursor->cursor->x,
+				seat->cursor->cursor->y);
+			output = wlr_output_layout_get_center_output(desktop->layout);
+		}
+		if (output) {
+			layer_surface->output = output;
+		} else {
+			wlr_layer_surface_close(layer_surface);
+			return;
+		}
 	}
 
 	roots_surface->surface_commit.notify = handle_surface_commit;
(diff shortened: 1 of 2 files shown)


rootston/layer_shell.c
```diff
@@ -396,7 +396,12 @@ void handle_layer_shell_surface(struct wl_listener *listener, void *data) {
 			wlr_output_layout_output_at(desktop->layout,
 					seat->cursor->cursor->x,
 					seat->cursor->cursor->y);
-		assert(output); // And this one
+		if (!output) {
+			wlr_log(L_ERROR, "Couldn't find output at (%.0f,%.0f)",
+				seat->cursor->cursor->x,
+				seat->cursor->cursor->y);
+			output = wlr_output_layout_output_at(desktop->layout, 0, 0);
```
emersion: Better to use `get_center_output`?

Also there might be zero outputs, in which case it would be wise to close the layer surface immediately.

types/wlr_layer_shell.c
```diff
@@ -287,7 +287,13 @@ static void handle_wlr_surface_committed(struct wlr_surface *wlr_surface,
 		surface->added = true;
 		wlr_signal_emit_safe(&surface->shell->events.new_surface,
 				surface);
-		assert(surface->output);
+		if (!surface->output) {
+			wl_resource_post_error(surface->resource,
```
emersion: Sorry for the late review, I didn't notice you updated the PR.

Protocol errors should only be used for protocol violations. Here the client cannot guess that the compositor won't find any suitable output. So I think the compositor should instead close immediately the layer surface (and this can become an assert like `surface->output != NULL || surface->closed`).

emersion: Oh, you already close it, good. So I guess what needs to be done here is do not send an error and re-add an assert.

protocol/wlr-layer-shell-unstable-v1.xml
```diff
@@ -65,6 +65,7 @@
       <entry name="role" value="0" summary="wl_surface has another role"/>
       <entry name="invalid_layer" value="1" summary="layer value is invalid"/>
       <entry name="already_constructed" value="2" summary="wl_surface has a buffer attached or committed"/>
+      <entry name="bad_output" value="3" summary="Output could not be determined"/>
```
SirCmpwn: `No suitable output could be assigned`

agx:
Yeah, I was a bit undecided if this is a protocol error since the client requested s.th. the compositor could not fulfill or not. Updated.

PR #891, opened 2018-04-23
examples/layer-shell: take grab on the popup

agx:
This demonstrates popus can be close by clicking on the area outside the
popup and parent.

There's no meaningful serial at this point so just pick one.


emersion:
The protocol states that:

>This request must be used in response to some sort of user action like a
button press, key press, or touch down event. The serial number of the
event should be passed as 'serial'.

Maybe a click on the layer shell should open the popup in the example?

SirCmpwn:
>There's no meaningful serial at this point so just pick one.

This seems... wrong. There is a meaningful serial, you just gotta pass it into the create_popup function from the pointer event that calls it

agx:
My bad, I mixed up popup and initial window creation. We have a serial for the click, passed no now.

examples/layer-shell.c
```diff
@@ -209,7 +209,7 @@ static const struct xdg_popup_listener xdg_popup_listener = {
 	.popup_done = xdg_popup_done,
 };
 
-static void create_popup() {
+static void create_popup(int serial) {
```
emersion: This isn't an `int`, it's an `uint32_t`.

agx:
Indeed. Fixed as well.

emersion:
Thanks!

PR #893, opened 2018-04-23
Fix protocol error when a client destroys and re-creates a subsurface

### LINKED ISSUE #888: Popups in Gedit crash the whole app or the compositor
This happened at commit https://github.com/swaywm/wlroots/pull/885/commits/fa84b267e0fdc3bcdd43dc2d06b55ffad03059a3

Steps to reproduce: Start rootston, start gedit, move cursor over the "Open" menu. A popup should open, but instead gedit crashes, or sometimes rootston. Doesn't happen before this commit. I'm still investigating what exactly is going on.

emersion:
Fixes #888
Fixes swaywm/sway#1842

### DIFF
--- a/types/wlr_compositor.c
+++ b/types/wlr_compositor.c
@@ -42,7 +42,8 @@ static void subcompositor_handle_get_subsurface(struct wl_client *client,
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


ammen99:
I can confirm this fixes the issue

PR #895, opened 2018-04-24
layer-shell: use output_damage_{whole,from}_local_surface

agx:
Reuse what already handles rotation and scaling. This unbreaks popups
on rotated or scaled outputs.

Can be tested with layer-shell example.

### DIFF
--- a/rootston/layer_shell.c
+++ b/rootston/layer_shell.c
@@ -255,7 +255,8 @@ static void unmap(struct wlr_layer_surface *layer_surface) {
 	struct wlr_output *wlr_output = layer_surface->output;
 	if (wlr_output != NULL) {
 		struct roots_output *output = wlr_output->data;
-		wlr_output_damage_add_box(output->damage, &layer->geo);
+		output_damage_whole_local_surface(output, layer_surface->surface,
+			layer->geo.x, layer->geo.y, 0);
 	}
 }
 
@@ -282,7 +283,8 @@ static void handle_map(struct wl_listener *listener, void *data) {
 	struct roots_layer_surface *layer = layer_surface->data;
 	struct wlr_output *wlr_output = layer_surface->output;
 	struct roots_output *output = wlr_output->data;
-	wlr_output_damage_add_box(output->damage, &layer->geo);
+	output_damage_whole_local_surface(output, layer_surface->surface,
+		layer->geo.x, layer->geo.y, 0);
 	wlr_surface_send_enter(layer_surface->surface, wlr_output);
 }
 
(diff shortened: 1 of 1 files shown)


rootston/layer_shell.c
```diff
@@ -297,35 +297,32 @@ static void popup_handle_map(struct wl_listener *listener, void *data) {
 	struct roots_layer_surface *layer = popup->parent;
 	struct wlr_output *wlr_output = layer->layer_surface->output;
 	struct roots_output *output = wlr_output->data;
-	struct wlr_box geom;
-	memcpy(&geom, &popup->wlr_popup->geometry, sizeof(struct wlr_box));
-	geom.x += layer->geo.x;
-	geom.y += layer->geo.y;
-	wlr_output_damage_add_box(output->damage, &geom);
+	int ox = popup->wlr_popup->geometry.x;
```
emersion: Can you sum `layer->geo.x` in the `ox` definition? Otherwise it's not really the position in output-local coords.

agx:
@emersion yeah that much less confusing. Updated.

rootston/layer_shell.c
```diff
@@ -297,35 +297,32 @@ static void popup_handle_map(struct wl_listener *listener, void *data) {
...
 	struct roots_output *output = wlr_output->data;
-	struct wlr_box geom;
-	memcpy(&geom, &popup->wlr_popup->geometry, sizeof(struct wlr_box));
-	geom.x += layer->geo.x;
-	geom.y += layer->geo.y;
-	wlr_output_damage_add_box(output->damage, &geom);
+	int ox = popup->wlr_popup->geometry.x + layer->geo.x;
+	int oy = popup->wlr_popup->geometry.y + layer->geo.y;
+	output_damage_from_local_surface(output, popup->wlr_popup->base->surface,
+		ox, oy, 0);
```
emersion: I think this should be `output_damage_whole_local_surface`

agx:
BTW I wonder if layer_shell::unmap shouldn't better use output_damage_whole_local_surface as well?

rootston/layer_shell.c
emersion: Basically, `from` can only be used in commit handlers

emersion:
>BTW I wonder if layer_shell::unmap shouldn't better use output_damage_whole_local_surface as well?

Probably. Same for `map`. Patches welcome!

agx:
@emersion  Ah, o.k. I probably didn't notice since the commit came right after the map. Thanks!

agx:
Fixed layer_shell::{,un}map too

PR #903, opened 2018-04-24
render/egl: add wlr_egl_destroy_surface

emersion:
This also fixes the `eglDestroySurface` error under the Wayland backend.

### DIFF
--- a/backend/headless/output.c
+++ b/backend/headless/output.c
@@ -29,9 +29,7 @@ static bool output_set_custom_mode(struct wlr_output *wlr_output, int32_t width,
 		refresh = HEADLESS_DEFAULT_REFRESH;
 	}
 
-	if (output->egl_surface) {
-		wlr_egl_destroy_surface(&backend->egl, output->egl_surface);
-	}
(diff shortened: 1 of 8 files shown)


render/egl.c
```diff
@@ -499,3 +499,7 @@ int wlr_egl_get_dmabuf_modifiers(struct wlr_egl *egl,
 	}
 	return num;
 }
+
+void wlr_egl_destroy_surface(struct wlr_egl *egl, EGLSurface surface) {
+	eglDestroySurface(egl->display, surface);
```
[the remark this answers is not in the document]
emersion: Good idea. Just done that.

backend/headless/output.c
```diff
@@ -30,7 +30,7 @@ static bool output_set_custom_mode(struct wlr_output *wlr_output, int32_t width,
 	}
 
 	if (output->egl_surface) {
```
ascent12: Another check to remove here.

emersion: Right, fixed

PR #904, opened 2018-04-25
Normalise scroll values

### LINKED ISSUE #848: Xwayland double axis events
When scrolling on an Xwayland window, sometimes there will be two axis events sent. It's not every time, but happens very frequently. It's easiest to notice on a web browser, where the page will randomly scroll a lot further than normal.

Trying a wayland application with WAYLAND_DEBUG=1, I couldn't ever get it to send more than 1 event.
Trying libinput debug_events, I also couldn't get it to have more than 1 event.
Looking at xev, I can see that there a multiple events being sent to it, so it's definitely related to Xwayland.

This affects both rootston and sway, so it's probably a wlroots problem.

ascent12:
Fixes #848 

Apparently Xwayland will accumulate the scroll amount between events, and each 10 "units" of scroll send one X11 scroll event.
Since libinput was sending 15 units per mouse scroll for me, 2 actual scrolls would end up as 3 X11 scrolls.

You can see the weston fix here: https://github.com/wayland-project/weston/blob/master/libweston/libinput-device.c#L184

I also took the liberty of cleaning up the libinput axis function a bit.

### DIFF
--- a/backend/libinput/pointer.c
+++ b/backend/libinput/pointer.c
@@ -95,9 +95,9 @@ static enum wlr_axis_source axis_source_to_wlr(
 		return WLR_AXIS_SOURCE_CONTINUOUS;
 	case LIBINPUT_POINTER_AXIS_SOURCE_WHEEL_TILT:
 		return WLR_AXIS_SOURCE_WHEEL_TILT;
-	default:
-		abort();
 	}
+
+	abort();
 }
 
 static enum wlr_axis_orientation axis_orientation_to_wlr(
@@ -107,9 +107,9 @@ static enum wlr_axis_orientation axis_orientation_to_wlr(
 		return WLR_AXIS_ORIENTATION_VERTICAL;
 	case LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL:
 		return WLR_AXIS_ORIENTATION_HORIZONTAL;
-	default:
-		abort();
 	}
+
+	abort();
 }
 
 static double normalize_axis(struct libinput_event_pointer *pevent,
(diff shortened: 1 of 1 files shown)


backend/libinput/pointer.c
```diff
@@ -84,6 +84,51 @@ void handle_pointer_button(struct libinput_event *event,
...
+}
+
+static enum wlr_axis_orientation axis_orientation_to_wlr(
+		enum libinput_pointer_axis orientation) {
+	switch (orientation) {
+	case LIBINPUT_POINTER_AXIS_SCROLL_VERTICAL:
+		return WLR_AXIS_ORIENTATION_VERTICAL;
+	case LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL:
+		return WLR_AXIS_ORIENTATION_HORIZONTAL;
+	default:
```
emersion: I usually like to omit the `default` if we handle all the enum items - this allows the compiler to check at compilation if we're missing one case because libinput has added a new one.

backend/libinput/pointer.c
```diff
@@ -92,43 +137,32 @@ void handle_pointer_axis(struct libinput_event *event,
...
+		.time_msec = usec_to_msec(time_usec),
+		.source = axis_source_to_wlr(source),
+	};
+
 	enum libinput_pointer_axis axies[] = {
 		LIBINPUT_POINTER_AXIS_SCROLL_VERTICAL,
 		LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL,
 	};
+
 	for (size_t i = 0; i < sizeof(axies) / sizeof(axies[0]); ++i) {
```
emersion: The plural of `axis` is `axes`. Or is there another version of the plural form I'm not aware of?

ascent12: You're correct. This is incorrect.

ascent12:
I've looked at the Xwayland code. They're not using the discrete event.

SirCmpwn:
>I've looked at the Xwayland code. They're not using the discrete event.

Patch them?

SirCmpwn:
Xwayland bug.

PR #906, opened 2018-04-25
[WIP] Add DMA-BUF export support

emersion:
EGL docs: https://www.khronos.org/registry/EGL/extensions/MESA/EGL_MESA_image_dma_buf_export.txt

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -177,9 +177,12 @@ bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *remote_display,
 	egl->egl_exts.dmabuf_import_modifiers =
 		check_egl_ext(egl->exts_str, "EGL_EXT_image_dma_buf_import_modifiers")
 		&& eglQueryDmaBufFormatsEXT && eglQueryDmaBufModifiersEXT;
+	egl->egl_exts.dmabuf_export =
+		check_egl_ext(egl->exts_str, "EGL_MESA_image_dma_buf_export");
 
 	egl->egl_exts.bind_wayland_display =
 		check_egl_ext(egl->exts_str, "EGL_WL_bind_wayland_display");
+
 	print_dmabuf_formats(egl);
 
 	return true;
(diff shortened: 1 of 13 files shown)


render/egl.c
```diff
@@ -413,6 +416,40 @@ EGLImageKHR wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
+		int32_t width, int32_t height, uint32_t flags,
+		struct wlr_dmabuf_buffer_attribs *attribs) {
+	memset(attribs, 0, sizeof(struct wlr_dmabuf_buffer_attribs));
+
+	if (!egl->egl_exts.dmabuf_export || !eglExportDMABUFImageQueryMESA ||
+			!eglExportDMABUFImageMESA) {
+		return false;
+	}
+
+	EGLuint64KHR modifier;
```
agx: Should this be 
```
EGLuint64KHR modifiers[WLR_LINUX_DMABUF_MAXPLANES}; 
```
since the modifiers are per plane?

emersion: The first call assumes the caller doesn't know how many planes the image has, so there's only one modifier. This is a limitation of the EGL API I guess.

agx: Looking at mesa it's an array:

https://sources.debian.org/src/mesa/17.3.9-1/src/gallium/drivers/etnaviv/etnaviv_screen.c/?hl=591#L604

agx:
Nice! Would be awesome to have a test client.

render/egl.c
```diff
@@ -413,6 +416,40 @@ EGLImageKHR wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
+	if (!eglExportDMABUFImageQueryMESA(egl->display, image,
+			(int *)&attribs->format, &attribs->n_planes, &modifier)) {
+		return false;
+	}
+	if (attribs->n_planes > WLR_LINUX_DMABUF_MAX_PLANES) {
+		wlr_log(L_ERROR, "EGL returned more planes than supported");
+		return false;
+	}
+	for (int i = 0; i < attribs->n_planes; ++i) {
+		attribs->modifier[i] = modifier;
```
agx: Need to copy per plane modifiers here (rather than using the one for the first plane for all planes).

render/egl.c
emersion: The EGL docs say that when making the `eglExportDMABUFImageQueryMESA` call the caller doesn't know about the number of planes. The `eglExportDMABUFImageMESA` call is used to retrieve per-plane properties.

agx: Sorry for being unclear, the main thing I'm trying to say is that the loop in https://github.com/swaywm/wlroots/pull/906/files#diff-7d28873cf02c85696a511bcb3a1521daR439 should use per plane modifiers.

emersion: Hmm, I still don't get it x)

Are per-plane modifiers encoded in this single int somehow?

render/egl.c
agx:
```
attribs->modifier[i] = modifiers[i];
```

emersion: How do I know the length of the array to pass to `eglExportDMABUFImageQueryMESA`?

render/egl.c
agx: Note the plural form ```modifiers``` in https://www.khronos.org/registry/EGL/extensions/MESA/EGL_MESA_image_dma_buf_export.txt and how https://sources.debian.org/src/mesa/17.3.9-1/src/gallium/drivers/etnaviv/etnaviv_screen.c/?hl=591#L614 fills in the values per plane.

emersion: I think this function is used for `eglQueryDmaBufModifiersEXT`, not for `eglExportDMABUFImageQueryMESA`.

render/egl.c
agx: Rereading the mesa code I'm plain wrong here. It's returning the supported modifiers. Scratch all that and sorry for the confusion.

render/egl.c
agx: See below, I mixed up things. Sorry for the confusion.

emersion: No worries, this was confusing for me too!

PR #911, opened 2018-04-26
Fix WLR_DIRECTION enums

RyanDwyer:
Having 0 as an enum value causes wlr_output_layout_adjacent_output() to
never match WLR_DIRECTION_UP.

See here: https://github.com/swaywm/wlroots/blob/31857c9ed40a6bf34273b0459d2aa604642d1807/types/wlr_output_layout.c#L458

To test:

* In sway, have two outputs arranged vertically.
* With focus on the bottom output, do `focus up`

### DIFF
--- a/include/wlr/types/wlr_output_layout.h
+++ b/include/wlr/types/wlr_output_layout.h
@@ -105,9 +105,9 @@ struct wlr_output *wlr_output_layout_get_center_output(
 		struct wlr_output_layout *layout);
 
 enum wlr_direction {
-	WLR_DIRECTION_UP = 0,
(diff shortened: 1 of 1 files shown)


emersion:
If we do bitwise operations like this, shouldn't the enum have power of 2 values?

SirCmpwn:
We're treating these like a bitfield? If so, then yeah, they should be powers of two. But are we doing that?

emersion:
Well…

```c
if (direction & WLR_DIRECTION_UP)
```

SirCmpwn:
Then yeah, that's wrong.

PR #912, opened 2018-04-26
Remove bitchecks from wlr_output_layout_adjacent_output()

RyanDwyer:


emersion:
Hmm, the old code seems to allow the user to pass a direction like `WLR_DIRECTION_LEFT | WLR_DIRECTION_UP`.

PR #914, opened 2018-04-26
Misc segfault fixes

emersion:


emersion:
>Where these spotted by accident?

Somewhere in #882

PR #915, opened 2018-04-26
compositor: redesign how resources are managed

emersion:
All public resource creators now take a new ID for the resource
and an optional list where the resource link is added. When the
resource is destroyed it is its own responsibility to remove
itself from the list. This removes the need for the caller to add
a destroy listener.

This commit fixes a few segfaults with resources not removed from
the list when destroyed.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -615,6 +615,8 @@ static void subsurface_destroy(struct wlr_subsurface *subsurface) {
 		wl_list_remove(&subsurface->parent_destroy.link);
 	}
 
+	wl_list_remove(wl_resource_get_link(subsurface->resource));
+
 	wl_resource_set_user_data(subsurface->resource, NULL);
 	if (subsurface->surface) {
 		subsurface->surface->role_data = NULL;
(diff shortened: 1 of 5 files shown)


SirCmpwn:
Thanks!

types/wlr_surface.c
```diff
@@ -856,7 +878,20 @@ struct wlr_subsurface *wlr_surface_make_subsurface(struct wlr_surface *surface,
 	}
 	subsurface->synchronized = true;
 	subsurface->surface = surface;
+	subsurface->resource =
+		wl_resource_create(client, &wl_subsurface_interface, version, id);
```
Ongy: Should these have an upper bound?
The implementation is for surfaces of version 1, which would not work with higher version surfaces (libwayland-client aborts when the listener has a NULL handler, not sure about libwayland-server).

emersion: These interfaces are a special case since their constructor is public. I'm still not sure we should keep these public tbh. But yeah, we should add version checks, these would be useful if the compositor decides to use the constructor on his own without `wlr_subcompositor` and supports a more recent version of wl-subcompositor.

PR #918, opened 2018-04-26
xwayland: forward configure events to compositor when unmapped

### LINKED ISSUE #899: Xwayland configure requests ignored when the view is unmapped
Should forward these to the compositor instead.

emersion:
Fixes #899


SirCmpwn:
Thanks!

PR #923, opened 2018-04-27
cursor: use NAN for unspecified axes, refactor absolute warping code

### LINKED ISSUE #907: wlr_cursor doesn't support absolute coordinates outside 0..1
But libinput docs say it's possible, if the tool is outside the working region of the tablet for instance.

`wlr_cursor` will just ignore them, because negative values mean "ignore me". Maybe an infinite/NAN value would be better?

Note that the cursor shouldn't go outside the mapped region, even if it receives absolute coordinates outside 0..1.

emersion:
Fixes #907

### DIFF
--- a/types/wlr_cursor.c
+++ b/types/wlr_cursor.c
@@ -1,5 +1,6 @@
 #include <assert.h>
 #include <limits.h>
+#include <math.h>
 #include <stdlib.h>
 #include <wayland-server.h>
 #include <wlr/types/wlr_cursor.h>
@@ -175,21 +176,20 @@ static struct wlr_cursor_device *get_cursor_device(struct wlr_cursor *cur,
 }
 
 static void cursor_warp_unchecked(struct wlr_cursor *cur,
-		double x, double y) {
+		double lx, double ly) {
 	assert(cur->state->layout);
 
 	struct wlr_cursor_output_cursor *output_cursor;
 	wl_list_for_each(output_cursor, &cur->state->output_cursors, link) {
-		double output_x = x;
-		double output_y = y;
+		double output_x = lx, output_y = ly;
 		wlr_output_layout_output_coords(cur->state->layout,
 			output_cursor->output_cursor->output, &output_x, &output_y);
-		wlr_output_cursor_move(output_cursor->output_cursor, output_x,
-			output_y);
+		wlr_output_cursor_move(output_cursor->output_cursor,
+			output_x, output_y);
 	}
 
-	cur->x = x;
-	cur->y = y;
+	cur->x = lx;
+	cur->y = ly;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


types/wlr_cursor.c
```diff
@@ -233,69 +233,81 @@ static struct wlr_box *get_mapping(struct wlr_cursor *cur,
...
+}
 
+static void cursor_warp_closest(struct wlr_cursor *cur,
+		struct wlr_input_device *dev, double lx, double ly) {
+	struct wlr_box *mapping = get_mapping(cur, dev);
 	if (mapping) {
-		if (wlr_box_contains_point(mapping, x, y)) {
-			cursor_warp_unchecked(cur, x, y);
-			result = true;
+		if (!wlr_box_contains_point(mapping, lx, ly)) {
```
Ongy: 1) Is there a reason for this indirection over `closest_{x,y}`?
2) Is this explicit check necessary? Without checking the file, I'd expect this to be the first thing `_contains_point` does

Same in the case below

types/wlr_cursor.c
```diff
@@ -233,69 +233,81 @@ static struct wlr_box *get_mapping(struct wlr_cursor *cur,
...
 
 void wlr_cursor_move(struct wlr_cursor *cur, struct wlr_input_device *dev,
 		double delta_x, double delta_y) {
 	assert(cur->state->layout);
 
-	double x = cur->x + delta_x;
-	double y = cur->y + delta_y;
-
-	struct wlr_box *mapping = get_mapping(cur, dev);
+	double lx = !isnan(delta_x) ? cur->x + delta_x : cur->x;
```
Ongy: The negation here is odd. IMO it would be better for readability to not negate and swap the clauses (same for y)

types/wlr_cursor.c
emersion: >Is there a reason for this indirection over closest_{x,y}?

`double` vs `int` thing

>Is this explicit check necessary? Without checking the file, I'd expect this to be the first thing _contains_point does

Ditto

types/wlr_cursor.c
emersion: Hmmm, not sure about this change. I like having the important case first. But I don't like ternary operations like this in the first place, so…

types/wlr_cursor.c
Ongy: I'm not quite sure what you are trying to tell me here.
Both `current_{x,y}` and and `l{x,y}` are doubles. What am I missing?

emersion: `wlr_box_closest_point` takes integers

emersion: Fixed

types/wlr_cursor.c
Ongy: I just thought of it as a bit surprising, but if you want to keep it this way, fine by me

PR #924, opened 2018-04-27
render: introduce wlr_multi_renderer

### LINKED ISSUE #778: Multiple renderers at the same time
We can probably export and import dma-bufs to do this one.

emersion:
I started #906, and then realized there are a lot of issues trying to import/export lazily because switching EGL contexts is required when exporting and importing. So hopefully this PR will make #906 easier to implement.

This introduces `wlr_multi_renderer`, which is similar to `wlr_multi_backend`. The `wlr_multi_backend` now creates a `wlr_multi_renderer` which creates textures on all renderers. When rendering a texture, the compositor must use `wlr_multi_texture_get_child` to get the texture for a specific renderer from a `wlr_multi_texture`.

In the future, it will be possible to optimize this multi renderer with DMA-BUF and the master/slave mechnism explained in #778.

`wlr_multi_renderer` was needed anyway for backend hotplugging (because you don't want to create surfaces with a renderer that could go away).

You can try to run rootston on this branch, you'll get two outputs: one Wayland output and one X11 output. Cursor input is a bit messed up, but rendering works!

- [x] Create textures from pixel buffers
- [x] Create textures from wl-drm
- [x] Create textures from DMA-BUF
- [x] Properly destroy textures
- [x] Properly destroy renderer
- [x] ~~Investigate why using `wlr_egl_bind_display` multiple times fails on shutdown~~

Follow-up issues:
- Hotplugging doesn't work if it changes the set of available formats
- ~~Input events on the Wayland and X11 backends are messed up (#925, fixed by #930)~~
- Allow the headless backend to re-use an existing renderer (see #778)
- Use DMA-BUF to share textures between backends (see #778)

Fixes #778

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -166,10 +166,6 @@ struct wlr_backend *wlr_drm_backend_create(struct wl_display *display,
 		goto error_event;
 	}
 
-	if (!wlr_egl_bind_display(&drm->renderer.egl, display)) {
-		wlr_log(L_INFO, "Failed to bind egl/wl display");
-	}
-
 	drm->display_destroy.notify = handle_display_destroy;
 	wl_display_add_destroy_listener(display, &drm->display_destroy);
 
--- a/backend/multi/backend.c
+++ b/backend/multi/backend.c
@@ -4,6 +4,7 @@
 #include <wlr/backend/drm.h>
 #include <wlr/backend/interface.h>
 #include <wlr/backend/session.h>
+#include <wlr/render/multi.h>
 #include <wlr/util/log.h>
 #include "backend/multi.h"
 #include "util/signal.h"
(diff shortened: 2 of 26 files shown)


SirCmpwn:
Concept :+1: :+1: :+1:

Will review the code shortly

backend/backend.c
```diff
@@ -113,7 +113,7 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
 		struct wlr_backend *wl_backend = attempt_wl_backend(display);
 		if (wl_backend) {
 			wlr_multi_backend_add(backend, wl_backend);
-			return backend;
+			//return backend;
```
SirCmpwn: wat

SirCmpwn:
Code looks pretty good

backend/backend.c
emersion: This is the hack that spawns both a Wayland and a X11 backend. I should have added a comment to make it clear that it's not there to stay. That's the reason why this PR is marked as "DO NOT MERGE".

emersion:
Removed the hack, rebased.

Paging @Ongy and @Timidger since this is a breaking change.

ascent12:
I've been slow to get onto reviewing this PR, but I wanted to think about it before I commented.

I can say I'm really not liking the direction that wlroots is going in terms of the backends and renderer. This PR is a hack in an attempt to try and get the graphics stack to work in a way which is wasn't designed to. This extreme flexibility with backends/renderers you're trying to go for is leading to a crapload of pointless extra complexity.

I don't even know why you'd even want to mix the "EGL native" (DRM/X11/Wayland) backends together. I cannot see a viable reason for them ever being used at the same time. Other backends are much easier, and can do basically everything with a single GL context. I said that on #wayland a little while ago.

We really need to re-evaluate what we want the backends and renderers to actually be, and what is actually supported by the graphics stack.

:-1:

SirCmpwn:
I think I'm inclined to agree with you. The best use-case I could think of was RDP, but that is probably better served through @atomnuker's screencapture protocol plus a protocol for virtual input devices.

PR #930, opened 2018-04-29
Add wlr_input_device.output_name

### LINKED ISSUE #925: Absolute touch events don't work with multiple screens
Absolute touch events are mapped to the whole layout by default. This works if there's only one screen, but fails if there are multiple.

We should provide screen information in the input device.

The same issue forces us to have a rather complicated code to correctly support multiple outputs in the Wayland and X11 backends. This approach fails short when there are multiple backends (#924).

emersion:
This allows the backend to suggest an output mapping for an input device. This is mainly used by the Wayland/X11 backends to suggest a mapping between pointers that give absolute events relative to one surface and the surface itself. This could be used by a backend capable of matching touchscreens to their output.

rootston will try to get a mapping from the config, and default to what the backend suggests.

This removes the nonsense layout coordinates conversions in the Wayland and X11 backends.

Test plan: try opening multiple outputs on the Wayland/X11 backends, check that input events are not messed up. Try adding scaling and transformations to outputs.

Fixes #925

### DIFF
--- a/backend/libinput/events.c
+++ b/backend/libinput/events.c
@@ -1,3 +1,4 @@
+#define _POSIX_C_SOURCE 200809L
 #include <assert.h>
 #include <libinput.h>
 #include <stdlib.h>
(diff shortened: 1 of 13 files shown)


emersion:
This PR is ready for review.

emersion:
>From the reviewing side it would have been awesome if the mechnical wlr_wl_backend_output -> wlr_wl_output rename would have been in a separate commit so it's easier to follow logic changes.

Ugh, you're right, I'm still not used to do this. Will try harder.

PR #931, opened 2018-04-29
[WIP] Redesign resource destruction

### LINKED ISSUE #909: wlr_* structs should be free'd when the wl_resource is
Rationale: it shouldn't be possible to destroy a wlr_* struct without destroying its resource.

Right now, we generally have a `wlr_thing_destroy` that destroys the `wlr_thing` without destroying the resource. Then, we have a `wl_resource` destroy handler which calls `wlr_thing_destroy`. This is an issue since calling `wlr_thing_destroy` won't free the resource (ie. leak memory, and more importantly allow the client to send requests to it and make the compositor crash).

We can't call `wl_resource_destroy` from `wlr_thing_destroy` because we can't call `wl_resource_destroy` from the resource destroy handler. We also still want to destroy the `wlr_thing` if the client disconnects.

So instead we should make `wlr_thing_destroy` just call `wl_resource_destroy` (if we want to keep this function at all), and the resource destroy handler do the actual cleaning.

TODO:
- [x] Core Wayland protocol
- [ ] gamma control
- [ ] idle inhibit
- [ ] idle
- [x] input inhibitor
- [ ] layer shell
- [x] linux dmabuf
- [ ] primary selection
- [x] ~~screenshooter~~ will be replaced
- [x] ~~server decoration~~ will be replaced
- [x] xdg output
- [ ] zxdg-shell v6, xdg-shell stable

---

wlroots has migrated to gitlab.freedesktop.org. This issue has been moved to:

https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/909

emersion:
This PR updates `CONTRIBUTING.md` with new guidelines regarding protocol implementation and updates all protocol implementations to reflect that.

Feedback on the `CONTRIBUTING.md` additions is welcome, even if this PR is marked as WIP.

Fixes #909


CONTRIBUTING.md
```diff
@@ -232,3 +232,118 @@ error_session:
 	return NULL;
 }
 ```
+
+## Wayland protocol implementation
+
+Each protocol generally lives in a file with the same name, containing one
+struct per interface. For instance, `xdg_shell` lives in `types/xdg_shell.h` and
```
SirCmpwn: per protocol

CONTRIBUTING.md
```diff
@@ -232,3 +232,118 @@ error_session:
...
+struct wlr_surface *wlr_surface_from_resource(struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &wl_surface_interface,
+		&surface_impl));
+	return wl_resource_get_user_data(resource);
+}
+```
+
+### Destructing resources
+
+Resource structs should only be destroyed when their resource is destroyed, ie.
```
SirCmpwn: You haven't defined "resource structs" yet, you should pick some consistent terminology and use it throughout

CONTRIBUTING.md
```diff
@@ -232,3 +232,118 @@ error_session:
...
+inert resources should be ignored, except the destructor. This is achieved by:
+
+1. When the resource becomes inert: destroy the struct and call
+   `wl_resource_set_user_data(resource, NULL)`. Do not destroy the resource.
+2. For each request made to a resource that can be inert: add a NULL check to
+   ignore the request if the resource is inert.
+3. When the client calls the destructor request on the resource: call
+   `wl_resource_destroy(resource)` as usual.
+4. When the resource is destroyed, if the resource isn't inert, destroy the
+   struct.
```
SirCmpwn: Another reason a resource could be inert is if the compositor wants to get rid of it. Since only clients can destroy resources we're kind of SoL

emersion:
Updated to address both comments.

emersion:
Addressed @agx comments, thanks for your review!

CONTRIBUTING.md
```diff
@@ -232,3 +232,125 @@ error_session:
 	return NULL;
 }
 ```
+
+## Wayland protocol implementation
+
+Each protocol generally lives in a file with the same name, containing one
+struct per interface. For instance, `xdg_shell` lives in `types/wlr_xdg_shell.h`
+and has a `wlr_xdg_surface` struct.
```
SirCmpwn: "usually containing at least one struct for each interface in the protocol."

SirCmpwn:
LGTM

emersion:
This PR is ready for review.

Updated all of the core Wayland protocol. I don't want this PR to get too big, I'll do protocol extensions in other PRs.

SirCmpwn:
Added @Timidger and @Ongy cause it might break some of their shit

Timidger:
So I have listeners set up so that the compositor writer doesn't need to destroy any of their data manually. Eg when the output is destroyed it destroys their data and removes the signal listeners.

I haven't read through this in detail, but does wlroots now deal with those simple non-inert cases? Eg I don't need to have most of the destructors I have?

emersion:
>I haven't read through this in detail, but does wlroots now deal with those simple non-inert cases? Eg I don't need to have most of the destructors I have?

wlroots has been automatically destroying resources bound to the display when it's destroyed for quite some time. Some global implementations were missing a destroy signal though. And this patch mostly destroys the resources allocated by the global when it's destroyed.

What you should do: if you need to free some stuff when a wlroots struct is destroyed, listen to the destroy signal. Manually destroying wlroots structs when the display is destroyed is useless, it's done automatically.

Timidger:
Ah ok I have been doing the correct thing the whole time then. Good to know, thanks for CCing me.

PR #935, opened 2018-04-30
wlr_cursor: transform absolute coordinates

### LINKED ISSUE #928: Touch events don't take output output rotation into account
With e.g. 

```
[output:HDMI-A-1]
rotate = 90
```

in rootston touch events are still reported towards the unrotated screen. If there's e.g. a top panel with a button on the left and the screen is rotated 90⁰ one still has to click into the upper left corner of the sreen to activate it. The menu then "correctly" opens on the rotated location.

agx:
Honor output rotation when processing touch coordinates.

I can't test the tablet related parts, touch works as expected now.

Fixes #928

### DIFF
--- a/types/wlr_cursor.c
+++ b/types/wlr_cursor.c
@@ -330,11 +330,75 @@ static void handle_pointer_motion(struct wl_listener *listener, void *data) {
 	wlr_signal_emit_safe(&device->cursor->events.motion, event);
 }
 
+static void apply_output_transform(double *x, double *y,
+		enum wl_output_transform transform) {
+	double dx, dy;
+	double width = 1.0, height = 1.0;
+
+	switch (transform) {
+	case WL_OUTPUT_TRANSFORM_NORMAL:
+		dx = *x;
+		dy = *y;
+		break;
+	case WL_OUTPUT_TRANSFORM_90:
+		dx = *y;
+		dy = width - *x;
+		break;
+	case WL_OUTPUT_TRANSFORM_180:
+		dx = width - *x;
+		dy = height - *y;
+		break;
+	case WL_OUTPUT_TRANSFORM_270:
+		dx = height - *y;
+		dy = *x;
+		break;
+	case WL_OUTPUT_TRANSFORM_FLIPPED:
+		dx = width - *x;
+		dy = *y;
+		break;
+	case WL_OUTPUT_TRANSFORM_FLIPPED_90:
+		dx = height - *y;
+		dy = width - *x;
+		break;
+	case WL_OUTPUT_TRANSFORM_FLIPPED_180:
+		dx = *x;
+		dy = height - *y;
+		break;
+	case WL_OUTPUT_TRANSFORM_FLIPPED_270:
+		dx = *y;
+		dy = *x;
+		break;
+	}
+	*x = dx;
+	*y = dy;
+}
+
+
+static struct wlr_output *get_mapped_output(struct wlr_cursor_device *cursor_device) {
+	if (cursor_device->mapped_output) {
+		return cursor_device->mapped_output;
+	}
+
+	struct wlr_cursor *cursor = cursor_device->cursor;
+	assert(cursor);
+	if (cursor->state->mapped_output) {
+		return cursor->state->mapped_output;
+	}
+	return NULL;
+}
+
+
 static void handle_pointer_motion_absolute(struct wl_listener *listener,
 		void *data) {
 	struct wlr_event_pointer_motion_absolute *event = data;
 	struct wlr_cursor_device *device =
 		wl_container_of(listener, device, motion_absolute);
+
+	struct wlr_output *output =
+		get_mapped_output(device);
+	if (output) {
+		apply_output_transform(&event->x, &event->y, output->transform);
+	}
 	wlr_signal_emit_safe(&device->cursor->events.motion_absolute, event);
 }
 
@@ -362,6 +426,12 @@ static void handle_touch_down(struct wl_listener *listener, void *data) {
 	struct wlr_event_touch_down *event = data;
 	struct wlr_cursor_device *device;
 	device = wl_container_of(listener, device, touch_down);
+
+	struct wlr_output *output =
+		get_mapped_output(device);
+	if (output) {
+		apply_output_transform(&event->x, &event->y, output->transform);
+	}
 	wlr_signal_emit_safe(&device->cursor->events.touch_down, event);
 }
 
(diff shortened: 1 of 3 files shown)


agx:
No idea where the static went. Fixed along with the extra space. 

I was on the fence regarding the asserts as well. The assert would indicate a logic bug which is nice during development. Disabling all asserts would be a matter of passing NDEBUG during "release builds".

agx:
On Mon, Apr 30, 2018 at 06:23:12AM -0700, emersion wrote:
> emersion commented on this pull request.
> 
> 
> 
> > @@ -330,12 +330,65 @@ static void handle_pointer_motion(struct wl_listener *listener, void *data) {
>  	wlr_signal_emit_safe(&device->cursor->events.motion, event);
>  }
>  
> +void apply_output_transform(double *x, double *y,
> 
> Note: we already have a function for this, but it's taking ints :(

That's why I came up with the float one. Having (hopefully) faster int
arithmetic for all the other cases is IMHO fine.

emersion:
>I was on the fence regarding the asserts as well. The assert would indicate a logic bug which is nice during development. Disabling all asserts would be a matter of passing NDEBUG during "release builds".

I mean, what if a touch device isn't mapped to an output? We should only transform if the device is mapped to an output.

agx:
@emersion I had x11 and wayland on my radar from your other comment but then tested incorrectly and thought it wasn't necessary. Fixed now.

agx:
On Mon, Apr 30, 2018 at 07:21:56AM -0700, emersion wrote:
> emersion commented on this pull request.
> 
> 
> 
> > @@ -406,7 +504,18 @@ static void handle_tablet_tool_proximity(struct wl_listener *listener,
>  	struct wlr_event_tablet_tool_proximity *event = data;
>  	struct wlr_cursor_device *device;
>  	device = wl_container_of(listener, device, tablet_tool_proximity);
> -	wlr_signal_emit_safe(&device->cursor->events.tablet_tool_proximity, event);
> +	struct wlr_cursor *cursor = device->cursor;
> +	assert(cursor);
> +	double lx, ly;
> +	wlr_cursor_absolute_to_layout_coords(cursor, event->device,
> +		event->x, event->y, &lx, &ly);
> +	struct wlr_output *output =
> +		wlr_output_layout_output_at(cursor->state->layout, lx, ly);
> 
> (There's a `get_mapping` function somewhere)

I saw that one but wasn't sure if it applies. We don't need a box but
really the output transform. Should I do this via a get_mapped_output():

```c
   if (c_device->mapped_output)
      return c_device->mapped_output;
   else if (cur->state->mapped_output)
      return cur->state->mapped_output
   else
      /* fall back to wlr_cursor_absolut_layout_to_coords */
```

emersion:
>Should I do this via a get_mapped_output()

Yeah, you're right, we need an output here.

agx:
Code now takes mapped outputs into account.

types/wlr_cursor.c
```diff
@@ -330,11 +330,83 @@ static void handle_pointer_motion(struct wl_listener *listener, void *data) {
...
+	}
+	*x = dx;
+	*y = dy;
+}
+
+
+/* get output based on mapped_output or input device coordinates */
+static struct wlr_output *get_output(struct wlr_cursor_device *cursor_device,
+									 struct wlr_input_device *input_device,
+									 double x, double y)
```
emersion: Style: indent with two tabs

types/wlr_cursor.c
```diff
@@ -330,11 +330,83 @@ static void handle_pointer_motion(struct wl_listener *listener, void *data) {
...
+	struct wlr_cursor *cursor = cursor_device->cursor;
+	assert(cursor);
+	if (cursor->state->mapped_output) {
+		return cursor->state->mapped_output;
+	}
+
+	double lx, ly;
+	wlr_cursor_absolute_to_layout_coords(cursor_device->cursor, input_device,
+		x, y, &lx, &ly);
+	return wlr_output_layout_output_at(cursor->state->layout, lx, ly);
```
emersion: It shouldn't try to be smart and guess the output based on coords IMHO.

agx: If I don't do that  I don't end up with an output at all and so no transformation can be done although it is rotated.

agx:
Style issues adressed.

types/wlr_cursor.c
emersion: Hmm, yeah, after thinking about it it's probably fine.

emersion: >If I don't do that I don't end up with an output at all and so no transformation can be done although it is rotated.

Hmm, nah, in the end I really think we shouldn't do that. You should map the touch device to the output, otherwise it'll break support for multiple outputs.

agx: @emersion you mean map configuration wise or code wise?

emersion: That's up to the compositor, so code wise. See https://github.com/swaywm/wlroots/issues/925#issuecomment-385175579 for a list of possible solutions. In rootston that would mean add some configuration.

The behaviour you implemented works with 1 output because the function always return the output. But it really doesn't make sense with two outputs. I just tested by removing the backend-wise mapping in the X11 backend and configuring one screen to be rotated, the cursor jumps when you move it. (If you try with the Wayland backend you need to disable hardware cursors)

agx: Please ignore my ignorance but which code did you exactly disable? If I run this PR (which includes the removal of X11/Wayland specific coordinates) I get the same behaviour as on master.

I know about #925, that's why I suggested to map via wlr_input_device.output_name. But that should probably be done in addition to the current code. IMHO special configuration should not be required for the simple case of a monitor with touch capabilities.

agx:
On Thu, May 03, 2018 at 01:18:54AM -0700, emersion wrote:
>     1. Comment out this line:
>        [2]https://github.com/swaywm/wlroots/blob/master/backend/x11/output.c#L172

That was the missing piece. It all make sense now. Thanks!

agx:
Updated the PR to only use mapped outputs. I wonder if we should  (in rootston) match a touch input device to an output if there is only a single touch device and a single output?

emersion:
>I wonder if we should (in rootston) match a touch input device to an output if there is only a single touch device and a single output?

I'm not against it, but we should be careful to remove the mapping when a second output kicks in. We may also want to implement the heuristics mentioned in #925 (output type, aspect ratio).

agx:
I've opened https://github.com/swaywm/wlroots/issues/940 to so we don't lose the suggestions in a close PR.

SirCmpwn:
Thanks!

PR #936, opened 2018-05-01
output: always use hardware cursors if available

### LINKED ISSUE #281: Full hardware cursors support

emersion:
This changes the `wlr_output_impl.set_cursor` function to take a
`wlr_texture` instead of a byte buffer. This simplifies the
DRM and Wayland backends since they were creating textures from
the byte buffer anyway.

With this commit, performance should be improved when moving the
cursor since outputs don't need to be re-rendered anymore.

Test plan: check that client cursors work on the DRM and Wayland backends.

Fixes #281

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -698,6 +698,27 @@ static void output_cursor_update_visible(struct wlr_output_cursor *cursor) {
 	cursor->visible = visible;
 }
 
+static bool output_cursor_attempt_hardware(struct wlr_output_cursor *cursor) {
+	struct wlr_texture *texture = cursor->texture;
+	if (cursor->surface != NULL) {
+		texture = cursor->surface->texture;
+	}
+
+	struct wlr_output_cursor *hwcur = cursor->output->hardware_cursor;
+	if (cursor->output->impl->set_cursor && (hwcur == NULL || hwcur == cursor)) {
+		if (cursor->output->impl->move_cursor && hwcur != cursor) {
+			cursor->output->impl->move_cursor(cursor->output,
+				(int)cursor->x, (int)cursor->y);
+		}
+		if (cursor->output->impl->set_cursor(cursor->output, texture,
+				cursor->hotspot_x, cursor->hotspot_y, true)) {
+			cursor->output->hardware_cursor = cursor;
+			return true;
+		}
+	}
+	return false;
+}
+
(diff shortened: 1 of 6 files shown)


ascent12:
This adds a hard-dependency on wlr_renderer, which is what we were trying to avoid by taking the pixels directly.

emersion:
>This adds a hard-dependency on wlr_renderer, which is what we were trying to avoid by taking the pixels directly.

I don't get it. Is that bad? The renderer comes from the backend anyway.

Do you prefer to read the pixels from the texture, and then recreate an new texture with the exact same contents on the exact same renderer?

types/wlr_output.c
```diff
@@ -698,6 +698,27 @@ static void output_cursor_update_visible(struct wlr_output_cursor *cursor) {
...
+static bool output_cursor_attempt_hardware(struct wlr_output_cursor *cursor) {
+	struct wlr_texture *texture = cursor->texture;
+	if (cursor->surface != NULL) {
+		texture = cursor->surface->texture;
+	}
+
+	struct wlr_output_cursor *hwcur = cursor->output->hardware_cursor;
+	if (cursor->output->impl->set_cursor && (hwcur == NULL || hwcur == cursor)) {
+		if (cursor->output->impl->move_cursor && hwcur != cursor) {
+			cursor->output->impl->move_cursor(cursor->output,
```
SirCmpwn: Can you assert cursor->output->impl->set_cursor here

emersion: Ah, you mean a backend implementing `set_cursor` must also implement `move_cursor`? Seems like a thing that should be in `wlr_output_init` too.

SirCmpwn: aye

SirCmpwn:
cc'd external integrations as this contains breaking changes

Please send a patch to sway

emersion:
This patch **doesn't** change the compositor API.

It only changes the API for backend implementors.

SirCmpwn:
Which is a public API.

emersion:
... Which they don't use because they don't implement a backend.

SirCmpwn:
But they implement API wrappers. Maybe they have a wrapper around the backend interface.

PR #937, opened 2018-05-02
data-device: reorder functions

emersion:
This is some preliminary work for #931. It just groups functions by type.


emersion:
Updated to split the data device stuff into multiple files.

PR #938, opened 2018-05-02
seat: split into multiple files

emersion:
And remove a useless surface resource destroy listener.

Again, this is some preliminary work for #931.


emersion:
Would you prefer a PR that splits these in different files?

It's pretty difficult to work on these files right now.

SirCmpwn:
Sure.

emersion:
Updated to split into multiple files.

agx:
LGTM

PR #939, opened 2018-05-03
rootston: remove mode and transform signals when output is destroyed

ammen99:
Otherwise rootston will crash when an output is re-plugged.


emersion:
Thanks!

PR #943, opened 2018-05-03
Document env vars used by wlroots/rootston

agx:
Since this influences the inner workings it might make sense to have them documented.

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -1,0 +1,20 @@
+wlroots reads these environment variables
+
+wlroots specific
+----------------
+* *WLR_DRM_DEVICES*: specifies the DRM devices (as a colon separated list) instead of auto probing them. The first existing device in this list is considered the primary DRM device.
+* *WLR_DRM_NO_ATOMIC*: set to 1 to use legacy DRM interface instead of atomic mode setting
+* *WLR_LIBINPUT_NO_DEVICES*: set to 1 to not fail without any input devices
+* *WLR_WL_OUTPUTS*: when using the wayland backend specifies the number of outputs
+* *WLR_X11_OUTPUTS*: when using the X11 backend specifies the number of outputs
+
+rootston specific
+------------------
+* *XKB_DEFAULT_RULES*, *XKB_DEFAULT_MODEL*, *XKB_DEFAULT_LAYOUT*, *XKB_DEFAULT_VARIANT*, *XKB_DEFAULT_OPTIONS*: xkb setup
+
+generic
+-------
+* *DISPLAY*: if set probe X11 backend in *wlr_backend_autocreate*
+* *WAYLAND_DISPLAY*, *_WAYLAND_DISPLAY*, *WAYLAND_SOCKET*: if set probe Wayland backend in *wlr_backend_autocreate*
+* *XCURSOR_PATH*: directory where xcursors are located
+* *XDG_RUNTIME_DIR*: used to store non persistent temporary files


docs/env_vars.md
```diff
@@ -0,0 +1,17 @@
...
+* *WLR_DRM_DEVICES*: specifies the DRM device instead of auto probing it
+* *WLR_DRM_NO_ATOMIC*: set to 1 to use legacy DRM interface instead of atomic mode setting
+* *WLR_LIBINPUT_NO_DEVICES*: set to 1 to not fail without any inpout devices
+* *WLR_WL_OUTPUTS*: when using the wayland backend specifies the number of outputs
+* *WLR_X11_OUTPUTS*: when using the X11 backend specifies the number of outputs
+
+generic
+-------
+* *DISPLAY*: if set probe X11 backend in *wlr_backend_autocreate*
+* *WAYLAND_DISPLAY*, *_WAYLAND_DISPLAY*, *WAYLAND_SOCKET*: if set probe Wayland backend in *wlr_backend_autocreate*
```
emersion: I don't think we support `WAYLAND_SOCKET`? (Yes we should)

docs/env_vars.md
```diff
@@ -0,0 +1,17 @@
...
+* *XCURSOR_PATH*: path to xcursor library
```
emersion: s/library/directory/?

docs/env_vars.md
```diff
@@ -0,0 +1,17 @@
...
+* *XDG_RUNTIME_DIR*: used to store non persistent temporary files
+* *XKB_DEFAULT_RULES*, *XKB_DEFAULT_MODEL*, *XKB_DEFAULT_LAYOUT*, *XKB_DEFAULT_VARIANT*, *XKB_DEFAULT_OPTIONS*: xkb setup
```
emersion: This is used by rootston and the examples, this should be written down (as in: don't expect them to work in any wlroots-based compositor)

docs/env_vars.md
agx: It's locked at in backend/backend.c for autoprobing backends (only if it's set or not)

docs/env_vars.md
agx: Fixed.

docs/env_vars.md
agx: Oh, I misgrepped that one. I've added a rootston specific section (I think we can ignore the examples for now since anybody looking into these is likely to study the sources).

docs/env_vars.md
emersion: Oh, you're right. It's just that `wlr_wl_backend_create` doesn't support being passed a socket.

docs/env_vars.md
```diff
@@ -0,0 +1,21 @@
+wlroots reads these environment variables
+
+wlroots specific
+----------------
+* *WLR_DRM_DEVICES*: specifies the DRM device instead of auto probing it
```
ascent12: It should be specified that this uses colon-separated paths, like PATH.

docs/env_vars.md
```diff
@@ -0,0 +1,21 @@
...
+* *WLR_DRM_NO_ATOMIC*: set to 1 to use legacy DRM interface instead of atomic mode setting
```
ascent12: Technically it's if it's defined at all, not if it's set to 1.

docs/env_vars.md
ascent12: Also, the first one specified is the "primary" GPU.

docs/env_vars.md
```diff
@@ -0,0 +1,21 @@
...
+* *WLR_LIBINPUT_NO_DEVICES*: set to 1 to not fail without any inpout devices
```
SirCmpwn: >inpout

docs/env_vars.md
SirCmpwn: code should be fixed

agx:
Updated the review comments and opened https://github.com/swaywm/wlroots/pull/945 to make WLR_DRM_NO_ATOMIC docs true.

emersion:
Thanks!

PR #945, opened 2018-05-03
drm: only disable atomic modesetting when WLR_DRM_NO_ATOMIC == 1

agx:
Don't allow any value of the env var to disable atomic mode setting.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -32,7 +32,8 @@ bool check_drm_features(struct wlr_drm_backend *drm) {
 		return false;
 	}
 
-	if (getenv("WLR_DRM_NO_ATOMIC")) {
+	const char *no_atomic = getenv("WLR_DRM_NO_ATOMIC");
+	if (no_atomic && strcmp(no_atomic, "1") == 0) {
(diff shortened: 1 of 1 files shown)


backend/drm/drm.c
```diff
@@ -32,7 +32,13 @@ bool check_drm_features(struct wlr_drm_backend *drm) {
...
 	}
 
-	if (getenv("WLR_DRM_NO_ATOMIC")) {
+	const char *no_atomic = getenv("WLR_DRM_NO_ATOMIC");
+	if (no_atomic) {
+		if (strcmp(no_atomic, "1") != 0) {
+			no_atomic = NULL;
+		}
+	}
+	if (no_atomic) {
```
emersion: This code logic is pretty weird, I'd just

```c
if (no_atomic && strcmp(no_atomic, "1") == 0) {
```

agx:
On Thu, May 03, 2018 at 06:35:23AM -0700, emersion wrote:
>  if (no_atomic && strcmp(no_atomic, "1") == 0) {

Updated. Thanks for the hint.

emersion:
Thanks!

PR #946, opened 2018-05-03
Fix gcc maybe-uninitialized in wlr_cursor apply_output_transform

Hello71:
gcc 8.1.0: types/wlr_cursor.c: error: 'dx', 'dy' may be used uninitialized in this function

maybe this should log an error, but I don't know how you do "log this error once".

### DIFF
--- a/types/wlr_cursor.c
+++ b/types/wlr_cursor.c
@@ -368,6 +368,10 @@ static void apply_output_transform(double *x, double *y,
 		dx = *y;
 		dy = *x;
 		break;
+	default:
+		// displaying cursor wrong is better than crashing
+		dx = *x;
+		dy = *y;
 	}
 	*x = dx;
 	*y = dy;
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Fucking gcc

Just initialize them before the switch, don't add a default.

SirCmpwn:
Have we reported this bug to gcc yet?

Hello71:
according to https://stackoverflow.com/questions/45409452/is-gccs-wmaybe-uninitialized-generating-a-spurious-warning-with-o1-for-a-swit, an enum may have values outside of the declared ones. I don't totally understand the C standard's wording here, but it appears to agree with this theory.

Hello71:
ugh... can't fixup on the github editor?

emersion:
Please try to compile before committing.

```c
../types/wlr_cursor.c: In function 'apply_output_transform':
../types/wlr_cursor.c:338:2: error: enumeration value 'WL_OUTPUT_TRANSFORM_NORMAL' not handled in switch [-Werror=switch]
  switch (transform) {
  ^~~~~~
```

SirCmpwn:
Don't use the github editor to author commits.

Hello71:
eh, I just turned off Werror. you guys can figure it out, it'll be faster than me trying to figure out your coding standards

PR #951, opened 2018-05-04
Fix gcc maybe-uninitialized in wlr_cursor apply_output_transform

chebykinn:
Fix this weird gcc warning, same as #946, but it was closed.


emersion:
Thanks!

PR #957, opened 2018-05-05
output-damage: limit the number of damaged rectangles

### LINKED ISSUE #949: Don't try to render damage from too many rectangles
Drawing damage from too many rectangles is not efficient. We should fallback to rendering the bounds of the damaged region when the region is too complicated.

This is how how GTK+ does it: https://streamable.com/rb6rr

Proposed plan: add a `uint32_t max_rects` field to `wlr_output_damage`. This is set to a reasonable value by default (something between 20 and 30?), and is override-able by the compositor. If the number of rectangles is too high, it'll just return the bounds of the damaged region. Since `wlr_output_damage` is optional anyway, I don't think this will do any harm. Compositors that wish to disable this feature can set it to MAX_INT.

emersion:
Test plan: set the limit to 1, open htop on gnome-terminal.

The current value is kind of arbitrary, it just roughly reflects what GTK+ does.

Fixes #949

### DIFF
--- a/types/wlr_output_damage.c
+++ b/types/wlr_output_damage.c
@@ -58,6 +58,7 @@ struct wlr_output_damage *wlr_output_damage_create(struct wlr_output *output) {
 	}
 
 	output_damage->output = output;
+	output_damage->max_rects = 20;
 	wl_signal_init(&output_damage->events.frame);
 	wl_signal_init(&output_damage->events.destroy);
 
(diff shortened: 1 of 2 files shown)


types/wlr_output_damage.c
```diff
@@ -58,6 +58,7 @@ struct wlr_output_damage *wlr_output_damage_create(struct wlr_output *output) {
 	}
 
 	output_damage->output = output;
+	output_damage->max_rects = 20;
```
agx: I'd avoid magic constants in the code:
```
output_damage->max_rects = WLR_OUTPUT_DAMAGE_MAX_RECTS;
```
and then define it in a wlr_output.h.

emersion: Well, it's not a constant, it's a default value. Does it make sense to add something like `WLR_OUTPUT_DAMAGE_DEFAULT_MAX_RECTS`?

agx: Sure it's a default but that's a compile time constant. I'd do the #define nevertheless since it's more descriptive and can be reused at other locations if needed.

PR #960, opened 2018-05-07
tablet-unstable-v2 support

Ongy:
I'm losing motivation to work on this on my own, so I'm opening the PR, for a bit of feedback and pressure :)

This implements the tablet-unstable-v2 protocol from wayland-protocols for wlroots, and uses it in `rootston`.

The most notable changes:
* `wlr_tablet_*` has been expanded to contain any information required for tablet-v2
* the libinput backend builds the new table objects completly

**Edit:** Cleaned this part a bit, and clarified what I mean.

Planed for this PR:
* ~Inert resources/resource destruction~
* General cleaning sweep over the changes
* ~all axis events~
* ~Clean up serial handling~

Follow up issues:
* wayland backend client implementation
* (Maybe) give tablet a dedicated cursor/focus
* ~Use relative movements for Mouse (and Lens?) tools [by default]~
* Implement grabs (and the send/notify duality)


backend/libinput/tablet_tool.c
```diff
@@ -1,25 +1,185 @@
...
+	}
+}
+
 struct wlr_tablet_tool *create_libinput_tablet_tool(
 		struct libinput_device *libinput_dev) {
 	assert(libinput_dev);
-	struct wlr_tablet_tool *wlr_tablet_tool = calloc(1, sizeof(struct wlr_tablet_tool));
+	struct wlr_libinput_tablet *libinput_tablet_tool =
+		calloc(1, sizeof(struct wlr_libinput_tablet));
+	struct wlr_tablet_tool *wlr_tablet_tool = &libinput_tablet_tool->wlr_tool;
```
emersion: NULL check should be before this line

include/wlr/types/wlr_tablet_tool.h
```diff
@@ -5,8 +5,48 @@
...
+	WLR_TABLET_TOOL_TYPE_ERASER,	/**< Eraser */
+	WLR_TABLET_TOOL_TYPE_BRUSH,	/**< A paintbrush-like tool */
+	WLR_TABLET_TOOL_TYPE_PENCIL,	/**< Physical drawing tool, e.g.
+				             Wacom Inking Pen */
+	WLR_TABLET_TOOL_TYPE_AIRBRUSH,	/**< An airbrush-like tool */
+	WLR_TABLET_TOOL_TYPE_MOUSE,	/**< A mouse bound to the tablet */
+	WLR_TABLET_TOOL_TYPE_LENS,	/**< A mouse tool with a lens */
+};
+
+struct wlr_tablet_tool_tool {
```
emersion: I'm still not liking this name

include/wlr/types/wlr_tablet_pad.h
```diff
@@ -20,16 +20,40 @@ struct wlr_tablet_pad {
...
 		struct wl_signal strip;
+		struct wl_signal attach_tablet; //struct wlr_tablet_tool
 	} events;
 
+	size_t button_count;
+	size_t ring_count;
+	size_t strip_count;
+
+	struct wl_list groups; // wlr_tablet_pad_group::link
+	struct wl_list paths; // wlr_tablet_path::link
```
emersion: Is it possible to use a regular array for paths? (`char **paths` + `size_t paths_len`)

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,130 @@
+#ifndef WLR_TYPES_WLR_TABLET_V2_H
```
emersion: We don't have a consistent naming scheme for unstable protocol implementations.
1. Some implementations are missing the version number, we should fix that.
2. We're stripping the "z" prefix. That means that if a xdg-shell stable v2 protocol is publised, it will conflict with unstable versions.

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,130 @@
...
+
+struct wlr_tablet_v2_tablet_tool *wlr_make_tablet_tool(
+		struct wlr_tablet_manager_v2 *manager,
+		struct wlr_seat *wlr_seat,
+		struct wlr_tablet_tool_tool *wlr_tool);
+
+struct wlr_tablet_manager_v2 *wlr_tablet_v2_create(struct wl_display *display);
+void wlr_tablet_v2_destroy(struct wlr_tablet_manager_v2 *manager);
+
+uint32_t wlr_send_tablet_v2_tablet_tool_proximity_in(
```
emersion: It seems that grabs are not implemented. This should be added to the list of follow-up issues.

types/wlr_tablet_v2.c
```diff
@@ -0,0 +1,1291 @@
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+#endif
+
+#include <wlr/types/wlr_tablet_v2.h>
```
emersion: Style nitpick: group all these include together, without blank newlines.

types/wlr_tablet_v2.c
```diff
@@ -0,0 +1,1291 @@
...
+	wl_list_remove(&tablet->tool_destroy.link);
+	free(tablet);
+}
+
+static void handle_tablet_tool_v2_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+static struct zwp_tablet_tool_v2_interface tablet_tool_impl = {
+	.set_cursor = NULL,
```
emersion: This should be implemented

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,130 @@
...
+
+	size_t group_count;
+	uint32_t *groups;
+
+	struct wl_listener pad_destroy;
+
+	struct wlr_tablet_pad_client_v2 *current_client;
+};
+
+struct wlr_tablet_v2_tablet *wlr_make_tablet(
```
emersion: We generally call these `<type>_create`

backend/libinput/events.c
```diff
@@ -186,6 +186,9 @@ static void handle_device_removed(struct wlr_libinput_backend *backend,
 	}
 	struct wlr_input_device *dev, *tmp_dev;
 	wl_list_for_each_safe(dev, tmp_dev, wlr_devices, link) {
+		if (dev->type == WLR_INPUT_DEVICE_TABLET_TOOL) {
+			wlr_libinput_tablet_tool_destroy(dev);
```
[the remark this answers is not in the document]
Ongy: Ah, I think this should be done over the implementation destructor, it's a hack here.
I missed that way to do it, when I wrote the code (and some backend functions still used `wlr_` prefix, so I had used it here), thanks for the pointer.

include/wlr/types/wlr_tablet_tool.h
Ongy: Neither do I. I'd still prefer a rename to the way tablet-v2 handles it, or good suggestions in general :)

include/wlr/types/wlr_tablet_pad.h
Ongy: Probably I'd say. It's nice/easy on libinput, semi-nice on wayland backend (have to aggregate).
I'd prefer the old `list_t` (I think `wlr_list` now) over blank array.

Ongy:
Thanks for the initial feedback :)

I just remembered what I was stuck on earlier:
How are serials managed in wlroots? And who has ownership in general? rootston or wlroots?
tablet-v2 has a somewhat ugly situation where we are supposed to send button events for all held buttons on `proximity_in`, but each of them requires a serial.
Should this be pushed into rootston so the compositor can keep track of serials, or contain it in the implementation (since it's forced), and then store the serials in the `wlr_` structs that are exposed?

emersion:
>(Re-establish) pointer fallback semantics (currently tablet support is required for input)

How does this work in practice? Do clients like Xwayland and GTK+ have a proper tablet protocol support?

Ongy:
`strings /bin/Xwayland | grep tablet` shows nothing. So at least current arch's Xwayland doesn't support it.

gnome-terminal supports it. There's some corner cases that don't work properly, but that's issues on the GTK side of things.

Any other client I should test?

emersion:
>strings /bin/Xwayland | grep tablet shows nothing. So at least current arch's Xwayland doesn't support it.

My Xwayland seems to support it.

Ongy:
`s/current/last week's/`
ok, current Xwayland supports tablet. Krita "worked", I think there might be a bug with the pressure axis in my code (or libinput backend), I'll investigate that

backend/libinput/tablet_tool.c
```diff
@@ -1,25 +1,179 @@
...
+
+static void destroy_tool_tool(struct wlr_libinput_tablet_tool *tool) {
+	wlr_signal_emit_safe(&tool->wlr_tool.events.destroy, &tool->wlr_tool);
+	libinput_tablet_tool_ref(tool->libinput_tool);
+	libinput_tablet_tool_set_user_data(tool->libinput_tool, NULL);
+	free(tool);
+}
+
+
+static void libinput_tablet_tool_destroy(struct wlr_tablet_tool *tool) {
```
emersion: The `libinput_` prefix should be reserved for libinput.

backend/libinput/tablet_tool.c
```diff
@@ -1,25 +1,179 @@
...
+	wl_signal_init(&ret->wlr_tool.events.destroy);
+
+	libinput_tablet_tool_set_user_data(tool, ret);
+	return ret;
+}
+
+static void ensure_tool_reference(struct wlr_libinput_tablet_tool *tool,
+		struct wlr_tablet_tool *wlr_dev) {
+	struct tablet_tool_list_elem *pos;
+	struct wlr_libinput_tablet *tablet = wl_container_of(wlr_dev, tablet, wlr_tool);
```
emersion: Is it possible to add assertions to all of these `wl_container_of`?

Ongy: What kind of assertions? Assert non-NULL, or should we have a better way to check for valid arguments?

emersion: No need to assert non-NULL. I was thinking about checking that the `wlr_dev` comes from the libinput backend.

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,172 @@
...
+	struct wl_list link; // wlr_tablet_seat_v2::tablets
+	struct wlr_tablet_tool_tool *wlr_tool;
+	struct wl_list clients; // wlr_tablet_tool_client_v2::tablet_link
+
+	struct wl_listener tool_destroy;
+
+	struct wlr_tablet_tool_client_v2 *current_client;
+	struct wlr_surface *focused_surface;
+	struct wl_listener surface_destroy;
+	struct wl_listener client_destroy;
```
emersion: There are a lot of different clients in here. I think this should be renamed to `seat_client_destroy` for clarity.

types/wlr_tablet_tool_tool.c
```diff
@@ -0,0 +1,16 @@
...
+#include <string.h>
+#include <wayland-server.h>
+#include <wlr/interfaces/wlr_tablet_tool.h>
+#include <wlr/types/wlr_tablet_tool.h>
+
+void wlr_tablet_tool_tool_init(struct wlr_tablet_tool_tool *tool) {
+	// Intentionaly empty (for now)
+}
+
+void wlr_tablet_tool_tool_destroy(struct wlr_tablet_tool_tool *tool) {
```
emersion: Should be in a header file, or removed.

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+	wl_list_remove(&seat->seat_destroy.link);
+
+	struct wlr_tablet_seat_client_v2 *client;
+	struct wlr_tablet_seat_client_v2 *tmp;
+	wl_list_for_each_safe(client, tmp, &seat->clients, seat_link) {
+		wlr_tablet_seat_client_v2_destroy(client->resource);
+	}
+}
+
+static struct wlr_tablet_seat_v2 *create_tablet_seat(
```
emersion: Should be named `create_tablet_seat_v2`

SirCmpwn:
Did not exhaustively review the code but the approach looks good to me and I tested it with krita, seems to work.

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+	wl_list_init(&tablet_seat->pads);
+
+	tablet_seat->seat_destroy.notify = handle_wlr_seat_destroy;
+	wl_signal_add(&wlr_seat->events.destroy, &tablet_seat->seat_destroy);
+
+	wl_list_insert(&manager->seats, &tablet_seat->link);
+	return tablet_seat;
+}
+
+struct wlr_tablet_seat_v2 *get_or_create_tablet_seat(
```
emersion: Should be `static`

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+
+	struct wlr_tablet_seat_client_v2 *seat_client =
+		calloc(1, sizeof(struct wlr_tablet_seat_client_v2));
+	if (tablet_seat == NULL) {
+		wl_client_post_no_memory(wl_client);
+		return;
+	}
+
+	seat_client->resource =
+		wl_resource_create(wl_client, &zwp_tablet_seat_v2_interface, 1, id);
```
emersion: This should use the manager's version

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+	wl_list_init(&seat_client->tablets);
+	wl_list_init(&seat_client->pads);
+
+	seat_client->seat_client_destroy.notify = handle_seat_client_destroy;
+	wl_signal_add(&seat->events.destroy, &seat_client->seat_client_destroy);
+
+	wl_list_insert(&manager->tablet_seats, &seat_client->client_link);
+	wl_list_insert(&tablet_seat->clients, &seat_client->seat_link);
+
+	// We need to emmit the devices allready on the seat
```
emersion: Typo: emit

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+		 */
+		wl_resource_set_implementation(seat_resource, &seat_impl, NULL,
+			wlr_tablet_seat_client_v2_destroy);
+		return;
+	}
+	struct wlr_seat_client *seat = wlr_seat_client_from_resource(seat_resource);
+	struct wlr_tablet_seat_v2 *tablet_seat =
+		get_or_create_tablet_seat(manager->manager, seat->seat);
+
+	if (!tablet_seat) {// This can only happen when we ran out of memory
```
emersion: Style: space between `{` and `//`

include/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,93 @@
...
+struct wlr_tablet_seat_client_v2 {
+	struct wl_list seat_link;
+	struct wl_list client_link;
+	struct wl_client *wl_client;
+	struct wl_resource *resource;
+
+	struct wlr_tablet_manager_client_v2 *client;
+	struct wlr_seat_client *seat;
+
+	struct wl_listener seat_client_destroy;
```
emersion: Should be removed

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+#include <wlr/util/log.h>
+#include "tablet-unstable-v2-protocol.h"
+
+struct wlr_tablet_manager_client_v2 {
+	struct wl_list link;
+	struct wl_client *client;
+	struct wl_resource *resource;
+	struct wlr_tablet_manager_v2 *manager;
+
+	struct wl_listener client_destroy;
```
emersion: Should be removed

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	struct wlr_tablet_seat_client_v2 *pos;
+	struct wlr_tablet_seat_client_v2 *tmp;
+	wl_list_for_each_safe(pos, tmp, &client->tablet_seats, client_link) {
+		wlr_tablet_seat_client_v2_destroy(pos->resource);
+	}
+
+	wl_list_remove(&client->link);
+	//wl_list_remove(&client->client_destroy.link);
```
emersion: Should be removed

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	wl_list_init(&tablet->clients);
+	wl_list_init(&tablet->seats);
+
+	tablet->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &tablet->display_destroy);
+
+	tablet->wl_global = wl_global_create(display,
+		&zwp_tablet_manager_v2_interface, 1, tablet, tablet_v2_bind);
```
emersion: Can you define `TABLET_MANAGER_VERSION` at the beginning of the file, so that it's obvious which version we implement?

types/tablet_v2/wlr_tablet_v2.c
Ongy: It's used by the other files I split the implementation into, but shouldn't be API.
What's the proper naming scheme for them? Afaik it shouldn't have the `wlr_` prefix.

SirCmpwn:
How're we looking here guys

emersion:
This is a massive PR, so I don't think I'll be able to review it very carefully. Give me time to check the rest of the protocol implementation though. I think we should just check that it works properly after that.

I'm still not happy with `tool_tool` though.

types/tablet_v2/wlr_tablet_v2_tablet.c
```diff
@@ -0,0 +1,127 @@
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+#endif
+
+#include "tablet-unstable-v2-protocol.h"
```
emersion: Local includes should be at the bottom.

types/tablet_v2/wlr_tablet_v2_tablet.c
```diff
@@ -0,0 +1,127 @@
...
+void add_tablet_client(struct wlr_tablet_seat_client_v2 *seat,
+		struct wlr_tablet_v2_tablet *tablet) {
+	struct wlr_tablet_client_v2 *client =
+		calloc(1, sizeof(struct wlr_tablet_client_v2));
+	if (!client) {
+		return;
+	}
+
+	client->resource =
+		wl_resource_create(seat->wl_client, &zwp_tablet_v2_interface, 1, 0);
```
emersion: The version should be retrieved from the tablet client's resource.

types/tablet_v2/wlr_tablet_v2_tablet.c
```diff
@@ -0,0 +1,127 @@
...
+	if (!client->resource) {
+		free(client);
+		return;
```
emersion: Should `wl_resource_post_no_memory()` here

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,564 @@
...
+	struct tablet_pad_auxiliary_user_data *aux = wl_resource_get_user_data(resource);
+	if (!aux) {
+		return;
+	}
+
+	struct wlr_tablet_v2_event_feedback evt = {
+		.serial = serial,
+		.description = description,
+		.index = aux->index
+		};
```
emersion: Style: extra tab here

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,564 @@
...
+
+static struct zwp_tablet_pad_group_v2_interface tablet_pad_group_impl = {
+	.destroy = handle_tablet_pad_group_v2_destroy,
+};
+
+static void add_tablet_pad_group(struct wlr_tablet_v2_tablet_pad *pad,
+		struct wlr_tablet_pad_client_v2 *client,
+		struct wlr_tablet_pad_group *group, size_t index) {
+	client->groups[index] =
+		wl_resource_create(client->client, &zwp_tablet_pad_group_v2_interface, 1, 0);
```
emersion: Again, take version from client

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,564 @@
...
+	if (!client->groups[index]) {
+		wl_client_post_no_memory(client->client);
+		return;
+	}
+	struct tablet_pad_auxiliary_user_data *user_data =
+		calloc(1, sizeof(struct tablet_pad_auxiliary_user_data));
+	if (!user_data) {
+		return;
```
emersion: Again, post no memory

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,564 @@
...
+		for (size_t i = 0; i < pos->strip_count; ++i) {
+			destroy_tablet_pad_strip_v2(pos->strips[i]);
+		}
+
+		for (size_t i = 0; i < pos->ring_count; ++i) {
+			destroy_tablet_pad_ring_v2(pos->rings[i]);
+		}
+	}
+
+	wl_list_remove(&pad->clients);
```
emersion: Tablet pads don't get destroyed?

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,564 @@
...
+	 * proximity_in. Someone wants to do the math how long that would take?
+	 */
+	uint32_t serial = wl_display_next_serial(wl_client_get_display(client));
+
+	zwp_tablet_pad_v2_send_enter(pad_client->resource, serial,
+		tablet_client->resource, surface->resource);
+
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	uint32_t time = now.tv_nsec / 1000;
```
emersion: Can't we get the timestamp from libinput?

Ongy: Currently this is bound to the seat keyboard focus. If that's changed, or we require the time to be passed to that as well then probably? But focus change there can also happen from a client unmap, which doesn't have the timestamp from libinput.

emersion: Yeah, probably better to keep it that way.

types/tablet_v2/wlr_tablet_v2_pad.c
emersion: Bump

Ongy: this seems to be a weird display bug on the github side? It's fixed in my local checkout and using the web-ui I can see the change when I only view changes made in my latest commit, but not on the current state of the file

types/tablet_v2/wlr_tablet_v2_tablet.c
```diff
@@ -0,0 +1,131 @@
...
+
+void add_tablet_client(struct wlr_tablet_seat_client_v2 *seat,
+		struct wlr_tablet_v2_tablet *tablet) {
+	struct wlr_tablet_client_v2 *client =
+		calloc(1, sizeof(struct wlr_tablet_client_v2));
+	if (!client) {
+		return;
+	}
+
+	int version = wl_resource_get_version(seat->resource);
```
emersion: This should be an `uint32_t`

rootston/seat.c
```diff
@@ -5,16 +5,20 @@
...
 #include <wlr/types/wlr_xcursor_manager.h>
+#include <wlr/types/wlr_tablet_v2.h>
 #include <wlr/util/log.h>
 #include "rootston/cursor.h"
 #include "rootston/input.h"
 #include "rootston/keyboard.h"
 #include "rootston/seat.h"
 #include "rootston/xcursor.h"
 
+#include <libinput.h>
```
emersion: Style: this should grouped with other external headers, sorted alphabetically.

rootston/seat.c
```diff
@@ -97,13 +101,96 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
...
+	 * We need to handle them ourselves, not pass it into the cursor
+	 * without any consideration
+	 */
+	handle_tablet_tool_position(cursor, event->device->data, event->tool,
+		event->updated_axes & WLR_TABLET_TOOL_AXIS_X,
+		event->updated_axes & WLR_TABLET_TOOL_AXIS_Y,
+		event->x, event->y, event->dx, event->dy);
+
+	if (event->updated_axes & WLR_TABLET_TOOL_AXIS_PRESSURE) {
+		wlr_send_tablet_v2_tablet_tool_pressure(roots_tool->tablet_v2_tool, event->pressure);
```
emersion: Style: 80 char limit

rootston/seat.c
```diff
@@ -112,7 +199,86 @@ static void handle_tool_tip(struct wl_listener *listener, void *data) {
...
+	struct wlr_tablet_v2_event_cursor *evt = data;
+
+
+	struct wlr_seat_pointer_request_set_cursor_event event = {
+		.surface = evt->surface,
+		.hotspot_x = evt->hotspot_x,
+		.hotspot_y = evt->hotspot_y,
+		.serial = evt->serial,
+		.seat_client = evt->seat_client,
+		};
```
emersion: Style: is this GitHub's fault again?

rootston/seat.c
```diff
@@ -112,7 +199,86 @@ static void handle_tool_tip(struct wl_listener *listener, void *data) {
...
+		roots_tool->seat = cursor->seat;
+		tool->data = roots_tool;
+		roots_tool->tablet_v2_tool =
+			wlr_tablet_tool_create(desktop->tablet_v2,
+				cursor->seat->seat, tool);
+		roots_tool->tool_destroy.notify = handle_tablet_tool_tool_destroy;
+		wl_signal_add(&tool->events.destroy, &roots_tool->tool_destroy);
+
+		roots_tool->set_cursor.notify = handle_tablet_tool_set_cursor;
+		wl_signal_add(&roots_tool->tablet_v2_tool->events.set_cursor, &roots_tool->set_cursor);
```
emersion: Style: 80 char limit

rootston/seat.c
```diff
@@ -819,6 +1158,13 @@ void roots_seat_set_focus(struct roots_seat *seat, struct roots_view *view) {
 		wlr_seat_keyboard_notify_enter(seat->seat, view->wlr_surface,
 			keyboard->keycodes, keyboard->num_keycodes,
 			&keyboard->modifiers);
+		/* FIXME: Move this to a better place */
+		struct roots_tablet_pad *pad;
```
emersion: Yeah, this definitely shouldn't be here. Probably near `wlr_pointer_send_enter`.

rootston/seat.c
```diff
@@ -532,6 +849,7 @@ static void handle_tablet_tool_destroy(struct wl_listener *listener,
 	wlr_cursor_detach_input_device(seat->cursor->cursor, tablet_tool->device);
 	wl_list_remove(&tablet_tool->device_destroy.link);
 	wl_list_remove(&tablet_tool->link);
+	wl_list_remove(&tablet_tool->pads);
```
emersion: This doesn't do anything useful. You should iterate the list and destroy each element if you want to free it.

rootston/seat.c
```diff
@@ -518,9 +691,153 @@ static void seat_add_touch(struct roots_seat *seat,
...
+	wlr_send_tablet_v2_tablet_pad_button(pad->tablet_v2_pad,
+		event->button, event->time_msec,
+		(enum zwp_tablet_pad_v2_button_state)event->state);
+}
+
 static void seat_add_tablet_pad(struct roots_seat *seat,
 		struct wlr_input_device *device) {
-	// TODO
+	struct roots_tablet_pad *tablet_pad =
+		calloc(sizeof(struct roots_tablet_pad), 1);
```
emersion: Nitpick: doesn't matter a lot, but strictly speaking, the order of the arguments are swapped.

rootston/seat.c
```diff
@@ -518,9 +691,153 @@ static void seat_add_touch(struct roots_seat *seat,
...
+	tablet_pad->device = device;
+	tablet_pad->seat = seat;
+	wl_list_init(&tablet_pad->tablet_link);
+	wl_list_insert(&seat->tablet_pads, &tablet_pad->link);
+
+	tablet_pad->device_destroy.notify = handle_tablet_pad_destroy;
+	wl_signal_add(&tablet_pad->device->events.destroy,
+		&tablet_pad->device_destroy);
+
+	tablet_pad->tablet_destroy.notify = handle_pad_tool_destroy;
```
emersion: We prefer not to set handlers without `wl_signal_add` - and always set handlers when we use `wl_signal_add`. This makes it easier to understand which function is called when using `wl_signal_add`.

rootston/seat.c
Ongy: This is a list of roots_tablet_pads, they get freed on their own.
This list is intended for managing focus when it's decoupled from the usual seat, though I discarded that, at least for this PR, since it's large enough.

It's currently unused, though set up. Should I remove it?

rootston/seat.c
Ongy: Those are tablet pads that are set here, they are closer to keyboards than pointers (tools are more pointer-ish).

In theory, this should probably be decoupled from the seat, but this PR is getting large enough, which is why I didn't do that for now and kept this working situation.

emersion: Oh, right. Yeah, that's fine then.

rootston/seat.c
emersion: Yeah, I think so. Also, never `wl_list_remove` a head of a list.

emersion:
What about the `tool_tool` rename?

rootston/seat.c
Ongy: Why would I not remove it? This makes sure there's no use-after-free into the head when the other parts are removed. The "list" becomes useless, but everything stays in good cases

rootston/seat.c
```diff
@@ -97,13 +101,102 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
...
 static void handle_tool_axis(struct wl_listener *listener, void *data) {
 	struct roots_cursor *cursor =
 		wl_container_of(listener, cursor, tool_axis);
 	struct roots_desktop *desktop = cursor->seat->input->server->desktop;
 	wlr_idle_notify_activity(desktop->idle, cursor->seat->seat);
 	struct wlr_event_tablet_tool_axis *event = data;
-	roots_cursor_handle_tool_axis(cursor, event);
+	struct roots_tablet_tool_tool *roots_tool = event->tool->data;
+
+	if (!roots_tool) { // Should this be an assert?
```
emersion: Yes.

rootston/seat.c
```diff
@@ -112,7 +205,87 @@ static void handle_tool_tip(struct wl_listener *listener, void *data) {
...
+	struct roots_tablet_tool_tool *roots_tool = event->tool->data;
+
+	if (event->state == WLR_TABLET_TOOL_TIP_DOWN) {
+		wlr_send_tablet_v2_tablet_tool_down(roots_tool->tablet_v2_tool);
+	} else {
+		wlr_send_tablet_v2_tablet_tool_up(roots_tool->tablet_v2_tool);
+	}
+}
+
+static void handle_tablet_tool_tool_destroy(struct wl_listener *listener, void *data) {
```
emersion: `tool_tool`?

SirCmpwn:
Can you rebase this? Let's get it merged.

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+
+void wlr_send_tablet_v2_tablet_tool_proximity_in(
+	struct wlr_tablet_v2_tablet_tool *tool,
+	struct wlr_tablet_v2_tablet *tablet,
+	struct wlr_surface *surface);
+
+void wlr_send_tablet_v2_tablet_tool_down(struct wlr_tablet_v2_tablet_tool *tool);
+void wlr_send_tablet_v2_tablet_tool_up(struct wlr_tablet_v2_tablet_tool *tool);
+
+void wlr_send_tablet_v2_tablet_tool_motion(
```
acrisci: The naming convention we use for these things is more like `wlr_tablet_v2_tablet_tool_send_motion()`

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+	struct wlr_tablet_v2_tablet_tool *tool, double x, double y);
+
+void wlr_send_tablet_v2_tablet_tool_rotation(
+	struct wlr_tablet_v2_tablet_tool *tool, double degrees);
+
+void wlr_send_tablet_v2_tablet_tool_slider(
+	struct wlr_tablet_v2_tablet_tool *tool, double position);
+
+void wlr_send_tablet_v2_tablet_tool_wheel(
+	struct wlr_tablet_v2_tablet_tool *tool, double delta, int32_t clicks);
```
acrisci: Would prefer `delta_degrees` because it could be radians. Also, we need to document what the angle is relative to (protocol seems to indicate "clockwise from neutral")

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+#include <wayland-server.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_input_device.h>
+
+#include "tablet-unstable-v2-protocol.h"
+
+/* This can probably be even lower,the tools don't have a lot of buttons */
+#define WLR_TABLET_V2_TOOL_BUTTONS_CAP 16
+
+struct wlr_tablet_client_v2;
```
acrisci: Ugh, we're all over the place with naming the versions in the protocols. At least they'll go away when they go stable. Do we want this or `wlr_tablet_v2_client`?

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+	struct wlr_surface *surface);
+
+void wlr_send_tablet_v2_tablet_tool_down(struct wlr_tablet_v2_tablet_tool *tool);
+void wlr_send_tablet_v2_tablet_tool_up(struct wlr_tablet_v2_tablet_tool *tool);
+
+void wlr_send_tablet_v2_tablet_tool_motion(
+	struct wlr_tablet_v2_tablet_tool *tool, double x, double y);
+
+void wlr_send_tablet_v2_tablet_tool_pressure(
+	struct wlr_tablet_v2_tablet_tool *tool, double pressure);
```
acrisci: The protocol says `pressure` is a `uint`.

acrisci:
Just tried to open a `gnome-terminal` with this branch and got:

```
==6673==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000300b70 at pc 0x7f5e429347c8 bp 0x7fff748bd0f0 sp 0x7fff748bd0e0
WRITE of size 8 at 0x602000300b70 thread T0
    #0 0x7f5e429347c7 in add_tablet_pad_group ../types/tablet_v2/wlr_tablet_v2_pad.c:190
    #1 0x7f5e42935da3 in add_tablet_pad_client ../types/tablet_v2/wlr_tablet_v2_pad.c:315
    #2 0x7f5e42932c3c in get_tablet_seat ../types/tablet_v2/wlr_tablet_v2.c:202
    #3 0x7f5e3e51203d in ffi_call_unix64 (/usr/lib64/libffi.so.6+0x603d)
    #4 0x7f5e3e5119fe in ffi_call (/usr/lib64/libffi.so.6+0x59fe)
    #5 0x7f5e42635f2c  (/usr/lib64/libwayland-server.so.0+0xbf2c)
    #6 0x7f5e426323de  (/usr/lib64/libwayland-server.so.0+0x83de)
    #7 0x7f5e42633f01 in wl_event_loop_dispatch (/usr/lib64/libwayland-server.so.0+0x9f01)
    #8 0x7f5e42632601 in wl_display_run (/usr/lib64/libwayland-server.so.0+0x8601)
    #9 0x41e85c in main ../rootston/main.c:74
    #10 0x7f5e417d518a in __libc_start_main (/usr/lib64/libc.so.6+0x2318a)
    #11 0x406a99 in _start (/home/tony/projects/wlroots/build/rootston/rootston+0x406a99)

0x602000300b74 is located 0 bytes to the right of 4-byte region [0x602000300b70,0x602000300b74)
allocated by thread T0 here:
    #0 0x7f5e42ca2e50 in calloc (/usr/lib64/libasan.so.5+0xeee50)
    #1 0x7f5e42935603 in add_tablet_pad_client ../types/tablet_v2/wlr_tablet_v2_pad.c:265
    #2 0x7f5e42932c3c in get_tablet_seat ../types/tablet_v2/wlr_tablet_v2.c:202
    #3 0x7f5e3e51203d in ffi_call_unix64 (/usr/lib64/libffi.so.6+0x603d)

SUMMARY: AddressSanitizer: heap-buffer-overflow ../types/tablet_v2/wlr_tablet_v2_pad.c:190 in add_tablet_pad_group
  0x0c0480058110: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fd
  0x0c0480058120: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fa
  0x0c0480058130: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fa
  0x0c0480058140: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fa
  0x0c0480058150: fa fa fd fd fa fa fd fd fa fa fd fd fa fa fd fd
=>0x0c0480058160: fa fa fd fd fa fa fd fd fa fa fd fa fa fa[04]fa
  0x0c0480058170: fa fa 01 fa fa fa 01 fa fa fa fd fd fa fa fd fd
  0x0c0480058180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0480058190: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c04800581a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c04800581b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==6673==ABORTING
```

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+	struct wlr_tablet_tool_client_v2 *tool = tablet_tool_client_from_resource(resource);
+	if (!tool) {
+		return;
+	}
+
+	struct wlr_surface *surface = NULL;
+	if (surface_resource != NULL) {
+		surface = wlr_surface_from_resource(surface_resource);
+		if (wlr_surface_set_role(surface, "wl_pointer-cursor", surface_resource,
+				WL_POINTER_ERROR_ROLE) < 0) {
```
acrisci: I think the error to use here is `ZWP_TABLET_TOOL_V2_ERROR_ROLE`.

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+		int32_t hotspot_x, int32_t hotspot_y) {
+	struct wlr_tablet_tool_client_v2 *tool = tablet_tool_client_from_resource(resource);
+	if (!tool) {
+		return;
+	}
+
+	struct wlr_surface *surface = NULL;
+	if (surface_resource != NULL) {
+		surface = wlr_surface_from_resource(surface_resource);
+		if (wlr_surface_set_role(surface, "wl_pointer-cursor", surface_resource,
```
acrisci: I think the `error_resource` for this call is the client resource, not the surface resource.

include/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,93 @@
...
+};
+
+struct wlr_tablet_seat_client_v2 {
+	struct wl_list seat_link;
+	struct wl_list client_link;
+	struct wl_client *wl_client;
+	struct wl_resource *resource;
+
+	struct wlr_tablet_manager_client_v2 *client;
+	struct wlr_seat_client *seat;
```
acrisci: Would prefer to call this `seat_client`.

include/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,93 @@
...
+void destroy_tablet_pad_v2(struct wl_resource *resource);
+struct wlr_tablet_pad_client_v2 *tablet_pad_client_from_resource(struct wl_resource *resource);
+void add_tablet_pad_client(struct wlr_tablet_seat_client_v2 *seat, struct wlr_tablet_v2_tablet_pad *pad);
+
+void destroy_tablet_tool_v2(struct wl_resource *resource);
+struct wlr_tablet_tool_client_v2 *tablet_tool_client_from_resource(struct wl_resource *resource);
+void add_tablet_tool_client(struct wlr_tablet_seat_client_v2 *seat, struct wlr_tablet_v2_tablet_tool *tool);
+
+struct wlr_tablet_seat_client_v2 *tablet_seat_client_from_resource(struct wl_resource *resource);
+void wlr_tablet_seat_client_v2_destroy(struct wl_resource *resource);
```
acrisci: Does this need to be exported?

types/tablet_v2/wlr_tablet_v2.c
```diff
@@ -0,0 +1,310 @@
...
+};
+
+struct wlr_tablet_seat_client_v2 *tablet_seat_client_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &zwp_tablet_seat_v2_interface,
+		&seat_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+void wlr_tablet_seat_client_v2_destroy(struct wl_resource *resource) {
```
acrisci: I think we normally have two different functions for this, where one takes a resource and the other takes the object.

emersion: Yeah, I'm not a fan of exposing functions which take a generic `wl_resource` as an argument.

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+
+void wlr_send_tablet_v2_tablet_tool_distance(
+	struct wlr_tablet_v2_tablet_tool *tool, double distance);
+
+void wlr_send_tablet_v2_tablet_tool_tilt(
+	struct wlr_tablet_v2_tablet_tool *tool, double x, double y);
```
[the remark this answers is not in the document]
Ongy: Relative to what? And why would they be something with s?
They aren't exactly coordinates, either.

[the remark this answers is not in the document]
emersion: Relative to the surface.

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+
+void wlr_send_tablet_v2_tablet_tool_rotation(
+	struct wlr_tablet_v2_tablet_tool *tool, double degrees);
+
+void wlr_send_tablet_v2_tablet_tool_slider(
+	struct wlr_tablet_v2_tablet_tool *tool, double position);
```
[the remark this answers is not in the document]
Ongy: I kept the values it takes closer to the wlroots side of things, which has values in `[0, 1]`, while the protocol has them in `[0, 65535]` and convert in the send functions.

Which I prefer, but I can take the final ints as well

include/wlr/types/wlr_tablet_v2.h
Ongy: Same as slider

include/wlr/types/wlr_tablet_v2.h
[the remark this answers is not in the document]
emersion: Yeah, +1 for [0, 1]

types/tablet_v2/wlr_tablet_v2.c
Ongy: Yea, this one should be more internal, it's in the right header, but should have the wlr prefix dropped.

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,573 @@
...
+	struct tablet_pad_auxiliary_user_data *aux = wl_resource_get_user_data(resource);
+	if (!aux) {
+		return;
+	}
+
+	struct wlr_tablet_v2_event_feedback evt = {
+		.serial = serial,
+		.description = description,
+		.index = aux->index
+	};
```
acrisci: Indentation here doesn't seem right.

include/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,93 @@
...
+	struct wlr_tablet_seat_client_v2 *seat;
+
+	struct wl_event_source *frame_source;
+};
+
+struct wlr_tablet_client_v2 *tablet_client_from_resource(struct wl_resource *resource);
+void destroy_tablet_v2(struct wl_resource *resource);
+void add_tablet_client(struct wlr_tablet_seat_client_v2 *seat, struct wlr_tablet_v2_tablet *tablet);
+
+void destroy_tablet_pad_v2(struct wl_resource *resource);
```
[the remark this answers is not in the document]
Ongy: Where are more private headers than `include/types/`? The public ones are in `include/wlr/types/`

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,573 @@
...
+}
+
+static void handle_tablet_pad_ring_v2_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static struct zwp_tablet_pad_ring_v2_interface tablet_pad_ring_impl = {
+	.set_feedback = handle_tablet_pad_ring_v2_set_feedback,
+	.destroy = handle_tablet_pad_ring_v2_destroy,
```
acrisci: Normally, anything that begins with `handle_` is a signal handler and anything you need to distinguish to be a protocol function should be of the form `blet_pad_ring_v2_protocol_set_feedback`. This is kind of throwing me off.

include/types/wlr_tablet_v2.h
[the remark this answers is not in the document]
acrisci: Oh ok yeah this is a private header nevermind.

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,573 @@
...
+	memcpy(button_array.data, group->buttons, group->button_count * sizeof(int));
+	zwp_tablet_pad_group_v2_send_buttons(client->groups[index], &button_array);
+	wl_array_release(&button_array);
+
+	client->strip_count = group->strip_count;
+	for (size_t i = 0; i < group->strip_count; ++i) {
+		size_t strip = group->strips[i];
+		struct tablet_pad_auxiliary_user_data *user_data =
+			calloc(1, sizeof(struct tablet_pad_auxiliary_user_data));
+		if (!user_data) {
```
acrisci: Need to post no memory here.

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,573 @@
...
+		size_t ring = group->rings[i];
+		struct tablet_pad_auxiliary_user_data *user_data =
+			calloc(1, sizeof(struct tablet_pad_auxiliary_user_data));
+		if (!user_data) {
+			continue;
+		}
+		user_data->pad = client;
+		user_data->index = ring;
+		client->rings[ring] =
+			wl_resource_create(client->client, &zwp_tablet_pad_ring_v2_interface, 1, 0);
```
acrisci: Need out of memory check for this.

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,573 @@
...
+		struct wlr_tablet_v2_tablet_pad *pad) {
+	struct wlr_tablet_pad_client_v2 *client =
+		calloc(1, sizeof(struct wlr_tablet_pad_client_v2));
+	if (!client) {
+		wl_client_post_no_memory(seat->wl_client);
+		return;
+	}
+	client->pad = pad;
+
+	client->groups = calloc(sizeof(int), wl_list_length(&pad->wlr_pad->groups));
```
acrisci: Are these `calloc` args backwards?

include/wlr/types/wlr_tablet_v2.h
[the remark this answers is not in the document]
Ongy: https://cgit.freedesktop.org/wayland/wayland-protocols/tree/unstable/tablet/tablet-unstable-v2.xml#n492 tilt isn't relative to the surface though.

The only case where this could be converted by some surface property is if it's rotated, but I'm not sure if that should rotate the x/y values here.

Ongy:
Can you re-test the asan bug? I think I had a wrong `sizeof` in there, should be fixed.

include/wlr/types/wlr_tablet_v2.h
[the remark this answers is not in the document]
acrisci: Oh ok so the coordinate system is relative to the device itself. I still think any bare x/y is a mistake because there are so many coordinate systems in this library. What about calling it `tilt_x` and `tilt_y`?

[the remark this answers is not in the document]
emersion: Meh, that doesn't really help…

acrisci:
New error:

```
==9740==ERROR: AddressSanitizer: heap-use-after-free on address 0x60b000220e08 at pc 0x7f8987ee651e bp 0x7ffc441d7b70 sp 0x7ffc441d7b60
READ of size 8 at 0x60b000220e08 thread T0
    #0 0x7f8987ee651d in destroy_tablet_pad_group_v2 ../types/tablet_v2/wlr_tablet_v2_pad.c:171
    #1 0x7f8987be4025  (/usr/lib64/libwayland-server.so.0+0x8025)
    #2 0x7f8987be8851  (/usr/lib64/libwayland-server.so.0+0xc851)
    #3 0x7f8987be4140 in wl_client_destroy (/usr/lib64/libwayland-server.so.0+0x8140)
    #4 0x7f8987be421c  (/usr/lib64/libwayland-server.so.0+0x821c)
    #5 0x7f8987be5f01 in wl_event_loop_dispatch (/usr/lib64/libwayland-server.so.0+0x9f01)
    #6 0x7f8987be4601 in wl_display_run (/usr/lib64/libwayland-server.so.0+0x8601)
    #7 0x41e85c in main ../rootston/main.c:74
    #8 0x7f8986d8718a in __libc_start_main (/usr/lib64/libc.so.6+0x2318a)
    #9 0x406a99 in _start (/home/tony/projects/wlroots/build/rootston/rootston+0x406a99)

0x60b000220e08 is located 72 bytes inside of 112-byte region [0x60b000220dc0,0x60b000220e30)
freed by thread T0 here:
    #0 0x7f8988254880 in __interceptor_free (/usr/lib64/libasan.so.5+0xee880)
    #1 0x7f8987ee5ba8 in destroy_tablet_pad_v2 ../types/tablet_v2/wlr_tablet_v2_pad.c:56
    #2 0x7f8987ee45fd in tablet_seat_client_v2_destroy ../types/tablet_v2/wlr_tablet_v2.c:112
    #3 0x7f8987ee4dbf in wlr_tablet_manager_v2_destroy ../types/tablet_v2/wlr_tablet_v2.c:232
    #4 0x7f8987be4025  (/usr/lib64/libwayland-server.so.0+0x8025)

    #0 0x7f8988254e50 in calloc (/usr/lib64/libasan.so.5+0xeee50)
    #1 0x7f8987ee76ad in add_tablet_pad_client ../types/tablet_v2/wlr_tablet_v2_pad.c:268
    #2 0x7f8987ee4bcc in get_tablet_seat ../types/tablet_v2/wlr_tablet_v2.c:202
    #3 0x7f8983ac403d in ffi_call_unix64 (/usr/lib64/libffi.so.6+0x603d)

SUMMARY: AddressSanitizer: heap-use-after-free ../types/tablet_v2/wlr_tablet_v2_pad.c:171 in destroy_tablet_pad_group_v2
  0x0c168003c170: fa fa fa fa fa fa fd fd fd fd fd fd fd fd fd fd
  0x0c168003c180: fd fd fd fd fa fa fa fa fa fa fa fa fd fd fd fd
  0x0c168003c190: fd fd fd fd fd fd fd fd fd fd fa fa fa fa fa fa
  0x0c168003c1a0: fa fa fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c168003c1b0: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
=>0x0c168003c1c0: fd[fd]fd fd fd fd fa fa fa fa fa fa fa fa 00 00
  0x0c168003c1d0: 00 00 00 00 00 00 00 00 00 00 00 00 fa fa fa fa
  0x0c168003c1e0: fa fa fa fa 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c168003c1f0: 00 00 fa fa fa fa fa fa fa fa 00 00 00 00 00 00
  0x0c168003c200: 00 00 00 00 00 00 00 fa fa fa fa fa fa fa fa fa
  0x0c168003c210: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==9740==ABORTING
```

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+	struct wlr_tablet_manager_v2 *manager,
+	struct wlr_seat *wlr_seat,
+	struct wlr_input_device *wlr_device);
+
+struct wlr_tablet_v2_tablet_tool *wlr_tablet_tool_create(
+	struct wlr_tablet_manager_v2 *manager,
+	struct wlr_seat *wlr_seat,
+	struct wlr_tablet_tool *wlr_tool);
+
+struct wlr_tablet_manager_v2 *wlr_tablet_v2_create(struct wl_display *display);
```
acrisci: Should be `wlr_tablet_manager_v2_create()`.

include/wlr/types/wlr_tablet_v2.h
```diff
@@ -0,0 +1,171 @@
...
+	struct wlr_seat_client *seat_client;
+};
+
+struct wlr_tablet_v2_event_feedback {
+	const char *description;
+	size_t index;
+	uint32_t serial;
+};
+
+struct wlr_tablet_v2_tablet *wlr_tablet_create(
```
acrisci: Need versions for these things or they'll conflict with stable. `wlr_tablet_v2_tablet_create()`.

include/wlr/types/wlr_tablet_v2.h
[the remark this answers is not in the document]
acrisci: Oh ok, that for sure needs documentation.

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,583 @@
...
+		zwp_tablet_pad_v2_send_button(pad->current_client->resource,
+				time, button, state);
+	}
+}
+
+void wlr_send_tablet_v2_tablet_pad_strip(struct wlr_tablet_v2_tablet_pad *pad,
+		uint32_t strip, double position, bool finger, uint32_t time) {
+	if (!pad->current_client &&
+			pad->current_client->strips &&
+			pad->current_client->strips[strip]) {
```
acrisci: This guard looks messed up.

types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -0,0 +1,583 @@
...
+		zwp_tablet_pad_strip_v2_send_position(resource, position * 65535);
+	}
+	zwp_tablet_pad_strip_v2_send_frame(resource, time);
+}
+
+void wlr_send_tablet_v2_tablet_pad_ring(struct wlr_tablet_v2_tablet_pad *pad,
+		uint32_t ring, double position, bool finger, uint32_t time) {
+	if (!pad->current_client ||
+			!pad->current_client->rings ||
+			!pad->current_client->rings[ring]) {
```
acrisci: Also need to check that rings is less than ring count right?

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+	case WLR_TABLET_TOOL_TYPE_BRUSH:
+		return ZWP_TABLET_TOOL_V2_TYPE_BRUSH;
+	case WLR_TABLET_TOOL_TYPE_PENCIL:
+		return ZWP_TABLET_TOOL_V2_TYPE_PENCIL;
+	case WLR_TABLET_TOOL_TYPE_AIRBRUSH:
+		return ZWP_TABLET_TOOL_V2_TYPE_AIRBRUSH;
+	case WLR_TABLET_TOOL_TYPE_MOUSE:
+		return ZWP_TABLET_TOOL_V2_TYPE_MOUSE;
+	case WLR_TABLET_TOOL_TYPE_LENS:
+		return ZWP_TABLET_TOOL_V2_TYPE_LENS;
```
acrisci: There's no wlr type, but this is missing the tablet tool finger type.

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+	// Send the expected events
+	if (tool->wlr_tool->hardware_serial) {
+			zwp_tablet_tool_v2_send_hardware_serial(
+			client->resource, 
+			tool->wlr_tool->hardware_serial >> 32,
+			tool->wlr_tool->hardware_serial & 0xFFFFFFFF);
+	}
+	if (tool->wlr_tool->hardware_wacom) {
+			zwp_tablet_tool_v2_send_hardware_id_wacom(
+			client->resource,
```
acrisci: trailing

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+			tool->wlr_tool->hardware_serial >> 32,
+			tool->wlr_tool->hardware_serial & 0xFFFFFFFF);
+	}
+	if (tool->wlr_tool->hardware_wacom) {
+			zwp_tablet_tool_v2_send_hardware_id_wacom(
+			client->resource, 
+			tool->wlr_tool->hardware_wacom >> 32,
+			tool->wlr_tool->hardware_wacom & 0xFFFFFFFF);
+	}
+	zwp_tablet_tool_v2_send_type(client->resource,
```
acrisci: trailing

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+		return;
+	}
+	wl_resource_set_implementation(client->resource, &tablet_tool_impl,
+		client, destroy_tablet_tool_v2);
+	zwp_tablet_seat_v2_send_tool_added(seat->resource, client->resource);
+
+	// Send the expected events
+	if (tool->wlr_tool->hardware_serial) {
+			zwp_tablet_tool_v2_send_hardware_serial(
+			client->resource,
```
acrisci: trailing

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+struct wlr_tablet_tool_client_v2 *tablet_tool_client_from_resource(struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &zwp_tablet_tool_v2_interface,
+		&tablet_tool_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+
+/* Actual protocol foo */
+// https://www.geeksforgeeks.org/move-zeroes-end-array/
+static size_t push_zeroes_to_end(uint32_t arr[], size_t n) {
```
acrisci: This is used in `wlr_keyboard.c` as well. Maybe we need a utility function?

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+	size_t i = 0;
+	for (; i < tool->num_buttons; ++i) {
+		if (tool->pressed_buttons[i] == button) {
+			found = true;
+			break;
+		}
+	}
+
+	if (state == ZWP_TABLET_PAD_V2_BUTTON_STATE_PRESSED && !found &&
+			tool->num_buttons < WLR_TABLET_V2_TOOL_BUTTONS_CAP) {
```
acrisci: We need to at least warn here if they try to press 17 buttons.

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+		i = tool->num_buttons++;
+		tool->pressed_buttons[i] = button;
```
acrisci: What if they pass button = 0?

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -0,0 +1,515 @@
...
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	zwp_tablet_tool_v2_send_frame(tool->resource, timespec_to_msec(&now));
+	tool->frame_source = NULL;
+}
+
+static void queue_tool_frame(struct wlr_tablet_tool_client_v2 *tool) {
+	struct wl_display *display = wl_client_get_display(tool->client);
+	struct wl_event_loop *loop = wl_display_get_event_loop(display);
+	if (!tool->frame_source) {
```
acrisci: Could the frame be queued for a different client?

types/tablet_v2/wlr_tablet_v2_tool.c
Ongy: Provided that we find a tablet tool like the Naga :)
But agreed, it should be warned about

types/tablet_v2/wlr_tablet_v2_tool.c
Ongy: AFAIK the buttons start at 0x100 (though that may be linux specific), I should ask about this

types/tablet_v2/wlr_tablet_v2_tool.c
Ongy: I don't quite the question tbh.
The idea here is to queue it for the client struct of the tool (it's stored there as well).
The only situation where two clients could have a need for tablet events is when they change focus, which has to go over `property_out` which forces the frame inline.

rootston/seat.c
```diff
@@ -97,13 +101,102 @@ static void handle_touch_motion(struct wl_listener *listener, void *data) {
...
+	double sx, sy;
+	struct roots_view *view = NULL;
+	struct roots_seat *seat = cursor->seat;
+	struct roots_desktop *desktop = seat->input->server->desktop;
+	struct wlr_surface *surface = desktop_surface_at(desktop,
+			cursor->cursor->x, cursor->cursor->y, &sx, &sy, &view);
+	struct roots_tablet_tool *roots_tool = tool->data;
+	
+	if (!surface) {
+		wlr_send_tablet_v2_tablet_tool_proximity_out(roots_tool->tablet_v2_tool);
```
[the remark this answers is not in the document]
Ongy: Grabs aren't implemented yet, it's mentioned in the initial post as follow up issue to implement.

That includes stuff like keeping back the focus change when buttons are pressed as well (drag&drop).
I think it's a feature we will need in practice, but this is massive enough and I want to get it in, that part should be mostly stacked on top of the current work, without many changes to what I have done so far.

types/tablet_v2/wlr_tablet_v2.c
Ongy: This *is* in the private header :) `include/types/` not `/include/wlr/tablet/` :+1:

Ongy: Ah wait, no. I got confused for a second. And I still didn't get any answer about what the proper non-colliding naming scheme would be here

emersion: `tablet_seat_get_or_create` seems reasonable

types/tablet_v2/wlr_tablet_v2_tool.c
Ongy: At least freebsd uses a copy of the same codes. 0 is KEY_RESERVED on both.

SirCmpwn:
I want to merge this. Can you resolve conflicts?

SirCmpwn:
Thanks!

Ongy:
This one was quite the ordeal!

Thanks for the patience and reviews :)

PR #965, opened 2018-05-09
Fix hardware cursors scale & transform

emersion:
Test plan: try different combinations of output scale/transform, try compositor and client cursors, try clients that support HiDPI (e.g. GTK+) and clients that don't (e.g. Xwayland clients).

cc @Ongy @Timidger: this is a breaking change for the output implementation API.


atomnuker:
Works fine.

SirCmpwn:
Thanks!

PR #969, opened 2018-05-11
xdg-output: redesign resource destroy

emersion:
And rename `xdg_*` functions to prevent clash with protocol structs.

Updates #909


SirCmpwn:
Thanks!

PR #974, opened 2018-05-13
fix check of EGL_WL_bind_wayland_display presence

agreppin:
Tested  on virtualbox with Fedora 28 guest, which does not have the `EGL_WL_bind_wayland_display` extension. The function pointer `eglBindWaylandDisplayWL `is non-null and causes a crash when used (SIGSEGV in `__strlen_avx2`).

Instead use `struct wlr_egl.egl_exts.bind_wayland_display` which is correctly set in `wlr_egl_init`, but not yet used...

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -209,7 +209,7 @@ void wlr_egl_finish(struct wlr_egl *egl) {
 }
 
 bool wlr_egl_bind_display(struct wlr_egl *egl, struct wl_display *local_display) {
-	if (!eglBindWaylandDisplayWL) {
+	if (!egl->egl_exts.bind_wayland_display) {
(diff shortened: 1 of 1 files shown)


emersion:
Nice catch. Is it possible to test for both?

SirCmpwn:
We should probably remove `eglBindWaylandDisplayWL` entirely

emersion:
>We should probably remove eglBindWaylandDisplayWL entirely

What do you mean? It's what makes client EGL work without wl_shm.

SirCmpwn:
Sorry, I was misunderstanding this.

SirCmpwn:
Thanks @agreppin!

PR #977, opened 2018-05-14
Split xdg-shell into multiple files

emersion:
I had to rename xdg-shell-unstable-v6 functions because of conflicts with xdg-shell stable (both the generated Wayland protocol code and the new private functions).


SirCmpwn:
Breaking changes

Ongy:
Are they really breaking changes? Nothing under `include/wlr/` changed, just `include/types/`, so it should be internal changes.

And just skimming the diffs, it looks like mostly "if it compiles, it works" changes, should I do indepth review?

emersion:
>Are they really breaking changes? Nothing under include/wlr/ changed, just include/types/, so it should be internal changes.

Removed one function which was incorrectly used on popups with a layer surface parent.

>And just skimming the diffs, it looks like mostly "if it compiles, it works" changes, should I do indepth review?

Mostly. I also tested it to make sure assertions still hold, but yeah, there are no logic changes.

emersion:
Bump

SirCmpwn:
Thanks!

PR #978, opened 2018-05-14
backend: add WLR_BACKEND env variable

### LINKED ISSUE #942: Add a WLR_BACKEND env var to choose the autocreate backend

emersion:
Fixes #942


SirCmpwn:
This should support a comma delimited list of multiple backends

emersion:
>This should support a comma delimited list of multiple backends

Hmm, we should definitely prevent multiple renderers at the same time.

SirCmpwn:
I really wish multiple renderers was a thing

emersion:
Me too. Too bad that +1000/-0 PR was sent to /dev/null. :'(

Life is hard

emersion:
Done

SirCmpwn:
Something about this still doesn't sit right with me. Maybe we should call it `__UNSTABLE__WLR_BACKENDS` and tell people it's subject to change.

PR #979, opened 2018-05-14
backend/wayland: print output name in window title

emersion:
This makes it easier to know which output is which window.

This just mirrors what the X11 backend does.


SirCmpwn:
Thanks!

PR #981, opened 2018-05-15
output: fix lingering hardware cursors

### LINKED ISSUE #976: Lingering cursor on output border
"Sometimes" (I'm not sure yet when exactly) a cursor lingers in the visible area on an output, after it was moved from it.

I have only seen this on waymonad with the DRM backend so far, it's hard to reproduce reliably, so I can't exactly test on rootston =.=

My observations so far:
* Cursor hotspot is outside visible area
* Usually had mpv on the output the cursor lingers on
* Not every cursor is visible (larger ones are, thinner ones aren't)

Just looking at it, I guess a small cursor isn't moved far enough outside the visible area, and uploading a new cursor makes a lingering cursor visible.
But looking at the code, I'd expec the cursor to move again when it's enlarged to be visible, so I'm a little confused

emersion:
I cannot test this patch, please try it for me.

Fixes #976 ~~(I hope)~~

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -706,11 +706,11 @@ static bool output_cursor_attempt_hardware(struct wlr_output_cursor *cursor) {
 
 	struct wlr_output_cursor *hwcur = cursor->output->hardware_cursor;
 	if (cursor->output->impl->set_cursor && (hwcur == NULL || hwcur == cursor)) {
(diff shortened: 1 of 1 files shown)


emersion:
All right, can confirm it fixes the issue :3

SirCmpwn:
Thanks!

Issue #984, opened 2018-05-18
Alpha blend mode is broken


emersion:
1. Start rootston with the Wayland backend
2. Open gnome-calculator
3. The window shadows (which are semi-transparent) clear the gray background, making the Wayland backend window semi-transparent

Can also reproduce on DRM by making a screenshot.

No idea how to fix this, the current blend mode looks good to me.

ammen99:
I have a hypothesis, but I'm no expert on GL, so what I say might be completely wrong. But looking at the documentation here: https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glBlendFunc.xhtml, it seems that by using `(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)` the alpha channel will be multiplied by itself and as a result the total alpha will be less than 1(after blending).I mean, the equation for the final alpha is actually `Sa * Sa + 1 - Sa` where `Sa` is the source alpha and the buffer RGBA values(which are taken for screenshot or when nested) becomes less than 1.

 A solution could be: In the fragment shader, pre-multiply RGB values by the alpha value, then use `(GL_ONE, GL_ONE_MINUS_SRC_ALPHA)`. This seems to fix the problem, if it looks good to you, I can make a PR later

ammen99:
I didn't know about `glBlendFuncSeparate` it might be the right fix (with rgb SRC_ALPHA, ONE_MINUS_SRC_ALPHA and for alpha ONE ONE_MINUS_SRC_ALPHA)

However I'm thinking what should be the resulting alpha when we blend two textures with alpha 0.5

emersion:
>However I'm thinking what should be the resulting alpha when we blend two textures with alpha 0.5?

That's a good question. For our uses-case it shouldn't happen too often (in fact only on Wayland/X11 backends with transparent backgrounds, unless a transparent screen is used but these are still pretty rare). Not sure but I'd say 0.5?

Paging @ascent12

VincentVanlaer:
[...]

PR #986, opened 2018-05-18
gles2: pre-multiply alpha and fix blending function

ammen99:
This is a basic implementation of what we discussed at #984

It seems to fix the issue with transparency in shadows or other translucent clients when running rootston on nested backend(wayland).


ammen99:
Seems I've forgotten to pre-multiply RGBX shaders, sorry for that. The second commit should address this issue.

emersion:
Can you squash these two commits together? (`git rebase -i`)

Works with `weston-simple-egl`. Shadows work correctly with GTK+ apps. This breaks titlebar transparency in Sway, because the title texture isn't pre-multiplied. I guess this should be fixed in Sway instead.

PR #892, opened 2018-04-23
[WIP] On-screen keyboard support

dcz-purism:
This pull request combines work on:
- `input-method` protocol #856 
- `text-input` protocol #776 
- rootston glue

Its purpose is to develop these in sync, and to remove the need to juggle different long-running feature branches as different parts are being developed. This has proven to take up effort and lead to awkward situations, e.g. many of the rebased rootston commits were developed against an earlier version of `wlr_input_method.h` and are currently unbuildable.

One interesting thing is the addition of code duplicating some functionality of layer surfaces (panels, unfocusable windows) in rootston views, in order to support `input-method-panel`.

PR #999, opened 2018-05-24
Support virtual keyboard protocol

dcz-purism:
This change is to support the recently proposed `virtual-keyboard-v1` protocol.

Because virtual keyboards are privileged, the commits setting up an authorization system - meant to be general for all input methods - are separate.

The main commit provides support in wlroots as well as rootston. The keyboard does not attempt to synchronize its modifier status with the rest of the system, and code for having multiple instances running at the same time has not been extensively tested, however it's rather simple and wasn't failing either.

### DIFF
--- a/include/rootston/virtual_keyboard.h
+++ b/include/rootston/virtual_keyboard.h
@@ -1,0 +1,7 @@
+#ifndef ROOTSTON_VIRTUAL_KEYBOARD_H
+#define ROOTSTON_VIRTUAL_KEYBOARD_H
+
+#include <wayland-server-core.h>
+
+void handle_virtual_keyboard(struct wl_listener *listener, void *data);
+#endif
--- a/rootston/virtual_keyboard.c
+++ b/rootston/virtual_keyboard.c
@@ -1,0 +1,21 @@
+#define _POSIX_C_SOURCE 199309L
+
+#include <wlr/util/log.h>
+#include <wlr/types/wlr_virtual_keyboard_v1.h>
+#include "rootston/virtual_keyboard.h"
+#include "rootston/seat.h"
+
+void handle_virtual_keyboard(struct wl_listener *listener, void *data) {
+	struct roots_desktop *desktop =
+		wl_container_of(listener, desktop, virtual_keyboard_new);
+	struct wlr_virtual_keyboard_v1 *keyboard = data;
+
+	struct roots_seat *seat = input_seat_from_wlr_seat(desktop->server->input,
+		keyboard->seat);
+	if (!seat) {
+		wlr_log(L_ERROR, "could not find roots seat");
+		return;
+	}
+
+	roots_seat_add_device(seat, &keyboard->input_device);
+}
--- a/include/rootston/desktop.h
+++ b/include/rootston/desktop.h
@@ -14,6 +14,7 @@
 #include <wlr/types/wlr_output.h>
 #include <wlr/types/wlr_primary_selection.h>
 #include <wlr/types/wlr_screenshooter.h>
+#include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include <wlr/types/wlr_wl_shell.h>
 #include <wlr/types/wlr_xcursor_manager.h>
 #include <wlr/types/wlr_xdg_shell_v6.h>
@@ -51,6 +52,7 @@ struct roots_desktop {
 	struct wlr_input_inhibit_manager *input_inhibit;
 	struct wlr_linux_dmabuf *linux_dmabuf;
 	struct wlr_layer_shell *layer_shell;
+	struct wlr_virtual_keyboard_manager_v1 *virtual_keyboard;
 
 	struct wl_listener new_output;
 	struct wl_listener layout_change;
@@ -61,6 +63,7 @@ struct roots_desktop {
 	struct wl_listener decoration_new;
 	struct wl_listener input_inhibit_activate;
 	struct wl_listener input_inhibit_deactivate;
+	struct wl_listener virtual_keyboard_new;
 
 #ifdef WLR_HAS_XWAYLAND
 	struct wlr_xwayland *xwayland;
--- a/include/wlr/types/wlr_virtual_keyboard_v1.h
+++ b/include/wlr/types/wlr_virtual_keyboard_v1.h
@@ -1,0 +1,37 @@
+#ifndef WLR_TYPES_WLR_VIRTUAL_KEYBOARD_V1_H
+#define WLR_TYPES_WLR_VIRTUAL_KEYBOARD_V1_H
+
+#include <wayland-server.h>
+#include <wlr/interfaces/wlr_input_device.h>
+#include <wlr/interfaces/wlr_keyboard.h>
+
+struct wlr_virtual_keyboard_manager_v1 {
+	struct wl_global *global;
+	struct wl_list resources; // struct wl_resource*
+	struct wl_list virtual_keyboards; // struct wlr_virtual_keyboard_v1*
+
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal new_virtual_keyboard; // struct wlr_virtual_keyboard_v1*
+	} events;
+};
+
+struct wlr_virtual_keyboard_v1 {
+	struct wl_resource *resource;
+	struct wlr_input_device input_device;
+	struct wlr_seat *seat;
+
+	struct wl_list link;
+
+	struct {
+		struct wl_signal destroy; // struct wlr_virtual_keyboard_v1*
+	} events;
+};
+
+struct wlr_virtual_keyboard_manager_v1* wlr_virtual_keyboard_manager_v1_create(
+	struct wl_display *display);
+void wlr_virtual_keyboard_manager_v1_destroy(
+	struct wlr_virtual_keyboard_manager_v1 *manager);
+
+#endif
--- a/protocol/meson.build
+++ b/protocol/meson.build
@@ -38,6 +38,7 @@ protocols = [
 	'idle.xml',
 	'screenshooter.xml',
 	'server-decoration.xml',
+	'virtual-keyboard-unstable-v1.xml',
 	'wlr-layer-shell-unstable-v1.xml',
 	'wlr-input-inhibitor-unstable-v1.xml',
 ]
--- a/protocol/virtual-keyboard-unstable-v1.xml
+++ b/protocol/virtual-keyboard-unstable-v1.xml
@@ -1,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="virtual_keyboard_unstable_v1">
+  <copyright>
+    Copyright © 2008-2011  Kristian Høgsberg
+    Copyright © 2010-2013  Intel Corporation
+    Copyright © 2012-2013  Collabora, Ltd.
+    Copyright © 2018       Purism SPC
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+  </copyright>
+
+  <interface name="zwp_virtual_keyboard_v1" version="1">
+    <description summary="virtual keyboard">
+      The virtual keyboard provides an application with requests which emulate
+      the behaviour of a physical keyboard.
+
+      This interface can be used by clients on its own to provide raw input
+      events, or it can accompany the input method protocol.
+    </description>
+
+    <request name="keymap">
+      <description summary="keyboard mapping">
+        Provide a file descriptor to the compositor which can be
+        memory-mapped to provide a keyboard mapping description.
+
+        Format carries a value from the keymap_format enumeration.
+      </description>
+      <arg name="format" type="uint" summary="keymap format"/>
+      <arg name="fd" type="fd" summary="keymap file descriptor"/>
+      <arg name="size" type="uint" summary="keymap size, in bytes"/>
+    </request>
+
+    <enum name="error">
+      <entry name="no_keymap" value="0" summary="No keymap was set"/>
+    </enum>
+
+    <request name="key">
+      <description summary="key event">
+        A key was pressed or released.
+        The time argument is a timestamp with millisecond granularity, with an
+        undefined base. All requests regarding a single object must share the
+        same clock.
+
+        Keymap must be set before issuing this request.
+
+        State carries a value from the key_state enumeration.
+      </description>
+      <arg name="time" type="uint" summary="timestamp with millisecond granularity"/>
+      <arg name="key" type="uint" summary="key that produced the event"/>
+      <arg name="state" type="uint" summary="physical state of the key"/>
+    </request>
+
+    <request name="modifiers">
+      <description summary="modifier and group state">
+        Notifies the compositor that the modifier and/or group state has
+        changed, and it should update state.
+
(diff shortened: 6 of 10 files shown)


dcz-purism:
The test client is at https://code.puri.sm/Librem5/weston/src/virtual_keyboard - please use the `weston-keyboard` program:

```
./configure --enable-clients
./weston-keyboard
```

rootston/config.c
```diff
@@ -19,14 +19,15 @@
 
 static void usage(const char *name, int ret) {
 	fprintf(stderr,
-		"usage: %s [-C <FILE>] [-E <COMMAND>]\n"
+		"usage: %s [-C <FILE>] [-E <COMMAND>] [-E <EXEC>]\n"
 		"\n"
```
agx: Duplicat -E, should be -K <KEYBOARD> ?

include/rootston/input_method.h
```diff
@@ -0,0 +1,11 @@
+//#include <wlr/types/wlr_input_method.h>
```
agx: Dead code?

rootston/input_method.c
```diff
@@ -0,0 +1,65 @@
...
+#include <wlr/util/log.h>
+#include "rootston/desktop.h"
+
+static pid_t *im_pid;
+static const char *im_cmd;
+
+static void spawn_im_process() {
+	*im_pid = fork();
+	if (*im_pid == 0) {
+		sleep(1);
```
agx: sleep for what? If a sleep is needed for any reason it shoud likely be polling for something to be not racy

dcz-purism: Initially, I added the sleep there because when `spawn_im_process` is first called, the server is not completely initialized, causing an immediate crash. Later on, I realized this would be useful to rate-limit the spawns.

emersion: -1

Rate-limiting shouldn't be done via `sleep`, and rootston shouldn't rate-limit. Child processes should only be spawned when the server is completely initialized.

include/wlr/types/wlr_virtual_keyboard.h
```diff
@@ -0,0 +1,34 @@
...
+        struct wl_client *client, void *data);
+
+struct wlr_virtual_keyboard_manager {
+    struct wl_global *global;
+    struct wl_resource *resource;
+    wlr_virtual_keyboard_check_credentials_func_t check_credentials;
+    void *credentials_data;
+    struct {
+        struct wl_signal new_virtual_keyboard; // struct wlr_virtual_keyboard*
+    };
```
agx: Other places use  
```
struct {
} events;
```

include/wlr/types/wlr_virtual_keyboard.h
```diff
@@ -0,0 +1,34 @@
...
+};
+
+struct wlr_virtual_keyboard {
+    struct wl_resource *resource;
+    struct wlr_input_device input_device;
+    struct wlr_seat *seat;
+    struct {
+		struct wl_signal destroy; // struct wlr_virtual_keyboard*
+    };
```
agx: Same here.

rootston/virtual_keyboard.c
```diff
@@ -0,0 +1,29 @@
...
+		keyboard->seat);
+	if (!seat) {
+		wlr_log(L_ERROR, "could not find roots seat");
+		return;
+	}
+
+	roots_seat_add_device(seat, &keyboard->input_device);
+	wl_signal_add(&keyboard->destroy, &desktop->virtual_keyboard_destroy);
+	desktop->virtual_keyboard_destroy.notify = handle_virtual_keyboard_destroy;
+}
```
agx: Why register it when ```handle_virtual_keyboard_destroy``` is a noop?

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,220 @@
...
+static const struct zwp_virtual_keyboard_v1_interface virtual_keyboard_impl;
+
+static struct wlr_virtual_keyboard *virtual_keyboard_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource,
+		&zwp_virtual_keyboard_v1_interface, &virtual_keyboard_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void	virtual_keyboard_keymap(struct wl_client *client,
```
agx: style nitpick: Extra space

rootston/input_method.c
dcz-purism: How to wait for full initialization?

include/wlr/types/wlr_virtual_keyboard.h
dcz-purism: Thanks, I knew something was off here.

rootston/input_method.c
dcz-purism: When I say rate-limit, I mean the situation where a faulty process will crash very soon after spawning and get restarted. I think it the infinite spawning loop should be prevented from excessively consuming resources, although not necessarily using sleep.

rootston/virtual_keyboard.c
dcz-purism: That's vestigial from an earlier version - removing.

types/wlr_virtual_keyboard.c
dcz-purism: This was actually a tab :)

SirCmpwn:
I don't like rootston having special support for spawing the IM and keeping up with its process. I also don't like that the wlroots implementation has a special function to check for authoriziation to the protocol. This isn't how we're going to secure protocols.

For now, just let any client use this protocol and yank out all of the `-K` and process handling stuff. If you want it to restart automatically just use a shell script that runs it in a loop.

rootston/input_method.c
```diff
@@ -0,0 +1,69 @@
...
+		wlr_log(L_INFO, "Skipping permission check for input method.");
+		return true;
+	}
+	pid_t allowed_pid = *((pid_t*)data);
+	pid_t pid;
+	uid_t uid;
+	gid_t gid;
+	wl_client_get_credentials(client, &pid, &uid, &gid);
+	return pid == allowed_pid;
+}
```
SirCmpwn: Basically this whole file should go away

rootston/config.c
```diff
@@ -400,14 +401,17 @@ struct roots_config *roots_config_create_from_args(int argc, char *argv[]) {
...
 		switch (c) {
 		case 'C':
 			config->config_path = strdup(optarg);
 			break;
 		case 'E':
 			config->startup_cmd = strdup(optarg);
 			break;
+		case 'K':
+			config->im_cmd = strdup(optarg);
+			break;
```
SirCmpwn: -1

include/wlr/types/wlr_virtual_keyboard.h
```diff
@@ -0,0 +1,34 @@
...
+#include <wlr/interfaces/wlr_input_device.h>
+#include <wlr/interfaces/wlr_keyboard.h>
+
+typedef bool (*wlr_virtual_keyboard_check_credentials_func_t)(
+	struct wl_client *client, void *data);
+
+struct wlr_virtual_keyboard_manager {
+	struct wl_global *global;
+	struct wl_resource *resource;
+	wlr_virtual_keyboard_check_credentials_func_t check_credentials;
```
SirCmpwn: -1

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,220 @@
...
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_virtual_keyboard.h>
+#include <wlr/util/log.h>
+#include <xkbcommon/xkbcommon.h>
+#include "util/signal.h"
+#include "virtual-keyboard-unstable-v1-protocol.h"
+
+
+static void keyboard_led_update(struct wlr_keyboard *wlr_kb, uint32_t leds) {
+	// unsupported by virtual keyboard protocol
```
SirCmpwn: Hmm, but what if it was

rootston/input_method.c
Ongy: You can either split your setup and call the spawn part directly before going into the main loop, or add an idle event at setup time to start the client from the mainloop.

agx:
> For now, just let any client use this protocol and yank out all of the -K and process handling stuff. If you 
> want it to restart automatically just use a shell script that runs it in a loop.

I think that's fine, we can carry this as a patch on top of rootston for the librem5 and rather work on the general logic to authorize clients since layer-shell, etc. are also affected.

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,209 @@
...
+	xkb_keymap_unref(keymap);
+	xkb_context_unref(context);
+	return;
+keymap_fail:
+	munmap(data, size);
+fd_fail:
+	xkb_context_unref(context);
+context_fail:
+	wl_client_post_no_memory(client);
+	return;
```
Shugyousha: This return is not needed.

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,209 @@
...
+
+	struct wlr_virtual_keyboard *virtual_keyboard = calloc(1,
+		sizeof(struct wlr_virtual_keyboard));
+	if (!virtual_keyboard) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	struct wlr_keyboard* keyboard = calloc(1, sizeof(struct wlr_keyboard));
+	if (keyboard == NULL) {
```
Shugyousha: Should be

if (!keyboard) {

for consistency.

emersion: Nah, both are fine

Shugyousha: The style is mixed in the same function though... your call.

dcz-purism:
Yanked and fixed.

include/rootston/virtual_keyboard.h
```diff
@@ -0,0 +1,3 @@
+#include <wayland-server-core.h>
```
SirCmpwn: Needs guards

SirCmpwn:
This is looking really good. Just had one minor issue, and I still need to test it on my touchscreen device.

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,208 @@
...
+	struct wl_resource *resource = wl_resource_create(client,
+		&zwp_virtual_keyboard_manager_v1_interface, version, id);
+
+	if (!resource) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	wl_resource_set_implementation(resource, &manager_impl, manager, NULL);
+	manager->resource = resource;
```
emersion: If two clients bind to the global, `manager->resource` will be overwritten. You should use a list and `wl_resource_get_link` instead.

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,208 @@
...
+	wlr_signal_emit_safe(&manager->events.new_virtual_keyboard, virtual_keyboard);
+}
+
+static const struct zwp_virtual_keyboard_manager_v1_interface manager_impl = {
+	.create_virtual_keyboard = virtual_keyboard_manager_create_virtual_keyboard,
+};
+
+static void virtual_keyboard_manager_bind(struct wl_client *client, void *data,
+		uint32_t version, uint32_t id) {
+	assert(client);
```
emersion: This assert is unnecessary.

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,208 @@
...
+		WLR_INPUT_DEVICE_KEYBOARD, &input_device_impl, "virtual keyboard",
+		0xabcd, 0x1234);
+
+	struct wlr_seat_client *seat_client = wlr_seat_client_from_resource(seat);
+
+	virtual_keyboard->input_device.keyboard = keyboard;
+	virtual_keyboard->resource = keyboard_resource;
+	virtual_keyboard->seat = seat_client->seat;
+	wl_signal_init(&virtual_keyboard->events.destroy);
+	wlr_signal_emit_safe(&manager->events.new_virtual_keyboard, virtual_keyboard);
```
emersion: We should have a list of all virtual keyboards in the manager.

types/wlr_virtual_keyboard.c
```diff
@@ -0,0 +1,208 @@
...
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	wl_resource_set_implementation(keyboard_resource, &virtual_keyboard_impl,
+		virtual_keyboard, virtual_keyboard_destroy_resource);
+
+	wlr_input_device_init(&virtual_keyboard->input_device,
+		WLR_INPUT_DEVICE_KEYBOARD, &input_device_impl, "virtual keyboard",
+		0xabcd, 0x1234);
```
emersion: Hmm, better to use 0 instead of these constants :)

SirCmpwn:
How does this work relate to https://github.com/swaywm/wlroots/pull/892?

include/wlr/types/wlr_virtual_keyboard.h
```diff
@@ -0,0 +1,28 @@
...
+	struct wl_resource *resource;
+	struct wlr_input_device input_device;
+	struct wlr_seat *seat;
+	struct {
+		struct wl_signal destroy; // struct wlr_virtual_keyboard*
+	} events;
+};
+
+struct wlr_virtual_keyboard_manager *wlr_virtual_keyboard_manager_create(
+	struct wl_display *display);
```
emersion: The compositor should have a way to destroy the manager here

include/wlr/types/wlr_virtual_keyboard.h
```diff
@@ -0,0 +1,28 @@
+#ifndef WLR_TYPES_WLR_VIRTUAL_KEYBOARD_H
+#define WLR_TYPES_WLR_VIRTUAL_KEYBOARD_H
```
emersion: All of the symbols should be include the unstable version too.

SirCmpwn:
Tested on touch screen :+1:

types/wlr_virtual_keyboard.c
dcz-purism: What for?

dcz-purism:
@SirCmpwn this provides some features which will be cut out from `input-method` protocol, so it's part of the work needed to get #892 done.

include/wlr/types/wlr_virtual_keyboard.h
dcz-purism: What do you mean?

types/wlr_virtual_keyboard.c
emersion: Let the users iterate on the list and properly cleanup when destroying the manager.

include/wlr/types/wlr_virtual_keyboard.h
emersion: This protocol is `virtual_keyboard_unstable_v1`. All symbols should be chosen so that there's no conflict when a new unstable version is published (e.g. `virtual_keyboard_unstable_v2`) or when the stable version is adopted (`virtual_keyboard`).

types/wlr_virtual_keyboard.c
dcz-purism: What's the advantage of letting manager do the cleanup over letting keyboards clean up themselves?
As I understand, there's not much utility in destroying the manager while objects are bound. In addition to that, the manager is acting as a factory rather than a collection, so there's no harm in leaving keyboards alive after destroying it.
In addition to that, having a list of active keyboards in the manager will need introducing another object to remember which keyboards are held by which user, making the whole thing a doubly-nested structure:

- manager global
- manager binds
- keyboards

The way it is now seems just as functional and can stay simple.

dcz-purism: Giving it a bit more thought, a binds list is not implied by a keyboard list, but I stand by my comment that it's not necessary.

emersion: No need for complicated list. The global, a list of resources and a list of keyboards in the manager is enough.

The list is useful to let compositors inspect it without having to maintain their own list.

include/wlr/types/wlr_virtual_keyboard.h
dcz-purism: Does the file name need to change?

emersion: Yes.

dcz-purism:
Added list of manager resources and keyboards, renamed public interface to include `_unstable_v1`.

types/wlr_virtual_keyboard_unstable_v1.c
```diff
@@ -0,0 +1,236 @@
...
+		struct wl_resource *resource, uint32_t format, int32_t fd,
+		uint32_t size) {
+	struct wlr_virtual_keyboard_unstable_v1 *keyboard =
+		virtual_keyboard_from_resource(resource);
+
+	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	if (!context) {
+		goto context_fail;
+	}
+	void *data = mmap(NULL, size, PROT_READ, MAP_PRIVATE, fd, 0);
```
emersion: In case of success, when is this unmapped?

include/wlr/types/wlr_virtual_keyboard_unstable_v1.h
```diff
@@ -0,0 +1,38 @@
...
+
+	struct wl_list link;
+
+	struct {
+		struct wl_signal destroy; // struct wlr_virtual_keyboard_unstable_v1*
+	} events;
+};
+
+struct wlr_virtual_keyboard_manager_unstable_v1*
+	wlr_virtual_keyboard_manager_unstable_v1_create(
```
emersion: We currently name unstable protocols without "unstable" in the symbols, just with `_v1`.

We should probably change that to adding a `z` prefix or something, but that's unrelated to this PR.

dcz-purism:
Removed "unstable" and fixed unmapping.

emersion:
Thanks!

PR #1000, opened 2018-05-25
[RFC] backends: implement custom EGL and renderer initialization

### LINKED ISSUE #987: [RFC] Proposal: Allow compositors to setup EGL manually
Currently the backend creates EGL itself. However, this doesn't let compositors specify EGL context attributes and perhaps some other options. While I certainly don't have much experience or knowledge of wlroots' backend code, I think something like the following can be implemented:

1) `wlr_backend_autocreate()` and similar are used to initialize the backend, but NOT the renderer.
2) Implement a new function, called for example `wlr_backend_attempt_egl(backend, attribs, ...)` in order to create the EGL context. This function can also create the GLES or whatever renderer, or perhaps this can be yet another function which must be called manually.

In this way we can not only support the different attribs of EGL, but also choose which renderer to create. For example when we have a vulkan renderer in the future, the compositor will be able to choose whether to use it or the "old" GL renderer.

ammen99:
Compositors now have more control over how the backend creates its
renderer. Currently all backends create an EGL/GLES2 renderer, so
the necessary attributes for creating the context are passed to a
user-provided callback function. It is responsible for initializing the
provided `wlr_egl` and to return a renderer. On fail, return NULL.

Fixes #987 

Things I'm not sure about:
1) Where to place and how to name `wlr_backend_renderer_autocreate()` - should it be in `backend.c` or `wlr_renderer.c`?
2) Should we allow passing NULL to `wlr_backend_autocreate()` or force the user to specify `wlr_backend_renderer_autocreate` as renderer initialization function?
3) Example - is another example needed? Or maybe implement such a function in rootston to demonstrate how this is done?
4) Testing - I've tested rootston only with the default initialization, and my own compositor with custom initialization. 
5) The headless backend didn't need a GLES2 renderer, or at least it didn't fail without one. This PR currently forces it to have a render, should I fix this somehow?

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -72,8 +72,9 @@ static size_t parse_outputs_env(const char *name) {
 	return outputs;
 }
 
-static struct wlr_backend *attempt_wl_backend(struct wl_display *display) {
-	struct wlr_backend *backend = wlr_wl_backend_create(display, NULL);
+static struct wlr_backend *attempt_wl_backend(struct wl_display *display,
+		wlr_renderer_create_func_t create_renderer_func) {
+	struct wlr_backend *backend = wlr_wl_backend_create(display, NULL, create_renderer_func);
 	if (backend == NULL) {
 		return NULL;
 	}
@@ -88,8 +89,8 @@ static struct wlr_backend *attempt_wl_backend(struct wl_display *display) {
 
 #ifdef WLR_HAS_X11_BACKEND
 static struct wlr_backend *attempt_x11_backend(struct wl_display *display,
-		const char *x11_display) {
-	struct wlr_backend *backend = wlr_x11_backend_create(display, x11_display);
+		const char *x11_display, wlr_renderer_create_func_t create_renderer_func) {
+	struct wlr_backend *backend = wlr_x11_backend_create(display, x11_display, create_renderer_func);
 	if (backend == NULL) {
 		return NULL;
 	}
@@ -104,8 +105,8 @@ static struct wlr_backend *attempt_x11_backend(struct wl_display *display,
 #endif
 
 static struct wlr_backend *attempt_headless_backend(
-		struct wl_display *display) {
-	struct wlr_backend *backend = wlr_headless_backend_create(display);
+		struct wl_display *display, wlr_renderer_create_func_t create_renderer_func) {
+	struct wlr_backend *backend = wlr_headless_backend_create(display, create_renderer_func);
 	if (backend == NULL) {
 		return NULL;
 	}
@@ -119,7 +120,8 @@ static struct wlr_backend *attempt_headless_backend(
 }
 
 static struct wlr_backend *attempt_drm_backend(struct wl_display *display,
-		struct wlr_backend *backend, struct wlr_session *session) {
+		struct wlr_backend *backend, struct wlr_session *session,
+		wlr_renderer_create_func_t create_renderer_func) {
 	int gpus[8];
 	size_t num_gpus = wlr_session_find_gpus(session, 8, gpus);
 	struct wlr_backend *primary_drm = NULL;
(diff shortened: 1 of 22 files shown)


backend/backend.c
```diff
@@ -165,15 +168,32 @@ static struct wlr_backend *attempt_backend_by_name(struct wl_display *display,
...
+			return attempt_drm_backend(display, backend, *session, create_renderer_func);
 		}
 	}
 
 	wlr_log(L_ERROR, "unrecognized backend '%s'", name);
 	return NULL;
 }
 
-struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
+struct wlr_renderer *wlr_backend_renderer_autocreate(struct wlr_egl *egl,
```
emersion: This should be named `wlr_renderer_autocreate` and should be in `wlr/render/wlr_renderer.h`

backend/backend.c
```diff
@@ -165,15 +168,32 @@ static struct wlr_backend *attempt_backend_by_name(struct wl_display *display,
...
+	EGLenum platform, void *remote_display, EGLint *config_attribs, EGLint visual_id) {
+
+	if (!wlr_egl_init(egl, platform, remote_display, config_attribs, visual_id)) {
+		wlr_log(L_ERROR, "Could not initialize EGL");
+		return NULL;
+	}
+
+	struct wlr_renderer *renderer = wlr_gles2_renderer_create(egl);
+	if (!renderer)
```
emersion: Style: missing brackets

backend/backend.c
```diff
@@ -165,15 +168,32 @@ static struct wlr_backend *attempt_backend_by_name(struct wl_display *display,
...
 		}
 	}
 
 	wlr_log(L_ERROR, "unrecognized backend '%s'", name);
 	return NULL;
 }
 
-struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
+struct wlr_renderer *wlr_backend_renderer_autocreate(struct wlr_egl *egl,
+	EGLenum platform, void *remote_display, EGLint *config_attribs, EGLint visual_id) {
```
emersion: Style: missing one indentation level

backend/backend.c
ammen99: Should the implementation also be in `render/wlr_renderer.c`?

emersion:
>Should we allow passing NULL to wlr_backend_autocreate() or force the user to specify wlr_backend_renderer_autocreate as renderer initialization function?

I think we should allow passing NULL.

>Example - is another example needed?

I don't think that's needed.

>Testing - I've tested rootston only with the default initialization, and my own compositor with custom initialization.

Can you think of any customization you wouldn't be able to do with this design?

>The headless backend didn't need a GLES2 renderer, or at least it didn't fail without one. This PR currently forces it to have a render, should I fix this somehow?

I don't understand. The headless backend is creating a renderer and initializing EGL. Try running `env WLR_BACKENDS=headless build/rootston/rootston`, and then spawning an app in it, and then screenshooting the output.

backend/backend.c
emersion: I think so, yes.

ammen99:
> I don't understand. The headless backend is creating a renderer and initializing EGL. Try running env WLR_BACKENDS=headless build/rootston/rootston, and then spawning an app in it, and then screenshooting the output.

I mean that before this, backend creation was successful even if EGL init succeeds, but GLES2 init fails

> Can you think of any customization you wouldn't be able to do with this design?

Didn't get what you mean (in connection to my question about testing). Otherwise I think this will be flexible enough for 1.0

emersion:
>I mean that before this, backend creation was successful even if EGL init succeeds, but GLES2 init fails

Oh yes, that should be fixed.

> Otherwise I think this will be flexible enough for 1.0

Yeah that was my question.

ammen99:
@ascent12 It is more than enough for me. I'll squash the commits shortly

SirCmpwn:
Thanks!

emersion:
cc @Ongy @Timidger: this changes the backend creation API by adding a new parameter

PR #1009, opened 2018-05-27
xdg popup parent is now wlr_xdg_surface

Timidger:
This is to make it the same as the v6 version. If you want the `wlr_surface` it's a field in the `wlr_xdg_surface` so while it's a breaking change it's not hard to get the previous behaviour.


Timidger:
I see. That's a little confusing then, since from an outsider's perspective it's odd that layer-shell requires changes to how xdg shell surfaces work (especially going from v6 to stable, though I know you want to remove v6 eventually).

PR #1011, opened 2018-05-28
log: add wlr_ prefix

emersion:
See https://github.com/swaywm/sway/pull/2009

Not sure what to do with `L_ERROR` et al.


SirCmpwn:
How about s/L_DEBUG/WLR_DEBUG/g et al?

emersion:
>Does #2009 need to be updated?

Yeah, I bet it won't work anymore, because it redefines `wlr_log_init`. It also needs to setup wlroots to use the sway logging callback.

emersion:
This kind of PRs needs to be merged pretty quickly otherwise there are tons of conflicts. Will resend another one.

PR #1014, opened 2018-05-29
Fix wlr_box_contains_point comparison

RyanDwyer:


emersion:
Does this break stuff?™

RyanDwyer:
Possibly, but only if a compositor has implemented a workaround for it (like subtracting 1 from the dimensions before passing it to this function), or if they accidentally depend on the calculation being incorrect. Either way, it's their fault.

emersion:
Thanks!

PR #1015, opened 2018-05-29
Only allow one modifier per DMA-BUF, split attributes struct in render/

### LINKED ISSUE #908: Consider dropping per-plane DMA-BUF modifiers
The kernel people say it was a mistake.

Thanks @atomnuker for pointing that out.

References:
* http://kiwitree.net/~chadv/vulkan/1.0-VK_EXT_image_drm_format_modifier/html/vkspec.html#_issues_58
* https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bae781b259269590109e8a4a8227331362b88212

emersion:
The DMA-BUF attributes struct is starting to be used in various parts of wlroots, some of which have nothing to do with the linux-dmabuf protocol. This extracts the common struct into `wlr/render/dmabuf.h`.

This also removes the ability to have per-plane modifiers.

Fixes #908

### DIFF
--- a/include/wlr/render/dmabuf.h
+++ b/include/wlr/render/dmabuf.h
@@ -1,0 +1,29 @@
+#ifndef WLR_RENDER_DMABUF_H
+#define WLR_RENDER_DMABUF_H
+
+// So we don't have to pull in linux specific drm headers
+#ifndef DRM_FORMAT_MOD_INVALID
+#define DRM_FORMAT_MOD_INVALID ((1ULL<<56) - 1)
+#endif
+
+#define WLR_DMABUF_MAX_PLANES 4
+
+enum wlr_dmabuf_attributes_flags {
+	WLR_DMABUF_ATTRIBUTES_FLAGS_Y_INVERT = 1,
+	WLR_DMABUF_ATTRIBUTES_FLAGS_INTERLACED = 2,
+	WLR_DMABUF_ATTRIBUTES_FLAGS_BOTTOM_FIRST = 4,
+};
+
+struct wlr_dmabuf_attributes {
+	int32_t width, height;
+	uint32_t format;
+	uint32_t flags; // enum wlr_dmabuf_attributes_flags
+	uint64_t modifier;
+
+	int n_planes;
+	uint32_t offset[WLR_DMABUF_MAX_PLANES];
+	uint32_t stride[WLR_DMABUF_MAX_PLANES];
+	int fd[WLR_DMABUF_MAX_PLANES];
+};
+
+#endif
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -243,9 +243,9 @@ static int gles2_get_dmabuf_modifiers(struct wlr_renderer *wlr_renderer,
 }
 
 static bool gles2_check_import_dmabuf(struct wlr_renderer *wlr_renderer,
-		struct wlr_dmabuf_buffer *dmabuf) {
+		struct wlr_dmabuf_attributes *attribs) {
 	struct wlr_gles2_renderer *renderer = gles2_get_renderer(wlr_renderer);
-	return wlr_egl_check_import_dmabuf(renderer->egl, dmabuf);
+	return wlr_egl_check_import_dmabuf(renderer->egl, attribs);
 }
 
 static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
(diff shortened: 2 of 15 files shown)


include/wlr/render/dmabuf.h
```diff
@@ -0,0 +1,29 @@
+#ifndef WLR_UTIL_DMABUF_H
+#define WLR_UTIL_DMABUF_H
+
+// So we don't have to pull in linux specific drm headers
```
ascent12: This only requires drm_fourcc.h, which FreeBSD actually has.
Too bad their DRM stuff is ridiculously ancient, and they don't have any of the modifier stuff. I don't even know if they even have dma-bufs.

render/gles2/renderer.c
```diff
@@ -243,9 +243,9 @@ static int gles2_get_dmabuf_modifiers(struct wlr_renderer *wlr_renderer,
 }
 
 static bool gles2_check_import_dmabuf(struct wlr_renderer *wlr_renderer,
-		struct wlr_dmabuf_buffer *dmabuf) {
+		struct wlr_dmabuf_attributes *attribs) {
 	struct wlr_gles2_renderer *renderer = gles2_get_renderer(wlr_renderer);
```
agx: We might want to do sanity checks on the buffer as well (e.g. for YUV) at some point that's why I passed in the buffer and not only the attribs but since it's not a public function we can change that back later.

include/wlr/render/dmabuf.h
```diff
@@ -0,0 +1,29 @@
...
+#define DRM_FORMAT_MOD_INVALID ((1ULL<<56) - 1)
+#endif
+
+#define WLR_DMABUF_MAX_PLANES 4
+
+enum wlr_dmabuf_attributes_flags {
+	WLR_DMABUF_BUFFER_ATTRIBS_FLAGS_Y_INVERT = 1,
+	WLR_DMABUF_BUFFER_ATTRIBS_FLAGS_INTERLACED = 2,
+	WLR_DMABUF_BUFFER_ATTRIBS_FLAGS_BOTTOM_FIRST = 4,
+};
```
agx: When the _buffer_ is dropped from the struct name (which is fine since we're now talking about a dmabuf not the buffer of the dmabuf wayland extensions) the _BUFFER_ should IMHO be dropped here as well. Otherwise it looks fine.

emersion: Ah, forgot this, thanks

render/gles2/renderer.c
emersion: I'm starting to think it would be better to remove this function and try to import+destroy the DMA-BUF in the protocol implementation instead. Thoughts?

agx: You mean rather do it in wlr_linux_dmabuf? I'd rather leave it where it is since YUV will need more stuff in egl etc (I should really sit down and do the YUV stuff since some things will become clearer then, but I'm currently a bit short on time due to RL stuff)

emersion: Okay, let's keep it there for now. I'm curious though:

1. "We might want to do sanity checks on the buffer as well" → what piece of information relevant to GL is there in `wlr_dmabuf_buffer` and not in `wlr_dmabuf_attributes`?
2. Can't we do these checks in `wlr_gles2_texture_from_dmabuf`?

agx: 1.)  Things like say NV12 having two planes can be done with attributes only but e.g. checking fd seekability can not
2.) From my understanding of the protocol you want to do these things _before_ you finally switch the resource from params_resource to buffer_resoure so doing it in wlr_gles2_texture_from_dmabuf would be too late

emersion: >checking fd seekability can not

Hum, why that? Can't you just `fseek(attribs->fd[i])` in the renderer code?

>From my understanding of the protocol you want to do these things before you finally switch the resource from params_resource to buffer_resoure so doing it in wlr_gles2_texture_from_dmabuf would be too late

Well, the idea would be to call `wlr_gles2_texture_from_dmabuf` before switching the resource to a buffer, and destroy the texture immediately. `gles2_check_import_dmabuf` already does this with `wlr_egl`.

agx: > Hum, why that? Can't you just fseek(attribs->fd[i]) in the renderer code?
That would work too. I'm fine with that either way.

> Well, the idea would be to call wlr_gles2_texture_from_dmabuf before switching the resource to a buffer, and destroy > the texture immediately. gles2_check_import_dmabuf already does this with wlr_egl.
I guess I'm confused now. So the idea would be to replace the call of ```wlr_renderer_check_import_dmabuf``` to ```wlr_gles2_texture_from_dmabuf``` + destroy to get rid of that function? Fine with me as well although having a explicit _check_ function makes the code more self documenting (at least for me).

emersion: >Fine with me as well although having a explicit check function makes the code more self documenting (at least for me).

Yeah, the idea is just to reduce the `wlr_renderer` API surface.

SirCmpwn:
Thanks!

PR #1017, opened 2018-05-29
Fix double-free in the rotation.c example.

tobiasblass:
The wl_display_destroy function already destroys the backend's renderer.
Freeing it by hand causes a segmentation fault.

### DIFF
--- a/examples/rotation.c
+++ b/examples/rotation.c
@@ -268,6 +268,5 @@ int main(int argc, char *argv[]) {
 	wl_display_run(display);
 
 	wlr_texture_destroy(state.cat_texture);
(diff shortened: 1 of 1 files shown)


emersion:
Hmm, the render is owned by the backend, so destroying the backend should work, but indeed destroying the renderer doesn't.

Can you fix `output-layout`, `tablet` and `touch` too please?

tobiasblass:
Indeed, destroying the backend works fine. `output-layout` is not affected, though, since it doesn't destroy the display in the end. I'm not sure if this is intended or not.

emersion:
Thanks for looking into this!

PR #1018, opened 2018-05-30
Compositor pointer grab for the focused view (fix #613)

### LINKED ISSUE #613: Pointer button release outside of view does not send it to surface via seat
To reproduce:

1. Click and drag text in `gnome-terminal`.
2. Hold the pointer button down after you exit the view.
3. Release the button outside of the view.
4. Move the pointer back into the view

The seat client of the view should have received a button up.

Instead, it did not. The text will still be highlighted as you move the pointer around in the surface.

acrisci:
Feedback on the approach is appreciated.

Tests:

* Test drag and drop still works.
* Test #613 is fixed.
* Click on the window and drag outside and make sure the view you clicked on is getting all those motion events outside the surface still.
    - When you release the button, make sure the pointer leaves the surface
    - When you press additional buttons outside the surface, make sure that surface is getting the buttons.
* Test xwayland decorations work correctly (cursors were broken before)

Compare the behavior to Weston and X11 (because the behaviors are different and Weston still has bugs related to this, but X11 works pretty well).

Open question:

Should we deny popups while the view has a pointer grab?

TODO:

* [ ] style cleanups
* [ ] handle the grabbed view closing during the grab
* [ ] handle management of `cursor->grabbed_view` a bit better

fixes #613


rootston/seat.c
```diff
@@ -652,8 +652,8 @@ static void seat_view_destroy(struct roots_seat_view *seat_view) {
 		seat->cursor->mode = ROOTS_CURSOR_PASSTHROUGH;
 	}
 
-	if (seat_view == seat->cursor->pointer_view) {
-		seat->cursor->pointer_view = NULL;
+	if (seat_view == seat->cursor->deco_view) {
```
emersion: Same for `pointer_view`?

rootston/cursor.c
```diff
@@ -228,6 +250,20 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 	struct wlr_surface *surface = desktop_surface_at(desktop,
 			lx, ly, &sx, &sy, &view);
 
+	if (!is_touch && cursor->pointer_view && cursor->pointer_view != view) {
```
emersion: This is racy: there's no guarantee there will be a motion event on this view before a button event.

What about saving on button press the view on which the press occurred?

rootston/cursor.c
```diff
@@ -115,6 +115,21 @@ static void roots_passthrough_cursor(struct roots_cursor *cursor,
...
+				seat->seat->pointer_state.focused_surface) {
+		sx = cursor->cursor->x - cursor->pointer_view->x;
+		sy = cursor->cursor->y - cursor->pointer_view->y;
+		wlr_seat_pointer_notify_motion(seat->seat, time, sx, sy);
+
+		struct roots_drag_icon *drag_icon;
+		wl_list_for_each(drag_icon, &seat->drag_icons, link) {
+			roots_drag_icon_update_position(drag_icon);
+		}
+		return;
```
emersion: Not a fan of these early returns

acrisci:
Here's a different approach that seems to work ok.

martinetd:
I'll just say it once because nobody else did it, but an issue that says "fix 613" is pretty hard to follow unless (even if?) you're close to the project, and a commit message that says "fix 613" or "fix bugs" is petty much useless to anyone taking a look at the git log later on.

However good the code is, the wrapping does matter, a bit.

On a more positive note, I'll have a look at the code now I looked up that this is about "keep track of what buttons are pressed"

acrisci:
Thanks for keeping me honest, @martinetd . The truth is I had no idea what I was going to do to fix this when I opened the PR. I come from the computer science school of "try random stuff until it works".

rootston/cursor.c
```diff
@@ -148,14 +154,36 @@ static void roots_passthrough_cursor(struct roots_cursor *cursor,
...
 }
 
+static void roots_grabbed_cursor(struct roots_cursor *cursor,
+		uint32_t time) {
+	struct roots_seat *seat = cursor->seat;
+	struct roots_view *view = cursor->grabbed_view;
+	double sx = cursor->cursor->x - view->x;
+	double sy = cursor->cursor->y - view->y;
+	if (view && view->wlr_surface) {
+		wlr_seat_pointer_notify_enter(seat->seat, view->wlr_surface, sx, sy);
```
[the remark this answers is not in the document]
acrisci: The seat keeps track of the surface that last got enter ("pointer focus") and won't do anything if you try to send enter to the same surface twice. If you send enter to a different surface, it will automatically send leave to the last surface that got enter.

The alternative would be to explicitly guard it, but the guard already happens internally so we don't bother.

rootston/cursor.c
```diff
@@ -148,14 +154,37 @@ static void roots_passthrough_cursor(struct roots_cursor *cursor,
...
+
 static void roots_cursor_update_position(
 		struct roots_cursor *cursor, uint32_t time) {
 	struct roots_seat *seat = cursor->seat;
 	struct roots_view *view;
+
+	if (cursor->mode == ROOTS_CURSOR_PASSTHROUGH &&
+			seat->seat->pointer_state.button_count > 0 &&
+			cursor->grabbed_view &&
+			!wlr_seat_pointer_has_grab(seat->seat)) {
```
martinetd: (rewriting previous comment because github marked it as outdated before it went live)
I don't like changing cursor mode on move, if there is no drawback to entering ROOT_CURSOR_GRABBED with no button count I'd set it at click-within-a-grab time. When you set grabbed_view?

rootston/cursor.c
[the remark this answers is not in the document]
martinetd: I don't think this needs guarding in particular, but it feels wrong to me that there could be any way for another view to have received an enter event between cursor leaving the surface area and the move -- so at best this enter here is noop because of the interal guard, but I'm worried it could hide another unwanted behaviour we wouldn't notice with this.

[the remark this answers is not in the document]
acrisci: This is just the standard idiom for sending pointer motion when the surface you want to send motion to is known. Motion and enter were separated because they are separate in the protocol and we want the seat to do whatever the protocol can do (although I don't think we have a known use case for enter without motion).

rootston/cursor.c
acrisci: This would be necessary for the behavior removed in 26c6c69, which is to deny popups (or other grabbing actions) when the pointer has a grab. This seems to be the behavior in X11 (try it) although not in Weston. Although in the X11 case, I'm not sure if this is enforced by the Xorg server or if clients voluntarily do not show popups in this case. The actual sending of buttons is the same in X11 and Weston as it is here however (you can see that clearly with urxvt).

rootston/cursor.c
```diff
@@ -148,14 +154,45 @@ static void roots_passthrough_cursor(struct roots_cursor *cursor,
...
+	struct roots_seat *seat = cursor->seat;
+	struct wlr_surface *grabbed_surface = cursor->grabbed_surface;
+	if (grabbed_surface != NULL) {
+		double sx = seat->seat->pointer_state.sx;
+		double sy = seat->seat->pointer_state.sy;
+		double dx = cursor->cursor->x - cursor->offs_x;
+		double dy = cursor->cursor->y - cursor->offs_y;
+
+		cursor->offs_x = cursor->cursor->x;
+		cursor->offs_y = cursor->cursor->y;
```
acrisci: This isn't good, right? Unfortunately there is no easy way to translate layout coordinates to surface coordinates for some arbitrary surface.

acrisci: Actually, I'm coming around to this way because translating layout coordinates to surface coordinates for an arbitrary surface is difficult, and this is how all the other grabs operate - by comparing movement to a reference point when the grab begins.

panaman67:
@acrisci Whats the status on this PR?

acrisci:
I've been on a break from my sway work, but at the time this was written, I thought this was the best approach to fix this bug. If it's still an issue, I should probably come back and complete this.

emersion:
>I thought this was the best approach to fix this bug

I agree

>I should probably come back and complete this.

Feel free not to do it if you're not interested in this anymore :)

If you are, you'll be welcome of course.

PR #1019, opened 2018-05-30
virtual-keyboard: fix wlr_virtual_keyboard_manager_v1_destroy

emersion:
* Stop listening to wl_display destroy signal, fixes a segfault if the manager is destroyed prior to the display
* Destroy resources created by the manager, fixes memory leaks if the compositor isn't using `wl_display_destroy_clients`

cc @dcz-purism


dcz-purism:
looks good!

PR #1020, opened 2018-05-30
render: remove wlr_renderer_check_import_dmabuf

emersion:


types/wlr_linux_dmabuf.c
```diff
@@ -212,7 +225,7 @@ static void params_create_common(struct wl_client *client,
 
 		// planes > 0 might be subsampled according to fourcc format
 		if (i == 0 && buffer->attributes.offset[i] +
-				buffer->attributes.stride[i] * height >= size) {
+				buffer->attributes.stride[i] * height > size) {
```
agx: Isn't >= correct here since we're starting at 0? In any case this should be in a separate commit.

emersion: Sooo. I changed this in my previous PR (#1015) and it breaks `weston-simple-dmabuf-drm --import-format=XRGB` (from weston HEAD). Do you think it's a weston bug?

agx: Uhh...I missed that change in your previous PR. 
The current code ```>=``` is more strict than just ```>``` or am I missing something? So I wonder how using ```>=``` would make something work that didn't work with just ```>``` ?

Anyways, I'd drop that change.

emersion: >The current code >= is more strict than just > or am I missing something?

Yeah, the current code is more strict because it makes sense and should work - see all previous checks.

>So I wonder how using >= would make something work that didn't work with just > ?

The current code (`>=`) breaks `weston-simple-dmabuf-drm`. So this change (`>`) unbreaks it.

>Anyways, I'd drop that change.

By "drop", do you mean keep it as strict or make it more lenient just like before #1015?

agx: Rereading the code the ```>=``` version is correct. I missed the ```* height``` there. So let's just move the part that turns ```>``` into ```>=``` into a separate commit so we don't sneak in any unrelated changes.

agx: I flipped ```>=``` with ```>``` in the above (sorry, to many things going on). I meant to say: let's make it ```>``` in a separate commit since that's correct. stride * height == size is o.k.

emersion: Okay. Is there any explanation why `stride * height == size` is OK? I assume we'd need to update the check above this one as well, in case of images with height=1.

agx: Take the extreme case of height == 1 and width == stride == size. We have pixels [0, width-1] but ```offset + stride == size```. (so maybe the check should read

```
offset + (stride * height) -1 >= size
```

which is

```
offset + (stride * height) > size
```

emersion: Oh yeah, that does make sense

emersion:
Fixed bounds checking. Test plan: `weston-simple-dmabuf-drm --import-format={XRGB,NV12}` plus try some Xwayland apps.

types/wlr_linux_dmabuf.c
```diff
@@ -207,15 +207,16 @@ static void params_create_common(struct wl_client *client,
 			// Skip checks if kernel does no support seek on buffer
 			continue;
 		}
-		if (buffer->attributes.offset[i] >= size) {
+		if (buffer->attributes.offset[i] > size) {
```
agx: The ```>=``` is here to make sure we have at least some data to read. Changing it to ```>``` means we're o.k. with having an offset but nothing left to read. I'd prefere to leave it as is but I'm o.k. either way.

types/wlr_linux_dmabuf.c
```diff
@@ -207,15 +207,16 @@ static void params_create_common(struct wl_client *client,
...
 			wl_resource_post_error(params_resource,
 				ZWP_LINUX_BUFFER_PARAMS_V1_ERROR_OUT_OF_BOUNDS,
 				"invalid offset %i for plane %d",
 				buffer->attributes.offset[i], i);
 			goto err_out;
 		}
 
-		if (buffer->attributes.offset[i] + buffer->attributes.stride[i]	> size) {
+		if (buffer->attributes.offset[i] + buffer->attributes.stride[i] > size ||
+				buffer->attributes.stride[i] == 0) {
```
agx: That makes sense, also I wonder if we should reject stride < width?

types/wlr_linux_dmabuf.c
agx: Looks good now. Thanks!

types/wlr_linux_dmabuf.c
emersion: Hmm, the intention was to check here that the offset is okay, and in the next condition that the it's not empty.

types/wlr_linux_dmabuf.c
emersion: I'd rather not do that because there are obscure formats which have less than one byte per pixel.

agx: Yeah, it would only work for plane 0

emersion:
Thanks for your review!

PR #1021, opened 2018-06-01
Add wlr_surface_get_geometry

Ongy:
This function defaults and clips the xdg-surface geometry to the
bounding box of the surface + its subsurfaces, as specified by the
protocol spec.

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -342,6 +342,15 @@ struct wlr_xdg_surface *wlr_xdg_surface_from_wlr_surface(
 		struct wlr_surface *surface);
 
 /**
+ * Get the surface geometry.
+ * This is either the geometry as set by the client, or defaulted to the bounds
+ * of the surface + the subsurfaces (as specified by the protocol).
+ *
+ * The x and y value can be <0
+ */
+void wlr_xdg_surface_get_geometry(struct wlr_xdg_surface *surface, struct wlr_box *box);
+
+/**
  * Call `iterator` on each surface in the xdg-surface tree, with the surface's
  * position relative to the root xdg-surface. The function is called from root to
  * leaves (in rendering order).
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -251,6 +251,13 @@ static void xdg_surface_handle_set_window_geometry(struct wl_client *client,
 		return;
 	}
 
+	if (width <= 0 || height <= 0) {
+		wlr_log(L_ERROR, "Client tried to set invalid geometry");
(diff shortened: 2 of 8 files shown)


include/wlr/types/wlr_xdg_shell.h
```diff
@@ -341,6 +341,15 @@ bool wlr_surface_is_xdg_surface(struct wlr_surface *surface);
...
 		struct wlr_surface *surface);
 
+/**
+ * Get the surface geometry.
+ * This is either the geometry as set by the client, or defaulted to the bounds
+ * of the surface + the subsurfaces (as specified by the protocol).
+ *
+ * The x and y value can be <0
+ */
+void wlr_surface_get_geometry(const struct wlr_xdg_surface *surface, struct wlr_box *box);
```
emersion: This should be named `wlr_xdg_surface_get_geometry`

types/xdg_shell/wlr_xdg_shell.c
```diff
@@ -170,3 +186,37 @@ void wlr_xdg_shell_destroy(struct wlr_xdg_shell *xdg_shell) {
...
 	free(xdg_shell);
 }
+
+static void wlr_surface_get_bounding_box(const struct wlr_surface *surface, struct wlr_box *box) {
+	int min_x = 0, min_y = 0;
+	int max_x = surface->current->width;
+	int max_y = surface->current->height;
+
+	struct wlr_subsurface *sub;
+	wl_list_for_each(sub, &surface->subsurfaces, parent_link) {
```
emersion: This isn't going to work for nested subsurfaces. Use surface iterators instead (`wlr_xdg_surface_for_each_surface`).

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -251,6 +251,10 @@ static void xdg_surface_handle_set_window_geometry(struct wl_client *client,
 		return;
 	}
 
+	if (width <= 0 || height <= 0) {
+		//FIXME: Post an error here (which one? protocol patch?)
```
emersion: You can log the error, and maybe disconnect the client.

emersion:
This should be backported to xdg-shell-v6 when ready. Adding a function that operates on `wlr_surface` may be useful:

```c
void wlr_surface_get_extents(struct wlr_surface *surface, struct wlr_box *box);
```

types/xdg_shell/wlr_xdg_shell.c
Ongy: Urgh I forgot about those.
But it should be `wlr_surface_for_each_surface` so we don't consider popups

emersion: Right

emersion:
Thanks!

PR #1028, opened 2018-06-03
Request a high priority EGL context

emersion:


render/egl.c
```diff
@@ -177,10 +166,50 @@ bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *remote_display,
...
+	EGLint attribs[5];
+	attribs[atti++] = EGL_CONTEXT_CLIENT_VERSION;
+	attribs[atti++] = 2;
+
+	// Try to reschedule all of our rendering to be completed first. If it
+	// fails, it will fallback to the default priority (MEDIUM).
+	if (egl->egl_exts.context_priority) {
+		attribs[atti++] = EGL_CONTEXT_PRIORITY_LEVEL_IMG;
+		attribs[atti++] = EGL_CONTEXT_PRIORITY_HIGH_IMG;
+	}
```
Ongy: I'd rather see constants and the `EGL_NONE` in `if {} else {}` than this weird `atti++` with the assert here.

Unless there's a good reason to expect that this will be expanded in the future of course

include/wlr/render/egl.h
```diff
@@ -21,6 +21,7 @@ struct wlr_egl {
 		bool dmabuf_import;
 		bool dmabuf_import_modifiers;
 		bool bind_wayland_display;
+		bool context_priority;
```
ascent12: Why does this need to be stored?

emersion: It's just for consistency with other extensions. Can remove it if you don't like it.

ascent12: There are other extensions we're using which aren't listed here, so the whole thing is kind of inconsistent.

emersion: Added a more consistent EGL extension checking (GL extension checking doesn't happen here).

ascent12: -EGL_KHR_debug
-EGL_(EXT/KHR)_platform_wayland
-EGL_(EXT/KHR)_platform_x11
-EGL_(MESA/KHR)_platform_gbm
-EGL_MESA_platform_surfaceless
-EGL_EXT_platform_base (Although we fail if this doesn't exist)

emersion: Ah, right, I missed a few it seems.

Not sure how to check `EGL_KHR_debug`: we use it before initializing the display.

emersion: For all platforms, it's probably better to expose a `wlr_egl_check_extension` function.

ascent12: It's not something that I actually care about that much, but I was really just pointing out the inconsistency.
Most of these things will just fail right at the start if they don't exist, and we have no real reason to remember them.

emersion: Okay, probably better to remember only what we need to remember then.

SirCmpwn:
This doesn't make sense for all consumers of wlr_egl

emersion:
>This doesn't make sense for all consumers of wlr_egl

That's true, it doesn't make sense for the Wayland and the X11 backends.

emersion:
Updated to only request high priority context on DRM.

PR #1030, opened 2018-06-03
Remove redundant roots_seat_remove_device function

### LINKED ISSUE #998: Orphaned declaration of `roots_seat_remove_device`
https://github.com/swaywm/wlroots/blob/master/include/rootston/seat.h#L97 - no matching function in the body.

Does `handle_keyboard_destroy` perform the same job?

Shugyousha:
This function is unimplemented and is redundant because all devices added
with roots_seat_add_device get destruction handlers assigned already.

This fixes #998.


emersion:
Thanks!

PR #1031, opened 2018-06-03
seat: allow clients to bind to seat multiple times

### LINKED ISSUE #1023: Remove wlr_seat_client_for_wl_client
This function assumes a client only binds once to the seat. This isn't the case of Firefox on Wayland.

martinetd:
This lets clients bind to a seat multiple times by re-using the existing
wlr_seat_client whenever a duplicate request happens.
Previously, an independant wlr_seat_client would be created and only
events from one would be processed.

Fixes #1023.

~~FWIW I'm not 100% happy with the style (new_seat_client bool and the need for a new function in the API, and that function's name), but at least it seems to work.~~
Cc @ongy @Timidger as API changes, even if I didn't need to change anything to rootston/sway, it's just a ~~new helper function~~ change in the `wlr_seat_client` struct

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -13,11 +13,11 @@
  * managed by wlr_seat; some may be NULL.
  */
 struct wlr_seat_client {
-	struct wl_resource *wl_resource;
 	struct wl_client *client;
 	struct wlr_seat *seat;
 
 	// lists of wl_resource
+	struct wl_list wl_resources;
 	struct wl_list pointers;
 	struct wl_list keyboards;
 	struct wl_list touches;
--- a/types/seat/wlr_seat.c
+++ b/types/seat/wlr_seat.c
@@ -60,6 +60,12 @@ static void seat_client_handle_resource_destroy(
 		struct wl_resource *seat_resource) {
 	struct wlr_seat_client *client =
 		wlr_seat_client_from_resource(seat_resource);
+
+	wl_list_remove(wl_resource_get_link(seat_resource));
+	if (!wl_list_empty(&client->wl_resources)) {
+		return;
+	}
+
 	wlr_signal_emit_safe(&client->events.destroy, client);
 
 	if (client == client->seat->pointer_state.focused_client) {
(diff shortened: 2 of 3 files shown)


types/seat/wlr_seat.c
```diff
@@ -282,6 +307,20 @@ struct wlr_seat_client *wlr_seat_client_for_wl_client(struct wlr_seat *wlr_seat,
 	return NULL;
 }
 
+struct wl_client *wl_client_for_wlr_seat_client(
```
emersion: This is already `wlr_seat_client->client`

include/wlr/types/wlr_seat.h
```diff
@@ -13,11 +13,11 @@
  * managed by wlr_seat; some may be NULL.
  */
 struct wlr_seat_client {
-	struct wl_resource *wl_resource;
 	struct wl_client *client;
 	struct wlr_seat *seat;
 
 	// lists of wl_resource
+	struct wl_list binds;
```
emersion: I don't like a lot that there are two levels of lists. Why not create one `wlr_seat_client` per bind?

types/seat/wlr_seat.c
martinetd: Oh. That removes one thorn, I totally missed that. Thanks.

include/wlr/types/wlr_seat.h
martinetd: What do you call two levels of lists? There already are plenty of lists in wlr_seat_client.
On the other hand, creating one wlr_seat_client per bind will make us iterate over all the binds everywhere we want to notify the client about something, that's a much bigger code change that would break compositor compat (that is fine) and feels less practical to use.
Unless there's any reason to treat one bind differently from the other I'd rather keep it as simple as possible from an interface point of view

types/seat/wlr_seat.c
martinetd: (force pushed)

include/wlr/types/wlr_seat.h
emersion: I'll think about it. In the meantime, can you rename it to `resources`? We don't use the term "bind" anywhere.

martinetd: Hmm, looked around and we're not consistent at all. The one that comes up most is `wl_resources` but we have `resources`, `client_resources`... 

I personally don't like to say just resources as everything is resources, a few lines below (thanks github for showing context!!!) are "resources" to pointers, keyboards etc.

Anyway, renaming for now, but this might need a separate discussion.

BTW I think you're thinking too much, this is basically just what I said in #1023 - or did you understand something else in my comment?

emersion: Yeah, read it too quickly.

emersion:
Can you confirm it fixes firefox on wayland?

>I personally don't like to say just resources as everything is resources, a few lines below (thanks github for showing context!!!) are "resources" to pointers, keyboards etc.

Yeah, but the struct is named `wlr_seat_client`, so it makes sense that a `resource` field contains a `wl_seat`.

martinetd:
> Can you confirm it fixes firefox on wayland?

Yes it does. Well, it doesn't work great because of no popup, but there's input.

> Yeah, but the struct is named `wlr_seat_client`, so it makes sense that a `resource` field contains a `wl_seat`.

Well, sure, until it's created as `struct wlr_seat_client *client` and then `client->resource` is a `wl_seat` (thanksfully nobody tried to access that field with that naming until now, but we do have some of that naming)
Anyway, I'm just nitpicking at this point, it's more annoying to me that we have some `wl_resources` some `resources` and some others, I'll open an issue for that later

emersion:
>Yes it does. Well, it doesn't work great because of no popup, but there's input.

Good. Is the popup issue a firefox issue or a wlroots issue?

>Anyway, I'm just nitpicking at this point, it's more annoying to me that we have some wl_resources some resources and some others, I'll open an issue for that later

+1

types/seat/wlr_seat.c
```diff
@@ -109,33 +115,49 @@ static void seat_handle_bind(struct wl_client *client, void *_wlr_seat,
...
+		if (seat_client == NULL) {
+			wl_client_post_no_memory(client);
+			return;
+		}
+
+		new_seat_client = true;
 	}
-	seat_client->wl_resource =
+
+	struct wl_resource *wl_resource =
```
emersion: Maybe we can move this block before the previous one and remove `new_seat_client`?

martinetd: Good idea.
How can we cleanly undo a wl_resource_create if the calloc fails, though? just destroy it?

agx:
Ah...I've seen something similar in phosh when binding to a seat multiple times

https://source.puri.sm/Librem5/phosh/blob/master/src/phosh.c#L634

I can try the patches later this week.

types/seat/wlr_seat.c
emersion: Yeah, `wl_resource_destroy` should do it.

martinetd: okay, force pushed.

martinetd:
@agx: Cool! I'm sure the other problems with firefox-wayland are different anyway so should be separate, but this can probably wait until you've had time to test to be sure.

emersion:
Thanks!

agx:
Hi,
On Mon, Jun 04, 2018 at 06:37:08AM -0700, Dominique Martinet wrote:
>    [1]@agx: Cool! I'm sure the other problems with firefox-wayland are
>    different anyway so should be separate, but this can probably wait until
>    you've had time to test to be sure.

I did not look at the code but your patch breaks binding wl_seat
multiple times in phosh. Without your patch there's simply no input at
all, works fine with your patch applied.

emersion:
>I did not look at the code but your patch breaks binding wl_seat multiple times in phosh.

Can you elaborate?

emersion:
Ahah, got it, thanks for the feedback!

PR #1033, opened 2018-06-04
Fix mistakes in xdg_popup and xdg_popup_v6

RyanDwyer:
These are just things I spotted while staring down the code. I'm not sure how to test them, but they certainly look like mistakes.

### DIFF
--- a/types/xdg_shell/wlr_xdg_popup.c
+++ b/types/xdg_shell/wlr_xdg_popup.c
@@ -416,7 +416,7 @@ static bool xdg_popup_unconstrain_slide(struct wlr_xdg_popup *popup,
 		(popup->positioner.constraint_adjustment &
 		 XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_SLIDE_X);
 
-	bool slide_y = offset_x &&
+	bool slide_y = offset_y &&
 		(popup->positioner.constraint_adjustment &
 		 XDG_POSITIONER_CONSTRAINT_ADJUSTMENT_SLIDE_Y);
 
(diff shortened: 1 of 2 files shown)


acrisci:
These do look like mistakes, but positioner is still broken. It worked fine when I implemented this like 2 months ago.

To test this, open something with menus and try to open a menu that would go off the screen and it should either slide over, or flip to the other side of the parent menu so it appears on the screen.

acrisci:
@RyanDwyer you broke the positioner in 0f685c7e0.

This doesn't work anymore:

https://github.com/swaywm/wlroots/blob/78966417f42cee6f33139bcc653ff43efb19d3fe/rootston/xdg_shell.c#L79-L89

What was the reason for the change you made in that commit?

acrisci:
Tested positioners again and they work fine with reverting f8a8318, #1039, and this.

emersion:
>What was the reason for the change you made in that commit?

We decided it makes more sense to make boxes span [x; x + width[ and [y; y + height[ ("[" is inclusive, "]" is exclusive).

acrisci:
That makes no sense to me.

emersion:
>That makes no sense to me.

Well, that's how it works. If you have a `char arr[10]`, then `arr[0]` to `arr[9]` are valid but `arr[10]` isn't.

RyanDwyer:
Regarding `wlr_box_contains_point`, tt was brought up in chat, agreed that it was not the expected behaviour, and I happened to be the one who made the PR to change it.

For example, if you have two 1920x1080 outputs sitting side by side at 0,0 and 1920,0, and a view at position 1920,0, the view should not identify as being in the left output. So calling `wlr_box_contains_point(left output's box, 1920, 0)` should return false.

It looks like the positioner code depends on the previous behaviour of `wlr_box_contains_point`. I would think the best action is to update the positioner code to use the new behaviour rather than revert the `wlr_box_contains_point` PR.

Either way, you've confirmed the issues in this PR are indeed issues, so we should merge this one and then address the positioner separately.

acrisci:
I guess that makes sense.

PR #1034, opened 2018-06-04
wlr_linux_dmabuf: Fix printf conversion specifiers on armhf

rah2501:
The printf conversion specifiers in a call to wl_resource_post_error
do not specify the type correctly on armhf:

../types/wlr_linux_dmabuf.c: In function 'params_add':
../types/wlr_linux_dmabuf.c:104:21: error: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'uint64_t {aka long long unsigned int}' [-Werror=format=]
    "sent modifier %lu for plane %u, expected modifier %lu like other planes",
                   ~~^
                   %llu

To fix this, we use standard printf conversion specifier macros from
inttypes.h.


emersion:
Thanks!

PR #1035, opened 2018-06-04
wlr_surface_get_root_surface: walk up parent

martinetd:
This would dead-loop and never walk up if called on a subsurface.

This came up with firefox, this function would just freeze up everything at some point, if I understand the semantics of `root_surface` correctly it makes more sense to walk up to the parent.


emersion:
Thanks!

PR #1036, opened 2018-06-04
wlr_surface: walk up surfaces in reverse order

martinetd:
This displays the last surface up in front.

This is more of a monkeying patch (came up with FF-wayland as well, subsurfaces drawn for popups would be at the bottom -- FF draws a tree of surface with one root and two children, the order of the children was wrong).
Now that it's done it kind of makes sense too me, but maybe we should fix the order in which the subsurfaces are happened instead? Anyway, RFC.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -805,7 +805,7 @@ static void subsurface_handle_place_above(struct wl_client *client,
 	}
 
 	wl_list_remove(&subsurface->parent_pending_link);
-	wl_list_insert(sibling->parent_pending_link.prev,
+	wl_list_insert(&sibling->parent_pending_link,
 		&subsurface->parent_pending_link);
 
 	subsurface->reordered = true;
@@ -832,7 +832,7 @@ static void subsurface_handle_place_below(struct wl_client *client,
 	}
 
 	wl_list_remove(&subsurface->parent_pending_link);
-	wl_list_insert(&sibling->parent_pending_link,
+	wl_list_insert(sibling->parent_pending_link.prev,
 		&subsurface->parent_pending_link);
 
 	subsurface->reordered = true;
@@ -933,8 +933,8 @@ struct wlr_subsurface *wlr_subsurface_create(struct wlr_surface *surface,
 	subsurface->parent = parent;
 	wl_signal_add(&parent->events.destroy, &subsurface->parent_destroy);
 	subsurface->parent_destroy.notify = subsurface_handle_parent_destroy;
-	wl_list_insert(&parent->subsurfaces, &subsurface->parent_link);
-	wl_list_insert(&parent->subsurface_pending_list,
+	wl_list_insert(parent->subsurfaces.prev, &subsurface->parent_link);
+	wl_list_insert(parent->subsurface_pending_list.prev,
 		&subsurface->parent_pending_link);
 
 	surface->role_data = subsurface;
@@ -971,7 +971,7 @@ bool wlr_surface_point_accepts_input(struct wlr_surface *surface,
 struct wlr_surface *wlr_surface_surface_at(struct wlr_surface *surface,
 		double sx, double sy, double *sub_x, double *sub_y) {
 	struct wlr_subsurface *subsurface;
-	wl_list_for_each(subsurface, &surface->subsurfaces, parent_link) {
+	wl_list_for_each_reverse(subsurface, &surface->subsurfaces, parent_link) {
 		double _sub_x = subsurface->surface->current->subsurface_position.x;
 		double _sub_y = subsurface->surface->current->subsurface_position.y;
 		struct wlr_surface *sub = wlr_surface_surface_at(subsurface->surface,


emersion:
Hmm, one way or the other, `wlr_surface_surface_at` should iterate in the reverse order of `wlr_surface_for_each_surface`.

>maybe we should fix the order in which the subsurfaces are happened instead?

Yeah, IMHO it makes more sense for the "normal" order to be the rendering order, and the reverse order to be the input order.

martinetd:
> Hmm, one way or the other, wlr_surface_surface_at should iterate in the reverse order of wlr_surface_for_each_surface.

Eh, except that they both use the same underlying iterating functions... Should we add an additional "unwind direction" parameter?

> Yeah, IMHO it makes more sense for the "normal" order to be the rendering order, and the reverse order to be the input order.

Not clearly cut for me, especially if there are other users of the iterator that want the current "normal" order like `view_at`: it's logical for me for `view_at` to pick the first it sees, just as it's logical to draw in reverse order as we already draw overlapping windows from bottom to top.

emersion:
>Eh, except that they both use the same underlying iterating functions... Should we add an additional "unwind direction" parameter?

Right now they're both using `wl_list_for_each`, so one of them should be using `wl_list_for_each_reverse` instead. I don't think a "direction" argument is necessary.

>Not clearly cut for me, especially if there are other users of the iterator that want the current "normal" order like view_at

I don't see any use-case for iterating in input order without using `wlr_surface_at`.

>just as it's logical to draw in reverse order as we already draw overlapping windows from bottom to top.

Rootston view ordering is unrelated, it just happens to be this way because it's easier to focus a view (move it at the head of the list).

Either way, this should be specified in the header in `wlr_surface`.

martinetd:
Ah, sorry. For some reason I was convinced that `view_at` used the same iterating function. They use xdg popup only and not subsurfaces apparently? It's a bit weird to keep the same info in two places but I guess that means we don't have to worry about that here.

I tried to change the order appropriately, I'm not sure how I could test the place above/below functions though. Do we have a client that does this kind of stuff?

emersion:
>They use xdg popup only and not subsurfaces apparently?

They use the xdg-surface iterators, which iterates over xdg-popups *and* subsurfaces.

>I'm not sure how I could test the place above/below functions though. Do we have a client that does this kind of stuff?

Yeah, no Weston client actually uses it, and I'm not aware of any other client using it. Testing it would probably require an example client, I guess this could be done in another PR (we should create a follow-up issue).

emersion:
Can you change `wlr_surface_surface_at` to use `wl_list_for_each_reverse`?

martinetd:
> Can you change wlr_surface_surface_at to use wl_list_for_each_reverse?

Done (rebased & force-pushed)

I knew something was missing with `view_at` but couldn't point it out, ENOSLEEP. I'm surprised it works either way though, firefox uses subsurfaces that aren't popup in the xdg-shell sense for its menu and for the background, but click goes to the menu appropriately... Maybe I'm missing something about the background subsurface like flag to say it's not taking input? Oh, well, doesn't really matter except that it doesn't help with testing.

types/wlr_surface.c
```diff
@@ -805,7 +805,7 @@ static void subsurface_handle_place_above(struct wl_client *client,
 	}
 
 	wl_list_remove(&subsurface->parent_pending_link);
-	wl_list_insert(sibling->parent_pending_link.prev,
+	wl_list_insert(&sibling->parent_pending_link,
```
emersion: Wait, I think the `subsurface_handle_place_above` code was correct before. `place_above` means "place it above in rendering order" right?

martinetd: I naively thought above = will be drawn after in rendering order so it shows "on top" = inserted after in the list since we're walking it up the normal order.

Let's see...
```
    <request name="place_above">
      <description summary="restack the sub-surface">
        This sub-surface is taken from the stack, and put back just
        above the reference surface, changing the z-order of the sub-surfaces.
        The reference surface must be one of the sibling surfaces, or the
        parent surface. Using any other surface, including this sub-surface,
        will cause a protocol error.

        The z-order is double-buffered state, and will be applied on the
        next commit of the parent surface.
        See wl_surface.commit and wl_subcompositor.get_subsurface.

        A new sub-surface is initially added as the top-most in the stack
        of its siblings and parent.
      </description>

      <arg name="sibling" type="object" interface="wl_surface"
           summary="the reference surface"/>
    </request>
```
Not helpful for me, neither is the rest of the xml file after a very quick glance through, how do you understand that?

emersion: Our list is in the rendering order which is "natural", ie. the first item is at the bottom and the last item is at the top. So this request should place the subsurface right after the reference. `wl_list_insert(a, b)` inserts `b` right after `a`. So we should be fine if we `wl_list_insert(&sibling->parent_pending_link, &subsurface->parent_pending_link)`, ie. your code is correct?

emersion:
Thanks!

PR #1039, opened 2018-06-04
add missing anchor case in positioner invert-x for xdg-shell

acrisci:
I didn't test this because at the time we didn't have a good xdg-shell implementation.

The xdg-shell-v6 code is fine.


emersion:
Thanks1

PR #1040, opened 2018-06-04
fix wlr_box_intersection and closest_point

acrisci:
Now that box `x` + `width` is not within the box, update the intersection and closest point functions to work correctly.

There are probably a lot more off-by-one errors in the code. This fixes the positioner.


emersion:
Thanks!

PR #1043, opened 2018-06-05
drag-icon map and unmap fixes

acrisci:
The event is only emitted on unmap, so I renamed it to `unmap`. I don't like the old approach of emitting both map/unmap on the same event because it's inconsistent with the shell behavior. (I wrote this code and it just confused me).

Do we want to have a `map` event? We used to have one, right?

Reviewing to see if we need to `unmap` in other places.

### DIFF
--- a/rootston/seat.c
+++ b/rootston/seat.c
@@ -263,10 +263,10 @@ static void roots_drag_icon_handle_surface_commit(struct wl_listener *listener,
 	roots_drag_icon_damage_whole(icon);
 }
 
-static void roots_drag_icon_handle_map(struct wl_listener *listener,
+static void roots_drag_icon_handle_unmap(struct wl_listener *listener,
 		void *data) {
 	struct roots_drag_icon *icon =
-		wl_container_of(listener, icon, map);
+		wl_container_of(listener, icon, unmap);
 	roots_drag_icon_damage_whole(icon);
 }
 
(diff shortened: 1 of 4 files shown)


acrisci:
Ok, map/unmap will now additionally fire when the we go from null buffer to buffer (or reverse).

There is another bug I just noticed (on master too) where if you start the drag (before you move the pointer) it will show the drag icon in the top left corner. So maybe we are not emitting the map event correctly in this way after all.

acrisci:
```
[3367283.867]  -> wl_compositor@6.create_surface(new id wl_surface@39)
[3367283.898]  -> wl_data_device_manager@4.create_data_source(new id wl_data_source@40)
[3367283.929]  -> wl_data_source@40.offer("application/x-wayland-dnd-flower")
[3367283.953]  -> wl_data_source@40.offer("text/plain;charset=utf-8")
[3367283.976]  -> wl_data_source@40.set_actions(3)
[3367283.998]  -> wl_data_device@10.start_drag(wl_data_source@40, wl_surface@13, wl_surface@39, 90)
```

Am I mistaken, or should `wlr_surface_has_buffer()` be returning `false` for surface@39 according to the log above?

acrisci:
Ok nevermind that wasn't the problem. That function was actually returning false as expected. We just weren't updating the drag icon position at the right times in rootston.

I think this is good to go. Drag icons work now.

rootston/seat.c
```diff
@@ -260,6 +260,7 @@ static void roots_drag_icon_handle_surface_commit(struct wl_listener *listener,
 		void *data) {
 	struct roots_drag_icon *icon =
 		wl_container_of(listener, icon, surface_commit);
+	roots_drag_icon_update_position(icon);
 	roots_drag_icon_damage_whole(icon);
```
emersion: No need to damage after `roots_drag_icon_update_position`, this is already done in `roots_drag_icon_update_position`

emersion:
cc @Ongy @Timidger @ammen99: this is a breaking API change, see PR description.

emersion:
Thanks!

PR #1045, opened 2018-06-06
xdg-shells: add a set_parent signal to toplevel surfaces

ammen99:
Can be used in sway for auto-floating dialogues and similar and to properly position such transient windows relative to their parent.


emersion:
Thanks!

PR #1047, opened 2018-06-07
Fix GLES2 renderer to use glGetUniformLocations locations

NotKit:
This is needed as uniform locations are driver implementation-specific and glGetUniformLocation should be normally used according to OpenGL (ES) specification.

I tested this to work on Gemini PDA with Mali graphics proprietary driver and custom hwcomposer backend.

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -32,11 +32,37 @@ struct wlr_gles2_renderer {
 	const char *exts_str;
 
 	struct {
-		GLuint quad;
-		GLuint ellipse;
-		GLuint tex_rgba;
-		GLuint tex_rgbx;
-		GLuint tex_ext;
+		struct {
+			GLuint program;
+			GLint proj;
+			GLint color;
+		} quad;
+		struct {
+			GLuint program;
+			GLint proj;
+			GLint color;
+		} ellipse;
+		struct {
+			GLuint program;
+			GLint proj;
+			GLint invert_y;
+			GLint tex;
+			GLint alpha;
+		} tex_rgba;
+		struct  {
+			GLuint program;
+			GLint proj;
+			GLint invert_y;
+			GLint tex;
+			GLint alpha;
+		} tex_rgbx;
+		struct  {
+			GLuint program;
+			GLint proj;
+			GLint invert_y;
+			GLint tex;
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Nice

include/render/gles2.h
```diff
@@ -32,11 +32,37 @@ struct wlr_gles2_renderer {
...
+			GLint color;
+		} ellipse;
+		struct {
+			GLuint program;
+			GLint proj;
+			GLint invert_y;
+			GLint tex;
+			GLint alpha;
+		} tex_rgba;
+		struct  {
```
emersion: Style: there are two spaces here (ditto for subsequent structs)

include/render/gles2.h
```diff
@@ -32,11 +32,37 @@ struct wlr_gles2_renderer {
...
+			GLuint program;
+			GLint proj;
+			GLint color;
+		} quad;
+		struct {
+			GLuint program;
+			GLint proj;
+			GLint color;
+		} ellipse;
+		struct {
```
emersion: Since this struct is shared between `tex_rgba`, `tex_rgbx` and `tex_ext`, we can probably use the same `struct tex_shader` for these three shaders?

render/gles2/renderer.c
```diff
@@ -120,16 +120,36 @@ static bool gles2_render_texture_with_matrix(struct wlr_renderer *wlr_renderer,
...
+	GLint tex_loc = 0;
+	GLint alpha_loc = 0;
+
 	switch (texture->type) {
 	case WLR_GLES2_TEXTURE_GLTEX:
 	case WLR_GLES2_TEXTURE_WL_DRM_GL:
-		prog = texture->has_alpha ? renderer->shaders.tex_rgba :
-			renderer->shaders.tex_rgbx;
+		if (texture->has_alpha) {
+			prog = renderer->shaders.tex_rgba.program;
```
emersion: Using the same struct for these 3 would allow a more concise code:

```c
struct wlr_gles2_tex_shader *shader = NULL;

if (…) {
  shader = &renderer->shaders.tex_rgba;
}
…
```

emersion:
Thanks!

Issue #1046, opened 2018-06-06
Provide a way for compositors to keep the texture of a surface for an indefinite amount of time


ammen99:
Currently we can keep the texture only until it is re-uploaded. Having the ability to keep them for more time is useful when creating a closing animation and other things (like atomic layout in sway)

SirCmpwn:
Maybe it'd be nice to have a function which creates a new wlr_texture from an existing one and copies its GPU resources (or refcounts them if that'd be too expensive?).

emersion:
As I've said on IRC, copying a texture is expensive because it involves rendering to a pbuffer/FBO (there's no direct way to just copy it). Refcounting is not really possible because textures are mutable (for wl_shm, they're updated by wlr_surface, and for dma-buf or wl_drm, they're updated by the client).

We don't send `wl_buffer.release` at the right time in the first place. We currently sned it as soon as we create the texture. This is fine for wl_shm since we've copied the data to the GPU, but it's incorrect for wl_drm and dma-buf since the data is not copied (it's shared). We need to make sure the client doesn't re-use the wl_buffer before a new wl_buffer is attached (this _may_ explain some of our rendering/damage tracking issues).

SirCmpwn:
Do Bad Things happen if we just... don't release the texture?

emersion:
Yes, the memory is not released in the client, and the client has no choice but to allocate a new buffer at each new frame. We must manage this properly.

For this issue, what we want to basically to allow the compositor to prevent `wl_buffer.release` from being sent - and send it later. We could e.g. refcount `wl_buffer`, this would require to have something like `wlr_buffer`:

```c
struct wlr_buffer {
   struct wl_buffer wl_buffer;
   size_t refs;
   struct wlr_texture *texture; // destroyed when wl_buffer.release is sent
};
```

emersion:
Oh, and `wlr_buffer` might allow use to get rid of the `wl_buffer` type switch in `wlr_surface` - by making protocol interfaces that create `wl_buffer` responsible for providing a way to get a texture from the resource. That would be pretty cool.

emersion:
More detailed constraints:

* For wl_shm, we need to release the wl_buffer when the pixels have been uploaded to the GPU.
* For dma-buf, we need to release the wl_buffer when a new buffer is committed - unless the compositor still holds a reference to the texture.
[...]

PR #1048, opened 2018-06-07
surface: don't release wl_drm and linux-dmabuf buffers early

emersion:
For wl_drm and linux-dmabuf buffers, release now happens in `surface_move_state`.

Test plan: start mpv with `WAYLAND_DEBUG=1`.

Background: https://github.com/swaywm/wlroots/issues/1046#issuecomment-395433065


SirCmpwn:
Thanks!

PR #1051, opened 2018-06-09
Fix atti assert in wlr_egl_init

RedSoxFan:
If `request_high_priority` is `true`, then `atti` will be `5` in the assert. Since `5` is not less than `5` (`sizeof(attribs)/sizeof(attribs[0])`), the assert fails and sway does not start.

This changes the assert to be less than or equal to.


emersion:
Thanks!

PR #1053, opened 2018-06-10
Add xdg-decoration-unstable-v1 support

emersion:
This is a redo of #638, rebased against master and updated to use the latest v6 proposal (https://lists.freedesktop.org/archives/wayland-devel/2018-May/038206.html).

This adds support for xdg-toplevel-decoration in rootston. The preferred mode is "client". rootston will accept the client's mode requests.

A new example client is added. When launched, it will be yellow if using server-side decorations and green if using client-side decorations. The client's initial preferred mode can be given as an argument (either "client" or "server"). The client's preferred mode can be toggled by clicking on its surface.

Test plan: try the example

Follow-up refactoring: use the new `configure` and `ack_configure` events for xdg-toplevel and xdg-popup.


SirCmpwn:
That doesn't surprise anyone and doesn't change our plans.

Sunderland93:
Anyway, I'm very glad to see this protocol! You did amazing job!

emersion:
wayland-protocols have just been updated in the repos, so this is ready for review.

SirCmpwn:
What's the status on client implementations?

emersion:
>What's the status on client implementations?

We have a test client. Smithay have an opened PR, but they want to test on a compositor which supports it. I've been trying to work on the GTK implementation, but it's very messy (no, just using the previous work on the KDE protocol doesn't work). I haven't heard about anything wrt. Qt and KDE.

SirCmpwn:
Nice work, looks good

emersion:
Ping

```
johanhelsing | are there any test clients/compositors for xdg-decoration yet?
```

SirCmpwn:
Thanks!

Issue #233, opened 2017-10-07
xdg-popup: don't swallow the button event on ungrab


acrisci:
To reproduce, open two windows. Open a popup on one of the windows (like a right click menu) and now click on the other window. You will see that the popup is dismissed, but the other window does not get focused. This is because the grab "swallows" the button event to dismiss the popup so the other window never gets the button press. Better might be if the button event went through to the other window so the compositor can focus the window and the window can process the click.

Weston and i3 has the former behavior. It might be a design decision, or maybe it's just an artifact of our legacy. Windows(tm) has the behavior that I am suggesting here.

I guess it's a matter of opinion on what is correct here. Maybe we should implement this in such a way that compositor implementations can set their own policy.

SirCmpwn:
I agree with your proposed behavior. We should dismiss the popup and send the click to the other window.

SirCmpwn:
via GenkiSky on IRC

  <p>
    This is for wlroot's <a href="https://github.com/swaywm/wlroots/issues/233">Issue #233</a>.
    <em>Summary</em>: If you have two windows open, then open a XDG popup on one, and
    left-click on the other, the newly-clicked window will not receive keyboard
    input. This is especially noticeable as the newly-clicked window
    receives XDG activation, despite keyboard input still going to the old window.
  </p>

  <p>
    The high level cause is that the popup's keyboard grab ignores enter() events,
    which are given when rootston is trying to change focus. Here are some
    options for solutions.
  </p>

  <h3>Option A: Both close the popup menu, and change focus to the other window.</h3>
  <ul>
    <em>Benefits:</em>
      <ul>
        <li>This behavior is what is expected by some, in discussion on <a
            href="https://github.com/swaywm/wlroots/issues/233">Issue #233</a>.
        <li>It's a simpler mental model that you, as the compositor, can
          just cancel the keyboard grab at any point, without worry.
        <li>At least for one of the below patch options, no API change is needed.
      </ul>

    <em>Patch options:</em>
      <ul>
        <li><a href="impl_cancel.patch">impl_cancel.patch</a>: When changing
          focus, each keyboard-grab impl must explicitly cancel themselves.
        <li><a href="rootston_cancel.patch">rootston_cancel.patch</a>: When
          changing focus, wlroots library-user can force-cancel the keyboard-grab.
        <li><a href="default_cancel.patch">default_cancel.patch</a>: When changing
          focus, have all keyboard-grab impls cancel themselves, unless they
[...]

PR #1054, opened 2018-06-11
rootston: Cancel existing keyboard grab when changing focus

SirCmpwn:
This patch fixes issue #233. As described there, if you have two
windows open, then open a XDG popup on one, and left-click on the
other, the newly-clicked window will not receive keyboard input. This
is especially jarring as it still receives XDG activation.

The cause is that when trying to change our seat's focus, an existing
keyboard grab can ignore our attempt. So, we cancel existing grabs
before doing so. The commit message has lower-level details.

It's worth noting two alternative approaches that were discussed with
@acrisci on the IRC channel:

(A) Change each wlr_pointer_grab_interface's enter() implementation
to cancel itself and call wlr_seat_keyboard_notify_enter(), if it
expects to do nothing. However, it is messy to change every interface.

(B) Change wlroots to behave like weston, where a click on the new
window will neither set it active nor send it keyboard input. This
also involves changing every interface to return whether it is a
no-op, which is messy. And, this would be counter-intuitive behavior,
according to discussion on IRC and issue #233.


rootston/seat.c
```diff
@@ -825,6 +825,11 @@ void roots_seat_set_focus(struct roots_seat *seat, struct roots_view *view) {
 	view_activate(view, true);
 	seat->has_focus = true;
 
+	// We want to unconditionally send keyboard input to the view we are
+	// focusing here, so cancel any existing grabs.
+	struct wlr_seat_keyboard_grab *curr_grab = seat->seat->keyboard_state.grab;
```
acrisci: probably better to use `wlr_seat_keyboard_end_grab()`.

acrisci:
Yeah, this is probably the most explicit and simple way to do it. The downside is it requires the library user to manually manage grabs in this case. It would be better if the grab could do this for you, but the downside of that is that now you have to explicitly have a rule of "all keyboard grabs shall end their grab when you send keyboard_notify_enter to another surface". That's probably safe to do though and I'm not necessarily against it.

acrisci:
What's the status of this? Can I just commit the change I wanted and merge?

SirCmpwn:
Sure.

PR #1059, opened 2018-06-13
WiP: Add get_gamma to gamma-control

agx:
This adds get_gamma to get the current gamma lut to wlr_gamma_control.

TODOs:

- [ ] figure out what to do with gamma-control protocol. It's not marked as unstable but e.g. has no docs and isn't in wlr-protocols or even wayland-protocols either. Is if fit for wlroots 1.0? Idea would be to rename to wlr-gamma-control and mark it as unstable. Old gamma-control could be kept for backward compat but are there any users atm?
- [ ]  don't always use legacy drm interface

Test Plan
- [ ] try examples/gamma
- [ ] Use phosh's gamma branch (https://source.puri.sm/Librem5/phosh/merge_requests/58) and see gnome nightlight work / monitor the interataction via dbus-monitor


emersion:
Why do you need to read gamma?

Yeah, the gamma protocol is one of these Orbital protocols, so just like KDE protocols they don't have any stability promise and don't follow wayland-protocols conventions. It's used by the Wayland branch of redshift.

agx:
@emersion see above: the gnome-desktop library uses it (gnome_rr_crtc_get_gamma) and that again is used by gnome-settings-daemon (which is e.g. responsible for nightlight aka redshift) among other things.

emersion:
Sure, but why do these need to read gamma?

SirCmpwn:
I'm guessing it's for showing the user what the current gamma is so you can set some sliders or something up

agx:
Hi,
On Wed, Jun 13, 2018 at 11:13:33AM -0700, Drew DeVault wrote:
>    I'm guessing it's for showing the user what the current gamma is so you
>    can set some sliders or something up

Yeah. Most of the clients just use it to get the size of the lut but
since it's in a frequently used library I can't just ignore that part of
the API.

emersion:
>I'm guessing it's for showing the user what the current gamma is so you can set some sliders or something up

I don't think clients reverse the gamma lut to get back values for their sliders. Usually the workflow for redshift-like clients is:
* Get a temperature from a user preference or from the current time
* With this temperature, compute gamma luts
* Send gamma luts to the compositor

A client should never need to read the gamma luts. If it lets the user set something with a slider, it needs to save it somewhere anyway.

agx:
Well it apparently can be done on X and there's nothing wrong with allowing it IMHO and some abstractions like gnome-desktop expect it. I could mock that up in phosh to make these clients not fail but reporting sane values is IMHO better.

agx:
>Also, this protocol needs work. For instance:
> What happens if two clients want to manage gamma luts at the same time?
> What happens when the client disconnects or crashes?

Are you refering to gamma-control in general or the parts of the PR here? We're talking about reading here so two clients at the same time won't interfere or am I missing s.th.? Writing is a different matter but that's not part of the PR. IMHO Not every client should be allowed to bind as with surface layers, input methods and other things.

agx:
I'm not arguing along the lines of X11 but rather that if we allow writing we should allow readind as well.

emersion:
>Are you refering to gamma-control in general or the parts of the PR here?

No this comment ("this protocol needs work") was about the protocol in general, unrelated to this PR.

>I'm not arguing along the lines of X11 but rather that if we allow writing we should allow readind as well.

Unless clients have a good reason to read gamma lut, I wouldn't add it to the protocol. If there isn't a real use-case for this, I don't see why it should be part of the protocol (being compatible with the internal GNOME D-Bus protocol isn't a use-case).

SirCmpwn:
@subdiff may have some comments on this

subdiff:
Don't know much about this specific implementation here, but for Night Color I only send a temperature value to KWin and then KWin calculates the gamma ramp from the linear default case with a suitable multiplication as redshift does it.

For gamma control (not yet merged) the workspace needs to know current red, green, blue values of an output, but these are _not_ the ones received through DRM interface (since these have been additionally multiplied with temperature value of Night Color).

PR #1060, opened 2018-06-13
surface: remove wlr_surface.texture

emersion:
The texture is managed by the surface's wlr_buffer now. In
particular, the buffer can destroy the texture early if it becomes
invalid.

cc @ongy @Timidger @ammen99: this is a breaking change.

### DIFF
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -8,8 +8,15 @@
  * A client buffer.
  */
 struct wlr_buffer {
-	struct wl_resource *resource; // can be NULL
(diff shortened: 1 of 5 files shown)


Ongy:
> the buffer can destroy the texture early if it becomes invalid

Can you elaborate on this? I had seen that it's available over two paths now, and thought it's just for backwards compat naming.

Is there a semantic difference between the `wlr_buffer::texture` and the previous `wlr_surface::texture`?

Ongy:
ah, sorry just read the code comments in this PR, this makes it clear

emersion:
This has been merged into #1062.

PR #1063, opened 2018-06-14
Multiseat fixes

### LINKED ISSUE #1057: Support XDG_SEAT
* https://www.freedesktop.org/wiki/Software/systemd/multiseat/
* Weston patch: https://lists.freedesktop.org/archives/wayland-devel/2018-January/036628.html

ascent12:
Fixes #1057 

Various little changes to get multi-seat working properly. We were already 99% of the way there.

To test:
- Two GPUs are required
- Another keyboard and mouse helps too, but is not strictly required
- Assign all of the devices to another seat with `loginctl assign seat1 /sys/path/to/device`

logind:
- Start a display manager (I used GDM); it will appear on both seats
- Log in to both seats
There is a limitation with GDM where it doesn't seem to want to start a wayland session on seat1 (the option is not given).
I tested this successfully with sway on seat0 and i3 on seat1.

direct:
GDM doesn't seem to like the direct session. It's probably related to their hard dependency on logind, and I don't care enough to look into it.
- From seat0, run `XDG_SEAT=seat1 sway`
- From seat0, change VT and run `sway`

### DIFF
--- a/backend/session/direct.c
+++ b/backend/session/direct.c
@@ -79,7 +79,7 @@ static bool direct_change_vt(struct wlr_session *base, unsigned vt) {
 	struct direct_session *session = wl_container_of(base, session, base);
 
 	// Only seat0 has VTs associated with it
-	if (strcmp(session->base.seat, "seat0") == 0) {
+	if (strcmp(session->base.seat, "seat0") != 0) {
 		return true;
 	}
 
--- a/backend/session/direct.c
+++ b/backend/session/direct.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include <errno.h>
+#include <fcntl.h>
 #include <linux/input.h>
 #include <linux/kd.h>
 #include <linux/major.h>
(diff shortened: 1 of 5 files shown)


backend/session/direct.c
```diff
@@ -76,30 +77,39 @@ static void direct_session_close(struct wlr_session *base, int fd) {
 
 static bool direct_change_vt(struct wlr_session *base, unsigned vt) {
 	struct direct_session *session = wl_container_of(base, session, base);
+
+	// Only seat0 has VTs associated with it
+	if (strcmp(session->base.seat, "seat0") == 0) {
```
emersion: So this _doesn't_ switch VTs if the user is on `seat0`?

ascent12: That was a silly mistake.

n3rdopolis:
I know it's already merged, but I have tested this on qemu, and can confirm that this does work.  
Here is the qemu system I used to test multiseat
```
qemu-system-x86_64 -m 2047 -cdrom ~/my_built.iso -vga none -enable-kvm \
-device VGA,id=video0  -device secondary-vga,id=video1 \
-spice port=5930,disable-ticketing -cpu host \
-device ich9-usb-ehci1,id=usb \
-device ich9-usb-uhci1,masterbus=usb.0,firstport=0,multifunction=on \
-device ich9-usb-uhci2,masterbus=usb.0,firstport=2 \
-device ich9-usb-uhci3,masterbus=usb.0,firstport=4 \
-chardev spicevmc,name=usbredir,id=usbredirchardev1 \
-device usb-redir,chardev=usbredirchardev1,id=usbredirdev1 \
-chardev spicevmc,name=usbredir,id=usbredirchardev2 \
-device usb-redir,chardev=usbredirchardev2,id=usbredirdev2 \
-chardev spicevmc,name=usbredir,id=usbredirchardev3 \
-device usb-redir,chardev=usbredirchardev3,id=usbredirdev3 
```
(have to use USB redirected devices, for the second mouse and keyboard so QEMU handles them separately )

PR #1066, opened 2018-06-16
layer-shell: check if the surface is mapped in layer_surface_destroy()

ammen99:
If the layer surface has been closed by the compositor, using `layer_surface_close()`, then the unmap event is emitted immediately. However, when the layer surface is later destroyed by the client, the compositor used to get a second unmap, which is fixed with this commit.

As discussed in IRC, there could be another way to fix this - by not emitting the unmap on in `layer_surface_close()`. However, there is still the possibility the client first unmaps by committing a `NULL` buffer and then destroying the surface, so the change in this PR is needed anyway.


emersion:
Thanks!

PR #1067, opened 2018-06-17
surface: fix double wl_buffer.release events

emersion:
Prior to this commit, we re-uploaded the buffer even if a new one
wasn't attached. After uploading, we send wl_buffer.release. So,
this sequence of requests resulted in a double release:

    surface.attach(buffer, 0, 0)
    surface.commit()
    <- buffer.release()
    surface.commit()
    <- buffer.release()

Test plan: open Gedit, click on a button to open a popup. Crashes Gedit before this commit with:

```
(gedit:28185): Gdk-WARNING **: 12:33:31.056: (gdkwindow-wayland.c:797):buffer_release_callback: runtime check failed: (impl->staging_cairo_surface != cairo_surface)
```

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -325,19 +325,12 @@ static void surface_damage_subsurfaces(struct wlr_subsurface *subsurface) {
 	}
 }
 
-static void surface_apply_damage(struct wlr_surface *surface,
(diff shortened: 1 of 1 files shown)


atomnuker:
Tested, works perfectly

PR #1068, opened 2018-06-18
Drop all mastered DRM fds on exit

tobiasblass:
The helper process, that handles privileged DRM operations for the
parent process, now remembers all file descriptors it acquired DRM
mastership over. This allows it to release them properly on
termination, even if the parent process crashes.

This fixes an issue where the virtual console locks up if sway crashes or if I kill it with meta+shift+e (at least on my system).

### DIFF
--- a/backend/session/direct-ipc.c
+++ b/backend/session/direct-ipc.c
@@ -130,10 +130,10 @@ static void communicate(int sock) {
 	struct msg msg;
 	int drm_fd = -1;
 	bool running = true;
-	fd_set masteredFds;
-	int maxFd = -1;
+	fd_set mastered_fds;
+	int max_fd = -1;
 
-	FD_ZERO(&masteredFds);
+	FD_ZERO(&mastered_fds);
 	while (running && recv_msg(sock, &drm_fd, &msg, sizeof(msg)) >= 0) {
 		switch (msg.type) {
 		case MSG_OPEN:
@@ -163,8 +163,8 @@ static void communicate(int sock) {
 			}
 error:
 			if (ret == 0) {
-				FD_SET(fd, &masteredFds);
-				maxFd = fd > maxFd ? fd : maxFd;
+				FD_SET(fd, &mastered_fds);
+				max_fd = fd > max_fd ? fd : max_fd;
 			} else {
 				close(fd);
 			}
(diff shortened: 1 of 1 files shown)


backend/session/direct-ipc.c
```diff
@@ -181,7 +191,12 @@ static void communicate(int sock) {
 			break;
 		}
 	}
-
+	for (int fd=0; fd<maxFd; fd++) {
```
SirCmpwn: style: add spaces around operators

SirCmpwn:
LGTM

backend/session/direct-ipc.c
```diff
@@ -129,7 +130,10 @@ static void communicate(int sock) {
 	struct msg msg;
 	int drm_fd = -1;
 	bool running = true;
+	fd_set masteredFds;
```
emersion: Style: use underscores, no camel case

backend/session/direct-ipc.c
```diff
@@ -129,7 +130,10 @@ static void communicate(int sock) {
 	struct msg msg;
 	int drm_fd = -1;
 	bool running = true;
+	fd_set mastered_fds;
```
ascent12: I'm not a fan of using an fd_set for this, considering the fairly low maximum file descriptor (usually 1024). It's not particularly likely that we'll reach this number, but it's not the kind of limitation I would want to add to library code.

backend/session/direct-ipc.c
```diff
@@ -158,18 +162,24 @@ static void communicate(int sock) {
 				ret = errno;
 			}
 error:
+			if (ret == 0) {
```
ascent12: This function also handles opening input devices, so you're remembering and later trying to drmDropMaster them too.

backend/session/direct-ipc.c
tobiasblass: Ah, I thought that the limit on fd_set was also the largest possible file descriptor. You're right that it isn't, though, and there is a remark about it in the BUGS section of the fd_set manpage.
I'll change it to a wl_array.

backend/session/direct-ipc.c
tobiasblass: I misread the ENOTSUP if-statement above. I'll make the remembering conditional on DRM_MAJOR.

tobiasblass:
While working on the changes proposed by @ascent12 I noticed that the issue might be related to the Xwayland backend. Since I didn't open any X windows when testing the patch it looked to me like the issue was fixed.
I'll close the push request for now and investigate further. Sorry for not testing the patch more thoroughly

tobiasblass:
For reference: The virtual console lockup happens due to a segmentation fault in sway (https://github.com/swaywm/sway/issues/2149). I'm not sure why this segmentation fault causes a console lock up, but googling around suggests it could be a bug in the graphics driver.

emersion:
>I'm not sure why this segmentation fault causes a console lock up, but googling around suggests it could be a bug in the graphics driver.

From what I've seen, a compositor segmentation fault leaves the session child behind and doesn't cleanup properly.

tobiasblass:
> > I'm not sure why this segmentation fault causes a console lock up, but googling around suggests it > > could be a bug in the graphics driver.
>
> From what I've seen, a compositor segmentation fault leaves the direct session child behind and doesn't cleanup properly.

If you're referring to #1073, this doesn't fix it. I even tried terminating sway from another virtual console using swaymsg, made sure that no sway process was alive any more and used fuser to make sure that /dev/dri/card0 was not opened by anyone. The system still locked up when I changed back to the virtual console where I started sway.

PR #1070, opened 2018-06-19
[WIP] Remove usages of wlr_output_layout_get()

### LINKED ISSUE #812: Make wlr_output_layout_get_box take a box as parameter
Returning a box is strange and inconsistent.

If we choose to return a box, we could at least return a `const`.

### LINKED ISSUE #1013: Make `wlr_output_layout_output` private
The `wlr_output_layout_output` was designed to be private to the `wlr_output_layout`. There's no reason to expose it to user code.

Looking through the rootston code, it looks like the only use for it is to get the layout x/y coordinates of the output in the layout. This can be accomplished much more cleanly with `wlr_output_layout_get_box()` now.

Is there any use case for exposing the `wlr_output_layout_output` to user code?

acrisci:
Getting the box works just as good.

Fixes #1013
Fixes #812

What do you think about changing `wlr_output_layout_get_box()` to accept a reference to the box as an output variable instead of returning a pointer? That seems to be what we've done in other places.

### DIFF
--- a/include/types/wlr_output_layout.h
+++ b/include/types/wlr_output_layout.h
@@ -3,27 +3,21 @@
 
 #include <wlr/types/wlr_output_layout.h>
 
-struct wlr_output_layout_output_state {
+struct wlr_output_layout_output {
 	struct wlr_output_layout *layout;
-	struct wlr_output_layout_output *l_output;
-
+	struct wlr_output *output;
+	int x, y;
+	struct wl_list link;
 	bool auto_configured;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
 
 	struct wl_listener mode;
 	struct wl_listener scale;
 	struct wl_listener transform;
 	struct wl_listener output_destroy;
-};
-
-struct wlr_output_layout_output {
-	struct wlr_output *output;
-	int x, y;
-	struct wl_list link;
-	struct wlr_output_layout_output_state *state;
-
-	struct {
-		struct wl_signal destroy;
-	} events;
 };
 
 struct wlr_output_layout_output *wlr_output_layout_get(
--- a/include/types/wlr_output_layout.h
+++ b/include/types/wlr_output_layout.h
@@ -1,0 +1,32 @@
+#ifndef TYPES_WLR_OUTPUT_LAYOUT_H
+#define TYPES_WLR_OUTPUT_LAYOUT_H
+
+#include <wlr/types/wlr_output_layout.h>
+
+struct wlr_output_layout_output_state {
+	struct wlr_output_layout *layout;
+	struct wlr_output_layout_output *l_output;
+
+	bool auto_configured;
+
+	struct wl_listener mode;
+	struct wl_listener scale;
+	struct wl_listener transform;
+	struct wl_listener output_destroy;
+};
+
+struct wlr_output_layout_output {
+	struct wlr_output *output;
+	int x, y;
+	struct wl_list link;
+	struct wlr_output_layout_output_state *state;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+};
+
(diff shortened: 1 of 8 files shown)


Ongy:
> What do you think about changing wlr_output_layout_get_box() to accept a reference to the box as an output variable instead of returning a pointer?
yes. looking at the current code, I think it's non-reentrant.
While it's not a huge problem with wlroots being single-threaded in general, it's not a good idea either.

emersion:
>What do you think about changing wlr_output_layout_get_box() to accept a reference to the box as an output variable instead of returning a pointer? That seems to be what we've done in other places.

+1, see #812

emersion:
What about actually removing `wlr_output_layout_get` and making `wlr_output_layout_output` private?

SirCmpwn:
>What do you think about changing wlr_output_layout_get_box() to accept a reference to the box as an output variable instead of returning a pointer? That seems to be what we've done in other places.

+1

include/types/wlr_output_layout.h
```diff
@@ -0,0 +1,32 @@
+#ifndef TYPES_WLR_OUTPUT_LAYOUT_H
+#define TYPES_WLR_OUTPUT_LAYOUT_H
+
+#include <wlr/types/wlr_output_layout.h>
+
+struct wlr_output_layout_output_state {
```
acrisci: Any reason to keep this around?

types/wlr_cursor.c
```diff
@@ -270,13 +276,17 @@ void wlr_cursor_absolute_to_layout_coords(struct wlr_cursor *cur,
 		double *lx, double *ly) {
 	assert(cur->state->layout);
 
-	struct wlr_box *mapping = get_mapping(cur, dev);
-	if (!mapping) {
-		mapping = wlr_output_layout_get_box(cur->state->layout, NULL);
+	struct wlr_box mapping;
+	if (!get_mapping(cur, dev, &mapping) &&
+			!wlr_output_layout_get_box(cur->state->layout, NULL, &mapping)) {
+		// no outputs in the layout
```
acrisci: dunno what to do here.

include/types/wlr_output_layout.h
emersion: No, I think the whole point of this state struct was to keep some stuff private.

types/wlr_cursor.c
emersion: I think setting everything to zero is sane.

types/wlr_xdg_output.c
```diff
@@ -6,6 +6,7 @@
 #include <wlr/types/wlr_xdg_output.h>
 #include <wlr/util/log.h>
 #include "xdg-output-unstable-v1-protocol.h"
+#include "types/wlr_output_layout.h"
```
emersion: Hmmm. I'm not a fan of this.

acrisci: Yeah, this interface can probably just work with the `wlr_output`. `wlr_cursor.c` uses it too.

include/types/wlr_output_layout.h
```diff
@@ -0,0 +1,26 @@
...
+		struct wl_signal destroy;
+	} events;
+
+	struct wl_listener mode;
+	struct wl_listener scale;
+	struct wl_listener transform;
+	struct wl_listener output_destroy;
+};
+
+struct wlr_output_layout_output *wlr_output_layout_get(
```
emersion: This function will be exported because it has the `wlr_` prefix. If it's in a private header, it shouldn't.

acrisci:
If this is private, we need a new way to iterate over the outputs in the layout.

emersion:
>make wlr_output_layout_output private

Can this commit be removed for now so that this PR can br merged?

panaman67:
bump @acrisci @emersion

PR #1073, opened 2018-06-20
FIX: Suprocess loops endlessly when the control socket closes.

tobiasblass:
recvmsg(3) returns 0 if the connection partner has shut down its socket.
The communicate function considered 0 a successful message, though, and
keeps calling recvmsg(3) again and again.


SirCmpwn:
Minor nitpick: a better commit message would have been "Fix infinite loop when control socket closes".

Issue #1065, opened 2018-06-14
Allow compositors to actually access the surface->buffer on unmap


ammen99:
Once #1062 is done, compositors will again be able to keep the buffer/texture of a `wlr_surface` longer than the `wlr_surface` itself, of course if this is possible at all. One of the use cases for that is window closing animation. However, currently compositors can't access the buffer on unmap, if the given app unmaps by committing a `NULL` buffer, because the pending state is applied before emitting the map/unmap signals (these come from the `role_committed` callback of the various shells). An example of such clients are many Xwayland apps.

Now, to let users access the previous `surface->buffer` on unmap when a `NULL` buffer is committed, there must be a way to signal the different shells before committing the new buffer, so that they can emit the unmap event on time. A possible solution could be to add a new `role_precommit` callback or somehow pass the old state to the existing `role_commit`.

emersion:
The more general issue is that `commit` handlers need to access both the old state and the new one. For instance, if the surface size changes, you'll need to read both the old size and the new size to properly damage outputs (this is currently done by adding the biggest size to the surface damage in `wlr_surface` itself, which is kind of a hack I guess?).

emersion:
We could also add a `wlr_surface.previous` state that is copied from `wlr_surface.current` right before commit.

ammen99:
> The more general issue is that commit handlers need to access both the old state and the new one. For instance, if the surface size changes, you'll need to read both the old size and the new size to properly damage outputs (this is currently done by adding the biggest size to the surface damage in wlr_surface itself, which is kind of a hack I guess?).

I didn't know that was done, but it seems like a good thing. I expect that all compositors will do this anyway, so adding it here makes life easier for the compositor.

emersion:
>I expect that all compositors will do this anyway, so adding it here makes life easier for the compositor.

Yeah, but it's a lie, and makes some things more complicated. For instance, the damage region can have extents bigger than the buffer size - so we need to crop it back when updating wl_shm textures. Also, it's unclear how the buffer position is handled - in fact I don't think it's supported at all right now. Luckily nobody seems to be using it.

All in all, we should probably keep states as the client set them - and maybe add `wlr_surface.damage` with the resize behaviour for convenience.

ammen99:
> All in all, we should probably keep states as the client set them - and maybe add wlr_surface.damage with the resize behaviour for convenience.

[...]

PR #1076, opened 2018-06-20
surface: redesign state

emersion:
- [x] Rename `invalid` to `committed` - better ideas welcome!
- [x] Move `subsurface_position` in `wlr_subsurface`
- [x] Fix `sx`/`sy` handling
- [x] Don't monkey-patch `surface_damage` and `buffer_damage` - add a new buffer damage region instead
- [x] Make `current` and `pending` embedded structs
- [x] Add a `previous` state
- [x] Update cursor hotspot with buffer position

Test plan:
- Make sure damage tracking works on surfaces and subsurfaces (try gnome-calculator)
- Try with scaled and transformed outputs
- Try [wleird](https://github.com/emersion/wleird)'s cursor example

See #1065


ammen99:
> Remove subsurface_position

Is there another way to get the subsurface position relative to its parent?

emersion:
>Is there another way to get the subsurface position relative to its parent?

Yes, this will be moved in `wlr_subsurface`.

include/wlr/types/wlr_surface.h
```diff
@@ -8,57 +8,32 @@
...
-#define WLR_SURFACE_INVALID_OPAQUE_REGION 8
-#define WLR_SURFACE_INVALID_INPUT_REGION 16
-#define WLR_SURFACE_INVALID_TRANSFORM 32
-#define WLR_SURFACE_INVALID_SCALE 64
-#define WLR_SURFACE_INVALID_SUBSURFACE_POSITION 128
-#define WLR_SURFACE_INVALID_FRAME_CALLBACK_LIST 256
+enum wlr_surface_state_field {
+	WLR_SURFACE_STATE_BUFFER = 1,
+	WLR_SURFACE_STATE_SURFACE_DAMAGE = 2,
+	WLR_SURFACE_STATE_BUFFER_DAMAGE = 2,
```
synaptiko: Should it be `4` here?

emersion: Right, thanks!

emersion:
I'm not really happy with the current `wlr_surface_state` struct. It contains both things from the client (fields from `buffer` to `frame_callback_list`) accumulated before commit and things computed on commit (`width`, `height`, `s{x,y}`, and I wanted to add the `wlr_buffer`?).

emersion:
Still not sure how damage should be defined.
* What should be the (0, 0) point in the damage region? The top-left corner of the buffer or the top-left corner of the surface?
* Should it be in buffer-local coords or in surface-local coords? We loose precision if we do surface-local coords, and it's doesn't add a _lot_ of complexity in the damage code - just one `/ surface->current.scale` somewhere, and one `wlr_transform_compose(…, surface->current.transform)` somewhere else.

ammen99:
@emersion, as we're doing the `damage` for just convenience, I think it should be in surface local coordinates, because that's 
what a compositor typically uses. In this case it'll also make sense to make (0,0) the top left corner of the surface. I'm however not sure how to do precision issues, maybe `wlr_region_expand`?

SirCmpwn:
It should be buffer local.

ammen99:
> It should be buffer local.

I'm not trying to argue, just asking out of interest: what makes you think this way?

SirCmpwn:
Can we just go with "it's been a long day and explaining the full answer is beyond the limits of my patience right now"?

emersion:
Moved bits that are computed as a delta of two states (the enhanced damage, the surface position, the buffer itself) outside of `wlr_surface_state`.

Decided to use damage in buffer-local coordinates, so that you can apply scale and transform to this region in the exact same way you render the associated texture.

emersion:
This PR is ready for review.

ammen99:
So, after this, what is the intended solution for #1065?

emersion:
>So, after this, what is the intended solution for #1065?

Let's continue to discuss this in the other issue.

SirCmpwn:
https://github.com/swaywm/sway/pull/2185

Code looks good, running this as my daily driver for a while.

atomnuker:
Pointer position is still off after applying this and the sway update PR. Do you compensate for scaling?

emersion:
>Pointer position is still off after applying this and the sway update PR. Do you compensate for scaling?

Input doesn't deal with scaling and transform. Can you describe how to reproduce? Are you sure you're running with the latest version of this branch? (A previous commit fucked up the input region)

atomnuker:
Yep, this is with the latest version of this PR. There's no transform on anything, just scaling set in sway.
You've reproduced this before and I have mentioned this before too. Hotspot is fine, the image matches what's on the screen and sent as input, its just that if you open a right click menu its offset compared to the pointer location.

emersion:
FYI I have input issues with some floating windows on Sway master (without this patch applied): https://github.com/swaywm/sway/issues/2195

atomnuker:
In GTK - a few pixels (3 or 4) upwards and a pixel or two to the left of where the menu is created.
In xwayland (such as firefox), 6 or 7 pixels upwards (or downwards, if you try to create the menu too far down the screen) and about 1 or two pixels to the left.
Yes, this isn't a regression, its always been present, but since this commit touched some position code I thought it might also fix this issue.

emersion:
>Yes, this isn't a regression, its always been present, but since this commit touched some position code I thought it might also fix this issue.

I think this is more of a xdg-shell window geometry issue (this PR only updates the wl_surface implementation).

emersion:
Bump: I've been using this for a few days.

SirCmpwn:
aight

PR #1084, opened 2018-06-24
use-after-free fixes (xdg_shell popups, primary selection source, xwm parents)

martinetd:
see individual commits:
 - popups and their parent surface (happens with submenus in konsole)
 - primary selection (happens when overwriting selection in e.g. urxvt)
 - xwm parents/children (happens when closing gimp)

### DIFF
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -415,6 +415,12 @@ void destroy_xdg_surface(struct wlr_xdg_surface *surface) {
 
 	wlr_signal_emit_safe(&surface->events.destroy, surface);
 
+	struct wlr_xdg_popup *popup_state, *next;
+	wl_list_for_each_safe(popup_state, next, &surface->popups, link) {
+		xdg_popup_send_popup_done(popup_state->resource);
+		destroy_xdg_popup(popup_state->base);
+	}
+
 	switch (surface->role) {
 	case WLR_XDG_SURFACE_ROLE_TOPLEVEL:
 		destroy_xdg_toplevel(surface);
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -294,6 +294,12 @@ static void xwayland_surface_destroy(
 	wl_list_remove(&xsurface->link);
 	wl_list_remove(&xsurface->parent_link);
 
+	struct wlr_xwayland_surface *child, *next;
+	wl_list_for_each_safe(child, next, &xsurface->children, parent_link) {
+		wl_list_remove(&child->parent_link);
+		wl_list_init(&child->parent_link);
+	}
+
 	if (xsurface->surface_id) {
 		wl_list_remove(&xsurface->unpaired_link);
 	}
--- a/types/wlr_primary_selection.c
+++ b/types/wlr_primary_selection.c
@@ -229,9 +229,9 @@ void wlr_seat_set_primary_selection(struct wlr_seat *seat,
 	}
 
 	if (seat->primary_selection_source) {
+		wl_list_remove(&seat->primary_selection_source_destroy.link);
 		seat->primary_selection_source->cancel(seat->primary_selection_source);
 		seat->primary_selection_source = NULL;
-		wl_list_remove(&seat->primary_selection_source_destroy.link);
 	}
 
 	seat->primary_selection_source = source;
--- a/types/xdg_shell_v6/wlr_xdg_surface_v6.c
+++ b/types/xdg_shell_v6/wlr_xdg_surface_v6.c
@@ -100,6 +100,12 @@ void destroy_xdg_surface_v6(struct wlr_xdg_surface_v6 *surface) {
 
 	wlr_signal_emit_safe(&surface->events.destroy, surface);
 
+	struct wlr_xdg_popup_v6 *popup_state, *next;
+	wl_list_for_each_safe(popup_state, next, &surface->popups, link) {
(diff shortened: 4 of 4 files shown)


SirCmpwn:
Can we actually get away with destroying resources at this point?

acrisci:
How do you reproduce this?

martinetd:
> Can we actually get away with destroying resources at this point?

I thought destroying resource would do the right thing™?
The client doesn't like it (wayland fatal error "zxdg_shell_v6@20: error 2: xdg_popup was not created on the topmost popup") but I think it's the wrong one here, I need to take a closer look at some WAYLAND_DEBUG output.

> How do you reproduce this?

konsole, go to any menu with a submenu, it happens when you hover over the entry that has a submenu and it would normally open a 2nd column... But you need to hack in some redefinition of `wl_list_remove` in wlroots/sway for asan to work, I just copy/pasted the implementation from wayland into some file in util/ so wlroots will use the local symbols instead of libwayland's and these have been compiled with asan.
You should be able to reproduce without this hack if you run sway/rootston in valgrind

SirCmpwn:
>I thought destroying resource would do the right thing™?

Not unless the client has indicated they'll no longer use it (which may be true in this case... or might not).

martinetd:
> Not unless the client has indicated they'll no longer use it (which may be true in this case... or might not).

Then it's probably wrong, I guess we could just empty the list here but this might mean fixing how the popup behaves if it has no parent, I don't think this was planned for.
I don't like having two places that can remove a link from a list though, especially since link after wl_list_remove is not considered empty for wl_list_empty....

acrisci:
Yeah, I see the memory errors in valgrind, so this is a problem. But I have a feeling this works only because the client is being disconnected because of the error.

martinetd:
yeah now SirCmpwn said it I agree this is wrong, I'll think about it while at work today

martinetd:
Rewrote that popup fix to just cleanup the list instead, even if that feels ugly to me, that looks safe enough; and it probably should never happen with correct clients anyway.

Pushed a couple more fixes at the same time:
 - xwm has the same problem with parent/children, fixed the same way, but that chlidren list is never used so probably better to just make it go away unless there's plans to use it later.
 - the primary selection one is more straight-forward, unless cancel has any reason to use that event, but it's supposed to come from source_finish so I think that's ok?

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -415,6 +415,12 @@ void destroy_xdg_surface(struct wlr_xdg_surface *surface) {
 
 	wlr_signal_emit_safe(&surface->events.destroy, surface);
 
+	struct wlr_xdg_popup *popup_state, *next;
```
emersion: Meh. I think we should send `popup_done` and make the popup inert.

xwayland/xwm.c
```diff
@@ -294,6 +294,12 @@ static void xwayland_surface_destroy(
 	wl_list_remove(&xsurface->link);
 	wl_list_remove(&xsurface->parent_link);
 
+	struct wlr_xwayland_surface *child, *next;
```
emersion: This one LGTM.

types/xdg_shell/wlr_xdg_surface.c
martinetd: Hm. Sending `popup_done` will not free the popup immediately but wait for the client to ack it, right?
If so, that won't fix the use-after-free that happens when the popup is destroyed after the surface has been, so we'd need that anyway.

I think it'd make sense to do both, though; I didn't realize that call existed.

emersion: >Sending popup_done will not free the popup immediately but wait for the client to ack it, right?

We should use inert resources. Here's how it works:
1. We send `popup_done`
2. We free our struct, and we set the resource's user_data field to NULL (this makes the resource inert)
3. For all requests on the popup, if the resource's user_data field is NULL, ignore the request
4. Except for the `destroy` request, which should still call `wl_resource_destroy`

We maybe have already a function that sends `popup_done`.

There's a paragraph about all of this at the end of `CONTRIBUTING.md`.

martinetd: The destroy function for popups actually kind of does that, so I think the version I just pushed should be good enough™.
Note that I can't really test, the bad client dies to a protocol violation before it can get the popup done or whatever happens; I'm pretty sure we're just splitting hair over something that'll never happen but might as well do it right.

> Except for the destroy request, which should still call wl_resource_destroy

I thought that was odd, so I tried, and this deadloops and overflows the stack with wl_resource_destroy calling the destroy handler calling wl_resource_destroy calling the....
So fixed `CONTRIBUTING.md` instead :>

emersion: >I thought that was odd, so I tried, and this deadloops and overflows the stack with wl_resource_destroy calling the destroy handler calling wl_resource_destroy calling the....

-1. The resource destroy handler is not the same as the destroy request handler.

martinetd: > -1. The resource destroy handler is not the same as the destroy request handler.

Ah. This xdg popup stuff is too confusing, I tried in the wrong one; removed that `CONTRIBUTING.md` commit just now. Thanks for putting up with me.

emersion: Nah, all of this protocol implementation stuff is complicated to understand and it's easy to get lost. That just convinces me that a "How to implement a Wayland protocol" blog post would help.

acrisci:
:+1:

PR #1089, opened 2018-06-26
Check for libavutil/hwcontext_drm.h

ascent12:
This is an optional feature of libavutil, so this will cause a build failure if it's not present (e.g. on Debian/Ubuntu).
Bumps the minimum meson version for the disabler object.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -5,6 +5,10 @@ libavutil = dependency('libavutil', required: false)
 libavcodec = dependency('libavcodec', required: false)
 libavformat = dependency('libavformat', required: false)
 
+if not cc.has_header('libavutil/hwcontext_drm.h', dependencies: libavutil)
+    libavutil = disabler()
+endif
+
 executable('simple', 'simple.c', dependencies: wlroots)
 executable('pointer', 'pointer.c', dependencies: wlroots)
 executable('touch', 'touch.c', 'cat.c', dependencies: wlroots)
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -1,9 +1,10 @@
 threads = dependency('threads')
 wayland_cursor = dependency('wayland-cursor')
 
-libavutil = dependency('libavutil', required: false)
-libavcodec = dependency('libavcodec', required: false)
-libavformat = dependency('libavformat', required: false)
+# These versions correspond to ffmpeg 4.0
+libavutil = dependency('libavutil', version: '>=56.14.100', required: false)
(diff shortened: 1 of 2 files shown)


ammen99:
I think that libavutil et al. should require version >= 4.0. We had a build error with ffmpeg 3.4.2, which was missing some symbols, but was resolved by updating to 4.0

agx:
Yes, compilation after adding

```
#include <libavutil/hwcontext.h>
```
to get AVHWFramesContext, etc. compilation still fails with

```
FAILED: examples/examples@@dmabuf-capture@exe/dmabuf-capture.c.o 
ccache cc -Iexamples/examples@@dmabuf-capture@exe -Iexamples -I../examples -Iinclude -I../include -I/usr/include/x86_64-linux-gnu -I/usr/include/libdrm -I/usr/include/pixman-1 -I/var/scratch/librem5/wlroots/_build -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -O0 -g -Wno-unused-parameter '-DWLR_SRC_DIR="/var/scratch/librem5/wlroots"' -DWL_HIDE_DEPRECATED -pthread  -MD -MQ 'examples/examples@@dmabuf-capture@exe/dmabuf-capture.c.o' -MF 'examples/examples@@dmabuf-capture@exe/dmabuf-capture.c.o.d' -o 'examples/examples@@dmabuf-capture@exe/dmabuf-capture.c.o' -c ../examples/dmabuf-capture.c
../examples/dmabuf-capture.c: In function ‘init_encoding’:
../examples/dmabuf-capture.c:627:57: error: ‘AV_CODEC_CAP_HARDWARE’ undeclared (first use in this function); did you mean ‘AV_CODEC_CAP_DR1’?
  ctx->is_software_encoder = !(out_codec->capabilities & AV_CODEC_CAP_HARDWARE);
                                                         ^~~~~~~~~~~~~~~~~~~~~
                                                         AV_CODEC_CAP_DR1
../examples/dmabuf-capture.c:627:57: note: each undeclared identifier is reported only once for each function it appears in
```

which is not defined in 3.4.2 (or even worse the same (1 << 17) value is used differently)

ascent12:
Updated with minimum versions. ffmpeg sure does have a weird numbering scheme for its libraries.

I haven't tested a failing case, because I don't have a computer with an old ffmpeg, and I can't be bothered having to compile it myself. I just took the values from when 4.0 was tagged, so it should work fine.

emersion:
Thanks!

PR #1091, opened 2018-06-27
wlr_idle: add helper to enable/disable all timers

martinetd:
There was no way to tell wlr_idle to stop processing input events
and rearm timers all the time, such an API is required to have
some form of idle inhibitor.

### DIFF
--- a/types/wlr_idle.c
+++ b/types/wlr_idle.c
@@ -33,6 +33,9 @@ static int idle_notify(void *data) {
 }
 
 static void handle_activity(struct wlr_idle_timeout *timer) {
+	if (!timer->enabled) {
+		return;
+	}
 	// rearm the timer
 	wl_event_source_timer_update(timer->idle_source, timer->timeout);
 	// in case the previous state was sleeping send a resume event and switch state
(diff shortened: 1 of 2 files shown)


martinetd:
Can add wlr_seat as argument, and NULL seat means do them all?

types/wlr_idle.c
```diff
@@ -135,14 +139,29 @@ static void create_idle_timer(struct wl_client *client,
...
+		// arm the timer
+		wl_event_source_timer_update(timer->idle_source, timer->timeout);
+	}
 }
 
 static const struct org_kde_kwin_idle_interface idle_impl = {
 	.get_idle_timeout = create_idle_timer,
 };
 
+void wlr_idle_update_enabled(struct wlr_idle *idle, bool enabled) {
```
emersion: I'd prefer `set_enabled` (we sort of try to only use `update` for impl → interface calls)

emersion:
>Can add wlr_seat as argument, and NULL seat means do them all?

Works for me

types/wlr_idle.c
emersion: All `wlr_output_update_*` are defined in `interfaces/wlr_output.h`, which means they should not be used by compositors, they should only be used by output implementations. `types/wlr_output.h` has some `wlr_output_set_*` functions I believe.

martinetd:
Force pushed the name change + added seat argument

emersion:
Thanks!

PR #1092, opened 2018-06-28
Idle inhibit cleanup

martinetd:
- Added *data pointers to the wlr types
- Cleanup destroy handlers like I think I understood what we talked about, please correct me if I got it wrong

### DIFF
--- a/types/wlr_idle_inhibit_v1.c
+++ b/types/wlr_idle_inhibit_v1.c
@@ -26,23 +26,30 @@ wlr_idle_inhibitor_v1_from_resource(struct wl_resource *resource) {
 	return wl_resource_get_user_data(resource);
 }
 
-static void idle_inhibitor_destroy(struct wl_resource *resource) {
-	struct wlr_idle_inhibitor_v1 *inhibitor =
-		wlr_idle_inhibitor_v1_from_resource(resource);
+static void idle_inhibitor_v1_destroy(struct wlr_idle_inhibitor_v1 *inhibitor) {
+	if (!inhibitor) {
+		return;
+	}
 
 	wlr_signal_emit_safe(&inhibitor->events.destroy, inhibitor->surface);
 
+	wl_resource_set_user_data(inhibitor->resource, NULL);
 	wl_list_remove(&inhibitor->link);
 	wl_list_remove(&inhibitor->surface_destroy.link);
 	free(inhibitor);
+}
+
+static void idle_inhibitor_v1_handle_resource_destroy(struct wl_resource *resource) {
+	struct wlr_idle_inhibitor_v1 *inhibitor =
+		wlr_idle_inhibitor_v1_from_resource(resource);
+	idle_inhibitor_v1_destroy(inhibitor);
 }
 
 static void idle_inhibitor_handle_surface_destroy(
 		struct wl_listener *listener, void *data) {
 	struct wlr_idle_inhibitor_v1 *inhibitor =
 		wl_container_of(listener, inhibitor, surface_destroy);
-
-	wl_resource_destroy(inhibitor->resource);
+	idle_inhibitor_v1_destroy(inhibitor);
 }
 
 static void idle_inhibitor_v1_handle_destroy(struct wl_client *client,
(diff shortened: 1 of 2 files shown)


types/wlr_idle_inhibit_v1.c
```diff
@@ -41,8 +44,11 @@ static void idle_inhibitor_handle_surface_destroy(
 		struct wl_listener *listener, void *data) {
 	struct wlr_idle_inhibitor_v1 *inhibitor =
 		wl_container_of(listener, inhibitor, surface_destroy);
+	struct wl_resource *resource = inhibitor->resource;
 
-	wl_resource_destroy(inhibitor->resource);
+	/* make resource inert */
+	idle_inhibitor_v1_handle_resource_destroy(resource);
```
emersion: Hmmm. Please use the template from CONTRIBUTING.md, I don't like how generic `wl_resource`s are passed around.

martinetd: Can do for this one, but the `idle_inhibit_manager_v1_handle_resource_destroy` does not allocate its own type (only set its link + user data) so I don't really have a choice there.

types/wlr_idle_inhibit_v1.c
```diff
@@ -142,13 +156,16 @@ void wlr_idle_inhibit_v1_destroy(struct wlr_idle_inhibit_manager_v1 *idle_inhibi
 	struct wlr_idle_inhibitor_v1 *inhibitor;
 	struct wlr_idle_inhibitor_v1 *tmp;
 	wl_list_for_each_safe(inhibitor, tmp, &idle_inhibit->inhibitors, link) {
-		wl_resource_destroy(inhibitor->resource);
+		struct wl_resource *resource = inhibitor->resource;
```
emersion: I think it's fine to destroy resources when tearing down the whole manager. Gobals like this aren't supposed to be plug-and-play anyway, and making resources inert introduces complexity for all interfaces of the protocol and is error-prone.

types/wlr_idle_inhibit_v1.c
```diff
@@ -142,13 +149,13 @@ void wlr_idle_inhibit_v1_destroy(struct wlr_idle_inhibit_manager_v1 *idle_inhibi
...
 	wl_list_for_each_safe(inhibitor, tmp, &idle_inhibit->inhibitors, link) {
-		wl_resource_destroy(inhibitor->resource);
+		idle_inhibitor_v1_destroy(inhibitor);
 	}
 
 	struct wl_resource *resource;
 	struct wl_resource *tmp_resource;
 	wl_resource_for_each_safe(resource, tmp_resource, &idle_inhibit->wl_resources) {
-		wl_resource_destroy(inhibitor->resource);
+		wl_resource_destroy(resource);
```
martinetd: fun little bug we had before here...

types/wlr_idle_inhibit_v1.c
```diff
@@ -85,25 +94,30 @@ static void manager_create_inhibitor(struct wl_client *client,
...
 }
 
 static const struct zwp_idle_inhibit_manager_v1_interface idle_inhibit_impl = {
-	.destroy = manager_destroy,
-	.create_inhibitor = manager_create_inhibitor,
+	.destroy = manager_handle_destroy,
+	.create_inhibitor = manager_handle_create_inhibitor,
 };
 
 static void handle_display_destroy(struct wl_listener *listener, void *data) {
```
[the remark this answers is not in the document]
emersion: Yeah, you're right - #1096

types/wlr_idle_inhibit_v1.c
emersion: Damn

types/wlr_idle_inhibit_v1.c
```diff
@@ -26,9 +26,11 @@ wlr_idle_inhibitor_v1_from_resource(struct wl_resource *resource) {
...
 }
 
-static void idle_inhibitor_destroy(struct wl_resource *resource) {
-	struct wlr_idle_inhibitor_v1 *inhibitor =
-		wlr_idle_inhibitor_v1_from_resource(resource);
+static void idle_inhibitor_v1_destroy(struct wlr_idle_inhibitor_v1 *inhibitor) {
+	if (!inhibitor) {
+		return;
+	}
+	wl_resource_set_user_data(inhibitor->resource, NULL);
```
emersion: We should probably do this after emitting the event. Some listener may be calling `wlr_idle_inhibitor_v1_from_resource`.

martinetd: Right, force-pushed. Head is turned off at this point so this comes straight from CONTRIBUTING.md, if/when you update it since the original `subsurface_destroy` has an emit event as well I'd suggest putting it in the "template" as well

emersion:
Thanks!

PR #1093, opened 2018-06-28
xdg_shell popup: fix potential segv in handle_destroy

martinetd:
surface could be NULL there if the popup had been made inert before... I think?


emersion:
Thanks!

PR #1095, opened 2018-06-28
Allow to add additional modes to outputs

### LINKED ISSUE #1080: Support custom modelines on outputs
I'd be nice to support this to allow for specific modes. This is especilly nice in QEMU where you can then set e.g. phone like screen dimensions. Weston supports this like:

https://github.com/krh/weston/blob/master/weston.ini#L49

agx:
This adds support for adding additional modes to an output. This can be useful when a user wants to define custom modes (similar to xrandr --newmode). 

This will become more even more interestin when we want to allow clients to add custom modes e.g. via
https://github.com/swaywm/wlr-protocols/issues/15

Fixes #1080


emersion:
Hmm. I'm not sure I like this. What's the upside of doing this instead of setting the mode you want?

Also this modeline parsing is very meh. And: what should this do on non-DRM backends?

backend/drm/drm.c
```diff
@@ -566,6 +567,75 @@ static bool drm_connector_set_mode(struct wlr_output *output,
...
+		mode->flags |= DRM_MODE_FLAG_NHSYNC;
+	else
+		return false;
+
+	if (strcasecmp(vsync, "+vsync") == 0)
+		mode->flags |= DRM_MODE_FLAG_PVSYNC;
+	else if (strcasecmp(vsync, "-vsync") == 0)
+		mode->flags |= DRM_MODE_FLAG_NVSYNC;
+	else
+		return false;
```
SirCmpwn: braces are mandatory

backend/drm/drm.c
```diff
@@ -566,6 +567,75 @@ static bool drm_connector_set_mode(struct wlr_output *output,
...
+	if (sscanf(s, "%f %hd %hd %hd %hd %hd %hd %hd %hd %15s %15s",
+		   &fclock,
+		   &mode->hdisplay,
+		   &mode->hsync_start,
+		   &mode->hsync_end,
+		   &mode->htotal,
+		   &mode->vdisplay,
+		   &mode->vsync_start,
+		   &mode->vsync_end,
+		   &mode->vtotal, hsync, vsync) != 11)
```
SirCmpwn: Here too

backend/drm/drm.c
```diff
@@ -566,6 +567,75 @@ static bool drm_connector_set_mode(struct wlr_output *output,
...
+		return false;
+
+	if (strcasecmp(vsync, "+vsync") == 0)
+		mode->flags |= DRM_MODE_FLAG_PVSYNC;
+	else if (strcasecmp(vsync, "-vsync") == 0)
+		mode->flags |= DRM_MODE_FLAG_NVSYNC;
+	else
+		return false;
+
+	snprintf(mode->name, sizeof mode->name, "%dx%d@%.3f",
```
SirCmpwn: `sizeof(...)`

SirCmpwn:
Rather than trying to bubble this through the wlr_output interface I think it would be better for users hoping to set this to use grab drm.h and use some DRM-specific function.

agx:
@emersion the upside of being able to add modes is that you can set them in advance (without running them) and a GUI can later on offer them for activation (see https://github.com/swaywm/wlr-protocols/issues/15). It will also appear in the list of available modes returned by wl_output so even without it if we allow to switch a compositor to cycle through a list of available modes this would be in there (so e.g. a sysadmin can set the modes for a user to use at a later point - we'll likely want a wlr_output_remove_mode at a later point to remove a mode that is supported by an output from the list of available modes).

Non-drm backends can simply ignore the unused parameters and use <hdisplay>x<vdisplay>@<refresh> to add to the list of available modes.

I'm also not sure what's so bad about the modeline parsing. We're shifting around Vesa modes implicitly in drm_mode all the time. The alternative would be to fill drmModeModeInfo and pass that around and leave the parsing to the compositor.

agx:
@SirCmpwn you mean adding a drm_connector_add_mode to drm.h ? We could do that but other backends would miss out then (see above) and I think we need add_mode / remove_mode functions functions to give fine enough control to the user.

emersion:
>@emersion the upside of being able to add modes is that you can set them in advance (without running them) and a GUI can later on offer them for activation (see swaywm/wlr-protocols#15).

I really don't understand why _adding_ mode is so important when you can set any custom mode. A client that lets the user choose a mode for an output can also offer to set a custom mode and remember a list of previously used custom modes.

You can also make your compositor do this job of remembering additional custom modes and exposing them through an output management protocol.

>It will also appear in the list of available modes returned by wl_output so even without it if we allow to switch a compositor to cycle through a list of available modes this would be in there

Exposing modes in wl_output was a mistake in the Wayland protocol design - we only do so for user convenience and debugging. GNOME only exposes the current mode for instance.

>Non-drm backends can simply ignore the unused parameters and use x@ to add to the list of available modes.

Remember that not all outputs support modes - all outputs except the DRM ones will have zero modes. This function will be a no-op on these backends, as adding a mode to the list would mark the output as supporting modes.

>I'm also not sure what's so bad about the modeline parsing. We're shifting around Vesa modes implicitly in drm_mode all the time. The alternative would be to fill drmModeModeInfo and pass that around and leave the parsing to the compositor.

It's not convenient for an API. If you want to programmatically add a new mode you need to format a string that'll get parsed anyway right after. Strings are fine in the config file, not in a library API.

>@SirCmpwn you mean adding a drm_connector_add_mode to drm.h ? We could do that but other backends would miss out then (see above) and I think we need add_mode / remove_mode functions functions to give fine enough control to the user.

I think he meant getting the DRM connector handle from the `wlr_output` to allow compositors to add modes themselves.

agx:
> I really don't understand why adding mode is so important when you can set any custom mode. A client 
that
> lets the user choose a mode for an output can also offer to set a custom mode and remember a list of 
> previously used custom modes.

Yeah I could do that but I think proper mode handling should be part of a compositor library. If that's not the case that's ok but and every compositor can do this by itself but..
 
> I think he meant getting the DRM connector handle from the wlr_output to allow compositors to add modes > themselves.

...that only works if I can rely on adding a mode always stays as simple as:

```
wl_list_insert(&conn->output.modes, &mode->wlr_mode.link);
```
is that the case?

> Remember that not all outputs support modes 
From what I have seen  they always have at least one. Otherwise```wlr_output_set_custom_mode``` would be a bad name.

Regarding modeline parsing
> It's not convenient for an API. If you want to programmatically add a new mode you need to format a string > that'll get parsed anyway right after. Strings are fine in the config file, not in a library API.
I agree here and that's why I talked about drmModeModeInfo above. We can pass around the parameters as separate variables but I don't think that's a much better API either. Third way would be to just pass in width, height and refresh and then make up the rest to fill drmModeModeInfo.

SirCmpwn:
>I think he meant getting the DRM connector handle from the wlr_output to allow compositors to add modes themselves.

That's not what I meant. @agx had the right idea, I want a function in `drm.h` which asserts the output passed is a DRM output and then does whatever it wants (up to and including modeline parsing). I think this is a pretty DRM specific feature, and other backends wouldn't benefit from it (at least not the ones we have today, save perhaps for headless, but w/e). Maybe the modeline parsing and adding of the mode can be separate functions and we can reuse them for headless in the future, idk.

emersion:
>I want a function in drm.h which asserts the output passed is a DRM output and then does whatever it wants (up to and including modeline parsing).

I'm okay with the DRM-specific function. This function could take a `drmModeModeInfo` struct directly so that compositors don't have to format a modeline when calling it. The modeline parsing can be moved into rootston's config parsing.

Still not convinced it's really useful, but this design makes it acceptable to me.

>we can reuse them for headless in the future

Some outputs are spec'ed to support modes, in which case the compositor will stick to the modes advertised by the output when modesetting. Some other outputs can have any size, thus don't support modes (ie. modes don't make sense on them): this is the case of all backends except DRM. Outputs support modes if the list of modes is not empty. Adding modes to the headless backend would break this design.

emersion:
>and output_set_custom_mode can free the output that it didn't create?

I don't get it. `output_set_custom_mode` doesn't free outputs. We should check its return value though.

>and while I'm looking x11 and wayland backends define a static function with the same name that's never called and probably ought to be removed.

It's never called, but it's exposed via the output interface (`.set_custom_mode = output_set_custom_mode`). So it should stay there. This output function is used by rootston and sway fwiw.

agx:
Made this drm specific. I moved the modeline parsing to rootston so it does not clutter the wlroots API (yet?). However we could move parse_modeline into drm.c as well allowing compositors to parse modelines and get a xf86ModeModeInfo back.

If this looks better I'd have to guard the places with #ifdef __linux__ in rooston which is the downside of not having it all in drm.c

SirCmpwn:
Looking better.

>If this looks better I'd have to guard the places with #ifdef linux in rooston which is the downside of not having it all in drm.c

Does BSD's drm implementation not support custom modes?

agx:
Ah...forgot about libdrm on FreeBSD. So it will have xf86ModeModeInfo as well.

backend/drm/drm.c
```diff
@@ -566,6 +566,31 @@ static bool drm_connector_set_mode(struct wlr_output *output,
...
+bool wlr_drm_connector_add_mode(struct wlr_output *output, drmModeModeInfo *modeinfo) {
+	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
+	struct wlr_drm_mode *mode = calloc(1, sizeof(*mode));
+
+	if (!mode) {
+		return false;
+	}
+
+	assert(modeinfo);
+	memcpy (&mode->drm_mode, modeinfo, sizeof(*modeinfo));
```
emersion: Style: no space before `(`

emersion:
The DRM backend part LGTM.

@ascent12 Thoughts?

backend/drm/drm.c
```diff
@@ -566,6 +566,31 @@ static bool drm_connector_set_mode(struct wlr_output *output,
...
+	mode->wlr_mode.refresh = mode->drm_mode.vrefresh;
+
+	wlr_log(L_INFO, "Registered custom mode "
+			"%"PRId32"x%"PRId32"@%"PRId32,
+			mode->wlr_mode.width, mode->wlr_mode.height,
+			mode->wlr_mode.refresh);
+	wl_list_insert(&conn->output.modes, &mode->wlr_mode.link);
+	return true;
+}
+
```
ascent12: There are extra blank lines on both sides of this function.

backend/drm/drm.c
```diff
@@ -566,6 +566,31 @@ static bool drm_connector_set_mode(struct wlr_output *output,
 	return false;
 }
 
+
+bool wlr_drm_connector_add_mode(struct wlr_output *output, drmModeModeInfo *modeinfo) {
```
ascent12: modeinfo can be const.

agx:
Hi,
On Thu, Jul 05, 2018 at 01:36:57PM -0700, emersion wrote:
>    Style: no space before (

Fixed. Thanks for having a look.

rootston/output.c
```diff
@@ -841,6 +842,10 @@ void handle_new_output(struct wl_listener *listener, void *data) {
 		roots_config_get_output(config, wlr_output);
 	if (output_config) {
 		if (output_config->enable) {
+			struct roots_output_mode_config *mode_config;
+			wl_list_for_each(mode_config, &output_config->modes, link) {
+				wlr_drm_connector_add_mode(wlr_output, &mode_config->info);
```
emersion: This needs to check that the `wlr_output` is a DRM output.

rootston/config.c
```diff
@@ -322,6 +370,15 @@ static int config_ini_handler(void *user, const char *section, const char *name,
 			wlr_log(L_DEBUG, "Configured output %s with mode %dx%d@%f",
 					oc->name, oc->mode.width, oc->mode.height,
 					oc->mode.refresh_rate);
+		} else if (strcmp(name, "modeline") == 0) {
+			struct roots_output_mode_config *mode = calloc(1, sizeof(*mode));
+
+			if (parse_modeline(value, &mode->info)) {
+				wl_list_insert(&oc->modes, &mode->link);
+			} else {
+				free (mode);
```
emersion: Extra space here

backend/drm/drm.c
emersion: Maybe we should ensure that the type is `DRM_MODE_TYPE_USERDEF`?

agx:
Hi,
On Fri, Jul 06, 2018 at 04:47:51AM -0700, emersion wrote:
>    @emersion commented on this pull request.
> 
>    --------------------------------------------------------------------------
> 
>    In [1]backend/drm/drm.c:
> 
>  > @@ -566,6 +566,31 @@ static bool drm_connector_set_mode(struct wlr_output *output,
>          return false;
>   }
> 
>  +
>  +bool wlr_drm_connector_add_mode(struct wlr_output *output, drmModeModeInfo *modeinfo) {
> 
>    Maybe we should ensure that the type is DRM_MODE_TYPE_USERDEF?

We're defining tht mode so we're setting it as DRM_MODE_TYPE_USERDEF,
what would we check, that it doesn't match an existing mode?

agx:
Hi,
On Fri, Jul 06, 2018 at 06:25:34AM -0700, emersion wrote:
>      We're defining tht mode so we're setting it as DRM_MODE_TYPE_USERDEF,
>      what would we check, that it doesn't match an existing mode?
> 
>    You set DRM_MODE_TYPE_USERDEF in rootston. That means API users can add a
>    mode with this flag set to something else. We should probably forbid that.

Good point! I've fixed that. While doing that I'm wondering if it
wouldn't be better to return 0 in case of success and < 0 in case of
error so we can distinguish parsing errors from OOM? I mostly used bool
since other APIs seem to do so as well.

backend/drm/drm.c
```diff
@@ -566,6 +566,31 @@ static bool drm_connector_set_mode(struct wlr_output *output,
...
+	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
+	struct wlr_drm_mode *mode = calloc(1, sizeof(*mode));
+
+	assert(modeinfo);
+	if (!mode || modeinfo->type != DRM_MODE_TYPE_USERDEF) {
```
emersion: This leaks `mode` in case `modeinfo->type` is incorrect. We can probably move the `modeinfo->type` check before allocating.

emersion:
>While doing that I'm wondering if it wouldn't be better to return 0 in case of success and < 0 in case of error so we can distinguish parsing errors from OOM?

I think using bools are fine

agx:
Hi,
On Fri, Jul 06, 2018 at 04:09:15AM -0700, emersion wrote:
>    @emersion commented on this pull request.
> 
>    --------------------------------------------------------------------------
> 
>    In [1]rootston/output.c:
> 
>  > @@ -841,6 +842,10 @@ void handle_new_output(struct wl_listener *listener, void *data) {
>                  roots_config_get_output(config, wlr_output);
>          if (output_config) {
>                  if (output_config->enable) {
>  +                       struct roots_output_mode_config *mode_config;
>  +                       wl_list_for_each(mode_config, &output_config->modes, link) {
>  +                               wlr_drm_connector_add_mode(wlr_output, &mode_config->info);
> 
>    This needs to check that the wlr_output is a DRM output.

Anoher good point. We're not keeping track of that yet, do we? I'd
rather not look at wlr_output->name.

emersion:
>Anoher good point. We're not keeping track of that yet, do we? I'd rather not look at wlr_output->name.

We can probably use `wlr_output_is_drm`

agx:
Hi,
On Fri, Jul 06, 2018 at 07:05:38AM -0700, emersion wrote:
>      Anoher good point. We're not keeping track of that yet, do we? I'd
>      rather not look at wlr_output->name.
> 
>    We can probably use wlr_output_is_drm

Did not spot the wlr_foo_is_bar pattern by looking at the impl
yet. Fixed as well as the leak. Thanks!

backend/drm/drm.c
SirCmpwn: bump

backend/drm/drm.c
SirCmpwn: bump

SirCmpwn:
LGTM but I'd like to see @ascent12's comments addressed

rootston/output.c
```diff
@@ -841,6 +842,17 @@ void handle_new_output(struct wl_listener *listener, void *data) {
...
 	if (output_config) {
 		if (output_config->enable) {
+			struct roots_output_mode_config *mode_config;
+
+			if (wlr_output_is_drm(wlr_output)) {
+				wl_list_for_each(mode_config, &output_config->modes, link) {
+					wlr_drm_connector_add_mode(wlr_output, &mode_config->info);
+				}
+			} else {
+				if (wl_list_length(&output_config->modes)) {
```
emersion: If you're going to update this PR, here's a nitpick: it's better to use `wl_list_empty` here instead of `wl_list_length` because it is constant time.

agx:
Hi,
On Fri, Jul 06, 2018 at 11:18:33AM -0700, Drew DeVault wrote:
>    LGTM but I'd like to see [1]@ascent12's comments addressed

Missed these initially, fixed now.

agx:
Hi,
On Fri, Jul 06, 2018 at 06:19:48PM +0000, emersion wrote:
>    @emersion commented on this pull request.
> 
>    --------------------------------------------------------------------------
> 
>    In [1]rootston/output.c:
> 
>  > @@ -841,6 +842,17 @@ void handle_new_output(struct wl_listener *listener, void *data) {
>                  roots_config_get_output(config, wlr_output);
>          if (output_config) {
>                  if (output_config->enable) {
>  +                       struct roots_output_mode_config *mode_config;
>  +
>  +                       if (wlr_output_is_drm(wlr_output)) {
>  +                               wl_list_for_each(mode_config, &output_config->modes, link) {
>  +                                       wlr_drm_connector_add_mode(wlr_output, &mode_config->info);
>  +                               }
>  +                       } else {
>  +                               if (wl_list_length(&output_config->modes)) {
> 
>    If you're going to update this PR, here's a nitpick: it's better to use
>    wl_list_empty here instead of wl_list_length because it is constant time.

Fixed as well. Thanks.

emersion:
Hmm, are you missing a `git push`? It seems that the parameter is not `const` and `wl_list_length` is still used?

agx:
Hi,
On Sat, Jul 07, 2018 at 03:05:58AM -0700, emersion wrote:
>    Hmm, are you missing a git push? It seems that the parameter is not const
>    and wl_list_length is still used?

Pused again, I've found another user of wl_list_length which I fixed as
well. I didn't see a comment about const having const xf86ModeModeinfo *
is a good thing.

rootston/output.c
emersion: Ping?

agx: Seems github hates me. Seems the PR updated now correctly.

SirCmpwn:
Thanks!

PR #1100, opened 2018-06-29
do not send focus request to a window that doesn't allow it

apreiml:
In this commit send focus considers the "Client accepts input or input focus" flag. If the client doesn't accept input, it will not set the input focus. The client will do this on her own.

This fixes a bug, where java awt applications (like intellij) do not get focus, if you tab cycle to them or move the focus using the keyboard in sway.

fixes https://github.com/swaywm/sway/issues/2008

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -214,15 +214,24 @@ static void xwm_send_focus_window(struct wlr_xwm *xwm,
 			XCB_INPUT_FOCUS_POINTER_ROOT,
 			XCB_NONE, XCB_CURRENT_TIME);
 		return;
-	} else if (xsurface->override_redirect) {
+	}
+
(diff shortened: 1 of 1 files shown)


xwayland/xwm.c
```diff
@@ -214,15 +214,24 @@ static void xwm_send_focus_window(struct wlr_xwm *xwm,
 			XCB_INPUT_FOCUS_POINTER_ROOT,
 			XCB_NONE, XCB_CURRENT_TIME);
 		return;
-	} else if (xsurface->override_redirect) {
+	}
+
+    if (xsurface->override_redirect) {
```
emersion: Style: we use tabs

emersion:
Thanks!

PR #1101, opened 2018-06-30
Static analysis fixes

martinetd:
A bunch of very small fixes.

I'd rather not name what tool it was as I'm using the paid version of a tool I have through work (If I say we could use the "free for open-source scan" it's probably enough to guess what it is, but I really hate the web UI and the free version doesn't give access to the command-line tools... So meh.)

If there is any commit worth discussing I can split it out in its own PR to discuss while we merge the rest, but I think it should all be fairly straightforward

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -203,6 +203,7 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display,
 				wlr_log(L_ERROR, "failed to start backend '%s'", name);
 				wlr_backend_destroy(backend);
 				wlr_session_destroy(session);
+				free(names);
 				return NULL;
 			}
 
@@ -210,12 +211,14 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display,
 				wlr_log(L_ERROR, "failed to add backend '%s'", name);
 				wlr_backend_destroy(backend);
 				wlr_session_destroy(session);
+				free(names);
 				return NULL;
 			}
 
 			name = strtok_r(NULL, ",", &saveptr);
 		}
 
+		free(names);
 		return backend;
 	}
 
--- a/backend/session/direct-ipc.c
+++ b/backend/session/direct-ipc.c
@@ -159,7 +159,9 @@ static void communicate(int sock) {
 			}
 error:
 			send_msg(sock, ret ? -1 : fd, &ret, sizeof(ret));
-			close(fd);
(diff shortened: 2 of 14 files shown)


backend/session/direct-ipc.c
```diff
@@ -159,7 +159,8 @@ static void communicate(int sock) {
 			}
 error:
 			send_msg(sock, ret ? -1 : fd, &ret, sizeof(ret));
-			close(fd);
+			if (fd >= 0)
+				close(fd);
```
SirCmpwn: braces

backend/wayland/wl_seat.c
```diff
@@ -56,6 +56,9 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 	}
 
 	struct wlr_wl_output *output = wl_surface_get_user_data(surface);
+	if (output == NULL) {
+		return;
+	}
```
SirCmpwn: This can be an assert

types/wlr_keyboard.c
```diff
@@ -156,7 +157,8 @@ void wlr_keyboard_destroy(struct wlr_keyboard *kb) {
 	}
 	xkb_state_unref(kb->xkb_state);
 	xkb_keymap_unref(kb->keymap);
-	close(kb->keymap_fd);
+	if (kb->keymap_fd >= 0)
+		close(kb->keymap_fd);
```
SirCmpwn: braces

SirCmpwn:
LGTM

backend/wayland/wl_seat.c
martinetd: Right-o, changed in pointer_handle_enter as well. Force pushed along with braces fix in their appropriate commits

backend/backend.c
```diff
@@ -203,19 +203,22 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display,
...
 				wlr_backend_destroy(backend);
 				wlr_session_destroy(session);
+				free(names);
 				return NULL;
 			}
 
 			name = strtok_r(NULL, ",", &saveptr);
 		}
 
+		free(names);
```
martinetd: (last push added this line, I hadn't looked deeply enough into what `names` was used for and it came up on recheck. Am done rechecking now)

emersion:
Thanks!

PR #1102, opened 2018-06-30
s/wlr_subsurface_from_surface/wlr_subsurface_from_wlr_surface/

martinetd:
This was the only x_from_wlr_surface function that lacked the wlr_ prefix,
let's have an API as uniform as possible.

Cc @ongy @Timidger @ammen99 as it is an API change


SirCmpwn:
Do you mind preparing a patch for the affected projects? Should be small.

SirCmpwn:
Thanks!

PR #1108, opened 2018-07-03
Apply opaque regions to surfaces' current state

RyanDwyer:
This change is needed to allow https://github.com/swaywm/sway/pull/2182 to see the opaque region in the current state.

I admit I'm a little confused by the property `invalid`. I assume it denotes the fields which are dirty and should be copied into the current state.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -258,6 +258,7 @@ static void surface_move_state(struct wlr_surface *surface,
 	}
 	if ((next->invalid & WLR_SURFACE_INVALID_OPAQUE_REGION)) {
 		// TODO: process buffer
+		pixman_region32_copy(&state->opaque, &next->opaque);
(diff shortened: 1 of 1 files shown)


emersion:
This is fixed in #/1076.

types/wlr_surface.c
```diff
@@ -258,6 +258,7 @@ static void surface_move_state(struct wlr_surface *surface,
 	}
 	if ((next->invalid & WLR_SURFACE_INVALID_OPAQUE_REGION)) {
 		// TODO: process buffer
+		pixman_region32_copy(&state->opaque, &next->opaque);
 		pixman_region32_clear(&next->opaque);
```
emersion: `next->opaque` should not be cleared.

RyanDwyer:
Closing this - I'll just wait until #1076 is merged.

PR #1110, opened 2018-07-03
fix: add stack update on focus change

apreiml:
Enable the stack update again for focus changes on non-focusable views.

I introduced this bug earlier in #1100.


xwayland/xwm.c
```diff
@@ -228,13 +228,13 @@ static void xwm_send_focus_window(struct wlr_xwm *xwm,
 		// if the surface doesn't allow the focus request, we will send him
 		// only the take focus event. It will get the focus by itself.
 		xwm_send_wm_message(xsurface, &message_data, XCB_EVENT_MASK_NO_EVENT);
-		return;
 	}
+	else {
```
emersion: `}` and `else` should be on the same line

SirCmpwn:
Thanks!

Issue #1041, opened 2018-06-05
Firefox-wayland crash when selection is available


martinetd:
I think I need help with this.
Here's the logs on our (rootston) side, because we don't have the last `.selection()` event on FF side as it crashes:

```
[1404702.976]  -> gtk_primary_selection_device@25.data_offer(new id gtk_primary_selection_offer@4278190080)
[1404703.009]  -> gtk_primary_selection_device@15.data_offer(new id gtk_primary_selection_offer@4278190080)
[1404703.039]  -> gtk_primary_selection_offer@4278190080.offer("text/plain")
[1404703.064]  -> gtk_primary_selection_offer@4278190080.offer("text/plain;charset=utf-8")
[1404703.089]  -> gtk_primary_selection_device@25.selection(gtk_primary_selection_offer@4278190080)
[1404703.113]  -> gtk_primary_selection_device@15.selection(gtk_primary_selection_offer@4278190080)
```

Where both selection devices have been created on different wl_seat (logs on ff side this time)
```
[191882.862]  -> gtk_primary_selection_device_manager@10.get_device(new id gtk_primary_selection_device@15, wl_seat@13)
[192201.245]  -> gtk_primary_selection_device_manager@22.get_device(new id gtk_primary_selection_device@25, wl_seat@23)
```

tbh even if it was a single seat it'd probably crash the app just the same to instanciate multiple device managers/selection devices and use them, so not sure we should focus on that part..

So apparently it's not OK to offer up the same selection to multiple devices.
Either we need to instanciate a different `gtk_primary_selection_offer` and offer one each, or we need to pick our poison and give it to one.
weston doesn't seem to have a complete selection implementation?
gnome-shell only gives it to one, presumably the latest device manager instanciated.

What shall we do?

Ongy:
I'd say if a client created to devices it signaled interest twice and this should be a bug on the firefox project and GNOME, not here.

martinetd:
What strikes me as possibly wrong is that we're using the same gtk_primary_selection_offer (same id) while the call is of `type=new_id` so means allocation.
I think if we send it twice, we need to allocate (and fill!) one per selection device.

martinetd:
Hm now I'm thinking a bit more we do the same with the selection `device_manager.get_device` that always return the same id for a given seat, so that might be ok. I don't know.

Ongy:
Oh, yea. If it's new_id it probably allocates a new proxy and could lead to double-free like problems.
[...]

PR #1113, opened 2018-07-05
[WIP] wlr_primary_selection: Allow multiple offers per source

martinetd:
Create as many offers as there are selection devices, so that we do
not send a client multiple times the same "new_id" offer (which
would make it crash)

Note something is weird with this and X11; once I've selected something in firefox then pastes to X11 do not work anymore (even from other wayland apps); I don't understand what though so could use some help with that.

Part of #1041 (this is not complete as it only does primary_selection, data_source needs a similar fix... but data source uses source->offer everywhere (like in drag n drop as well) so such an approach is going to be a royal pain)

Cc @aereaux as he said he would work on this a couple of weeks ago, if you want to take over please do (I can find something else to do :P); or if you just have thought about it and have an opinion it's welcome as well :)

### DIFF
--- a/include/wlr/types/wlr_primary_selection.h
+++ b/include/wlr/types/wlr_primary_selection.h
@@ -24,7 +24,7 @@ struct wlr_primary_selection_source {
 	void (*cancel)(struct wlr_primary_selection_source *source);
 
 	// source status
-	struct wlr_primary_selection_offer *offer;
+	struct wl_list offers;
 	struct wlr_seat_client *seat_client;
 
 	struct {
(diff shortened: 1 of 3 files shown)


martinetd:
BTW if there is no client in wleird that binds to the seat twice and deals with data devices I think it would be a nifty thing to add (I guess saying that means I'm welcome to do it... But it can't hurt to try, I already have a weird client called firefox here :P)

emersion:
Maybe `offer` should have multiple resources instead?

emersion:
> It does have the problem that we still need to match 1-to-1 the selection/data device to the resource.

Right, I was thinking that using `wl_resource_get_client` would  be enough, but it won't :(

martinetd:
Ah, we already use that user data to store a pointer to the wlr offer struct in the wl_resource offer, so I can't use it for the device :/

Tricky.

martinetd:
Superseded by #1380; thanks!

PR #1115, opened 2018-07-06
surface: replace wlr_surface_set_role_committed by a role_commit event

emersion:
I wanted to do this before adding a `role_precommit` event.

Fixes #527

PR #1116, opened 2018-07-07
surface: replace wlr_surface_set_role_committed with wlr_surface_role

### LINKED ISSUE #527: #524 follow-up

emersion:
Seems like I love doing refactoring. =.=

This is a redo of #1115, this time using the `struct wlr_surface_role` mentioned in #527.

Fixes #527

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -36,6 +36,11 @@ struct wlr_surface_state {
 	struct wl_listener buffer_destroy;
 };
 
+struct wlr_surface_role {
+	const char *name;
+	void (*commit)(struct wlr_surface *surface);
+};
+
 struct wlr_surface {
 	struct wl_resource *resource;
 	struct wlr_renderer *renderer;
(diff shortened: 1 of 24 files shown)


include/wlr/types/wlr_surface.h
```diff
@@ -36,6 +36,11 @@ struct wlr_surface_state {
 	struct wl_listener buffer_destroy;
 };
 
+struct wlr_surface_role {
+	const char *name;
```
emersion: Note that this is only here for debugging purposes, and isn't really useful anymore.

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -163,7 +163,8 @@ struct wlr_xdg_surface {
 	struct wlr_box next_geometry;
 	struct wlr_box geometry;
 
-	struct wl_listener surface_destroy_listener;
+	struct wl_listener surface_destroy;
+	struct wl_listener surface_commit;
```
acrisci: Why do we have two surface commit handlers now?

emersion: `xdg-shell` wants to do stuff before the role has been assigned (mainly check that there's no non-null commit before configure has been sent). So, we need another listener to do that.

Issue #1123, opened 2018-07-09
Shift count overflow due to size of time_t


5nefarious:
I am trying to build wlroots on an ARM system and am getting a compilation error. The error seems to be in both `types/wlr_export_dmabuf_v1.c` and `types/wlr_screencopy_v1.c`, where `tv_sec` is being split into high and low bits. `tv_sec` is of type `time_t`, and it looks like the size of this type is implementation-specific. On my system, I believe it must be 32 bits. Therefore, the following line in both files would cause undefined behavior:

    uint32_t tv_sec_hi = event->when->tv_sec >> 32;

Could the code that handles `time_t` types be made more robust to handle this situation?

emersion:
>Could the code that handles time_t types be made more robust to handle this situation?

Yeah, that's probably a good idea. Can you try to add a `sizeof` check, and don't do the bitshift (use 0) if `tv_sec` has no more than 4 bytes?

PR #1124, opened 2018-07-09
fix: handle 32 bit time_t (or smaller)

5nefarious:
See issue #1123


atomnuker:
Why don't you just cast it to a uint64_t? No branches.

Ongy:
`sizeof` is evaluated at compile time. This shouldn't be a branch in the binary either way.

emersion:
@5nefarious Thanks!

@Ongy Thanks for testing

PR #1127, opened 2018-07-09
surface: add wlr_surface_role.precommit 

### LINKED ISSUE #1065: Allow compositors to actually access the surface->buffer on unmap

emersion:
~~Depends on  #1116~~

Fixes #1065

### DIFF
--- a/include/types/wlr_xdg_shell.h
+++ b/include/types/wlr_xdg_shell.h
@@ -19,7 +19,8 @@ struct wlr_xdg_surface *create_xdg_surface(
 	uint32_t id);
 void unmap_xdg_surface(struct wlr_xdg_surface *surface);
 void destroy_xdg_surface(struct wlr_xdg_surface *surface);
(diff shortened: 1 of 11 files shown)


emersion:
Ping?

SirCmpwn:
Thanks!

PR #1128, opened 2018-07-11
Unbreak xvfb

agx:
x11: make sure event source is drained
    
 Otherwise running under Xvfb will not deliver any events. This results in e.g. weston-info reporting a 0x0 window size (which results in all sorts of problems) plus some other minor cleanups.

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -47,8 +47,16 @@ static bool output_set_custom_mode(struct wlr_output *wlr_output,
 	output_set_refresh(&output->wlr_output, refresh);
 
 	const uint32_t values[] = { width, height };
-	xcb_configure_window(x11->xcb_conn, output->win,
+	xcb_void_cookie_t cookie = xcb_configure_window_checked(x11->xcb_conn, output->win,
 		XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT, values);
+
+	xcb_generic_error_t *error;
+	if ((error = xcb_request_check(x11->xcb_conn, cookie))) {
+		wlr_log(WLR_ERROR, "Could not set window size to %dx%d\n", width, height);
+		free(error);
+		return false;
+	}
+
 	return true;
 }
 
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -264,6 +264,7 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
 		wlr_log(WLR_ERROR, "Could not create event source");
 		goto error_display;
 	}
+	wl_event_source_check(x11->event_source);
 
 	x11->screen = xcb_setup_roots_iterator(xcb_get_setup(x11->xcb_conn)).data;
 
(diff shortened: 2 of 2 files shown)


emersion:
`wl_event_loop_fd_func_t` does say the return value should be:

> If the event source is registered for re-check with wl_event_source_check(): 0 for all done, 1 for needing a re-check. If not registered, the return value is ignored and should be zero.

However `x11_event` always returns 0, which means we always say "this doesn't need to re-check" (same as not calling `wl_event_source_check`). So I don't quite understand what's happening.

(BTW, we have a headless backend)

agx:
Hi,
On Wed, Jul 11, 2018 at 08:51:39AM -0700, emersion wrote:
>    wl_event_loop_fd_func_t does say the return value should be:
> 
>      If the event source is registered for re-check with
>      wl_event_source_check(): 0 for all done, 1 for needing a re-check. If
>      not registered, the return value is ignored and should be zero.
> 
>    However x11_event always returns 0, which means we always say "this
>    doesn't need to re-check" (same as not calling wl_event_source_check). So
>    I don't quite understand what's happening.

See ``wl_event_loop_dispatch()``. It makes sure that we're dispatched
again after idle sources. Not sure why regular x.org does not need this
but having Xvfb working is a nice thing.

> 
>    (BTW, we have a headless backend)

Yeah, I know but I want some X11 things here to synthesize input events,
etc.

emersion:
>See ``wl_event_loop_dispatch()``. It makes sure that we're dispatched again after idle sources. Not sure why regular x.org does not need this but having Xvfb working is a nice thing.

Oh, right. Thanks for the explanation.

>Yeah, I know but I want some X11 things here to synthesize input events, etc.

(You can also do this with the headless backend - but ofc yeah, we want the X11 backend to work in Xvfb)

emersion:
Thanks!

backend/x11/output.c
```diff
@@ -47,8 +47,16 @@ static bool output_set_custom_mode(struct wlr_output *wlr_output,
...
 	const uint32_t values[] = { width, height };
-	xcb_configure_window(x11->xcb_conn, output->win,
+	xcb_void_cookie_t cookie = xcb_configure_window_checked(x11->xcb_conn, output->win,
 		XCB_CONFIG_WINDOW_WIDTH | XCB_CONFIG_WINDOW_HEIGHT, values);
+
+	xcb_generic_error_t *error;
+	if ((error = xcb_request_check(x11->xcb_conn, cookie))) {
+		wlr_log(WLR_ERROR, "Could not set window size to %dx%d\n", width, height);
+		free(error);
+		return false;
```
psychon: Should this also print information about what the error was?
Also, what kind of errors do you expect? According to https://www.x.org/releases/X11R7.5/doc/x11proto/proto.html (search for `ConfigureWindow`), the possible errors are Match, Value, and Window:

- BadMatch cannot happen here (happen when setting the border width of an InputOnly window or changing the stacking order in an incorrect way).
- BadValue can easily be checked against before sending the broken request (happens when setting width or height to zero).
- BadWindow... well, you have worse problems if you got your own window ID wrong.

So basically, this change introduces an unnecessary round-trip to the X11 server, I think.

Issue #277, opened 2017-10-16
Do not render regions under opaque surfaces


emersion:
ATM, we just iterate over all surfaces and render each of them. However, there are many situations when we don't need to render a surface: when a window is totally hidden behind other windows, or when the lockscreen is enabled.

We can use `wlr_surface.current.opaque_region` for that.

ascent12:
This is somewhat part of #190.
I've been looking into making the renderer more efficient, but I haven't had the time to implement it yet.

ammen99:
Just saw this issue, sounds like it is obsolete now that we don't have rootston anymore?

emersion:
Indeed, this is the compositor's job.

PR #1130, opened 2018-07-12
Add wlr_surface.opaque_region

### LINKED ISSUE #920: Set an infinite surface opaque region if buffer format doesn't have alpha
Creating an issue to remember about this optimization.

emersion:
This is the enhanced opaque region:
* Clipped to the surface bounds
* Set to the whole surface if the texture is using a format without an alpha channel

Fixes #920

### DIFF
--- a/include/wlr/render/interface.h
+++ b/include/wlr/render/interface.h
@@ -56,6 +56,7 @@ void wlr_renderer_init(struct wlr_renderer *renderer,
 
 struct wlr_texture_impl {
 	void (*get_size)(struct wlr_texture *texture, int *width, int *height);
+	bool (*is_opaque)(struct wlr_texture *texture);
 	bool (*write_pixels)(struct wlr_texture *texture,
 		enum wl_shm_format wl_fmt, uint32_t stride, uint32_t width,
 		uint32_t height, uint32_t src_x, uint32_t src_y, uint32_t dst_x,
--- a/include/wlr/render/wlr_texture.h
+++ b/include/wlr/render/wlr_texture.h
@@ -41,6 +41,11 @@ struct wlr_texture *wlr_texture_from_dmabuf(struct wlr_renderer *renderer,
 void wlr_texture_get_size(struct wlr_texture *texture, int *width, int *height);
 
 /**
+ * Returns true if this texture is using a fully opaque format.
+ */
+bool wlr_texture_is_opaque(struct wlr_texture *texture);
+
+/**
  * Update a texture with raw pixels. The texture must be mutable.
  */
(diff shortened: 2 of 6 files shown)


agx:
Looks good so far! I assume the users of ``wlr_surface_is_opaque()`` are comming up later in this PR?

emersion:
>I assume the users of wlr_surface_is_opaque() are comming up later in this PR?

Do you mean #277?

agx:
On Fri, Jul 13, 2018 at 02:57:12AM -0700, emersion wrote:
>      I assume the users of wlr_surface_is_opaque() are comming up later in
>      this PR?
> 
>    Do you mean [1]#277?

Something like that, yes. thanks! Will be interesting to see the numbers
how it speeds things up.

SirCmpwn:
Thanks!

emersion:
Yeah, I'm not sure I want to implement this in rootston. I'll probably do sway first, which is somewhat easier (if a layer covers the whole screen, don't render stuff underneath)

SirCmpwn:
I wonder if the need for something to rootston to exist is going to go down over time - maybe we should remove it at some point and just use everyone's respective compositors as a reference/testbed.

agx:
> Yeah, I'm not sure I want to implement this in rootston. I'll probably do sway first, which is somewhat 
> easier (if a layer covers the whole screen, don't render stuff underneath).
I might give it a try for rootston one it's in sway.

emersion:
On one hand I agree, on the other hand it's still my primary testbed - the fact that it's a simple floating compositor helps. For instance debugging damage tracking is way easier on rootston than on sway - because sway will do a lot of unnecessary repaints when e.g. opening a new window or changing focus.

emersion:
That's true, "can you reproduce in rootston" is still something I ask myself/others pretty often.

SirCmpwn:
Other wlroots-based compositors have the same advantage, though. "Can you reproduce it in waymonad?"

emersion:
>Other wlroots-based compositors have the same advantage, though. "Can you reproduce it in waymonad?"

"Real" compositors make much more complicated stuff. rootston is simple in comparison, and easier to debug.

SirCmpwn:
Fair enough.

agx:
On Fri, Jul 13, 2018 at 04:30:56AM -0700, Drew DeVault wrote:
>    Other wlroots-based compositors have the same advantage, though. "Can you
>    reproduce it in waymonad?"

But it's not shipped with the library (i.e. has the correct version
matching wlroots, does not require any additional tools to get it
working, etc).

Issue #1008, opened 2018-05-27
API stability guarantee


SirCmpwn:
We need to make a plan and set a deadline for enforcing it. I'm thinking we make our first stability guarantees by July 1st.

I think we should do this by headers, marking older headers as stable and newer headers as unstable, and over time graduate everything to stable.

Timidger:
I'm all for that. It makes my work on wlroots-rs easier as it means I can definitely be sure to finish wrapping something (which is non trivial with Rust's safety guarantees).

Ongy:
Missing stability promises are the reason hsroots lags behind and mostly just supports waymonad :)

And I think it's a good idea to have a set of unstable APIs as well. I don't think huge additions like the tablet_v2 PR should be stable from the get go.
Any plans on how to define them? Have them behind a `if WLR_USE_UNSTABLE`, under `include/wlr/unstable/`, a prose comment, or something completly different?

And for the transition from an unstable API to stable I suggest as rule of thumb:
Implementation needs to be finished in one external compositor, written by someone who didn't add it to wlroots.
I.e. waymonad wouldn't count for my additions to wlroots. This should ensure someone else used the API.

agx:
In terms of ABI stability we can change WLROOTS_0_0_0 in wlrotos.symbols to WLROOTS_0_1_0 and put the existing symbols in there adding the current wlr_* to a new WLROOTS_UNSTABLE {} section. (and doing the same for 0_2_0, ... and so on. 

For API stability version macros like glib based libraries use them could be used:

  https://github.com/GNOME/glib/blob/master/glib/gversionmacros.h

acrisci:
I feel like we're getting really close to being stable. But I'd like to go through and see if anything should be made private. It's easier to add them back in later versions than it is to remove them.

acrisci:
Also, where did we land on supporting multiple renderers? That will certainly be breaking.

emersion:
>Also, where did we land on supporting multiple renderers? That will certainly be breaking.

RIP

Timidger:
The multiple renders sounds like it would add a lot more time if we wanted it in 1.0. Is it a feature that couldn't wait until a 2.0 in the future?

emersion:
Multiple renderers won't happen.

acrisci:
[...]

PR #1131, opened 2018-07-13
Initial pass on API stability guarantees

SirCmpwn:
This introduces -DWLR_USE_UNSTABLE and adds information regarding the
stability status to all headers. I started with a conservative set of
headers to mark as stable:

- types/wlr_matrix.h
- util/edges.h
- util/log.h
- util/region.h
- xcursor.h

Updates #1008

### DIFF
--- a/include/wlr/types/wlr_matrix.h
+++ b/include/wlr/types/wlr_matrix.h
@@ -1,20 +1,57 @@
+/*
+ * This is a stable interface of wlroots. Future changes will be limited to:
+ *
+ * - New functions
+ * - New struct members
+ * - New enum members
+ *
+ * Note that wlroots does not make an ABI compatibility promise - in the future,
+ * the layout and size of structs used by wlroots may change, requiring code
+ * depending on this header to be recompiled (but not edited).
+ *
+ * Breaking changes are announced by email and follow a 1-year deprecation
+ * schedule. Send an email to ~sircmpwn/wlroots-announce+subscribe@lists.sr.ht
+ * to receive these announcements.
+ */
+
 #ifndef WLR_TYPES_WLR_MATRIX_H
 #define WLR_TYPES_WLR_MATRIX_H
 
 #include <wayland-server.h>
 #include <wlr/types/wlr_box.h>
 
+/** Writes the identity matrix into mat */
 void wlr_matrix_identity(float mat[static 9]);
+
+/** mat ← a × b */
 void wlr_matrix_multiply(float mat[static 9], const float a[static 9],
 	const float b[static 9]);
+
 void wlr_matrix_transpose(float mat[static 9], const float a[static 9]);
+
+/** Writes a 2D translation matrix to mat of magnitude (x, y) */
 void wlr_matrix_translate(float mat[static 9], float x, float y);
+
+/** Writes a 2D scale matrix to mat of magnitude (x, y) */
 void wlr_matrix_scale(float mat[static 9], float x, float y);
+
+/** Writes a 2D rotation matrix to mat at an angle of rad radians */
 void wlr_matrix_rotate(float mat[static 9], float rad);
+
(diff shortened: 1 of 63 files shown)


include/wlr/types/wlr_matrix.h
```diff
@@ -1,20 +1,57 @@
...
 #include <wlr/types/wlr_box.h>
 
+/** Writes the identity matrix into mat */
 void wlr_matrix_identity(float mat[static 9]);
+
+/** mat ← a × b */
 void wlr_matrix_multiply(float mat[static 9], const float a[static 9],
 	const float b[static 9]);
+
 void wlr_matrix_transpose(float mat[static 9], const float a[static 9]);
```
emersion: mat ← aᵀ

emersion:
It may make sense to update `wlroots.syms` to drop the `0_0_0`.

ascent12:
>Breaking changes are announced by email and follow a 1-year deprecation schedule

This is far too long of a promise. In fact, I'm against time-based promises at all, especially when it's something still as cutting edge as this.
It should be release-based, and probably just follow basic semver like most other people do (once we actually have a versioned release).

>util/log.h

I disagree with any of the wlr_log functions (excluding wlr_log_init) even being part of the API. We don't need to provide some generic logging framework; we just need to give people log messages when they want them.

It's because of those functions being exported that sway{lock,bg,idle,bar,msg} are massively over-linked; just because they were using some rather worthless logging functions.

SirCmpwn:
>This is far too long of a promise. In fact, I'm against time-based promises at all, especially when it's something still as cutting edge as this.
It should be release-based, and probably just follow basic semver like most other people do (once we actually have a versioned release).

I want to make proper stability promises, not some empty semver-based promise which'll just lead to releasing wlroots 138.19.2 in 6 months.

>I disagree with any of the wlr_log functions (excluding wlr_log_init) even being part of the API. We don't need to provide some generic logging framework; we just need to give people log messages when they want them.

The important part is making it so that the application can process *our* logging and presumably incoproprate it with their own logging.

Ongy:
Most logging systems I've seen, that are intended to be re-used, have a way to push down their own logging fuction, to e.g. redirect to syslog, from the current stderr approach.

Might be worth a consideration as well.

SirCmpwn:
wlroots has that.

SirCmpwn:
This can't sit dead in the water forever. We need to address these two points:

- I want to use a 1 year deprecation lifecycle, and I don't want to use semver. I don't want to make half promises about stability when in reality we're just cranking up that major version number all of the time as an excuse not to be. Once we finish implementing a protocol, until a new protocol version comes out our implementation is unlikely to need changes and when the new protocol *does* come out, it'll be backwards compatible because we only mark stable protocols as stable in wlroots.
- I want to include wlr_log. We're definitely going to keep using it internally - we need a logging system of some kind. What's the harm in exposing it? We need to at least expose a means for the user to provide their own callback instead of printf. I don't see why we can't expose the rest, too. It's not like we change it often, the only changes since December have been renaming things to use a WLR_ prefix.

emersion:
>I want to use a 1 year deprecation lifecycle, and I don't want to use semver.

That's pretty dangerous and could be potentially harmful. Also bye bye cool renderer stuff, bye bye DRM planes.

>I want to include wlr_log

At least don't mark it as stable.

emersion:
>We don't have to start labelling things as stable any time soon.

Ah. In this case the whole library won't ever be stable for years.

>I want to include wlr_log

/shrug

SirCmpwn:
>In this case the whole library won't ever be stable for years.

Right. We can roll this out nice and slow.

Ongy:
Can you add tablet-v2 header to the files addapted in this PR?

SirCmpwn:
Aight, we've gotta get some movement on this, so I'm going to merge this once the build passes.

PR #1134, opened 2018-07-16
Install generated protocols

ascent12:
This installs the server protocols we generate as <wlr/protocol/name.h>.

This removes the weird requirement for the user to run wayland-scanner and set their include paths up in a very specific way to use some of our public headers. It's not really expected for users to include these headers directly, but the option is there if they want to.

### DIFF
--- a/include/wlr/types/wlr_layer_shell.h
+++ b/include/wlr/types/wlr_layer_shell.h
@@ -3,9 +3,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <wayland-server.h>
+#include <wlr/protocol/wlr-layer-shell-unstable-v1-protocol.h>
 #include <wlr/types/wlr_box.h>
 #include <wlr/types/wlr_surface.h>
-#include "wlr-layer-shell-unstable-v1-protocol.h"
 
 /**
  * wlr_layer_shell allows clients to arrange themselves in "layers" on the
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -1,9 +1,9 @@
 #ifndef WLR_TYPES_WLR_XDG_SHELL_H
 #define WLR_TYPES_WLR_XDG_SHELL_H
+#include <wayland-server.h>
+#include <wlr/protocol/xdg-shell-protocol.h>
 #include <wlr/types/wlr_box.h>
 #include <wlr/types/wlr_seat.h>
-#include <wayland-server.h>
-#include "xdg-shell-protocol.h"
 
 struct wlr_xdg_shell {
 	struct wl_global *global;
--- a/include/wlr/types/wlr_xdg_shell_v6.h
+++ b/include/wlr/types/wlr_xdg_shell_v6.h
@@ -2,9 +2,9 @@
 #define WLR_TYPES_WLR_XDG_SHELL_V6_H
 
 #include <wayland-server.h>
+#include <wlr/protocol/xdg-shell-unstable-v6-protocol.h>
 #include <wlr/types/wlr_box.h>
 #include <wlr/types/wlr_seat.h>
-#include "xdg-shell-unstable-v6-protocol.h"
 
 struct wlr_xdg_shell_v6 {
 	struct wl_global *global;
--- a/protocol/meson.build
+++ b/protocol/meson.build
@@ -1,31 +1,13 @@
 wl_protocol_dir = wayland_protos.get_pkgconfig_variable('pkgdatadir')
 
-wayland_scanner = find_program('wayland-scanner')
-
-wayland_scanner_server = generator(
-	wayland_scanner,
-	output: '@BASENAME@-protocol.h',
-	arguments: ['server-header', '@INPUT@', '@OUTPUT@'],
-)
-
-# should check wayland_scanner's version, but it is hard to get
-if wayland_server.version().version_compare('>=1.14.91')
-	code_type = 'private-code'
-else
-	code_type = 'code'
+wl_scanner = find_program('wayland-scanner', required: false)
+if not wl_scanner.found()
+	wl_scanner_dep = dependency('wayland-scanner')
+	wl_scanner = find_program(wl_scanner_dep.get_pkgconfig_variable('wayland_scanner'))
 endif
 
-wayland_scanner_code = generator(
-	wayland_scanner,
-	output: '@BASENAME@-protocol.c',
-	arguments: [code_type, '@INPUT@', '@OUTPUT@'],
-)
-
-wayland_scanner_client = generator(
-	wayland_scanner,
-	output: '@BASENAME@-client-protocol.h',
-	arguments: ['client-header', '@INPUT@', '@OUTPUT@'],
-)
+wl_scanner_version = run_command(wl_scanner, '-v').stderr().strip().split(' ')[-1]
+code_type = wl_scanner_version.version_compare('>=1.14.91') ? 'public-code' : 'code'
 
 protocols = [
 	[wl_protocol_dir, 'stable/xdg-shell/xdg-shell.xml'],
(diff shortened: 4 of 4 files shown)


emersion:
The alternative would be to remove the `#include`s from the headers and expect the user to do it himself.

SirCmpwn:
I like the status quo.

emersion:
>I like the status quo.

-1, requiring compositors to generate headers and put them in a specific directory is very meh.

SirCmpwn:
It's not a specific directory, it's just "any directory you can specify with `-I`", e.g. any directory they damn well please.

emersion:
>set their include paths up in a very specific way to use some of our public headers

that

ammen99:
Just my 5 cents, but I also think the current configuration is far from ideal

SirCmpwn:
They're going to have to invoke wayland-scanner anyway to provide the protocol implementation to link with.

Ongy:
I don't particularily like this situation either, but these headers aren't wlroots headers.
Maybe wlr-protocols, but I don't think it should be special cased.

IMO we should have upstream installs (or distros) fix that.
Only installing the .xml files with `*-protocols` is the main culprint IMO, and (I may just be dumb) I don't see the benefit from that.

emersion:
>IMO we should have upstream installs (or distros) fix that.

If we want to go this route, wayland-protocols upstream should make `make install` generate headers and code. I'm not sure it's the right approach.

SirCmpwn:
I think people should just run wayland-scanner themselves.

ascent12:
>but these headers aren't wlroots headers.

We require several enums from various protocols as part of our public API. Since we can't get those values anywhere else, the headers need to be part of our public API too.

> Only installing the .xml files with *-protocols is the main culprint IMO, and (I may just be dumb) I don't see the benefit from that.

This have been brought up before, and it's not going to happen. Basically, installing headers doesn't solve who is going to generate and link the appropriate source files, and nobody wants a milltions libraries for each particular wayland protocol to have to link to. It's also a C-only solution, and doesn't solve the issue for any other language should do this.

>I like the status quo.
>I think people should just run wayland-scanner themselves.

You're going to have to give an actual justification for this, instead of just blindly saying no.
I consider the way we're currently doing this to be the worst possible way.

Here are the current solutions to this problem I've already considered:

## Status Quo
i.e. `#include "name-protocol.h"`

### Cons:
- Requires the user to set their include paths in a very specific way
- Requires the user to name the protocol header a very specific way (Requires -protocol suffix)
- Duplicate effort from the user, since we've already gone through the effort of generating the linking the code/header
- Annoying for users

### Pros:
- None

As I mentioned before, I think this is definitely the worst solution. I've never seen such stupid requirements for using a library, and I think forcing the user to do that is unacceptable. I don't see the point of doing this _at all_. There is no advantage of doing this crap.

## Don't include anything at all
i.e. (in user code)
```
#include "name.h"
#include <wlr/some-header.h>
```
### Cons:
- Still duplicate effort, since we've already generated the code/headers.
- Still annoying for users

### Pros:
- (Compared to above) Fewer asinine restrictions placed on generated protocol headers.

Better than above, but still pretty annoying.

## Redefine everything
Remove the dependency on the protocol headers entirely, and redefine everything ourselves.
```
enum wlr_xdg_positioner_anchor {
	WLR_XDG_POSITIONER_ANCHOR_NONE = 0,
	WLR_XDG_POSITIONER_ANCHOR_TOP = 1,
	WLR_XDG_POSITIONER_ANCHOR_BOTTOM = 2,
	WLR_XDG_POSITIONER_ANCHOR_LEFT = 3,
	WLR_XDG_POSITIONER_ANCHOR_RIGHT = 4,
	WLR_XDG_POSITIONER_ANCHOR_TOP_LEFT = 5,
	WLR_XDG_POSITIONER_ANCHOR_BOTTOM_LEFT = 6,
	WLR_XDG_POSITIONER_ANCHOR_TOP_RIGHT = 7,
	WLR_XDG_POSITIONER_ANCHOR_BOTTOM_RIGHT = 8,
};
```
instead of just using `enum xdg_positioner_anchor`.

### Cons:
- Waste of time/code having to copy everything
- Have to waste effort keeping this in sync with upstream

### Pros:
- Cleaner interface
- No stupid requirement for users

I really hate redefining types if I don't have to. I consider this basically on the level of those annoying as hell libraries that provide their own integer typedefs for no good reason. We don't need to slap our own wlr_ prefix on everything and is generally just a waste of time and effort

## Install headers (proposed solution)

### Cons:
- Have to install code which arguably isn't ours

### Pros:
- No stupid requirements for users
- No maintenance cost

And now what I think is the best (or least worst) solution. Easy for users and easy for us.
I think it's perfectly justified to install these headers, even if we didn't write them. I consider them more of an implementation detail, but due to the fact we need to use protocol enums, they really are part of our public API.

SirCmpwn:
# Status quo

**Pros**

- Sole source of truth for protocol definitions
- You need libwayland and its headers to build a wayland compositor anyway, it's not like wayland-scanner isn't going to be there
- If the user has any private protocols they're going to need wayland-scanner anyway
- Some people aren't going to use them anyway, e.g. way-cooler, waymonad, which use their own respective scanners

The burden on the users is quite minimal. wlroots has always been and will always be a batteries-not-included library.

ascent12:
>You need libwayland and its headers to build a wayland compositor anyway, it's not like wayland-scanner isn't going to be there
>If the user has any private protocols they're going to need wayland-scanner anyway

I'm not trying to make it so a user never needs to use wayland-scanner. Obviously they're going to need to use it for any private/extra protocols that wlroots doesn't handle.

>Some people aren't going to use them anyway, e.g. way-cooler, waymonad, which use their own respective scanners

wayland scanner is still needed if they're going to generate anything from our headers at all, because our headers won't even parse if the protocol headers are not there.

>Sole source of truth for protocol definitions

Considering how many times that the wlr-protocol headers have been copied between various repositories, I definitely wouldn't say that there is single source of truth right now. Actually in that regard, the proposed solution is slightly better, because you'll at least get the protocol definition that wlroots was built with. This is besides the point, though.

The "don't include anything at all" solution has all of the "advantages" you've listed, without the dumb requirements of header names and include paths. I still don't like that solution, but I object to it far less than what it is right now.

SirCmpwn:
>Considering how many times that the wlr-protocol headers have been copied between various repositories, I definitely wouldn't say that there is single source of truth right now

This is bad but I'd sooner make wlr-protocols installable than change this.

>Actually in that regard, the proposed solution is slightly better, because you'll at least get the protocol definition that wlroots was built with

Yeah but if wlroots is using one protocol version and you're using another things are going to break which is Bad

ascent12:
>Yeah but if wlroots is using one protocol version and you're using another things are going to break which is Bad

I don't see how the current solution prevents that at all.

SirCmpwn:
Well presumably you're then going to be building from the same protocols e.g. in /usr/share/wayland-protocols (or /usr/share/wlr-protocols if we go that route).

ascent12:
Status quo / don't install headers:
- Protocol X version is 1
- wlroots is built and installed to the system
- Protocol X is updated to version 2; wlroots has not been rebuilt
- User code is built against protocol X version 2 and wlroots
- Version mismatch

vs

Redefine everything/install headers
- Protocol X is version 1
- wlroots is built and installed to the system
- Protocol X is updated to version 2; wlroots has not been rebuilt
- User code is built against wlroots and uses the inbuilt version 1
- No version mismatch

As I see it, this is actually an argument _for_ the proposed solution.

SirCmpwn:
>User code is built against wlroots and uses the inbuilt version 1

But it won't, we don't expose the symbols from generated protocols (and we're damn well not about to start)

ascent12:
>we don't expose the symbols from generated protocols

We don't really need to. This is really about generated enum values and the like.
Also, wlroots is providing/linking the generated .c file, so the user shouldn't be generating that anyway.

SirCmpwn:
>Also, wlroots is providing/linking the generated .c file, so the user shouldn't be generating that anyway.

That's not how it works

Listen, is this *really* the hill you want to die on?

ascent12:
>That's not how it works

https://github.com/swaywm/wlroots/blob/master/protocol/meson.build#L65
What on earth is this then?

>Listen, is this really the hill you want to die on?

My main goal is to get rid of the extremely stupid requirements of specific header file names and include paths.
I find that to be completely unacceptable for a library to do.

SirCmpwn:
>https://github.com/swaywm/wlroots/blob/master/protocol/meson.build#L65

*We* fucking link to it, yes, but the end user doesn't get those fucking symbols, jesus christ

ammen99:
@SirCmpwn tbh I don't think those symbols (from protocol code files) are needed anyway ... wlroots wraps everything so the user ends up using the `wlr_*` functions

emersion:
>Don't include anything at all

This won't be possible: forward declaring enums is not allowed.

Issue #725, opened 2018-03-15
Tag generation makes build fail when a file is deleted


emersion:
When deleting a file while developing, the build always fails because the meson script generating tags uses git to list files.

Also happens when moving a file.

acrisci:
My vote is to remove tag generation.

acrisci:
Actually @agx did the implementation so maybe he's using it and will fix it.

agx:
On Fri, Mar 23, 2018 at 01:14:37AM +0000, Tony Crisci wrote:
> Actually @agx did the implementation so maybe he's using it and will fix it.
I'll have a look.

agx:
Does freebsd have '-f' - is so we can use that instead

agx:
Oh, I see it has. PR forthcoming.

PR #1136, opened 2018-07-17
Remove unused types/wlr_tablet_v2.c

emersion:
Related to https://github.com/swaywm/wlroots/issues/725


agx:
Can you provide some steps to reproduce? If I remove ``types/wlr_tablet_v2.c`` from master and rebuild I don't have any issues.

PR #1137, opened 2018-07-17
Remove tag generation

emersion:
Tag generation fails when a file is deleted is this is annoying, see https://github.com/swaywm/wlroots/pull/1136.

Fixes https://github.com/swaywm/wlroots/issues/725

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -217,31 +217,3 @@ pkgconfig.generate(
 	name: meson.project_name(),
 	description: 'Wayland compositor library',
 )
-
-git = find_program('git', required: false)
-if git.found()
-	all_files = run_command(
-		git,
-		['--git-dir=@0@/.git'.format(meson.current_source_dir()),
-		 'ls-files',
-		 ':/*.[ch]'])
-	all_files = files(all_files.stdout().split())
-
-	etags = find_program('etags', required: false)
-	if etags.found() and all_files.length() > 0
-		custom_target('etags',
-			build_by_default: true,
-			input: all_files,
-			output: 'TAGS',
-			command: [etags.path(), '-o', 'TAGS'] + all_files)
-	endif
-
-	ctags = find_program('ctags', required: false)
-	if ctags.found() and all_files.length() > 0
-		custom_target('ctags',
-			build_by_default: true,
-			input: all_files,
-			output: 'tags',
-			command: [ctags.path(), '-o', 'tags'] + all_files)
-	endif
-endif


agx:
Can you provide some steps to reproduce? If I remove ``types/wlr_tablet_v2.c`` from master and rebuild I don't have any issues.

emersion:
Are you sure you have `ctags` installed?

```
Program git found: YES (/usr/bin/git)
Program etags found: NO
Program ctags found: YES (/usr/bin/ctags)
```

When I `rm types/wlr_tablet_v2.c`, I get:

```
ninja: Entering directory `build'
ninja: error: '../types/wlr_tablet_v2.c', needed by 'tags', missing and no known rule to make it
```

Using `git rm` doesn't fix it. The only way to make the build not to fail is to commit the changes, which makes deleting a file really cumbersome.

agx:
Hi,
On Tue, Jul 17, 2018 at 03:59:06AM -0700, emersion wrote:
>    Are you sure you have ctags installed?
> 
>  Program git found: YES (/usr/bin/git)
>  Program etags found: NO
>  Program ctags found: YES (/usr/bin/ctags)
> 
>    When I rm types/wlr_tablet_v2.c, I get:
> 
>  ninja: Entering directory `build'
>  ninja: error: '../types/wlr_tablet_v2.c', needed by 'tags', missing
>  and no known rule to make it

I have both ctags and etags and doing:

    $ rm types/wlr_tablet_v2.c ; touch types/wlr_list.c ; ninja -C _build/

works fine altough it should trip up ninja:

    $ git ls-files :/*.[ch] | grep wlr_tablet_v2.c
    types/wlr_tablet_v2.c

since it's a dependency for the custom_targets.

>    Using git rm doesn't fix it. The only way to make the build not to fail is
>    to commit the changes, which makes deleting a file really
>    cumbersome.

Besides resorting to find . -name '*.[ch]' instead of git-ls I don't
think there's much we could do here.

SirCmpwn:
I don't want this removed, just fixed. I use ctags.

emersion:
I don't use them, and I don't want this bug. I've been trying to fix it for people using it but it doesn't seem like there's a good fix.

I'll just end up adding a config option to disable this shit.

SirCmpwn:
Shell out to a script which tests that each file exists before passing them to ctags? Use find instead of git ls-dir? I can think of a hundred solutions to this offhand, aren't you a Unix user?

emersion:
These are not _good_ solutions.

SirCmpwn:
What? Why not?

emersion:
`find` will include files in the build dir, which doesn't have a fixed name. Testing if each file exists doesn't work when a new file is added. Not to mention these are slow.

agx:
Hi,
On Tue, Jul 17, 2018 at 05:16:11AM -0700, Drew DeVault wrote:
>    Shell out to a script which tests that each file exists before passing
>    them to ctags? Use find instead of git ls-dir? I can think of a hundred
>    solutions to this offhand, aren't you a Unix user?

A "find . -name '*.[ch]'" as suggested in

  https://github.com/swaywm/wlroots/pull/1137#issuecomment-405546837

?

SirCmpwn:
>find will include files in the build dir, which doesn't have a fixed name

Let it, sometimes I need to read the protocol implementations too. These aren't really slow on our scale, it takes 0.27s for me to run `find . -name "*'c" -or -name "*-h"`

agx:
Hi,
On Tue, Jul 17, 2018 at 02:21:33PM +0200, Guido Günther wrote:
> Hi,
> On Tue, Jul 17, 2018 at 05:16:11AM -0700, Drew DeVault wrote:
> >    Shell out to a script which tests that each file exists before passing
> >    them to ctags? Use find instead of git ls-dir? I can think of a hundred
> >    solutions to this offhand, aren't you a Unix user?
> 
> A "find . -name '*.[ch]'" as suggested in
> 
>   https://github.com/swaywm/wlroots/pull/1137#issuecomment-405546837
> 
> ?
I'm hesitant to propose a PR since I can't reproduce the current problem
so I wouldn't be able to verify it improves things.

PR #1145, opened 2018-07-18
xwayland: handle override_redirect flag changes

emersion:
The override_redirect flag can change on configure notify and
on map notify. This adds an event to know when it changes.

This removes wlr_xwayland_surface_is_unmanaged which was wrongly
using the window type to decide whether the view should be
unmanaged.

A similar patch was proposed to Weston, but has never been
merged upstream [1].

[1]: https://patchwork.freedesktop.org/patch/211161/

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -150,6 +150,7 @@ struct wlr_xwayland_surface {
 		struct wl_signal set_parent;
 		struct wl_signal set_pid;
 		struct wl_signal set_window_type;
+		struct wl_signal set_override_redirect;
 		struct wl_signal ping_timeout;
 	} events;
 
@@ -199,9 +200,6 @@ void wlr_xwayland_surface_set_fullscreen(struct wlr_xwayland_surface *surface,
 
 void wlr_xwayland_set_seat(struct wlr_xwayland *xwayland,
 	struct wlr_seat *seat);
-
-bool wlr_xwayland_surface_is_unmanaged(
-	const struct wlr_xwayland_surface *surface);
 
 bool wlr_surface_is_xwayland_surface(struct wlr_surface *surface);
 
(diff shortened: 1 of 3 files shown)


emersion:
>However, this isn't the case for many others - menus, tooltips, etc.

These need to receive keyboard focus too (you can navigate in a right-click menu using your keyboard for instance).

The issue was that we unfocused surfaces when giving keyboard focus to override redirect windows in Sway. See the Sway PR: https://github.com/swaywm/sway/pull/2300

emersion:
Note: in i3 unmanaged windows are _exactly_ windows that have the override_redirect flag.

Ongy:
Did you test dunst? Or will that grab your keyboard away from the currently focused window?

The initial unmanaged thingy there was to behave correctly around that, while being able to give focus to dmenu

Probably not the most important application with macko around, but that was the one I had in mind at that point.

emersion:
>Did you test dunst? Or will that grab your keyboard away from the currently focused window?

So… It won't grab away focus if the currently focused window is an xwayland window, but it will if it's a wayland one. Honestly I don't want to support this special case, because when the window type is set to notification it doesn't necessarily means that key events are supposed to passthrough (dunst still gets keyboard events for _some_ keyboard shortcuts) and X11 clients set the notification type even if it's not a notification (I've seen it set for menus for instance).

Ongy:
I think dunst doesn't get input on the window, but just sniffs input (the usual global hotkey kind). Which of course doesn't work on wayland (unless an X window is focused).

I'm ok with not supporting this ATM since my main usecause gets managed by mako. But I did always intend this to be supplementary to the OR flag, not a replacement.

But this code isn't lost, it's in the history and known. If we find another usecase, it shouldn't be hard to restore as needed.

emersion:
>but just sniffs input

Well it's not clear how to detect if an xwayland window is sniffing or actually wants input. In any case using the window type to do this is incorrect.

The old code exploited the fact that we were missing some OR events, so dunst wasn't getting the OR flag (even if the client requested it) but had the window type hints. Now both menus and dunst have the OR flag and the window type hints.

PR #1152, opened 2018-07-21
Use posix_fallocate when available

emersion:
* `posix_fallocate` was never used
* ~~Use `memfd_create` and seals when possible~~

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,6 +130,10 @@ else
 	exclude_headers += 'xwayland.h'
 endif
 
+if cc.has_header_symbol('fcntl.h', 'posix_fallocate', prefix: '#define _POSIX_C_SOURCE 200112L')
+	conf_data.set('WLR_HAS_POSIX_FALLOCATE', true)
+endif
+
 includedir = get_option('includedir')
 exclude_headers += 'meson.build'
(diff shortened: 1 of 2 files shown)


emersion:
Wait, these checks are wrong:

```
Header <sys/memfd.h> has symbol "memfd_create": NO
Header <sys/memfd.h> has symbol "MFD_ALLOW_SEALING": NO
```

Hmm, what's wrong here?

emersion:
>It might actually be defined in some dumb GNU-specific header that memfd.h pulls in

Ah, it's defined in `mman.h` only if `_GNU_SOURCE` is enabled. Two solutions:

1. Don't use `memfd_create` and seals
2. Redefine constants we need and use a syscall, see https://gitlab.gnome.org/GNOME/gtk/blob/master/gdk/wayland/gdkdisplay-wayland.c#L1207

SirCmpwn:
I'm okay with either.

emersion:
Let's just fix that `posix_fallocate` detection for now.

emersion:
#1152 and #1153 can't be merged together, because this PR uses `#include <wlr/config.h>`. When one of those gets merged I'll fix the other one.

emersion:
Updated

SirCmpwn:
Thanks!

Issue #1096, opened 2018-06-28
All globals should have a destroy event


emersion:
Globals are destroyed with the display. However, if the compositor sets up a listener for a global (e.g. `new_surface` on `wlr_compositor`) it needs to remove the listener before the global is destroyed. It can't use the display destroy event since there's no ordering guarantee.

Thanks @martinetd for bringing this up in https://github.com/swaywm/wlroots/pull/1092#discussion_r198806702

emersion:
We also need to decide what to do when the compositor wants to destroy the global.

* We currently destroy all associated resources. This effectively cleans up everything, but will disconnect clients that try to use the destroyed resources afterwards.
* We could make resources inert instead, but that requires to accommodate *all* interfaces to support inert resources.
* We shouldn't destroy the global while letting the resources alive and non-inert. This would just be asking for trouble.
* We could remove the destroy function, and keep destroying the global when the display is. In this case, I think keeping destroying resources like we currently do is fine.

Ongy:
> We shouldn't destroy the global while letting the resources alive and non-inert. This would just be asking for trouble.

Some protocols (e.g. tablet-v2) require something like this.
The actual global should become inert, but everything created from the global is explicitly not destroyed after the global got removed.

emersion:
Fixed by #1200

emersion:
>Some protocols (e.g. tablet-v2) require something like this.
The actual global should become inert, but everything created from the global is explicitly not destroyed after the global got removed.

When I say "destroy the global", I refer to the `wl_global` on the compositor side, not a specific client's global.

PR #1157, opened 2018-07-22
Implement wlr-gamma-control-unstable-v1

### LINKED ISSUE #1135: Gamma control not working when gamma size is too big
@minus7 has a graphics card with a big gamma size and this is causing issues with `redshift`:

```
Data too big for buffer (24596 > 4096).
Error sending request: Argument list too long
```

`WAYLAND_DEBUG=1` reveals:

```
[553959.018] gamma_control@3.gamma_size(4096)
```

So the gamma size is too large, and fills Wayland's internal send buffer.

Wayland is not designed to send big chunks of data. The gamma control protocol needs to be updated to use e.g. FDs to provide gamma ramps.

emersion:
See https://github.com/swaywm/wlr-protocols/pull/25

- [x] Reset gamma ramps
- [x] Consider setting O_NONBLOCK on the fd before reading

Fixes #1135

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -228,19 +228,12 @@ static bool drm_connector_swap_buffers(struct wlr_output *output,
 	return true;
 }
 
-static bool drm_connector_set_gamma(struct wlr_output *output,
-		uint32_t size, uint16_t *r, uint16_t *g, uint16_t *b) {
-	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
-	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)output->backend;
-
-	bool ok = false;
-	if (conn->crtc) {
-		ok = drm->iface->crtc_set_gamma(drm, conn->crtc, r, g, b, size);
-		if (ok) {
-			wlr_output_update_needs_swap(output);
-		}
+static void fill_empty_gamma_table(uint32_t size,
+		uint16_t *r, uint16_t *g, uint16_t *b) {
+	for (uint32_t i = 0; i < size; ++i) {
+		uint16_t val = (uint32_t)0xffff * (uint32_t)i / (uint32_t)(size - 1);
+		r[i] = g[i] = b[i] = val;
 	}
-	return ok;
 }
 
 static uint32_t drm_connector_get_gamma_size(struct wlr_output *output) {
@@ -252,6 +245,37 @@ static uint32_t drm_connector_get_gamma_size(struct wlr_output *output) {
 	}
 
 	return 0;
+}
+
+static bool drm_connector_set_gamma(struct wlr_output *output,
+		uint32_t size, uint16_t *r, uint16_t *g, uint16_t *b) {
+	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
+	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)output->backend;
+
+	if (!conn->crtc) {
+		return false;
+	}
+
+	uint16_t *reset_table = NULL;
+	if (size == 0) {
+		size = drm_connector_get_gamma_size(output);
+		reset_table = malloc(3 * size * sizeof(uint16_t));
+		if (reset_table == NULL) {
+			wlr_log(WLR_ERROR, "Failed to allocate gamma table");
+			return false;
+		}
+		r = reset_table;
+		g = reset_table + size;
+		b = reset_table + 2 * size;
+		fill_empty_gamma_table(size, r, g, b);
+	}
+
+	bool ok = drm->iface->crtc_set_gamma(drm, conn->crtc, r, g, b, size);
+	if (ok) {
+		wlr_output_update_needs_swap(output);
+	}
+	free(reset_table);
+	return ok;
 }
 
 static bool drm_connector_export_dmabuf(struct wlr_output *output,
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -228,19 +228,19 @@ static bool drm_connector_swap_buffers(struct wlr_output *output,
 	return true;
 }
 
-static void drm_connector_set_gamma(struct wlr_output *output,
+static bool drm_connector_set_gamma(struct wlr_output *output,
 		uint32_t size, uint16_t *r, uint16_t *g, uint16_t *b) {
 	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
 	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)output->backend;
-	bool ok;
 
+	bool ok = false;
 	if (conn->crtc) {
 		ok = drm->iface->crtc_set_gamma(drm, conn->crtc, r, g, b, size);
 		if (ok) {
 			wlr_output_update_needs_swap(output);
 		}
 	}
-
+	return ok;
 }
 
 static uint32_t drm_connector_get_gamma_size(struct wlr_output *output) {
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -231,7 +231,7 @@ static bool drm_connector_swap_buffers(struct wlr_output *output,
 static void fill_empty_gamma_table(uint32_t size,
 		uint16_t *r, uint16_t *g, uint16_t *b) {
 	for (uint32_t i = 0; i < size; ++i) {
-		uint16_t val = (uint32_t)0xffff * (uint32_t)i / (uint32_t)(size - 1);
+		uint16_t val = (uint32_t)0xffff * i / (size - 1);
 		r[i] = g[i] = b[i] = val;
 	}
 }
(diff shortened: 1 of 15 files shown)


types/wlr_gamma_control_v1.c
```diff
@@ -0,0 +1,244 @@
...
+	uint16_t *r = table;
+	uint16_t *g = table + ramp_size;
+	uint16_t *b = table + 2 * ramp_size;
+
+	bool ok = wlr_output_set_gamma(gamma_control->output, ramp_size, r, g, b);
+	free(table);
+	if (!ok) {
+		zwlr_gamma_control_v1_send_failed(gamma_control->resource);
+		gamma_control_destroy(gamma_control);
+		return;
```
SirCmpwn: This return is not necessary. Maybe there should be a few gotos somewhere?

backend/drm/drm.c
```diff
@@ -228,30 +228,54 @@ static bool drm_connector_swap_buffers(struct wlr_output *output,
 	return true;
 }
 
-static void drm_connector_set_gamma(struct wlr_output *output,
-		uint32_t size, uint16_t *r, uint16_t *g, uint16_t *b) {
+static void fill_empty_gamma_table(uint32_t size,
+		uint16_t *r, uint16_t *g, uint16_t *b) {
+	for (uint32_t i = 0; i < size; ++i) {
+		uint16_t val = (uint32_t)0xffff * (uint32_t)i / (uint32_t)(size - 1);
```
ascent12: Why does this have so much casting? `i` and `size` are `uint32_t`s anyway.

emersion: Yeah, true. Fixed.

I kept `uint32_t)0xffff` though.

agx:
A example client would be totally awesome.

martinetd:
@minus7 implemented it for redshift there https://github.com/minus7/redshift/commit/b98000947a73f863975e8c1ae6abe9834ed272b7

agx:
@martinetd thanks! I was actually thinking about a separate client in examples. If wlroots "drives" the protocols (like also with layer-surfaces) it's great to have a test client in wlroots as well so others wanting to implement that protocol (say weston, mutter, ...) don't have to run around and grab different sources. id not that's perfectly fine but having it might help adoption.

emersion:
>A example client would be totally awesome.

I considered doing that, but I've realized computing the gamma ramps for a given temperature is non-trivial, so I wonder if we want this in `examples/`. See https://github.com/jonls/redshift/blob/master/src/colorramp.c

agx:
I noticed that while working on the client for reading gamma ramps:

https://github.com/swaywm/wlroots/pull/1059/files#diff-ed82c9cad3a00609d7ecb9d9eca3c333

but wondered if a sample client couldn't e.g. just half all the values. The main thing is to have something different than the default color ramp. Now that values get reset when the client exits we're sure to get a reasonable display again.

agx:
That said I'm not trying to shove any more work your way, I might just do that at a later point when I have some more time for compositor work again.

emersion:
Right, I could just go with a simple client that changes the brightness. Good idea, I'll do that.

progandy:
You could even add contrast, and a simple gamma value
```A = c * E^y + b```

**Comment by emersion** [comment]
@progandy Thanks, this seems easy enough for an example :)

**Comment by atomnuker** [comment]
Could you make an example client to force linear gamma? For hdmi, which some drivers force limited range rgb for (something which shouldn't exist). Should be even simpler.

**Comment by emersion** [comment]
>Could you make an example client to force linear gamma?

Actually that's how I reset gamma tables. Maybe we should just always set a linear gamma?

I've been searching for other ways to reset gamma, but haven't found one (setting an empty table is a no-op).

**Comment by agx** [comment]
@emersion liniear looks good when you don't have any other color temperature to match against:

https://github.com/GNOME/gnome-settings-daemon/blob/master/plugins/color/gsd-color-state.c#L551

**Comment by emersion** [comment]
Added an example client. There's a pretty high probability I messed up the function filling the gamma tables, @progandy thoughts?

File: examples/gamma-control.c [hunk 3]
```diff
@@ -115,21 +115,26 @@ static void fill_gamma_table(uint16_t *table, uint32_t ramp_size,
 	uint16_t *b = table + 2 * ramp_size;
 	for (uint32_t i = 0; i < ramp_size; ++i) {
 		double val = (double)i / (ramp_size - 1);
-		val = contrast * pow(val, 1.0 / gamma) + brightness;
+		val = contrast * pow(val, 1.0 / gamma) + (1 - brightness);
```
progandy: ```(brightness - 1)``` would be better. 
In my opinion ```brightness = 0``` should result in a black image, ```2``` pure white.

emersion: Oh right

include/wlr/types/wlr_gamma_control_v1.h
```diff
@@ -0,0 +1,31 @@
...
+};
+
+struct wlr_gamma_control_v1 {
+	struct wl_resource *resource;
+	struct wlr_output *output;
+	struct wl_list link;
+
+	struct wl_listener output_destroy_listener;
+
+	void* data;
```
minus7: Nitpick: placement of `*`. Is there no auto formatter?

types/wlr_gamma_control_v1.c
```diff
@@ -0,0 +1,265 @@
...
+
+	off_t fd_size = lseek(fd, 0, SEEK_END);
+	// Skip checks if kernel does no support seek on buffer
+	if (fd_size != -1 && (size_t)fd_size != table_size) {
+		wl_resource_post_error(gamma_control_resource,
+			ZWLR_GAMMA_CONTROL_V1_ERROR_INVALID_GAMMA,
+			"The gamma ramps don't have the correct size");
+		goto error_fd;
+	}
+	lseek(fd, 0, SEEK_SET);
```
minus7: I'd just skip the whole seeking because you test whether enough data was read later anyway. Also, why use a different error here than after failed `read()`?

minus7:
Otherwise *looks* good to me, but gamma changes don't apply instantly. They only apply after switching TTY or DPMS off/on'ing.

types/wlr_gamma_control_v1.c
```diff
@@ -0,0 +1,265 @@
...
+
+void wlr_gamma_control_manager_v1_destroy(
+		struct wlr_gamma_control_manager_v1 *manager) {
+	if (!manager) {
+		return;
+	}
+	wl_list_remove(&manager->display_destroy.link);
+	struct wlr_gamma_control_v1 *gamma_control, *tmp;
+	wl_list_for_each_safe(gamma_control, tmp, &manager->controls, link) {
+		wl_resource_destroy(gamma_control->resource);
```
minus7: Does this destroy the `gamma_control`s the client's still keeping a handle on?

emersion: Yes. See https://github.com/swaywm/wlroots/issues/1096#issuecomment-408670983

types/wlr_gamma_control_v1.c
emersion: This check was originally here because I wasn't setting `O_NONBLOCK`, but I guess now it can be merged with the `read` check. The error is different because a `read` can fail for other reasons than the file being too short (in which case the client did nothing wrong and we shouldn't send a protocol error), but I can handle both cases.

minus7:
Can anyone with amdgpu try if the gamma curve applies instantly?

Emantor:
Yep, I'll test when I get home.

SirCmpwn:
On it.

SirCmpwn:
I can reproduce it like @minus7 specified, only after tty switch (rootston).

SirCmpwn:
Works with legacy DRM. I think we should use legacy DRM for setting the gamma ramps for now and file a driver bug, maybe check our implementation against [weston](https://patchwork.freedesktop.org/series/45255/) first.

minus7:
Confirming that it works with legacy. If working around the problem causes the driver not to fix the problem then I'm against working around it. Preferrably I'd enable the workaround with an env var, so affected users can work around the problem for the time being.

emersion:
Yeah, I'd prefer not to add a workaround for a bug in _one_ driver too. In the meantime, you can already use the env var `WLR_DRM_NO_ATOMIC=1` if you're affected and want to use redshift.

SirCmpwn:
Can we get a rebase? Then I'm prepared to merge this.

minus7:
> In the meantime, you can already use the env var WLR_DRM_NO_ATOMIC=1 if you're affected and want to use redshift.

I'd prefer not to set everything to the old backend, but okay

SirCmpwn:
https://bugs.freedesktop.org/show_bug.cgi?id=107459

emersion:
Rebased.

SirCmpwn:
Thanks!

PR #1159, opened 2018-07-22
xdg-shell(-v6): add set_title and set_app_id toplevel signals

ammen99:
This is useful for example when rendering decorations

I haven't tested much, because I'm not aware of any clients that change app_id. But I successfully managed to get title updates from gedit, so I think it should work fine.


emersion:
Thanks!

PR #1160, opened 2018-07-23
Tablet grabs

Ongy:
Add grabs to the tablet-v2 input device structs.

Also use the tablet-tool grab to implement the logic to *not* change focus when a button is held or the tip is pressed down.

* ~Currently buttons aren't tracked, only the tip works for now.~

Testplan:

Use an application that gives feedback on tablet-tools (e.g. the example client in https://github.com/Ongy/wlroots/tree/tablet-v2-example) [which I should merge at some point] and start two instances.
Hold pen over 1 window and connect tip. Move over other and notice that focus stays with the first.

Release pen over 2nd window and focus should be updated automatically (though no rotate/tilt/pressure/motion etc.)

### DIFF
--- a/include/wlr/types/wlr_tablet_v2.h
+++ b/include/wlr/types/wlr_tablet_v2.h
@@ -10,6 +10,14 @@
 /* This can probably be even lower,the tools don't have a lot of buttons */
 #define WLR_TABLET_V2_TOOL_BUTTONS_CAP 16
 
+struct wlr_tablet_pad_v2_grab_interface;
+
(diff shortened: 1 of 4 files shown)


Ongy:
Note: On IRC @emersion and me talked about the semantics this should have, when the tool is outside the actual surface.

Right now, motion events are dropped, though we though, that it might be better to still send events, to e.g. allow pulling a scroll bar.

emersion:
Thanks!

PR #1164, opened 2018-07-29
Revert "rootston: Reap child processes"

emersion:
This reverts commit b6ed1f29a4dbba93eb53c32ec5492db8ee1d9343.

This commit breaks xwayland.

PR #1165, opened 2018-07-29
reintroduce xwayland is_unmanaged

Ongy:
https://github.com/swaywm/wlroots/pull/1145 removed the
wlr_xwayland_is_unamanged function while fixing OR, because it was
belieived that it's supposed to work around the broken OR handling.

This was a misunderstanding. is_unmanaged is (while sort of a hack)
intended to work around inherent differences between "real" X sessions
and our Xwayland/wayland situation.

The main reason it exists is to support applications like rofi and dzen,
while not handing focus to other OR windows (which should *not* be
required).
Traditionally, these applications just grabbed input from X and didn't
need to be focused by any logic in the WM. Which of course doesn't work
in wayland compositors. So we have to give them focus in some way.
Giving *every* OR window focus, breaks other applications that don't
expect focus to change.

A testcase that was pointed out to me where wlr_xwayland_is_unamanged was
breaking things is https://github.com/swaywm/sway/issues/2128 (syncplay,
gitk, gitgui)
Supposedly it broke using keyboard to navigate the menus.
I can't reproduce this with this patch. The popups can be navigated as
long as the parent has focus.

Testing:
Requires to merge https://github.com/swaywm/wlroots/pull/1164 since Xwayland can't be used at all without.

As I hope the commit coment points out, this is a focus issue. With this patch applied, dmenu/rofi will work fine. I will retract (or improve) when I get a counter-example. @emersion  you are up! 

`gitk` is somewhat broken, but thats unrelated to this patch. (Moving cursor to different menu while menu is shown is funky)

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -209,4 +209,29 @@ struct wlr_xwayland_surface *wlr_xwayland_surface_from_wlr_surface(
 
 void wlr_xwayland_surface_ping(struct wlr_xwayland_surface *surface);
 
+/** Metric to guess if an OR window should "receive" focus
+ *
+ * In the pure X setups, window managers usually straight up ignore override
+ * redirect windows, and never touch them. (we have to handle them for mapping)
+ *
+ * When such a window wants to receive keyboard input (e.g. rofi/dzen) it will
+ * use mechanics we don't support (sniffing/grabbing input).
+ * [Sadly this is unrelated to xwayland-keyboard-grab]
+ *
+ * To still support these windows, while keeping general OR semantics as is, we
+ * need to hand a subset of windows focus.
+ * The dirty truth is, we need to hand focus to any Xwayland window, though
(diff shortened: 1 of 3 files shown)


Ongy:
I always have the best thoughts, when I leave the PC:

It would probably be closer to reality to have a "fake" X window managed by the compositor that receives focus, since it's not actually the window, but Xwayland that needs to receive inputs. But keeping it on the window should provide more feasable with handling focus changes once it unmapps.

emersion:
LGTM. Can you add a comment to the function explaining how to use it, when to use it? Maybe we should consider renaming it too: rofi is unmanaged.

ammen99:
LGTM, I also think we need to have this distinction

Ongy:
Added a bunch of comment to the header, renamed to something that hopefully makes it more obvious and inverted the logic to fit the name better.

SirCmpwn:
Thanks!

PR #1166, opened 2018-07-29
examples: cleanup simple

emersion:
- Move a log to where it should be
- Enforce code style
- Rename wlr to backend


SirCmpwn:
Thanks!

PR #1167, opened 2018-07-29
rootston: focus newly-created surfaces

### LINKED ISSUE #1162: Dropdowns (xdg-popups) sometimes close when the pointer button is released
Steps to reproduce:
Open gnome-calculator, switch to advanced mode. Try to click on the second dropdown (the one on the right). Popup opens, but as soon as the user releases the pointer button (the one used to click on the dropdown) the popup closes.

The same thing happens with some other popups, for example in simple-scanner's preferences.

 This bug is specific to rootston, and doesn't happen in sway. I haven't been able to pin down the exact cause yet.

ammen99:
Whenever a new surface is created, we have to update the cursor focus,
even if there's no input event. So, we generate one motion event, and
reuse the code to update the proper cursor focus. We need to do this
for all surface roles - toplevels, popups, subsurfaces.

Fixes #1162

How to test: There are many ways to test, for example open an app and don't move the cursor. The cursor surface should change (and the view should get cursor focus). Or open weston-terminal, right click, hold the button and don't move the mouse - the first menu item should be highlighted.

Further work:
1. I'm not sure, but maybe we should do the same when a surface is unmapped?
2. Port this to sway - it doesn't handle cursor focus properly either

### DIFF
--- a/include/rootston/cursor.h
+++ b/include/rootston/cursor.h
@@ -80,4 +80,7 @@ void roots_cursor_handle_tool_tip(struct roots_cursor *cursor,
 void roots_cursor_handle_request_set_cursor(struct roots_cursor *cursor,
 		struct wlr_seat_pointer_request_set_cursor_event *event);
 
+void roots_cursor_update_position(struct roots_cursor *cursor,
+		uint32_t time);
+
 #endif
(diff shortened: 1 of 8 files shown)


RedSoxFan:
Thanks!

Issue #934, opened 2018-04-30
Multi GPU is broken


emersion:
Workaround: set `WLR_DRM_DEVICES=/dev/dri/card0`

ascent12:
I was asked to do a write up about this issue, and actually looking at the reported message and thinking about it, I've noticed something wrong:

https://github.com/swaywm/wlroots/blob/master/backend/multi/backend.c#L146-L150
Says we can't have multiple renderers, and this is partially true. Each of the secondary GPUs will have their own renderer internally, because they need to do a copy with the dma-buf they receive. This is just an internal detail, and shouldn't be exposed outside of the backend.

Basically, https://github.com/swaywm/wlroots/blob/master/backend/drm/backend.c#L51-L55 is wrong, and should be returning the renderer of the primary GPU, instead of its own renderer, since that's the thing which actually does all of the "real" rendering.

If we're lucky, the fix will just be to change DRM's backend_get_renderer. I'm not in the position to test this out as I'm writing this. I'll try it later.

Before the check in wlr_multi_backend_add was added, the place the error happened was https://github.com/swaywm/wlroots/blob/master/render/gles2/renderer.c#L28. Maybe what I said above would explain why this used to fail too.

Otherwise there is some weird misuse of renderers and EGL contexts somewhere, which needs to be tracked down. I tried looking before, but it certainly wasn't obvious.

---

Another important thing is that this needs to be tested on a wide range of setups. I originally got it working on an AMD HD 7950 and an Nvidia GTX 5XX card, but when it used to work someone did report that it did not work for them. So I'm not really confident that the actual implementation is correct.

I have a couple of more GPUs to work with and test now (AMD R9 2XX and an RX 5XX), so hopefully I can be more confident with the results I get. I'm really not looking forwards to having to keep turning my computer off and replacing GPUs though.

However, I don't have a laptop with a discrete GPU, or even an Intel iGPU desktop to work with (although I may be able to get my hands on one easily). Once this is actually thought to be fixed, any extra help testing would be great. The more hardware setups tested, the better.

---

That said, I'm not particularly happy with the way multi-GPU is currently implemented. Most of it is fine. but it makes the DRM code is fair bit more complicated to follow (it was already complicated as hell). The way it was originally implemented was to basically bolt it on top of the single-GPU code, and reuses variables is weird ways.
I think there needs to be a refactor of DRM to make that whole bit a lot easier to follow.

engineeringdivision-play:
[...]

PR #1168, opened 2018-08-02
Fix support for multiple GPU

manio:
Discussion: https://github.com/swaywm/wlroots/issues/934

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -51,7 +51,12 @@ static void backend_destroy(struct wlr_backend *backend) {
 static struct wlr_renderer *backend_get_renderer(
 		struct wlr_backend *backend) {
 	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)backend;
-	return drm->renderer.wlr_rend;
+
+	if (drm->parent) {
+		return drm->parent->renderer.wlr_rend;
+	} else {
+		return drm->renderer.wlr_rend;
+	}
 }
 
 static struct wlr_backend_impl backend_impl = {
--- a/backend/multi/backend.c
+++ b/backend/multi/backend.c
@@ -143,7 +143,7 @@ bool wlr_multi_backend_add(struct wlr_backend *_multi,
 	struct wlr_renderer *multi_renderer =
 		multi_backend_get_renderer(&multi->backend);
 	struct wlr_renderer *backend_renderer = wlr_backend_get_renderer(backend);
-	if (multi_renderer != NULL && backend_renderer != NULL) {
+	if (multi_renderer != NULL && backend_renderer != NULL && multi_renderer != backend_renderer) {
 		wlr_log(WLR_ERROR, "Could not add backend: multiple renderers at the "
 			"same time aren't supported");
 		return false;
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -116,7 +116,7 @@ static bool gles2_render_texture_with_matrix(struct wlr_renderer *wlr_renderer,
 	struct wlr_gles2_renderer *renderer =
 		gles2_get_renderer_in_context(wlr_renderer);
 	struct wlr_gles2_texture *texture =
-		get_gles2_texture_in_context(wlr_texture);
+		gles2_get_texture(wlr_texture);
 
 	struct wlr_gles2_tex_shader *shader = NULL;
 	GLenum target = 0;
@@ -213,7 +213,7 @@ static const enum wl_shm_format *gles2_renderer_formats(
 static bool gles2_resource_is_wl_drm_buffer(struct wlr_renderer *wlr_renderer,
 		struct wl_resource *resource) {
 	struct wlr_gles2_renderer *renderer =
-		gles2_get_renderer_in_context(wlr_renderer);
+		gles2_get_renderer(wlr_renderer);
 
 	if (!eglQueryWaylandBufferWL) {
 		return false;
@@ -227,7 +227,7 @@ static bool gles2_resource_is_wl_drm_buffer(struct wlr_renderer *wlr_renderer,
 static void gles2_wl_drm_buffer_get_size(struct wlr_renderer *wlr_renderer,
 		struct wl_resource *buffer, int *width, int *height) {
 	struct wlr_gles2_renderer *renderer =
-		gles2_get_renderer_in_context(wlr_renderer);
+		gles2_get_renderer(wlr_renderer);
 
 	if (!eglQueryWaylandBufferWL) {
 		return;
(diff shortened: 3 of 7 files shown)


backend/drm/backend.c
```diff
@@ -51,7 +51,11 @@ static void backend_destroy(struct wlr_backend *backend) {
 static struct wlr_renderer *backend_get_renderer(
 		struct wlr_backend *backend) {
 	struct wlr_drm_backend *drm = (struct wlr_drm_backend *)backend;
-	return drm->renderer.wlr_rend;
+
+	if (drm->parent)
```
emersion: Style: we always use brackets.

Returning the parent's renderer is the Right Thing To Do™.

backend/multi/backend.c
```diff
@@ -140,15 +140,6 @@ bool wlr_multi_backend_add(struct wlr_backend *_multi,
 		return true;
 	}
 
-	struct wlr_renderer *multi_renderer =
```
emersion: I'd really like to keep this.

render/gles2/renderer.c
```diff
@@ -25,7 +25,6 @@ static struct wlr_gles2_renderer *gles2_get_renderer(
 static struct wlr_gles2_renderer *gles2_get_renderer_in_context(
 		struct wlr_renderer *wlr_renderer) {
 	struct wlr_gles2_renderer *renderer = gles2_get_renderer(wlr_renderer);
-	assert(wlr_egl_is_current(renderer->egl));
```
emersion: We still need to assert this for functions called between `wlr_renderer_begin` and `wlr_renderer_end`. I'd suggest keeping this function as is and adding a new `gles2_get_renderer_switch_context` for textures.

backend/multi/backend.c
manio: If I return here (no matter if I return true or false) then the further code for allocating subbackend is not called and this is leading to only support one GPU...

render/gles2/texture.c
```diff
@@ -25,6 +25,8 @@ static struct wlr_gles2_texture *gles2_get_texture(
 struct wlr_gles2_texture *get_gles2_texture_in_context(
 		struct wlr_texture *wlr_texture) {
 	struct wlr_gles2_texture *texture = gles2_get_texture(wlr_texture);
+	if (!wlr_egl_is_current(texture->egl))
```
emersion: This `if` is strange. No need for an assert after switching contexts.

backend/multi/backend.c
emersion: You can add `backend_renderer != multi_renderer`

ascent12:
Hmm, this is a smaller change than I was expecting.
There are still lots of things I want to do to make Multi-GPU better (like removing the renderer copy and fixing the code structure), which is probably why I was being damn slow/lazy when I was looking into this myself instead of looking at the core issue.

I'll get around to testing a few of the various GPU configurations I have access to this weekend.

To anyone else trying this out, pay attention to:
- Hotplugging
- Using non-native resolutions and rotations/reflections
- Whether things like hardware cursors and gamma control work
- The overall performance (stuttering, etc.)
- Different configurations of which GPU is primary (e.g. run with WLR_DRM_DEVICES=/dev/dri/card0:/dev/dri/card1 then WLR_DRM_DEVICES=/dev/dri/card1:/dev/dri/card0)
- Whether this works on all of rootston, sway, and any other wlroots-based compositor
- Whether tri or quad-GPU systems work well (I only have access to motherboards with enough PCI-e space for 2 GPUs)
- Whether it works cross driver (i915 + nouveau, amdgpu + radeon, etc.)

manio:
I've updated the code one more time. I was switching context when I did not have to and this was leading to problems (non visual, but in the log, like):
```[GLES2] GL_INVALID_OPERATION ...```

This is now fixed. All @emersion comments are now took into account.

There are still some problems left:
- Screen mirror + rotation
- Hardware cursor on secondary display

But I'd like to have it covered in another issues/pull requests (step by step).

**Comment by manio** [comment]
Screen rotation fixed :)

File: render/gles2/renderer.c [hunk 5]
```diff
@@ -323,7 +323,7 @@ static struct wlr_texture *gles2_texture_from_dmabuf(
 static void gles2_init_wl_display(struct wlr_renderer *wlr_renderer,
 		struct wl_display *wl_display) {
 	struct wlr_gles2_renderer *renderer =
-		gles2_get_renderer_in_context(wlr_renderer);
+		gles2_get_renderer(wlr_renderer);
```
emersion: Hmm. We should probably switch context here.

render/gles2/renderer.c
```diff
@@ -227,7 +227,7 @@ static bool gles2_resource_is_wl_drm_buffer(struct wlr_renderer *wlr_renderer,
 static void gles2_wl_drm_buffer_get_size(struct wlr_renderer *wlr_renderer,
 		struct wl_resource *buffer, int *width, int *height) {
 	struct wlr_gles2_renderer *renderer =
-		gles2_get_renderer_in_context(wlr_renderer);
+		gles2_get_renderer(wlr_renderer);
```
emersion: Not sure if we should switch context here. @ascent12, thoughts?

render/gles2/renderer.c
```diff
@@ -116,7 +116,7 @@ static bool gles2_render_texture_with_matrix(struct wlr_renderer *wlr_renderer,
 	struct wlr_gles2_renderer *renderer =
 		gles2_get_renderer_in_context(wlr_renderer);
 	struct wlr_gles2_texture *texture =
-		get_gles2_texture_in_context(wlr_texture);
+		gles2_get_texture(wlr_texture);
```
emersion: Can you try adding back this assertion? I'd like to assert that the texture has been created in the same context as the renderer.

ascent12:
I've pushed the fix for the black hardware cursor on the secondary GPU.

render/gles2/renderer.c
manio: Tried one more time after your comment. If I revert this change I have the assert on startup.

render/gles2/renderer.c
manio: I think it should be as it is now. I've also retested reverting this change - after some working in sway and opening some XWayland apps the assert is raising.

manio:
@ascent12 
Thank you for your work! Confirming - the cursor is now working, yay! :)

If you don't have nothing against I think it's now in good shape to merge this PR.

render/gles2/renderer.c
manio: Ah - I probably took you wrong. You are telling about switching context and I was thinking about reverting this change to original function (which is asserting and not switching context, as this was your point regarding rendering functions)... Yeah - I can switch context here as well.

n3rdopolis:
I'm having a little trouble starting this on my qemu VM with multiple GPUs.
The relevant part of my QEMU command is
```
qemu-system-x86_64 -m 2047 -vga none -enable-kvm -device VGA,id=video0  -device secondary-vga,id=video1 -spice port=5930,disable-ticketing -cpu host -cdrom PATHTOMYISO.iso
```
to get two GPUs. and then to see the VM with the two GPUs:
```
remote-viewer spice://127.0.0.1:5930 --spice-preferred-compression=off
```
rootston and sway both abort with
```
sway: ../backend/drm/renderer.c:238: copy_drm_surface_mgpu: Assertation 'tex' failed.
```

Sway can start when there is one GPU, Let me know if you need more info

SirCmpwn:
Let's take a look at qemu separately. Can you open a ticket?

Thanks for the patch, @manio!

PR #1173, opened 2018-08-03
Add missing dependency libinput to types/meson.build

sdilts:
The file `types/tablet_v2/wlr_tablet_v2.c` includes `libinput.h`, and therefore depends on `libinput`. This fixes a compilation error on some systems that is caused by `libinput.h` not being in the include path.


ascent12:
Thanks!

PR #1177, opened 2018-08-05
Check for DRM prime

ascent12:
This will indicate the user properly when multi-GPU is not supported by their GPU driver.

@n3rdopolis 
I suspect that is what is causing https://github.com/swaywm/wlroots/pull/1168#issuecomment-410449388

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -27,6 +27,23 @@
 #include "util/signal.h"
 
 bool check_drm_features(struct wlr_drm_backend *drm) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

n3rdopolis:
Just saw I was pinged here (I did see the commit earlier), this does work with the QEMU's VGA driver (which I primarily use), but as a test, trying with qxl, I get a similar error message as before, but I opened a separate issue for that.

PR #1179, opened 2018-08-06
rootston: add keybinding to toggle decoration mode

emersion:
This is useful for `xdg-decoration` client devs to test decoration mode changes triggered server-side.

Test plan:
* Add `Ctrl+Shift+d = toggle_decoration_mode` to `rootston.ini`
* Open `rootston` and then spawn `build/examples/toplevel-decoration`
* Press the keybinding

### DIFF
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -39,6 +39,8 @@ struct roots_xdg_surface_v6 {
 	uint32_t pending_move_resize_configure_serial;
 };
 
+struct roots_xdg_toplevel_decoration;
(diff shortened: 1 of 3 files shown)


johanhelsing:
I'm not familiar with wlroots code, but looks sensible to me.

Seems to work well with the toplevel-decoration test, as well as my patch for xdg-decoration in Qt clients: https://codereview.qt-project.org/#/c/235936/3

SirCmpwn:
Thanks!

PR #1182, opened 2018-08-07
session: load GPU devices even if they have zero connectors

ammen99:
On some systems (most notably laptops with two GPUs) there are GPUs that
don't have attached outputs. However, we still want to load those GPUs
because they could still be used by the compositor for rendering.

### DIFF
--- a/backend/session/session.c
+++ b/backend/session/session.c
@@ -220,17 +220,9 @@ static int open_if_kms(struct wlr_session *restrict session, const char *restric
 		goto out_fd;
 	}
 
-	if (res->count_crtcs <= 0 || res->count_connectors <= 0 ||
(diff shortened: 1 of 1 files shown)


backend/session/session.c
```diff
@@ -220,8 +220,8 @@ static int open_if_kms(struct wlr_session *restrict session, const char *restric
 		goto out_fd;
 	}
 
-	if (res->count_crtcs <= 0 || res->count_connectors <= 0 ||
-		res->count_encoders <= 0) {
+	if (res->count_crtcs < 0 || res->count_connectors < 0 ||
+		res->count_encoders < 0) {
```
ascent12: I don't think those values would ever be negative.
You may as well just remove the check entirely.

Issue #1170, opened 2018-08-02
Division by zero when using X11 backend


sdilts:
When using the X11 backend, `wlr_matrix_projection` is called with `width=0` and `height=0`, which causes division by zero. This appears to occur when the backend is first starting up, as the initial width and height is set to `(0,0)`. A relevant printout from gdb while running rootson is below. You can see some debug statements that were added to make the behavior more clear. 

``` gdb
Thread 1 "rootston" hit Breakpoint 1, wlr_matrix_projection (mat=0x926a94, width=0, height=0, transform=
    WL_OUTPUT_TRANSFORM_NORMAL) at ../types/wlr_matrix.c:124
124		memset(mat, 0, sizeof(*mat) * 9);
(gdb) print width
$1 = 0
(gdb) print height
$2 = 0
(gdb) n
126		wlr_log(WLR_DEBUG, "About to divide: Width: %d, Height: %d", width, height);
(gdb) n
2018-08-02 11:10:47 - [types/wlr_matrix.c:126] About to divide: Width: 0, Height: 0
128		const float *t = transforms[transform];
(gdb) n
129		float x = 2.0f / width;
(gdb) n
130		float y = 2.0f / height;
(gdb) n
132		wlr_log(WLR_DEBUG, "x: %lf, y: %lf", x, y);
(gdb) backtrace full
#0  0x00007ffff7b97c8b in wlr_matrix_projection (mat=0x926a94, width=0, height=0, transform=
    WL_OUTPUT_TRANSFORM_NORMAL) at ../types/wlr_matrix.c:132
        t = 0x7ffff7bb87e0 <transforms>
        x = inf
        y = inf
#1  0x00007ffff7b9a139 in output_update_matrix (output=0x9269a0) at ../types/wlr_output.c:141
#2  0x00007ffff7b9a316 in wlr_output_update_custom_mode (output=0x9269a0, width=0, height=0, refresh=60000) at ../types/wlr_output.c:187
        resource = 0x7ffff7ff5000
#3  0x00007ffff7b78929 in output_set_refresh (wlr_output=0x9269a0, refresh=60000)
    at ../backend/x11/output.c:36
        output = 0x9269a0
#4  0x00007ffff7b78c80 in wlr_x11_output_create (backend=0x631700) at ../backend/x11/output.c:124
        __PRETTY_FUNCTION__ = "wlr_x11_output_create"
        x11 = 0x631700
        output = 0x9269a0
        wlr_output = 0x9269a0
[...]

PR #1184, opened 2018-08-10
Keep X11 backend windows from having an invalid size

sdilts:
Fixes issue #1170. Sets the initial size for the X11 output window  to `(1024,768)` and makes the backend ignore events that try to set one of the window dimensions to zero.

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -190,11 +190,17 @@ struct wlr_output *wlr_x11_output_create(struct wlr_backend *backend) {
 
 void handle_x11_configure_notify(struct wlr_x11_output *output,
 		xcb_configure_notify_event_t *ev) {
-	wlr_output_update_custom_mode(&output->wlr_output, ev->width,
-		ev->height, output->wlr_output.refresh);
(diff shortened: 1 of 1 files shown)


backend/x11/output.c
```diff
@@ -137,8 +140,8 @@ struct wlr_output *wlr_x11_output_create(struct wlr_backend *backend) {
 	};
 	output->win = xcb_generate_id(x11->xcb_conn);
 	xcb_create_window(x11->xcb_conn, XCB_COPY_FROM_PARENT, output->win,
-		x11->screen->root, 0, 0, 1024, 768, 1, XCB_WINDOW_CLASS_INPUT_OUTPUT,
-		x11->screen->root_visual, mask, values);
+		x11->screen->root, wlr_output->width, wlr_output->height, 1024, 768, 1,
```
emersion: The two values that were previously `0, 0` are x and y, I think those should remain 0. The two values that are `1024, 768` are the window size, which should be set to `wlr_output->{width,height}`.

emersion:
Thanks!

PR #1187, opened 2018-08-13
move wlr_drm_connector_add_mode to wlr/backend/drm.h

ammen99:
It is meant to be used by compositors, so it should be in the public headers.

### DIFF
--- a/include/wlr/backend/drm.h
+++ b/include/wlr/backend/drm.h
@@ -28,6 +28,12 @@ struct wlr_backend *wlr_drm_backend_create(struct wl_display *display,
 bool wlr_backend_is_drm(struct wlr_backend *backend);
 bool wlr_output_is_drm(struct wlr_output *output);
 
+/**
+ * Add mode to the list of available modes
+ */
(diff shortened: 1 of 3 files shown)


include/wlr/backend/drm.h
```diff
@@ -13,6 +13,7 @@
 #include <wlr/backend.h>
 #include <wlr/backend/session.h>
 #include <wlr/types/wlr_output.h>
+#include <xf86drmMode.h>
```
emersion: Maybe it would be better to forward declare `typedef struct _drmModeModeInfo drmModeModeInfo;` to allow users to use this file without having to link with libdrm?

ammen99: Hmm, makes sense

ammen99:
I updated the PR to forward-declare `drmModeModeInfo` and also fixed the rootston include.

emersion:
rootston should now include `xf86drmMode.h` since it uses `wlr_drm_connector_add_mode`

ammen99:
It is included in config.h

emersion:
Thanks!

PR #1194, opened 2018-08-22
Build system changes

### LINKED ISSUE #1082: Use meson feature type for options
Upcoming feature in Meson to add inbuilt support for the common "enabled"/"disabled"/"auto" option when picking dependencies, while adding some bells and whistles compared to doing a combo type which helps packagers and other people from accidentally enabling a dependency.

Despite not actually being in a tagged version of Meson, it's already in their docs:
http://mesonbuild.com/Build-options.html#features
It says it was added in 0.47.0, while 0.46.1 is the current release.

I'm just creating this ticket now before I forget.

While we're breaking the "API" of the build system (i.e. breaking AUR packages), we may as well change our option names to follow the general naming guide and drop the "enabled-" prefix.

ascent12:
This is a pretty large refactoring of the build system to make use of newer meson features.
It bumps the minimum meson version to 0.47.0 and changes our `meson_options.txt`, breaking build scripts.

The individual commits largely reflect each major change, but the main changes are:
- Use new [feature option](https://mesonbuild.com/Build-options.html#features) type for most features. The `enable-` prefix was dropped, so you don't have to do anything silly like `meson configure -Denable-xwayland=enabled`, and is more consistent with [meson's style guide](https://mesonbuild.com/Style-guide.html) anyway.
An advantage of this is that you can explicitly turn all features off/on with a single switch, preventing any accidental linking of dependencies (good for distro maintainers).
- Change how libsystemd/libelogind is selected with a combo option called `logind-provider`. They're mutually exclusive, so I think this is cleaner.
- Install all headers explicitly. I just think this is nicer, but I wouldn't really care that much if someone wanted to revert this.
- Generate `config.h`/`version.h` from `config.h.in`/`version.h.in` instead of from scratch.
This allows us to add comments if we wanted to, and users can see the `/* #undef WLR_HAS_FEATURE */` inside the header. Again, I'm not super concerned if someone opposes this.
- Various small changes; see the commits.

There are some other changes I want to make, but it'll probably need more discussion before they get done:
- Remove all of the internal static libraries that get linked together.
I think these are largely pointless, and just complicate the build system with unnecessary dependency management and boilerplate.
Just build them as a single library; that's how it's going to end up anyway.
- Change the indentation to be consistent with [meson's style guide](https://mesonbuild.com/Style-guide.html).
I think it's better to follow conventions. It's not like our build system needs to be the same as our actual source code.

Closes #1082


ascent12:
Another thing which I wanted to bring up was that I want to remove the ctags/etags from the build system.
I really think it has no business being there, as it's very specific to a developer's own workflow, which I don't think needs to live inside the repository. We don't include things like `*.swp` in `.gitignore`, which I consider to be on the same level.

Not to mention, it's reliant on files being checked into git, which is somewhat flakey.
If someone wants to generate this stuff automatically, just write a git hook. It'll basically get you the same result.

SirCmpwn:
>Another thing which I wanted to bring up was that I want to remove the ctags/etags from the build system

No

>Not to mention, it's reliant on files being checked into git, which is somewhat flakey

So fix this? As mentioned before it's not hard to replace this with `find -name "*'c"`.

emersion:
>Use new feature option type for most features. 

Nice.

>Install all headers explicitly. I just think this is nicer, but I wouldn't really care that much if someone wanted to revert this.

On one hand, we do list all C source files, so it makes sense to do the same for headers. However while it's easy to detect when a C file is missing (linker errors), it's trickier to notice we forgot to add a header to the list.

>Change the indentation to be consistent with meson's style guide.

Meh.

>Another thing which I wanted to bring up was that I want to remove the ctags/etags from the build system.

(Yes.)

All the other things: I don't really care, I'm fine with either.

SirCmpwn:
The changes look good to me. If you want to refactor etags/ctags before merge let me know.

ascent12:
>If you want to refactor etags/ctags before merge let me know

Nah, I'd be fine with this being merged now.

SirCmpwn:
Thanks!

SirCmpwn:
This has been reverted - didn't realize that this breaks the build when wlroots is compiled as a meson subproject.

meson.build
```diff
@@ -18,15 +18,7 @@ so_version = ['0', '0', '0']
...
-	'-DWLR_SRC_DIR="@0@"'.format(meson.source_root()),
-	language: 'c',
-)
-add_project_arguments(
-	'-I@0@'.format(meson.build_root()),
-	language: 'c',
-)
-add_project_link_arguments(
-	'-Wl,-rpath,@0@'.format(meson.build_root()),
+	'-DWLR_SRC_DIR="@0@"'.format(meson.current_source_dir()),
```
martinetd: the breakage comes from this line (source_root to current_source_dir), I think; that'll be different as a submodule.
But frankly the problem is in glgen, trying to include the full path from build dir is insane; something like this fixes @atomnuker's problem:
```diff
diff --git a/glgen.sh b/glgen.sh
index b27750e4..2dc24eec 100755
--- a/glgen.sh
+++ b/glgen.sh
@@ -77,7 +77,7 @@ EOF
 
 cat > $OUT_C << EOF
 #include <wlr/util/log.h>
-#include "$OUT_H"
+#include "$(basename "$OUT_H")"
 $DEFN
 
 bool load_$BASE(void) {
```

ascent12:
After a quick bisect, the commit causing the issue is the -I removal in https://github.com/swaywm/wlroots/pull/1194/commits/869dca1bce662e685a83a33f7550644817265fb9, but it's pointing out a flaw in glgen.sh rather than the commit itself.

Basically, the header is generated in a path from the parent project, which isn't in our include paths, and we can't add it cleanly.

Coming up with a solution isn't hard, but I can't decide which is the best one.

- Use absolute paths
```diff
diff --git a/glgen.sh b/glgen.sh
index b27750e4..c8a9d0c6 100755
--- a/glgen.sh
+++ b/glgen.sh
@@ -7,13 +7,26 @@
 # to fail if it can't load the function. You'll need to check if that function
 # is NULL before using it.
 
+# There is no POSIX-compliant utility to do this directly, so we just have to
+# do it ourselves.
+absolute() {
+    case $1 in
+    /*)
+        echo "$1"
+        ;;
+    *)
+        echo "$PWD/$1"
+        ;;
+    esac
+}
+
 if [ $# -ne 3 ]; then
        exit 1
 fi
 
 SPEC=$1
-OUT_C=$2
-OUT_H=$3
+OUT_C=$(absolute "$2")
+OUT_H=$(absolute "$3")
 
 BASE=$(basename "$SPEC" .txt)
 INCLUDE_GUARD=$(printf %s "$OUT_H" | tr -c [:alnum:] _ | tr [:lower:] [:upper:])
```
Including an absolute path is weird, but perfectly valid.

- Remove the include from glapi.c
```diff
diff --git a/glgen.sh b/glgen.sh
index b27750e4..a73ee0e7 100755
--- a/glgen.sh
+++ b/glgen.sh
@@ -77,7 +77,11 @@ EOF
 
 cat > $OUT_C << EOF
 #include <wlr/util/log.h>
-#include "$OUT_H"
+
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#include <GLES2/gl2.h>
+#include <GLES2/gl2ext.h>
 $DEFN
 
 bool load_$BASE(void) {
```
It'd normally be considered bad practive for a file to not include its own header, but I'm more forgiving to it, since it's machine generated code. It doesn't actually need any facilities from the header.

- Assume that the generated .c and .h will be in the same directory.
@martinetd's solution above:
```diff
diff --git a/glgen.sh b/glgen.sh
index b27750e4..2dc24eec 100755
--- a/glgen.sh
+++ b/glgen.sh
@@ -77,7 +77,7 @@ EOF
 
 cat > $OUT_C << EOF
 #include <wlr/util/log.h>
-#include "$OUT_H"
+#include "$(basename "$OUT_H")"
 $DEFN
 
 bool load_$BASE(void) {
```
I'd also clean up the script arguments and how meson calls it just to remove some redundancy.
The script was written to be somewhat general purpose and not tailored extremely close to how wlroots needs to use it, so it'd be nice to not add that assumption.

PR #1195, opened 2018-08-22
virtual_keyboard: fix virtual keyboard destruction

dcz-purism:
After destroying the virtual keyboard, listeners could still be pointing to its wlr_input_device signals. This patch sets NULL as the active keyboard if the virtual keyboard is destroyed while in use.

### DIFF
--- a/types/wlr_virtual_keyboard_v1.c
+++ b/types/wlr_virtual_keyboard_v1.c
@@ -99,7 +99,7 @@ static void virtual_keyboard_destroy_resource(struct wl_resource *resource) {
 		virtual_keyboard_from_resource(resource);
 	wlr_signal_emit_safe(&keyboard->events.destroy, keyboard);
 	wl_list_remove(&keyboard->link);
-	wlr_keyboard_destroy(keyboard->input_device.keyboard);
+	wlr_input_device_destroy(&keyboard->input_device);
 	free(keyboard);
 }
 
--- a/types/seat/wlr_seat_keyboard.c
+++ b/types/seat/wlr_seat_keyboard.c
@@ -110,7 +110,7 @@ static void handle_keyboard_repeat_info(struct wl_listener *listener,
 static void handle_keyboard_destroy(struct wl_listener *listener, void *data) {
 	struct wlr_seat_keyboard_state *state =
 		wl_container_of(listener, state, keyboard_destroy);
-	state->keyboard = NULL;
+	wlr_seat_set_keyboard(state->seat, NULL);
 }
 
 void wlr_seat_set_keyboard(struct wlr_seat *seat,


types/wlr_virtual_keyboard_v1.c
```diff
@@ -99,6 +99,7 @@ static void virtual_keyboard_destroy_resource(struct wl_resource *resource) {
 		virtual_keyboard_from_resource(resource);
 	wlr_signal_emit_safe(&keyboard->events.destroy, keyboard);
 	wl_list_remove(&keyboard->link);
+	wlr_seat_set_keyboard(keyboard->seat, NULL);
```
emersion: We should only do this if `keyboard->seat->keyboard_state.keyboard == keyboard->input_device->keyboard` (ie. if the active keyboard on the seat is the virtual keyboard).

emersion: Oh, a better fix would be to listen to the keyboard `destroy` signal in the seat.

types/wlr_virtual_keyboard_v1.c
```diff
@@ -99,6 +99,8 @@ static void virtual_keyboard_destroy_resource(struct wl_resource *resource) {
 		virtual_keyboard_from_resource(resource);
 	wlr_signal_emit_safe(&keyboard->events.destroy, keyboard);
 	wl_list_remove(&keyboard->link);
+	wlr_signal_emit_safe(&keyboard->input_device.events.destroy,
```
emersion: We should replace this line (and the next line) with `wlr_input_device_destroy`

rootston/seat.c
```diff
@@ -613,6 +613,9 @@ static void handle_keyboard_destroy(struct wl_listener *listener, void *data) {
 	wl_list_remove(&keyboard->device_destroy.link);
 	wl_list_remove(&keyboard->keyboard_key.link);
 	wl_list_remove(&keyboard->keyboard_modifiers.link);
+	if (seat->seat->keyboard_state.keyboard == keyboard->device->keyboard) {
+		wlr_seat_set_keyboard(seat->seat, NULL);
```
emersion: >Because I haven't seen any code possibly sending NULL to seat_set_keyboard, it seems that an active keyboard being destroyed would have never cleared the problematic listeners, even with other keyboard types. That's why I added the fix to the general handler in seat.

Hmm, we already set it to NULL in `handle_keyboard_destroy` in `seat/wlr_seat_keyboard.c`. I don't understand why this is needed.

dcz-purism: That's because `wlr_sear_set_keyboard` also removes listeners pointing to signals on `wlr_input_device` in `seat/wlr_seat_keyboard.c:127`. Those signals become invalid after destruction, so the disconnection must happen beforehand.

If I'm reading this right, the listeners are attached to the seat, so they will keep references even after referenced keyboards are gone.

emersion: Maybe calling `wlr_seat_set_keyboard(seat, NULL)` in `handle_keyboard_destroy` is the right fix?

dcz-purism: That's what's happening here - note that I moved the call.

emersion: But this is in rootston - not in wlroots.

dcz-purism:
I guess this should become 2 separate commits.

PR #1196, opened 2018-08-22
rootston/surface_layers: fix clicking after surface moved

dcz-purism:
Surfaces are not notified of cursor position changes if the surface moves, only if the cursor moves. This workaround emits an extra cursor position event before every click to make sure the click lands in the right place.

A better solution would be to update the cursor position while the surface is moving, but it's difficult to provide a correct time value on input events, because the time values are taken from external sources.

Another solution is to set a flag when the surface is moved and clear it on click, after sending the motion event. This reduces the number of motion events: from 1 every click to 1 every click after movement.

### DIFF
--- a/rootston/layer_shell.c
+++ b/rootston/layer_shell.c
@@ -1,6 +1,12 @@
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200112L
+#endif
 #include <assert.h>
+#include <errno.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
+#include <time.h>
 #include <wayland-server.h>
 #include <wlr/types/wlr_box.h>
 #include <wlr/types/wlr_surface.h>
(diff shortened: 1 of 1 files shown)


SirCmpwn:
>A better solution would be to update the cursor position while the surface is moving, but it's difficult to provide a correct time value on input events, because the time values are taken from external sources.

The external sources always just use CLOCK_MONOTONIC and we can too.

dcz-purism:
I accidentally pushed to much... please ignore until I test this.

dcz-purism:
Handling of moving surfaces is now in 2 places: in `arrange_layers` and `handle_surface_commit`. A naive implementation would send 2 cursor move events on each resize: when the `layer_surface` is resized, and then when the surface inside is changed. I eliminated that by not sending any cursor move events when a `layer_surface` changes size, only when it moves.

agx:
@dcz-purism is this testable via the layer-shell example?

dcz-purism:
The layer-shell example doesn't move on its own; I've been testing it using virtboard: https://source.puri.sm/Librem5/virtboard/tree/input_method revision [81c74611](https://source.puri.sm/Librem5/virtboard/commit/81c74611fbd378721aea90a62f01f7833590c3d5)

To see the issue:

1. Run virtboard as the only application
2. Click the arrow button (nothing else should be visible before clicking)
3. Without moving the cursor, click again (the keyboard should be expanded before clicking)

Before the patch, the keyboard will register a key press close to its top left (`q`).

After the patch, the keyboard will register a key press in the bottom part (the arrow button), and hide back again.

RedSoxFan:
Thanks!

PR #1199, opened 2018-08-25
introduce wlr_output_layout_farthest_output

apreiml:
Similar to wlr_output_layout_adjacent_output but will return the one that is the farthest away from the reference in given direction.

I refactored the adjacent one to share some code.


RedSoxFan:
Thanks!

PR #1200, opened 2018-08-26
Add destroy signals to types that are destroyed by wl_display_destroy

alexbakker:
As discussed on IRC. This should make it a bit easier for compositors to keep track of which resources get destroyed.


SirCmpwn:
Thanks!

PR #1201, opened 2018-08-27
Implement wlr-foreign-toplevel-management-unstable-v1

ammen99:
This is a proof-of-concept implementation of https://github.com/swaywm/wlr-protocols/pull/26
Provides the basic wlroots implementation, rootston glue and a simple example client.


types/wlr_foreign_toplevel_v1.c
```diff
@@ -0,0 +1,477 @@
...
+}
+
+void wlr_foreign_toplevel_manager_v1_destroy(
+		struct wlr_foreign_toplevel_manager_v1 *manager) {
+	if (manager == NULL) {
+		return;
+	}
+
+	struct wlr_foreign_toplevel_v1 *toplevel, *tmp_toplevel;
+	wl_list_for_each_safe(toplevel, tmp_toplevel, &manager->toplevels, link) {
```
ammen99: I wasn't sure about that, we shouldn't auto-destroy things that the compositor created (the `wlr_foreign_toplevel_v1`)?

examples/foreign-toplevel.c
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	fprintf(stderr, "No toplevel with the given id: %d\n", id);
+	exit(EXIT_FAILURE);
+}
+
+int main(int argc, char **argv) {
+	int focus_id = -1, close_id = -1;
+	int maximize_id = -1, unmaximize_id = -1;
+	int	minimize_id = -1, restore_id = -1;
```
SirCmpwn: Weird tab here

examples/foreign-toplevel.c
```diff
@@ -0,0 +1,362 @@
...
+	if (print_endl) {
+		printf("\n");
+	}
+}
+
+static void	toplevel_handle_title(void *data,
+		struct zwlr_foreign_toplevel_handle_v1 *zwlr_toplevel,
+		const char *title) {
+	struct toplevel_v1 *toplevel = data;
+	if (toplevel->pending.title) {
```
emersion: No need to check for NULL, `free(NULL)` is a no-op

examples/foreign-toplevel.c
```diff
@@ -0,0 +1,362 @@
...
+	} else {
+		printf(" inactive");
+	}
+
+	if (print_endl) {
+		printf("\n");
+	}
+}
+
+static void	toplevel_handle_title(void *data,
```
emersion: Extra tab here

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+
+struct wlr_foreign_toplevel_manager_v1 {
+	struct wl_global *global;
+	struct wl_list resources;
+	struct wl_list toplevels; // wlr_foreign_toplevel_handle_v1::link
+
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal destroyed;
```
emersion: This should be named `destroy`

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+	struct wlr_output *output;
+};
+
+struct wlr_foreign_toplevel_handle_v1 {
+	struct wl_list resources;
+	struct wl_list link;
+
+	char *title;
+	char *app_id;
+	struct wl_list outputs; // wlr_foreign_toplevel_output
```
emersion: I'd like not to have a list of outputs here. We should add it to `wlr_surface` later instead.

(Additionally, this doesn't handle outputs being destroyed)

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+	uint32_t state; // wlr_foreign_toplevel_state
+
+	struct {
+		struct wl_signal request_maximize;
+		struct wl_signal request_minimize;
+		struct wl_signal request_activate;
+		struct wl_signal request_set_rectangle;
```
emersion: I think we can drop the `request_` prefix for this one, as it's just a hint.

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal destroyed;
+	} events;
+
+	void *data;
+};
+
+enum wlr_foreign_toplevel_handle_state {
```
emersion: We decided it's better to use directly enums generated by wayland-scanner.

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+		struct wl_signal request_minimize;
+		struct wl_signal request_activate;
+		struct wl_signal request_set_rectangle;
+		struct wl_signal request_close;
+	} events;
+
+	void *data;
+};
+
+struct wlr_foreign_toplevel_handle_maximized_event {
```
emersion: Where are these used?

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+	struct wlr_foreign_toplevel_handle_v1 *toplevel, struct wlr_output *output);
+
+void wlr_foreign_toplevel_handle_v1_set_maximized(
+	struct wlr_foreign_toplevel_handle_v1 *toplevel, bool maximized);
+void wlr_foreign_toplevel_handle_v1_set_minimized(
+	struct wlr_foreign_toplevel_handle_v1 *toplevel, bool minimized);
+void wlr_foreign_toplevel_handle_v1_set_activated(
+	struct wlr_foreign_toplevel_handle_v1 *toplevel, bool activated);
+
+void wlr_foreign_toplevel_handle_v1_send_done(
```
emersion: This should not be exposed. Instead, add a idle event source when changing a value and send `done` when it's fired.

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,112 @@
...
+	int32_t x, y, width, height;
+};
+
+struct wlr_foreign_toplevel_manager_v1* wlr_foreign_toplevel_manager_v1_create(
+	struct wl_display *display);
+void wlr_foreign_toplevel_manager_v1_destroy(
+	struct wlr_foreign_toplevel_manager_v1 *manager);
+
+struct wlr_foreign_toplevel_handle_v1*
+wlr_foreign_toplevel_manager_v1_create_toplevel_handle(
```
emersion: This could be named `wlr_foreign_toplevel_handle_v1_create`

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+}
+
+static void toplevel_handle_send_maximized_event(struct wl_resource *resource,
+		bool state) {
+	struct wlr_foreign_toplevel_handle_v1 *toplevel =
+		toplevel_handle_from_resource(resource);
+	if (!toplevel) {
+		return;
+	}
+	struct wlr_foreign_toplevel_handle_maximized_event event;
```
emersion: Use initializers instead:

```
struct thing = {
  .field = …
};
```

This ensures no field is undefined.

include/wlr/types/wlr_foreign_toplevel_management_v1.h
emersion: Ah, I see. Events should be annotated with the value they're fired with.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	if (manager == NULL) {
+		return;
+	}
+
+	struct wlr_foreign_toplevel_handle_v1 *toplevel, *tmp_toplevel;
+	wl_list_for_each_safe(toplevel, tmp_toplevel, &manager->toplevels, link) {
+		wl_list_remove(&toplevel->link);
+		/* we must make sure the toplevel link is valid when the compositor
+		 * destroys the wlr_foreign_toplevel_handle_v1 */
+		wl_list_init(&toplevel->link);
```
emersion: Just destroy them

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	}
+
+	struct wl_resource *resource, *tmp_resource;
+	wl_resource_for_each_safe(resource, tmp_resource, &manager->resources) {
+		foreign_toplevel_manager_resource_destroy(resource);
```
emersion: This should just be `wl_resource_destroy`

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	zwlr_foreign_toplevel_handle_v1_send_state(resource, &states);
+	wl_array_release(&states);
+
+	zwlr_foreign_toplevel_handle_v1_send_done(resource);
+}
+
+static void foreign_toplevel_manager_bind(struct wl_client *client, void *data,
+		uint32_t version, uint32_t id) {
+	struct wlr_foreign_toplevel_manager_v1 *manager = data;
+	assert(client && manager);
```
emersion: No need for this

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+}
+
+static struct wl_resource *create_toplevel_resource_for_resource(
+		struct wlr_foreign_toplevel_handle_v1 *toplevel,
+		struct wl_resource *manager_resource)
+{
+	struct wl_client *client = wl_resource_get_client(manager_resource);
+	struct wl_resource *resource = wl_resource_create(client,
+			&zwlr_foreign_toplevel_handle_v1_interface,
+			FOREIGN_TOPLEVEL_MANAGEMENT_V1_VERSION, 0);
```
emersion: Use the version from the `manager_resource` instead

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	}
+	if (toplevel->app_id) {
+		free(toplevel->app_id);
+	}
+
+	free(toplevel);
+}
+
+static void foreign_toplevel_resource_destroy(struct wl_resource *resource) {
+	if (wl_resource_get_user_data(resource)) {
```
emersion: No need to check for this. Just ensure to call `wl_list_init` after removing from the list.

include/wlr/types/wlr_foreign_toplevel_management_v1.h
ammen99: but here the idea is different, the compositor can set a bit mask of states (as opposed to the protocol, where we use an array), then it gets stored into state. If you're against this, I could change it to getting the ready `wl_array` from the compositor.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	wl_resource_for_each(resource, &toplevel->resources) {
+		zwlr_foreign_toplevel_handle_v1_send_done(resource);
+	}
+}
+
+void wlr_foreign_toplevel_handle_v1_destroy(
+		struct wlr_foreign_toplevel_handle_v1 *toplevel) {
+	struct wl_resource *resource, *tmp;
+	wl_resource_for_each_safe(resource, tmp, &toplevel->resources) {
+		zwlr_foreign_toplevel_handle_v1_send_closed(resource);
```
emersion: This desn't remove resources from the list

include/wlr/types/wlr_foreign_toplevel_management_v1.h
ammen99: oh, that's even better. Hadn't noticed we do things like that.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	} else {
+		// XXX: log an error? crash?
+	}
+}
+
+static void fill_array_from_toplevel_state(struct wl_array *array,
+		uint32_t state) {
+	if (state & WLR_FOREIGN_TOPLEVEL_STATE_MAXIMIZED) {
+		uint32_t *index = wl_array_add(array, sizeof(uint32_t));
+		*index = ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_MAXIMIZED;
```
emersion: This is missing NULL checks

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,503 @@
...
+	if (!toplevel) {
+		return;
+	}
+	wlr_signal_emit_safe(&toplevel->events.request_close, toplevel);
+}
+
+static void foreign_toplevel_handle_set_rectangle(struct wl_client *client,
+		struct wl_resource *resource, struct wl_resource *surface,
+		int32_t x, int32_t y, int32_t width, int32_t height)
+{
```
emersion: Style: `{` on the same line as `)`

include/wlr/types/wlr_foreign_toplevel_management_v1.h
emersion: I see. It's fine

types/wlr_foreign_toplevel_management_v1.c
ammen99: what should happen if it returns NULL?

include/wlr/types/wlr_foreign_toplevel_management_v1.h
ammen99: Yep, we need some way to track this, and except if we have some list of outputs in wlr_surface, I see no way around it. But even then we will need to associate a toplevel with a surface, which the current implementation avoids (not sure if that brings any value though). It could be a follow-up issue, but I'll surely update it to handle output destruction.

emersion: Okay, that seems reasonable.

types/wlr_foreign_toplevel_management_v1.c
ammen99: should I add a destroy event to the `wlr_toplevel_handle_v1`? Otherwise it seems we're just auto-destroying things the compositor explicitly created.

ammen99:
@emersion Thanks for the review.

I updated the code to address (most) of the issues. I'll squash commits again when this is deemed ready for merging.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+
+struct wlr_foreign_toplevel_manager_v1* wlr_foreign_toplevel_manager_v1_create(
+		struct wl_display *display) {
+	struct wlr_foreign_toplevel_manager_v1 *manager = calloc(1,
+			sizeof(struct wlr_foreign_toplevel_manager_v1));
+	if (!manager) {
+		return NULL;
+	}
+
+	manager->event_loop = wl_display_get_event_loop(display);
```
emersion: No need to store this in a field, we can call it every time we need it

emersion: (It won't return NULL)

```c
WL_EXPORT struct wl_event_loop *
wl_display_get_event_loop(struct wl_display *display)
{
	return display->loop;
}
```

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+}
+
+static void foreign_toplevel_resource_destroy(struct wl_resource *resource) {
+	wl_list_remove(wl_resource_get_link(resource));
+}
+
+static struct wl_resource *create_toplevel_resource_for_resource(
+		struct wlr_foreign_toplevel_handle_v1 *toplevel,
+		struct wl_resource *manager_resource)
+{
```
emersion: Style: `{` on the same line as `)`

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+		struct wlr_output *output) {
+	struct wlr_foreign_toplevel_handle_output *toplevel_output;
+	wl_list_for_each(toplevel_output, &toplevel->outputs, link) {
+		if (toplevel_output->output == output) {
+			return; // we have already sent output_enter event
+		}
+	}
+
+	toplevel_output =
+	    calloc(1, sizeof(struct wlr_foreign_toplevel_handle_output));
```
emersion: This is missing a NULL check

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+	if (toplevel->idle_source)
+		return;
+
+	toplevel->idle_source = wl_event_loop_add_idle(toplevel->manager->event_loop,
+		toplevel_idle_send_done, toplevel);
+}
+
+void wlr_foreign_toplevel_handle_v1_set_title(
+		struct wlr_foreign_toplevel_handle_v1 *toplevel, const char *title) {
+	if (toplevel->title) {
```
emersion: No need for these NULL checks

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+	wl_resource_for_each(resource, &toplevel->resources) {
+		zwlr_foreign_toplevel_handle_v1_send_done(resource);
+	}
+
+	toplevel->idle_source = NULL;
+}
+
+static void toplevel_update_idle_source(
+	struct wlr_foreign_toplevel_handle_v1 *toplevel) {
+	if (toplevel->idle_source)
```
emersion: Style: missing braces

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+	if (!r) {
+		wl_resource_post_no_memory(resource);
+		wl_array_release(&states);
+		return;
+	}
+
+	zwlr_foreign_toplevel_handle_v1_send_state(resource, &states);
+	wl_array_release(&states);
+
+	toplevel_update_idle_source(toplevel);
```
emersion: This will send "done" to all resources -- we only want to send "done" to this one

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+	if (width < 0 || height < 0) {
+		wl_resource_post_error(resource,
+			ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_ERROR_INVALID_RECTANGLE,
+			"invalid rectangle passed to set_rectangle: width/height < 0");
+		return;
+	}
+
+	struct wlr_foreign_toplevel_handle_set_rectangle_event event = {
+		.toplevel = toplevel,
+		.surface = wl_resource_get_user_data(surface),
```
emersion: It's better to never use plain `wl_resource_get_user_data` like this. There are type-safe helper functions available, here `wlr_surface_from_resource`.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,583 @@
...
+		.x = x,
+		.y = y,
+		.width = 0,
```
emersion: This probably shouldn't always be zero.

rootston/xwayland.c
```diff
@@ -262,6 +286,11 @@ static void handle_unmap(struct wl_listener *listener, void *data) {
 
 	wl_list_remove(&roots_surface->surface_commit.link);
 
+	if (view->toplevel_handle) {
```
emersion: Can't this be in `view_unmap`?

rootston/xwayland.c
```diff
@@ -249,6 +267,12 @@ static void handle_map(struct wl_listener *listener, void *data) {
 			view->titlebar_height = 12;
 		}
 
+		view_create_foreign_toplevel_handle(view);
```
emersion: Can't this be in `view_setup`?

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,120 @@
...
+
+		//wlr_foreign_toplevel_handle_set_rectangle_event
+		struct wl_signal set_rectangle;
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_foreign_toplevel_handle_maximized_event {
```
emersion: This is missing the `_v1` suffix (along with other structs in this file)

types/wlr_foreign_toplevel_management_v1.c
ammen99: i think I saw this check somewhere else too, that's why i did it. will remove.

ammen99: yeah, found it, https://github.com/swaywm/wlroots/blob/master/types/wlr_idle.c#L227 :)

types/wlr_foreign_toplevel_management_v1.c
ammen99: ok, but it would be ok if I directly send the done event to the resource without going through the idle loop in this case (because we already sent all we know about the toplevel)?

types/wlr_foreign_toplevel_management_v1.c
emersion: Hmm, you still need to access the `display`, so having a field is probably fine

types/wlr_foreign_toplevel_management_v1.c
emersion: Yeah

types/wlr_foreign_toplevel_management_v1.c
ammen99: lol, no idea how this came to be.

types/wlr_foreign_toplevel_management_v1.c
ammen99: do we again `post_no_memory` to the client or just silently fail?

types/wlr_foreign_toplevel_management_v1.c
ammen99: Yes, thanks for the suggestion. I had even used `wl_surface` instead of `wlr_surface` ...

types/wlr_foreign_toplevel_management_v1.c
ammen99: didn't quite understand, is the code fine if I remove the check or not (plus if display is NULL this is crash)?

types/wlr_foreign_toplevel_management_v1.c
emersion: For allocations not caused by client requests, we can just log an error.

types/wlr_foreign_toplevel_management_v1.c
emersion: `display` won't be NULL. Yes, please remove the check.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -0,0 +1,584 @@
...
+}
+
+static void foreign_toplevel_handle_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	struct wlr_foreign_toplevel_handle_v1 *toplevel =
+		toplevel_handle_from_resource(resource);
+	if (!toplevel) {
+		return;
+	}
+	toplevel_handle_from_resource(resource);
```
emersion: I think this can be removed.

emersion: Actually, "destroy" request handlers should always just call `wl_resource_destroy`.

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -0,0 +1,120 @@
...
+#endif
+
+#ifndef WLR_TYPES_WLR_FOREIGN_TOPLEVEL_MANAGEMENT_V1_H
+#define WLR_TYPES_WLR_FOREIGN_TOPLEVEL_MANAGEMENT_V1_H
+
+#include <wayland-server.h>
+#include <wlr/types/wlr_output.h>
+
+struct wlr_foreign_toplevel_manager_v1 {
+	struct wl_event_loop *event_loop;
```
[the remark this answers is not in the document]
ammen99: Should I keep the pointer to the toplevel manager or drop it(it is used only for the event loop, which as you say should  come from the resource/client)?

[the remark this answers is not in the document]
emersion: Dropping it is the reason why I'm suggesting to use `wl_client_get_display` :)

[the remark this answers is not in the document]
ammen99: okay, will do it sometime later today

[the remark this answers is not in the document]
ammen99: This won't work very well. Where do I get the client for `wl_client_get_display()`? The resources list might be empty.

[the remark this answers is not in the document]
ammen99: Hmm, actually we can do this: Use the client from the first resource (though I'm not sure how to access the first element in a `wl_list`, I'd be helpful if somebody hints me). If no clients, don't send done, because no client will receive it anyway. Is this fine?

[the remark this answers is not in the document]
emersion: Okay, my bad, let's just keep it that way. I thought it was in a request handler, but it's not. So it's better to keep things simple instead of having this minor improvement.

For the record, here's how you access the first element of a list:

```c
struct thing *thing = wl_container_of(parent->children.next, thing, link);
```

ammen99:
Once again thanks for the reviews @emersion 
I squashed things together and fixed the last issue we know of.

emersion:
Thanks!

PR #1202, opened 2018-08-27
Init the new destroy signals added by #1200

alexbakker:
I forgot to init the destroy signals that were added by #1200. Sorry about that, I should have tested it.


SirCmpwn:
Thanks!

PR #1208, opened 2018-08-30
Fixes examples/dmabuf-capture being built with unmet dependencies

arandomhuman:
Even if the libav* variables were disabler objects, the build targer
dmabuf-capture was being built. Modified the script to support a generic
solution.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -20,31 +20,31 @@ endif
 examples = {
 	'simple': {
 		'src': 'simple.c',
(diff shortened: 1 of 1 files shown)


SirCmpwn:
Thanks!

ascent12:
>Even if the libav* variables were disabler objects, the build targer
dmabuf-capture was being built

Really? If that's true, this should be raised as an upstream Meson bug instead of doing this.

PR #1210, opened 2018-08-31
add missing wlr_box.h include in wlr_screencopy_v1.h

ammen99:
`wlr_screencopy_v1.h` depends on `wlr_box`, but doesn't have the include.


SirCmpwn:
Thanks!@

PR #1212, opened 2018-09-01
xdg-shells: make xdg_surface implementation handle inert objects

ammen99:
Original issue: https://github.com/WayfireWM/wayfire/issues/73

### DIFF
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -99,6 +99,9 @@ void destroy_xdg_toplevel(struct wlr_xdg_surface *surface) {
 static void xdg_surface_handle_ack_configure(struct wl_client *client,
 		struct wl_resource *resource, uint32_t serial) {
 	struct wlr_xdg_surface *surface = wlr_xdg_surface_from_resource(resource);
+	if (!surface) {
+		return;
+	}
 
 	if (surface->role == WLR_XDG_SURFACE_ROLE_NONE) {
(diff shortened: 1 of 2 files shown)


emersion:
Not sure this is the right thing to do.

>An xdg_surface must only be destroyed after its role object has been destroyed.

But it's not specified what happens if the wl-surface is destroyed before the xdg-surface.

ammen99:
Well, I'm open to other ideas to solve the problem. But the main reason for this crash was that `ack_configure` came after the `destroy` request, and I thought these kind of races are solved by inertness.

emersion:
Hmm, how is it possible? If the client sends a destroy request, it can't send an `ack_configure` after without a use-after-free.

SirCmpwn:
Maybe we just protocol error the client in this case?

SirCmpwn:
bump

ammen99:
Hi, I've forgotten about this one. So, here's a backtrace: https://pastebin.com/6yhNkqca and here's a `WAYLAND_DEBUG` of the tool: https://pastebin.com/2Z6NUSz2

Looking at them again, I'm not so sure anymore what the correct way to fix this is.

panaman67:
@ammen99 Whats the status on this PR?

emersion:
@panaman67 Please refrain from randomly pinging people. I appreciate the enthusiasm, but if you want to help adopt these PRs instead.

ammen99:
@panaman67 The status is "I don't know what the correct fix is".

PR #1213, opened 2018-09-01
Add wlr_log_get_verbosity method

arandomhuman:
Returns the verbosity passed to wlr_log_init().

Used to close stdout/stderr for Xwayland.

### DIFF
--- a/rootston/config.c
+++ b/rootston/config.c
@@ -26,7 +26,10 @@ static void usage(const char *name, int ret) {
 		"                See `rootston.ini.example` for config\n"
 		"                file documentation.\n"
 		" -E <COMMAND>   Command that will be ran at startup.\n"
-		" -D             Enable damage tracking debugging.\n",
+		" -D             Enable damage tracking debugging.\n"
+		" -l <LEVEL>     Set log verbosity, where,\n"
+		"                0:SILENT, 1:ERROR, 2:INFO, 3+:DEBUG\n"
+		"                (default: DEBUG)\n",
 		name);
 
 	exit(ret);
@@ -455,7 +458,8 @@ struct roots_config *roots_config_create_from_args(int argc, char *argv[]) {
 	wl_list_init(&config->bindings);
 
 	int c;
-	while ((c = getopt(argc, argv, "C:E:hD")) != -1) {
+	unsigned int log_verbosity = WLR_DEBUG;
+	while ((c = getopt(argc, argv, "C:E:hDl:")) != -1) {
 		switch (c) {
 		case 'C':
 			config->config_path = strdup(optarg);
@@ -466,11 +470,18 @@ struct roots_config *roots_config_create_from_args(int argc, char *argv[]) {
 		case 'D':
 			config->debug_damage_tracking = true;
 			break;
+		case 'l':
+			log_verbosity = strtoul(optarg, NULL, 10);
+			if (log_verbosity >= WLR_LOG_IMPORTANCE_LAST) {
+				log_verbosity = WLR_LOG_IMPORTANCE_LAST - 1;
+			}
+			break;
 		case 'h':
 		case '?':
 			usage(argv[0], c != 'h');
 		}
 	}
+	wlr_log_init(log_verbosity, NULL);
 
 	if (!config->config_path) {
 		// get the config path from the current directory
--- a/util/log.c
+++ b/util/log.c
@@ -85,3 +85,7 @@ const char *_wlr_strip_path(const char *filepath) {
 	}
 	return filepath;
 }
+
+enum wlr_log_importance wlr_log_get_verbosity(void) {
+	return log_importance;
+}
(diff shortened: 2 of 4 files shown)


util/log.c
```diff
@@ -85,3 +85,7 @@ const char *_wlr_strip_path(const char *filepath) {
 	}
 	return filepath;
 }
+
+enum wlr_log_importance wlr_log_get_verbosity() {
```
SirCmpwn: `wlr_log_get_verbosity(void)`

arandomhuman: Done.

rootston/config.c
```diff
@@ -466,11 +470,18 @@ struct roots_config *roots_config_create_from_args(int argc, char *argv[]) {
...
+			if (log_verbosity >= WLR_LOG_IMPORTANCE_LAST) {
+				log_verbosity = WLR_LOG_IMPORTANCE_LAST - 1;
+			}
+			break;
 		case 'h':
 		case '?':
 			usage(argv[0], c != 'h');
 		}
 	}
+	wlr_log_init(log_verbosity, NULL);
```
emersion: Hmm. Now `wlr_log_init` is called twice. Do we care? We should probably update the wlr-log docs to allow this.

xwayland/xwayland.c
```diff
@@ -123,9 +123,19 @@ static void exec_xwayland(struct wlr_xwayland *wlr_xwayland) {
 		wlr_xwayland->wl_fd[1], wlr_xwayland->display, wlr_xwayland->x_fd[0],
 		wlr_xwayland->x_fd[1], wlr_xwayland->wm_fd[1]);
 
-	// TODO: close stdout/err depending on log level
+	// Closes stdout/stderr depending on log verbosity
+	enum wlr_log_importance verbosity = wlr_log_get_verbosity();
+	int devnull = open("/dev/null", O_WRONLY | O_CREAT, 0666);
```
emersion: Check for -1

rootston/config.c
arandomhuman: The function doesn't really care how many times it's called, so I let it be. It acts kind of like an "upgrade" of the log level (from DEBUG to whatever is given).

SirCmpwn: I don't care if we call it several times

xwayland/xwayland.c
```diff
@@ -123,9 +123,23 @@ static void exec_xwayland(struct wlr_xwayland *wlr_xwayland) {
...
+	if (devnull < 0) {
+		wlr_log_errno(WLR_ERROR, "XWayland: failed to open /dev/null");
+		return -1;
```
arandomhuman: I'm not sure how to properly exit with an error here, since the function type is void. Simply a `return` or `_exit(EXIT_FAILURE)` (which seemed too harsh for something this nonessential)?

emersion: `_exit(EXIT_FAILURE)` is fine. If we can't open `/dev/null` something really bad is going on anyway.

rootston/config.c
emersion: Can you add this to the docs?

xwayland/xwayland.c
```diff
@@ -123,9 +123,23 @@ static void exec_xwayland(struct wlr_xwayland *wlr_xwayland) {
...
+	if (verbosity < WLR_INFO) {
+		dup2(devnull, STDOUT_FILENO);
+	}
+	if (verbosity < WLR_ERROR) {
+		dup2(devnull, STDERR_FILENO);
+	}
 
 	execvp("Xwayland", argv);
+
+	close(devnull);
```
emersion: This is never reached. We should probably close before exec-ing?

ascent12: It will be reached on failure, but there deserves to be some proper error handling here.

arandomhuman: You mean checking the result of execvp, right?

ascent12: Actually, looking at how exec_wayland is used, there is already error handling outside of the function.
Closing devnull here will mess with errno, though.

arandomhuman: The error checking code in `xwayland_start_server` (line 349) could probably be shifted to after the `execvp` call (since execvp returns iff it fails)?

This in particular:
```
wlr_log_errno(WLR_ERROR, "failed to exec Xwayland");
_exit(EXIT_FAILURE);
```

Should I add it to the PR?

ascent12: Yeah, you may as well. Also mark `exec_wayland` as [noreturn](https://en.cppreference.com/w/c/language/_Noreturn).

arandomhuman: Added to commit 8184ccc.

rootston/config.c
arandomhuman: Sorry, I missed this earlier. Added a short blurb in the header.

SirCmpwn:
Thanks!

Issue #774, opened 2018-03-29
Renderer redesign v5


ascent12:
Boy, I really love breaking this, don't I?

This is all pretty breaking stuff, so I'm not suggesting we do this immediately, but I'll put this idea out there to get feedback on it. This may be a wlroots 2.0 kind of change, unless it's wanted sooner.

Obviously, changes are needed to facilitate a Vulkan renderer, and possibly other renderer types that we may think up in the future.
@SirCmpwn suggested merging `wlr_egl` and `wlr_renderer` together (before retracting it), and while I originally didn't like the idea, I've come around to thinking that it's the way we should probably go.

## Current issues

The main problem is that EGL types are deeply embedded into our code. A `wlr_backend` has an `EGLDisplay`, a `wlr_output` has an `EGLSurface`, and our renderer expects nothing else. This is fine if we can create those from various 'native' types using `EGL_KHR_platform_*`, but it doesn't work in cases where no such extension exists. We've only managed to scrape by in the other cases because `EGL_MESA_platform_surfaceless` exists.

Vulkan is where this really becomes a problem. There are similar types, but they're not available in all of the same places. `EGLDisplay` == `VkDevice`and `EGLSurface` == `VkSurfaceKHR` + `VkSwapchainKHR`.
However, we have no way to create the `VkSurfaceKHR` from GBM or in the headless backend. Vulkan can definitely do surfaceless rendering, just not with a `VkSurfaceKHR`.

Basically, this model can only work with the Wayland and X11 backends, unless more Vulkan extensions are added, but that's not something I would bank on. This abstraction is just too limiting.

## Renderer-specific backends to backend-specific renderers

So instead of having the backends need to add support for every renderer and cram it into some `EGLSurface`-like, we just have the renderer take the 'native' backend resources, and create rendering resources from them however it sees fit.

So we tear all of the EGL code out of the backends and put them into `wlr_renderer`. With this, `wlr_egl` doesn't really have a purpose any more, and would just be merged into `wlr_renderer` too.

This will allow the renderer to work differently based on what backend its using. Backend-specific code isn't great, and will make things more complicated, but I think it's going to be a necessary evil.

It does make adding new backends a bit harder, but I hope the headless rendering case would be able to work with most other backends that people would want to add.

## The Design

I even bothered to draw a fancy little diagram.

![wlroots_renderer](https://user-images.githubusercontent.com/5235114/38083927-7f9fb520-33a7-11e8-99a7-d3e475609a3a.png)

[...]

Issue #642, opened 2018-02-18
Vulkan Support


ascent12:
This would be nice for even better performance, and actually allowing real multi-gpu to happen.

Currently, there are several things missing from Vulkan to truly make this a reality and be a first-class renderer for wlroots. It would probably be best for this to remain out-of-tree until they've been resolved. 
 
There are also things in wlroots that needs to change for a Vulkan (or any non-OpenGL) renderer to even be viable. These things would all be incredibly breaking changes, and may be more appropriate for a 2.0 release.

## Removal of hard-dependency on EGL
Basically, Vulkan doesn't need EGL and manages its own **window system integration**.
I think this would involve wrapping all of that functionality up in a *wlr_wsi* struct, and have loadable backends, just like everything else we do.

This might also open up the possibility for other types of renderers to happen (e.g. software) and backends which don't have an EGL surface type (e.g. fbdev). I think this would be a very difficult interface to design, though.

## Removal of all other GL calls in wlroots
It would really be nice to not need OpenGL at all, but there are a couple of places in wlroots where we need to ability to draw stuff.

For things like software cursors, it might be better to do it with callbacks, and get the library user to do it for us.

For the multi-gpu stuff in DRM, I think it can mostly be removed by the inclusion of Vulkan. If we're going to get true multi-gpu rendering, I'm going to remove the half-assed fake multi-gpu support.

# Current Issues

## Binding wl_display

Right now, there is no equivalent to [WL_bind_wayland_display](https://cgit.freedesktop.org/mesa/mesa/tree/docs/specs/WL_bind_wayland_display.spec). I guess an EGLDisplay would need to be kept around just for this, until it gets implemented.
There is a github issue about it here: https://github.com/KhronosGroup/Vulkan-Docs/issues/294

## No support for GBM surfaces

Vulkan currently doesn't have a GBM surface type or any other surface type we can use for the DRM backend, unlike EGL.

Oddly enough, there is [VK_KHR_display](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_KHR_display), which does a whole bunch of the stuff DRM is doing, but it's not suitable for us. I think it's mainly for controlling VR headsets, not implementing full display servers.

I think this *may* be addressed whenever this GBM replacement that Nvidia was supposedly developing is finished. We'll have to wait and see what happens.

SirCmpwn:
[...]

Issue #1174, opened 2018-08-03
Set eglSwapInterval to 0 in wayland backend


emersion:
Since we track frame events manually, no need to require EGL to do it for us. This message from Daniel Stone on the ML reminded me of this:

>When you use EGL in a client, is it important to never let eglSwapBuffers() block for you. You need to track frame callbacks manually or set eglSwapInterval to zero, preferably both, so your application code does not get blocked by EGL API that was designed for applications far less smart than yours.

emersion:
Fixed in https://github.com/swaywm/wlroots/pull/1338

PR #1214, opened 2018-09-01
Implement basic vulkan renderer

nyorain:
Finally had some time to put this together.

# Current state

- working and tested in basic scenarios (displaying terminal, cursor etc) on all backends,
  all examples (and rootston) work, screenshotting as well
- buffers from hardware accelerated clients (dmabuf/wl_drm) are not (yet) implemented for
  the vulkan renderer. Don't know how to test dmabuf and to get mesa egl/vulkan
  clients to work the vulkan renderer would have to implement
  wl_drm at the moment; i'm not sure how to do that correctly.
  Since it was was mentioned that this might not be needed in the future
  at all, leaving such buffers for another pr.
  The dmabuf implementation would also profit from the mentioned (and
  hopefully soon releasing) vulkan dmabuf modifiers extension.
  See @fooishbar's comments ([[1]](https://github.com/swaywm/wlroots/issues/642#issuecomment-371633233), [[2]](https://github.com/KhronosGroup/Vulkan-Docs/issues/294#issuecomment-371611750))
- As the pr says this renderer is fairly basic and some things can be optimized. The pr is rather large like this already, so i'd rather postpone that (see bottom of the pr)
- We also make some assumptions in the renderer not guaranteed by the spec. Some can be fixed with some work on the renderer, others would probably require large workarounds
or changes to wlr interfaces. Those are usually explicitly marked in the code.

You can try it out by starting programs with `WLR_RENDERER=vulkan ...`.
Note that besides the vulkan dependency it currently requires glslangValidator (arch package glslang) to compile the shaders from glsl into spirv c headers. This dependency is only required at build time and only to recompile the shaders; we could also include the
compiled shader headers in the repo, removing this dependency. 

# Interface changes

As expected, this has to touch wlr interfaces (only unstable ones, thanks for waiting with stabilizing them). The main changes are in renderer, output and backend interfaces.
This also introduces a new abstraction *wlr_render_surface*, generally is the design inspired and close to the ideas from [renderer redesign v5](https://github.com/swaywm/wlroots/issues/774). This allowed me to remove all direct egl and gl usages from backends. egl initialization is still done by the backend and the render_surface creation interface is explicit (using wl/xcb/gbm primitives) so that we don't have any `wlr_backend_is_x` chains in the renderers. But otherwise renderers must be aware of all backends they support and do so explicitly. Therefore, this moves some code
from the backends (mainly drm, `drm/renderer.c`) to the vulkan/gles2 renderers.

In the renderer redesign issue, @SirCmpwn mentioned that this approach makes it hard to write new backends or renderers since for a new backend the render interface must be
extended and support implemented for existent renderers (if possible) and for a new renderer one has to explicitly implement support for the supported backends.

The thing with this is: there is no way without that difficulty.
When creating a new renderer, it has to know how to render on the backends
its supports. This cannot be done in a general way since e.g. vulkan has explicit methods for creating surfaces and egl on wayland requires creation of a wayland-egl-window and rendering on the drm/gbm backend requires a lot of completely different backend AND renderer dependent primitives.
And when creating a new backend, it is either possible to use the primitives
the renderer interface already uses (e.g. gbm) or the render interface
just has to be extended to be able to pass the new primitives from backend
to renderer.
We could (what i did first, see some older commits) create
a more general `create_render_surface` interface, where a
`void*` handle is passed to the renderer. But this requires that renderer
and backend both know what the handle should be (and casting
it into structures containing all information from the backend, see
e.g. the gbm case where a single handle isn't enough information) and to `if (backend_is_x(backend)) ...` chains in the renderer which seemed more limited and generally not like a good idea.

This digs somewhat into the current drm backend implementation to make this work, so it should be tested on as many setups as possible and definitely receive the mandatory @ascent12 review. I realize that this pr is rather large and may require some changes before everyone is happy merging it so please give me the chance to get it into that state with as much reviews and comments as possible. Most of the later commit messages are also rather detailed and outline design decisions as problems came up. 
I'm not sure how many of you are experienced with vulkan, but some reviews of the vulkan usage would be appreciated as well (although i heavily relied on the debug layers while developing). Otherwise, see [the vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html) for information.

If you (categorically/at the moment) reject such a pr or a vulkan renderer altogether, please also let me know up front. Although i spent quite some time on this i did so without communicating/asking and mainly because it was a really interesting task. So i'd understand if you chose not to merge this. If you do, i will gladly maintain this renderer in the future. Will also be as often as possible on the irc the next time, just ping me for questions and discussion.

This also fixes some non-vulkan issues though i found on the way so you at least probably
want the wlr_output_damage_begin fix and the improved wayland output cursor setting.

# Notes

A list of notes/decisions i wrote down while writing the renderer as well as ideas for future improvements and optimizations.

- with vulkan we can efficiently read from textures but wlr has no
  interface for this. It is only needed in one place at the moment iirc:
  in the drm backend when setting the cursor (there it is currently
  first rendered and then read from the dummy render surface). Since it's called every time 
  the cursor is changed when using the drm 
  backend, this should be a rather important optimization (and
  i guess we can implement it more efficiently in gles2 as well using a dummy framebuffer 
  we read the texture from instead of rendering the texture first?)
- vulkan and gles2 gbm render surfaces currently always use 3 buffers statically.
  that can be improved with explicit fencing (at least for atomic modesetting)
  and releasing on pageflip again (see drm backend comment)
  but since this requires more changes to the drm backend/wlr interface and is mainly a
  optimization thing, it is left for later
- ~~vk_khr_incremental_present is currently not used (optimization)~~ implemented, buffer age for swapchain surfaces fixed by pre-acquiring an image
- ~~all textures use currently linear tiling and are created on host
  visible memory. This has a rather large performance impact and
  can be improved with some uploading logic~~ staging logic for optimal shm textures now implemented
- ~~can textures ever be resized? if so this is currently not implemented
  on the vulkan renderer~~ nope, not allowed
- we currently just choose the first physical device.
  A better approach would be to choose them by what they support but
  also make this configurable. Multi gpu support is currently not
  implemented in any way and due to the way we interface with the
  drm backend, the vulkan renderer might not work at all when using
  the drm backend with multiple gpus.
- ~~in vulkan gbm render surfaces, we currently create gbm_bo's and import them into vulkan.
  we could also create vulkan images, export them as dmabufs and import them   as gbm_bo's. Might have an impact on performance/portability (?)~~ no, wouldn't be possible to specify gbm_use_scanout this way; can be improved with modifiers extension
- ~~since the currently packaged mesa version does not implement vk_khr_display
  this does not implement such a backend. That is a somewhat independent
  topic anyways. Would like to implement it though in the future (the next
  mesa version implements it iirc).~~ Started a VkDisplayKHR backend implementation [here](https://github.com/nyorain/wlroots/commit/5a6595202a41892d9fc105a6193c548df8822b1c) on top of this pr, roughly working with mesa 18.2. To not further blow this pr up, will submit it as a new one afterwards.
- ~~the vulkan renderer currently implements a custom swapchain for gbm render surfaces.
  It might be possible to use gbm_surface there as well but since the 
  documentation (in the gbm implementation, not the header) explicitly mentions 
  eglSwapBuffers and when it must be called, i'm not sure and haven't tried it.~~ nope, gbm_surface does not work with vulkan
- the current implementation requires vulkan 1.1. If this could be problem on any setup we
  want to support, let me know. The same can be achieved (with some work and additional 
  extension loading) with vulkan 1.0.


emersion:
Nice PR!

>screenshotting as well

I assume screencopy works but dmabuf-export doesn't.

>Don't know how to test dmabuf and to get mesa egl/vulkan
clients to work the vulkan renderer would have to implement
wl_drm at the moment; i'm not sure how to do that correctly.

Can you try `weston-simple-dmabuf`?

I don't think we want to implement wl_drm -- but we definitely want dmabuf.

>As the pr says this renderer is fairly basic and some things can be optimized. The pr is rather large like this already, so i'd rather postpone that (see bottom of the pr)

+1

>This allowed me to remove all direct egl and gl usages from backends

Have you designed this with a maybe future software renderer in mind?

>The thing with this is: there is no way without that difficulty.
When creating a new renderer, it has to know how to render on the backends
its supports. This cannot be done in a general way since e.g. vulkan has explicit methods for creating surfaces and egl on wayland requires creation of a wayland-egl-window and rendering on the drm/gbm backend requires a lot of completely different backend AND renderer dependent primitives.

I think we can switch to a per-renderer code in the backends, instead of a per-backend code in the renderer. What do you think?

>can textures ever be resized?

No, you need to destroy the previous one and create a new one.

>Multi gpu support is currently not
implemented in any way and due to the way we interface with the
drm backend, the vulkan renderer might not work at all when using
the drm backend with multiple gpus.

We should probably use the same device as the DRM backend?

include/wlr/types/wlr_output_damage.h
```diff
@@ -59,8 +60,8 @@ void wlr_output_damage_destroy(struct wlr_output_damage *output_damage);
  * `wlr_output_damage_swap_buffers` needs to be called. The region of the output
  * that needs to be repainted is added to `damage`.
  */
-bool wlr_output_damage_make_current(struct wlr_output_damage *output_damage,
-	bool *needs_swap, pixman_region32_t *damage);
+bool wlr_output_damage_begin(struct wlr_output_damage *output_damage,
+	struct wlr_renderer *renderer, bool *needs_swap, pixman_region32_t *damage);
```
emersion: The renderer can be obtained from the output.

SirCmpwn:
For the abstraction of backend specific details we can probably abstract it in a simliar way to how we did wlr_renderer's various texture things. We can have a list of named APIs (e.g. EGL) which are supported both by backends and renderers and then they review each others lists when considering which of their own features to enable (or the minimum features required for one or the other to function correctly, possibly leading to an error).

types/wlr_output_damage.c
```diff
@@ -101,14 +101,20 @@ void wlr_output_damage_destroy(struct wlr_output_damage *output_damage) {
 	free(output_damage);
 }
 
-bool wlr_output_damage_make_current(struct wlr_output_damage *output_damage,
-		bool *needs_swap, pixman_region32_t *damage) {
+bool wlr_output_damage_begin(struct wlr_output_damage *output_damage,
```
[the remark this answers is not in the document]
nyorain: Yes, that's better. The function started rendering for some time but that was a bad idea.

include/wlr/types/wlr_output_damage.h
nyorain: oh we don't even need the renderer anymore, good catch.

ascent12:
After some discussions on IRC, I've decided that it would be best to change the way the DRM buffer handling works before considering this for merging.

I'm also inclined to agree with @atomnuker about waiting for the proper modifier extension to happen. At least until everything is done "properly" and is shown to work well on a wide range of hardware/drivers,  I would only merge this is something listed as experimental that would need to be explicitly enabled (i.e. remove it from any auto-detection logic).

I should have some time to make some DRM changes in the near future, so we'll just leave this PR on hold until then. However, I do suspect that this PR may require quite a large reworking.

nyorain:
I'm okay with putting this on hold for now, merging this as experimental works as well but doesn't make much sense when there really is a rework needed. Can't do anything about the modifier extension at the moment but with drm buffer handling you mean the way the gbm render surfaces are implemented at the moment, right? Do you mean explicit fencing (which seems to be required to do vulkan on drm without khr_display correctly) with the drm changes or do you also plan to move the entire custom gbm_bo handling into the drm backend? What else isn't done "properly" and can be done before you consider accepting this?

ascent12:
>with drm buffer handling you mean the way the gbm render surfaces are implemented at the moment, right?
>do you also plan to move the entire custom gbm_bo handling into the drm backend?

Right.

>Do you mean explicit fencing (which seems to be required to do vulkan on drm without khr_display correctly)

Maybe implicit fencing will work here? I'm not 100% sure.
Explicit fencing is one of those things I wanted to get around to anyway, but it might not be needed for this.

>What else isn't done "properly" and can be done before you consider accepting this?

The modifier thing is what I was mainly thinking of. I haven't poured over the PR very deeply yet, but I want to make sure everything is done how it's supposed to be done.

fooishbar:
> Do you mean explicit fencing (which seems to be required to do vulkan on drm without khr_display correctly)

Explicit fencing is required, yes.

SirCmpwn:
I would rather not let this pull request grow stale, full of conflicts, etc. This is 4,000 lines of code that is going to quickly rot if we leave it here until we finish refactoring elsewhere. Having this in the tree wouldn't make the other backends worse, would it?

SirCmpwn:
This only took 3 days

nyorain:
hm so what if i disable it by default in meson, remove it as fallback when the gles2 renderer fails initializing (in renderer_autocreate) and maybe add a warning in meson/when it's used at runtime that it's only experimental and not ready as default renderer, is that ok for everybody? This pr obviously touches some parts other than the new renderer, like moving around the code for gles2, adding the whole wlr_render_surface thing and changing/fixing the way we set the cursor on wayland (due to blocking eglSwapBuffers) so it would be a good idea to at least make sure it does not break anything on any system for the gles2 renderer (could only test it on two systems). Other than that a large part of this pr are the interface changes that allow non-gles2 renderers anyways and yes, having this merged soon would relieve me of some work making sure it stays up to date. I (and whoever else may be interested) could then start sending smaller prs that add the missing features.

emersion:
>changing/fixing the way we set the cursor on wayland (due to blocking eglSwapBuffers)

https://github.com/swaywm/wlroots/issues/1174

SirCmpwn:
>hm so what if i disable it by default in meson, remove it as fallback when the gles2 renderer fails initializing (in renderer_autocreate) and maybe add a warning in meson/when it's used at runtime that it's only experimental and not ready as default renderer, is that ok for everybody?

+1. Make sure the `.build.yml` explicitly enables it, though.

nyorain:
Regarding #1174, it fixes that for the cursor surface only by tracking frame callbacks.

include/render/vulkan.h
```diff
@@ -0,0 +1,252 @@
...
+	struct wlr_renderer *renderer, uint32_t width, uint32_t height);
+struct wlr_render_surface *vulkan_render_surface_create_xcb(
+	struct wlr_renderer *renderer, uint32_t width, uint32_t height,
+	void *xcb_connection, uint32_t xcb_window);
+struct wlr_render_surface *vulkan_render_surface_create_wl(
+	struct wlr_renderer *renderer, uint32_t width, uint32_t height,
+	struct wl_display *disp, struct wl_surface *surf);
+struct wlr_render_surface *vulkan_render_surface_create_gbm(
+	struct wlr_renderer *renderer, uint32_t width, uint32_t height,
+	struct gbm_device *gbm_dev, uint32_t flags);
```
SirCmpwn: I guess these are unimplemented?

backend/drm/renderer.c
```diff
@@ -52,112 +48,34 @@ void finish_drm_renderer(struct wlr_drm_renderer *renderer) {
...
-	return true;
-
-error_gbm:
-	gbm_surface_destroy(surf->gbm);
-error_zero:
-	memset(surf, 0, sizeof(*surf));
-	return false;
+	*surf = wlr_render_surface_create_gbm(
+		renderer->wlr_rend, width, height, renderer->gbm, flags);
+	return (*surf);
```
SirCmpwn: These parens can go

SirCmpwn:
I guess this comment got left behind, but I want to rethink this:

>I would rather have per-backend code in the renderer than per-renderer code in the backend.

More people will write custom renderers than custom backends, let's optimize for that use-case.

ammen99:
@nyorain It seems that you changed the meaning if `create_renderer_func` parameters to backends, now they initialize just the renderer, and don't have control over EGL. However, this means compositors cannot add additional attributes to the EGL config, which I implemented in #987. I don't have the time to look in-depth at your PR, so I can't propose a solution, but I think this needs to be addressed.

include/render/vulkan.h
nyorain: Implemented in render/vulkan/render_surface.c

nyorain:
> I would rather have per-backend code in the renderer than per-renderer code in the backend.

Yeah, that's how it's handled at the moment. If every backend had renderer specific code, writing custom renderers would not be possible without touching the backends. At the moment it's impossible to create a completely new backend (i.e. using new primitives; unable to use an existent render surface implementation) without touching the renderer but that shouldn't happen too often and we can't get around this.

nyorain:
@ammen99 thanks for the notice, this should be fixed now. Since the vulkan renderer obviously doesn't need a wlr_egl object and wlroots isn't that egl fixed anymore it's slightly different than before. Your custom renderer create func could look like this:

```c
struct wlr_renderer *create_custom_gles_renderer(struct wlr_backend *backend) {
	EGLenum platform;
	void *remote_display;
	const EGLint *config_attribs;
	EGLint visualid;

	if (!wlr_backend_egl_params(backend, &platform, &remote_display,
			&config_attribs, &visualid)) {
		wlr_log(WLR_ERROR, "Backend does not support egl");
		return NULL;
	}

	// create changed_config_attribs
	// process the given config attribs: change them/add new ones
	// for adding new ones you have to copy them first into a new buffer

	return wlr_gles2_renderer_create_for_egl(backend, platform, remote_display,
		changed_config_attribs, visualid);
}
```

As before, the compositor can still have the final word about the egl config attribs used (but should query the backends requirements/default attribs). By default the backends egl params are used without any processing (wlr_gles2_renderer_create).

SirCmpwn:
Can you enumerate the touchpoints between the renderers and the backends?

nyorain:
It boils down to
  - `wlr_render_surface_create{xcb,wl,headless,gbm}`, called from backend. The backends then own those render surfaces and will interact with them, currently only via `wlr_render_surface_swap_buffers` and `wlr_render_surface_resize` and the drm backend additionally via `wlr_render_surface_get_bo` to retrieve the front buffer object.
  -  backend-specific attributes retrieved by the renderer/renderer_create_func. At the moment only `wlr_backend_egl_params` (but it might include e.g. required vulkan extensions later on).

Renderers are still created in the backend as before via a given renderer create callback. Instead of egl specific arguments this callback gets passed the backend, the create_func or renderer itself can then retrieve properties from the backend, e.g. these egl arguments via `wlr_backend_egl_params`. There are currently no hardcoded backend type switches (i.e. `wlr_backend_is_xxx`) in the renderer implementations.

For every output the backend creates, it will ask the renderer to create a `wlr_render_surface`. This is where the renderers have backend-specific code but this is done explicitly via the new `render_surface_create_{xcb,wl,headless,gbm}` functions in the renderer interface. We need them since the method of creating render surfaces in the renderers are backend dependent. But they can't that easily be moved into the backend either since e.g. the gles2 renderer needs the extra wl_egl_window primitive on the wayland backend and the way gles2/vulkan implement the gbm render surface is completely different. Another problem is that render surfaces are not only created for outputs but sometimes also e.g. for cursor surfaces (in the wl, drm backends).
Instead of calling `eglSwapBuffers`, outputs will now call `wlr_render_surface_swap_buffers` in their own swap_buffers implementation, which in turn will call `eglSwapBuffers` or issue a `vkQueuePresentKHR` respectively. The same for `wlr_render_surface_resize`, which will recreate the egl surface/resize the wl_egl_window or recreate the swapchain respectively.

SirCmpwn:
I don't understand how the vulkan renderer gets away without having an implementation of `wlr_render_surface_create{xcb,wl,headless,gbm}`

nyorain:
`wlr_render_surface_create{xcb,wl,headless,gbm}` is implemented in wlr_render_surface.c and just calls into the renderer interface. The vulkan renderer implements the work at the bottom of `render/vulkan/render_surface.c`.

XakepSDK:
PR dead?

emersion:
Nope, we're just waiting for progress on https://github.com/swaywm/wlroots/pull/1355 before working again on this one. The renderer redesign should make a lot of things simpler.

nyorain:
also waiting for mesa to implement VK_EXT_image_drm_format modifier. There was at least an implementation for anv (intel) before the extension was even released but that hasn't made it upstream yet afaik.

martell:
@nyorain @emersion can you share some insight with respect to `VK_EXT_image_drm_format_modifier`

I've crawled the web for this a few times since the holidays and I found the initial intel branch / patchset but after this became part of the official vulkan spec I don't see much more with respect needing implementation.

Surely this isn't all that is needed to implement the extension.
https://patchwork.freedesktop.org/patch/186926/

I'm willing to do the MESA work with respect to `VK_EXT_image_drm_format_modifier` so that we can get a vulkan backed wlroots for sway.
Just need a pointer :)

ascent12:
@martell That's not the main thing this PR is blocked on.
I'm doing a bunch of changes to the existing renderer, which would allow a vulkan renderer to be structured much more cleanly, among a bunch of other new things. See #1355 and #1352.
Progress has been fairly slow, though.

martell:
@ascent12 I was tracking that PR for awhile, I was going to comment on the PR asking the status, I saw that it was forced pushed 18 days ago.
damn keeping that conflict free must be annoying :) `backend/drm/drm.c`
I can not help much with something you are all already spec'ing out and working on but the mesa bit seemed a bit isolated which is why I chimed in here.
Also the sooner it gets into mesa the sooner it gets into a stable release, I doubt we can put it in 19.x.

martell:
Tracking stuff related to mesa from the outside is a little difficult but I came across this today when looking into this again https://github.com/st3r4g/swvkc

There is a link to this PR https://gitlab.freedesktop.org/mesa/mesa/merge_requests/515
Which has been superseeded by https://gitlab.freedesktop.org/mesa/mesa/merge_requests/1466
So there is an implementation of `VK_EXT_image_drm_format_modifier` for `anv`.

This seems like a good reference for someone just starting out with contributing so I'll try pick up a Radeon RX 5500 when it comes out ( early December ) and I will implement and test a `VK_EXT_image_drm_format_modifier` implementation for `radv`

martell:
On the wlroots side it seems that switching to GBM is the next step for vulkan support.
`VK_EXT_image_drm_format_modifier` also expects to use gbm ref: @ascent12 's comment here under the weston PR for the example gbm client. https://gitlab.freedesktop.org/wayland/weston/merge_requests/183#note_177045

and there has been code pulled out from the renderer redesign for gbm here. 
https://github.com/swaywm/wlroots/pull/1598

This looks like a massive undertaking.

There also seems to be some bits related to `zwp_linux_dmabuf_v1` and multi gpu issues that are being addressed with a `wp_linux_dmabuf_hints` by @emersion here https://patchwork.freedesktop.org/patch/263061/?series=52370&rev=2
which is needed for https://gitlab.freedesktop.org/wayland/wayland/issues/59
so that `wl_drm` can be dropped ?

Though I'm not quite sure if this is the same lagacy drm that @ascent12 references in the above PR.

Sorry if this is noisy btw, I'm trying to gather and expose context here.

letalvoj:
If this get's ever finished, would it be technically possible to use the new unofficial https://github.com/Yours3lf/rpi-vk-driver Vulkan driver to run a snappy VM on an ancient RPi? That be super awesome.

nyorain:
In theory, yes. In practice, this renderer is not too useful unless `VK_EXT_image_drm_format_modifier` is supported by the driver, so it depends on the likeliness of that being supported in the rpi driver in the future.

Btw, when `VK_EXT_image_drm_format_modifier` gets finally merged at least for the intel driver and this could be useful, would you be interested in a fixed-up and rebased version of this, @emersion, @ascent12, @ddevault? Or is vulkan support definitely a no-go until the main parts of renderer v6 are in? I'd be still interested in working on this (and on wlr in general), but it's been some time since I've looked into the wlr codebase and rebasing this would be quite some work I reckon. If you don't see this being used in the end, feel free to close the PR.

ddevault:
It probably does not make sense to continue investing in this until the renderer redesign settles down. There are still a lot of moving parts and most of this code would have to be rewritten or overhauled.

emersion:
>when VK_EXT_image_drm_format_modifier gets finally merged at least for the intel driver and this could be useful

`VK_EXT_image_drm_format_modifier` is merged, we're waiting on Vulkan drivers to catch up now.

When that's done, if you're interested to work on this, you could try adding a Vulkan renderer to [glider](https://github.com/emersion/glider). Glider is the test compositor I've been using, and it's close to what renderer v6 will look like. The renderer basically needs to import GBM buffers and render to them.

I've written about renderer v6 in my last status update: https://emersion.fr/blog/2020/status-update-19/

Example of `VK_EXT_image_drm_format_modifier` usage in ffmpeg: https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/hwcontext_vulkan.c#L1955

emersion:
I think we're pretty close to being able to have a Vulkan renderer now that the `wlr_renderer` API doesn't depend on EGL anymore. I've begun to work on a Vulkan allocator in https://github.com/swaywm/wlroots/pull/2648. Would be nice to try to rebase this pull request on top of master.

nyorain:
Alright, I'll look into it but this will take a while. I don't think it's useful anyways before a mesa driver supports `VK_EXT_image_drm_format_modifier` since that will be required even for basic testing.

emersion:
`VK_EXT_image_drm_format_modifier` support shouldn't be too far away. We can already try these:

* ANV patch: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/1466
* RADV patch: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/7667

martell:
I noticed the turnip driver just casually enabled `VK_EXT_image_drm_format_modifier` a few months ago https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/6940 but it is not in `docs/features.txt`.

nyorain:
Superseded by #2771.

Issue #409, opened 2017-11-11
Segfault on launch with 3 screens when only 2 are supported


emersion:
I have a laptop with two screens connected (so 3 screens total). The GPU only supports 2 at the same time.

This used to work.

What happens exactly:

1. I launch rootston
2. rootston hangs, no screen enabled
3. Switched to another VT
4. Switched back to rootston
5. rootston segfaults

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f260577d8b1 in wlr_drm_connector_cleanup (conn=0x55f1f4a6feb0) at ../backend/drm/drm.c:914
914				if (!crtc->planes[i]) {
(gdb) bt
#0  0x00007f260577d8b1 in wlr_drm_connector_cleanup (conn=0x55f1f4a6feb0) at ../backend/drm/drm.c:914
#1  0x00007f260577e346 in wlr_drm_connector_set_mode (output=0x55f1f4a6feb0, mode=0x55f1f4a1dd10) at ../backend/drm/drm.c:487
#2  0x00007f26057976ae in wlr_output_set_mode (output=0x55f1f4a6feb0, mode=0x55f1f4a1dd10) at ../types/wlr_output.c:153
#3  0x00007f260577c11f in session_signal (listener=0x55f1f4299380, data=0x55f1f427b250) at ../backend/drm/backend.c:70
#4  0x00007f260578bbbb in wl_signal_emit (signal=0x55f1f427b258, data=0x55f1f427b250) at /usr/include/wayland-server-core.h:388
#5  0x00007f260578bb40 in resume_device (msg=0x55f1f4acc150, userdata=0x55f1f427b250, ret_error=0x7ffc410449d0) at ../backend/session/logind.c:288
#6  0x00007f2602aeef07 in  () at /usr/lib/libsystemd.so.0
#7  0x00007f2602aeec3f in  () at /usr/lib/libsystemd.so.0
#8  0x00007f2602aef19e in  () at /usr/lib/libsystemd.so.0
#9  0x00007f2602aeec3f in  () at /usr/lib/libsystemd.so.0
#10 0x00007f2602aef19e in  () at /usr/lib/libsystemd.so.0
#11 0x00007f2602aeec3f in  () at /usr/lib/libsystemd.so.0
#12 0x00007f2602aef19e in  () at /usr/lib/libsystemd.so.0
#13 0x00007f2602aeec3f in  () at /usr/lib/libsystemd.so.0
[...]

PR #1215, opened 2018-09-01
Do not modeset disabled outputs

emersion:
This allows to get a usable setup by disabling an output when a graphics card only supports 2 screens and 3 are connected. See #409

This doesn't fix the bug described in #409 though.


SirCmpwn:
Thanks!

PR #1216, opened 2018-09-01
Init dmabuf global in renderer

emersion:
Fixes https://github.com/swaywm/wlroots/issues/1190


emersion:
@Ongy @ammen99: this requires changes in your compositors (tl;dr don't call `wlr_linux_dmabuf_v1_create`).

PR #1219, opened 2018-09-01
backend/drm: allow disabling outputs in NEEDS_MODESET state

emersion:
This correctly frees CRTCs when disabling outputs without setting
a mode.

Fixes the situation in which some CRTCs were in use by disabled outputs and we want to re-allocate them to other outputs. The first modeset attempt still fails, so this isn't a full fix.

Fixes part of #409


SirCmpwn:
Thanks!

PR #1222, opened 2018-09-02
xwayland: Add WM_STATE modal property

RyanDwyer:
Adds a `modal` property to indicate whether the surface wants to be a modal.

<s>Also adds an `attention` property to indicate whether the surface wants attention, and introduces a `request_attention` event.</s> I've removed this because the `DEMANDS_ATTENTION` atom doesn't seem to do what I thought it would.


emersion:
Thanks!

PR #1225, opened 2018-09-02
fix incorrect NULL check in get_tablet_seat

cnt0:
there was an obvious typo leading to NULL pointer dereference


emersion:
Thanks!

PR #1226, opened 2018-09-02
xwayland: Introduce set_role event

RyanDwyer:
Do we want to name the event `set_role` or `set_window_role`?


xwayland/xwm.c
```diff
@@ -365,6 +368,36 @@ static void read_surface_class(struct wlr_xwm *xwm,
 	wlr_signal_emit_safe(&surface->events.set_class, surface);
 }
 
+#ifdef WLR_HAS_XCB_ICCCM
```
emersion: Why does this require ICCCM?

emersion:
>Do we want to name the event set_role or set_window_role?

I'm fine with `set_role`

xwayland/xwm.c
RyanDwyer: Thanks for the explanation. I've removed the `#ifdef`.

PR #1228, opened 2018-09-03
Miscellaneous memory leak fixes

arandomhuman:
The changes in wlroot include a very minor leak in x11/backend.c (7bc2657) and a bug in tablet_manager_v2 (9f511ae).

The other commits to the examples include setting keymaps more securely (7105864), and removing pointers in multi-pointer (6af77e3).

Please let me know if some commits are better merged squashed (say, all the edits to examples).


emersion:
Thanks!

PR #1229, opened 2018-09-04
backend/drm: better hotplug handling

emersion:
This commit handles better situations in which the number of
connected outputs is greater than the number of available CRTCs.
It'll enable as many outputs as possible, and transfer CRTCs to
outputs that need one on unplug.

This changes CRTC and plane reallocation to happen after scanning
DRM connectors instead of on modeset.

This cleanups CRTCs and planes on unplug to allow them to be
re-used for other outputs.

On modeset, if an output doesn't have a CRTC, the desired mode is
saved and used later when the output gains a CRTC.

Future work includes giving priority to enabled outputs over
disabled ones for CRTC allocation. This requires the compositor to
know about all outputs (even outputs without CRTCs) to properly
modeset outputs enabled in the compositor config file and disable
outputs disabled in the config file.

Known issues/more future work:
- ~~Falls back to software cursors in some situations~~ Falls back to software cursors when output doesn't have CRTC, but upgrades to hardware cursors when the output gains a CRTC
- Fails a modeset on shutdown, seems harmless
- Outputs without CRTCs show up as disabled outputs, we might want to find a better solution

Fixes part of #409

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -404,9 +404,11 @@ static void realloc_planes(struct wlr_drm_backend *drm, const uint32_t *crtc_in,
 			struct wlr_drm_plane *new = &drm->type_planes[type][crtc_res[i]];
 
 			if (*old != new) {
-				wlr_log(WLR_DEBUG, "Assigning plane %d -> %d to CRTC %d",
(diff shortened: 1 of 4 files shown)


SirCmpwn:
Thanks!

PR #1231, opened 2018-09-05
Remove indent_size from .editorconfig

ascent12:
I've recently bothered to configure my editor to actually use this.
Some of us like to use different indent sizes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,10 +5,8 @@ end_of_line = lf
 insert_final_newline = true
 charset = utf-8
 indent_style = tab
-indent_size = 4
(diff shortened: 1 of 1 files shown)


SirCmpwn:
My only concern is that this might make it harder for people with some configurations to notice when they have hit 80 columns.

emersion:
`.xml` protocol files use the Wayland code style, so the indent size should remain at least there.

And yeah, I had issues in the past setting a different indent size and not realizing I hit the 80 char limit.

SirCmpwn:
Sure would be nice if clang-format understood sway style...

PR #1233, opened 2018-09-08
Fix wlr_xwayland_destroy

nyorain:
In rootston, wlr_xwayland_destroy is always called (when compiled with xwayland) which causes a segmentation fault on exit when disabling xwayland per config (e.g. example config).
We could also just fix this by checking in rootston, but this follows the wlr convention of most destroy functions that passing NULL is ok.


emersion:
Thanks!

PR #1234, opened 2018-09-10
fix cursor loss w/ legacy drm and software cursor (tested on etnaviv)

mntmn:
I tested sway today on an etnaviv platform (imx6qp/reform) and it worked fine except that the cursor disappeared as soon as I moved it into the swaybar and would not reappear. This was because swaybar clears and then immediately sets a cursor. Both fail in DRM, but the failure of clearing is not seen, leading to a wrong state. This PR adds a missing error handling for the drmModeSetCursor() case where an empty cursor is set.

### DIFF
--- a/backend/drm/legacy.c
+++ b/backend/drm/legacy.c
@@ -39,7 +39,10 @@ bool legacy_crtc_set_cursor(struct wlr_drm_backend *drm,
 	}
 
 	if (!bo) {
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

(Successful build: https://builds.sr.ht/~emersion/job/7265)

PR #1235, opened 2018-09-10
backend/drm: de-allocate unused CRTCs

emersion:
They can be re-used by another output after a subsequent hotplug.


SirCmpwn:
Thanks!

PR #1236, opened 2018-09-10
Add a `tags` option, don't ls-files if not needed

emersion:
Related to #725


SirCmpwn:
Now that the build has been resolved...

We've talked about this *several* times. The solution has been *explicitliy* laid out for you before. This is not the correct solution. Figure it the fuck out or quit fucking bothering me about this shit.

PR #1239, opened 2018-09-11
rootston: Keep older anchored layers near anchor points

dcz-purism:
Layer surfaces are attached to edges of the screen starting with the youngest, causing new ones to always displace existing ones. This changes the order to oldest first, keeping the positions more often.

With this change, layers attached to edges have a more intuitive feeling of "building upon" the edge.


SirCmpwn:
Thanks!

PR #1243, opened 2018-09-14
layer-shell: add _v1 suffix

emersion:
cc @Ongy, @ammen99: this is a breaking change.


SirCmpwn:
Thanks! @swaywm/wlroots-go @swaywm/wlroots-ocaml @swaywm/cl-wlroots

PR #1244, opened 2018-09-14
xdg-output: add _v1 suffix

emersion:
cc @Ongy: this is a breaking change (this should matter to bindings implementors, but not really to compositor implementors).

This doesn't require a sway update.


SirCmpwn:
Thanks! @swaywm/wlroots-go @swaywm/wlroots-ocaml @swaywm/cl-wlroots

PR #1245, opened 2018-09-17
Disable hardware cursors when capturing output

emersion:
We need to track the number of times hardware cursors have been disabled to properly support e.g. screenshooting and screen recording at the same time.

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -107,6 +107,7 @@ struct wlr_output {
 
 	struct wl_list cursors; // wlr_output_cursor::link
 	struct wlr_output_cursor *hardware_cursor;
+	int software_cursor_locks; // number of locks forcing software cursors
 
 	// the output position in layout space reported to clients
 	int32_t lx, ly;
@@ -196,6 +197,14 @@ bool wlr_output_export_dmabuf(struct wlr_output *output,
 void wlr_output_set_fullscreen_surface(struct wlr_output *output,
 	struct wlr_surface *surface);
 struct wlr_output *wlr_output_from_resource(struct wl_resource *resource);
+/**
+ * Locks the output to only use software cursors instead of hardware cursors.
+ * This is useful if hardware cursors need to be temporarily disabled (e.g.
+ * during screen capture). There must be as many unlocks as there have been
+ * locks to restore the original state. There should never be an unlock before
+ * a lock.
+ */
+void wlr_output_lock_software_cursors(struct wlr_output *output, bool lock);
 
 
 struct wlr_output_cursor *wlr_output_cursor_create(struct wlr_output *output);
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -273,6 +273,13 @@ void wlr_output_init(struct wlr_output *output, struct wlr_backend *backend,
 	wl_signal_init(&output->events.destroy);
 	pixman_region32_init(&output->damage);
 
+	const char *no_hardware_cursors = getenv("WLR_NO_HARDWARE_CURSORS");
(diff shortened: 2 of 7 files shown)


include/wlr/types/wlr_output.h
```diff
@@ -107,6 +107,7 @@ struct wlr_output {
 
 	struct wl_list cursors; // wlr_output_cursor::link
 	struct wlr_output_cursor *hardware_cursor;
+	int hardware_cursors_enabled; // > 0 when enabled
```
SirCmpwn: I think this variable name is a little misleading. If it were just "are hardware cursors enabled" it should be a bool. I'd rather be more transparent about how it's used.

arandomhuman: Since it's roughly capturing `enable requests - disable requests`, maybe a better name could be `hardware_cursors_tokens` or `hardware_cursors_enable_requests`?

Would it make sense for the value to be > 1? If not, why not enable them at 0, and count disable requests?

types/wlr_output.c
```diff
@@ -666,6 +673,32 @@ struct wlr_output *wlr_output_from_resource(struct wl_resource *resource) {
...
+		assert(output->impl->set_cursor);
+		output->impl->set_cursor(output, NULL, 1,
+			WL_OUTPUT_TRANSFORM_NORMAL, 0, 0, true);
+		output_cursor_damage_whole(output->hardware_cursor);
+		output->hardware_cursor = NULL;
+	}
+
+	// If it's possible to use hardware cursors again, don't switch immediately
+	// since a recorder is likely to lock software cursors for the next frame
+	// again.
```
atomnuker: Well, when are hardware cursors restored then?

emersion: The next time the cursor is updated again.

atomnuker: Doesn't sound very robust. Could the cursor perhaps be updated after a client detaches from the manager interface?

emersion: This would complicate this already-complicated code. It's not like using software cursors for a few seconds is the end of the world.

atomnuker: It could be an entire eternity if for instance this is a fullscreen shell with a single output. I don't think it would add more than a few lines.

emersion: Doing this would require to track bound globals (ie. allocate a per-bound-global struct to contain the lock state), which we don't currently do. I prefer to keep this as a follow-up issue since we don't support fullscreen shell (so the situation you describe is very unlikely to happen).

SirCmpwn:
Thanks!

types/wlr_output.c
atomnuker: The situation is very likely to happen even in sway, all you have to do is not move your mouse over another window after you stop capturing. Not restoring the state and requiring compositors to hack around this for you is a hack in of itself. Moreover, the function doesn't actually do what it states it should do, and changing this behavior in the future could be an API break.

emersion: Even if you don't move your mouse over another client, you're very likely to update it anyway by moving over a text input or a link. We're of course not requiring compositors to hack around this. This API is nowhere near stable so I'm not concerned about changing its semantics.

SirCmpwn: >It's not like using software cursors for a few seconds is the end of the world.

This is why I don't care

atomnuker: It would have been absolutely no problem if it was just a few seconds, as it stands though it could go to infinity in case of e.g. embedded, single program systems where the cursor doesn't change if it goes over text or not.

emersion: Do we really care about recording the screen with the cursor on such systems?

Also support for writeback DRM feature will allow us to stop doing that on the DRM backend (and drivers that support it).

PR #1246, opened 2018-09-17
output-layout: don't expose outputs without a mode

### LINKED ISSUE #1242: DRM: decide what to do with outputs missing a CRTC
Right now, they show up as disabled without a mode (and with a zero size). However I'm not sure this is the right thing to do.

Some backends don't support a mode, so outputs without a mode can be perfectly fine outputs.

Disabled outputs can be just turned off by the user (DPMS), they just don't emit frame events, but still show up as Wayland globals.

The issue is that outputs without a CRTC show up in the output layout and are exposed as a Wayland global. Some sway clients handle this poorly, and I don't think these outputs should be visible to clients anyway.

As said before, we still want these outputs to be visible to the compositor so it can modeset them and disable other outputs depending on the user configuration.

Possible solutions:
* In the output layout, don't expose outputs that have a non-empty list of modes but don't have a current mode. We'll probably do similar checks in sway too (when e.g. spawning `swaybg`).
* Add a new output property basically saying "this output isn't available right now"

emersion:
See #1242

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -134,10 +134,20 @@ static void output_layout_reconfigure(struct wlr_output_layout *layout) {
 	wlr_signal_emit_safe(&layout->events.change, layout);
 }
 
+static void output_update_global(struct wlr_output *output) {
+	// Don't expose the output if it doesn't have a current mode
+	if (wl_list_empty(&output->modes) || output->current_mode != NULL) {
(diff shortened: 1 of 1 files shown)


SirCmpwn:
What does this do for e.g. the wayland backend?

SirCmpwn:
Thanks!

PR #1248, opened 2018-09-17
Assert all the things!

### LINKED ISSUE #507: Double check that we're asserting in all of the right places
Every type-specific function should assert the resource is actually of that type. There are probably other places we can assert, too.

emersion:
Fixes #507

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -205,16 +205,22 @@ void finish_drm_resources(struct wlr_drm_backend *drm) {
 	free(drm->planes);
 }
 
+static struct wlr_drm_connector *get_drm_connector_from_output(
+		struct wlr_output *wlr_output) {
+	assert(wlr_output_is_drm(wlr_output));
+	return (struct wlr_drm_connector *)wlr_output;
+}
+
 static bool drm_connector_make_current(struct wlr_output *output,
 		int *buffer_age) {
-	struct wlr_drm_connector *conn = (struct wlr_drm_connector *)output;
+	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
 	return make_drm_surface_current(&conn->crtc->primary->surf, buffer_age);
 }
 
 static bool drm_connector_swap_buffers(struct wlr_output *output,
 		pixman_region32_t *damage) {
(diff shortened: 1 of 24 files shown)


backend/drm/drm.c
```diff
@@ -205,16 +205,22 @@ void finish_drm_resources(struct wlr_drm_backend *drm) {
 	free(drm->planes);
 }
 
+static struct wlr_drm_connector *get_drom_connector_from_output(
```
SirCmpwn: >drom

backend/libinput/events.c
```diff
@@ -150,10 +163,12 @@ static void handle_device_added(struct wlr_libinput_backend *backend,
...
 		wlr_signal_emit_safe(&backend->backend.events.new_input, wlr_dev);
 	}
-	if (libinput_device_has_capability(libinput_dev, LIBINPUT_DEVICE_CAP_GESTURE)) {
+	if (libinput_device_has_capability(libinput_dev,
+			LIBINPUT_DEVICE_CAP_GESTURE)) {
 		// TODO
 	}
-	if (libinput_device_has_capability(libinput_dev, LIBINPUT_DEVICE_CAP_SWITCH)) {
+	if (libinput_device_has_capability(libinput_dev,
+			LIBINPUT_DEVICE_CAP_SWITCH)) {
```
SirCmpwn: I think these might be better balanced if you cut after the `(`

backend/headless/backend.c
```diff
@@ -7,10 +7,17 @@
 #include <wlr/util/log.h>
 #include "backend/headless.h"
 #include "glapi.h"
+#include "util/signal.h"
+
+struct wlr_headless_backend *headless_backend_from_backend(
+		struct wlr_backend *wlr_backend) {
```
agx: Should this be `get_headless_backend_from_backend()` to be consistent with the other `get_foo_from_bar()`'s above?

emersion: The others have a `get_` prefix to prevent clashes with library names (`drm`, `libinput`, `wl`, etc...).

emersion:
Addressed @SirCmpwn's comments

SirCmpwn:
Thanks!

PR #1249, opened 2018-09-17
Send unique keymap file descriptors

### LINKED ISSUE #1192: wl_keyboard keymap can be written to by clients
The FD is shared between all clients. We need to isolate each client.

See https://gitlab.freedesktop.org/wayland/weston/commit/76829fc4eaea329d2a525c3978271e13bd76c078

arandomhuman:
Fixes #1192 

To prevent wl_keyboard keymap being written to by clients, use a unique
file descriptor for each wl_keyboard resource.

Reference: [weston, commit 76829fc4eaea329d2a525c3978271e13bd76c078](https://gitlab.freedesktop.org/wayland/weston/commit/76829fc4eaea329d2a525c3978271e13bd76c078)

### DIFF
--- a/types/seat/wlr_seat_keyboard.c
+++ b/types/seat/wlr_seat_keyboard.c
@@ -3,6 +3,8 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <sys/mman.h>
+#include <unistd.h>
 #include <wayland-server.h>
 #include <wlr/types/wlr_data_device.h>
 #include <wlr/types/wlr_input_device.h>
@@ -10,6 +12,7 @@
 #include <wlr/util/log.h>
 #include "types/wlr_seat.h"
 #include "util/signal.h"
+#include "util/os-compatibility.h"
 
 static void default_keyboard_enter(struct wlr_seat_keyboard_grab *grab,
 		struct wlr_surface *surface, uint32_t keycodes[], size_t num_keycodes,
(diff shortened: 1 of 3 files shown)


types/seat/wlr_seat_keyboard.c
```diff
@@ -330,6 +334,9 @@ static void seat_client_send_keymap(struct wlr_seat_client *client,
 		return;
 	}
 
+	int keymap_fd = -1;
+	void *ptr = NULL;
```
emersion: These can be defined when needed

types/seat/wlr_seat_keyboard.c
```diff
@@ -338,8 +345,25 @@ static void seat_client_send_keymap(struct wlr_seat_client *client,
 			continue;
 		}
 
+		keymap_fd = os_create_anonymous_file(keyboard->keymap_size);
```
emersion: This is never closed on the compositor side. It should be closed right after `wl_keyboard_send_keymap`

types/wlr_keyboard.c
```diff
@@ -194,26 +190,13 @@ void wlr_keyboard_set_keymap(struct wlr_keyboard *kb,
 		kb->mod_indexes[i] = xkb_map_mod_get_index(kb->keymap, mod_names[i]);
 	}
 
-	keymap_str = xkb_keymap_get_as_string(kb->keymap,
+	kb->keymap_string = xkb_keymap_get_as_string(kb->keymap,
```
emersion: The previous `kb->keymap_string` should be free'd before replacing it with the new one. Preferably it should be replaced only if `xkb_keymap_get_as_string` doesn't fail.

types/seat/wlr_seat_keyboard.c
```diff
@@ -11,6 +13,8 @@
 #include "types/wlr_seat.h"
 #include "util/signal.h"
 
+int os_create_anonymous_file(off_t size);
```
agx: Rather include os-compatibility.h

agx: Rather include "util/os-compatibility.h"?

types/seat/wlr_seat_keyboard.c
arandomhuman: I'd rather initialize variables on declaration; I'll remove it if you prefer that.

emersion: No, I mean these can be declared later, on lines 347 and 353.

types/wlr_keyboard.c
```diff
@@ -126,7 +122,7 @@ void wlr_keyboard_init(struct wlr_keyboard *kb,
 	wl_signal_init(&kb->events.modifiers);
 	wl_signal_init(&kb->events.keymap);
 	wl_signal_init(&kb->events.repeat_info);
-	kb->keymap_fd = -1;
+	kb->keymap_string = NULL;
```
emersion: No need to initialize to NULL, this is calloc'ed.

types/wlr_keyboard.c
```diff
@@ -139,8 +135,8 @@ void wlr_keyboard_destroy(struct wlr_keyboard *kb) {
 	}
 	xkb_state_unref(kb->xkb_state);
 	xkb_keymap_unref(kb->keymap);
-	if (kb->keymap_fd >= 0) {
-		close(kb->keymap_fd);
+	if (kb->keymap_string) {
```
emersion: You can just `free` without an `if`, `free(NULL)` is a no-op.

types/seat/wlr_seat_keyboard.c
arandomhuman: Ah, got it. I thought it would throw "already declared" errors; the more you know.

emersion:
Thanks!

PR #1250, opened 2018-09-17
Add support for WLR_SESSION env variable

### LINKED ISSUE #1064: Add environment variable to specify which wlr_session is used
This would be helpful for debugging. I've just been commenting out the logind session in session.c until now, but that's cumbersome because it requires a rebuild.
`WLR_SESSION=logind`, `WLR_SESSION=direct`, etc.

arandomhuman:
Fixes #1064 

Valid values are "logind"/"systemd" and "direct". If WLR_SESSION is set,
only its value is potentially tried; it will not try any other option.

### DIFF
--- a/backend/session/session.c
+++ b/backend/session/session.c
@@ -66,10 +66,25 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 
 struct wlr_session *wlr_session_create(struct wl_display *disp) {
 	struct wlr_session *session = NULL;
-	const struct session_impl **iter;
 
-	for (iter = impls; !session && *iter; ++iter) {
-		session = (*iter)->create(disp);
+	const char *env_wlr_session = getenv("WLR_SESSION");
+	if (env_wlr_session) {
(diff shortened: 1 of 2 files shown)


backend/session/session.c
```diff
@@ -66,10 +66,23 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
...
-	const struct session_impl **iter;
 
-	for (iter = impls; !session && *iter; ++iter) {
-		session = (*iter)->create(disp);
+	const char *env_wlr_session = getenv("WLR_SESSION");
+	if (env_wlr_session) {
+		if (!strcmp(env_wlr_session, "logind") || !strcmp(env_wlr_session, "systemd")) {
+		#if defined(WLR_HAS_SYSTEMD) || defined(WLR_HAS_ELOGIND)
+			session = session_logind.create(disp);
+		#endif
```
ascent12: Can you add an #else with a log message stating that logind support wasn't compiled in?

arandomhuman:
Done, both. Completely forgot about the docs.

SirCmpwn:
Thanks!

PR #1252, opened 2018-09-19
buffer: disconnect clients that commit an unknown buffer type

emersion:
If this happens, something went really wrong and we should do more
than just logging an error. See [1].

[1]: https://gitlab.freedesktop.org/wayland/weston/issues/148


SirCmpwn:
Thanks!

PR #1256, opened 2018-09-21
Apply multiseat fixes to FreeBSD direct backend

sghctoma:
This commit applies the multiseat fixes that are part of PR #1063 (commits 964e0a50 and 47895d2) to the FreeBSD direct backend.


emersion:
Thanks!

PR #1257, opened 2018-09-21
Fix partial dependency in protocol/meson.build

sghctoma:
External dependencies in Meson do not have `include_directories`, therefore `includes: true` means nothing for the `wayland-client` partial dependency in protocol/meson.build. Because of this, the `-I` CFLAGs for wayland-client are not used by the build command. This commit fixes this by using `compile_args`.


emersion:
Thanks!

PR #1261, opened 2018-09-24
Fix damage tracking for surfaces with transforms

arandomhuman:
Fixes bugs with running `weston-simple-damage --transform=90` (for example) and `weston-fullscreen` (use 't' to get a non-zero transform, then resize/scale/move mouse around). Both clients are now working (mostly) perfectly.

*Caveat:* `weston-simple-damage --rotating-transform` leads to a very dirty buffer. Damage tracking is correct though (check in rootston). Weston also behaves the same way. ~~No idea how to fix this.~~ It's a bug with `weston-simple-damage`: it damages the previous location incorrectly when the transformation changes.

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -229,4 +229,12 @@ struct wlr_surface *wlr_surface_from_resource(struct wl_resource *resource);
 void wlr_surface_for_each_surface(struct wlr_surface *surface,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 
+/**
+ * Get the effective damage to the surface in terms of surface local
+ * coordinates. This includes damage induced by resizing and moving the
+ * surface. The damage is not expected to be bounded by the surface itself.
+ */
+void wlr_surface_get_effective_damage(struct wlr_surface *surface,
+	pixman_region32_t *damage);
+
 #endif
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -17,14 +17,14 @@
 #include <wlr/types/wlr_output.h>
 
 enum wlr_surface_state_field {
-	WLR_SURFACE_STATE_BUFFER = 1,
-	WLR_SURFACE_STATE_SURFACE_DAMAGE = 2,
-	WLR_SURFACE_STATE_BUFFER_DAMAGE = 4,
-	WLR_SURFACE_STATE_OPAQUE_REGION = 8,
-	WLR_SURFACE_STATE_INPUT_REGION = 16,
-	WLR_SURFACE_STATE_TRANSFORM = 32,
-	WLR_SURFACE_STATE_SCALE = 64,
-	WLR_SURFACE_STATE_FRAME_CALLBACK_LIST = 128,
+	WLR_SURFACE_STATE_BUFFER = 1 << 0,
+	WLR_SURFACE_STATE_SURFACE_DAMAGE = 1 << 1,
+	WLR_SURFACE_STATE_BUFFER_DAMAGE = 1 << 2,
+	WLR_SURFACE_STATE_OPAQUE_REGION = 1 << 3,
+	WLR_SURFACE_STATE_INPUT_REGION = 1 << 4,
+	WLR_SURFACE_STATE_TRANSFORM = 1 << 5,
+	WLR_SURFACE_STATE_SCALE = 1 << 6,
+	WLR_SURFACE_STATE_FRAME_CALLBACK_LIST = 1 << 7,
 };
 
 struct wlr_surface_state {
@@ -32,7 +32,7 @@ struct wlr_surface_state {
 
 	struct wl_resource *buffer_resource;
 	int32_t dx, dy; // relative to previous position
-	pixman_region32_t surface_damage, buffer_damage;
+	pixman_region32_t surface_damage, buffer_damage; // clipped to bounds
 	pixman_region32_t opaque, input;
 	enum wl_output_transform transform;
 	int32_t scale;
@@ -68,11 +68,12 @@ struct wlr_surface {
 	 * The last commit's buffer damage, in buffer-local coordinates. This
 	 * contains both the damage accumulated by the client via
 	 * `wlr_surface_state.surface_damage` and `wlr_surface_state.buffer_damage`.
-	 * If the buffer has changed its size or moved, the whole buffer is
-	 * damaged.
+	 * If the buffer has been resized, the whole buffer is damaged.
 	 *
 	 * This region needs to be scaled and transformed into output coordinates,
-	 * just like the buffer's texture.
+	 * just like the buffer's texture. In addition, if the buffer has shrunk the
+	 * old size needs to be damaged and if the buffer has moved the old and new
+	 * positions need to be damaged.
 	 */
 	pixman_region32_t buffer_damage;
 	/**
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -696,13 +696,10 @@ static void damage_from_surface(struct wlr_surface *surface, int sx, int sy,
 	int center_x = box.x + box.width/2;
 	int center_y = box.y + box.height/2;
 
-	enum wl_output_transform transform = surface->current.transform;
-
 	pixman_region32_t damage;
 	pixman_region32_init(&damage);
-	pixman_region32_copy(&damage, &surface->buffer_damage);
-	wlr_region_transform(&damage, &damage, transform,
-		surface->current.buffer_width, surface->current.buffer_height);
+	wlr_surface_get_effective_damage(surface, &damage);
+
 	wlr_region_scale(&damage, &damage,
 		wlr_output->scale / (float)surface->current.scale);
 	if (ceil(wlr_output->scale) > surface->current.scale) {
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -700,8 +700,7 @@ static void damage_from_surface(struct wlr_surface *surface, int sx, int sy,
 	pixman_region32_init(&damage);
 	wlr_surface_get_effective_damage(surface, &damage);
 
-	wlr_region_scale(&damage, &damage,
-		wlr_output->scale / (float)surface->current.scale);
+	wlr_region_scale(&damage, &damage, wlr_output->scale);
 	if (ceil(wlr_output->scale) > surface->current.scale) {
 		// When scaling up a surface, it'll become blurry so we need to
 		// expand the damage region
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -696,8 +696,7 @@ static void damage_from_surface(struct wlr_surface *surface, int sx, int sy,
 	int center_x = box.x + box.width/2;
 	int center_y = box.y + box.height/2;
 
-	enum wl_output_transform transform =
-		wlr_output_transform_invert(surface->current.transform);
+	enum wl_output_transform transform = surface->current.transform;
 
 	pixman_region32_t damage;
 	pixman_region32_init(&damage);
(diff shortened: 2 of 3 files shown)


types/wlr_surface.c
```diff
@@ -198,7 +199,8 @@ static void surface_update_damage(pixman_region32_t *buffer_damage,
 		pixman_region32_init(&surface_damage);
 		pixman_region32_copy(&surface_damage, &current->surface_damage);
 		wlr_region_transform(&surface_damage, &surface_damage,
-			current->transform, current->buffer_width, current->buffer_height);
+			wlr_output_transform_invert(current->transform),
+			current->width, current->height);
```
emersion: Re using surface-local coordinates instead of buffer coordinates here: good catch, makes sense

types/wlr_surface.c
```diff
@@ -185,26 +188,23 @@ static void surface_update_damage(pixman_region32_t *buffer_damage,
 			prev_y = tmp;
 		}
 
-		pixman_region32_union_rect(buffer_damage, buffer_damage,
-			prev_x * previous->scale, prev_y * previous->scale,
-			previous->buffer_width, previous->buffer_height);
-		pixman_region32_union_rect(buffer_damage, buffer_damage, 0, 0,
-			current->buffer_width, current->buffer_height);
+		pixman_region32_union_rect(&surface_damage, &surface_damage, prev_x,
```
emersion: I'd prefer using buffer coordinates here.

types/wlr_surface.c
```diff
@@ -301,29 +301,8 @@ static void surface_apply_damage(struct wlr_surface *surface) {
...
-
-		pixman_region32_intersect_rect(&damage, &damage, 0, 0,
-			surface->current.buffer_width, surface->current.buffer_height);
-
 		struct wlr_buffer *updated_buffer =
-			wlr_buffer_apply_damage(surface->buffer, resource, &damage);
-
-		pixman_region32_fini(&damage);
-
+			wlr_buffer_apply_damage(surface->buffer, resource, &surface->buffer_damage);
```
emersion: This doesn't work, because in case the surface moves (`sx` and `sy` are changed), `surface->buffer_damage` is not empty (contains the previous and the new buffer rectangles). However in this case we don't need to re-upload the texture.

types/wlr_surface.c
```diff
@@ -307,8 +309,8 @@ static void surface_apply_damage(struct wlr_surface *surface) {
 		pixman_region32_init(&surface_damage);
 		pixman_region32_copy(&surface_damage, &surface->current.surface_damage);
 		wlr_region_transform(&surface_damage, &surface_damage,
-			surface->current.transform,
-			surface->current.buffer_width, surface->current.buffer_height);
+			wlr_output_transform_invert(surface->current.transform),
```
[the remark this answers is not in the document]
arandomhuman: As far as I understand, the transform (as given in `wlr_region_transform`) is what is to be applied to the buffer to get it in terms of surface coordinates. Here, we go from surface -> buffer coordinates, which means the inverse transform needs to be applied.
~~(Also, it just works™)~~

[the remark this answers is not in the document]
emersion: Oh right, this makes sense.

Note to myself: we need to update compositors' rendering code to remove the `wlr_output_transform_invert` there

types/wlr_surface.c
arandomhuman: That's the source of the bug (in `weston-fullscreen` resizing when transformed). If transformed, the buffer's origin and the surface's origin don't match, so the "overflowing" damage here doesn't correspond to where the surface was previously.
TBH, I don't agree with setting the damage beyond the bounds of the surface here at all: it breaks the implicit assumption of the damage being limited by the surface. IMO, overflowing resizing damage should be handled by the compostor/rootston.

types/wlr_surface.c
arandomhuman: Hm, so there needs to a way to keep track of *why* the buffer was damaged, maybe? Or accumulating move damage separately. See my comment about removing move/resize damage from this stage in the pipeline as well.

types/wlr_surface.c
emersion: We expose both the original damage (via `surface->current`) and the easy-to-use damage. It's up to the compositor to manage it itself.

There's no assumption at all about the damage being inside of the buffer bounds. Maybe we should make this more explicit in the header docs.

arandomhuman: Fair enough. In any case, buffer coordinates really aren't feasible here, since they'd have to be transformed in exactly the same way the surface coordinates are, and that's redundant effort.

types/wlr_surface.c
emersion: Nah, the buffer could have been moved _and_ damaged. We just need to use `surface->current.{buffer,surface}_damage`.

Originally the `buffer_damage` included these move/resize damaged regions as well because there wasn't `surface->previous`. Now that we have this, maybe we could make `buffer_damage` only contain the "real" damage as you'd like, but also provide a new utility function to compute the damage including move/resize. What do you think?

types/wlr_surface.c
emersion: Hmm, what do you mean? We can just use `current->buffer_width` and `current->buffer_height` as it was done previously. No need to use `current->width` and `current->height`.

emersion:
BTW, feel free to join `#sway-devel` on Freenode

types/wlr_surface.c
arandomhuman: Let me take the easiest case: flipped. Suppose originally the surface and buffer was 20x10, with the upper left corner at (10,0). Say someone resized the surface/buffer to 10x10.

If we use `current->buffer_*` here, the buffer damage is ((0,0),(20,10)), which on flipping with buffer size 10x10 (when getting back the surface damage during output damage calculation) gives the damage on the output as (10,0) + ((-10,0),(10,10)) == ((0,0),(20,10)) (translation), which is wrong: it should be ((10,0),(30,10)), the previous surface bounds.

types/wlr_surface.c
arandomhuman: I think a separation does make sense: a client `buffer_damage` (used in `surface_apply_damage`) plus a `total_damage`/`output_damage`/something? We could set them independently in `surface_update_damage` and take the union later, or treat the second variable as the union anyway.

Or, we could accumulate the total damage (resize/move) in terms of surface coordinates, which makes output damage calculation easier (no need to transform again).

emersion: >I think a separation does make sense: a client buffer_damage (used in surface_apply_damage) plus a total_damage/output_damage/something?

I was just thinking of exposing the client damage in `surface->buffer_damage` and adding a new function (`wlr_surface_effective_damage`?) that would compute the total damage.

>Or, we could accumulate the total damage (resize/move) in terms of surface coordinates, which makes output damage calculation easier (no need to transform again).

I decided not to do so so that if you have a surface with scale=2 and an output with scale=2 you don't have to divide by 2 and then multiply by 2 the region, you can just keep it as-is. The Wayland protocol kind of encourages people to use buffer damage instead of surface damage (aka it would've been better if we only had buffer damage).

arandomhuman: > I was just thinking of exposing the client damage in surface->buffer_damage and adding a new function (wlr_surface_effective_damage?) that would compute the total damage.

I agree with this approach, but where would you call it? If it's during output damage calculation, it would presumably break sway. Maybe after calling `surface_apply_damage` in `surface_commit_pending`?

> The Wayland protocol kind of encourages people to use buffer damage instead of surface damage

I agree; I was just recommending surface damage for resize/move damages, not client damage.

(I'm sitting on IRC if you'd rather discuss there.)

emersion: >If it's during output damage calculation, it would presumably break sway.

We can break this API and update sway. I'd just let compositors call it if they need it.

arandomhuman: If that's the case, +1 for `wlr_surface_(get/calculate_?)effective_damage`. Presumably, one: it'll handle resize/move damage; two: handle buffer damage to surface damage transformation; and three: return in terms of surface coordinates?

emersion: I'd like to keep buffer coords. Compositors need to convert to output coords anyway, so doing buffer->surface in addition to surface->output is little work. Here's how it looks:

https://github.com/swaywm/wlroots/blob/936ec022664625f70fbcb22de6794d774cca159f/rootston/output.c#L708

arandomhuman: I'm not a very big fan of the idea of dealing with move/resize damage in terms of buffer coords. Two points I've already said before: one, they'd go surface->buffer->surface->output; two, semantically it's not buffer damage. Finally: if/when viewports are implemented, the conversions would get messy pretty quickly.

(Sorry, busy weekend)

emersion: Eh, I see. Yeah, that makes sense, even if it makes me a little bit sad inside.

So do we agree on:
* Keep state (`pending` and `current`) as is
* Make `buffer_damage` the sum of surface damage and buffer damage submitted by the client and clipped to the surface bounds (used when uploading textures and by compositors not using the helper function)
* Add a new `wlr_surface_get_effective_damage` helper function that handles move/resize and is in surface-local coords

emersion: We could also handle resize in `buffer_damage`, and handle move in `wlr_surface_get_effective_damage`. Not sure what's best. This could make sense since the whole texture is re-uploaded when resized.

arandomhuman: Ack on everything.

I think it might make sense to split resize damage: `surface_update_damage` damages the whole buffer (in buffer coords) if required (which preserves surface bounds, and integrates with viewports easily); while `wlr_surface_get_effective_damage` gives the surface damage to the compositor, including `previous` surface dimensions.

arandomhuman: Maybe #277 is relevant here: the function could take care of opaque regions implicitly?

emersion: >I think it might make sense to split resize damage

+1

>Maybe #277 is relevant here: the function could take care of opaque regions implicitly?

Hmm, #277 is a lot more complicated than this, and requires knowledge about other surfaces. We'd basically need to walk down the surfaces stack to collect the opaque region, and then walk up to render according to damage + opaque. It should be implemented in another standalone PR.

arandomhuman: Hm, alright. I'll add this function as soon as possible then. I'm assuming you're okay with all the current changes in the PR?

emersion: Yeah, I think so, they make sense. I'll do another final review to make sure I've not missed anything after you send a new version.

arandomhuman:
Right now, damage on resize and move for rotated clients on rootston is broken, especially after a 90* or 270* turn (unrelated to this PR). I'll try and take a look.

SirCmpwn:
We can deal with rotated clients separately imo.

rootston/output.c
```diff
@@ -696,14 +696,10 @@ static void damage_from_surface(struct wlr_surface *surface, int sx, int sy,
...
-
 	pixman_region32_t damage;
 	pixman_region32_init(&damage);
-	pixman_region32_copy(&damage, &surface->buffer_damage);
-	wlr_region_transform(&damage, &damage, transform,
-		surface->current.buffer_width, surface->current.buffer_height);
+	wlr_surface_get_effective_damage(surface, &damage);
+
 	wlr_region_scale(&damage, &damage,
 		wlr_output->scale / (float)surface->current.scale);
```
emersion: You need to remove `surface->current.scale` here.

Can you test with `scale=2` in your config file?

include/wlr/types/wlr_surface.h
```diff
@@ -229,4 +229,12 @@ struct wlr_surface *wlr_surface_from_resource(struct wl_resource *resource);
 void wlr_surface_for_each_surface(struct wlr_surface *surface,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 
+/**
+ * Get the effective damage to the surface in terms of surface local
```
emersion: Can you also update the `buffer_damage` docs?

SirCmpwn:
Thanks!

PR #1262, opened 2018-09-24
Improve session handling

nyorain:
Sessions can now be retrieved from a backend in a more general manner.
Multi-backend gets back its `session` field that contains the session
if one was created, removing the interfacing from multi backend with the
drm backend directly. This adds the possibility to use sessions even
without the drm backend.

It additionally fixes the bug that 2 session objects got created when
WLR_BACKENDS were set to "libinput,drm".

To allow vt switching without drm backend (and drm fd) on logind, start
listening to PropertiesChanged signals from dbus and parse the session
"Active" property when no master fd was created (this does not change
current drm backend behaviour in any way).

---

The main reason for this pr is to allow future session-based backends other than the drm one (like a VkDisplayKHR backend, i did a rather basic implementation [here](https://github.com/nyorain/wlroots/tree/vk_display) but it depends on the vulkan pr so it may take a while until it hits master (if it will at all...) and i figured it's a good idea submitting these rather independent changes/fixes already). The interface changes make session interaction in general simpler though, see the rootston/keyboard.c change for vt switching. Rootston no longer has to directly interface with the multi backend.

This is my first time really touching dbus/the sdbus library so a review of that part by someone with more experience is appreciated. Note that weston does handle it the same way; although it leaves the decision how/when to (de)activate sessions explicitly to the backend (it seems) and does not decide it implicitly by whether or not a drm device is taken.

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -35,12 +35,41 @@ struct logind_session {
 
 	char *id;
 	char *path;
+
+	// specifies whether a drm device was taken
+	// if so, the session will be (de)activated with the drm fd,
+	// otherwise with the dbus PropertiesChanged on "active" signal
+	bool has_drm;
 };
 
 static struct logind_session *logind_session_from_session(
 		struct wlr_session *base) {
 	assert(base->impl == &session_logind);
 	return (struct logind_session *)base;
+}
+
+static void parse_active(struct logind_session *session,
+		struct sd_bus_message *msg) {
+	int ret;
+	ret = sd_bus_message_enter_container(msg, 'v', "b");
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to parse D-Bus active property (1): %s",
+			strerror(-ret));
+		return;
+	}
+
+	bool active;
+	ret = sd_bus_message_read_basic(msg, 'b', &active);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to parse D-Bus active property (2): %s",
+			strerror(-ret));
+		return;
+	}
+
+	if (!session->has_drm) {
+		session->base.active = active;
+		wlr_signal_emit_safe(&session->base.session_signal, session);
+	}
 }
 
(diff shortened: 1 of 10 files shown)


backend/session/logind.c
```diff
@@ -307,6 +341,107 @@ static int resume_device(sd_bus_message *msg, void *userdata, sd_bus_error *ret_
...
+		goto error;
+	}
+
+	// PropertiesChanged 3: changed properties without values
+	sd_bus_message_enter_container(msg, 'a', "s");
+	while ((ret = sd_bus_message_read_basic(msg, 's', &s)) > 0) {
+		if (strcmp(s, "Active") == 0) {
+			sd_bus_error error = SD_BUS_ERROR_NULL;
+			sd_bus_message *answer = NULL;
+			ret = sd_bus_call_method(session->bus, "org.freedesktop.login1",
```
ascent12: `sd_bus_get_property`

backend/session/logind.c
```diff
@@ -307,6 +341,107 @@ static int resume_device(sd_bus_message *msg, void *userdata, sd_bus_error *ret_
...
+	// if we have drm fd we don't depend on this
+	if (session->has_drm) {
+		return 0;
+	}
+
+	// PropertiesChanged 1: interface
+	const char *interface;
+	ret = sd_bus_message_read_basic(msg, 's', &interface); // skip path
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to parse D-Bus PropertiesChanged (1) %s",
```
ascent12: I don't think these numbers are necessary. The line number is enough to distingush these calls.

backend/session/logind.c
nyorain: nice, didn't know that

backend/session/logind.c
nyorain: agreed

nyorain:
Currently on vacation, will come back to this next week.

nyorain:
Should be fixed now.

SirCmpwn:
Thanks!

PR #1265, opened 2018-09-26
Set b_lundef to false on FreeBSD in meson.build

sghctoma:
The Meson option `b_lundef` need to be set to `false` on FreeBSD, because the symbol `environ` is in crt1.o, which is not linked with shared libraries. Until now, it was done by passing the `-Db_lundef=false` flag manually, but with Meson >=0.48.0 it is possible to set this option conditionally inside meson.build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,8 @@ Run these commands:
     meson build
     ninja -C build
 
-On FreeBSD, you need to pass an extra flag to prevent a linking error:
-`meson build -D b_lundef=false`.
+If you use Meson older than 0.48.0 on FreeBSD, you need to pass an extra flag
+to prevent a linking error: `meson build -D b_lundef=false`.
 
 Install like so:
 
--- a/README.md
+++ b/README.md
@@ -74,9 +74,6 @@ Run these commands:
     meson build
     ninja -C build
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -74,8 +74,8 @@ Run these commands:
     meson build
     ninja -C build
 
-On FreeBSD, you need to pass an extra flag to prevent a linking error:
-`meson build -D b_lundef=false`.
+If you use Meson older than 0.48.0 on FreeBSD, you need to pass an extra flag
+to prevent a linking error: `meson build -D b_lundef=false`.
```
SirCmpwn: You should just set the minimum version and remove this from the README entirely. Specifying override_options on meson <0.48.0 will break anyway.

sghctoma: Done.

meson.build
```diff
@@ -111,6 +111,12 @@ wlr_deps += [
 	math,
 ]
 
+if host_machine.system().startswith('freebsd')
```
emersion: It doesn't matter a lot, but isn't it possible to check for equality here?

sghctoma: I copied this from sway's meson.build, and didn't consider if `startswith` really is necessary here. Checked it now, it works with `==`, commiited the change.

emersion:
Thanks!

PR #1269, opened 2018-09-28
xwayland/xwm: Stop including xcb_image.h

ivyl:
It's not used (XCB_IMAGE_FORMAT_Z_PIXMAP comes from xproto.h) and we
don't even have a pkg-config dependency on xcb-image, making the build
to fail on that inclusion on systems without the package.


emersion:
Thanks!

Issue #1172, opened 2018-08-02
Implement presentation-time


emersion:
```
23:27 <ascent12> presentation-time is one of those things I was going to get
around to doing in my "DRM 2.0" thing, but it would be easy enough for someone
to just do now if they wanted to. the "page_flip_callback" function in drm.c
basically contains all of the values you'd need.
```

ascent12:
>page_flip_callback

Correcting myself: `page_flip_handler`.

Also, when doing this (at least initially), it's probably only worth implementing/exposing this for the DRM backend. We could do it with the Wayland backend by doing this as a client and passing the values through, and I even think X11 had an extension related to this, but it's probably not worth the effort.

emersion:
Talked to some guys at XDC today, it seems that this extension suffers from a chicken-and-egg problem: no client implements it because no compositor implements it. Use cases include VR, games, medical apps, video players.

PR #1272, opened 2018-09-29
Implement presentation-time

emersion:
Test plan: `weston-presentation-shm` under DRM

Fixes https://github.com/swaywm/wlroots/issues/1172


types/meson.build
```diff
@@ -19,13 +19,17 @@ lib_wlr_types = static_library(
 		'xdg_shell_v6/wlr_xdg_shell_v6.c',
 		'xdg_shell_v6/wlr_xdg_surface_v6.c',
 		'xdg_shell_v6/wlr_xdg_toplevel_v6.c',
+		'tablet_v2/wlr_tablet_v2_pad.c',
```
Ongy: Is this alphabetical? X -> t -> w is weird, inside the subdirs it looks reasonable alphabetical.

types/wlr_output.c
```diff
@@ -560,6 +560,17 @@ void wlr_output_schedule_frame(struct wlr_output *output) {
 		wl_event_loop_add_idle(ev, schedule_frame_handle_idle_timer, output);
 }
 
+void wlr_output_send_present(struct wlr_output *output, struct timespec *when,
+		unsigned seq, uint32_t flags) {
+	struct wlr_output_event_present event = {
+		.output = output,
```
Ongy: What's the reason for adding the output pointer here?
I'd expect it to be derivable from the actual handler that's called.

emersion: wlroots has no idea on which output the surface is on. How could you "derive it from the handler"?

emersion: Oh, yeah, sorry, I misunderstood. I just added it for consistency with other events. It's not strictly necessary I guess.

types/wlr_output.c
```diff
@@ -18,6 +18,8 @@
 
 #define OUTPUT_VERSION 3
 
+#define DEFAULT_PRESENT_CLOCK CLOCK_MONOTONIC
```
SirCmpwn: -1

emersion: What do you mean?

SirCmpwn: This doesn't need to be abstracted, just use CLOCK_MONOTONIC explicitly.

emersion: There's more to come.

SirCmpwn: Okay, but I'm probably not going to change my mind.

emersion:
This is ready for review.

backend/drm/drm.c
```diff
@@ -1150,6 +1160,19 @@ static void page_flip_handler(int fd, unsigned seq,
 		post_drm_surface(&conn->crtc->primary->mgpu_surf);
 	}
 
+	struct timespec present_time = {
+		.tv_sec = tv_sec,
+		.tv_nsec = tv_usec * 1000,
+	};
+	struct wlr_output_event_present present_event = {
```
ammen99: shouldn't `output` also be set?

emersion: So this part is a little misleading. It's set in `wlr_output_send_present`. I just didn't want to create another struct. Maybe I can just set it here too, if that matters…

ammen99: oh, yeah, I've missed that. Idk if something should be changed then ...

emersion:
Rebased against master.

SirCmpwn:
Thanks!

PR #1274, opened 2018-10-01
[WIP] Implement relative-pointer

arandomhuman:
This is mostly a feature complete implementation of the relative-pointer protocol, but since this is my first try, I've marked it WIP for now. Suggestions on how to correct/improve code, structure, whatever, are welcome.

What works:
 - weston-resizor (mostly) works. Some eccentricity is because of issue #613; specifically (lack of) [this behavior](https://github.com/swaywm/wlroots/issues/613#issuecomment-363923280). This makes the cursor release the lock as soon as the surface is sufficiently small (i.e., the pointer location is out of bounds). The x11 backend is generating absolute events for pointer motion for me (why?), so the scaling is out of whack.
 - [Neverball](https://neverball.org/) on Xwayland

PR Todo:
 - Add an example/append to the pointer-constraints example

Otherwise:
 - Maintain event timestamps in usecs instead of converting to msecs?


include/wlr/types/wlr_seat.h
```diff
@@ -304,6 +305,18 @@ void wlr_seat_pointer_clear_focus(struct wlr_seat *wlr_seat);
 void wlr_seat_pointer_send_motion(struct wlr_seat *wlr_seat, uint32_t time,
 		double sx, double sy);
 
+/** Send relative motion events to the surface with pointer focus. Coordinates
```
SirCmpwn: nit: for this and the other doc comments, move the text to the next line

```c
/**
  * Send relative motion...
  * ...32 bits.
  */
```

rootston/cursor.c
```diff
@@ -310,6 +310,9 @@ void roots_cursor_handle_motion(struct roots_cursor *cursor,
 	double dx = event->delta_x;
 	double dy = event->delta_y;
 
+	wlr_seat_pointer_notify_relative_motion(cursor->seat->seat, (uint64_t)
```
SirCmpwn: nit: put the cast next to the statement it's casting, don't split the line here

SirCmpwn: I wonder if these calls could live in wlr_cursor?

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,202 @@
...
+
+static const struct zwp_relative_pointer_manager_v1_interface relative_pointer_manager_v1_impl;
+static const struct zwp_relative_pointer_v1_interface relative_pointer_v1_impl;
+
+
+/* Callback functions
+ */
+
+static void relative_pointer_manager_v1_handle_resource_destroy(struct wl_resource *resource)
+{
```
SirCmpwn: Opening braces go on the same line

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,202 @@
...
+		relative_pointer, relative_pointer_v1_handle_resource_destroy);
+
+	struct wlr_relative_pointer_manager_v1 *relative_pointer_manager =
+		wl_resource_get_user_data(resource);
+	wlr_signal_emit_safe(&relative_pointer_manager->events.new_relative_pointer,
+		relative_pointer);
+
+	wlr_log(WLR_DEBUG,
+		"relative_pointer_v1 0x%" PRIXPTR " created for client 0x%" PRIXPTR,
+		(uintptr_t) relative_pointer, (uintptr_t) client);
```
SirCmpwn: Just use `%p`

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,202 @@
...
+
+
+void wlr_relative_pointer_v1_send_relative_motion(struct wl_resource *resource,
+	uint64_t time, double dx, double dy, double dx_unaccel, double
+	dy_unaccel)
+{
+	assert(wl_resource_instance_of(resource, &zwp_relative_pointer_v1_interface,
+		&relative_pointer_v1_impl));
+	zwp_relative_pointer_v1_send_relative_motion(resource,
+		(uint32_t) (time >> 32), (uint32_t) time, wl_fixed_from_double(dx),
```
SirCmpwn: nit: remove the extra space between cast and statement

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,202 @@
...
+	wl_global_destroy(relative_pointer_manager->global);
+	free(relative_pointer_manager);
+
+	wlr_log(WLR_DEBUG, "relative_pointer_v1 manager destroyed");
+}
+
+
+void wlr_relative_pointer_v1_send_relative_motion(struct wl_resource *resource,
+	uint64_t time, double dx, double dy, double dx_unaccel, double
+	dy_unaccel)
```
SirCmpwn: Don't split lines along a parameter definition

SirCmpwn:
Overall this looks pretty good

rootston/cursor.c
arandomhuman: Can't opine on that yet; not too familiar with the pipeline. It has to be done after pointer focus is calculated, at least.

rootston/cursor.c
```diff
@@ -310,8 +310,8 @@ void roots_cursor_handle_motion(struct roots_cursor *cursor,
 	double dx = event->delta_x;
 	double dy = event->delta_y;
 
-	wlr_seat_pointer_notify_relative_motion(cursor->seat->seat, (uint64_t)
-		event->time_msec, dx, dy, dx, dy);
+	wlr_seat_pointer_notify_relative_motion(cursor->seat->seat,
+		(uint64_t) event->time_msec, dx, dy, dx, dy);
```
SirCmpwn: Extra space here

types/wlr_relative_pointer_v1.c
```diff
@@ -190,13 +183,12 @@ void wlr_relative_pointer_v1_destroy(struct wlr_relative_pointer_manager_v1 *rel
...
-{
+		uint64_t time, double dx, double dy,
+		double dx_unaccel, double dy_unaccel) {
 	assert(wl_resource_instance_of(resource, &zwp_relative_pointer_v1_interface,
 		&relative_pointer_v1_impl));
 	zwp_relative_pointer_v1_send_relative_motion(resource,
-		(uint32_t) (time >> 32), (uint32_t) time, wl_fixed_from_double(dx),
-		wl_fixed_from_double(dy), wl_fixed_from_double(dx_unaccel),
-		wl_fixed_from_double(dy_unaccel));
+		(uint32_t) (time >> 32), (uint32_t) time,
```
SirCmpwn: still the extra spaces here

rootston/cursor.c
arandomhuman: I kinda find them dirty, but eh, your call. Formatting's all done. Let me add an example before merging.

emersion: There's no need to add an example if there's already a weston example that does the same thing.

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,75 @@
...
+	struct wl_pointer *pointer;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_relative_pointer_manager_v1 *wlr_relative_pointer_v1_create(struct wl_display *display);
```
emersion: This should be named `wlr_relative_pointer_manager_v1_create`.

Ditto for `destroy`

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,75 @@
...
+ * pointer.
+ *
+ * Signals:
+ * - destroy -> struct wlr_relative_pointer_manager_v1 *manager
+ *   :: the manager was destroyed
+ * - new_relative_pointer -> struct wlr_relative_pointer_v1 *relative_pointer
+ *   :: a new relative_pointer was created
+ */
+
+struct wlr_relative_pointer_manager_v1 {
```
emersion: This should listen to the `wl_display` destroy signal.

rootston/cursor.c
```diff
@@ -349,6 +352,11 @@ void roots_cursor_handle_motion_absolute(struct roots_cursor *cursor,
 	wlr_cursor_absolute_to_layout_coords(cursor->cursor, event->device, event->x,
 		event->y, &lx, &ly);
 
+	double dx = lx - cursor->cursor->x;
+	double dy = ly - cursor->cursor->y;
+	wlr_seat_pointer_notify_relative_motion(cursor->seat->seat, (uint64_t)
+		event->time_msec, dx, dy, dx, dy);
```
emersion: Can you put `(uint64_t)` and `event->time_msec` on the same line?

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Signals:
+ * - destroy -> struct wlr_relative_pointer_v1 *relative_pointer
+ *   :: the relative_pointer was destroyed
+ */
+
+struct wlr_relative_pointer_v1 {
+	struct wl_client *client;
+	struct wl_resource *resource;
+	struct wl_pointer *pointer;
```
emersion: `wl_pointer` is a client-side struct, it shouldn't be used here?

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * A wp_relative_pointer object is an extension to the wl_pointer interface
+ * used for emitting relative pointer events. It shares the same focus as
+ * wl_pointer objects of the same seat and will only emit events when it has
+ * focus.
+ *
+ * Signals:
+ * - destroy -> struct wlr_relative_pointer_v1 *relative_pointer
+ *   :: the relative_pointer was destroyed
```
emersion: This can be documented near the signal itself.

types/seat/wlr_seat_pointer.c
```diff
@@ -304,6 +305,28 @@ void wlr_seat_pointer_notify_motion(struct wlr_seat *wlr_seat, uint32_t time,
...
+		double dx_unaccel, double dy_unaccel) {
+	struct wlr_seat_client *client = wlr_seat->pointer_state.focused_client;
+	if (client == NULL) {
+		return;
+	}
+
+	struct wl_resource *resource;
+	wl_resource_for_each(resource, &client->relative_pointers) {
+		struct wlr_relative_pointer_v1 *relative_pointer =
+			wl_resource_get_user_data(resource);
```
emersion: There should be a safe helper function that asserts `wl_resource_is_instance_of` instead of directly using `wl_resource_get_user_data`

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,192 @@
...
+	wl_signal_init(&relative_pointer->events.destroy);
+
+	wl_list_insert(&seat_client->relative_pointers,
+		wl_resource_get_link(relative_pointer_resource));
+
+	wl_resource_set_implementation(relative_pointer_resource, &relative_pointer_v1_impl,
+		relative_pointer, relative_pointer_v1_handle_resource_destroy);
+
+	struct wlr_relative_pointer_manager_v1 *relative_pointer_manager =
+		wl_resource_get_user_data(resource);
```
emersion: There should be a safe helper instead of directly using `wl_resource_get_user_data` (this one can be local to this file)

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,192 @@
...
+		return NULL;
+	}
+
+	wl_list_init(&relative_pointer_manager->resources);
+
+	wl_signal_init(&relative_pointer_manager->events.destroy);
+	wl_signal_init(&relative_pointer_manager->events.new_relative_pointer);
+
+	relative_pointer_manager->global = wl_global_create(display,
+		&zwp_relative_pointer_manager_v1_interface, 1,
```
emersion: Is it possible to define the version with `RELATIVE_POINTER_MANAGER_VERSION` at the beginning of the file? It makes it way easier to locate it.

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,192 @@
...
+		return NULL;
+	}
+
+	wlr_log(WLR_DEBUG, "relative_pointer_v1 manager created");
+
+	return relative_pointer_manager;
+}
+
+
+void wlr_relative_pointer_v1_destroy(struct wlr_relative_pointer_manager_v1 *relative_pointer_manager) {
```
emersion: Style: 80 char limit

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,192 @@
...
+		&zwp_relative_pointer_v1_interface, wl_resource_get_version(resource), id);
+	if (relative_pointer_resource == NULL) {
+		free(relative_pointer);
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	relative_pointer->client = client;
+	relative_pointer->resource = relative_pointer_resource;
+	relative_pointer->pointer = (struct wl_pointer *) pointer;
```
emersion: Style: extra space

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,192 @@
...
+static void relative_pointer_manager_v1_handle_resource_destroy(struct wl_resource *resource) {
+	wl_list_remove(wl_resource_get_link(resource));
+}
+
+
+static void relative_pointer_v1_handle_resource_destroy(struct wl_resource *resource) {
+	wl_list_remove(wl_resource_get_link(resource));
+
+	struct wlr_relative_pointer_v1 *relative_pointer = wl_resource_get_user_data(resource);
+	free(relative_pointer);
```
emersion: The `destroy` event should probably be emitted here

include/wlr/types/wlr_relative_pointer_v1.h
arandomhuman: What should be the correct behavior [here](https://github.com/swaywm/wlroots/pull/1274/files#diff-4aa3486289ec9789afb8643384856f95R326)? To send the frame event to every pointer resource?

I thought the `wl_pointer` needed to be stored to compare against events/pointer resources, but if I'm not wrong, every seat has a unique pointer, right? So, `wlr_seat_client_from_pointer_resource` should be enough?

include/wlr/types/wlr_relative_pointer_v1.h
arandomhuman: I find it cleaner separately. Do you prefer it alongside the signal?

include/wlr/types/wlr_relative_pointer_v1.h
emersion: `wl_pointer` is a client struct that wraps `wl_proxy`. On the server-side we're using `wl_resource`, we shouldn't cast to `wl_pointer`.

This line of code should be `struct wl_resource *pointer`. An even better solution would be to store a pointer to the seat client (`struct wlr_seat_client *seat_client`).

include/wlr/types/wlr_relative_pointer_v1.h
emersion: Yeah, that's what we do elsewhere.

types/wlr_relative_pointer_v1.c
arandomhuman: Should it be undef'ed at the end of the file?

emersion: Since this is a `.c` file, it doesn't need to, it won't be included in another file anyway.

include/wlr/types/wlr_relative_pointer_v1.h
arandomhuman: I see. Could you check and tell me what the correct behavior should be when sending frame events in the file I linked above? If sending it to every pointer resource of the same seat-client suffices, there's no need to store anything.

emersion: Ah, right, each seat client can have multiple pointer resources, so storing `wlr_seat_client` won't do, we need to store the `wl_resource` and send the frame event only to this one.

arandomhuman: Done. Exposed `wlr_relative_pointer_v1_from_resource` as well, for the frame event handling.

arandomhuman:
@emersion, I think I've finished with most of your comments. Will push changes plus an example in a bit.
Re: resources in the seat, would you rather it be named relative_pointers_v1?

types/wlr_relative_pointer_v1.c
emersion: I think @SirCmpwn wants to have a hard 80-char limit (please confirm). It looks like this:

```c
void wlr_relative_pointer_v1_destroy(
		struct wlr_relative_pointer_manager_v1 *relative_pointer_manager) {
	…
}
```

emersion:
>Will push changes plus an example in a bit.

Only push an example if there's no matching Weston example.

>Re: resources in the seat, would you rather it be named relative_pointers_v1?

Yeah, that seems like a good idea.

rootston/cursor.c
arandomhuman: I checked; the analogous `wlr_seat_pointer_notify_motion` is called only in rootston. Same treatment for this one.

arandomhuman:
I've pushed an example nevertheless because the weston example isn't working satisfactorily (see top post). I'll remove it if required. Column limits are left; I'll get to that then squash my commits.

emersion:
This PR needs a rebase.

SirCmpwn:
Would you consider sending a patch to fix the weston example?

clarcharr:
Any updates on this? Would love to see this merged so the corresponding functionality in sway can be added.

ddevault:
Anyone is welcome to adopt this patch to put the finishing touches on.

ForTheReallys:
I can pick this up.  Do I just create a separate pull request?

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/1432

PR #1275, opened 2018-10-01
Fix a crash on zxdg_decoration_manager_v1 destroy

linkmauve:
When this request was called from a client, the unimplemented callback
was NULL and thus was crashing rootston.


SirCmpwn:
Thanks!

Issue #995, opened 2018-05-24
Check DRM link-status property


ascent12:
There is some user-space stuff we're supposed to be doing to make sure black screens don't happen in some situations. I don't know a huge deal about this or even how to test it, though.
I don't think any other wayland compositor handles this, but Xorg does.

Some information is here.
https://fosdem.org/2018/schedule/event/kms_dp/attachments/slides/2703/export/events/attachments/kms_dp/slides/2703/Fosdem2018__final.pdf

emersion:
Some more information: https://patchwork.kernel.org/patch/9428849/

emersion:
Weston issue, with what needs to be done: https://gitlab.freedesktop.org/wayland/weston/issues/124#note_30387

We don't have a signal to update the list of available modes yet.

emersion:
We need to: on hotplug event, check for link-status, if it's bad, force a new modeset with the highest mode available. The kernel will prune bad modes itself.

We need to make sure we update our list of modes in this case.

PR #1280, opened 2018-10-04
backend/drm: add support for the link-status property

emersion:
Fixes https://github.com/swaywm/wlroots/issues/995

### DIFF
--- a/include/backend/drm/properties.h
+++ b/include/backend/drm/properties.h
@@ -14,12 +14,13 @@ union wlr_drm_connector_props {
 	struct {
 		uint32_t edid;
 		uint32_t dpms;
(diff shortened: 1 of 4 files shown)


include/backend/drm/properties.h
```diff
@@ -18,8 +18,9 @@ union wlr_drm_connector_props {
 		// atomic-modesetting only
 
 		uint32_t crtc_id;
+		uint32_t link_status;
```
ascent12: Thus is not an atomic-only property, so it should be placed above the above comment.

emersion: Right. Should be fixed.

emersion:
@ascent12 bump

PR #1281, opened 2018-10-04
Update sizes for new EDID database.

aereaux:
This fixes an error where in certain compiler configurations (Arch Linux?) compilation will fail because the sizes were not large enough.


emersion:
Thanks!

PR #1282, opened 2018-10-05
Prevent excessive frame events when compositor doesn't swap buffers

RyanDwyer:
When we send an output frame event, we should not assume that the compositor is going to call `wlr_output_swap_buffers` in response to it. If the compositor does not swap the buffers, the idle event source still exists and is executed every time the Wayland event loop becomes idle, which means we send frame events repeatedly until the buffers are swapped.

This moves the removal of the idle event source out of `wlr_output_swap_buffers` and into `wlr_output_send_frame`, where it is guaranteed to be removed.

Fixes https://github.com/swaywm/sway/issues/2748

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -472,10 +472,6 @@ bool wlr_output_swap_buffers(struct wlr_output *output, struct timespec *when,
 		wlr_log(WLR_ERROR, "Tried to swap buffers when a frame is pending");
 		return false;
 	}
-	if (output->idle_frame != NULL) {
-		wl_event_source_remove(output->idle_frame);
-		output->idle_frame = NULL;
-	}
 
 	int width, height;
 	wlr_output_transformed_resolution(output, &width, &height);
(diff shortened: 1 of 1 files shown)


RyanDwyer:
Can you try with this? It just removes the event source to before emitting the frame event rather than afterwards.

    diff --git a/types/wlr_output.c b/types/wlr_output.c
    index c407ca0b..b2ba35c6 100644
    --- a/types/wlr_output.c
    +++ b/types/wlr_output.c
    @@ -533,14 +533,14 @@ bool wlr_output_swap_buffers(struct wlr_output *output, struct timespec *when,
     }
     
     void wlr_output_send_frame(struct wlr_output *output) {
    -	output->frame_pending = false;
    -	wlr_signal_emit_safe(&output->events.frame, output);
    -	output->frame_pending = true;
    -
     	if (output->idle_frame != NULL) {
     		wl_event_source_remove(output->idle_frame);
     		output->idle_frame = NULL;
     	}
    +
    +	output->frame_pending = false;
    +	wlr_signal_emit_safe(&output->events.frame, output);
    +	output->frame_pending = true;
     }
     
     static void schedule_frame_handle_idle_timer(void *data) {

ianyfan:
Doesn't work

emersion:
Ehhh? I really don't understand what this PR is doing. `send_frame` is called by backends after a frame has been scheduled. This completely breaks how outputs work.

PR #1286, opened 2018-10-07
backend/drm: better handle GPUs without CRTCs

emersion:
Also make sure we don't have zero-length VLAs as this is undefined behaviour.


SirCmpwn:
Thanks!

PR #1287, opened 2018-10-07
data-device: remove wlr_data_source.seat_client

emersion:
Since the source doesn't always come from a client, this field
doesn't make sense. It is replaced by a new "finalized" field in
wlr_client_data_source. This is used to make sure set_actions is
not sent after start_drag has been sent.

A check in data_offer_choose_action has been removed: if an offer
has been sent then start_drag has been called, no need to check.

I also wanted to add a check for wl_data_source.offer, but it
turns out (1) this isn't in the spec (2) it breaks GTK+.

This is some preliminary work for Firefox on Wayland compatibility.


emersion:
Ping?

SirCmpwn:
Thanks!

PR #1289, opened 2018-10-07
Fix cursor visibility on surface commit

ixru:
When spawning a new window in Rootston the cursor will disappear if it is over the surface. It is supposed to be set to a surface specific cursor. I think this is caused by the client sending an empty cursor surface when requested causing cursor->visible to be set to false. The surface is then committed but the visibility is not updated so it stays hidden until it registers movement. 

I'm unsure if I put it in the correct place though. Maybe it should be called inside output_cursor_commit since it updates the dimenstions, I don't know what consequences that will have though.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -874,6 +874,7 @@ static void output_cursor_handle_commit(struct wl_listener *listener,
 	struct wlr_output_cursor *cursor =
 		wl_container_of(listener, cursor, surface_commit);
 	output_cursor_commit(cursor, true);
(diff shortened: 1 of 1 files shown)


emersion:
>Maybe  it should be called inside output_cursor_commit since it updates the  dimenstions, I don't know what consequences that will have though.

That seems like a good idea. Can you try doing this?

ixru:
This emits request_set_cursor twice on every surface entry so it's not perfect.

SirCmpwn:
Thanks!

PR #1292, opened 2018-10-08
meson.build: require logind v237

n3rdopolis:
Commit 7b523884249f904f45b1e9a7eae8152c2ac1248d uses a function added in logind v237


emersion:
Thanks!

PR #1294, opened 2018-10-08
rootston: export XCURSOR_SIZE and XCURSOR_THEME

emersion:
These can be used by toolkits (currently Qt) to choose a default
cursor theme and size. Note that this isn't a perfect solution:

- Per-seat configuration isn't possible
- It's not possible to set the default image
- Live config reload isn't possible

But it's easy to implement and simple. To fix these remaining
issues a separate protocol would be needed.


SirCmpwn:
Thanks!

PR #1295, opened 2018-10-08
Reuse `wlr_buffer` objects where possible

nyorain:
Hardware accelerated (ie. only imported) wlr_buffers were previously
destroyed and recreated (ie. reimported) on every usage.
Keep them around even when no one uses them anymore and reuse them (ie.
the ref count goes to 0).

---

As mentioned on IRC, the motivation is rather simple: keeping wl_drm/dmabuf texture objects around shouldn't cost much (the buffers are allocated on gpu anyways) but importing them might have some performance cost. On the egl code path it actually looks small enough (egl image import, texture creation) but even that might result in some work for the driver. On vulkan (more explicit than egl/gl) there is quite something to do (see a first vulkan [texture_from_dmabuf](https://github.com/nyorain/wlroots/blob/vk_ext/render/vulkan/texture.c#L347) implementation that is still missing something (and depending on a currently disabled extension)). Anyways, probably not something we want to do _every frame_ (for hardware accelerated clients) if we don't have to.

Via the `wlr_buffer.keep` flag we currently disable this keeping-alive mechanism for shm bufs: keeping them around even if not read anymore will allocate additional gpu memory and we usually don't have to create a new texture every frame anyways due to the previous buffer/texture re-using mechanism for shm buffers.

### DIFF
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -14,6 +14,7 @@
 
 /**
  * A client buffer.
+ * The buffers contents are reference counted.
  */
 struct wlr_buffer {
 	/**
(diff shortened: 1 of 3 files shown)


nyorain:
Actually nevermind. Guess this was intentionally written this way, as the whole implicit synchronization thing of egl/gl **requires** us to destroy the egl image everytime we release the buffer (although it still doesn't seem to be documented anywhere...). See https://bugs.freedesktop.org/show_bug.cgi?id=97353 (probably didn't find an issue while testing since it's only an issue is some special situations as most synchronization stuff).
Once the explicit synchronization protocol is in use feel free to come back to this.

Issue #1297, opened 2018-10-09
Assertion failure in get_drm_backend_from_backend in page_flip_handler


Emantor:
Backtrace:
```
(gdb) bt f
#0  0x00007f1228ff1d7f in raise () at /usr/lib/libc.so.6
#1  0x00007f1228fdc672 in abort () at /usr/lib/libc.so.6
#2  0x00007f1228fdc548 in _nl_load_domain.cold.0 () at /usr/lib/libc.so.6
#3  0x00007f1228fea396 in  () at /usr/lib/libc.so.6
#4  0x00007f1229e3e15e in get_drm_backend_from_backend (wlr_backend=0x7f1229178db0 <main_arena+784>) at ../subprojects/wlroots/backend/drm/backend.c:20
        __PRETTY_FUNCTION__ = "get_drm_backend_from_backend"
#5  0x00007f1229e431ce in page_flip_handler (fd=9, seq=2242, tv_sec=93, tv_usec=945761, data=0x55baf46a6650) at ../subprojects/wlroots/backend/drm/drm.c:1226
        conn = 0x55baf46a6650
        drm = 0x55baf4642430
        present_time = {tv_sec = 0, tv_nsec = 8361308013513814784}
        present_event = {output = 0x7ffe9a633c10, when = 0x7f12290a6749 <read+89>, seq = 2590193888, refresh = 32766, flags = 32}
#6  0x00007f1228cd9605 in drmHandleEvent () at /usr/lib/libdrm.so.2
#7  0x00007f1229e4337d in handle_drm_event (fd=9, mask=1, data=0x0) at ../subprojects/wlroots/backend/drm/drm.c:1268
        event = {version = 4, vblank_handler = 0x0, page_flip_handler = 0x7f1229e4318f <page_flip_handler>, page_flip_handler2 = 0x0, sequence_handler = 0x0}
#8  0x00007f1229403702 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#9  0x00007f12294022ac in wl_display_run () at /usr/lib/libwayland-server.so.0
#10 0x000055baf3d276bd in server_run (server=0x55baf3d7eb40 <server>) at ../sway/server.c:174
#11 0x000055baf3d26dd2 in main (argc=2, argv=0x7ffe9a634108) at ../sway/main.c:384
        verbose = 0
        debug = 1
        validate = 0
        long_options =
[...]

PR #1298, opened 2018-10-09
output: remove idle_frame event source when destroying output

emersion:
This prevents the idle event to be activated on a destroyed
output.

This also makes the backend responsible for free-ing modes, as it
is the one allocating them and adding them to the list. Note that
the DRM backend (the only one using modes) already frees them.

Fixes https://github.com/swaywm/wlroots/issues/1296


Emantor:
Tested and presumably fixes #1296, now I only see #1297 as the crash reason.

SirCmpwn:
Thanks!

PR #1299, opened 2018-10-09
backend/drm: set drmEventContext version to 2

Emantor:
As per [1] set drmEventContext version to 2, since wlroots does not use the
page_flip_handler2.

[1]: https://s-opensource.org/2017/04/12/libdrm-event-handling-youre-probably-wrong/


SirCmpwn:
Thanks!

PR #1303, opened 2018-10-12
text-input: fix releasing destroy handlers

dcz-purism:
It was possible for listeners on text-input objects to be still attached to signals even after the objects were gone, causing crashes depending on the order objects were destroyed. Not after this patch.

### DIFF
--- a/types/wlr_text_input_v3.c
+++ b/types/wlr_text_input_v3.c
@@ -9,6 +9,12 @@
 #include "text-input-unstable-v3-protocol.h"
 #include "util/signal.h"
 
+static void text_input_clear_focused_surface(struct wlr_text_input_v3 *text_input) {
+	wl_list_remove(&text_input->surface_destroy.link);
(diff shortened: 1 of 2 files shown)


types/wlr_text_input_v3.c
```diff
@@ -61,6 +61,9 @@ void wlr_text_input_v3_send_done(struct wlr_text_input_v3 *text_input) {
 
 static void wlr_text_input_destroy(struct wlr_text_input_v3 *text_input) {
 	wlr_signal_emit_safe(&text_input->events.destroy, text_input);
+	if (text_input->focused_surface) {
```
emersion: Instead of adding conditions like this, you can always call `wl_list_remove`, and call `wl_list_init` when the surface is removed.

dcz-purism:
I found another place where the handler needs to be released in text-input, and I added `list_init` to all the relevant places. With inits, the patch touches more places, but conditions are gone.

emersion:
Thanks!

PR #1304, opened 2018-10-12
Support older wlr_linux_dmabuf_v1 clients

nyorain:
If a client uses an older version of the dmabuf protocol, use the
`formats` event instead of `modifiers` (since that didn't exist in older
versions).
With a bit of necessary guessing, support dmabuf importing even when
EGL_EXT_image_dma_buf_import_modifiers isn't present instead of
failing up front.

---

Makes weston-simple-dmabuf-v4l work (at least with XR24 formats since the gl renderer does not support yuv). See the help manual of that program for information (loading the dummy vivid kernel module). Mainly useful to test dmabuf support (e.g. in renderers, think you can guess what i needed this for).

Note that the way we support older versions of the protocol is implemented different from weston: in weston, the modifier the client passes the compositor are simply ignored (see [here](https://gitlab.freedesktop.org/wayland/weston/blob/master/libweston/linux-dmabuf.c#L115)) which may (theoretically?) cause problems later on i guess (reading garbage). We never ignore the modifier, but assume that without the modifiers egl extension it expects linear data (DRM_FORMAT_MOD_LINEAR).
Otherwise we implement roughly the same guessing that weston does when the egl modifiers extension is not present (note that this is still the case for amd drivers. Not sure if it will soon change since amd doesn't really seem to use public modifiers so we should really support this case as good as we can).

I guess @agx should have a look at this since it seems like he implemented the protocol in the first place.

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -3,6 +3,7 @@
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
 #include <stdlib.h>
+#include <drm_fourcc.h>
 #include <wlr/render/egl.h>
 #include <wlr/util/log.h>
 #include "glapi.h"
@@ -382,13 +383,21 @@ EGLImageKHR wlr_egl_create_image_from_wl_drm(struct wlr_egl *egl,
 
 EGLImageKHR wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
 		struct wlr_dmabuf_attributes *attributes) {
-	if (!egl->exts.image_base_khr) {
+	if (!egl->exts.image_base_khr || !egl->exts.image_dmabuf_import_ext) {
+		wlr_log(WLR_ERROR, "dmabuf import extension not present");
 		return NULL;
 	}
 
 	bool has_modifier = false;
-	if (attributes->modifier != DRM_FORMAT_MOD_INVALID) {
+
+	// we assume the same way we assumed formats without the import_modifiers
+	// extension that mod_linear is supported. The special mod mod_invalid
+	// is sometimes used to signal modifier unawareness which is what we
+	// have here
+	if (attributes->modifier != DRM_FORMAT_MOD_INVALID &&
+			attributes->modifier != DRM_FORMAT_MOD_LINEAR) {
 		if (!egl->exts.image_dmabuf_import_modifiers_ext) {
+			wlr_log(WLR_ERROR, "dmabuf modifiers extension not present");
 			return NULL;
 		}
 		has_modifier = true;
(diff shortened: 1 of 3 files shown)


render/egl.c
```diff
@@ -460,12 +469,33 @@ EGLImageKHR wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
+	// which formats are supported. These two are on almost always
+	// supported; it's the intended way to just try to create buffers.
+	// Just a guess but better than not supporting dmabufs at all,
+	// given that the modifiers extension isn't supported everywhere.
+	if (!egl->exts.image_dmabuf_import_modifiers_ext) {
+		static const int fallback_formats[] = {
+			DRM_FORMAT_ARGB8888,
+			DRM_FORMAT_XRGB8888,
+		};
+		static unsigned num = 2;
```
emersion: This can be `sizeof(fallback_formats) / sizeof(fallback_formats[0])`

render/egl.c
```diff
@@ -460,12 +469,33 @@ EGLImageKHR wlr_egl_create_image_from_dmabuf(struct wlr_egl *egl,
...
+
+		*formats = calloc(num, sizeof(int));
+		if (!*formats) {
+			wlr_log_errno(WLR_ERROR, "Allocation failed");
+			return -1;
+		}
+
+		memcpy(*formats, fallback_formats, num * sizeof(**formats));
+		return 2;
```
emersion: `return num`

render/egl.c
```diff
@@ -488,12 +518,16 @@ int wlr_egl_get_dmabuf_formats(struct wlr_egl *egl,
...
-	if (!egl->exts.image_dmabuf_import_ext ||
-			!egl->exts.image_dmabuf_import_modifiers_ext) {
+	if (!egl->exts.image_dmabuf_import_ext) {
 		wlr_log(WLR_DEBUG, "dmabuf extension not present");
 		return -1;
 	}
 
+	if(!egl->exts.image_dmabuf_import_modifiers_ext) {
+		// we don't know any supported modifiers
+		return 0;
```
emersion: Just to prevent mistakes, can we set `*modifiers` to NULL here?

types/wlr_linux_dmabuf_v1.c
```diff
@@ -410,10 +407,18 @@ static void linux_dmabuf_send_modifiers(struct wlr_linux_dmabuf_v1 *linux_dmabuf
...
 		for (int j = 0; j < num_modifiers; j++) {
-			uint32_t modifier_lo = modifiers[j] & 0xFFFFFFFF;
-			uint32_t modifier_hi = modifiers[j] >> 32;
-			zwp_linux_dmabuf_v1_send_modifier(resource, formats[i],
-				modifier_hi, modifier_lo);
+			if (version >= ZWP_LINUX_DMABUF_V1_MODIFIER_SINCE_VERSION) {
+				uint32_t modifier_lo = modifiers[j] & 0xFFFFFFFF;
+				uint32_t modifier_hi = modifiers[j] >> 32;
+				zwp_linux_dmabuf_v1_send_modifier(resource,
+								  formats[i],
```
emersion: Style: don't align those, indenting them is enough

types/wlr_linux_dmabuf_v1.c
```diff
@@ -410,10 +407,18 @@ static void linux_dmabuf_send_modifiers(struct wlr_linux_dmabuf_v1 *linux_dmabuf
...
+								  modifier_hi,
+								  modifier_lo);
+			} else if (modifiers[j] == DRM_FORMAT_MOD_LINEAR ||
+				   modifiers == &modifier_invalid) {
+				zwp_linux_dmabuf_v1_send_format(resource,
+								formats[i]);
```
emersion: Ditto

nyorain:
Thanks, should all be fixed now

emersion:
I tried to test but failed to run `weston-simple-dmabuf-v4l`. How did you test? (I think I'm missing a `modprobe` or something)

It fails to create XR24 buffers.

nyorain:
Unless you have a webcam or something (even if you do it probably won't work with gl since most can only use some kind of yuv buffers), you have to use modprobe the `vivid` kernel module which adds a dummy video output that can use XR24 buffers (you should see colored blocks with some information).

nyorain:
btw this does also work on master on platforms that implement the EGL_EXT_image_dma_buf_import_modifiers extension (but only since the weston client [does not respect the advertised formats](https://gitlab.freedesktop.org/wayland/weston/blob/master/clients/simple-dmabuf-v4l.c#L860), otherwise it would need this patch. Sadly i know no correct older client to show this). But with this it should also work on platforms without the egl extension (everything amd at the moment).

SirCmpwn:
Thanks!

PR #1307, opened 2018-10-13
Fix a few bugs found by the static analyzer

emersion:
See https://builds.sr.ht/~sircmpwn/job/8500#task-clang

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -217,16 +217,16 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display,
 				backend, &multi->session, name, create_renderer_func);
 			if (subbackend == NULL) {
 				wlr_log(WLR_ERROR, "failed to start backend '%s'", name);
-				wlr_backend_destroy(backend);
 				wlr_session_destroy(multi->session);
+				wlr_backend_destroy(backend);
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Why are static analysis failures non-fatal?

backend/backend.c
```diff
@@ -217,16 +217,16 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display,
 				backend, &multi->session, name, create_renderer_func);
 			if (subbackend == NULL) {
 				wlr_log(WLR_ERROR, "failed to start backend '%s'", name);
-				wlr_backend_destroy(backend);
 				wlr_session_destroy(multi->session);
+				wlr_backend_destroy(backend);
```
nyorain: I don't think that will work as backends (see drm) might expect the session they were created with to be valid on destruction. Probably best to use a separate `session` variable instead of `multi->session` so we can keep the destruction order as it was (Was probably my oversight in the first place, sry). Other places as well.

agx:
@SirCmpwn I guess due to https://github.com/mesonbuild/meson/issues/4334

SirCmpwn:
Thanks!

PR #1308, opened 2018-10-13
text-input: Accept disable requests

dcz-purism:
The disable signal on text-input objects must not be ignored. It is only sent when the previous state was enabled.

This led to input method not being disabled when the application requested it.


SirCmpwn:
Thanks!

PR #1309, opened 2018-10-14
Fix headless backend

SirCmpwn:
It was only working if you ran it underneath an already-working Wayland
compositor. Running with the headless backend on the API would break.


emersion:
Thanks!

PR #1310, opened 2018-10-14
Fix include path for (libdrm/)drm_fourcc.h

colemickens:
This PR broke a private nixpkgs definition I have for wlroots: https://github.com/swaywm/wlroots/pull/1304

It is fixed by changing `#include <drm_fourcc.h>` to `#include <libdrm/drm_fourcc.h>`, which follows what is already done in the dmabuf example.

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -3,7 +3,7 @@
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

ascent12:
This shouldn't be necessary. The necessary include directories should be included as part of libdrm's pkg-config file.
If it's not, it's a bug with your distro.

PR #1312, opened 2018-10-14
Fix libdrm includes

ascent12:
This removes any assumptions about how the libdrm headers are installed, and uses the pkg-config include directories as we're "supposed to".
This only adds a partial dependency, since we don't actually need to link against libdrm.


SirCmpwn:
Thanks

PR #1313, opened 2018-10-15
render: fix meson including libdrm

colemickens:
See https://github.com/swaywm/wlroots/pull/1257 for context for why this seems to be necessary.

This fixes the build issues on my system, I believe the right way.

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -80,7 +80,7 @@ examples = {
 			libavcodec,
 			libavformat,
 			libavutil,
-			drm.partial_dependency(includes: true), # <drm_fourcc.h>
+			drm.partial_dependency(compile_args: true), # <drm_fourcc.h>
 			threads,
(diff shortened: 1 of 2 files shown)


colemickens:
cc: @ascent12 as FYI since we've been touching this area back and forth (thanks again!)

ascent12:
I just asked on #mesonbuild. Yeah, compile_args is the correct thing we're supposed to use here.

ascent12:
Also, can you change the one for dmabuf-capture in examples/meson.build?

colemickens:
Well, I'm back to being confused. I build the examples in my package definition, but it was building `dmabuf-capture` just fine.

Anyway, it works with the change applied to the example too, and it seems the right thing to do. I've amended and pushed. Thanks!

ascent12:
Thanks

PR #1314, opened 2018-10-15
Remove fmt parameter from wlr_texture_write_pixels

nyorain:
This uses dummy format values (0xFFFFFFFF) for immutable wlr_gles2_texture objects since we don't always know their formats (and don't write the textures anyways so why bother with additional conversions in that case).
The buffer damage handling had to be changed a bit; the previous implementation seemed to rely on being able to change the textures format (which was broken for the gles2 renderer since has_alpha was not updated on format change).
Breaking change, i guess this is relevant for sway and @ongy @Timidger @ammen99 (add anyone i missed) if it gets merged.

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -19,7 +19,7 @@
 extern PFNGLEGLIMAGETARGETTEXTURE2DOESPROC glEGLImageTargetTexture2DOES;
 
 struct wlr_gles2_pixel_format {
-	uint32_t wl_format;
+	enum wl_shm_format wl_format;
 	GLint gl_format, gl_type;
 	int depth, bpp;
 	bool has_alpha;
@@ -72,7 +72,7 @@ struct wlr_gles2_texture {
 	enum wlr_gles2_texture_type type;
 	int width, height;
 	bool has_alpha;
-	uint32_t wl_format; // used to interpret upload data
+	enum wl_shm_format wl_format; // used to interpret upload data
 	bool inverted_y;
 
 	// Not set if WLR_GLES2_TEXTURE_GLTEX
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -72,6 +72,7 @@ struct wlr_gles2_texture {
 	enum wlr_gles2_texture_type type;
 	int width, height;
 	bool has_alpha;
+	uint32_t wl_format; // used to interpret upload data
 	bool inverted_y;
 
 	// Not set if WLR_GLES2_TEXTURE_GLTEX
--- a/include/wlr/render/wlr_texture.h
+++ b/include/wlr/render/wlr_texture.h
@@ -55,9 +55,11 @@ bool wlr_texture_is_opaque(struct wlr_texture *texture);
 
 /**
  * Update a texture with raw pixels. The texture must be mutable.
+ * The given data is interpreted as being in the format the
+ * texture was created with.
  */
 bool wlr_texture_write_pixels(struct wlr_texture *texture,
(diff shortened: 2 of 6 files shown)


include/wlr/render/wlr_texture.h
```diff
@@ -55,9 +55,11 @@ bool wlr_texture_is_opaque(struct wlr_texture *texture);
 
 /**
  * Update a texture with raw pixels. The texture must be mutable.
+ * The given data is interpreted as being in the format the
+ * texture was created with.
```
SirCmpwn: Can you phrase this instead as:

```
/**
  * Update a texture with raw pixels. The texture must be mutable, and the input
  * data must have the same pixel format that the texture was created with.
  */
```

render/gles2/texture.c
```diff
@@ -55,11 +55,9 @@ static bool gles2_texture_write_pixels(struct wlr_texture *wlr_texture,
 		return false;
 	}
 
-	const struct wlr_gles2_pixel_format *fmt = get_gles2_format_from_wl(wl_fmt);
-	if (fmt == NULL) {
-		wlr_log(WLR_ERROR, "Unsupported pixel format %"PRIu32, wl_fmt);
-		return false;
-	}
+	const struct wlr_gles2_pixel_format *fmt = get_gles2_format_from_wl(
+		texture->wl_format);
```
SirCmpwn: nit: this would be better balanced as

```c
const struct wlr_gles2_pixel_format *fmt =
		get_gles2_format_from_wl(texture->wl_format);
```

render/gles2/texture.c
```diff
@@ -203,6 +202,7 @@ struct wlr_texture *wlr_gles2_texture_from_wl_drm(struct wlr_egl *egl,
 	texture->wl_drm = data;
 
 	EGLint fmt;
+	texture->wl_format = 0xFFFFFFFF; // texture can't be written anyways
```
SirCmpwn: You can just leave this unset in these cases

emersion: "Unset" means 0, which is a valid format (ARGB8888 iirc)

include/render/gles2.h
```diff
@@ -72,6 +72,7 @@ struct wlr_gles2_texture {
 	enum wlr_gles2_texture_type type;
 	int width, height;
 	bool has_alpha;
+	uint32_t wl_format; // used to interpret upload data
```
emersion: This can be `enum wl_shm_format`

render/gles2/texture.c
emersion: Maybe it's worth it to add a `immutable` attribute

include/render/gles2.h
nyorain: ok, will change it in wlr_gles2_pixel_format as well then.

render/gles2/texture.c
nyorain: That's implicitly stored via texture->type. `WLR_GLES2_TEXTURE_GLTEX` signals that it's mutable. And yep, that was my thought with not leaving it 0. We don't rely on this value anywhere btw i just didn't want to store something wrong.

SirCmpwn:
cc @swaywm/wlroots-go @swaywm/cl-wlroots @swaywm/wlroots-rs @swaywm/wlroots-ocaml

render/gles2/texture.c
emersion: I just felt bad about setting an enum to a value outside the enum.

emersion:
Thanks!

PR #1315, opened 2018-10-16
This commit will add the building of a manual by doxygen to the build phase

AdminXVII:
By default, it will build the manual and install it with the build process if Doxygen is present.
Optionally, it will also build HTML documentation with a custom theme. This could be used as github pages if the maintainers sees it fit.

### DIFF
--- a/man/Doxyfile.in
+++ b/man/Doxyfile.in
@@ -1,0 +1,344 @@
+# Doxyfile 1.8.13
+
+#---------------------------------------------------------------------------
+# Project related configuration options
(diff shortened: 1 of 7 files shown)


SirCmpwn:
No thank you.

ascent12:
Yeah, I'm not a fan of Doxygen.

AdminXVII:
Ok. I still believe that man pages are going to a must in the near future, as well as web documentation more user friendly than a pack of header files. If you have some tools you prefer, or even grep and al., I could integrate it with the build system.

PR #1317, opened 2018-10-17
FreeBSD fixes

sghctoma:
This PR fixes two issues on FreeBSD:

1. `CLOCK_MONOTONIC` appeared in IEEE Std. 1003.1-200x, it was not part of
POSIX.1b (the 1993 version). FreeBSD treats it accordingly, thus `_POSIX_C_SOURCE` needs to be set to `200112L`.
2. `epoll` is not part of FreeBSD, it is provided as a 3rd party library, so it needs to be set as a dependency.


SirCmpwn:
Thanks!

PR #1318, opened 2018-10-17
text-input: Don't reject enable requests when unfocused

dcz-purism:
The prevoius implementation would always raise an error in the following sequence:

1. → enter
2. → leave
3. ← enable

The text-input type is not equipped to manage the validity of clents' requests, which should be handled in the compositor, as rootston does.

There *is* one scenario where this error should stay; namely immediately after creating the text input, but handling it introduces additional state and complexity, so I decided against it.

This should also make terminals work again, fixing https://gitlab.gnome.org/GNOME/gtk/merge_requests/369 from the other side.


SirCmpwn:
Thanks!

dcz-purism:
Thanks for the quick merge!

PR #1319, opened 2018-10-18
xwayland: handle configure event mask

emersion:
Fixes https://github.com/swaywm/sway/issues/731

Thanks @ixru for debugging this.


SirCmpwn:
Nice, thanks!

PR #1320, opened 2018-10-19
output-layout, cursor: fix duplicate cursors

emersion:
wlr_cursor: make sure the output doesn't have a cursor before
creating a new one

wlr_output_layout: don't emit the "add" event when the output is
already in the layout

Fixes https://github.com/swaywm/sway/issues/2486
Fixes https://github.com/emersion/grim/issues/20

### DIFF
--- a/types/wlr_cursor.c
+++ b/types/wlr_cursor.c
@@ -620,8 +620,14 @@ static void handle_layout_output_destroy(struct wl_listener *listener,
 
 static void layout_add(struct wlr_cursor_state *state,
 		struct wlr_output_layout_output *l_output) {
-	struct wlr_cursor_output_cursor *output_cursor =
-		calloc(1, sizeof(struct wlr_cursor_output_cursor));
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Thanks!

rtyler:
I can confirm that this fixed the double-cursor issue for me, thus far at least :smile:

koynoyno:
Manjaro Sway 21.1.0 (Kernel 5.10). In my case the reason of the double cursor was recalculation of the cursor position after connection the second (top) monitor.

```
output HDMI-A-1 res 1920x1080 pos 0 0
output eDP-1 res 1920x1080 pos 0 1080
```

I changed it to

```
output HDMI-A-1 res 1920x1080 pos 0 -1080
output eDP-1 res 1920x1080 pos 0 0
```

and the problem is gone.

kennylevinsen:
@koynoyno Please open a new issue rather than commenting on a PR merged 3 years ago.

PR #1326, opened 2018-10-22
Fix wlroots.pc file path on FreeBSD

sghctoma:
I did not catch this earlier because I'm using the ports system to install wlroots, which automatically deals with such issues, but the default directory for pkg-config files is not the right one for FreeBSD.

This commit changes the installation directory of wlroots.pc on FreeBSD to <prefix>/libdata/pkgconfig, and leaves it as the default on other systems.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -31,6 +31,7 @@ conf_data = configuration_data()
 wlr_inc = include_directories('.', 'include')
 
 cc = meson.get_compiler('c')
+prefix = get_option('prefix')
 
 # Clang complains about some zeroed initializer lists (= {0}), even though they
 # are valid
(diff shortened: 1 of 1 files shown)


emersion:
Shouldn't this be fixed in Meson instead?

SirCmpwn:
>`meson.build:164:1: ERROR:  Unknown variable "libdir".`

sghctoma:
@emersion 
You are right, fixing this in Meson seems more appropriate. I'll make a PR there.

@SirCmpwn 
Oops, I didn't test the non-FreeBSD path before PR-ing. Sorry for that.
But if this gets resolved in Meson, it does not matter anyways. Or should I push a fix, just in case Meson won't accept a PR to change the default pkgconfig directory on FreeBSD?

SirCmpwn:
meson is usually pretty responsive to these sorts of issues, I'd just see if they'll take the fix.

sghctoma:
Here's the PR: https://github.com/mesonbuild/meson/pull/4410

emersion:
Closing because this is a Meson bug.

PR #1330, opened 2018-10-24
[WIP] wlr_backend_drm: Add wlr_drm_backend_device_name

NoraCodes:
Gets the name of the device used by the DRM backend. Also refactored to use this internally. This is in support of an upcoming patch related to swaywm/sway#2896

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -140,17 +140,48 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	backend_destroy(&drm->backend);
 }
 
+char* wlr_drm_backend_device_name_from_fd(int gpu_fd) {
+	char *name = drmGetDeviceNameFromFd2(gpu_fd);
+	drmVersion *version = drmGetVersion(gpu_fd);
+
+	if (!name || !version) {
+		return NULL;
+	}
+
+	int len = strlen(name) + strlen(version->name);
+	char *buf = calloc(sizeof(char), 4 + len);
+	if (!buf) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+
+	snprintf(buf, 4 + sizeof(char) * len, "%s (%s)", name, version->name);
+	free(name);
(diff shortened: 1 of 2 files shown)


ascent12:
I don't think an extra function is necessary. I'd just prefer a blacklist of driver names and not initialise the backend if it fails.
I don't want library users to write driver specific code (neither do I, but this will be the only exception).

LeoTindall:
Well, I was hoping to keep the same separation of concerns here - no reason to move the rejection from sway to wlroots that I can see.

If it's preferable, I can do it here, but then overriding the driver blacklist will require a flag be passed every step of the way from library user code into the DRM backend creation and that seems gross.

ascent12:
Leave it as just a warning message then. It will inevitably fail later anyway.

RyanDwyer:
@ascent12 Not necessarily. In a hybrid setup where nvidia is the secondary GPU, the user will not see the warning.

LeoTindall:
The reason this approach is better than the "fail if proprietary drivers are loaded" approach is precisely that. If the user is on hybrid graphics and the wlroots backend is using a supported device, there's no reason to bail out.

ascent12:
@RyanDwyer Since we auto-load every GPU by default, the user will more than likely receive the message.

RyanDwyer:
Yeah, it'll be logged to stderr which they never see because their session is running on their primary GPU.

clarcharr:
Perhaps a simple `wlr_drm_backend_is_supported` might be better? That way, sway and other compositors don't need to add any special logic, just clarify that you're on an unsupported GPU and give you a way to check for more info.

emersion:
>Perhaps a simple wlr_drm_backend_is_supported might be better?

Note that there is only one `wlr_backend` exposed to wlroots users. Secondary GPUs are represented as internal sub-backends.

panaman67:
@NoraCodes Whats the status of this PR?

emersion:
There is now `get_drm_fd`, which allows wlroots users to get this information. Closing.

PR #1332, opened 2018-10-26
util: use shm_open for in-memory files 

emersion:
`shm_open` is a POSIX function creating an in-memory file. Using it simplifies the code and removes the dependency on `XDG_RUNTIME_DIR`. The only downside is that we need to generate a random name for the shm file.


SirCmpwn:
Can you rebase this so we get a FreeBSD build

emersion:
>Can you rebase this so we get a FreeBSD build

:D

Rebased

SirCmpwn:
Thanks!

PR #1337, opened 2018-10-28
idle: allow zero timeout

emersion:
Prior to this commit, setting up a zero timeout resulted in a timer that never
expires, which isn't particularly useful.

This commit allows setting up timers that expire immediately, which is useful
to immediately enter idle state.

### DIFF
--- a/types/wlr_idle.c
+++ b/types/wlr_idle.c
@@ -27,6 +27,9 @@ static void idle_timeout_destroy(struct wlr_idle_timeout *timer) {
 
 static int idle_notify(void *data) {
 	struct wlr_idle_timeout *timer = data;
+	if (timer->idle_state) {
+		return 0;
+	}
 	timer->idle_state = true;
 	org_kde_kwin_idle_timeout_send_idle(timer->resource);
 	return 1;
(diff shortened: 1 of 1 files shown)


types/wlr_idle.c
```diff
@@ -36,13 +39,18 @@ static void handle_activity(struct wlr_idle_timeout *timer) {
...
 	// in case the previous state was sleeping send a resume event and switch state
 	if (timer->idle_state) {
 		timer->idle_state = false;
 		org_kde_kwin_idle_timeout_send_resumed(timer->resource);
 	}
+
+	// rearm the timer
+	wl_event_source_timer_update(timer->idle_source, timer->timeout);
+	if (timer->timeout == 0) {
+		idle_notify(timer);
```
martinetd: How will the wayland marshalling handle if the client destroys the timer in the earlier `send_resumed` message?
(I guess you do that in the swayidle PR and it doesn't blow up, but am a bit curious - I guess the client only sends a request to destroy and we'll destroy later so it'll work out on the server side, but the client side would receive an idle event on something it thinks doesn't exist anymore? Or does the wayland marshalling recognize it got destroyed and just discard it?)

emersion: In `swayidle`, I destroy the client-side object in the resumed handler. The server sends resume immediately followed by idle. libwayland handles this well and discards the idle event.

PR #1338, opened 2018-10-28
render: on Wayland, make eglSwapBuffers non-blocking

emersion:
We always manage frame callbacks manually on the Wayland backend. Thus we don't want `eglSwapBuffers` to register an additional frame callback and to block.


SirCmpwn:
Thanks!

Issue #1325, opened 2018-10-22
FTBFS without xwayland: variable ‘cursor_default’ set but not used


JasonWoof:
I'm getting a compile-time error on latest git master: 6c7e0e48

```
../rootston/desktop.c: In function ‘desktop_create’:
../rootston/desktop.c:885:14: error: variable ‘cursor_default’ set but not used [-Werror=unused-but-set-variable]
  const char *cursor_default = ROOTS_XCURSOR_DEFAULT;
              ^~~~~~~~~~~~~~
```

The variable is only used inside `#ifdef WLR_HAS_XWAYLAND`

So this makes it compile:

```
diff --git a/rootston/desktop.c b/rootston/desktop.c
index 06b785ad..69c71867 100644
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -882,14 +882,18 @@ struct roots_desktop *desktop_create(struct roots_server *server,
 	desktop->tablet_v2 = wlr_tablet_v2_create(server->wl_display);
 
 	const char *cursor_theme = NULL;
+#ifdef WLR_HAS_XWAYLAND
 	const char *cursor_default = ROOTS_XCURSOR_DEFAULT;
+#endif
 	struct roots_cursor_config *cc =
 		roots_config_get_cursor(config, ROOTS_CONFIG_DEFAULT_SEAT_NAME);
 	if (cc != NULL) {
 		cursor_theme = cc->theme;
+#ifdef WLR_HAS_XWAYLAND
 		if (cc->default_image != NULL) {
 			cursor_default = cc->default_image;
 		}
+#endif
 	}
 
 	char cursor_size_fmt[16];
```

SirCmpwn:
Can you send that patch as a pull request?

PR #1339, opened 2018-10-28
Fix "variable ‘cursor_default’ set but not used"

MrSorcus:
Based on https://github.com/swaywm/wlroots/issues/1325


emersion:
Thanks!

PR #1344, opened 2018-10-28
backend/drm: add WLR_DRM_NO_ATOMIC_GAMMA env variable

emersion:
This allows the legacy interface to be used instead of atomic if needed. This
is a workaround to make some Intel GPUs work (if this variable is unset) as
well as to make some AMD GPUs work (if this variable is set).

See https://github.com/swaywm/sway/issues/2700

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -208,10 +208,12 @@ static bool atomic_crtc_set_gamma(struct wlr_drm_backend *drm,
 		struct wlr_drm_crtc *crtc, size_t size,
 		uint16_t *r, uint16_t *g, uint16_t *b) {
 	// Fallback to legacy gamma interface when gamma properties are not available
-	// (can happen on older intel gpu's that support gamma but not degamma)
-	// TEMP: This is broken on AMDGPU. Always fallback to legacy until they get
-	// it fixed. Ref https://bugs.freedesktop.org/show_bug.cgi?id=107459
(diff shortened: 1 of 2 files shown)


backend/drm/atomic.c
```diff
@@ -208,10 +208,12 @@ static bool atomic_crtc_set_gamma(struct wlr_drm_backend *drm,
...
 		uint16_t *r, uint16_t *g, uint16_t *b) {
 	// Fallback to legacy gamma interface when gamma properties are not available
-	// (can happen on older intel gpu's that support gamma but not degamma)
-	// TEMP: This is broken on AMDGPU. Always fallback to legacy until they get
-	// it fixed. Ref https://bugs.freedesktop.org/show_bug.cgi?id=107459
-	if (crtc->props.gamma_lut == 0 || true) {
+	// (can happen on older Intel GPUs that support gamma but not degamma).
+	// TEMP: This is broken on AMDGPU. Provide a fallback to legacy until they
+	// get it fixed. Ref https://bugs.freedesktop.org/show_bug.cgi?id=107459
+	const char *no_atomic_str = getenv("WLR_DRM_NO_ATOMIC_GAMMA");
```
martinetd: I actually looked at this and was wondering about doing a getenv everytime we set gamma.. It's not a fast path but it's the first getenv we'll have that's not at init.
Well, I guess it's probably ok

Issue #1349, opened 2018-10-31
gles2: use GL_OES_read_format to get preferred read format


emersion:
This allows to query the preferred read format for better performance.

Spec:

* https://www.khronos.org/registry/OpenGL/extensions/OES/OES_read_format.txt
* https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_read_format_bgra.txt

Issue #1348, opened 2018-10-31
Grim cannot take a screenshot in sway-1.0 master, GL_INVALID_OPERATION


mithodin:
With the latest master of sway and wlroots, grim fails to take a screenshot, saying "failed to copy output X11-1". I have filed a bug report there (https://github.com/emersion/grim/issues/22), but they claim it is a wlroots issue.

Therefore, I'm posting this bug here. I have attached a logfile. The last entry pops up every time I try to take a screenshot.

Now to avoid the (justified) anger, I am aware that using nvidia drivers is unsupported and if this bug is caused by that, I will be happy to just shut up about it, because I agree that nvidia needs to fix their shit. In my defense, I do not control the hardware on this machine and I need cuda support, so I'm kinda stuck with those shitty drivers. Sorry.

[sway.log](https://github.com/swaywm/wlroots/files/2533333/sway.log)

emersion:
Can you try to reproduce with rootston? Please enable debug logs and upload the whole file.

>I am aware that using nvidia drivers is unsupported and if this bug is caused by that

Since you use the X11 backend, it _should_ work. We don't support nvidia on DRM.

mithodin:
Here's a log. I'm starting rootston just like I normally would sway, open a terminal, try grim test.png, and close the session.

[rootston.log](https://github.com/swaywm/wlroots/files/2533748/rootston.log)

PR #1350, opened 2018-10-31
render/gles2: remove assumptions about supported formats

emersion:
We were assuming GL_BGRA_EXT was always supported.

We now check that it's supported for rendering. We fail if it isn't because
this format is specified as "always supported" by the Wayland protocol.

We also check if it's supported for reading pixels. A new preferred_read_format
function returns the preferred format that can be used to read pixels. This is
used by the screencopy protocol.

Fixes https://github.com/swaywm/wlroots/issues/1348
Fixes https://github.com/swaywm/wlroots/issues/1349

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -30,8 +30,17 @@ bool init_drm_renderer(struct wlr_drm_backend *drm,
 		create_renderer_func = wlr_renderer_autocreate;
 	}
 
+	static EGLint config_attribs[] = {
+		EGL_RED_SIZE, 1,
+		EGL_GREEN_SIZE, 1,
+		EGL_BLUE_SIZE, 1,
+		EGL_ALPHA_SIZE, 1,
+		EGL_NONE,
+	};
+
 	renderer->wlr_rend = create_renderer_func(&renderer->egl,
-		EGL_PLATFORM_GBM_MESA, renderer->gbm, NULL, GBM_FORMAT_ARGB8888);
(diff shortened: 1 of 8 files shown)


emersion:
@mithodin Can you try this pull request?

mithodin:
Maybe I'm a little dumb, how do I check out the pull request?

emersion:
No worries.

```
git fetch origin pull/ID/head:BRANCHNAME
git checkout BRANCHNAME
```

https://help.github.com/articles/checking-out-pull-requests-locally/

Emantor:
Grim does not support this format, 875709016 is `WL_SHM_FORMAT_XBGR8888` while grim supports `WL_SHM_FORMAT_XRGB8888` (note the switched sequence of the R,G, and B letter).
The pull request should be fine, grim will need a patch to reformat the data for cairo.

mithodin:
I'll go report this in the grim repo then.

emersion:
As @Emantor said, yeah now wlroots chooses the correct format but grim doesn't support it. Can you open an issue in the grim repo to support XBGR8888?

Sorry for the roundtrip, but wlroots needed this fix anyway and I wasn't sure whether grim needed to support an additional format or not.

mithodin:
No worries. Glad to help at least a bit by reporting the things in the right place.

emersion:
@ascent12 bump

SirCmpwn:
Thanks!

PR #1351, opened 2018-10-31
backend/drm: check DRM_CAP_ADDFB2_MODIFIERS

emersion:
See https://gitlab.freedesktop.org/wayland/weston/merge_requests/45


emersion:
Actually I'm mistaken, we don't use this feature yet: https://gitlab.freedesktop.org/wayland/weston/commit/a864f58f44d701164dbb32bfcdde8c6d761f28ee

PR #1355, opened 2018-11-03
[WIP] Renderer v6

### LINKED ISSUE #1352: [RFC] Renderer redesign v6
I've been doing some casual research into some DRM/Vulkan related stuff and came up with an idea.

With the changes required to the renderer for Vulkan etc., there is going to be a very significant difference between the rendering paths of the DRM backend and the rest of them, where we need to handle buffers ourselves in DRM, and use a provided swapchain (EGLSurface or VkSurface+VkSwapchain) for the others.

Having two code paths is kind of crap, so why don't we just use one?
Basically, we throw EGLSurfaces/VkSurface/VkSwapchains out entirely, and instead write our own swapchain.

We allocate two (or more?) GBM buffers, wrap them in our swapchain, and use the appropriate backend-specific way to display it. This can work for all backends, not just GBM.

## Wayland
Use `zwp_linux_dmabuf_v1` and  maybe `wp_presentation` directly.
This does require the DRM render node to be given to us, which is supposed to come in the next version of the protocol.
`zwp_linux_explicit_syncronization` would also be relevant, whenever that ends up being done.

## X11
Use the `dri3` and `present` extension directly. The dri3 extension can give us the DRM render node.

## DRM
Business as usual.

## Headless
Just pick some arbitrary render node and use that.

## Potential future backends
GBM buffers are essentially dmabufs, so any backend which can handle dmabufs can work.
They're also mmapable, so this doesn't exclude purely software based backends and renderers.

# Advantages
- Opens the possibility of #778 to be revived again, or at least sharing a rendering node and EGL/VK context between different backend types.
- Adds a lot of consistency between backends; moves a lot of code that would be DRM-specific.
- Makes the API between EGL and Vulkan more similar. We're just dealing with EGLImages and VkImages, which are analogues to each other, and there is no such concept of a native platform.
- Potentially allows non-dmabuf based screencapture protocols to be removed, as we will now be certain we have a dmabuf.

# Disadvantages
- More work for us to do. We have to do all of the X11 and Wayland bits that Mesa would have provided for us. This includes things like damage handling.
I've already done some investigation to how to do it on X11, and it doesn't seem that hard. I'm sure everyone also understands Wayland, so that's easy too.

ascent12:
Closes #1352 

Just to be clear, this is still VERY much a work-in-progress. If I were to guess, I would say that it's only about ~~5%~~ ~~10%~~ 20% done.
I'm posting this now just to let people see what I'm doing and get feedback before I go too deep into changing everything.

~~The code *compiles*, but that's about it. At this point, the entirety of the backends, examples, and rootston have been removed, just until they're updated. There are several #if 0 blocks around, as a "I'll get to it later" thing.~~

I'm intending to build a mailbox swapchain (Vulkan terminology), which arises quite naturally from using dmabufs, and can allow for people who want to write renderers that go faster than the display's refresh rate. [Here](https://software.intel.com/en-us/articles/api-without-secrets-introduction-to-vulkan-part-2#inpage-nav-5-6) is a pretty good explanation of different swapchain types.


ascent12:
I've gotten to the point where `examples/simple` works with the wayland backend. So some of the overall changes could be reviewed or commented on.

ascent12:
>Doesn't work if the parent Wayland compositor doesn't support the linux-dmabuf protocol (not sure what the situation on X11 is)

This is a use case I'm really not interested in. Considering there are no known (relevant) compositors that don't support this, it's a lot of effort for nobody.
It's not an inherent flaw of this design, though. It's just creating a wl_shm buffer and performing a copy into that.

>Makes it impossible to have a software renderer

Not impossible; just not 100% guaranteed to work. gbm_bos are mmapable in most cases. On DRM, it would be possible to use dumb-buffers too.
Again, a lot of effort for basically nobody.

>Ties wlroots to GBM, making it impossible to use another allocator (ie. the future Unix allocator, though I'm not holding my breath)

Considering it hasn't been touched in over 8 months, I've given up on it for the time being. Right now, GBM is the only viable option, so I'm completely fine with depending on that.

emersion:
>This is a use case I'm really not interested in. Considering there are no known (relevant) compositors that don't support this, it's a lot of effort for nobody.
It's not an inherent flaw of this design, though. It's just creating a wl_shm buffer and performing a copy into that.

You still need to get a render node. This isn't even in the current linux-dmabuf protocol.

>Not impossible; just not 100% guaranteed to work. gbm_bos are mmapable in most cases. On DRM, it would be possible to use dumb-buffers too.
Again, a lot of effort for basically nobody.

This would not be viable, as you'll end up doing roundtrips between the GPU and the main memory all the time.

I'd like to implement a software renderer for a display manager.

>(note that devices are advertised per surface, not globally)

This means the FD needs to be per-output, not per-backend.

ascent12:
>You still need to get a render node. This isn't even in the current linux-dmabuf protocol.

Sure. Obviously I'll add that when it's actually made part of the protocol.

>This would not be viable, as you'll end up doing roundtrips between the GPU and the main memory all the time.

I suppose. The allocation model for software renderers is completely different than what's here (backend allocated instead of renderer allocated), and it seems like trying to cram the allocations into the same API would be a mistake. I have an idea about how that may work, but I'll think about it some more.

>This means the FD needs to be per-output, not per-backend.

That certainly causes some issues. That basically mandates that multiple renderers becomes a solved problem.
I am going to wait for the next linux-dmabuf version to see what actually gets merged.

backend/backend.c
```diff
@@ -50,47 +55,60 @@ void wlr_backend_destroy(struct wlr_backend *backend) {
...
-		return backend->impl->get_renderer(backend);
+int wlr_backend_get_render_fd(struct wlr_backend *backend) {
+	if (backend->impl->get_render_fd) {
+		return backend->impl->get_render_fd(backend);
+	}
+
+	return -1;
+}
+
+struct wlr_format_set *wlr_backend_get_formats(struct wlr_backend *backend) {
```
emersion: This can probably be const

backend/wayland/output.c
```diff
@@ -26,56 +32,31 @@ static struct wl_callback_listener frame_listener;
...
-	struct wlr_wl_output *output =
-		get_wl_output_from_output(wlr_output);
+	wl_callback_destroy(output->frame_callback);
 
-	if (output->frame_callback != NULL) {
-		wlr_log(WLR_ERROR, "Skipping buffer swap");
-		return false;
-	}
+	if (output->scheduled) {
+		struct wlr_wl_buffer *buffer = gbm_bo_get_user_data(output->scheduled);
```
emersion: This is pretty weird. The frame event means "please render", not "please commit". Instead we should call `wlr_output_send_frame` here, and let the compositor swap buffers to commit.

backend/wayland/output.c
```diff
@@ -197,26 +176,110 @@ static bool output_move_cursor(struct wlr_output *_output, int x, int y) {
...
+
+	wl_signal_emit(&output->wlr_output.events.release_buffer, &event);
+
+	/*
+	 * It's possible for the compositor to hold onto all of our buffers at
+	 * the same time, before releasing one of them back to us. This'll kill
+	 * our rendering loop, so we start it again by sending a frame event.
+	 */
+	if (!output->frame_callback) {
+		wlr_output_send_frame(&output->wlr_output);
```
emersion: This is pretty fragile.

emersion:
>I got a bit carried away with the changes, and started adding Xinput2 support, but I haven't gotten around to finishing that yet. There are still some other X11 changes that need to be made too.

Please don't make so massive that it becomes impossible to review (it's already). You can open another PR if you want to do this. ;_;

>I think the multi backend needs to be massively overhauled. I haven't thought of how yet.

Yeah I'm not sure how we can fix this.

ascent12:
I'll probably rebase it pretty soon so there is a sane commit history to review and work off of. Perhaps I could also change things so it's more of a "introduce new thing -> gradually change everything to use it -> remove old thing" instead of "change everything at once".

emersion:
> > This means the FD needs to be per-output, not per-backend.
>
>That certainly causes some issues. That basically mandates that multiple renderers becomes a solved problem.
>I am going to wait for the next linux-dmabuf version to see what actually gets merged.

Note that this doesn't necessarily means multiple renderers. The compositor could still use only one GPU for rendering, and the others just for direct scan-out.

We still need to advertise multiple GPUs to clients.

emersion:
Mesa uses `drmGetDevices2` and `DRM_NODE_RENDER` to get a render node (see `loader_open_render_node` in `src/loader/loader.c`).

render/gles2/renderer.c
```diff
@@ -582,60 +692,98 @@ struct wlr_renderer *wlr_gles2_renderer_create(struct wlr_egl *egl) {
...
+		const char *frag_src;
+		size_t num_uniforms;
+		struct uniform *uniforms;
+	} info[] = {
+		{
+			.prog = &gles->shaders.quad.program,
+			.vert_src = quad_vertex_src,
+			.frag_src = quad_fragment_src,
+			.num_uniforms = 2,
+			.uniforms = (struct uniform []) {
```
emersion: Does this work? All elements aren't of the same size anymore…

emersion:
Have you designed the API such that the DRM `IN_FORMATS` property can be supported later?

ascent12:
I've rebased this, but I haven't gotten around to doing all of it yet. Some of the old work is still sitting on https://github.com/ascent12/wlroots/tree/renderer_v6_save.

It no longer removes old interfaces, which will instead be delayed until the very end.

ascent12:
I've added some functions to `wlr_output` related to presentation. They're so that supporting things like `wp_viewporter` and `zwp_fullscreen_shell_v1` are possible in the future.
The functions accumulate state and are all presented in one go. This is so it's easier to add more of these types of functions in the future and not have to keep breaking the API.

include/wlr/types/wlr_output.h
```diff
@@ -270,4 +307,69 @@ enum wl_output_transform wlr_output_transform_invert(
...
+ *
+ * This is designed to match the wp_viewporter protocol interface.
+ *
+ * src_x, src_y, src_w and src_h control cropping.
+ * dest_w and dest_h control scaling.
+ *
+ * This is not applied until the next call to wlr_output_present.
+ *
+ * Not every backend supports this.
+ * TODO: Add a way to query if it's supported ahead of time.
```
emersion: This function should probably return a `bool`

emersion:
Maybe we could rename `present` to `commit` and `set_image` to `attach`.

include/wlr/types/wlr_output.h
ascent12: Yeah, I was originally thinking that, but if this ends up being not supported, this affects how you're going to render things, and you'd probably call this function _after_ you've done that.
I think the way to check needs to be done earlier and without changing the output state.

emersion: I see. Then just add a function that checks that `impl->set_viewport != NULL` and assert this in `wlr_output_set_viewport`?

emersion:
Needs rebase

panaman67:
@ascent12 Any chance at a rebase my friend?

ascent12:
@panaman67 I intend to try and get most of this merged through separate PRs; it'll be too monstrously massive doing it this way. Maybe when I get to the point of actually changing wlr_renderer after laying all of the groundwork elsewhere, I could do that here to tie everything in a nice little bow, but this is mainly just sitting here for me to cherry-pick commits off.
I can't really be bothered with a non-trivial rebase for this branch, because it's probably not going to help very much right now, and is just going to get a bunch of new conflicts before it's ready.

I've been very fickle about what I've been working on, and have only been working on this on-and-off in my own spare time mainly, so it's been slow progress. I'd certainly accept any help from anybody who is interested in getting their hands dirty in wlroots internals and wants to get work like this done more quickly.

emersion:
Hmm. https://gitlab.freedesktop.org/wayland/weston/merge_requests/183#note_177014

emersion:
Almost all of this has now been merged, incrementally and slightly updated (mostly to accommodate for the software renderer). Thanks for working on this PR, it's been instrumental for the renderer v6 refactoring!

PR #1356, opened 2018-11-04
render: correctly set EGL_RENDERABLE_TYPE 

emersion:
Depends on https://github.com/swaywm/wlroots/pull/1350

### DIFF
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -198,7 +198,30 @@ void wlr_renderer_init_wl_display(struct wlr_renderer *r,
 struct wlr_renderer *wlr_renderer_autocreate(struct wlr_egl *egl,
 		EGLenum platform, void *remote_display, EGLint *config_attribs,
 		EGLint visual_id) {
-	if (!wlr_egl_init(egl, platform, remote_display, config_attribs, visual_id)) {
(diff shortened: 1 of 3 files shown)


render/wlr_renderer.c
```diff
@@ -198,7 +198,30 @@ void wlr_renderer_init_wl_display(struct wlr_renderer *r,
 struct wlr_renderer *wlr_renderer_autocreate(struct wlr_egl *egl,
 		EGLenum platform, void *remote_display, EGLint *config_attribs,
 		EGLint visual_id) {
-	if (!wlr_egl_init(egl, platform, remote_display, config_attribs, visual_id)) {
+	// Append GLES2-specific bits to the provided EGL config attributes
+	EGLint gles2_config_attribs[] = {
```
ammen99: Maybe we need to add a note for the `create_renderer_func` parameters that this attribute needs to be set?

emersion: Probably not. `create_renderer_func` might yield a renderer for GLES3, for instance. `create_renderer_func` is responsible for creating the EGL context, so I don't think we need to document how to properly create it here.

ammen99: Ok, I see your point.

SirCmpwn:
Thanks!

PR #1357, opened 2018-11-05
Use _POSIX_C_SOURCE, use shm_open

emersion:
* Prefer `_POSIX_C_SOURCE` over `_XOPEN_SOURCE` where possible (we only need `_XOPEN_SOURCE >= 500` for `M_PI`)
* Simplify examples by using `shm_open` instead of `XDG_RUNTIME_DIR`
* Don't link client examples to wlroots if possible


SirCmpwn:
Thanks!

PR #1360, opened 2018-11-07
[RFC] Layer-shell: Add a request for modifiers.

harishkrupo:
Add a new request to the zwlr_layer_surface_v1 interface for requesting
to get keybord modifiers. This enables applications like the status bar
to get the current layout and present it.

Signed-off-by: Harish Krupo <harishkrupo@gmail.com>

### DIFF
--- a/include/wlr/types/wlr_layer_shell_v1.h
+++ b/include/wlr/types/wlr_layer_shell_v1.h
@@ -52,6 +52,7 @@ struct wlr_layer_surface_v1_state {
 		uint32_t top, right, bottom, left;
 	} margin;
 	bool keyboard_interactive;
+  bool modifier_events;
 	uint32_t desired_width, desired_height;
 	uint32_t actual_width, actual_height;
 };
@@ -94,6 +95,8 @@ struct wlr_layer_surface_v1 {
 		struct wl_signal new_popup;
 	} events;
 
+  struct wl_resource *seat_resource; //For requesting modifier events
+  bool resource_added_to_seat;
 	void *data;
 };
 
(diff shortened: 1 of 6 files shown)


harishkrupo:
I am currently working on a module for [waybar](https://github.com/Alexays/Waybar) ([PR](https://github.com/Alexays/Waybar/pull/85)) to show the current keyboard layout. Each time the user switches the layout using the shortcut set in xkb_options (eg. `xkb_options win_space_toggle`), the layout switches and the modifier event is sent to the currently focused application. Due to this the bar doesn't get the event and is unable to update the layout.
To remedy this, I have added a new request in the layer_surface interface which adds the client into the notifier list and sends the modifier change information to the client.
This is my first time contributing to wlroots. This patch is a quick hack to get it working. Kindly review the patch and suggest changes. If this is not the way to go then please suggest what else can be done to receive the layout change information.

emersion:
I don't think this is in scope for layer-shell. Maybe for another protocol?

Please make a PR/open an issue in the wlr-protocols repo first so we can discuss.

SirCmpwn:
There are a few problems here:

- Code style is wrong all over the place
- Changes to protocols should first be discussed in wlr-protocols
- I don't want to add this to layer-shell

harishkrupo:
> I don't think this is in scope for layer-shell. Maybe for another protocol?
> 
> Please make a PR/open an issue in the wlr-protocols repo first so we can discuss.

Sure, will do so.

harishkrupo:
> There are a few problems here:
> 
> * Code style is wrong all over the place

Yes, this is just for comments. I will fix them :)

> * Changes to protocols should first be discussed in wlr-protocols

Okay, will do so.
> * I don't want to add this to layer-shell

emersion:
Closing because this definitely shouldn't go in layer-shell.

PR #1361, opened 2018-11-08
Force ftruncate for shared memory objects on FreeBSD

sghctoma:
FreeBSD does not allow to use `posix_fallocate` on shared memory objects.

I also think that this function should not be used here anyways, since shared memory objects are not regular files, and `posix_fallocate` should give an ENODEV error if the fd refers to a non-regular file. At least that is my understanding of the POSIX specification.

### DIFF
--- a/include/wlr/config.h.in
+++ b/include/wlr/config.h.in
@@ -14,6 +14,4 @@
 #mesondefine WLR_HAS_XCB_ICCCM
 #mesondefine WLR_HAS_XCB_XKB
 
-#mesondefine WLR_HAS_POSIX_FALLOCATE
-
 #endif
--- a/meson.build
+++ b/meson.build
@@ -72,10 +72,6 @@ if logind.found()
 	wlr_deps += logind
 endif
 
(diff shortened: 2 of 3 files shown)


emersion:
>I also think that this function should not be used here anyways, since shared memory objects are not regular files, and posix_fallocate should give an ENODEV error if the fd refers to a non-regular file. At least that is my understanding of the POSIX specification.

Indeed, this seems correct. Can we drop `posix_fallocate` altogether for all platforms?

ascent12:
I think `ftruncate` is the POSIX-compliant way to set the size of a shared memory object.

emersion:
>I think ftruncate is the POSIX-compliant way to set the size of a shared memory object.

Yeah, it's used in the example of the POSIX spec.

SirCmpwn:
>Can we drop posix_fallocate altogether for all platforms?

This

sghctoma:
Pushed a commit that uses `ftruncate` on all platforms.

sghctoma:
Sure, I've removed it.
Didn't want to touch it earlier, because it was not introduced with the shm stuff, but now I see that it was only used in os-compatibility.c, which is not there anymore.

emersion:
Thanks!

PR #1362, opened 2018-11-09
Fix compilation error on musl

julio641742:
I have the following error when compiling latest sway+wlroots on Void Linux with musl 
```
[130/204] Compiling C object 'types/types@@wlr_types@sta/data_device_wlr_data_offer.c.o'.
FAILED: types/types@@wlr_types@sta/data_device_wlr_data_offer.c.o
cc -Itypes/types@@wlr_types@sta -Itypes -I../types -I. -I../ -Iinclude -I../include -Iprotocol -I/usr/include/pixman-1 -flto -fdiagnostics-color=always -DNDEBUG -pipe -D_FILE_OFFSET_BITS=64 -Werror -std=c11 -Wno-unused-parameter '-DWLR_SRC_DIR="/builddir/wlroots-master"' -DWLR_USE_UNSTABLE -DWL_HIDE_DEPRECATED -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -fPIC  -MD -MQ 'types/types@@wlr_types@sta/data_device_wlr_data_offer.c.o' -MF 'types/types@@wlr_types@sta/data_device_wlr_data_offer.c.o.d' -o 'types/types@@wlr_types@sta/data_device_wlr_data_offer.c.o' -c ../types/data_device/wlr_data_offer.c
../types/data_device/wlr_data_offer.c: In function 'data_offer_choose_action':
../types/data_device/wlr_data_offer.c:53:15: error: implicit declaration of function 'ffs' [-Werror=implicit-function-declaration]
  return 1 << (ffs(available_actions) - 1);
               ^~~
ninja: build stopped: subcommand failed.
```
Investigating the error, I found that it was introduced when commit 2d0c5ec was merged.
That commit removed the `_XOPEN_SOURCE` definition. If you read the `strings.h` musl implementation [here](https://git.musl-libc.org/cgit/musl/tree/include/strings.h#n23) we can see that the definition is needed if the `ffs` function is to be included.

```
#if defined(_XOPEN_SOURCE) || defined(_GNU_SOURCE)  || defined(_BSD_SOURCE)
int ffs (int);
int ffsl (long);
int ffsll (long long);
#endif
```

### DIFF
--- a/types/data_device/wlr_data_offer.c
+++ b/types/data_device/wlr_data_offer.c
@@ -1,3 +1,4 @@
+#define _XOPEN_SOURCE 700
 #include <assert.h>
 #include <stdlib.h>
 #include <strings.h>


emersion:
Thanks!

emersion:
We should add an Alpine build to the CI to prevent this kind of things.

SirCmpwn:
>We should add an Alpine build to the CI to prevent this kind of things.

Be the change you want to see in the world

emersion:
https://builds.sr.ht/~emersion/job/10064 :(

SirCmpwn:
I thought I updated that in testing. I guess not. I'll do it this weekend if you don't beat me to it.

emersion:
https://github.com/alpinelinux/aports/pull/5591

Now we can wait a few months :P

sghctoma:
Additionally, we could maybe do what was done in swaywm/sway#2715 (i.e. restrict the default build environment to C11 features on FreeBSD). This could also help catching these mistakes, but with the following drawbacks:

- As @SirCmpwn commented on the above mentioned PR, `_C11_SOURCE` is indeed a FreeBSD-specific macro.
- `__BSD_VISIBLE` would be needed in backend/session/direct-freebsd.c, because sys/consio.h and sys/kbio.h both use `u_short`, `u_char`, etc. Using an internal feature macro is bad, but on the other hand, it's already used in xwayland/xwayland.c.

If you are OK with these, I could send a PR.

SirCmpwn:
I'm opposed to using `_BSD_SOURCE` for the same reasons I'm opposed to using `_GNU_SOURCE`

emersion:
That makes sense. I'm not sure what's the relationship between `_BSD_SOURCE` and `__BSD_VISIBLE`.

Also maybe this is something that could be improved by opening an issue on the FreeBSD tracker?

emersion:
>Instead of defining __BSD_VISIBLE, _C11_SOURCE could be undefined for files that need FreeBSD specific definitions, though.

I'm okay with that.

>Yep, because of SOCK_CLOEXEC. Maybe it could be eliminated using the socket+fcntl combo I've used in swaywm/sway#2715, but I'm not really sure if it would have any side effects here. I mean, SOCK_CLOEXEC exists to prevent the potential race condition caused by the two separate function calls, and I don't know if that race applies here.

Yeah, I think it's fine. (Since we only have a single thread, I don't see how the race condition could happen)

>While we are speaking of xwayland/xwayland.c: the _DEFAULT_SOURCE macro there is a Glibc thing, so maybe that should also be eliminated. I'll look into this.

+1, I think it's used somewhere else too (xcursor?). But it's needed there for `DT_*` constants IIRC. Not sure what to do.

>That said, backend/session/direct-freebsd.c is FreeBSD-specific anyways, so I don't think there would be a real downside using the macro.

Yes.

>BTW, backend/session/direct.c is not standard either (e.g. major is not POSIX), so using _POSIX_C_SOURCE without any other macros there should be an error, but Glibc is a little lenient with these type of stuff (as further evidenced by this issue).

We should probably re-define those, they seem to be just macros.

>I'm not sure this is something that FreeBSD should resolve. FreeBSD is very strict with the feature test macros, as it should be. If a code is declared to be POSIX by defining _POSIX_C_SOURCE, it should be POSIX. The same goes for C99, C11, etc.

Yeah, your explanation makes sense.

PR #1364, opened 2018-11-09
Tell Mesa not to attempt X11 headers

ManDay:
If no X11 related things are being built, tell Mesa's eglplatform.h not to attempt inclusion of associated X11 header.


include/wlr/render/egl.h
```diff
@@ -9,6 +9,10 @@
 #ifndef WLR_RENDER_EGL_H
 #define WLR_RENDER_EGL_H
 
+#ifndef WLR_HAS_X11_BACKEND && WLR_HAS_XWAYLAND
```
SirCmpwn: This isn't how `ifdef/ifndef` works. You want something more like `#if defined(...) && defined(...)`.

ManDay: Fixed. Thank you!

SirCmpwn:
Thanks!

PR #1366, opened 2018-11-10
Render software cursors in compositor

### LINKED ISSUE #1363: Make the compositor render software cursors
Making wlroots responsible for rendering them causes some issues. It adds rendering code when swapping buffers, after `wlr_renderer_end`. It causes issues like https://github.com/swaywm/wlroots/issues/1291, because `wlr_output` is not presentation-time-aware (and shouldn't be).

We should probably remove the rendering code from `wlr_output_swap_buffers` and make the compositor do it. We could provide a helper function to render software cursors that could be called by the compositor in its rendering process.

Same applies to fullscreen surfaces -- but we should remove `wlr_output_set_fullscreen_surface` altogether, replacing it with something else.

@ascent12 How does this interact with the renderer redesign?

---

wlroots has migrated to gitlab.freedesktop.org. This issue has been moved to:

https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/1363

emersion:
This PR removes `wlr_output_set_fullscreen_surface`, adds a new `wlr_output_render_software_cursors` helper, and removes the 180 damage flip in `wlr_output_swap_buffers` (which is GL-specific).

This also fixes frame events for hardware cursors, which were sent too soon.

While this improves the situation, this doesn't completely fix <span></span>#1363, because redesigning how cursor surfaces are handled and how output damage is managed requires more thought.

See #1363

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -106,11 +106,6 @@ struct wlr_output {
 
 	struct wl_event_source *idle_frame;
 
-	struct wlr_surface *fullscreen_surface;
-	struct wl_listener fullscreen_surface_commit;
-	struct wl_listener fullscreen_surface_destroy;
-	int fullscreen_width, fullscreen_height;
-
 	struct wl_list cursors; // wlr_output_cursor::link
 	struct wlr_output_cursor *hardware_cursor;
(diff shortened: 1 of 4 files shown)


emersion:
Rebased

SirCmpwn:
Thanks!

PR #1368, opened 2018-11-11
X11 backend improvements

ascent12:
This contains a lot of the X11 backend refactoring and changes I did in my #1355 that really deserved to be part of a different PR, with the history cleaned up.

This removes xcb-xkb, and adds xcb-xinput and xcb-xfixes as dependencies.

~~While working on this PR, I actually encountered another bug which affects this:~~

~~I've changed the X11 backend to use xfixes to hide the cursor, but we need to turn this off once the pointer leaves our window. However, it seems that sway/wlroots is not sending the "leave" event, so the cursor remains hidden.
While the X11 backend is open in sway, your cursor will be invisible over other Xwayland windows.
This does not affect weston or native X11.
I need to debug this more, but is technically not a bug in this PR.~~
Never mind all that. It was because I wasn't flushing the connection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,6 @@ If you choose to enable X11 support:
 * x11-xcb
 * xcb-errors (optional, for improved error reporting)
 * x11-icccm (optional, for improved Xwayland introspection)
-* xcb-xkb (optional, for improved keyboard handling on the X11 backend)
 
 Run these commands:
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ If you choose to enable X11 support:
 * xcb
 * xcb-composite
 * xcb-xfixes
+* xcb-xinput
 * xcb-image
 * xcb-render
 * x11-xcb
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -104,46 +104,6 @@ static bool backend_start(struct wlr_backend *backend) {
 	struct wlr_x11_backend *x11 = get_x11_backend_from_backend(backend);
 	x11->started = true;
 
-	struct {
-		const char *name;
-		xcb_intern_atom_cookie_t cookie;
-		xcb_atom_t *atom;
-	} atom[] = {
-		{
-			.name = "WM_PROTOCOLS",
-			.atom = &x11->atoms.wm_protocols,
-		},
-		{
-			.name = "WM_DELETE_WINDOW",
-			.atom = &x11->atoms.wm_delete_window,
-		},
-		{
-			.name = "_NET_WM_NAME",
-			.atom = &x11->atoms.net_wm_name,
-		},
-		{
-			.name = "UTF8_STRING",
-			.atom = &x11->atoms.utf8_string,
-		},
-	};
-
-	for (size_t i = 0; i < sizeof(atom) / sizeof(atom[0]); ++i) {
-		atom[i].cookie = xcb_intern_atom(x11->xcb,
-			true, strlen(atom[i].name), atom[i].name);
-	}
-
-	for (size_t i = 0; i < sizeof(atom) / sizeof(atom[0]); ++i) {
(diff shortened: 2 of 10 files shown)


backend/x11/meson.build
```diff
@@ -1,10 +1,8 @@
 x11_libs = []
 x11_required = [
-	'xcb',
 	'x11-xcb',
-]
-x11_optional = [
-	'xcb-xkb',
+	'xcb',
+	'xcb-xinput',
```
emersion: What about xfixes?

backend/x11/output.c
```diff
@@ -142,45 +148,53 @@ struct wlr_output *wlr_x11_output_create(struct wlr_backend *backend) {
...
-	parse_xcb_setup(wlr_output, x11->xcb_conn);
+	parse_xcb_setup(wlr_output, x11->xcb);
 
 	uint32_t mask = XCB_CW_EVENT_MASK;
 	uint32_t values[] = {
 		XCB_EVENT_MASK_EXPOSURE |
-		XCB_EVENT_MASK_KEY_PRESS | XCB_EVENT_MASK_KEY_RELEASE |
-		XCB_EVENT_MASK_BUTTON_PRESS | XCB_EVENT_MASK_BUTTON_RELEASE |
-		XCB_EVENT_MASK_POINTER_MOTION |
 		XCB_EVENT_MASK_STRUCTURE_NOTIFY
```
emersion: This can be on a single line.

README.md
```diff
@@ -62,12 +62,12 @@ If you choose to enable X11 support:
 * xcb
 * xcb-composite
 * xcb-xfixes
+* xcb-xinput
```
emersion: (This could be marked as optional, only needed for the X11 backend)

ascent12: All of the xcb* dependencies could, because Xwayland is optional too.

emersion: Right

backend/x11/backend.c
```diff
@@ -1,25 +1,29 @@
...
 #include <wayland-server.h>
+#include <xcb/xcb.h>
+#include <xcb/xfixes.h>
+#include <xcb/xinput.h>
+
 #include <wlr/backend/interface.h>
 #include <wlr/backend/x11.h>
-#include <wlr/config.h>
 #include <wlr/interfaces/wlr_input_device.h>
 #include <wlr/interfaces/wlr_keyboard.h>
```
SirCmpwn: What's up with all the weird newlines?

emersion: I like the way includes are groupped like this, but it's inconsistent with other files in the repo.

SirCmpwn: I just alphasort headers. No need to think too hard about it.

emersion: You don't, otherwise `#include "thing"` would show up before `#incldue <thing>`.

You create two groups, alphasort each, and don't separate the two with a newline.

emersion:
Needs rebase.

ascent12:
Rebased.

backend/x11/meson.build
```diff
@@ -16,14 +15,6 @@ foreach lib : x11_required
 	x11_libs += dep
 endforeach
 
-foreach lib : x11_optional
-	dep = dependency(lib, required: get_option(lib))
-	if dep.found()
-		x11_libs += dep
-		conf_data.set10('WLR_HAS_' + lib.underscorify().to_upper(), true)
```
emersion: `WLR_HAS_XCB_XKB` needs to be removed from the toplevel `meson.build`

emersion: Also from `include/wlr/config.h.in`

ascent12: Done

emersion:
Thanks!

PR #1369, opened 2018-11-11
Wayland backend improvements

ascent12:
This contains a lot of the Wayland backend refactoring and changes I did in my #1355 that really deserved to be part of a different PR, with the history cleaned up.


SirCmpwn:
Thanks!

Issue #1365, opened 2018-11-09
Use #if instead of #ifdef for config.h


emersion:
`#ifdef`s silently fail when `config.h` is not included. This has been a source of silent mistakes in the past. Instead we could `#define` constants in `config.h` with 0 or 1 and use `#if`.

Meson has:

```
conf.set10(name, bool)
```

ascent12:
Something people don't generally know that the C preprocessor is that if you `#if` an identifier that doesn't exist as a macro, it'll just get replaced with 0. That alone won't fix it, but I guess the `conf.set10` will.

I'm for this change, but I also think we may want to unconditionally add `#include <wlr/config..h>` basically everywhere, at the beginning of every file, including all of our public headers, even if it's not technically required for that file.

emersion:
>Something people don't generally know that the C preprocessor is that if you #if an identifier that doesn't exist as a macro, it'll just get replaced with 0. That alone won't fix it, but I guess the conf.set10 will.

We can set `-Wundef` to make sure this doesn't happen.

PR #1372, opened 2018-11-12
Use #if instead of #ifdef for wlroots config data

emersion:
This prevents some annoying issues when e.g. not including wlr/config.h or
making a typo in the guard name.

For instance, [it caught a typo](https://github.com/swaywm/wlroots/pull/1372/files#diff-282b30593da47b681ecb7626496f2032R14) of `WLR_HAS_XWAYLAND` (which was written as `WLR_HAS_WAYLAND`).

Fixes https://github.com/swaywm/wlroots/issues/1365

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -17,8 +17,7 @@
 #include <wlr/util/log.h>
 #include "backend/multi.h"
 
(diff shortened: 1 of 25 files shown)


SirCmpwn:
Thanks!

PR #1376, opened 2018-11-13
[WIP] Implement linux-dmabuf-unstable-v1 version 4

emersion:
This new protocol version adds the rendering device FD to be able to completely replace `wl_drm`.

The wayland-protocols patch is in review: https://patchwork.freedesktop.org/patch/261368/

I'm also working on a Mesa patch.


emersion:
Mesa patch: https://lists.freedesktop.org/archives/mesa-dev/2018-November/209804.html

panaman67:
@emersion What is the status on this PR?

emersion:
Updated with the latest protocol proposal.

It may be a better idea to get the render FD from the renderer instead of the backend.

emersion:
> get the render FD from the renderer

This is now implemented.

emersion:
This PR now implements per-surface hints. Here's a patch that sets these for fullscreen surfaces in Sway: https://github.com/swaywm/sway/pull/6313

types/wlr_linux_dmabuf_v1.c
```diff
@@ -400,6 +412,255 @@ static void linux_dmabuf_create_params(struct wl_client *client,
...
+
+		for (size_t j = 0; j < fmt->len; j++) {
+			uint32_t modifier_lo = fmt->modifiers[j] & 0xFFFFFFFF;
+			uint32_t modifier_hi = fmt->modifiers[j] >> 32;
+			zwp_linux_dmabuf_hints_v1_send_tranche_modifier(resource,
+				fmt->format, modifier_hi, modifier_lo);
+		}
+	}
+
+	zwp_linux_dmabuf_hints_v1_send_tranche_done(resource);
```
fooishbar, line 537: Forgetting to send `flags`?

emersion: Oops. Should be fixed now, thanks!

fooishbar:
> > It may be a better idea to get the render FD from the renderer instead of the backend.
> 
> After thinking about it, I think it's a better idea to get the render FD from the renderer. The primary device is defined as the device which must always be able to access the DMA-BUFs. We want the renderer to always be able to import DMA-BUFs (as a fallback if direct scan-out doesn't work), so it's natural to have this information come from the renderer.

Yeah, strong agree. Your main device should be your renderer of last resort - EGL/GLES since that 'always' works - and then the backend should only be providing per-tranche target devices when it wants to try something potentially more optimal (e.g. direct to KMS or media encoder).

fooishbar:
Generally this LGTM. I'm not au fait with wlroots so not qualified to give a full review, but I couldn't pick any holes in this or the sway bit either. Thanks a lot Simon!

render/drm_format_set.c
```diff
@@ -93,6 +93,26 @@ bool wlr_drm_format_set_add(struct wlr_drm_format_set *set, uint32_t format,
...
+	dst->formats = calloc(dst->cap, sizeof(struct wlr_drm_format *));
+	if (dst->formats == NULL) {
+		return false;
+	}
+	for (size_t i = 0; i < src->len; i++) {
+		const struct wlr_drm_format *fmt = src->formats[i];
+		size_t fmt_size = sizeof(*fmt) + sizeof(fmt->modifiers[0]) * fmt->cap;
+		dst->formats[i] = malloc(fmt_size);
+		if (dst->formats[i] == NULL) {
+			return false;
```
leandrohrb, line 109: Forgot to free the `dst->formats[i]` that were allocated until this failure and also `dst->formats`.

types/wlr_linux_dmabuf_v1.c
```diff
@@ -400,6 +412,257 @@ static void linux_dmabuf_create_params(struct wl_client *client,
...
+	if (dst->tranches == NULL) {
+		return false;
+	}
+
+	for (size_t i = 0; i < src->tranches_len; i++) {
+		dst->tranches[i].target_device = src->tranches[i].target_device;
+		dst->tranches[i].flags = src->tranches[i].flags;
+		if (!wlr_drm_format_set_copy(&dst->tranches[i].formats,
+				&src->tranches[i].formats)) {
+			return false;
```
leandrohrb, line 506: Forgot to free the other `dst->tranches[i].formats` that were allocated and also `dst->tranches`.

leandrohrb:
Thanks for writing this up, LGTM!

Maybe would be a good to add hysteresis to this MR as well, as suggested for Weston? https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/544#note_919169

emersion:
Thanks for the comments.

> Maybe would be a good to add hysteresis to this MR as well, as suggested for Weston?

I'd rather keep that for later.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1376

PR #1377, opened 2018-11-15
Add support for libinput_switch input devices

tokyovigilante:
These are used primarily by laptops to signal
the state of the lid (open/closed) and tablet
mode if supported, based on ACPI events.

AFAICT this should be feature-complete for wlroots itself, I've added what looks correct for rootston but am unsure if this is all I need. 

Seems to be working well within sway, but I assume sway will also need some formal support for these new events.

```
2018-11-15 00:13:53 - [backend/libinput/events.c:94] Added Lid Switch [0:5]
2018-11-15 00:13:53 - [backend/libinput/events.c:173] Switch capability match for device Lid Switch
2018-11-15 00:13:53 - [backend/libinput/switch.c:20] Created switch for device Lid Switch
2018-11-15 00:13:53 - [backend/libinput/events.c:173] Switch capability match for device ThinkPad Extra Buttons
2018-11-15 00:13:53 - [backend/libinput/switch.c:20] Created switch for device ThinkPad Extra Buttons
2018-11-15 00:14:37 - [backend/libinput/switch.c:41] Got switch event type 2 state 1 from ThinkPad Extra Buttons
2018-11-15 00:14:58 - [backend/libinput/switch.c:41] Got switch event type 2 state 0 from ThinkPad Extra Buttons
2018-11-15 00:15:12 - [backend/libinput/switch.c:41] Got switch event type 1 state 1 from Lid Switch
2018-11-15 00:15:16 - [backend/libinput/switch.c:41] Got switch event type 1 state 0 from Lid Switch
```
Just an RFC currently, this is my first delve into the guts of wlroots.

### DIFF
--- a/include/rootston/bindings.h
+++ b/include/rootston/bindings.h
@@ -4,6 +4,6 @@
 #include "rootston/seat.h"
 #include "rootston/input.h"
 
-void execute_binding_command (struct roots_seat *seat, struct roots_input *input, const char *command);
+void execute_binding_command(struct roots_seat *seat, struct roots_input *input, const char *command);
 
 #endif //ROOTSTON_BINDINGS_H
--- a/include/rootston/bindings.h
+++ b/include/rootston/bindings.h
@@ -1,0 +1,9 @@
+#ifndef ROOTSTON_BINDINGS_H
+#define ROOTSTON_BINDINGS_H
+
+#include "rootston/seat.h"
+#include "rootston/input.h"
+
+void execute_binding_command (struct roots_seat *seat, struct roots_input *input, const char *command);
+
+#endif //ROOTSTON_BINDINGS_H
--- a/rootston/bindings.c
+++ b/rootston/bindings.c
@@ -1,0 +1,109 @@
+#include <string.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include <wlr/util/log.h>
+
+#include "rootston/bindings.h"
+
+static bool outputs_enabled = true;
+
+static const char *exec_prefix = "exec ";
+
+static void double_fork_shell_cmd(const char *shell_cmd) {
+    pid_t pid = fork();
+    if (pid < 0) {
+        wlr_log(WLR_ERROR, "cannot execute binding command: fork() failed");
+        return;
+    }
+
+    if (pid == 0) {
+        pid = fork();
+        if (pid == 0) {
+            execl("/bin/sh", "/bin/sh", "-c", shell_cmd, NULL);
+            _exit(EXIT_FAILURE);
+        } else {
+            _exit(pid == -1);
+        }
+    }
+
+    int status;
+    while (waitpid(pid, &status, 0) < 0) {
+        if (errno == EINTR) {
+            continue;
+        }
+        wlr_log_errno(WLR_ERROR, "waitpid() on first child failed");
+        return;
+    }
+
+    if (WIFEXITED(status) && WEXITSTATUS(status) == 0) {
+        return;
+    }
+
+    wlr_log(WLR_ERROR, "first child failed to fork command");
+}
+
+void execute_binding_command (struct roots_seat *seat, struct roots_input *input, const char *command) {
+    if (strcmp(command, "exit") == 0) {
+        wl_display_terminate(input->server->wl_display);
+    } else if (strcmp(command, "close") == 0) {
+        struct roots_view *focus = roots_seat_get_focus(seat);
+        if (focus != NULL) {
+            view_close(focus);
+        }
+    } else if (strcmp(command, "fullscreen") == 0) {
+        struct roots_view *focus = roots_seat_get_focus(seat);
+        if (focus != NULL) {
+            bool is_fullscreen = focus->fullscreen_output != NULL;
+            view_set_fullscreen(focus, !is_fullscreen, NULL);
+        }
+    } else if (strcmp(command, "next_window") == 0) {
+        roots_seat_cycle_focus(seat);
+    } else if (strcmp(command, "alpha") == 0) {
+        struct roots_view *focus = roots_seat_get_focus(seat);
+        if (focus != NULL) {
+            view_cycle_alpha(focus);
+        }
+    } else if (strncmp(exec_prefix, command, strlen(exec_prefix)) == 0) {
+        const char *shell_cmd = command + strlen(exec_prefix);
+        double_fork_shell_cmd(shell_cmd);
+    } else if (strcmp(command, "maximize") == 0) {
+        struct roots_view *focus = roots_seat_get_focus(seat);
+        if (focus != NULL) {
+            view_maximize(focus, !focus->maximized);
+        }
+    } else if (strcmp(command, "nop") == 0) {
+        wlr_log(WLR_DEBUG, "nop command");
+    } else if (strcmp(command, "toggle_outputs") == 0) {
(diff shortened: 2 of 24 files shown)


tokyovigilante:
Updated with correct `wlr_event.state` and `wlr_event.type` enums to fix BSD build and removed debug logs.

emersion:
This looks good for the wlroots side. Now we need to decide how to expose this in rootston config (and later in sway).

tokyovigilante:
Thanks. I'd assume they probably just need to be exposed as hooks, I can't imagine there will be a set of actions which are universally appropriate, rather will depend on user hardware and preference.

emersion:
What about something like this

```
[device:Just A Switch]
switch_toggle = exec echo toggle
switch_toggle:on = exec echo on
```

tokyovigilante:
On further thought, I wonder if it's better to just listen to the switch types, regardless of the device that generates them? 

There are [only two](https://wayland.freedesktop.org/libinput/doc/latest/switches.html),` LIBINPUT_SWITCH_LID` and `LIBINPUT_SWITCH_TABLET_MODE`, and I doubt any device would have more than one of each, so perhaps it's better to ignore the device name (which will likely vary from laptop to laptop) then just react to the events by specifying a binding command, so:

```
[switches]
lid:open = exec echo open 
lid:close = exec echo close
lid:toggle = exec echo toggle
tablet_mode:true = exec echo tablet
tablet_mode:false = exec echo laptop
tablet_mode;toggle = exec echo toggle
```

emersion:
Yeah, maybe that's better. Maybe we could match the part on the left of the colon with either the device name or the type.

emersion:
```
16:51 <tokyovigilante> Does rootston process general libinput events? I'm trying to add tablet and laptop switch support but the devices never seem to be initialised in rootston
16:59 <damjan> tokyovigilante: isn't that done in wlroots?
17:00 <tokyovigilante> yeah I've done added support for it in wlroots, but rootston doesn't seem to see those devices to listen to events from
17:01 <tokyovigilante> It only seems to see the keyboard and pointer. I'm testing rootston from inside a sway session, maybe that's the problem
```

libinput is not used if you run a nested compositor (e.g. if you run rootston inside of sway).

tokyovigilante:
Oh right, makes sense. I did try to run from a TTY but get a crash with exit on startup without obvious debug message. Will try and solve.

tokyovigilante:
This latest set adds some support for Rootston, and logs a debug message when the switches are triggered. I'm not sure whether to add another class for the switch command handling itself, as this would duplicate a lot of the code in keyboard.c, or to factor this code out to a separate shared class.

emersion:
> I'm not sure whether to add another class for the switch command handling itself, as this would duplicate a lot of the code in keyboard.c, or to factor this code out to a separate shared class.

You mean another file?

Probably better to split the command handling stuff in a separate file.

tokyovigilante:
Oh yeah sorry another file. OK I'll refactor it.

tokyovigilante:
OK, I think this is ready for a review, all working on my system with the config:

```
[switch:lid]
on = exec echo "Lid opened"
off = exec echo "Lid closed"
toggle = exec echo "Lid moved"

[switch:tablet]
on = exec echo "Entered tablet mode"
off = exec echo "Entered laptop mode"
toggle = exec echo "Toggled tablet mode"

[switch:ThinkPad Extra Buttons]
on = exec echo "Matched named device switch on"
off = exec echo "Matched named device switch off"
```
Only wrinkle is that the lid switch events don't seem to be received by `rootston`. I wonder if they are intercepted by ACPI as `systemd` is set to suspend automatically when the lid is shut.

tokyovigilante:
Whoops, rebased.

emersion:
Please remove "WIP" from the PR title once it's ready for review!

tokyovigilante:
Oh sorry! Have been working on Support (which is definitely still WIP) but think this PR is in reasonable shape.

include/rootston/bindings.h
```diff
@@ -0,0 +1,9 @@
+#ifndef ROOTSTON_BINDINGS_H
+#define ROOTSTON_BINDINGS_H
+
+#include "rootston/seat.h"
+#include "rootston/input.h"
+
+void execute_binding_command (struct roots_seat *seat, struct roots_input *input, const char *command);
```
ddevault: style: no space before (

rootston/bindings.c
```diff
@@ -0,0 +1,109 @@
+#include <string.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include <wlr/util/log.h>
+
+#include "rootston/bindings.h"
```
ddevault: style: put all of the includes together, `<>` then `""`, alpha sorted.

rootston/bindings.c
```diff
@@ -0,0 +1,109 @@
...
+                mode = mode == WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE
+                    ? WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE
+                    : WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE;
+                wlr_xdg_toplevel_decoration_v1_set_mode(
+                    decoration->wlr_decoration, mode);
+            }
+        }
+    } else if (strcmp(command, "break_pointer_constraint") == 0) {
+        struct wl_list *list =
+            &input->seats;
```
ddevault: style: why is this line broken up?

tokyovigilante:
Thanks, will make the style changes ASAP. Is there a style guide about?

On Mon, Dec 10, 2018, at 3:16 AM, Drew DeVault wrote:
> *@ddevault* requested changes on this pull request.


> Overall this looks pretty good, nice work


> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[1], or mute the thread[2].

Links:

  1. https://github.com/swaywm/wlroots/pull/1377#pullrequestreview-182984582

emersion:
Yes, see CONTRIBUTING.md.

rootston/bindings.c
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    if (WIFEXITED(status) && WEXITSTATUS(status) == 0) {
+        return;
+    }
+
+    wlr_log(WLR_ERROR, "first child failed to fork command");
+}
+
+void execute_binding_command (struct roots_seat *seat, struct roots_input *input, const char *command) {
```
emersion: Style: 80 char limit

ascent12: And the space before the (

rootston/switch.c
```diff
@@ -0,0 +1,33 @@
...
+    struct roots_switch_config *sc;
+    wl_list_for_each(sc, bound_switches, link) {
+        bool device_match = false;
+        bool state_match = false;
+        if ((sc->name != NULL && strcmp(event->device->name, sc->name) == 0) ||
+                (sc->name == NULL && event->switch_type == sc->switch_type)) {
+            device_match = true;
+        }
+        if (!device_match) {
+            break;
```
emersion: I guess this should be `continue`?

I think the logic here can be improved, no need for these `bool` variables.

tokyovigilante:
Style nits attended to, and I've improved the config matching logic somewhat.

emersion:
Thanks!

tokyovigilante:
No, thank you!

PR #1380, opened 2018-11-21
gtk-primary-selection: support multiple devices

emersion:
When a client was creating multiple data devices for the same seat, we were
only creating one resource. This is a protocol error.

Instead, create one offer per data device.

This commit also makes offers inert when their source is destroyed.

Fixes part of https://github.com/swaywm/wlroots/issues/1041
Supersedes https://github.com/swaywm/wlroots/pull/1113

### DIFF
--- a/include/wlr/types/wlr_primary_selection.h
+++ b/include/wlr/types/wlr_primary_selection.h
@@ -24,8 +24,6 @@ struct wlr_primary_selection_device_manager {
 	void *data;
 };
 
-struct wlr_primary_selection_offer;
-
 struct wlr_primary_selection_source {
 	// source metadata
 	struct wl_array mime_types;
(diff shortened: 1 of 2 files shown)


baloo:
I gave the patch in the PR a shot with firefox 65 (hg master) and sway (master). It doesn't crash anymore when I'm selecting text in the address bar.

uwqwxing886:
After applying the patch, the `ipc_channel_posix.cc` errors seen in #3170 and other issues no longer occur when running firefox.

```
[Child 20714, Chrome_ChildThread] WARNING: pipe error (40): Connection reset by peer: file /var/cache/pkg/firefox-wayland/src/mozilla-unified/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 363
```

PR #1382, opened 2018-11-21
Skip interlaced modes during drm mode detection

atomnuker:
Fixes swaywm/sway#3167
A better solution could be made if there's a need to.


emersion:
Thanks!

PR #1384, opened 2018-11-21
data-device: allow multiple devices for the same seat

emersion:
This commit makes it possible for a single client to have multiple data devices
for the same seat. This fixes issues with Firefox.

This mainly removes wlr_data_source.offer. We make sure we create one data
offer per device. We now make the offer inert when the source is destroyed.

Fixes the second half of https://github.com/swaywm/wlroots/issues/1041

### DIFF
--- a/include/types/wlr_data_device.h
+++ b/include/types/wlr_data_device.h
@@ -19,6 +19,7 @@ extern const struct wlr_surface_role drag_icon_surface_role;
 struct wlr_data_offer *data_offer_create(struct wl_client *client,
 	struct wlr_data_source *source, uint32_t version);
 void data_offer_update_action(struct wlr_data_offer *offer);
+void data_offer_destroy(struct wlr_data_offer *offer);
 
 struct wlr_client_data_source *client_data_source_create(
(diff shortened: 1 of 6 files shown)


uwqwxing886:
This fixes the Firefox crash when pasting the clipboard contents into the address bar. Is there anything else that should be tested?

Firefox is still laggy but that's an issue with the Firefox wayland port itself.

atomnuker:
Works fine, this is the last thing needed to make firefox with wayland work here.

cg9999:
With this fix, firefox is now perfectly usable. The lag issue has also been fixed by other commits.
Only minor issue left for me seems to be some redraws missed behind closing menus and tooltips.

emersion:
This breaks copy-pasting in gnome-terminal. To reproduce:

```
pkill gnome-terminal
/usr/lib/gnome-terminal-server
gnome-terminal; gnome-terminal
```

And try copy-pasting between the two terminal windows.

emersion:
Fixed, this is ready for review.

atomnuker:
Works fine now.

Issue #1367, opened 2018-11-10
Add support for primary-selection


emersion:
New Wayland protocol, replacing the GTK one. Will be published in a few days in the next wayland-protocols version.

https://patchwork.freedesktop.org/patch/257267/

emersion:
Landed in wayland-protocols 1.17: https://cgit.freedesktop.org/wayland/wayland-protocols/commit/?id=2c3b11d76fd498fc3856cdd36ac76c1193686bf1

emersion:
Once https://github.com/swaywm/wlroots/pull/1397 lands, I'll be happy with gtk-primary-selection's implementation design. I'd like to do https://github.com/swaywm/wlroots/issues/1138 next, and after that we can think of copy-pasting the GTK implementation and rename it for zwp-primary-selection-unstable-v1.

What changes with this protocol is that we now have 3 different protocols for primary selection: GTK, wayland-protocols and X11.

Currently the GTK protocol is used as a central point to synchronize the primary selection. We have two options:

1. Let the compositor synchronize everything. This is similar to how we handle the different shells. This involves calling `set_selection` for the two other protocols when it's called from one protocol.
2. Define a common interface for primary selections. Each protocol can listen when the selection is set, and broadcast its own (that's currently how it works with X11 selection). This has the drawback of assuming the primary selection interface won't change even if we don't have a stable Wayland protocol yet (but this is unlikely because it's a copy of the stable `wl_data_device` interface).

I think going for (2) is worth it. Proposal:

* A `request_set_selection` signal compositors will listen to
* A `wlr_seat_set_primary_selection` function compositors can call
* A `set_selection` signal primary selections can listen to in order to synchronize themselves

PR #1387, opened 2018-11-23
Rename wlr_primary_selection to wlr_gtk_primary_selection

emersion:
Preliminary work for https://github.com/swaywm/wlroots/issues/1367

`wlr_primary_selection` should not be used for the GTK variant of the protocol since it'll cause name clashes.

This PR doesn't contain any functional changes.


emersion:
Ping

ddevault:
Thanks!

PR #1390, opened 2018-11-23
output: add wlr_output_preferred_read_format()

### LINKED ISSUE #1389: GL_INVALID_OPERATION in glGetIntegerv(GL_IMPLEMENTATION_COLOR_READ_FORMAT: no GL_READ_BUFFER)
This happens when using `wlr-screencopy-v1`. How to reproduce (at least on my setup):

1. Open sway/rootston/wayfire (probably all wlroots compositors)
2. Start wf-recorder (https://github.com/ammen99/wf-recorder)
3. Open a dmabuf/EGL client - it happens immediately after I open a Xwayland client, `weston-simple-dmabuf-drm` or `weston-simple-egl`, but `weston-simple-shm` doesn't break it.

The above GL error is printed to the log, which causes a change in the buffer format advertised to the client. This isn't handled by `wf-recorder`, causing it to misbehave (and being force disconnected by the compositor) but that is a bug in wf-recorder. This bug is about the GL errors in `wlroots`

My guess is that we need to properly bind the framebuffer when importing a dmabuf, or maybe we need to reset some other state.

ammen99:
The read format is dependent on the output, so we first need to make it
current. This fixes a race condition in wlr-screencopy-v1 where a dmabuf
client would cause EGL_NO_SURFACE to be bound at the time when
screencopy needs to query for the preferred format, causing GL errors.

Fixes #1389

### DIFF
--- a/include/wlr/render/wlr_renderer.h
+++ b/include/wlr/render/wlr_renderer.h
@@ -97,11 +97,6 @@ int wlr_renderer_get_dmabuf_formats(struct wlr_renderer *renderer,
 int wlr_renderer_get_dmabuf_modifiers(struct wlr_renderer *renderer, int format,
 	uint64_t **modifiers);
 /**
- * Get the preferred format for reading pixels.
- */
-bool wlr_renderer_preferred_read_format(struct wlr_renderer *renderer,
-	enum wl_shm_format *fmt);
-/**
  * Reads out of pixels of the currently bound surface into data. `stride` is in
  * bytes.
  *
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -193,6 +193,12 @@ void wlr_output_effective_resolution(struct wlr_output *output,
  */
 bool wlr_output_make_current(struct wlr_output *output, int *buffer_age);
 /**
+ * Get the preferred format for reading pixels.
+ * This function might change the current rendering context.
+ */
+bool wlr_output_preferred_read_format(struct wlr_output *output,
+	enum wl_shm_format *fmt);
+/**
  * Swaps the output buffers. If the time of the frame isn't known, set `when` to
  * NULL. If the compositor doesn't support damage tracking, set `damage` to
  * NULL.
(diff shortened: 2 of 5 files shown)


include/wlr/types/wlr_output.h
```diff
@@ -192,6 +192,12 @@ void wlr_output_effective_resolution(struct wlr_output *output,
  * unknown. This is useful for damage tracking.
  */
 bool wlr_output_make_current(struct wlr_output *output, int *buffer_age);
+/**
+ * Get the preferred format for reading pixels.
+ * This also makes the output rendering context current.
```
emersion: Hmm, I don't want users to start relying on this implementation detail, so I'd rather not document it, or say something like "this function can change the current rendering context".

include/wlr/types/wlr_output.h
```diff
@@ -192,6 +192,12 @@ void wlr_output_effective_resolution(struct wlr_output *output,
  * unknown. This is useful for damage tracking.
  */
 bool wlr_output_make_current(struct wlr_output *output, int *buffer_age);
+/**
+ * Get the preferred format for reading pixels.
+ * This function might change the current rendering context.
+ */
+bool wlr_output_preferred_read_format(struct wlr_output *output,
+	enum wl_shm_format *fmt);
```
emersion: This should be removed

include/wlr/render/wlr_renderer.h
```diff
@@ -98,6 +98,8 @@ int wlr_renderer_get_dmabuf_modifiers(struct wlr_renderer *renderer, int format,
 	uint64_t **modifiers);
 /**
  * Get the preferred format for reading pixels.
+ * The caller needs to ensure the appropriate rendering context has been loaded,
+ * for ex. by calling wlr_output_make_current()
  */
 bool wlr_renderer_preferred_read_format(struct wlr_renderer *renderer,
```
emersion: This should be removed

ammen99: oops, I've forgotten it.

emersion:
Thanks!

Issue #1388, opened 2018-11-23
Don't pollute wlr_seat with primary selection stuff


emersion:
`wlr_seat_client` has a `primary_selection_devices` field. This isn't a good idea because this will conflict with future primary-selection protocols (a new unstable version has landed in wayland-protocols).

We don't want to add references to unstable protocols in stable protocols. Instead we should track per-seat state in primary-selection data structures directly (just like the tablet stuff).

emersion:
Fixed in https://github.com/swaywm/wlroots/pull/1397

Issue #1383, opened 2018-11-21
Make data offers inert when the source changes


emersion:
This applies to:

- [ ] `wl_data_device` selection offers
- [x] `wl_data_device` drag-and-drop offers
- [x] Primary selection offers (https://github.com/swaywm/wlroots/pull/1397)

PR #1397, opened 2018-11-27
gtk-primary-selection: refactor everything, untie from seat

emersion:
This PR completely refactors wlr_gtk_primary_selection. The goal is to
remove gtk-primary-selection state from the seat and better handle inert
resources where it makes sense.

wlr_seat_client.primary_selection_devices has been removed and replaced by
wlr_gtk_primary_selection_device. This allows us to support multiple
primary-selection protocols and removes unstable protocols state from wlr_seat.

wlr_seat_set_primary_selection has been refactored to be a standalone interface
that can be used by all primary-selection protocols (GTK, wayland-protocols and
Xwayland). A common wlr_primary_selection_source has been created.

wlr_seat_set_primary_selection doesn't take a serial anymore, serial validation
is expected to be done by each protocol. It makes it easier for Xwayland and
compositors to set the current selection.

Devices are now made inert when the seat goes away.

See https://github.com/swaywm/wlroots/issues/1388
See https://github.com/swaywm/wlroots/issues/1383


emersion:
I'll run this locally during a few days to make sure I haven't missed anything during testing.

emersion:
Haven't hit any regression, this is ready for review.

ddevault:
Thanks!

PR #1402, opened 2018-11-29
data-device, primary-selection: add request_set_selection

### LINKED ISSUE #1138: Compositor cannot customize the wl_data_device.set_selection policy
The handler directly calls `wlr_seat_set_selection`, without giving the compositor a chance to accept or reject the request.

We should do something similar to `wl_pointer.set_cursor` which has a `request_set_cursor` event.

Same applies to `start_drag`.

emersion:
~~Depends on https://github.com/swaywm/wlroots/pull/1397~~
~~Depends on https://github.com/swaywm/wlroots/pull/1485~~

This makes compositors able to block and/or customize set_selection requests
coming from clients. For instance, it's possible for a compositor to disable
rich selection content (by removing all MIME types except text/plain). This
commit implements the design proposed in [1].

Two new events are added to wlr_seat: request_set_selection and
request_set_primary_selection. Compositors need to listen to these events and
either destroy the source or effectively set the selection.

Fixes #1138

[1]: https://github.com/swaywm/wlroots/issues/1367#issuecomment-442403454

### DIFF
--- a/include/rootston/seat.h
+++ b/include/rootston/seat.h
@@ -38,6 +38,8 @@ struct roots_seat {
 	struct wl_list tablets;
 	struct wl_list tablet_pads;
 
+	struct wl_listener request_set_selection;
+	struct wl_listener request_set_primary_selection;
 	struct wl_listener new_drag_icon;
 	struct wl_listener destroy;
 };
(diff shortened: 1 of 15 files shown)


emersion:
This is now ready for review.

PR #1405, opened 2018-12-02
xwayland: fix Firefox/GTK3/Wayland -> Xwayland copy-paste not working

jnpkrn:
Firefox in such a setting doesn't offer "text/plain;charset=utf-8",
but rather something that equals to "UTF8_STRING" directly in the
MIME matching logic of Xwayland part, and it shall rather be preserved
as-is for this to work towards X programs (rxvt-unicode, for instance).
That leads to prioritizing xwm_get_atom_name over a static translation.

It seems X11 atoms vs. proper MIME types is an ongoing topic,
see https://gitlab.gnome.org/GNOME/gtk/issues/1177
("Make X11 run transformation steps mime-type <> atom").

### DIFF
--- a/xwayland/selection/outgoing.c
+++ b/xwayland/selection/outgoing.c
@@ -260,20 +260,32 @@ static void xwm_selection_send_data(struct wlr_xwm_selection *selection,
 		return;
 	}
 
-	bool found = false;
-	char **mime_type_ptr;
+	bool found = false, found_pseudo = false;
+	const char **mime_type_ptr;
+	const char *pseudo_mime = "";  // assume never matching any content type
+
+	if (strcmp(mime_type, "text/plain;charset=utf-8") == 0) {
+		pseudo_mime = "UTF8_STRING";
+	} else if (strcmp(mime_type, "text/plain") == 0) {
+		pseudo_mime = "TEXT";
+	}
+
 	wl_array_for_each(mime_type_ptr, mime_types) {
-		char *t = *mime_type_ptr;
+		const char *t = *mime_type_ptr;
 		if (strcmp(t, mime_type) == 0) {
 			found = true;
 			break;
+		} else if (strcmp(t, pseudo_mime) == 0) {
+			found_pseudo = true;
 		}
 	}
-	if (!found) {
+	if (!found && !found_pseudo) {
 		wlr_log(WLR_ERROR, "not sending selection: "
 			"requested an unsupported MIME type %s", mime_type);
 		xwm_selection_send_notify(selection->xwm, req, false);
 		return;
+	} else if (!found) {
+		mime_type_ptr = &pseudo_mime;
 	}
 
 	struct wlr_xwm_selection_transfer *transfer =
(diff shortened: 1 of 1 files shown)


jnpkrn:
Will report back if I'll hit any regression.

Btw. I find this handy for a very basic troubleshooting:
https://www.virtualbox.org/attachment/ticket/3755/viewclipformats.cpp

jnpkrn:
OK, I'll have a stab at it.

jnpkrn:
Did you mean something like this?

This WFM, though I will keep an eye if there's indeed no regression.
In positive case, I need to add a documentation of i/o handling
in `xwm_mime_map_fnc`.

emersion:
Ugh, this seems a lot more complicated than it needs to be. Can you try simplifying your changes? I don't think we need an abstraction with callbacks and methods for two special cases.

jnpkrn:
Agreed, it was overcomplicated, in part because I lack the wider view,
e.g. are there are other areas to receive such handling beside plain
text (the "complicated variant" would make for a nice chaining), or
is there any configurable adjustment to be hooked in eventually
(yeah, doesn't look so) etc.

jnpkrn:
One of the other possible simplifications would be to move
the remaining parts of logic in `xwm_mime_type_to_atom` and
`xwm_mime_type_from_atom` (secondary lookups) directly into
`xwm_mime_map`, which might make sense given the particular
mappings are only used in those contexts.

I.e., instead of two functions with scattered mapping redundancy,
there would be a single, internally overloaded (but still hopefully
not overblown) function centralized around the single mapping and
playing an extra role of one-way fallback selector.

That would be a 3rd commit that I can add for assessment whether
that would make this overall approach a viable solution.

emersion:
I'd prefer to stick with a simple callback-less approach. Your first iteration with my first comment addressed would be fine.

ddevault:
Can you rebase this to get the FreeBSD fixes?

jnpkrn:
Done (in both cases).  Thanks for looking.

jnpkrn:
Strangest of all, the patch seems no longer needed since
Firefox works even without it?  Colour me baffled.

jnpkrn:
FTR. the externally observable difference using `viewclipformat` utility
from above is this:

```
--- f63.txt	2018-12-21 22:38:19.074189175 +0100
+++ f64.txt	2018-12-21 22:38:17.979199018 +0100
@@ -1,11 +1,13 @@
 TIMESTAMP
 TARGETS
 text/html
 text/_moz_htmlcontext
 text/_moz_htmlinfo
 UTF8_STRING
 COMPOUND_TEXT
 TEXT
 STRING
+UTF8_STRING
+TEXT
 text/x-moz-url-priv
 SAVE_TARGETS
```

No idea why the former xwayland code would not handle gracefully
the former and seems to be doing so with the latter.

Anyway, with this fix (in either form, see above the alternative),
the things are handled gracefully in both cases.

jnpkrn:
I am withdrawing this for the time being, it might have been just some
glitch in Firefox Wayland integration prior to v64 (and/or in the gtk
binding it used, not sure if from distro or some embedded version).

What I _do_ suffer from now (Firefox 66, updated released versions of
sway/wlroots), though, is that incorrect new-line endings get copied
in addition to the xwayland programs (e.g. gvim, or urxvt):

- expected: `\n`
- obtained during paste: `\r\n`

That is just for the record.

emersion:
>incorrect new-line endings

As we don't modify the data in any way, this is a Firefox bug.

Issue #1407, opened 2018-12-03
Fire the xdg-surface destroy signal when the role object is destroyed


emersion:
Right now, the lifecycle of a `wlr_xdg_surface` object is as follows:

* Client creates a xdg-surface, we allocate a `wlr_xdg_surface` but the compositor doesn't know about it
* Client creates a xdg-role-object (e.g. xdg-toplevel). We allocate e.g. a `wlr_xdg_toplevel`, set the xdg-role and fire the `new_xdg_surface` event.
* Client maps the xdg-surface, we emit the `map` event. Client can map/unmap at will.
* Client destroys the xdg-role-object, this unmaps the surface. We destroy the `wlr_xdg_toplevel`, reset the xdg-role and do nothing else.
* Client destroys the xdg-surface, we fire the `destroy` event on `wlr_xdg_surface`

Right after the client destroys the xdg-role-object, the compositor is left with a weird xdg-surface: its xdg-role is set to none. Various xdg-shell functions that operate on the role won't work anymore because the xdg-toplevel is gone.

Instead, we could emit `destroy` when the xdg-role-object is destroyed, and re-emit `new_xdg_surface` if the client re-creates it again. This is more consistent with the initial xdg-toplevel creation case.

PR #1410, opened 2018-12-04
xdg-shell: emit xdg_surface destroy when role object is destroyed

emersion:
Fixes https://github.com/swaywm/wlroots/issues/1407


ddevault:
Thanks!

PR #1412, opened 2018-12-04
backend/drm: fix VLA overflow

emersion:
wlroots segfaults on aarch64 because of this. Found the cause thanks to ASAN.

We were using a CRTC index instead of an output index.


ddevault:
Thanks!

PR #1413, opened 2018-12-04
backend/session/logind: improve logging

emersion:
Will maybe help with https://github.com/swaywm/sway/issues/3136

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -51,8 +51,8 @@ static struct logind_session *logind_session_from_session(
 static int logind_take_device(struct wlr_session *base, const char *path) {
 	struct logind_session *session = logind_session_from_session(base);
 
-	int ret;
 	int fd = -1;
+	int ret;
 	sd_bus_message *msg = NULL;
 	sd_bus_error error = SD_BUS_ERROR_NULL;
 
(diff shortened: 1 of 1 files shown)


backend/session/logind.c
```diff
@@ -94,27 +95,27 @@ static int logind_take_device(struct wlr_session *base, const char *path) {
 error:
 	sd_bus_error_free(&error);
 	sd_bus_message_unref(msg);
-	return fd;
+	return -1;
```
kragniz: why did you change this to always return -1?

backend/session/logind.c
```diff
@@ -70,22 +69,24 @@ static int logind_take_device(struct wlr_session *base, const char *path) {
...
 		goto error;
 	}
 
+	int fd = -1;
 	int paused = 0;
 	ret = sd_bus_message_read(msg, "hb", &fd, &paused);
 	if (ret < 0) {
 		wlr_log(WLR_ERROR, "Failed to parse D-Bus response for '%s': %s",
-			path, strerror(-ret));
+			path, error.message);
```
progandy: ```error.message``` is only set by ```sd_bus_call_method```. Why would you use it here?

backend/session/logind.c
progandy: @ddevault It is not a pure error branch, but a *finally* branch.

emersion: Oh right. I'll s/error/out/ because it's confusing.

backend/session/logind.c
emersion: Good catch

backend/session/logind.c
```diff
@@ -70,28 +69,30 @@ static int logind_take_device(struct wlr_session *base, const char *path) {
 		session->path, "org.freedesktop.login1.Session", "TakeDevice",
 		&error, &msg, "uu", major(st.st_rdev), minor(st.st_rdev));
 	if (ret < 0) {
-		wlr_log(WLR_ERROR, "Failed to take device '%s': %s", path, error.message);
-		goto error;
+		wlr_log(WLR_ERROR, "Failed to take device '%s': %s", path,
+			error.message);
+		goto out;
```
progandy: ```fd``` is not defined if this goto is triggered, but it is needed for the return.

emersion: Yeah, just pushed a fix

ddevault:
Thanks!

Issue #1408, opened 2018-12-03
Rootston crash on when client killed with Layer Shell popup


wmww:
To reproduce: Open a layer shell popup, then kill the client with ctrl+C. Rootston crashes because:

First, it destroys the parent layer surface:
```
Thread 1 "rootston" hit Breakpoint 1, 0x00007ffff7915f20 in free () from /home/wmww/code/wlroots/build/rootston/../../../../../../lib64/libc.so.6
#0  0x00007ffff7915f20 in free () at /home/wmww/code/wlroots/build/rootston/../../../../../../lib64/libc.so.6
#1  0x0000555555565049 in handle_destroy (listener=0x55555589c618, data=0x5555558bc4e0) at ../rootston/layer_shell.c:336
#2  0x00007ffff7f96ec6 in wlr_signal_emit_safe (signal=0x5555558bc600, data=0x5555558bc4e0) at ../util/signal.c:29
#3  0x00007ffff7f810d2 in layer_surface_destroy (surface=0x5555558bc4e0) at ../types/wlr_layer_shell_v1.c:182
#4  0x00007ffff7f81609 in handle_surface_destroyed (listener=0x5555558bc5e8, data=0x555555918ff0) at ../types/wlr_layer_shell_v1.c:319
#5  0x00007ffff7f96ec6 in wlr_signal_emit_safe (signal=0x5555559192b0, data=0x555555918ff0) at ../util/signal.c:29
#6  0x00007ffff7f8ee59 in   (resource=0x555555919310) at ../types/wlr_surface.c:584
#7  0x00007ffff7f0ec7f in destroy_resource (element=0x555555919310, data=<optimized out>, flags=0) at src/wayland-server.c:688
#8  0x00007ffff7f13052 in for_each_helper (func=func@entry=0x7ffff7f0ec10 <destroy_resource>, data=data@entry=0x7fffffffe314, entries=<optimized out>, entries=<optimized out>) at src/wayland-util.c:372
#9  0x00007ffff7f1355f in wl_map_for_each (map=0x55555591a7a0, func=0x7ffff7f0ec10 <destroy_resource>, data=0x7fffffffe314) at src/wayland-util.c:385
#10 0x00007ffff7f0edff in wl_client_destroy (client=0x55555591a770) at src/wayland-server.c:847
#11 0x00007ffff7f0eedb in destroy_client_with_error (reason=<optimized out>, client=<optimized out>) at src/wayland-server.c:307
[...]

PR #1418, opened 2018-12-05
layer-shell-v1, xdg-shell: destroy popups on layer shell destroy

emersion:
This commit makes it so that wlr_xdg_surface_send_close only sends the "done"
event once to popup objects, so we can call it multiple times without issues.

A new function wlr_xdg_popup_destroy destroys popup by closing them and then
making them inert.

@wmww: can you test this?

Fixes https://github.com/swaywm/wlroots/issues/1408


wmww:
Thanks for the quick response. Unfortunately, that doesn't fix the issue. When I get a chance, I'll dig into what its trying to do, and why it doesn't work.

emersion:
It's kind of hard to debug for me, because I can't reproduce.

panaman67:
This should be ok with a rebase correct?

PR #1420, opened 2018-12-06
wlr_keyboard: add xkb_state parameter to wlr_keyboard_set_layout

Duncaen:
This allows the compositor to share the keyboard state between multiple
keyboards with the same layout.


ddevault:
I'm not sure this justifies the breaking change. Is there something to gain here other than a nice-to-have?

Duncaen:
I'm closing this PR, the change is only useful in very few cases and a change as proposed in https://github.com/swaywm/wlroots/issues/1421 would be better in general.

PR #1426, opened 2018-12-09
backend/drm: fix disappeared output indices

emersion:
This commit changes `scan_drm_connectors` to add new outputs to the end of the
list. That way, it's easier to understand what's going on with indices.

When we need to destroy outputs, we now walk the list in reverse order. This
ensures indices remain correct while iterating and removing items from the
list.

We now also make outputs without a CRTC disappear (those are in
WLR_DRM_CONN_NEEDS_MODESET state).

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -993,11 +993,10 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
 
 static uint32_t get_possible_crtcs(int fd, drmModeRes *res,
 		drmModeConnector *conn, bool is_mst) {
-	drmModeEncoder *enc;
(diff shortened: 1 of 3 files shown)


ascent12:
I've pushed a change which fixes the no encoder thing (well, it works for me at least).
I suppose we should wait for @jeremejevs or someone else to also report if the fix works.

emersion:
As a side note, Weston (and Mutter) seem to do things [quite differently](https://gitlab.freedesktop.org/wayland/weston/blob/master/libweston/compositor-drm.c#L5830). They only pick a CRTC that works on *all* encoders for the connector. If they don't find one, they just pick one at random and hope it works.

ddevault:
Thanks!

PR #1429, opened 2018-12-09
gtk-primary-selection: basic serial validation

emersion:
I've discovered that some clients _rely_ on the compositor for rejecting `set_selection` requests if the serial is invalid. For instance Gedit under Wayland issues multiple `set_selection` requests with the same serial, and not doing this basic serial validation will break clipboard.

While I haven't found any clients broken for primary selection, it's probably better to add back this basic serial validation, at least until we have a new improved validation mechanism.


ddevault:
Thanks!

PR #1432, opened 2018-12-09
Relative pointers

ForTheReallys:
I've taken @arandomhuman 's pull request and rebased it here.  I did run into the same issue he did though.  When running on the x11 or wayland backend, the cursor uses absolute events which causes the relative pointer to misbehave depending on the cursor's position.  Let me know if you have any suggestions.


ascent12:
>which causes the relative pointer to misbehave depending on the cursor's position
How exactly? Is that to do with a cursor leaving the surface and then going back in?

ForTheReallys:
Yes that was one of the issues.  I did a bit of digging and it looks like the problem is nested pointer constraints.  When nesting compositors, the inner compositor is locking the cursor but the outer one is not.  That sounds like it should be another pull request to me though.

ascent12:
Yeah, I think that's not worth worrying about too much right now.

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A wp_relative_pointer object is an extension to the wl_pointer interface
+ * used for emitting relative pointer events. It shares the same focus as
+ * wl_pointer objects of the same seat and will only emit events when it has
+ * focus.
+ */
+
+struct wlr_relative_pointer_v1 {
+	struct wl_client *client;
```
emersion: No need to have this field. Instead, `wl_resource_get_client` can be used.

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,77 @@
...
+	struct wl_resource *resource;
+	struct wl_resource *pointer;
```
emersion: This field exposes a generic resource. No need for this. Instead, this struct should have a reference to the seat.

include/wlr/types/wlr_seat.h
```diff
@@ -27,6 +27,7 @@ struct wlr_seat_client {
 	// lists of wl_resource
 	struct wl_list resources;
 	struct wl_list pointers;
+	struct wl_list relative_pointers_v1;
```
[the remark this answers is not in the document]
ddevault: +1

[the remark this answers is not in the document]
emersion: Nah, no need for this. We need to have a per-seat structure for relative pointers. There are other examples of this in wlroots, for instance gtk-primary-selection (https://github.com/swaywm/wlroots/blob/master/include/wlr/types/wlr_gtk_primary_selection.h#L29) and tablet-v2.

include/wlr/types/wlr_relative_pointer_v1.h
ForTheReallys: I think we actually need *pointer.  We need to send the frame to the client.  Is there a way to know this pointer from the seat?

include/wlr/types/wlr_seat.h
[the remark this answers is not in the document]
ForTheReallys: I just made moved the list into the manager and out of the seat.  Let me know what you think.

include/wlr/types/wlr_relative_pointer_v1.h
emersion: Yes. Something like `seat.pointer_state.resource`.

include/wlr/types/wlr_seat.h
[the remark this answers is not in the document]
emersion: No, this isn't enough. When an event happens on one seat it sends it to all seats' relative pointer resources.

include/wlr/types/wlr_relative_pointer_v1.h
ForTheReallys: Oh ok I found it in seat_client.  I didn't realize sending events sent frames to all pointers in its list.

types/seat/wlr_seat_pointer.c
```diff
@@ -5,6 +5,7 @@
 #include <time.h>
 #include <wayland-server.h>
 #include <wlr/types/wlr_input_device.h>
+#include <wlr/types/wlr_relative_pointer_v1.h>
```
emersion: This doesn't seem needed.

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,78 @@
...
+#ifndef WLR_TYPES_WLR_RELATIVE_POINTER_V1_H
+#define WLR_TYPES_WLR_RELATIVE_POINTER_V1_H
+
+#include <wayland-server.h>
+
+
+/**
+ * This protocol specifies a set of interfaces used for making clients able to
+ * receive relative pointer events not obstructed by barriers (such as the
+ * monitor edge or other pointer barriers).
```
emersion: "or pointer constraints"

include/wlr/types/wlr_seat.h
```diff
@@ -313,6 +313,19 @@ void wlr_seat_pointer_clear_focus(struct wlr_seat *wlr_seat);
...
+/**
+ * Send relative motion events to the surface with pointer focus. Coordinates
+ * for the motion event are relative to current pointer location, both
+ * accelerated and unaccelerated. Compositors should use
+ * `wlr_seat_pointer_notify_relative_motion()` to send relative motion events
+ * to respect relative pointer requests by clients.
+ *
+ * Note that the timestamp is 64 bit, split into high 32 bits and low 32 bits.
+ */
+void wlr_seat_pointer_notify_relative_motion(struct wlr_seat *wlr_seat,
```
emersion: This has been replaced with `wlr_relative_pointer_v1_send_relative_motion`

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,78 @@
...
+};
+
+struct wlr_relative_pointer_manager_v1 *wlr_relative_pointer_manager_v1_create(
+	struct wl_display *display);
+void wlr_relative_pointer_manager_v1_destroy(
+	struct wlr_relative_pointer_manager_v1 *relative_pointer_manager);
+
+void wlr_relative_pointer_v1_send_relative_motion(
+	struct wlr_relative_pointer_v1 *relative_pointer, uint64_t time,
+	double dx, double dy, double dx_unaccel, double dy_unaccel);
```
emersion: A doc comment would be nice, at least to say `time` is in milliseconds (`time` could also be renamed `time_msec` to make this clear).

rootston/cursor.c
```diff
@@ -305,11 +305,47 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
...
+		seat->input->server->desktop->relative_pointer_manager;
+
+	struct wlr_seat_client *client = seat->seat->pointer_state.focused_client;
+	if (client == NULL) {
+		return;
+	}
+
+	struct wl_resource *resource;
+	wl_resource_for_each(resource, &relative_pointer_manager->relative_pointers) {
+
```
emersion: Style: extra newline

rootston/cursor.c
```diff
@@ -305,11 +305,47 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
 	}
 }
 
+static void notify_relative_motion(struct roots_seat *seat, uint64_t time_msec,
+		double dx, double dy, double dx_unaccel, double dy_unaccel) {
+
```
emersion: Style: extra newline

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * A global interface used for getting the relative pointer object for a given
+ * pointer.
+ */
+
+struct wlr_relative_pointer_manager_v1 {
+	struct wl_global *global;
+	struct wl_list resources;
+	struct wl_list relative_pointers;
```
emersion: This should be a list of `wlr_relative_pointer_v1`, not a list of resources.

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,78 @@
...
+
+
+/**
+ * A global interface used for getting the relative pointer object for a given
+ * pointer.
+ */
+
+struct wlr_relative_pointer_manager_v1 {
+	struct wl_global *global;
+	struct wl_list resources;
```
emersion: A little comment `// wl_resource_get_link` making it clear it's a list of resources would be nice

rootston/cursor.c
```diff
@@ -305,11 +305,47 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
...
+		wlr_relative_pointer_v1_send_relative_motion(pointer, time_msec,
+				dx, dy, dx_unaccel, dy_unaccel);
+	}
+
+	wl_resource_for_each(resource, &client->pointers) {
+		if (wlr_seat_client_from_pointer_resource(resource) == NULL) {
+			continue;
+		}
+		wl_pointer_send_frame(resource);
```
emersion: This should not be done in rootston. This should probably be moved in `wlr_relative_pointer_v1_send_relative_motion`.

rootston/cursor.c
```diff
@@ -305,11 +305,47 @@ static void roots_cursor_press_button(struct roots_cursor *cursor,
...
+
+}
+
 void roots_cursor_handle_motion(struct roots_cursor *cursor,
 		struct wlr_event_pointer_motion *event) {
 	double dx = event->delta_x;
 	double dy = event->delta_y;
 
+	notify_relative_motion(cursor->seat,
+		(uint64_t)event->time_msec * 1000, dx, dy, dx, dy);
```
emersion: This doesn't send unaccelerated values. I wonder if we can get those from libinput?

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,234 @@
...
+		double dx_unaccel, double dy_unaccel) {
+	zwp_relative_pointer_v1_send_relative_motion(relative_pointer->resource,
+		(uint32_t)(time >> 32), (uint32_t)time,
+		wl_fixed_from_double(dx), wl_fixed_from_double(dy),
+		wl_fixed_from_double(dx_unaccel), wl_fixed_from_double(dy_unaccel));
+}
+
+
+struct wlr_relative_pointer_v1 *wlr_relative_pointer_v1_from_resource(struct wl_resource *resource) {
+	return relative_pointer_from_resource(resource);
```
emersion: We can just rename `relative_pointer_from_resource` to `wlr_relative_pointer_v1_from_resource`

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,234 @@
...
+	relative_pointer_manager->display_destroy_listener.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &relative_pointer_manager->display_destroy_listener);
+
+	wlr_log(WLR_DEBUG, "relative_pointer_v1 manager created");
+
+	return relative_pointer_manager;
+}
+
+
+void wlr_relative_pointer_manager_v1_destroy(struct wlr_relative_pointer_manager_v1 *relative_pointer_manager) {
```
emersion: Style: `relative_pointer_manager` could just be named `manager`, it's clear what we're talking about since we're in `wlr_relative_pointer_v1.c`.

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,234 @@
...
+	wlr_log(WLR_DEBUG, "relative_pointer_v1 manager unbound from client %p",
+		client);
+}
+
+
+static void relative_pointer_manager_v1_handle_get_relative_pointer(struct wl_client *client,
+		struct wl_resource *resource, uint32_t id, struct wl_resource *pointer) {
+	struct wlr_seat_client *seat_client =
+		wlr_seat_client_from_pointer_resource(pointer);
+	assert(seat_client->client == client);
```
emersion: I don't think this is necessary.

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,234 @@
...
+	wl_resource_destroy(resource);
+
+	wlr_log(WLR_DEBUG, "relative_pointer_v1 manager unbound from client %p",
+		client);
+}
+
+
+static void relative_pointer_manager_v1_handle_get_relative_pointer(struct wl_client *client,
+		struct wl_resource *resource, uint32_t id, struct wl_resource *pointer) {
+	struct wlr_seat_client *seat_client =
```
emersion: We should listen for the seat being destroyed, and make the resource inert (see CONTRIBUTING.md) when this happens. Otherwise we'll have a dangling invalid pointer to a `wlr_seat`.

rootston/cursor.c
ForTheReallys: We can but I think we'd have to add to add those fields to the wlr_event_pointer_motion struct.

types/wlr_relative_pointer_v1.c
ForTheReallys: Should we also listen for the seat_client being destroyed as well?

emersion: I don't think so, because it's fine to create a `wl_seat`, create a `wl_pointer`, create a relative pointer, and then destroy the seat and continue to use both pointers.

rootston/cursor.c
emersion: Yeah, that would be nice.

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,261 @@
...
+		(uint32_t)(time >> 32), (uint32_t)time,
+		wl_fixed_from_double(dx), wl_fixed_from_double(dy),
+		wl_fixed_from_double(dx_unaccel), wl_fixed_from_double(dy_unaccel));
+
+	struct wl_resource *resource;
+	wl_resource_for_each(resource, &client->pointers) {
+		if (wlr_seat_client_from_pointer_resource(resource) == NULL) {
+			continue;
+		}
+		wl_pointer_send_frame(resource);
```
emersion: Hmm. Thought about this for a while, and I think we should only send frame events to the `wl_pointer` provided in `get_relative_pointer`. So in the end we'd need to save it somewhere and use it here.

If this is the case we should make sure we make the relative pointer inert when the `wl_pointer` is destroyed. We can listen to such destroy events with `wl_resource_add_destroy_listener`.

Does this make sense? Sorry for asking to move things around multiple times.

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,261 @@
...
+	wl_resource_destroy(resource);
+}
+
+static void relative_pointer_handle_seat_destroy(struct wl_listener *listener,
+		void *data) {
+	struct wlr_relative_pointer_v1 *relative_pointer =
+		wl_container_of(listener, relative_pointer, seat_destroy);
+
+	relative_pointer_destroy(relative_pointer);
+	wl_resource_set_user_data(relative_pointer->resource, NULL);
```
emersion: This is a use-after-free.

emersion: Actually, this instruction (which makes the resource inert) should be moved in `relative_pointer_destroy`.

include/wlr/types/wlr_relative_pointer_v1.h
emersion: Can you add a little comment `// wlr_relative_pointer_v1::link` to make it clear it's a list of these?

include/wlr/types/wlr_relative_pointer_v1.h
```diff
@@ -0,0 +1,80 @@
...
+
+struct wlr_relative_pointer_manager_v1 {
+	struct wl_global *global;
+	struct wl_list resources; // wl_resource_get_link()
+	struct wl_list relative_pointers;
+
+	struct {
+		struct wl_signal destroy;
+		/**
+		 * returns struct wlr_relative_pointer_v1 *relative_pointer
```
emersion: Hmm, it doesn't really "return" a value. We generally just add `// wlr_relative_pointer_v1` on the same line as the event.

types/wlr_relative_pointer_v1.c
ForTheReallys: Oh shoot you're right.  I can't believe I didn't see that.  Will do

types/wlr_relative_pointer_v1.c
ForTheReallys: Yep that makes sense.  No problem.  I'm just glad we're making progress

include/wlr/types/wlr_relative_pointer_v1.h
emersion: Bump

types/wlr_relative_pointer_v1.c
```diff
@@ -0,0 +1,253 @@
...
+	wl_resource_set_implementation(relative_pointer_resource, &relative_pointer_v1_impl,
+		relative_pointer, relative_pointer_v1_handle_resource_destroy);
+
+	struct wlr_relative_pointer_manager_v1 *relative_pointer_manager =
+		relative_pointer_manager_from_resource(resource);
+
+	wl_list_insert(&relative_pointer_manager->relative_pointers,
+			&relative_pointer->link);
+
+	wl_signal_add(&relative_pointer->seat->events.destroy,
```
emersion: You can't add a single listener for multiple signals. You need two listeners.

ddevault:
@sghctoma can you take a look at this FreeBSD issue?

sghctoma:
At first glance, this is caused by today's libinput-related updates to the port tree; I think moving the v4l_compat remove+evdev_proto install to the beginning of the ports-build phase in .builds/freebsd.yml would solve the problem. I don't have my laptop with me right now, but will check this later.

On a related note, 12.0 is out, and it seems that libinput is finally getting updated too, which means I will probably be able to make the build manifest much simpler/cleaner in the coming days.

emersion:
Tried it, didn't help: https://builds.sr.ht/~emersion/job/17108

Adding FreeBSD 12 support to builds.sr.ht is not yet done. It's on my todo-list, though.

sghctoma:
You did not move the v4l_compat remove+evdev_proto steps to the beginning of ports-build, but removed them :) Here is a successful build: https://builds.sr.ht/~sghctoma/job/17236
I have made a PR with the necessary change.

I will look into the qcow2 image builder script in the next few days.

emersion:
Ah, thanks, I misunderstood.

I tried implementing the qcow2 image builder, comments welcome:

https://lists.sr.ht/%7Esircmpwn/sr.ht-dev/%3Cp1KeOGPsQeLC1Fbioq-bm5_PWG1E53KZVx-LhAHZsKD30OxGdT-INL5Lc1BAcihRCujAyuL--RRMRLkkQGoZ86Z5mY1qQHhu4nQrTN9IGuk%3D%40emersion.fr%3E#%3CekoLq15gPoxAa0Vyi2O4ci4n2_Kqy2Vgu-6OHJPxNrm2cFbVZ2AWLmk2ITV9nVn8YhY39ztV0eUGT4coN5GKj0gh6AykwjfPOiXgQzS4WfM=@emersion.fr%3E

emersion:
@ForTheReallys Can you rebase this PR?

include/wlr/types/wlr_relative_pointer_v1.h
emersion: Bump

types/wlr_relative_pointer_v1.c
emersion: Bump

include/wlr/types/wlr_relative_pointer_v1.h
emersion: Bump

types/wlr_relative_pointer_v1.c
ForTheReallys: Sorry about that.  I thought I did this with my last commit.

emersion:
Can you have a look at the comments: https://github.com/swaywm/wlroots/pull/1432/files

There are a bunch of them that are still relevant.

ForTheReallys:
Sure thing. I'll look at them later today.

ForTheReallys:
Sorry for responding so late, the holidays have been busy.  I think I've covered everything now.  Let me know if there's anything else.  Thanks!

emersion:
Thanks!

PR #1433, opened 2018-12-12
xdg-popup: fix wlr_xdg_popup_get_toplevel_coords()

### LINKED ISSUE #1425: Add support for unconstraining of layer-shell popups
Currently the logic for unconstraining xdg-shell popups depends on the fact that the popup's topmost parent is an `xdg_shell_surface` which of course is wrong when the popup has been created on a layer-shell surface, resulting in a crash.

ammen99:
The previous code resulted in a crash when we try to unconstrain a popup
which is part of a layer-shell surface popup tree.

Fixes #1425

@emersion is this what you imagined?


ddevault:
I'm okay with not renaming this, not a big deal imo.

PR #1434, opened 2018-12-12
xwayland: remove clearenv() call

martinetd:
This has been causing troubles for some of our users and only been there
for legacy reasons, we trust Xwayland just as much as your next program
and weston doesn't take any such care when starting it.


martinetd:
No, we need that for setenv that's still used in this file - I had tried

martinetd:
Well we can get away with a less inclusive define I guess, will set _POSIX_C_SOURCE to 200112L

emersion:
Thanks!

PR #1439, opened 2018-12-20
rootston: create foreign toplevel handle before sending outputs

ammen99:
Otherwise the initial list of outputs isn't sent to foreign-toplevel clients.


emersion:
This shouldn't matter, since the foreign-toplevel-mgmt implementation sends the list of current outputs, right?

emersion:
Thanks!

PR #1440, opened 2018-12-21
Fix FreeBSD build manifest

sghctoma:
Recent libinput-related changes in the ports tree made it necessary to
install devel/evdev-proto before x11/libinput (see the comments in #1432).


emersion:
Thanks!

PR #1443, opened 2018-12-22
Protect against redefining MESA_EGL_NO_X11_HEADERS

Hjdskes:
When building wlroots for an embedded Linux (hobby) project with Wayland only, I run into errors where `MESA_EGL_NO_X11_HEADERS` is redefined several times.

This PR simply adds a check if it is already defined, before defining it.


emersion:
Thanks!

PR #1444, opened 2018-12-23
backend/drm: get physical size from libdrm

emersion:
Values from libdrm are likely more reliable than raw values from the EDID. We
were already using values from libdrm, but they were overwritten by parse_edid.
See drm.c:

    wlr_conn->output.phys_width = drm_conn->mmWidth;
    wlr_conn->output.phys_height = drm_conn->mmHeight;


ddevault:
Thanks!

PR #1447, opened 2018-12-30
examples/layer-shell.c: check popup exists before drawing

ianyfan:
Fixes a crash that occasionally occurs if you start this example and right click around a lot, I assume due to the popup being destroyed between scheduling a draw frame and the actual drawing.


emersion:
Thanks!

PR #1448, opened 2019-01-02
rootston: fix crash with fullscreen surfaces

myfreeweb:
Fixes RetroArch fullscreen mode.

### DIFF
--- a/rootston/output.c
+++ b/rootston/output.c
@@ -67,6 +67,10 @@ static void surface_for_each_surface(struct wlr_surface *surface,
 static void view_for_each_surface(struct roots_view *view,
 		struct layout_data *layout_data, wlr_surface_iterator_func_t iterator,
 		void *user_data) {
+	if (!view->wlr_surface) {
+		return;
+	}
+
 	layout_data->x = view->box.x;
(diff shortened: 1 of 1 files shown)


emersion:
Hmm. Maybe we could un-fullscreen when un-mapping views instead?

emersion:
Okay, I see that this also happens before views are mapped. Can you move the check at the beginning of `view_for_each_surface`? With ioquake3 this segfault happens in the presentation feedback codepath.

myfreeweb:
Updated. Does this fix ioq3?

rootston/output.c
```diff
@@ -67,6 +67,9 @@ static void surface_for_each_surface(struct wlr_surface *surface,
 static void view_for_each_surface(struct roots_view *view,
 		struct layout_data *layout_data, wlr_surface_iterator_func_t iterator,
 		void *user_data) {
+	if (!view->wlr_surface)
```
emersion: Style: please always use brackets

emersion:
Yes, this does fix ioquake3 :)

emersion:
Thanks!

PR #1449, opened 2019-01-03
Merge tinywl

ddevault:


tinywl/Makefile
```diff
@@ -0,0 +1,27 @@
+WAYLAND_PROTOCOLS=/usr/share/wayland-protocols
```
emersion: `pkg-config` should be used instead

emersion:
I'm not too worried about maintenance costs.

+1

PR #1451, opened 2019-01-04
relative-pointer-v1: move pointer iteration in wlroots

emersion:
This makes sure we always send events to the right client.

### DIFF
--- a/rootston/cursor.c
+++ b/rootston/cursor.c
@@ -35,7 +35,8 @@ void roots_cursor_destroy(struct roots_cursor *cursor) {
 	// TODO
 }
 
-static void seat_view_deco_motion(struct roots_seat_view *view, double deco_sx, double deco_sy) {
+static void seat_view_deco_motion(struct roots_seat_view *view,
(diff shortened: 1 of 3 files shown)


rootston/cursor.c
```diff
@@ -309,24 +309,9 @@ static void notify_relative_motion(struct roots_seat *seat, uint64_t time_msec,
 		double dx, double dy, double dx_unaccel, double dy_unaccel) {
 	struct wlr_relative_pointer_manager_v1 *relative_pointer_manager =
 		seat->input->server->desktop->relative_pointer_manager;
-
```
ammen99: what is the point of having this function at all, it is just a one-line wrapper for `wlr_relative_pointer_manager_v1_send_relative_motion` ...

ammen99:
LGTM aside from the `notify_relative_motion` which doesn't really make sense now IMO

rootston/cursor.c
emersion: Good catch, fixed

PR #1454, opened 2019-01-07
Simplify FreeBSD build

jbeich:
FreeBSD Ports recently updated libinput et al. which is now in both `/latest` and `/quarterly` package sets. `epoll-shim.pc` is still missing, so I've filed https://github.com/FreeBSDDesktop/epoll-shim/pull/1

### DIFF
--- a/backend/x11/input_device.c
+++ b/backend/x11/input_device.c
@@ -2,11 +2,7 @@
 
 #include <wlr/config.h>
 
-#ifdef __linux__
 #include <linux/input-event-codes.h>
-#elif __FreeBSD__
-#include <dev/evdev/input-event-codes.h>
-#endif
 
 #include <xcb/xcb.h>
 #include <xcb/xfixes.h>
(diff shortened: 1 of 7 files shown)


backend/x11/input_device.c
```diff
@@ -2,11 +2,7 @@
 
 #include <wlr/config.h>
 
-#ifdef __linux__
```
emersion: Hmm, what's the story about these? Are `linux/*` includes available on FreeBSD now?

jbeich: devel/evdev-proto currently provides `<linux/input.h>`, `<linux/input-event-codes.h>`, `<linux/uinput.h>`. Those files are usually newer than what's shipped in FreeBSD base.

@wulf7, do you know when consumers should use `<dev/evdev/*>` and when `<linux/*>`?

wulf7: The rule is simple: always use <linux/*> except you are writing something FreeBSD-only like kernel drivers

emersion: Okay, thanks!

PR #1461, opened 2019-01-13
Remove wlr_xdg_surface_send_close.

### LINKED ISSUE #1417: Remove wlr_xdg_surface_send_close
Replace with two separate functions: `wlr_xdg_toplevel_send_close` and `wlr_xdg_popup_destroy`.

The problem is that closing a popup has different semantics from closing a toplevel.

- When closing a toplevel, the compositor merely asks the client to shut down -- the client can ignore the request
- When closing a popup, we notify the client that the popup has been closed, the client cannot do anything about it. In this case we can destroy (aka `free`) the `wlr_xdg_surface` and make the resource inert.

c-edw:
Closes #1417.

I hope this is along the lines of what is needed.

Sending PR on Sway for this shortly.

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -296,9 +296,14 @@ uint32_t wlr_xdg_toplevel_set_tiled(struct wlr_xdg_surface *surface,
 		uint32_t tiled_edges);
 
 /**
- * Request that this xdg surface closes.
+ * Request that this xdg toplevel closes.
  */
-void wlr_xdg_surface_send_close(struct wlr_xdg_surface *surface);
+void wlr_xdg_toplevel_send_close(struct wlr_xdg_surface *surface);
+
+/**
+ * Request that this xdg popup closes.
+ **/
+void wlr_xdg_popup_destroy(struct wlr_xdg_surface *surface);
 
 /**
  * Get the geometry for this positioner based on the anchor rect, gravity, and
--- a/rootston/xdg_shell.c
+++ b/rootston/xdg_shell.c
@@ -248,9 +248,9 @@ static void close(struct roots_view *view) {
 	struct wlr_xdg_surface *surface = view->xdg_surface;
 	struct wlr_xdg_popup *popup = NULL;
 	wl_list_for_each(popup, &surface->popups, link) {
-		wlr_xdg_surface_send_close(popup->base);
+		wlr_xdg_popup_destroy(popup->base);
 	}
-	wlr_xdg_surface_send_close(surface);
+	wlr_xdg_toplevel_send_close(surface);
 }
 
 static void destroy(struct roots_view *view) {
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -517,21 +517,17 @@ void wlr_xdg_surface_ping(struct wlr_xdg_surface *surface) {
 		surface->client->ping_serial);
 }
 
-void wlr_xdg_surface_send_close(struct wlr_xdg_surface *surface) {
-	switch (surface->role) {
-	case WLR_XDG_SURFACE_ROLE_NONE:
-		assert(0 && "not reached");
-		break;
-	case WLR_XDG_SURFACE_ROLE_TOPLEVEL:
-		if (surface->toplevel) {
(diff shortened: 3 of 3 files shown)


types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -517,21 +517,17 @@ void wlr_xdg_surface_ping(struct wlr_xdg_surface *surface) {
...
-			xdg_toplevel_send_close(surface->toplevel->resource);
-		}
-		break;
-	case WLR_XDG_SURFACE_ROLE_POPUP:
-		if (surface->popup) {
-			xdg_popup_send_popup_done(surface->popup->resource);
-		}
-		break;
+void wlr_xdg_toplevel_send_close(struct wlr_xdg_surface *surface) {
+	if (surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL
```
emersion: I think it makes sense to make these assertions instead

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -296,9 +296,14 @@ uint32_t wlr_xdg_toplevel_set_tiled(struct wlr_xdg_surface *surface,
...
- * Request that this xdg surface closes.
+ * Request that this xdg toplevel closes.
  */
-void wlr_xdg_surface_send_close(struct wlr_xdg_surface *surface);
+void wlr_xdg_toplevel_send_close(struct wlr_xdg_surface *surface);
+
+/**
+ * Request that this xdg popup closes.
+ **/
+void wlr_xdg_popup_destroy(struct wlr_xdg_surface *surface);
```
ammen99: Why isn't this named `wlr_xdg_popup_done/send_done`? And at least, because it takes a popup, why not take a `wlr_xdg_popup` as a parameter instead?

emersion: >Why isn't this named wlr_xdg_popup_done/send_done?

Because it effectively _destroys_ the `wlr_xdg_popup`. I prefer to make it clear the reference is not valid anymore after calling this function.

>And at least, because it takes a popup, why not take a wlr_xdg_popup as a parameter instead?

This is a more general issue. A lot of functions in this header could be more type-safe by taking a toplevel/popup instead of a xdg_surface. I agree this should be fixed, but probably as a separate PR.

ammen99: > Because it effectively _destroys_ the `wlr_xdg_popup`. I prefer to make it clear the reference is not valid anymore after calling this function.
> 
> >

Well, it just sends a done event if I understand the code correctly. The protocol says "The client should destroy the xdg_popup object at this point." which I interpret that the popup might be "alive" for some time after this request. Feel free to just tell me I'm wrong if that's the case.

emersion: You're right. However, I'd like to free the popup and make the resource inert in this case. That's what we try to do when the compositor decides to free a resource and notifies the client about it.

emersion: The client _should_ destroy the resource. But making the resource inert doesn't mean we free it. Inert means the resource can no longer be used by the client (only the `destroy` request can be used) and we can free our state.

ammen99: Okay, maybe my understanding of the protocol language isn't good enough. I thought that a client may choose to keep its popup open (so this was more like a hint).

types/xdg_shell/wlr_xdg_surface.c
c-edw: Yep, I agree.

rootston/xdg_shell.c
```diff
@@ -248,9 +248,9 @@ static void close(struct roots_view *view) {
 	struct wlr_xdg_surface *surface = view->xdg_surface;
 	struct wlr_xdg_popup *popup = NULL;
 	wl_list_for_each(popup, &surface->popups, link) {
-		wlr_xdg_surface_send_close(popup->base);
+		wlr_xdg_popup_destroy(popup->base);
 	}
-	wlr_xdg_surface_send_close(surface);
+	wlr_xdg_toplevel_send_close(surface);
```
c-edw: Is it safe to assume this surface is a toplevel?

emersion: Yes

emersion:
Thanks!

PR #1462, opened 2019-01-13
Fix software cursor rendering for tinywl and some examples

alexbakker:
Since #1366 was merged, a call to ``wlr_output_render_software_cursors`` is needed to render software cursors. This PR adds a call to that function for:
- examples/pointer
- examples/multi-pointer
- tinywl

I think that's all of them, but let me know if I missed one.

### DIFF
--- a/tinywl/tinywl.c
+++ b/tinywl/tinywl.c
@@ -610,6 +610,14 @@ static void output_frame(struct wl_listener *listener, void *data) {
 				render_surface, &rdata);
 	}
 
+	/* Hardware cursors are rendered by the GPU on a separate plane, and can be
+	 * moved around without re-rendering what's beneath them - which is more
+	 * efficient. However, not all hardware supports hardware cursors. For this
+	 * reason, wlroots provides a software fallback, which we ask it to render
+	 * here. wlr_cursor handles configuring hardware vs software cursors for you,
+	 * and this function is a no-op when hardware cursors are in use. */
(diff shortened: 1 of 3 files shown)


tinywl/tinywl.c
```diff
@@ -610,6 +610,9 @@ static void output_frame(struct wl_listener *listener, void *data) {
 				render_surface, &rdata);
 	}
 
+	/* Render software cursors */
+	wlr_output_render_software_cursors(output->wlr_output, NULL);
```
ddevault: Can you expand on this comment, explaining the difference between hardware and software cursors, why this is necessary, and what wlroots does for us?

alexbakker: To be honest, I don't know much about how the internals work. I just looked at the patches rootston and sway received for #1366 and did the same here.

ddevault:
```
/*
 * Hardware cursors are rendered by the GPU on a separate plane, and can be
 * moved around without re-rendering what's beneath them - which is more
 * efficient. However, not all hardware supports hardware cursors. For this
 * reason, wlroots provides a software fallback, which we ask it to render
 * here. wlr_cursor handles configuring hardware vs software cursors for you,
 * and this function is a no-op when hardware cursors are in use.
 */
```

alexbakker: Added, thanks.

ddevault:
Thanks!

PR #1467, opened 2019-01-15
Add ConsoleKit2 launch support

### LINKED ISSUE #1178: Add ConsoleKit2 support
I've been trying to package wlroots for [Alpine Linux](https://alpinelinux.org)/[postmarketOS](https://postmarketos.org), with the intention to eventually both run Sway and Phosh. wlroots compiled without any issues on Musl (for which I applaude you guys) and rootston runs fine when running as root.

From what I understand though, to run it without root rights, it needs a logind implementation. Since we do not have systemd (and thus no logind) or elogind, I'd like to request support for [ConsoleKit2](https://github.com/consolekit2/consolekit2) as well.

myfreeweb:
Fixes #1178.

@PureTryOut please test on Linux, I've only tested on FreeBSD.

---

This is similar to the logind backend, but connecting to the event loop is a bit more complicated (uses the method stolen from Weston). Plus, there's a couple lines to handle FreeBSD's "meaningless" major/minor numbers (since there's no persistent `DRM_MAJOR` to check against, I store the `makedev()` number for the first DRM device and compare against that).

### DIFF
--- a/backend/meson.build
+++ b/backend/meson.build
@@ -50,6 +50,11 @@ if logind.found()
 	backend_deps += logind
 endif
 
+if consolekit.found()
+	backend_files += files('session/consolekit2.c')
+	backend_deps += [dbus, threads, consolekit]
+endif
+
 subdir('x11')
 
 lib_wlr_backend = static_library(
--- a/backend/session/consolekit2.c
+++ b/backend/session/consolekit2.c
@@ -1,0 +1,857 @@
+#define _POSIX_C_SOURCE 200809L
+#ifdef __FreeBSD__
+// for major/minor
+#define __BSD_VISIBLE 1
+#endif
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <pthread.h>
+#include <wayland-server.h>
+#include <wlr/backend/session/interface.h>
+#include <wlr/config.h>
+#include <wlr/util/log.h>
+#include "util/signal.h"
+
+#ifdef __linux__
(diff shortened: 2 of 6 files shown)


PureTryOut:
Wow thanks! I don't have much time this week, so testing will sadly have to wait till sunday.

emersion:
I'm not personally willing to maintain this libdbus code.

myfreeweb:
I don't see the need for any maintenance at all though… libdbus API is not going to change ever. `wl_event_` has been very stable too. The event loop integration doesn't touch anything else.

emersion:
That's not how it works. When we merge something, it means we officially support it. It means users can (and will) report bugs and it means contributors can send patches. The rest of wlroots can change too: the way threads/signals are handled now in the ConsoleKit2 session could break because of this, and the session interface can change too.

This patch is 879 lines of code we'll need to maintain. I don't want to have ConsoleKit2 as a second-class citizen in wlroots, nor do I want to commit to fix it when we'll find bugs in it.

ddevault:
Emersion has extracted sd-bus into a standalone library:

https://github.com/emersion/basu

I recommend basing this on sd-bus and shipping basu on BSD.

emersion:
There's discussion about this in https://github.com/emersion/mako/pull/92

basu needs to strip down more things, and it's still coming from systemd so it's still suing some Linux-specific features. Alternatively, @sghctoma has been working on a sd-bus shim on top of libdbus.

ddevault:
Aye, some work is still necessary in these areas. In either case, the final merge will look a lot different from this, so let's close it.

ddevault:
By the way, it wouldn't bother me if ports wants to merge this patch downstream in the meanwhile.

PureTryOut:
I haven't been able to test it yet, but we'll probably make use of this downstream indeed till there is an upstream solution. Everything better then setting a suid bit :sweat_smile:

PR #1470, opened 2019-01-16
Remove all VLAs.

c-edw:
This follows on from [#3438](https://github.com/swaywm/sway/pull/3438) on Sway. Please check there for the original justification.

This is a bigger change so I'd like to take the time to get this right and minimise the chance of breaking anything.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -445,15 +445,23 @@ static void realloc_planes(struct wlr_drm_backend *drm, const uint32_t *crtc_in,
 		bool *changed_outputs) {
 	wlr_log(WLR_DEBUG, "Reallocating planes");
 
+	uint32_t *possible = NULL;
+	uint32_t *crtc = NULL;
+	uint32_t *crtc_res = NULL;
+
 	// overlay, primary, cursor
 	for (size_t type = 0; type < 3; ++type) {
 		if (drm->num_type_planes[type] == 0) {
 			continue;
 		}
 
-		uint32_t possible[drm->num_type_planes[type] + 1];
-		uint32_t crtc[drm->num_crtcs + 1];
-		uint32_t crtc_res[drm->num_crtcs + 1];
+		possible = malloc(sizeof(uint32_t) * drm->num_type_planes[type]);
+		crtc = malloc(sizeof(uint32_t) * drm->num_crtcs);
+		crtc_res = malloc(sizeof(uint32_t) * drm->num_crtcs);
+
+		if (possible == NULL || crtc == NULL || crtc_res == NULL) {
+			goto error_possible;
+		}
 
 		for (size_t i = 0; i < drm->num_type_planes[type]; ++i) {
 			possible[i] = drm->type_planes[type][i].possible_crtcs;
@@ -501,7 +509,18 @@ static void realloc_planes(struct wlr_drm_backend *drm, const uint32_t *crtc_in,
 				*old = new;
 			}
 		}
+
+		free(possible);
+		free(crtc);
+		free(crtc_res);
 	}
+
+	return;
+error_possible:
+	free(crtc_res);
+	free(crtc);
+	free(possible);
+	wlr_log(WLR_ERROR, "Failed to allocate memory");
 }
 
 static void drm_connector_cleanup(struct wlr_drm_connector *conn);
@@ -564,7 +583,7 @@ bool wlr_drm_connector_add_mode(struct wlr_output *output,
 
 	struct wlr_drm_mode *mode = calloc(1, sizeof(*mode));
 	if (!mode) {
-		return false;
+		goto error_mode;
 	}
 	memcpy(&mode->drm_mode, modeinfo, sizeof(*modeinfo));
 
@@ -578,6 +597,9 @@ bool wlr_drm_connector_add_mode(struct wlr_output *output,
 			mode->wlr_mode.refresh);
 	wl_list_insert(&conn->output.modes, &mode->wlr_mode.link);
 	return true;
+error_mode:
+	wlr_log(WLR_ERROR, "Failed to allocate memory");
+	return false;
 }
 
 static void drm_connector_transform(struct wlr_output *output,
@@ -602,8 +624,7 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 		// We don't have a real cursor plane, so we make a fake one
 		plane = calloc(1, sizeof(*plane));
 		if (!plane) {
-			wlr_log_errno(WLR_ERROR, "Allocation failed");
-			return false;
+			goto error_plane;
 		}
 		crtc->cursor = plane;
 	}
@@ -720,6 +741,9 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 		wlr_output_update_needs_swap(output);
 	}
 	return ok;
+error_plane:
+	wlr_log(WLR_ERROR, "Failed to allocate memory");
+	return false;
 }
 
 static bool drm_connector_move_cursor(struct wlr_output *output,
@@ -884,15 +908,24 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
 
 	wlr_log(WLR_DEBUG, "Reallocating CRTCs");
 
-	uint32_t crtc[drm->num_crtcs + 1];
+	uint32_t *crtc = malloc(sizeof(uint32_t) * drm->num_crtcs);
+	if (crtc == NULL) {
+		goto error_crtc;
+	}
 	for (size_t i = 0; i < drm->num_crtcs; ++i) {
 		crtc[i] = UNMATCHED;
 	}
 
-	struct wlr_drm_connector *connectors[num_outputs + 1];
+	struct wlr_drm_connector **connectors =
+		calloc(num_outputs, sizeof(struct wlr_drm_connector));
+	if (connectors == NULL) {
+		goto error_connectors;
+	}
 
-	uint32_t possible_crtc[num_outputs + 1];
-	memset(possible_crtc, 0, sizeof(possible_crtc));
+	uint32_t *possible_crtc = calloc(num_outputs, sizeof(uint32_t));
+	if (possible_crtc == NULL) {
+		goto error_possible_crtc;
+	}
 
 	wlr_log(WLR_DEBUG, "State before reallocation:");
 	ssize_t i = -1;
@@ -919,12 +952,18 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
 		}
 	}
 
-	uint32_t crtc_res[drm->num_crtcs + 1];
+	uint32_t *crtc_res = malloc(sizeof(uint32_t) * drm->num_crtcs);
+	if (crtc_res == NULL) {
+		goto error_crtc_res;
+	}
 	match_obj(wl_list_length(&drm->outputs), possible_crtc,
 		drm->num_crtcs, crtc, crtc_res);
 
-	bool matched[num_outputs + 1];
-	memset(matched, false, sizeof(matched));
+	bool *matched = calloc(num_outputs, sizeof(bool));
+	if (matched == NULL) {
+		goto error_matched;
+	}
+
 	for (size_t i = 0; i < drm->num_crtcs; ++i) {
 		if (crtc_res[i] != UNMATCHED) {
 			matched[crtc_res[i]] = true;
@@ -997,6 +1036,24 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
 
 		wlr_output_damage_whole(&conn->output);
 	}
+
+	free(matched);
+	free(crtc_res);
+	free(possible_crtc);
+	free(connectors);
+	free(crtc);
+
+	return;
+error_matched:
+	free(crtc_res);
+error_crtc_res:
+	free(possible_crtc);
+error_possible_crtc:
+	free(connectors);
+error_connectors:
+	free(crtc);
+error_crtc:
+	wlr_log(WLR_ERROR, "Failed to allocate memory");
 }
 
 static uint32_t get_possible_crtcs(int fd, drmModeRes *res,
@@ -1050,11 +1107,16 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 	size_t seen_len = wl_list_length(&drm->outputs);
 	// +1 so length can never be 0, which is undefined behaviour.
 	// Last element isn't used.
-	bool seen[seen_len + 1];
-	memset(seen, false, sizeof(seen));
+	bool *seen = calloc(seen_len, sizeof(bool));
+	if (seen == NULL) {
+		goto error_seen;
+	}
 	size_t new_outputs_len = 0;
-	struct wlr_drm_connector *new_outputs[res->count_connectors + 1];
-
+	struct wlr_drm_connector **new_outputs =
+		calloc(res->count_connectors, sizeof(struct wlr_drm_connector*));
+	if (new_outputs == NULL) {
+		goto error_new_outputs;
+	}
 	for (int i = 0; i < res->count_connectors; ++i) {
 		drmModeConnector *drm_conn = drmModeGetConnector(drm->fd,
 			res->connectors[i]);
@@ -1161,8 +1223,7 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 			for (int i = 0; i < drm_conn->count_modes; ++i) {
 				struct wlr_drm_mode *mode = calloc(1, sizeof(*mode));
 				if (!mode) {
-					wlr_log_errno(WLR_ERROR, "Allocation failed");
-					continue;
+					goto error_mode;
 				}
 
 				if (drm_conn->modes[i].flags & DRM_MODE_FLAG_INTERLACE) {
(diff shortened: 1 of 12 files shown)


emersion:
* Some VLAs have `+ 1` to ensure they don't have a zero length (undefined behaviour). This can be removed when using dynamic allocation.
* All dynamic memory allocations in wlroots should be guarded with NULL checks.
* `malloc` followed by `memset` can be replaced with `calloc`.
* Care must be taken to properly `free` memory blocks, even in error conditions (note: `free(NULL)` is a no-op).

render/egl.c
```diff
@@ -342,7 +346,7 @@ bool wlr_egl_swap_buffers(struct wlr_egl *egl, EGLSurface surface,
 		int nrects;
 		pixman_box32_t *rects =
 			pixman_region32_rectangles(&flipped_damage, &nrects);
-		EGLint egl_damage[4 * nrects];
+		EGLint *egl_damage = malloc(sizeof(EGLint) * (4 * nrects));
```
atomnuker: This isn't freed?
Also per-frame or per-damage malloc is a terrible terrible idea. Don't. Ever.
Make an array somewhere in the main context and increase its size only if needed.

c-edw: Yep, thanks for spotting it. I'm not really looking at the greater context of this code so I had no idea that this was a hot code path. I'll look into implementing what you suggested.

c-edw:
Looking into resolving clang build failures, it works on my machine. I will try building on Arch inside Docker.

render/egl.c
```diff
@@ -95,7 +95,7 @@ static void print_dmabuf_formats(struct wlr_egl *egl) {
 		return;
 	}
 
-	char *str_formats = malloc(num * 5);
+	char *str_formats = malloc((num * 5) + 1);
 	if (str_formats == NULL) {
```
martinetd: (latest commit) I'd rather change the rest -- either return early from the function or check to `str_formats == NULL && num != 0` -- bailing out early or `foo = malloc(0); ...that doesn't use foo..; free(foo);`

c-edw: Could you clarify for me please?

martinetd: Sorry, I thought that +1 was the same as an earlier comment because of the commit message e.g. "the VLA code used +1 because a VLA of zero size is undefined behaviour, while it's perfectly OK to do malloc(0)" -- but it's actually needed for the terminal nul byte.

This one is fine, but if there are other malloc(x+1) where x can be 0 because of the VLa undefined behaviour, we don't really need that +1 anymore, it's ok to malloc(0) -- that'll return NULL or some memory pointer you're not supposed to use depending on the system, but it's perfectly fine as long as it's not accessed. Sorry for not taking the time to do a proper review.


That being said, this code has two problems:
 - If you add a malloc, you always need to add a free somewhere. I didn't pay attention  earlier but as it is this `str_formats` is leaking.
 - if num is 0, nothing prints the terminal nul byte for wlr_log to be safe - it will likely print garbage after the allocated region. You need to at least set the last byte of str_formats to 0 or '\0' (or use calloc).

c-edw: This is very helpful, thank you. `str_formats` was indeed leaking and I've now corrected that. I've done my best to remove the unnecessary `+ 1`s which were avoiding the UB you mentioned. To fix the `num == 0` issue, I simply guarded the code with a condition which tests for `num > 0`. It looks like the code didn't need to execute if `num == 0`.

c-edw:
I think this is more or less ready (famous last words). If somebody could give it a full review I would appreciate it.

backend/drm/drm.c
```diff
@@ -941,6 +993,8 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
 		}
```
martinetd: (fuck github, for not letting one comment out of default visible context) --
one line above this is a return, so leaking things.

Personal opinion here but I don't really like the style of freeing exactly what's needed just as it's not needed anymore and am in the camp of freeing everything at the end with gotos jumping there on errors if possible. I guess either are fine, just got to be careful...

backend/drm/util.c
```diff
@@ -323,10 +324,14 @@ static bool match_obj_(struct match_state *st, size_t skips, size_t score, size_
...
-		size_t num_res, const uint32_t res[static restrict num_res],
-		uint32_t out[static restrict num_res]) {
-	uint32_t solution[num_res];
+size_t match_obj(size_t num_objs, const uint32_t *objs,
+		size_t num_res, const uint32_t *res,
+		uint32_t *out) {
+	uint32_t *solution = malloc(sizeof(uint32_t) * num_res);
+	if (solution == NULL) {
+		wlr_log(WLR_ERROR, "Failed to allocate memory.");
+		return 0;
```
martinetd: Should we make this ssize_t and return -ENOMEM to propagate? There's only 2 call sites that don't use the return value at all.

render/egl.c
```diff
@@ -19,7 +19,11 @@ static bool egl_get_config(EGLDisplay disp, EGLint *attribs, EGLConfig *out,
...
 
-	EGLConfig configs[count];
+	EGLConfig *configs = malloc(sizeof(EGLConfig) * count);
+	if (configs == NULL) {
+		wlr_log(WLR_ERROR, "Failed to allocate memory");
+		return false;
+	}
 
 	ret = eglChooseConfig(disp, attribs, configs, count, &matched);
 	if (ret == EGL_FALSE) {
```
martinetd: (insert same fuck github comment) -- one return false below this line

render/egl.c
```diff
@@ -342,7 +351,20 @@ bool wlr_egl_swap_buffers(struct wlr_egl *egl, EGLSurface surface,
...
 		pixman_box32_t *rects =
 			pixman_region32_rectangles(&flipped_damage, &nrects);
-		EGLint egl_damage[4 * nrects];
+
+		static EGLint *egl_damage;
+		static uint32_t egl_damage_size = 0;
+
+		// Expand egl_damage if it isn't large enough.
+		if (egl_damage_size < (uint32_t) 4 * nrects) {
+			egl_damage = malloc(sizeof(EGLint) * (4 * nrects));
```
martinetd: I don't see where this is freed

render/wlr_renderer.c
```diff
@@ -203,7 +203,7 @@ struct wlr_renderer *wlr_renderer_autocreate(struct wlr_egl *egl,
 
 	size_t all_config_attribs_len = config_attribs_len +
 		sizeof(gles2_config_attribs) / sizeof(gles2_config_attribs[0]);
-	EGLint all_config_attribs[all_config_attribs_len];
+	EGLint *all_config_attribs = malloc(sizeof(EGLint) * all_config_attribs_len);
 	if (config_attribs_len > 0) {
 		memcpy(all_config_attribs, config_attribs,
 			config_attribs_len * sizeof(EGLint));
```
martinetd: There's one return NULL in the out of context blob below as well

backend/drm/drm.c
```diff
@@ -451,9 +451,24 @@ static void realloc_planes(struct wlr_drm_backend *drm, const uint32_t *crtc_in,
 			continue;
 		}
 
-		uint32_t possible[drm->num_type_planes[type] + 1];
-		uint32_t crtc[drm->num_crtcs + 1];
-		uint32_t crtc_res[drm->num_crtcs + 1];
+		uint32_t *possible = malloc(sizeof(uint32_t) * drm->num_type_planes[type]);
+		uint32_t *crtc = malloc(sizeof(uint32_t) * drm->num_crtcs);
+		uint32_t *crtc_res = malloc(sizeof(uint32_t) * drm->num_crtcs);
```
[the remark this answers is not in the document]
c-edw: I had something similar to your first example a few commits back, but clang wasn't happy with it. I think because some of the variables could've been freed before they were cleared in the `goto` block. Anyway - I much prefer your suggestion and clang doesn't complain about it, so I've made the change.

backend/drm/drm.c
```diff
@@ -593,17 +618,16 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 	struct wlr_drm_backend *drm = get_drm_backend_from_backend(output->backend);
 
 	struct wlr_drm_crtc *crtc = conn->crtc;
-	if (!crtc) {
-		return false;
+	if (crtc == NULL) {
+		goto allocate_fail;
```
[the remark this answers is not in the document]
c-edw: I think I was on autopilot. Reverted.

backend/drm/drm.c
```diff
@@ -593,17 +618,16 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
...
 	}
 
 	struct wlr_drm_plane *plane = crtc->cursor;
 	if (!plane) {
 		// We don't have a real cursor plane, so we make a fake one
 		plane = calloc(1, sizeof(*plane));
 		if (!plane) {
-			wlr_log_errno(WLR_ERROR, "Allocation failed");
-			return false;
+			goto allocate_fail;
```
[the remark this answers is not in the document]
c-edw: Since `allocate_fail` has been declared it makes sense to me to use it. You say we lose line information in the debug output which is unfortunate, but it's a compromise.

backend/drm/drm.c
```diff
@@ -884,15 +911,27 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
...
+	if (crtc == NULL) {
+		goto allocate_fail;
+	}
 	for (size_t i = 0; i < drm->num_crtcs; ++i) {
 		crtc[i] = UNMATCHED;
 	}
 
-	struct wlr_drm_connector *connectors[num_outputs + 1];
+	struct wlr_drm_connector **connectors
+		= calloc(num_outputs, sizeof(struct wlr_drm_connector));
```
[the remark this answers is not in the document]
c-edw: Fixed.

backend/drm/drm.c
c-edw: Yes, immediate freeing has been a headache, especially in these busy functions. I've decided to free in blocks at the end of the function instead. I honestly don't see much advantage in immediately freeing resources anyway.

backend/drm/util.c
c-edw: Not sure what the practice is here so I'll wait for somebody's input.

render/egl.c
c-edw: I'm not sure you're referring to here, probably because of the issue you pointed out with GitHub's comments. Perhaps you could copy the code context into a comment.

render/egl.c
c-edw: I've made `egl_damage` static since we want to keep some allocated memory quickly available for this function. It won't be freed until the programs exits, I guess.

render/wlr_renderer.c
c-edw: Again, not sure what you're referring to because of this issue with out-of-view comments. Sorry!

c-edw:
I have added `build-gcc/` and `build-clang/` to the gitignore, I hope that's alright. It makes things a bit easier when testing on the respective compilers.

render/egl.c
martinetd: Litteraly two lines below is a return false leaking configs:

```c
	ret = eglChooseConfig(disp, attribs, configs, count, &matched);
	if (ret == EGL_FALSE) {
		wlr_log(WLR_ERROR, "eglChooseConfig failed");
		return false;
	}
```

render/egl.c
martinetd: I see. But you still allocate a new one if the size grows, so you need to free the previous one there when allocating a new one at least.

render/wlr_renderer.c
martinetd: This leaks all_config_attribs:
```c
	if (!wlr_egl_init(egl, platform, remote_display, all_config_attribs,
			visual_id)) {
		wlr_log(WLR_ERROR, "Could not initialize EGL");
		return NULL;
	}
```

backend/drm/drm.c
```diff
@@ -997,6 +1041,17 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
...
 	}
+
+	free(possible_crtc);
+	free(matched);
+	free(crtc);
+	free(crtc_res);
+	free(connectors);
+
+	return;
+allocate_fail:
```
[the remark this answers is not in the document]
c-edw: Are you suggesting something like:

```c
error_crtc:
	wlr_log(WLR_ERROR, "Failed to allocate memory");
error_connectors:
	free(crtc);
	wlr_log(WLR_ERROR, "Failed to allocate memory");
error_possible_crtc:
	free(connectors);
	free(crtc);
	wlr_log(WLR_ERROR, "Failed to allocate memory");
error_crtc_res:
	free(possible_crtc);
	free(connectors);
	free(crtc);
	wlr_log(WLR_ERROR, "Failed to allocate memory");
error_matched:
	free(crtc_res);
	free(possible_crtc);
	free(connectors);
	free(crtc);
	wlr_log(WLR_ERROR, "Failed to allocate memory");
```

and just in this particular function or any that I've changed?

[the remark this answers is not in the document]
ddevault: You need more return statements, but yes, something like that. Some of these can be combined into one label, too.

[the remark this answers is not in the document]
emersion: Something among the lines of

```c
error_crtc:
	free(crtc);
error_connectors:
	free(connectors);
error_crtc_res:
	free(crtc_res);
error_matched:
	free(matched);
	wlr_log(WLR_ERROR, "Failed to allocate memory");
```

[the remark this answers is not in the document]
c-edw: Ah! I misunderstood how these work. I didn't think that it would fall through to the next statement even if a `return` was missing. Alright, I will get this done.

c-edw:
I went back and reviewed the diff in its entirety, spotted a few issues and made a few improvements. I would still like to hear opinions on @martinetd's suggestion:
```
Should we make this ssize_t and return -ENOMEM to propagate? There's only 2 call sites that don't use the return value at all.
```

backend/drm/util.c
emersion: `ssize_t` is only guaranteed to be able to hold `size_t` plus -1, nothing more. Maybe return -1 to indicate failure?

backend/drm/drm.c
```diff
@@ -884,15 +908,24 @@ static void realloc_crtcs(struct wlr_drm_backend *drm, bool *changed_outputs) {
...
+	if (crtc == NULL) {
+		goto error_crtc;
+	}
 	for (size_t i = 0; i < drm->num_crtcs; ++i) {
 		crtc[i] = UNMATCHED;
 	}
 
-	struct wlr_drm_connector *connectors[num_outputs + 1];
+	struct wlr_drm_connector **connectors =
+		calloc(num_outputs, sizeof(struct wlr_drm_connector));
```
ascent12: This is the wrong sizeof. This is an array of pointers, not just a normal array.

c-edw:
I changed it to return `-1`. If we want to set errno to ENOMEM can it be done in another PR?

backend/drm/util.c
```diff
@@ -323,10 +324,14 @@ static bool match_obj_(struct match_state *st, size_t skips, size_t score, size_
...
 }
 
-size_t match_obj(size_t num_objs, const uint32_t objs[static restrict num_objs],
-		size_t num_res, const uint32_t res[static restrict num_res],
-		uint32_t out[static restrict num_res]) {
-	uint32_t solution[num_res];
+ssize_t match_obj(size_t num_objs, const uint32_t *objs,
+		size_t num_res, const uint32_t *res,
+		uint32_t *out) {
+	uint32_t *solution = malloc(sizeof(uint32_t) * num_res);
```
ascent12: This has an upper bound of 32, because we're recursing over the bits in a uint32_t.
Remove the dynamic allocation.

backend/drm/util.c
```diff
@@ -323,10 +324,14 @@ static bool match_obj_(struct match_state *st, size_t skips, size_t score, size_
 	return match_obj_(st, skips, score, replaced, i + 1);
 }
 
-size_t match_obj(size_t num_objs, const uint32_t objs[static restrict num_objs],
-		size_t num_res, const uint32_t res[static restrict num_res],
-		uint32_t out[static restrict num_res]) {
-	uint32_t solution[num_res];
+ssize_t match_obj(size_t num_objs, const uint32_t *objs,
+		size_t num_res, const uint32_t *res,
+		uint32_t *out) {
```
ascent12: You've removed the restrict qualification off of all of the arguments.
Also, VLA arguments are completely safe. I oppose their removal.

c-edw: With `-Wvla`, it will complain about VLA arguments.

c-edw:
Does this apply to Sway and the related projects too? If so, it would've been useful to know before PRs were merged for all of them. wlroots is the last project for which I am sending a PR to remove VLAs.

emersion:
>Does  this apply to Sway and the related projects too? If so, it would've  been useful to know before PRs were merged for all of them. wlroots is  the last project for which I am sending a PR to remove VLAs.

Well, it all depends how VLAs are used. When it's for small, bounded, trusted inputs, I think it's fine too.  Typically for the DRM backend, you won't have hundreds of outputs.

However other changes to remove VLAs in other sway projects were welcome. For instance allocating 64k on the stack for `snprintf` wasn't a good idea.

I think this has to be done on a case-by-case basis. Also, when the input size is bounded, using a fixed-size array is probably better. IIRC for DRM some bitfields are used for connectors so we could bound their number to 64.

c-edw:
Any verdict on this? There's a change to the .gitignore I'd like to get on master, so if we decide not to merge this, then I'll send in a separate PR.

ddevault:
Let's not merge this, but if you instead want to evaluate each VLA one-by-one and make a case for removal in specific instances, let's do that.

handicraftsman:
C++ does not recognize VLAs and thus this library cannot be used in C++.

ddevault:
That's by design. This isn't a C++ library, it's a C library.

PR #1473, opened 2019-01-17
Introduce noop backend

RyanDwyer:
The noop backend is similar to headless, but it doesn't contain a renderer. It can be used as a place to stash views for when there's no physical outputs connected.

See https://github.com/swaywm/sway/pull/3441 for implementation in the compositor.


ddevault:
Thanks!

PR #1475, opened 2019-01-18
rootston: document why Xwayland is disconnected before clients

emersion:
A better solution would be to only restart Xwayland if it's exited with a non-zero exit code or if it's been killed. I don't think we have a way to do this though…


ddevault:
Thanks!

>A better solution would be to only restart Xwayland if it's exited with a non-zero exit code or if it's been killed. I don't think we have a way to do this though…

We can do this, can't we? But I dunno if it's particularly important/desirable to.

PR #1479, opened 2019-01-19
backend/drm: fix state for outputs loosing their CRTC

emersion:
When there aren't enough CRTCs for all outputs, we try to move a CRTC from a
disabled output to an enabled one. When this happens, the old output's state
wasn't changed, so the compositor thought it was still enabled and rendering.

This commit marks the old output as WLR_DRM_CONN_NEEDS_MODESET and sets its
current mode to NULL.


ddevault:
Thanks!

PR #1482, opened 2019-01-20
Add wlr_x11_output_set_title

Hjdskes:
This brings the X11 backend in line with the Wayland backend. I can imagine it's useful here too :) I tested this in Cage.


ddevault:
Thanks!

PR #1486, opened 2019-01-22
backend/wayland: fix resizing

### LINKED ISSUE #1371: Wayland backend not resizing properly
1. Open the Wayland backend in sway
2. Make the window floating, and then tiling again
3. The surface is smaller than the tile. In the extra space, sway's background is rendered

Also happens in GNOME Shell.

RedSoxFan:
Fixes #1371 

Before resizing the egl window, the buffers must be swapped

### DIFF
--- a/backend/wayland/output.c
+++ b/backend/wayland/output.c
@@ -43,8 +43,10 @@ static struct wl_callback_listener frame_listener = {
 static bool output_set_custom_mode(struct wlr_output *wlr_output,
 		int32_t width, int32_t height, int32_t refresh) {
 	struct wlr_wl_output *output = get_wl_output_from_output(wlr_output);
+	wlr_egl_swap_buffers(&output->backend->egl, output->egl_surface, NULL);
 	wl_egl_window_resize(output->egl_window, width, height, 0, 0);
 	wlr_output_update_custom_mode(&output->wlr_output, width, height, 0);
+
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


backend/wayland/output.c
```diff
@@ -253,8 +262,7 @@ static void xdg_toplevel_handle_configure(void *data,
 		return;
 	}
 	// loop over states for maximized etc?
-	wl_egl_window_resize(output->egl_window, width, height, 0, 0);
-	wlr_output_update_custom_mode(&output->wlr_output, width, height, 0);
+	output_set_custom_mode((struct wlr_output *)output, width, height, 0);
```
emersion: Please don't cast, instead use `&output->wlr_output`

emersion:
Hmm, is this normal EGL behaviour?

backend/wayland/output.c
RedSoxFan: Fixed. It looks like it's mixed throughout the file and the nearest one uses a cast.

emersion:
According to the [EGL spec](https://www.khronos.org/registry/EGL/specs/eglspec.1.5.pdf) section 3.10.1.1 "Native Window Resizing":

>If the native window corresponding to surface has been resized prior to the swap, surface must be resized to match. surface will normally be resized by the EGL implementation at the time the native window is resized.  If the implementation cannot do this transparently to the client,  then eglSwapBuffers must detect the change and resize surface prior to copying its pixels to the native window.

RedSoxFan:
It looks like the buffers just needed to be swapped before resizing

emersion:
Instead, does `wlr_output_damage_whole` after resizing fix it?

RedSoxFan:
Nope

emersion:
Thanks!

PR #1487, opened 2019-01-22
Make popup inert when destroyed.

### LINKED ISSUE #1417: Remove wlr_xdg_surface_send_close

c-edw:
Closes #1417.

### DIFF
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -524,9 +524,14 @@ void wlr_xdg_toplevel_send_close(struct wlr_xdg_surface *surface) {
 }
 
 void wlr_xdg_popup_destroy(struct wlr_xdg_surface *surface) {
+	if (surface == NULL) {
+		return;
+	}
 	assert(surface->popup);
 	assert(surface->role == WLR_XDG_SURFACE_ROLE_POPUP);
 	xdg_popup_send_popup_done(surface->popup->resource);
+	wl_resource_set_user_data(surface->popup->resource, NULL);
+	destroy_xdg_popup(surface);
 }
 
 static void xdg_popup_get_position(struct wlr_xdg_popup *popup,
--- a/types/xdg_shell/wlr_xdg_popup.c
+++ b/types/xdg_shell/wlr_xdg_popup.c
@@ -285,7 +285,8 @@ void destroy_xdg_popup(struct wlr_xdg_surface *xdg_surface) {
 	if (xdg_surface == NULL) {
 		return;
 	}
-	assert(xdg_surface->role == WLR_XDG_SURFACE_ROLE_POPUP);
+	assert(xdg_surface->role == WLR_XDG_SURFACE_ROLE_POPUP ||
+		xdg_surface->role == WLR_XDG_SURFACE_ROLE_NONE);
 	reset_xdg_surface(xdg_surface);
 }
 


c-edw:
Sorry for the delay, I haven't been feeling great. Are you asking me to swap out the asserts for NULL checks?

emersion:
>Sorry for the delay, I haven't been feeling great.

No worries. Take your time, we're not in a hurry. Take some time off from contributing if you need to. :)

>Are you asking me to swap out the asserts for NULL checks?

Maybe changed to something like this?

```c
assert(xdg_surface->role == WLR_XDG_SURFACE_ROLE_NONE || xdg_surface->role == WLR_XDG_SURFACE_ROLE_POPUP);
```

(Since `reset_xdg_surface` will set the role to "none")

c-edw:
Thanks, I appreciate it. I have read the section in CONTRIBUTING.md about inert resources like I probably should have. It says that handlers should ignore inert resources, but I'm not sure if this also means other code should pretend that these inert resources don't exist. I assumed yes.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -524,13 +524,22 @@ void wlr_xdg_toplevel_send_close(struct wlr_xdg_surface *surface) {
...
+	if (surface == NULL) {
+		return;
+	}
 	assert(surface->popup);
-	assert(surface->role == WLR_XDG_SURFACE_ROLE_POPUP);
+	assert(surface->role == WLR_XDG_SURFACE_ROLE_POPUP ||
+		surface->role == WLR_XDG_SURFACE_ROLE_NONE);
 	xdg_popup_send_popup_done(surface->popup->resource);
+	wl_resource_set_user_data(surface->popup->resource, NULL);
+	destroy_xdg_popup(surface->popup->base);
```
emersion: I believe this can just be `surface`.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -546,6 +555,9 @@ struct wlr_surface *wlr_xdg_surface_surface_at(
 		double *sub_x, double *sub_y) {
 	struct wlr_xdg_popup *popup_state;
 	wl_list_for_each(popup_state, &surface->popups, link) {
+		if (popup_state == NULL) {
```
emersion: This isn't possible. `destroy_xdg_popup` will remove the popup from the list.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -524,13 +524,22 @@ void wlr_xdg_toplevel_send_close(struct wlr_xdg_surface *surface) {
...
+	assert(surface->role == WLR_XDG_SURFACE_ROLE_POPUP ||
+		surface->role == WLR_XDG_SURFACE_ROLE_NONE);
 	xdg_popup_send_popup_done(surface->popup->resource);
+	wl_resource_set_user_data(surface->popup->resource, NULL);
+	destroy_xdg_popup(surface->popup->base);
 }
 
 static void xdg_popup_get_position(struct wlr_xdg_popup *popup,
 		double *popup_sx, double *popup_sy) {
+	if (popup == NULL) {
```
emersion: NULL checks should be added in protocol request handlers, not here. That means `xdg_popup_handle_grab`, but apparently that's already done, so we're good.

emersion:
Thanks!

PR #1488, opened 2019-01-22
Relative pointer/pointer constraints fixes

Lahvuun:
Needed to get pointer locking for Xonotic (and probably other SDL2 appplications) to work under sway.


Lahvuun:
> The current status quo is to make compositors generate headers for the Wayland protocols they use. Can you remove commit [6d95841](https://github.com/swaywm/wlroots/commit/6d95841d7c6739e62cbe1021666ea7ecbc9a4406)?

Done.

ddevault:
Thanks!

PR #1489, opened 2019-01-22
wlr_xdg_output_v1: only send name/description once

RedSoxFan:
The xdg-output protocol states that name and description should only be
sent once per output since they do not change.

This fixes a SIGABRT in swaybg when resizing an output on the x11 and
wayland backends.


emersion:
Thanks!

ianyfan:
Could you move that block into `output_manager_handle_get_xdg_output` so it wouldn't need the `updated` parameter?

emersion:
Indeed, that would be better!

PR #1490, opened 2019-01-22
wlr_xdg_output_v1: move where name & description are sent

ianyfan:
Since they are only sent once, they can be moved outside of `output_send_details`. (since order of events is not important).


RedSoxFan:
Oh actually can you remove the stdbool.h include?

emersion:
Thanks!

PR #1495, opened 2019-01-24
backend/x11 & backend/wayland: make set_title NULL-safe

Hjdskes:
Set the default "wlroots - " title when the title argument to the set_title functions is NULL. Otherwise, for at least the Wayland backend, we'd crash because xdg_toplevel_set_title doesn't handle a NULL pointer.

### DIFF
--- a/backend/wayland/output.c
+++ b/backend/wayland/output.c
@@ -314,10 +314,7 @@ struct wlr_output *wlr_wl_output_create(struct wlr_backend *wlr_backend) {
 		goto error;
 	}
 
-	char title[32];
-	if (snprintf(title, sizeof(title), "wlroots - %s", wlr_output->name)) {
-		xdg_toplevel_set_title(output->xdg_toplevel, title);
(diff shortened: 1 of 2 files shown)


backend/wayland/output.c
```diff
@@ -369,5 +366,14 @@ struct wlr_output *wlr_wl_output_create(struct wlr_backend *wlr_backend) {
 
 void wlr_wl_output_set_title(struct wlr_output *output, const char *title) {
 	struct wlr_wl_output *wl_output = get_wl_output_from_output(output);
+
+	if (title == NULL) {
+		char wl_title[32];
```
emersion: Can you move this declaration right before the `if`? I've seen cases where the compiler will decide to replace the local variable with other stuff outside of the block.

Hjdskes: Sure, force pushed the change.

ddevault:
Thanks!

PR #1500, opened 2019-01-25
Implement the pointer-gestures-unstable-v1 protocol

myfreeweb:
This protocol relays touchpad gesture events produced by libinput to supporting clients (e.g. Evince, Eye of GNOME).

---

Tested with gtk3-demo and the apps mentioned above. I should probably reduce the duplication in the code, but by how much? `swipe(…, BEGIN|UPDATE|END)` + `pinch(…, BEGIN|UPDATE|END)` or `gesture(…, SWIPE_BEGIN|SWIPE_UPDATE|SWIPE_END|PINCH_BEGIN|PINCH_UPDATE|PINCH_END)`?

### DIFF
--- a/backend/libinput/pointer.c
+++ b/backend/libinput/pointer.c
@@ -140,3 +140,123 @@ void handle_pointer_axis(struct libinput_event *event,
 	}
 	wlr_signal_emit_safe(&wlr_dev->pointer->events.frame, wlr_dev->pointer);
 }
+
+void handle_pointer_swipe_begin(struct libinput_event *event,
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_swipe_begin wlr_event = {
+		.device = wlr_dev,
+		.time_msec =
+			usec_to_msec(libinput_event_gesture_get_time_usec(gevent)),
+		.fingers = libinput_event_gesture_get_finger_count(gevent),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.swipe_begin, &wlr_event);
+}
+
+void handle_pointer_swipe_update(struct libinput_event *event,
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_swipe_update wlr_event = {
+		.device = wlr_dev,
+		.time_msec =
+			usec_to_msec(libinput_event_gesture_get_time_usec(gevent)),
+		.fingers = libinput_event_gesture_get_finger_count(gevent),
+		.dx = libinput_event_gesture_get_dx(gevent),
+		.dy = libinput_event_gesture_get_dy(gevent),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.swipe_update, &wlr_event);
+}
+
+void handle_pointer_swipe_end(struct libinput_event *event,
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_swipe_end wlr_event = {
+		.device = wlr_dev,
+		.time_msec =
+			usec_to_msec(libinput_event_gesture_get_time_usec(gevent)),
+		.cancelled = libinput_event_gesture_get_cancelled(gevent),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.swipe_end, &wlr_event);
+}
+
+void handle_pointer_pinch_begin(struct libinput_event *event,
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_pinch_begin wlr_event = {
+		.device = wlr_dev,
+		.time_msec =
+			usec_to_msec(libinput_event_gesture_get_time_usec(gevent)),
+		.fingers = libinput_event_gesture_get_finger_count(gevent),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.pinch_begin, &wlr_event);
+}
+
+void handle_pointer_pinch_update(struct libinput_event *event,
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_pinch_update wlr_event = {
+		.device = wlr_dev,
+		.time_msec =
+			usec_to_msec(libinput_event_gesture_get_time_usec(gevent)),
+		.fingers = libinput_event_gesture_get_finger_count(gevent),
+		.dx = libinput_event_gesture_get_dx(gevent),
+		.dy = libinput_event_gesture_get_dy(gevent),
+		.scale = libinput_event_gesture_get_scale(gevent),
+		.rotation = libinput_event_gesture_get_angle_delta(gevent),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.pinch_update, &wlr_event);
+}
+
+void handle_pointer_pinch_end(struct libinput_event *event,
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_pinch_end wlr_event = {
+		.device = wlr_dev,
+		.time_msec =
+			usec_to_msec(libinput_event_gesture_get_time_usec(gevent)),
+		.cancelled = libinput_event_gesture_get_cancelled(gevent),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.pinch_end, &wlr_event);
+}
--- a/include/wlr/types/wlr_pointer_gestures_v1.h
+++ b/include/wlr/types/wlr_pointer_gestures_v1.h
@@ -1,0 +1,73 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_POINTER_GESTURES_V1_H
+#define WLR_TYPES_WLR_POINTER_GESTURES_V1_H
+
+#include <wayland-server.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_surface.h>
+
+struct wlr_pointer_gestures_v1 {
+	struct wl_global *global;
+	struct wl_list resources; // wl_resource_get_link
+	struct wl_list swipes;    // wl_resource_get_link
+	struct wl_list pinches;   // wl_resource_get_link
+
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_pointer_gestures_v1 *wlr_pointer_gestures_v1_create(
+	struct wl_display *display);
+
+void wlr_pointer_gestures_v1_send_swipe_begin(
+	struct wlr_pointer_gestures_v1 *gestures,
+	struct wlr_seat *seat,
+	uint32_t time_msec,
+	uint32_t fingers);
+void wlr_pointer_gestures_v1_send_swipe_update(
+	struct wlr_pointer_gestures_v1 *gestures,
+	struct wlr_seat *seat,
+	uint32_t time_msec,
+	double dx,
+	double dy);
+void wlr_pointer_gestures_v1_send_swipe_end(
+	struct wlr_pointer_gestures_v1 *gestures,
+	struct wlr_seat *seat,
+	uint32_t time_msec,
+	bool cancelled);
+
+void wlr_pointer_gestures_v1_send_pinch_begin(
+	struct wlr_pointer_gestures_v1 *gestures,
+	struct wlr_seat *seat,
+	uint32_t time_msec,
+	uint32_t fingers);
+void wlr_pointer_gestures_v1_send_pinch_update(
+	struct wlr_pointer_gestures_v1 *gestures,
+	struct wlr_seat *seat,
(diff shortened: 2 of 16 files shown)


include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,83 @@
...
+#endif
+
+#ifndef WLR_TYPES_WLR_POINTER_GESTURES_V1_H
+#define WLR_TYPES_WLR_POINTER_GESTURES_V1_H
+
+#include <wayland-server.h>
+#include <wlr/types/wlr_pointer.h>
+
+struct wlr_pointer_gestures_v1 {
+	struct wl_global *wl_global;
```
emersion: This should be named `global`.

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,83 @@
...
+	struct wl_list resources;
+	struct wl_list swipes; // wlr_pointer_gesture_swipe_v1::link
```
emersion: Could these two lists just be lists of `wl_resource`?

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,83 @@
...
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void* data;
+};
+
+void wlr_pointer_gesture_swipe_notify_begin(
+	struct wlr_pointer_gesture_swipe_v1 *swipe,
```
emersion: It would make more sense to take a `struct wlr_pointer_gestures_v1 *` instead.

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,83 @@
...
+	struct wl_resource *surface,
```
emersion: We try not to use `wl_resource` in our APIs, because these are not type-checked to be `wl_surface` instances. Instead, the focused surface can be retrieved directly from the `wlr_seat`.

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,83 @@
...
+	struct wlr_event_pointer_swipe_begin *event);
```
emersion: It would be more consistent with other seat functions (see e.g. `wlr_seat_pointer_send_button`, `wlr_seat_pointer_send_axis`) to explicitly use multiple arguments instead of a `wlr_event_pointer_swipe_begin`. This would make it easier to simulate these events too.

rootston/seat.c
```diff
@@ -77,6 +77,78 @@ static void handle_cursor_axis(struct wl_listener *listener, void *data) {
...
 }
 
+static void handle_swipe_begin(struct wl_listener *listener, void *data) {
+	struct roots_cursor *cursor =
+		wl_container_of(listener, cursor, swipe_begin);
+	struct wlr_pointer_gestures_v1 *gestures = cursor->seat->input->server->desktop->pointer_gestures;
+	struct wlr_event_pointer_swipe_begin *event = data;
+	struct wl_resource *focus = roots_seat_get_focus(cursor->seat)->wlr_surface->resource;
+	struct wlr_pointer_gesture_swipe_v1 *swipe;
+	wl_list_for_each(swipe, &gestures->swipes, link) {
```
emersion: If two clients create a `zwp_pointer_gesture_swipe_v1` object, both will receive swipe events regardless of which one is focused. Additionally, the one that isn't focused will receive events with an unknown surface (because it belongs to a different client).

backend/libinput/pointer.c
```diff
@@ -136,3 +136,117 @@ void handle_pointer_axis(struct libinput_event *event,
...
+		struct libinput_device *libinput_dev) {
+	struct wlr_input_device *wlr_dev =
+		get_appropriate_device(WLR_INPUT_DEVICE_POINTER, libinput_dev);
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer gesture event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_gesture *gevent =
+		libinput_event_get_gesture_event(event);
+	struct wlr_event_pointer_swipe_update wlr_event = { 0 };
```
emersion: You can use struct initializers here.

```c
struct wlr_event_pointer_swipe_update wlr_event = {
	.device = wlr_dev,
	…
};
```

(The rest of the file should be fixed, but this probably belongs to another PR)

include/wlr/types/wlr_pointer_gestures_v1.h
myfreeweb: Well, I don't really need the `wlr_pointer_gesture_swipe_v1` objects at all, would it be okay to just get rid of them?

myfreeweb:
Updated with suggestions, also switched to actually using pointer focus instead of keyboard focus in rootston, because gestures should apply to where the cursor is.

> how should this integrate with grabs?

Comment on the [Weston patch](https://gitlab.freedesktop.org/myfreeweb/weston/commit/2fd0c8bdca3a8ac7711ec22b1291053203fc50fa):

> It could be deemed convenient in the future that gesture events go through the pointer grab interface. It could be useful in case we want gesture support in any weston_pointer_grab_interface, it is just not needed at the moment, so has been left aside.

I'm not sure what would be a use case for grabbing gestures. I've used Apple Macs for years, and there I've only encountered gestures that always go to the window below the cursor (pinch-zoom, two-finger swipe) and gestures that always go to the windowing system (four finger swipe up to reveal workspace overview, four finger swipe sideways to change workspaces, etc.) I don't recall ever seeing gestures that temporarily act differently depending on something.

include/wlr/types/wlr_pointer_gestures_v1.h
emersion: +1

myfreeweb: done

myfreeweb:
Hm, just had an infinite `wl_resource_for_each_safe` loop in `client_handle_destroy`… what am I doing wrong?

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,274 @@
...
+		calloc(1, sizeof(struct wlr_pointer_gestures_v1));
+	if (!gestures) {
+		return NULL;
+	}
+
+	wl_list_init(&gestures->resources);
+	wl_list_init(&gestures->swipes);
+	wl_list_init(&gestures->pinches);
+
+	gestures->display_destroy.notify = handle_display_destroy;
```
emersion: This should be done after creating the global. If the allocation fails, there will be a dangling listener otherwise.

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,274 @@
...
+static void handle_display_destroy(struct wl_listener *listener, void *data) {
+	struct wlr_pointer_gestures_v1 *tablet =
+		wl_container_of(listener, tablet, display_destroy);
+	wlr_pointer_gestures_v1_destroy(tablet);
+}
+
+void wlr_pointer_gestures_v1_destroy(struct wlr_pointer_gestures_v1 *gestures) {
+	struct wl_resource *resource, *tmp;
+	wl_resource_for_each_safe(resource, tmp, &gestures->resources) {
+		wl_resource_destroy(resource);
```
emersion: Should do the same for swipes and pinches

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,274 @@
...
+
+	struct wl_resource *resource = wl_resource_create(wl_client,
+			&zwp_pointer_gestures_v1_interface, version, id);
+	if (resource == NULL) {
+		wl_client_post_no_memory(wl_client);
+		return;
+	}
+
+	wl_resource_set_implementation(resource,
+			&gestures_impl, gestures, client_handle_destroy);
```
emersion: This doesn't happen when the client is destroyed, this happens when the resource is. This should be named `gestures_handle_resource_destroy`.

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,274 @@
...
+	.get_pinch_gesture = get_pinch_gesture,
+};
+
+static void client_handle_destroy(struct wl_resource *resource) {
+	struct wl_client *client = wl_resource_get_client(resource);
+	struct wlr_pointer_gestures_v1 *gestures =
+		pointer_gestures_from_resource(resource);
+	struct wl_resource *gesture, *tmp = NULL;
+	wl_resource_for_each_safe(gesture, tmp, &gestures->swipes) {
+		if (wl_resource_get_client(gesture) == client) {
```
emersion: This doesn't work if the client binds twice to the gestures global. Moreover, the protocol doesn't say swipe and pinch objects become invalid when the gestures object is destroyed. I think we shouldn't destroy these here.

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,274 @@
...
+
+void wlr_pointer_gesture_pinch_notify_end(
+		struct wlr_pointer_gestures_v1 *gestures,
+		struct wlr_surface *focus,
+		uint32_t time_msec,
+		bool cancelled) {
+	struct wl_resource *pinch;
+	wl_resource_for_each(pinch, &gestures->pinches) {
+		struct wl_client *pinch_client = wl_resource_get_client(pinch);
+		if (pinch_client != wl_resource_get_client(focus->resource)) {
```
emersion: This doesn't respect seats. The `wl_pointer` object used to create the pinch object is bound to a seat.

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,73 @@
...
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_pointer_gestures_v1 *wlr_pointer_gestures_v1_create(
+	struct wl_display *display);
+
+void wlr_pointer_gesture_swipe_notify_begin(
```
emersion: We should probably name these `wlr_pointer_send_gesture_swipe_begin`, because `notify` is reserved for functions respecting grabs.

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,73 @@
...
+	struct wlr_pointer_gestures_v1 *gestures,
+	struct wlr_surface *focus,
```
emersion: I'd like not to make the compositor responsible for deciding which surface these events are sent to. Events should only be sent to the currently focused surface. So we should remove the surface argument and replace it with a seat argument. The focused surface is in `wlr_seat.pointer_state.focused_surface`.

types/wlr_pointer_gestures_v1.c
myfreeweb: All of the client's resources are destroyed (with destructors run) when it disconnects, right?

emersion: Yes. The order of destruction isn't guaranteed. Also, there is one other situation when the resource destructor is called: when the client sends the `destroy` request on an object. So the client can e.g. create bind twice to the same global, then destroy one.

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,288 @@
...
+
+static void get_pinch_gesture(struct wl_client *client,
+		struct wl_resource *gestures_resource,
+		uint32_t id,
+		struct wl_resource *pointer_resource) {
+	struct wlr_pointer_gestures_v1 *gestures =
+		pointer_gestures_from_resource(gestures_resource);
+
+	struct wlr_seat *seat =
+		wlr_seat_client_from_pointer_resource(pointer_resource)->seat;
```
emersion: This can be NULL in case the pointer resource is inert (ie. the seat has disappeared but the client hasn't destroyed the pointer yet).

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,288 @@
...
+		bool cancelled) {
+	struct wl_resource *focus = seat->pointer_state.focused_surface->resource;
+	struct wl_resource *gesture;
+	wl_resource_for_each(gesture, &gestures->pinches) {
+		struct wlr_seat *gesture_seat = wlr_seat_from_pointer_resource(gesture);
+		struct wl_client *gesture_client = wl_resource_get_client(gesture);
+		if (gesture_seat != seat || gesture_client != wl_resource_get_client(focus)) {
+			continue;
+		}
+		uint32_t serial = wl_display_next_serial(
```
emersion: Maybe we can use one serial for everybody?

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,288 @@
...
+static const struct zwp_pointer_gesture_pinch_v1_interface pinch_impl;
+
+static struct wlr_pointer_gestures_v1 *pointer_gestures_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource,
+				&zwp_pointer_gestures_v1_interface, &gestures_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static struct wlr_seat *wlr_seat_from_pointer_resource(
```
emersion: We try to use the `wlr_` prefix only for exported functions. So it would be better to rename this `seat_from_pointer_resource`.

emersion:
This is looking pretty good, nice work. Will test tomorrow :)

types/wlr_pointer_gestures_v1.c
myfreeweb: Is that fine? I assumed serials must be globally unique…

types/wlr_pointer_gestures_v1.c
myfreeweb: I don't think that would cause any problems, the gesture resource will be created with `NULL` data, and that won't be dereferenced, just compared against the seat argument. So if the client creates a gesture object for a pointer from a disappearing seat, it will just be an object that doesn't deliver any events

emersion: If `wlr_seat_client_from_pointer_resource(pointer_resource)` is NULL, `->seat` will be something a little higher like 0x10. It would be silly to have invalid pointers that aren't NULL.

types/wlr_pointer_gestures_v1.c
emersion: The serial represents the event. It's used as a "timestamp" to prevent races, and also as a token. We'll send it to a single client, so it's fine to generate only one. In fact, it would simplify serial validation, if we need that in the future.

types/wlr_pointer_gestures_v1.c
myfreeweb: Oh, the seat_client *itself* can be NULL? Okay, `->seat` will be a segfault then

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,288 @@
...
+
+static const struct zwp_pointer_gestures_v1_interface gestures_impl = {
+	.get_swipe_gesture = get_swipe_gesture,
+	.get_pinch_gesture = get_pinch_gesture,
+};
+
+static void pointer_gestures_v1_bind(struct wl_client *wl_client, void *data,
+		uint32_t version, uint32_t id) {
+	struct wlr_pointer_gestures_v1 *gestures = data;
+	assert(wl_client && gestures);
```
emersion: Nitpick: libwayland already guarantees `wl_client != NULL`. Not sure the assertion about `gestures` has value.

include/wlr/types/wlr_pointer_gestures_v1.h
```diff
@@ -0,0 +1,73 @@
...
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_pointer_gestures_v1 *wlr_pointer_gestures_v1_create(
+	struct wl_display *display);
+
+void wlr_pointer_send_gesture_swipe_begin(
```
emersion: This should be named `wlr_pointer_gestures_v1_send_swipe_begin`.

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,288 @@
...
+				&zwp_pointer_gesture_pinch_v1_interface, &pinch_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+void wlr_pointer_send_gesture_swipe_begin(
+		struct wlr_pointer_gestures_v1 *gestures,
+		struct wlr_seat *seat,
+		uint32_t time_msec,
+		uint32_t fingers) {
+	struct wl_resource *focus = seat->pointer_state.focused_surface->resource;
```
emersion: The focused surface can be NULL.

types/wlr_pointer_gestures_v1.c
```diff
@@ -0,0 +1,335 @@
...
+
+static void get_swipe_gesture(struct wl_client *client,
+		struct wl_resource *gestures_resource,
+		uint32_t id,
+		struct wl_resource *pointer_resource) {
+	struct wlr_seat_client *seat_client =
+		wlr_seat_client_from_pointer_resource(pointer_resource);
+
+	if (seat_client == NULL) {
+		wl_client_post_no_memory(client);
```
myfreeweb: The protocol doesn't define any errors, so idk what else to do here, just an error on the gestures_resource?

myfreeweb: oh, right.

myfreeweb:
> I don't have the hardware to try this out

huh, Synaptics & co have been making multi-touch trackpads for *ages*… are you sure that on laptops you have access to, `sudo libinput list-devices | grep gesture` returns nothing?

emersion:
> huh, Synaptics & co have been making multi-touch trackpads for ages… are you sure that on laptops you have access to, sudo libinput list-devices | grep gesture returns nothing?

Yeah. I have a ThinkPad T420.

```
Device:           SynPS/2 Synaptics TouchPad
Kernel:           /dev/input/event8
Group:            8
Seat:             seat0, default
Size:             59x32mm
Capabilities:     pointer 
Tap-to-click:     disabled
Tap-and-drag:     enabled
Tap drag lock:    disabled
Left-handed:      disabled
Nat.scrolling:    disabled
Middle emulation: n/a
Calibration:      n/a
Scroll methods:   *two-finger edge 
Click methods:    none
Disable-w-typing: enabled
Accel profiles:   none
Rotation:         n/a
```

emersion:
Thanks!

PR #1511, opened 2019-01-29
Fix two issues found with Clang's static analyzer

emersion:
Clang still generates one false-positive:

```
../../../rootston/desktop.c:925:3: warning: 3rd function call argument is an uninitialized value
                roots_cursor_constrain(seat->cursor, wlr_constraint, sx, sy);
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


ddevault:
Thanks!

PR #1512, opened 2019-01-29
meson: remove b_lundef=false on FreeBSD

emersion:
This was necessary because we were using `environ` in the Xwayland code, but it has been removed now. It's better not to use these flags so that the linker errors out if we're using an undefined function (e.g. we include the header but forget to link against it).


ddevault:
Thanks!

Issue #1288, opened 2018-10-07
Drag-and-drop doesn't send motion events


emersion:
1. Open two `weston-dnd` instances
2. Drag a flower from one instance to the other
3. Flower is dropped where the cursor entered, not where the button was released

This happens because recently grab semantics have been changed (I suspect after @Ongy's work on tablets). When you don't call `wlr_seat_pointer_enter` in the `enter` grab handler then you always get `enter` events and you don't get any `motion` event.

For drag-and-drop, we don't want to change the pointer focus but we want to receive motion events relative to the surface under the cursor.

emersion:
This is the commit breaking drag-and-drop in rootston: https://github.com/swaywm/wlroots/commit/2e6eb097b6e23b8923bbfc68b1843d5ccde1955b

PR #1513, opened 2019-01-29
data-device: fix drag-and-drop

emersion:
Test plan: try out weston-dnd.

~~Depends on #1515~~
Fixes https://github.com/swaywm/wlroots/issues/1288

### DIFF
--- a/types/data_device/wlr_drag.c
+++ b/types/data_device/wlr_drag.c
@@ -122,6 +122,9 @@ static void drag_end(struct wlr_drag *drag) {
 			drag_icon_set_mapped(drag->icon, false);
 		}
 
+		assert(drag->seat->drag == drag);
+		drag->seat->drag = NULL;
+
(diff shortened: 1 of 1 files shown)


johnchen902:
Got lots of crashes today after updating from swaywm/sway@68a28e48 and 018727b1 to swaywm/sway@4f4424f6 and 77c25c15. Do not know how to reproduce.

Do you think this pull request will fix the crashes? Should I open a separate issue?

Here are two of the crashes today:

Crash 4: [sway-19223.core.gz](https://github.com/swaywm/wlroots/files/2811735/sway-19223.core.gz)

Crash 5: [sway-valgrind-17048.log](https://github.com/swaywm/wlroots/files/2811687/sway-valgrind-17048.log) (with `i965_dri.so` suppressed)

emersion:
Seems like it.

emersion:
This is now ready for review.

ddevault:
Thanks!

PR #1515, opened 2019-01-30
Add missing frame events to shell pointer grabs

emersion:
Test plan: open gnome-calculator, switch to "Programming mode", click on the "Binary/Octal/Decimal/Hexadecimal" combo button. Options should be highlighted as the pointer is moved.


ddevault:
Thanks!

PR #1520, opened 2019-01-31
data-control: Primary selection support

YaLTeR:
See https://github.com/swaywm/wlr-protocols/pull/34

There's quite a lot of changes in this PR, for two reasons:
1. There was an issue with the current implementation where data sources didn't track their usage correctly. Setting a data source as selection for two seats and subsequently changing selection of one of the seats freed the data source, even though it was still used for the second seat. Alternatively, setting a data source as selection for a single seat twice in a row would trigger this bug too (it would free the data source and then set the freed data source to the seat). The first commit adds appropriate tracking mechanisms to address this.
2. While data-control works in terms of `wlr_data_source`, the primary selection works in terms of `wlr_primary_selection_source` with no obvious way of converting between the two. Following @emersion's advise on the IRC, I changed the logic to have an internal data source object and create `wlr_data_source` and `wlr_primary_selection_source` objects as necessary. This also turned out to be required for the fix in the first point.

Anyway, I bombarded rootston with all the weird `wl-copy` and `wl-paste` invocations I could think of and fixed all of the crashes I encountered, so this should work.

This is my first experience with the wlroots code base, so while I read CONTRIBUTING.md, there can very well be some parts I changed in unidiomatic ways, which I'll be happy to fix if pointed out.

cc @emersion @bugaevc

Test version of wl-clipboard: https://github.com/bugaevc/wl-clipboard/tree/data-control-primary

Hack to make it work (currently has an issue):
```diff
diff --git a/src/wl-copy.c b/src/wl-copy.c
index 395c4cc..fa9a12b 100644
--- a/src/wl-copy.c
+++ b/src/wl-copy.c
@@ -300,10 +300,10 @@ void init_selection(char *mime_type) {
 }
 
 void init_primary_selection(char *mime_type) {
-    ensure_has_primary_selection();
+    /* ensure_has_primary_selection(); */
 
 #ifdef HAVE_WLR_DATA_CONTROL
-    if (data_control_supports_primary_selection) {
+    /* if (data_control_supports_primary_selection) { */
         struct zwlr_data_control_source_v1 *data_control_source =
             create_data_control_source(mime_type);
 
@@ -312,7 +312,7 @@ void init_primary_selection(char *mime_type) {
             data_control_source
         );
         return;
-    }
+    /* } */
 #endif
 
 #ifdef HAVE_WP_PRIMARY_SELECTION
diff --git a/src/wl-paste.c b/src/wl-paste.c
index 962fe3d..46fc03b 100644
--- a/src/wl-paste.c
+++ b/src/wl-paste.c
@@ -445,17 +445,17 @@ void init_selection() {
 }
 
 void init_primary_selection() {
-    ensure_has_primary_selection();
+    /* ensure_has_primary_selection(); */
 
 #ifdef HAVE_WLR_DATA_CONTROL
-    if (data_control_supports_primary_selection) {
+    /* if (data_control_supports_primary_selection) { */
         zwlr_data_control_device_v1_add_listener(
             data_control_device,
             &data_control_device_listener,
             NULL
         );
         return;
-    }
+    /* } */
 #endif
 
 #ifdef HAVE_WP_PRIMARY_SELECTION
```

### DIFF
--- a/types/wlr_data_control_v1.c
+++ b/types/wlr_data_control_v1.c
@@ -5,51 +5,26 @@
 #include <unistd.h>
 #include <wlr/types/wlr_data_control_v1.h>
 #include <wlr/types/wlr_data_device.h>
+#include <wlr/types/wlr_primary_selection.h>
 #include <wlr/util/log.h>
 #include "util/signal.h"
 #include "wlr-data-control-unstable-v1-protocol.h"
 
-#define DATA_CONTROL_MANAGER_VERSION 1
+#define DATA_CONTROL_MANAGER_VERSION 2
 
-struct client_data_source {
-	struct wlr_data_source source;
+struct data_control_source {
 	struct wl_resource *resource;
+	struct wl_array mime_types;
 	bool finalized;
-};
 
-static const struct wlr_data_source_impl client_source_impl;
-
-static struct client_data_source *
-		client_data_source_from_source(struct wlr_data_source *wlr_source) {
-	assert(wlr_source->impl == &client_source_impl);
-	return (struct client_data_source *)wlr_source;
-}
-
-static void client_source_send(struct wlr_data_source *wlr_source,
-		const char *mime_type, int fd) {
-	struct client_data_source *source =
-		client_data_source_from_source(wlr_source);
-	zwlr_data_control_source_v1_send_send(source->resource, mime_type, fd);
-	close(fd);
-}
-
-static void client_source_destroy(struct wlr_data_source *wlr_source) {
-	struct client_data_source *source =
-		client_data_source_from_source(wlr_source);
-	zwlr_data_control_source_v1_send_cancelled(source->resource);
-	// Make the resource inert
-	wl_resource_set_user_data(source->resource, NULL);
-	free(source);
-}
-
-static const struct wlr_data_source_impl client_source_impl = {
-	.send = client_source_send,
-	.destroy = client_source_destroy,
+	// Only one of these is non-NULL.
+	struct wlr_data_source *active_source;
+	struct wlr_primary_selection_source *active_primary_source;
 };
 
 static const struct zwlr_data_control_source_v1_interface source_impl;
 
-static struct client_data_source *source_from_resource(
+static struct data_control_source *source_from_resource(
 		struct wl_resource *resource) {
 	assert(wl_resource_instance_of(resource,
 		&zwlr_data_control_source_v1_interface, &source_impl));
@@ -58,7 +33,7 @@ static struct client_data_source *source_from_resource(
 
 static void source_handle_offer(struct wl_client *client,
 		struct wl_resource *resource, const char *mime_type) {
-	struct client_data_source *source = source_from_resource(resource);
+	struct data_control_source *source = source_from_resource(resource);
 	if (source == NULL) {
 		return;
 	}
@@ -66,12 +41,13 @@ static void source_handle_offer(struct wl_client *client,
 	if (source->finalized) {
 		wl_resource_post_error(resource,
 			ZWLR_DATA_CONTROL_SOURCE_V1_ERROR_INVALID_OFFER,
-			"cannot mutate offer after set_selection");
+			"cannot mutate offer after set_selection or "
+			"set_primary_selection");
 		return;
 	}
 
 	const char **mime_type_ptr;
-	wl_array_for_each(mime_type_ptr, &source->source.mime_types) {
+	wl_array_for_each(mime_type_ptr, &source->mime_types) {
 		if (strcmp(*mime_type_ptr, mime_type) == 0) {
 			wlr_log(WLR_DEBUG, "Ignoring duplicate MIME type offer %s",
 				mime_type);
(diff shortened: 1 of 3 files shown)


emersion:
>Setting a data source as selection for two seats

I don't think we want to allow this, this makes things waaaay too complicated. I've sent a message on the mailing list since it's not clear in the spec: https://lists.freedesktop.org/archives/wayland-devel/2019-January/039936.html

YaLTeR:
Only allowing a data source to participate in a single selection (regular or primary) would make things 10x easier.

emersion:
Let's just not allow it in our protocol anyway. I can't see any upsides to letting clients do this.

YaLTeR:
Rebased to master and updated according to not allowing a data source to participate in multiple selections.

YaLTeR:
Fixed a typo.

YaLTeR:
Updated according to comment in the protocol PR.

types/wlr_data_control_v1.c
```diff
@@ -5,51 +5,26 @@
...
-	// Make the resource inert
-	wl_resource_set_user_data(source->resource, NULL);
-	free(source);
-}
-
-static const struct wlr_data_source_impl client_source_impl = {
-	.send = client_source_send,
-	.destroy = client_source_destroy,
+	// Only one of these is non-NULL.
+	struct wlr_data_source *active_source;
```
emersion: Do we really need to keep the `data_control_source` around after `set_selection`?

YaLTeR: Well, it can receive the destroy event, in which case it must clean up the underlying `wlr_data_source` or `wlr_primary_selection_source`.

emersion: Err, right.

types/wlr_data_control_v1.c
```diff
@@ -106,17 +82,134 @@ static const struct zwlr_data_control_source_v1_interface source_impl = {
...
 
 static void source_handle_resource_destroy(struct wl_resource *resource) {
-	struct client_data_source *source = source_from_resource(resource);
+	struct data_control_source *source = source_from_resource(resource);
 	if (source == NULL) {
 		return;
 	}
-	wlr_data_source_destroy(&source->source);
+
+	char **p;
```
emersion: Can we move all of this into a `data_control_source_destroy` function? This would allow to call it from here and other places without using `wl_resource`, which is not type-safe.

types/wlr_data_control_v1.c
```diff
@@ -144,22 +258,78 @@ static const struct zwlr_data_control_offer_v1_interface offer_impl = {
...
+	wl_array_for_each(p, &source->mime_types) {
+		zwlr_data_control_offer_v1_send_offer(resource, *p);
+	}
+
+	return resource;
+}
+
+static struct wl_resource *create_primary_offer(
```
emersion: This can probably be merged with `create_offer`

types/wlr_data_control_v1.c
YaLTeR: Other places would still need to destroy the resource though which ends up running this code anyway. What would extracting it achieve?

emersion: No, you should never destroy resources yourself. `wl_resource_destroy` should only be called in the destroy request handler.

Extracting it would make sure a completely difference resource isn't passed.

bugaevc: > using `wl_resource`, which is not type-safe
(Completely unrelated to this pull request, but) do you happen to know why wayland-server is designed this way, especially given how wayland-client is typesafe (with all objects still being generic wl_proxies under the hood)?

emersion: It seems to be historical raisins, according to Pekka. No real reason.

YaLTeR: > Can we move all of this into a `data_control_source_destroy` function?

Something like this?

types/wlr_data_control_v1.c
```diff
@@ -105,18 +81,144 @@ static const struct zwlr_data_control_source_v1_interface source_impl = {
 	.destroy = source_handle_destroy,
 };
 
+static void data_control_source_destroy(struct data_control_source* source) {
```
emersion: Style: `*` on the right

types/wlr_data_control_v1.c
```diff
@@ -105,18 +81,144 @@ static const struct zwlr_data_control_source_v1_interface source_impl = {
...
 static void source_handle_resource_destroy(struct wl_resource *resource) {
-	struct client_data_source *source = source_from_resource(resource);
+	struct data_control_source *source = source_from_resource(resource);
+	if (source == NULL) {
+		return;
+	}
+
+	// Prevent destructors called from data_control_source_destroy() from
+	// accessing invalid data.
+	wl_resource_set_user_data(resource, NULL);
```
emersion: This should be done in `data_control_source_destroy`. See the CONTRIBUTING guide about inert resources.

types/wlr_data_control_v1.c
```diff
@@ -232,6 +461,38 @@ static void control_send_selection(struct wlr_data_control_device_v1 *device) {
...
+	if (version < ZWLR_DATA_CONTROL_DEVICE_V1_PRIMARY_SELECTION_SINCE_VERSION) {
+		return;
+	}
+
+	struct wlr_primary_selection_source *source =
+		device->seat->primary_selection_source;
+
+	if (device->primary_selection_offer_resource != NULL) {
+		// Make the offer inert
+		free(data_offer_from_offer_resource(
```
emersion: This should be extracted to a separate function.

YaLTeR:
Done.

emersion:
Thanks!

PR #1527, opened 2019-02-03
Fix another instance of swaywm/sway#3545.

johnchen902:
Fixes swaywm/sway#3545 should it be reopened.


emersion:
Hmmm, something is going wrong in the drag implementation, we shouldn't need to do that. Will investigate.

PR #1528, opened 2019-02-04
Add build-*/ to gitignore.

c-edw:
This brings it in line with Sway's gitignore.

The original justification was that it allows for simultaneous gcc and clang builds.


emersion:
Thanks!

PR #1529, opened 2019-02-05
data-device: destroy previous source when starting drag

emersion:
This supersedes f24e17259e49aef55b7ada54793a4cdb49ae94a1 and
04c9ca4198a729a95a6368bbbf0438d1ba3465fa. These commits were manually removing
wlr_data_source destroy handlers when starting a new drag. This is error-prone.

Instead, this commit destroys the previous source whenever we start a new drag.

### DIFF
--- a/xwayland/selection/dnd.c
+++ b/xwayland/selection/dnd.c
@@ -315,8 +315,6 @@ static void seat_handle_drag_source_destroy(struct wl_listener *listener,
 		wl_container_of(listener, xwm, seat_drag_source_destroy);
 
 	wl_list_remove(&xwm->seat_drag_source_destroy.link);
-	xwm->seat_drag_source_destroy.link.prev = NULL;
-	xwm->seat_drag_source_destroy.link.next = NULL;
 	xwm->drag_focus = NULL;
 }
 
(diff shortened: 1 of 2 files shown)


xwayland/selection/dnd.c
```diff
@@ -334,9 +332,6 @@ void xwm_seat_handle_start_drag(struct wlr_xwm *xwm, struct wlr_drag *drag) {
 		wl_signal_add(&drag->events.destroy, &xwm->seat_drag_destroy);
 		xwm->seat_drag_destroy.notify = seat_handle_drag_destroy;
 
-		if (xwm->seat_drag_source_destroy.link.prev != NULL) {
-			wl_list_remove(&xwm->seat_drag_source_destroy.link);
-		}
```
johnchen902: How about an assertion? I think it's fragile to assume `xwm->seat_drag_source_destroy` is not listening to any signal. A change to `seat_client_start_drag` may break this again.

emersion: Nope, because when the data source is free'ed, all listeners are guaranteed to have been removed (otherwise they are corrupt). In other words, we require all listeners to remove themselves in the `destroy` signal handler.

johnchen902: Yes, it works for now. But it may not work in the future as someone may change `seat_client_start_drag` (e.g. somebody need the source alive even after drag).

emersion: In this case there can be multiple `seat->drag_source` at a time, and listeners shouldn't assume there's only one. In this case, `seat->drag_source` would probably need to be changed to be a list.

Another solution would be to keep only a single `seat->drag_source`, but to add a new `set_drag_source` signal to the seat, emitted when `seat->drag_source` is set. I'm not sure if this makes a lot of sense though.

johnchen902: Alternatively, add some comment around `wlr_data_source_destroy(seat->drag_source);` to make sure anyone touching it is aware of the impact here.

emersion: Done

emersion:
This PR has been confirmed to work here: https://github.com/swaywm/sway/issues/3545#issuecomment-460624088

ddevault:
Thanks!

PR #1530, opened 2019-02-05
xdg-shell: destroy child popups on unmap

emersion:
It doesn't make sense to keep popups opened when unmapped. We also need to do
so in wlr_xdg_popup_destroy so that popups are destroyed in the correct order.


ddevault:
Thanks!

PR #1532, opened 2019-02-08
Mark old protocols as obsolete

emersion:
All of those have better replacements.

Not sure "deprecated" is the correct wording.


rkanati:
Maybe "Retained for compatibility" or "Not recommended for new applications"?

ddevault:
Proposed language:

```
/*
 * This protocol is obsolete and will be removed in a future version. The
 * recommended replacement is <x>.
 */
```

emersion:
Yeah, that's a more accurate statement. Updated.

ddevault:
Thanks!

PR #1533, opened 2019-02-11
Add a wlr_output_set_subpixel()

ggreer:
drmModeConnector.subpixel doesn't seem to detect subpixel order on many displays (especially laptops). Allow subpixel order to be manually set.

My plan is to make a corresponding PR for sway that adds a subpixel output option. Then https://github.com/swaywm/sway/issues/3163 will be fixed.

I'm very unfamiliar with the wlroots codebase so it's quite likely I made a mistake in this PR.

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -174,6 +174,7 @@ void wlr_output_set_transform(struct wlr_output *output,
 	enum wl_output_transform transform);
 void wlr_output_set_position(struct wlr_output *output, int32_t lx, int32_t ly);
 void wlr_output_set_scale(struct wlr_output *output, float scale);
+void wlr_output_set_subpixel(struct wlr_output *output, enum wl_output_subpixel subpixel);
 void wlr_output_destroy(struct wlr_output *output);
 /**
  * Computes the transformed output resolution.
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -248,6 +248,19 @@ void wlr_output_set_scale(struct wlr_output *output, float scale) {
 	wlr_signal_emit_safe(&output->events.scale, output);
 }
 
+void wlr_output_set_subpixel(struct wlr_output *output, enum wl_output_subpixel subpixel) {
+	if (output->subpixel == subpixel) {
+		return;
+	}
+
+	output->subpixel = subpixel;
+
+	struct wl_resource *resource;
+	wl_resource_for_each(resource, &output->resources) {
+		output_send_to_resource(resource);
+	}
+}
+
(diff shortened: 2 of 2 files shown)


include/wlr/util/subpixel.h
```diff
@@ -0,0 +1,8 @@
+#ifndef UTIL_SUBPIXEL_H
+#define UTIL_SUBPIXEL_H
+
+#include <wayland-server-protocol.h>
+
+const char *wlr_wl_output_subpixel_to_string(enum wl_output_subpixel subpixel);
```
emersion: We don't use this anywhere, so we might as well remove it.

types/wlr_output_damage.c
```diff
@@ -31,6 +31,12 @@ static void output_handle_scale(struct wl_listener *listener, void *data) {
 	wlr_output_damage_add_whole(output_damage);
 }
 
+static void output_handle_subpixel(struct wl_listener *listener, void *data) {
+	struct wlr_output_damage *output_damage =
+		wl_container_of(listener, output_damage, output_subpixel);
+	wlr_output_damage_add_whole(output_damage);
```
emersion: I don't think this is necessary.

include/wlr/util/subpixel.h
ggreer: I'm using it in https://github.com/swaywm/sway/pull/3645. I figured it would be better to put it in wlroots since other window managers would likely want to log the subpixel order.

If you still think it should be removed, I'll move it to my sway PR.

emersion: Yeah, I'd prefer to move this to sway.

types/wlr_output_damage.c
ggreer: I damaged the whole screen to get commands like `swaymsg output eDP-1 subpixel rgb` to update the hinting immediately. Without this I'd have to resize/quit/move apps for the subpixel hinting to update in title bars.

emersion: I see. This should be in sway then, because wlroots compositors might not need that. For instance rootston doesn't need to re-draw anything when subpixel changes.

emersion:
I'm a little sad to see subpixel becoming a dynamic thing. Also I'm not sure how Wayland clients will handle this (some will probably completely ignore the updated value).

ggreer:
I don't know how they implement it, but macOS & Windows both allow changing subpixel hinting without restarting anything. On both platforms, some apps pick up the new hinting immediately while others require relaunching.

I managed to get the same behavior with this PR + the Sway one. Thanks to your suggestions, it's fewer total lines of code.

emersion:
Thanks!

PR #1534, opened 2019-02-11
xwm: don't steal focus in xwayland_surface_create

RedSoxFan:
Fixes swaywm/sway#2899

The surface created in xwayland_surface_create would steal focus due to
being stacked on top. This changes is so that focus is restored.

After testing this with sway, I was not able to find any issues, but it is
possible that I missed something.

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -112,6 +112,9 @@ static int xwayland_surface_handle_ping_timeout(void *data) {
 	return 1;
 }
 
+static void xwm_send_focus_window(struct wlr_xwm *xwm,
+		struct wlr_xwayland_surface *xsurface);
+
 static struct wlr_xwayland_surface *xwayland_surface_create(
 		struct wlr_xwm *xwm, xcb_window_t window_id, int16_t x, int16_t y,
 		uint16_t width, uint16_t height, bool override_redirect) {
@@ -178,6 +181,8 @@ static struct wlr_xwayland_surface *xwayland_surface_create(
 		wlr_log(WLR_ERROR, "Could not add timer to event loop");
 		return NULL;
 	}
+
+	xwm_send_focus_window(xwm, xwm->focus_surface);
 
 	wlr_signal_emit_safe(&xwm->xwayland->events.new_surface, surface);
 


emersion:
Maybe we could remove `XCB_EVENT_MASK_FOCUS_CHANGE` instead?

RedSoxFan:
> Maybe we could remove `XCB_EVENT_MASK_FOCUS_CHANGE` instead?

Doesn't appear to have any effect

emersion:
>Doesn't appear to have any effect

Oh right, that's normal, this just asks the X11 server to notify us when the window gains focus.

I'd like to investigate why the window gains focus in the first place, to avoid back-and-forth focus changes.

psychon:
Per https://github.com/swaywm/sway/issues/2899, I understand that the problem here is actually about stacking order and not the input focus.

Per https://www.x.org/releases/X11R7.5/doc/x11proto/proto.html, docs for CreateWindow:

> The window is placed on top in the stacking order with respect to siblings.

Since XWM seems to only interact with the stacking order when a window is focused, I suspect that "some kind of interaction" with a non-focused window could have similar problems, but I cannot think of any concrete problems right now...

Anyway, for the issue at hand, I would propose the following change to wlroots to ensure that new windows end up at the bottom of the stacking order, but will not send a PR since I cannot actually test this. Thus, feel free to take this, format it correctly, test it, and submit a PR.
```diff
diff --git a/xwayland/xwm.c b/xwayland/xwm.c
index c1e95903..d4f1ce55 100644
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -915,6 +915,7 @@ static void xwm_handle_map_request(struct wlr_xwm *xwm,
 
        xsurface_set_wm_state(xsurface, ICCCM_NORMAL_STATE);
        xsurface_set_net_wm_state(xsurface);
+       xcb_configure_window(xwm->xcb_conn, ev->window, XCB_CONFIG_WINDOW_STACK_MODE, (uint32_t[]) { XCB_STACK_MODE_BELOW });
        xcb_map_window(xwm->xcb_conn, ev->window);
 }
 
```

RedSoxFan:
Superseded by #1541

PR #1538, opened 2019-02-11
[RFC] Add backend capability bitmask

ascent12:
This allows for a more robust an extensible way to advertise features that a backend is capable of, instead of just trying a function and seeing if it fails.
I've just added the enum here; I thought I would get some feedback before actually bothering to plumb this up.

This is something I want for the renderer PR and potential future features we'd want to add. It may be worth adding a separate bitmask for wlr_outputs too.

### DIFF
--- a/include/wlr/backend.h
+++ b/include/wlr/backend.h
@@ -15,8 +15,45 @@
 
 struct wlr_backend_impl;
 
+enum wlr_backend_capability {
+	/*
+	 * This backend supports input devices.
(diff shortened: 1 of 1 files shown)


ddevault:
+1 but I don't want to merge it until it's used somewhere

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1538

PR #1541, opened 2019-02-13
xwm: stack below on map

RedSoxFan:
Fixes swaywm/sway#2899
Supersedes #1534 

Since xwm only manipulates the stack when focusing a window, newly
mapped windows should be stacked below the focused window. This prevents
the newly mapped window from stealing focus due to being on the top of
the stack.

Thanks psychon for the insight! See https://github.com/swaywm/wlroots/pull/1534#issuecomment-463191842 for the full explanation

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -889,6 +889,12 @@ static void xwm_handle_map_request(struct wlr_xwm *xwm,
 
 	xsurface_set_wm_state(xsurface, ICCCM_NORMAL_STATE);
 	xsurface_set_net_wm_state(xsurface);
+
+	uint32_t values[1];
+	values[0] = XCB_STACK_MODE_BELOW;
+	xcb_configure_window(xwm->xcb_conn, ev->window,
+			XCB_CONFIG_WINDOW_STACK_MODE, values);
+
 	xcb_map_window(xwm->xcb_conn, ev->window);
 }
 


psychon:
> e.g. tooltips will stay at the bottom in X11's model but will be rendered on top in ours.

Shouldn't tooltips be override-redirect and thus not go through the MapRequest handling? I specifically did this change in response to a MapRequest and not a CreateNotify, because that way override-redirect windows are not touched.

emersion:
>Shouldn't tooltips be override-redirect and thus not go through the MapRequest handling?

Do tooltips send a MapRequest?

We're not just a WM, we also render surfaces, so we need to handle overide-redirect windows as well.

psychon:
> > Shouldn't tooltips be override-redirect and thus not go through the MapRequest handling?
>
> Do tooltips send a MapRequest?

Nope, I don't think so. The "override" in "override-redirect" refers to `XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT` (and `XCB_EVENT_MASK_RESIZE_REDIRECT`, but I never saw anything actually use that). This event bit is what causes map requests from clients to be ignored by the X11 server and instead a `XCB_MAP_REQUEST` event is sent to whichever client selected that event mask bit. Anything causing a map request should be a "normal" top level window (minus `_NET_WM_WINDOW_TYPE`, but Most (tm) WMs will e.g. put window decorations on non-override-redirect windows, so just specifying `_NET_WM_WINDOW_TYPE_TOOLTIP` is not enough to have a tooltip work).

emersion:
Okay. This should be a good solution then :D

emersion:
@RedSoxFan, can you force push? There was a short builds.sr.ht outage when you opened the PR.

RedSoxFan:
> @RedSoxFan, can you force push? There was a short builds.sr.ht outage when you opened the PR.

Done

emersion:
Thanks!

PR #1545, opened 2019-02-14
drm: do not modeset to current mode

RedSoxFan:
Fixes swaywm/sway#3659
Supersedes swaywm/sway#3161

There is no point in modesetting an output to a mode that it is already
set to. Modesetting will cause the output to briefly flicker which is
undesirable for a noop. This returns early in `drm_connector_set_mode`
when attempting to modeset to the current mode.


emersion:
Thanks!

PR #1547, opened 2019-02-14
Post protocol error on invalid wl_surface scale

dos1:
Letting the scale be set to 0 causes division by zero errors.


emersion:
Can you add the same to `surface_set_scale`?

emersion:
Thanks!

PR #1550, opened 2019-02-15
Revert "drm: do not modeset to current mode"

ddevault:
Reverts swaywm/wlroots#1545


emersion:
Yeah, would be better to keep #1545.

Issue #1536, opened 2019-02-11
Rootston crashes when used with tablet tool


wsinnema:
Rootston crashes when I do the following:

- Open termite
- Point a stylus on termite
- Put the stylus out of range for the screen
- Rootston crashes because tool->tool is NULL in handle_tablet_tool_v2_set_cursor

[Core dump for commit 9e49ceb12985697fbfc0b2fa8f86143b29cc837b ](https://github.com/swaywm/wlroots/files/2851818/rootston.core.gz)

emersion:
Please provide a stack trace

wsinnema:
#0  0x00007f1200423d62 in wl_signal_emit (signal=0x120, data=0x7ffdd8410a50) at /usr/include/wayland-server-core.h:467
#1  0x00007f1200424213 in handle_tablet_tool_v2_set_cursor (client=0x556f785e5000, resource=0x556f78554eb0, serial=8, surface_resource=0x556f785c2c90, hotspot_x=4, hotspot_y=4) at ../types/tablet_v2/wlr_tablet_v2_tool.c:47
#2  0x00007f11ffa356d0 in ffi_call_unix64 () at /lib64/libffi.so.6
#3  0x00007f11ffa350a0 in ffi_call () at /lib64/libffi.so.6
#4  0x00007f12003c46ff in  () at /lib64/libwayland-server.so.0
#5  0x00007f12003c10a3 in  () at /lib64/libwayland-server.so.0
#6  0x00007f12003c2702 in wl_event_loop_dispatch () at /lib64/libwayland-server.so.0
#7  0x00007f12003c12ac in wl_display_run () at /lib64/libwayland-server.so.0
#8  0x0000556f763e4686 in main (argc=1, argv=0x7ffdd8411128) at ../rootston/main.c:74

wsinnema:
It only happens on GTK applications, everything works fine on Konsole and Rootston (as client), but not with Termite and Epiphany

Ongy:
Nothing immediatly obvious.
Can't reproduce with gedit, and debug log shows the `set_cursor` calls.

Can you check valgrind and provide the client WAYLAND_DEBUG log?

wsinnema:
[valgrind.txt](https://github.com/swaywm/wlroots/files/2868657/valgrind.txt)
[client.txt](https://github.com/swaywm/wlroots/files/2868659/client.txt)

PR #1551, opened 2019-02-15
Prevent NULL dereference in tablet_tool handler

Ongy:
In case a tool was removed, but not yet destroyed by the client, the
tool_client's tool can be NULL. We have to check that as well in the
set_cursor handler to prevent using inert resources

Should fix https://github.com/swaywm/wlroots/issues/1536#issuecomment-463971534

This probably only ever happens with tablets that don't track unique tools (or maybe on unplug?), so I can't test this well.
@wsinnema please test this for me, I can't reproduce it with my hardware.


wsinnema:
It works for me

emersion:
Thanks!

Issue #1543, opened 2019-02-13
Proposal: helper to track surface damage


emersion:
```c
struct wlr_surface_damage_manager {
	struct wlr_surface *surface;
	struct {
		struct wl_signal damage;
	} events;
};

struct wlr_surface_damage_manager_damage_event {
	struct wlr_surface_damage_manager *manager;
	int32_t sx, sy;
	pixman_region32_t *damage;
};

// Listens to surface commit, map/unmap and damage from children subsurfaces.
struct wlr_surface_damage_manager *wlr_surface_damage_manager_from_subsurface(
	struct wlr_subsurface *subsurface);
// Listens to surface commit, map/unmap and damage from children
// subsurfaces/popups.
struct wlr_surface_damage_manager *wlr_surface_damage_manager_from_xdg_surface(
	struct wlr_xdg_surface *xdg_surface);
```

PR #1552, opened 2019-02-16
[WIP] surface-damage: add a helper to track surface damage

emersion:
One remaining major pain point when implementing damage tracking is tracking
damage from surfaces and all of their children. Compositors need to add
new listeners for each view child type, with the appropriate new data
structures.

This commit adds a new helper to ease damage tracking: wlr_surface_damage. It
tracks damage from a surface and all its children and reports it via a "damage"
event in surface-local coordinates. This way, compositors just need to add one
listener for the whole view. There is one function per shell to create a
wlr_surface_damage, just like other helpers like wlr_surface_for_each_surface.

Additional functions are provided to allow compositors to manually damage
surfaces. wlr_surface_damage_add_whole will damage the whole surface and all
its children, this is useful when moving a view.

Of course, this helper is optional, and damage tracking done entirely in the
compositor will still work.

- [x] wl-subsurface
- [x] xdg-shell
- [ ] xdg-shell-unstable-v6
- [ ] layer-shell

Fixes https://github.com/swaywm/wlroots/issues/1543

### DIFF
--- a/include/rootston/output.h
+++ b/include/rootston/output.h
@@ -40,6 +40,8 @@ struct roots_drag_icon;
 void output_damage_whole(struct roots_output *output);
 void output_damage_whole_view(struct roots_output *output,
 	struct roots_view *view);
+void output_damage(struct roots_output *output,
+	int ox, int oy, pixman_region32_t *damage);
 void output_damage_from_view(struct roots_output *output,
 	struct roots_view *view);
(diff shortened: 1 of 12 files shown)


emersion:
Closing in favor of https://github.com/swaywm/wlroots/pull/3117

Issue #1414, opened 2018-12-04
Consider adding map/unmap events to wlr_subsurface


emersion:
The protocol says a wl_subsurface becomes mapped when there's a non-NULL buffer commit and all the parents are mapped. Adding a signal for this would simplify the parent logic.

PR #1553, opened 2019-02-16
subsurface: add map/unmap events

emersion:
Fixes https://github.com/swaywm/wlroots/issues/1414

### DIFF
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -181,6 +181,8 @@ struct roots_subsurface {
 	struct roots_view_child view_child;
 	struct wlr_subsurface *wlr_subsurface;
 	struct wl_listener destroy;
+	struct wl_listener map;
+	struct wl_listener unmap;
 };
 
 struct roots_wl_shell_popup {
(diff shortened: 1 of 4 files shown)


ddevault:
Should we avoid iterating over unmapped subsurfaces when rendering? Should we instruct compositors to check whether or not a subsurface is mapped when rendering?

emersion:
Err, forgot that part. Hmm. It'd be better to handle all of this mess in wlroots. I'll work on a fix.

emersion:
Fixed

ddevault:
I don't follow how this new patch addresses my concern.

emersion:
Oh, yeah, the fix was more about the map/unmap semantics for subsurfaces (described in the original issue).

About how compositors should handle it… Right now for popups we iterate over unmapped ones too. I'm not sure iterating over only mapped surfaces is a good idea, as surface iterators can be used for e.g. `wlr_surface_send_{enter,leave}`. I'd say the compositor should handle it, although Rootston doesn't do it now (not for subsurfaces, not for popups, not for toplevels, not for xwayland surfaces).

ddevault:
Thanks!

PR #1554, opened 2019-02-16
wlr_output: do not modeset to current mode

RedSoxFan:
Supersedes #1545 which was reverted in #1550 due to
swaywm/sway#3685, which is a sway issue and has
swaywm/sway#3704 open to fix it. 

Fixes swaywm/sway#3659

There is no point in modesetting an output to a mode that it is already
set to. Modesetting will cause the output to briefly flicker which is
undesirable for a noop. This prevents modesetting any wlr_output,
regardless of the backend, to it's currently set mode.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -159,6 +159,9 @@ bool wlr_output_set_mode(struct wlr_output *output,
 	if (!output->impl || !output->impl->set_mode) {
 		return false;
 	}
+	if (output->current_mode == mode) {
+		return true;
+	}
 	return output->impl->set_mode(output, mode);
 }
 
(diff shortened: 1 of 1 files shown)


emersion:
Could we add a similar check for `wlr_output_set_custom_mode`?

RedSoxFan:
> Could we add a similar check for `wlr_output_set_custom_mode`?

Done

types/wlr_output.c
```diff
@@ -167,6 +170,11 @@ bool wlr_output_set_custom_mode(struct wlr_output *output, int32_t width,
 	if (!output->impl || !output->impl->set_custom_mode) {
 		return false;
 	}
+	if (output->current_mode->width == width &&
```
emersion: Some outputs may not have modes, in which case `output->current_mode` is NULL. We need to use `output->{width,height,refresh}` instead.

RedSoxFan: Fixed

emersion:
Thanks!

PR #1555, opened 2019-02-17
[tinywl] Check that wlr_surface is an xdg surface

Timidger:
Fixes a segfault that occurs when you click on a client.

### DIFF
--- a/tinywl/tinywl.c
+++ b/tinywl/tinywl.c
@@ -102,7 +102,7 @@ static void focus_view(struct tinywl_view *view, struct wlr_surface *surface) {
 		/* Don't re-focus an already focused surface. */
 		return;
 	}
-	if (prev_surface) {
+	if (prev_surface && wlr_surface_is_xdg_surface(seat->keyboard_state.focused_surface)) {
 		/*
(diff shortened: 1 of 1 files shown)


emersion:
What kind of surface is it, if it's not an xdg-surface?

Timidger:
Dunno, the abort didn't say and I just read the backtrace. Potentially this is a different issue.

emersion:
Is there an easy way to reproduce it?

You can print a surface role: `surface->role->name`.

Timidger:
This is very easy to reproduce. Just run it with any application (I found it worked regardless of the ones I tried, but gnome-calculator and alacritty definitely did it) and try to click on it (I ran it under Wayland backend).

In the core dump that memory can't be accessed, so maybe an uninitialized pointer?

emersion:
Cannot reproduce. Compiled with `-fsanitize=address`.

>so maybe an uninitialized pointer?

Seems unlikely, pointer focus is a very important pointer, we would've noticed it in sway/rootston.

Timidger:
I'll have time to investigate this in a few days.

Timidger:
Can't reproduce any more, might be an env issue.

PR #1556, opened 2019-02-18
xwayland, data-device: fix surface state on unmap

emersion:
This commit makes sure surface->mapped is true when the unmapped event is
emitted. This is necessary because listeners can only damage surfaces that are
mapped. This is similar to the fact that the destroy event is emitted before
any destruction is actually made.

Fixes https://github.com/swaywm/sway/issues/3568


ddevault:
Thanks!

PR #1557, opened 2019-02-18
data-device: make old offers inert before sending selection

emersion:
Fixes https://github.com/swaywm/wlroots/issues/1383


ddevault:
Thanks!

PR #1558, opened 2019-02-18
backend/drm: fix NULL dereference when unsetting cursor

emersion:
This segfault happens on multi-GPU systems.

Fixes https://github.com/swaywm/sway/issues/3717

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -720,9 +720,8 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 	}
 
 	struct gbm_bo *bo = plane->cursor_enabled ? plane->surf.back : NULL;
-
-	if (drm->parent) {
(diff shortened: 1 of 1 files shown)


backend/drm/drm.c
```diff
@@ -720,8 +720,7 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 	}
 
 	struct gbm_bo *bo = plane->cursor_enabled ? plane->surf.back : NULL;
-
-	if (drm->parent) {
+	if (bo && drm->parent) {
 		bo = copy_drm_surface_mgpu(&plane->mgpu_surf, plane->surf.back);
```
VincentVanlaer: We should also replace `plane->surf.back` by `bo` here, just to prevent future mistakes in case `plane->surf.back`is replaced by something else.

emersion: Oh yeah, that's a great idea

ddevault:
Thanks!

PR #1561, opened 2019-02-19
backend/drm: fix modeset on drm fd resume

### LINKED ISSUE #1560: can't change ttyt
sway or rootston
from commit 2d2c79e can not be changed to another tty.

RedSoxFan:
Fixes #1560 

On DRM resume, such as switching back to a TTY, the output needs to be
modeset to the current mode. However, wlr_output_set_mode will return
early when attempting to set the mode to the current mode. This just
steps around wlr_output_set_mode and calls drm_connector_set_mode
directly.

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -93,7 +93,8 @@ static void session_signal(struct wl_listener *listener, void *data) {
 		struct wlr_drm_connector *conn;
 		wl_list_for_each(conn, &drm->outputs, link){
 			if (conn->output.enabled) {
-				wlr_output_set_mode(&conn->output, conn->output.current_mode);
+				drm_connector_set_mode(&conn->output,
+						conn->output.current_mode);
 			} else {
 				enable_drm_connector(&conn->output, false);
 			}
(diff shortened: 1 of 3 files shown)


backend/drm/backend.c
```diff
@@ -93,7 +93,13 @@ static void session_signal(struct wl_listener *listener, void *data) {
...
 		wl_list_for_each(conn, &drm->outputs, link){
 			if (conn->output.enabled) {
-				wlr_output_set_mode(&conn->output, conn->output.current_mode);
+				// wlr_output_set_mode prevents setting the mode to the
+				// current mode since it is almost always a noop. This however
+				// is one exception so just set current_mode to NULL before
+				// setting the mode.
+				struct wlr_output_mode *mode = conn->output.current_mode;
+				conn->output.current_mode = NULL;
+				wlr_output_set_mode(&conn->output, mode);
```
emersion: Can't we just call our DRM-private function in `drm/drm.c`?

RedSoxFan: Yup. Just had to change the scope of it to so it was no longer static.

emersion:
Thanks!

PR #1562, opened 2019-02-19
xwayland: don't use SOCK_CLOEXEC

emersion:
SOCK_CLOEXEC isn't POSIX.

### DIFF
--- a/xwayland/xwayland.c
+++ b/xwayland/xwayland.c
@@ -1,8 +1,4 @@
 #define _POSIX_C_SOURCE 200112L
-#ifdef __FreeBSD__
-// for SOCK_CLOEXEC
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

jbeich:
Was the other place in xwayland left as is intentionally?

https://github.com/swaywm/wlroots/blob/50011e7170d1d6a58a6b1e4997da6ce105180481/xwayland/sockets.c#L2-L5
https://github.com/swaywm/wlroots/blob/50011e7170d1d6a58a6b1e4997da6ce105180481/xwayland/sockets.c#L32

emersion:
Good catch. Addressed in https://github.com/swaywm/wlroots/pull/1564

PR #1563, opened 2019-02-19
rootston: don't try to maximize fullscreen surfaces

dos1:
It doesn't make much sense and actually breaks stuff when using layer-shell
(fullscreen window gets resized, but it's still fullscreen, leading to black
bars where the shell layers are behind).

### DIFF
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -213,6 +213,10 @@ static struct wlr_output *view_get_output(struct roots_view *view) {
 }
 
 void view_arrange_maximized(struct roots_view *view) {
+	if (view->fullscreen_output != NULL) {
+		return;
+	}
+
 	struct wlr_box view_box;
 	view_get_box(view, &view_box);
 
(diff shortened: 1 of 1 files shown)


dos1:
Turns out there's another similar case in layer_shell - updated the patch to handle that as well.

Decided against moving the check to `view_arrange_maximized`, as it seems like `view_maximize` would still have to have that check anyway, but if you prefer it there instead of the one in `arrange_layers` I can move it.

emersion:
>Decided against moving the check to `view_arrange_maximized`, as it seems like `view_maximize` would still have to have that check anyway, but if you prefer it there instead of the one in `arrange_layers` I can move it.

Hmm. Yeah I'd prefer to have it in `view_arrange_maximized`, as there might be other `view_arrange_maximized` callers (now or in the future).

dos1:
Updated again :)

emersion:
Thanks!

PR #1564, opened 2019-02-20
xwayland: remove remaining SOCK_CLOEXEC

emersion:


ddevault:
FreeBSD build failed here

emersion:
Okay, we were using non-POSIX address families. Fixed.

ddevault:
Thanks!

PR #1565, opened 2019-02-21
backend/drm: fix memory leak in realloc crtcs

Emantor:
If *changed_outputs is not supplied by the calling function, track the local
allocation with a bool variable and free the allocation at the end of the
function.


ascent12:
Thanks

PR #1566, opened 2019-02-22
rootston: surface_at: check for fullscreen surfaces in between TOP and OVERLAY layers

dos1:
Input order didn't match rendering order, causing pointer events to go into surfaces
that were completely obscured by others (like a LAYER_TOP panel behind a fullscreen
window).

### DIFF
--- a/rootston/desktop.c
+++ b/rootston/desktop.c
@@ -755,20 +755,6 @@ static bool view_at(struct roots_view *view, double lx, double ly,
 static struct roots_view *desktop_view_at(struct roots_desktop *desktop,
 		double lx, double ly, struct wlr_surface **surface,
 		double *sx, double *sy) {
-	struct wlr_output *wlr_output =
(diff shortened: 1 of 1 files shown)


rootston/desktop.c
```diff
@@ -817,6 +803,15 @@ struct wlr_surface *desktop_surface_at(struct roots_desktop *desktop,
 					ox, oy, sx, sy))) {
 			return surface;
 		}
+
+		struct roots_output *output =
+			desktop_output_from_wlr_output(desktop, wlr_output);
+		if (output != NULL && output->fullscreen_view != NULL) {
+			if (view_at(output->fullscreen_view, lx, ly, &surface, sx, sy)) {
+				return surface;
+			}
```
emersion: If there's a fullscreen view, we don't want to check for views/layers underneath.

dos1: Ah, makes sense. Updated!

emersion:
Thanks!

PR #1571, opened 2019-02-24
backend/drm: Unset cursor on cleanup

ascent12:
Fixes https://github.com/swaywm/sway/issues/3017

This will prevent the cursor from persisting on the Linux framebuffer
terminal on exit.


emersion:
Thanks!

PR #1573, opened 2019-02-24
Install xcb-errors on FreeBSD builder

jbeich:
Other builders don't test optional dependencies. Let's see if this works after d1409e338175.


emersion:
Thanks!

Issue #1570, opened 2019-02-24
Drag and drop crash when dropping on a target that does not support DND


aidanharris:
The title is my best guess at what's going on. In sway when dropping onto a window that does not support DND (tested by dragging a folder from Dolphin into Alacritty) I get a crash. Patching wlroots as follows seems to fix things:

```diff

diff --git a/types/data_device/wlr_data_offer.c b/types/data_device/wlr_data_offer.c

index 086feb11..2a1cd6f2 100644

--- a/types/data_device/wlr_data_offer.c

+++ b/types/data_device/wlr_data_offer.c

@@ -98,7 +98,7 @@ static void data_offer_handle_receive(struct wl_client *client,

 

 static void data_offer_dnd_finish(struct wlr_data_offer *offer) {

 	struct wlr_data_source *source = offer->source;

-	if (source->actions < 0) {

+	if (source && source->actions < 0) {

 		return;

 	}

 

diff --git a/types/data_device/wlr_data_source.c b/types/data_device/wlr_data_source.c

index 0ab0d7f7..d2a72898 100644

--- a/types/data_device/wlr_data_source.c

+++ b/types/data_device/wlr_data_source.c

@@ -61,6 +61,8 @@ void wlr_data_source_dnd_drop(struct wlr_data_source *source) {

 }

 

 void wlr_data_source_dnd_finish(struct wlr_data_source *source) {

+	if (source == NULL)

+		return;

 	if (source->impl->dnd_finish) {

 		source->impl->dnd_finish(source);

 	}

```

```c

[New LWP 21890]

[New LWP 21954]

[Thread debugging using libthread_db enabled]

Using host libthread_db library "/lib64/libthread_db.so.1".

Core was generated by `sway'.

Program terminated with signal SIGSEGV, Segmentation fault.

#0  data_offer_dnd_finish (offer=offer@entry=0x555919284e60) at ../wlroots-9999/types/data_device/wlr_data_offer.c:101

101		if (source->actions < 0) {

[Current thread is 1 (Thread 0x7ff8915da9c0 (LWP 21890))]



Thread 2 (Thread 0x7ff8906d6700 (LWP 21954)):

#0  futex_wait_cancelable (private=0, expected=0, futex_word=0x555918ca4d28) at ../sysdeps/unix/sysv/linux/futex-internal.h:88

        __ret = -512

        oldtype = 0

        err = <optimized out>

        oldtype = <optimized out>

        err = <optimized out>

[...]

PR #1574, opened 2019-02-24
data-device: check if there's a source before sending dnd_finish

emersion:
This could cause a segfault in data_offer_destroy if the offer has version < 3.

Fixes https://github.com/swaywm/wlroots/issues/1570


ddevault:
Thanks!

Issue #1114, opened 2018-07-05
Allow headless backend to re-use an existing renderer


emersion:
This would allow things like having a DRM backend with some headless outputs.

emersion:
Implemented in 6129a6f93e97680dfe7e983db2ee7671edb5ffbe.

PR #1578, opened 2019-02-25
RDP (remote desktop) backend

ddevault:
- [x] Spin up FreeRDP server
- [x] Allocate headless outputs for each peer
- [x] Send frame updates to clients via RFX
- [x] Send frame updates to clients via NSC
- [x] Send raw frame updates for clients which support neither
- [x] Keyboard input
- [x] Pointer input
- [x] Clean up

### DIFF
--- a/backend/meson.build
+++ b/backend/meson.build
@@ -53,6 +53,22 @@ if logind.found()
 	backend_deps += logind
 endif
 
+if freerdp.found() and winpr2.found()
+	backend_files += files(
+		'rdp/backend.c',
+		'rdp/keyboard.c',
+		'rdp/listener.c',
+		'rdp/output.c',
+		'rdp/peer.c',
+		'rdp/pointer.c',
+	)
+	backend_deps += [
+		freerdp,
+		winpr2
+	]
+	conf_data.set10('WLR_HAS_RDP_BACKEND', true)
+endif
+
 subdir('x11')
 
 lib_wlr_backend = static_library(
--- a/backend/rdp/output.c
+++ b/backend/rdp/output.c
@@ -1,0 +1,288 @@
+#include <assert.h>
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#include <stdlib.h>
+#include <wayland-server.h>
+#include <wlr/types/wlr_output.h>
+#include <wlr/interfaces/wlr_output.h>
+#include <wlr/util/log.h>
+#include <wlr/render/wlr_renderer.h>
+#include "backend/rdp.h"
+#include "util/signal.h"
+
+static struct wlr_rdp_output *rdp_output_from_output(
+		struct wlr_output *wlr_output) {
+	assert(wlr_output_is_rdp(wlr_output));
+	return (struct wlr_rdp_output *)wlr_output;
+}
+
+static EGLSurface egl_create_surface(struct wlr_egl *egl, unsigned int width,
+		unsigned int height) {
+	EGLint attribs[] = {
+		EGL_WIDTH, width,
+		EGL_HEIGHT, height,
+		EGL_NONE,
+	};
+	EGLSurface surf = eglCreatePbufferSurface(egl->display, egl->config, attribs);
+	if (surf == EGL_NO_SURFACE) {
+		wlr_log(WLR_ERROR, "Failed to create EGL surface");
+		return EGL_NO_SURFACE;
+	}
+	return surf;
+}
+
+static bool output_set_custom_mode(struct wlr_output *wlr_output, int32_t width,
+		int32_t height, int32_t refresh) {
+	struct wlr_rdp_output *output =
+		rdp_output_from_output(wlr_output);
+	struct wlr_rdp_backend *backend = output->backend;
+
+	if (refresh <= 0) {
+		refresh = 60 * 1000; // 60 Hz
+	}
+
+	wlr_egl_destroy_surface(&backend->egl, output->egl_surface);
+
+	output->egl_surface = egl_create_surface(&backend->egl, width, height);
+	if (output->egl_surface == EGL_NO_SURFACE) {
+		wlr_log(WLR_ERROR, "Failed to recreate EGL surface");
+		wlr_output_destroy(wlr_output);
+		return false;
+	}
+
+	output->frame_delay = 1000000 / refresh;
+
+	if (output->shadow_surface) {
+		pixman_image_unref(output->shadow_surface);
+	}
+	output->shadow_surface = pixman_image_create_bits(PIXMAN_x8r8g8b8,
+			width, height, NULL, width * 4);
+
+	wlr_output_update_custom_mode(&output->wlr_output, width, height, refresh);
+	return true;
+}
+
+static void output_transform(struct wlr_output *wlr_output,
+		enum wl_output_transform transform) {
+	struct wlr_rdp_output *output =
+		rdp_output_from_output(wlr_output);
+	output->wlr_output.transform = transform;
+}
+
+static bool output_make_current(struct wlr_output *wlr_output, int *buffer_age) {
+	struct wlr_rdp_output *output =
+		rdp_output_from_output(wlr_output);
+	return wlr_egl_make_current(&output->backend->egl, output->egl_surface,
+		buffer_age);
+}
+
+static bool rfx_swap_buffers(
+		struct wlr_rdp_output *output, pixman_region32_t *damage) {
+	struct wlr_rdp_peer_context *context = output->context;
+	freerdp_peer *peer = context->peer;
+	rdpUpdate *update = peer->update;
+
+	Stream_Clear(context->encode_stream);
+	Stream_SetPosition(context->encode_stream, 0);
+	int width = damage->extents.x2 - damage->extents.x1;
+	int height = damage->extents.y2 - damage->extents.y1;
+
+	SURFACE_BITS_COMMAND cmd;
+	cmd.skipCompression = TRUE;
+	cmd.destLeft = damage->extents.x1;
+	cmd.destTop = damage->extents.y1;
+	cmd.destRight = damage->extents.x2;
+	cmd.destBottom = damage->extents.y2;
+	cmd.bmp.bpp = pixman_image_get_depth(output->shadow_surface);
+	cmd.bmp.codecID = peer->settings->RemoteFxCodecId;
+	cmd.bmp.width = width;
+	cmd.bmp.height = height;
+
+	uint32_t *ptr = pixman_image_get_data(output->shadow_surface) +
+		damage->extents.x1 + damage->extents.y1 *
+		(pixman_image_get_stride(output->shadow_surface) / sizeof(uint32_t));
+
+	RFX_RECT *rfx_rect;
+	int nrects;
+	pixman_box32_t *rects =
+		pixman_region32_rectangles(damage, &nrects);
+	rfx_rect = realloc(context->rfx_rects, nrects * sizeof(*rfx_rect));
+	if (rfx_rect == NULL) {
+		wlr_log(WLR_ERROR, "RDP swap buffers failed: could not realloc rects");
+		return false;
+	}
+	context->rfx_rects = rfx_rect;
+
+	for (int i = 0; i < nrects; ++i) {
+		pixman_box32_t *region = &rects[i];
+		rfx_rect = &context->rfx_rects[i];
+		rfx_rect->x = region->x1 - damage->extents.x1;
+		rfx_rect->y = region->y1 - damage->extents.y1;
+		rfx_rect->width = region->x2 - region->x1;
+		rfx_rect->height = region->y2 - region->y1;
+	}
+
+	rfx_compose_message(context->rfx_context, context->encode_stream,
+			context->rfx_rects, nrects, (BYTE *)ptr, width, height,
+			pixman_image_get_stride(output->shadow_surface));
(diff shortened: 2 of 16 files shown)


emersion:
Would be nice to let the compositor allocate the output, so that it can decide to either create a headless one or use an existing one. That would also remove all the OpenGL stuff from the backend.

ddevault:
Hmm, maybe. The way I saw this working out is that the compositor would mirror to the RDP outputs if they wanted to use an existing one. I also like that this lets you transparently add RDP support without any changes to the compositors.

emersion:
Still unsure whether this should be merged or not. A client with virtual-keyboard + virtual-pointer + virtual-output would probably be better.

ddevault:
I'm starting to wonder if it wouldn't be nice to have both.

emersion:
In any case, you could use a headless backend inside the RDP backend to avoid having to re-implement `wlr_output`.

I still think the compositor should be able to customize the output if needed, this could be optional. This can be useful for mirroring or remote controlling.

ddevault:
That seems like a nice to have for v2 of this patch.

emersion:
Note: this backend is better than Weston's because it uses OpenGL (the Weston backend forces the software renderer).

ddevault:
Or it would be if this didn't force llvmpipe for some reason <_< headless backend has the same problem. No luck figuring it out yet.

progandy:
@ddevault Maybe open a render node fd (```/dev/dri/render...```) and create a normal GBM backed mesa platform just without any crtcs or outputs? My very shallow reading of surfaceless mesa seems to indicate that it is built on top of llvmpipe.

ddevault:
Maybe? I'm not sure how that works to be honest. Interested in trying a patch?

emersion:
>Maybe open a render node fd (`/dev/dri/render...`) and create a normal GBM backed mesa platform just without any crtcs or outputs?

That's how the not-yet-merged Weston patch works: https://lists.freedesktop.org/archives/wayland-devel/2017-July/034423.html

emersion:
I'd first try to read the mesa code though.

progandy:
> Maybe? I'm not sure how that works to be honest. Interested in trying a patch?

Currently not really. I don't really know how that works either. I just remembered reading about render nodes for unprivileged gpu access without a DRM master for computation and DRI_PRIME.

I found this implementation in virgl: https://gitlab.collabora.com/virgl-es/virglrenderer/blob/master/src/virgl_egl_context.c

#1114 is probably related. A single DRM backend that supports real and virtual outputs if it is run as DRM master. If it is started on a render node, it will only have virtual outputs and be completely headless.

emersion:
Other idea: use `EGL_PLATFORM_DEVICE_EXT`. I might try this this weekend.

ascent12:
>use EGL_PLATFORM_DEVICE_EXT

Mesa does not support this.

emersion:
On my machine `EGL_PLATFORM_SURFACELESS_MESA` correctly selects i915 and doesn't fallback to software rendering. Maybe this is a Mesa bug?

Can you try with `export EGL_LOG_LEVEL=debug`, just to make sure we're not missing anything?

progandy:
I walked through the code and it look like surfaceless does try to create a render node: https://gitlab.freedesktop.org/mesa/mesa/blob/master/src/egl/drivers/dri2/platform_surfaceless.c#L358

* first try: render node and hardware rendering
* second try: DRM master with software rendering
* last try: swrast without DRM

I did a log as well, looks like r600 is loaded.

    EGL_LOG_LEVEL=debug WLR_RDP_TLS_CERT_PATH="$PWD/cert/cert.pem" WLR_RDP_TLS_KEY_PATH="$PWD/cert/key.pem" WLR_BACKENDS=rdp ./rootston 2>rootston-rdp.log

[rootston-rdp.log](https://github.com/swaywm/wlroots/files/2921365/rootston-rdp.log)

ddevault:
This is ready for review. It still ends up using llvmpipe on my system, but that can be solved separately.

backend/rdp/peer.c
```diff
@@ -0,0 +1,352 @@
...
+		(struct wlr_rdp_peer_context *)client->context;
+	struct wlr_rdp_backend *backend = context->backend;
+	rdpSettings *settings = client->settings;
+
+	if (!settings->SurfaceCommandsEnabled) {
+		wlr_log(WLR_ERROR, "RDP peer does not support SurfaceCommands");
+		return FALSE;
+	}
+
+	context->output = wlr_rdp_output_create(backend, context,
```
emersion: This is missing a NULL check

backend/rdp/peer.c
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	context->pointer = wlr_rdp_pointer_create(backend);
+	// Use wlroots' software cursors instead of remote cursors
+	POINTER_SYSTEM_UPDATE pointer_system;
+	rdpPointerUpdate *pointer = client->update->pointer;
+	pointer_system.type = SYSPTR_NULL;
+	pointer->PointerSystem(client->context, &pointer_system);
+
+	context->keyboard = wlr_rdp_keyboard_create(backend, settings);
```
emersion: Ditto

backend/rdp/peer.c
```diff
@@ -0,0 +1,352 @@
...
+			context->output->wlr_output.height);
+	nsc_context_reset(context->nsc_context,
+			context->output->wlr_output.width,
+			context->output->wlr_output.height);
+
+	if (context->flags & RDP_PEER_ACTIVATED) {
+		return TRUE;
+	}
+
+	context->pointer = wlr_rdp_pointer_create(backend);
```
emersion: Ditto

backend/rdp/peer.c
```diff
@@ -0,0 +1,352 @@
...
+}
+
+static int xf_suppress_output(rdpContext *context,
+		BYTE allow, const RECTANGLE_16 *area) {
+	struct wlr_rdp_peer_context *peer_context =
+		(struct wlr_rdp_peer_context *)context;
+	if (allow) {
+		peer_context->flags |= RDP_PEER_OUTPUT_ENABLED;
+	} else {
+		peer_context->flags &= (~RDP_PEER_OUTPUT_ENABLED);
```
emersion: Style: extra parentheses here

backend/rdp/peer.c
```diff
@@ -0,0 +1,352 @@
...
+	struct wlr_input_device *wlr_device = &context->pointer->wlr_input_device;
+	struct wlr_pointer *pointer = wlr_device->pointer;
+	bool frame = false;
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+
+	if (flags & PTR_FLAGS_MOVE) {
+		struct wlr_event_pointer_motion_absolute event = { 0 };
+		int width, height;
+		wlr_output_effective_resolution(
```
emersion: `wlr_output_effective_resolution` shouldn't be used here. Instead, `wlr_output->{width,height}` should be used and the input device should be mapped to the output.

(Applies to all absolute motion events)

backend/rdp/pointer.c
```diff
@@ -0,0 +1,33 @@
...
+#include <wlr/interfaces/wlr_input_device.h>
+#include <wlr/util/log.h>
+#include "backend/rdp.h"
+#include "util/signal.h"
+
+static struct wlr_input_device_impl input_device_impl = { 0 };
+
+struct wlr_rdp_input_device *wlr_rdp_pointer_create(
+		struct wlr_rdp_backend *backend) {
+	struct wlr_rdp_input_device *device =
```
emersion: This is missing a NULL check

backend/rdp/pointer.c
```diff
@@ -0,0 +1,33 @@
...
+	const char *name = "rdp";
+	struct wlr_input_device *wlr_device = &device->wlr_input_device;
+	wlr_input_device_init(wlr_device, WLR_INPUT_DEVICE_POINTER,
+			&input_device_impl, name, vendor, product);
+
+	if (!(wlr_device->pointer = calloc(1, sizeof(struct wlr_pointer)))) {
+		wlr_log(WLR_ERROR, "Failed to allocate RDP pointer device");
+		return NULL;
+	}
+	wlr_pointer_init(wlr_device->pointer, NULL);
```
emersion: Right before, you can map the pointer to the output. See e.g. how the Wayland backend does it.

```c
wlr_dev->output_name = strdup(output->wlr_output.name);
```

include/wlr/backend/rdp.h
```diff
@@ -0,0 +1,32 @@
...
+#include <wlr/types/wlr_input_device.h>
+#include <wlr/types/wlr_output.h>
+
+/**
+ * Creates an RDP backend. An RDP backend will create one output, keyboard, and
+ * pointer for each client that connects.
+ * TODO: add freerdp options like bind port to this
+ */
+struct wlr_backend *wlr_rdp_backend_create(struct wl_display *display,
+	wlr_renderer_create_func_t create_renderer_func,
```
emersion: Indentation is off here

include/wlr/backend/rdp.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef WLR_BACKEND_RDP_H
+#define WLR_BACKEND_RDP_H
+#include <wlr/backend.h>
+#include <wlr/types/wlr_input_device.h>
+#include <wlr/types/wlr_output.h>
+
+/**
+ * Creates an RDP backend. An RDP backend will create one output, keyboard, and
+ * pointer for each client that connects.
+ * TODO: add freerdp options like bind port to this
```
emersion: Since there is `wlr_rdp_backend_set_port`, is this still relevant?

include/wlr/types/wlr_output.h
```diff
@@ -277,4 +277,14 @@ enum wl_output_transform wlr_output_transform_invert(
 enum wl_output_transform wlr_output_transform_compose(
 	enum wl_output_transform tr_a, enum wl_output_transform tr_b);
 
+/**
+ * Marks the output as in need of a buffer swap.
+ */
+void wlr_output_update_needs_swap(struct wlr_output *output);
```
emersion: This shouldn't be in this header. There's already a definition of this in `include/wlr/interfaces/wlr_output.h`. This function is intended to be used by backends, not compositors.

include/wlr/types/wlr_output.h
```diff
@@ -277,4 +277,14 @@ enum wl_output_transform wlr_output_transform_invert(
...
+
+/**
+ * Damages the whole output.
+ */
+void wlr_output_damage_whole(struct wlr_output *output);
```
emersion: Ditto

backend/meson.build
```diff
@@ -53,6 +53,22 @@ if logind.found()
 	backend_deps += logind
 endif
 
+if freerdp.found()
```
emersion: `and winpr2.found()`?

ddevault:
Addressed @emersion's comments and rebased into a single (changelog-ready) commit.

backend/rdp/peer.c
```diff
@@ -0,0 +1,363 @@
...
+		rdpInput *input, UINT16 flags, UINT16 x, UINT16 y) {
+	struct wlr_rdp_peer_context *context =
+		(struct wlr_rdp_peer_context *)input->context;
+	struct wlr_input_device *wlr_device = &context->pointer->wlr_input_device;
+	struct wlr_pointer *pointer = wlr_device->pointer;
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	struct wlr_event_pointer_motion_absolute event = { 0 };
+	int width, height;
+	wlr_output_effective_resolution(
```
emersion: `wlr_output.{width,height}` should be used here

backend/rdp/output.c
```diff
@@ -0,0 +1,289 @@
...
+		wlr_log(WLR_ERROR, "Raw updates are not supported; use rfx or nsc");
+	}
+	wlr_output_send_present(wlr_output, NULL);
+	return ret;
+}
+
+static void output_destroy(struct wlr_output *wlr_output) {
+	struct wlr_rdp_output *output =
+		rdp_output_from_output(wlr_output);
+	wl_event_source_remove(output->frame_timer);
```
emersion: This will segfault if `output->frame_timer` is NULL (not yet initialized).

Backtrace: https://gist.github.com/kennylevinsen/54cb7d55c1ec1d7e86d84b11573e0d87

backend/rdp/output.c
```diff
@@ -0,0 +1,289 @@
...
+}
+
+static EGLSurface egl_create_surface(struct wlr_egl *egl, unsigned int width,
+		unsigned int height) {
+	EGLint attribs[] = {
+		EGL_WIDTH, width,
+		EGL_HEIGHT, height,
+		EGL_RED_SIZE, 1,
+		EGL_GREEN_SIZE, 1,
+		EGL_BLUE_SIZE, 1,
```
emersion: Actually, we already ask for red/green/blue in `backend.c`, and these are invalid for `eglCreatePbufferSurface`. Can you revert this change?

emersion:
Thanks!

RorraVox:
Can someone explain in simple words how start freerpd with sway? I use remmina i think should be compatible.
I cant find nothing about that...
Thank you.

twhitehead:
The original patch put a bunch of info into *docs/env_vars.md*. Looks like it was dropped in ebdbe177d62d3c8bab04f77b86326804ff185bef in favour of [wayvnc](https://github.com/any1/wayvnc) though.

Issue #1576, opened 2019-02-25
dnd sometimes unresponsive


p-hamann:
After a certain uptime dnd becomes unresponsive. Then dnd is functional again after restarting the compositor (sway in my case).

Btw, thank you for the great work :-)

emersion:
Apparently serial validation stops working at some point.

See:
* https://github.com/swaywm/sway/issues/3538#issuecomment-466766603
* https://github.com/swaywm/sway/issues/3538#issuecomment-466681822

ddevault:
Is this not a dupe?

emersion:
Nope, the sway one is "doesn't work when there's a floating window", this one is "works but doesn't after some time". The other issue has been used as a general "it doesn't work" issue, so it's pretty confusing.

p-hamann:
I still could not identify the triggering event. Anyway, after a while - sometimes after a short and sometimes longer time - dnd is unresponsive in all native wayland windows. In xwayland windows, on the other hand, it works consistently

PR #1586, opened 2019-02-28
seat: guard against button count corruption

emersion:
This is still a compositor bug, and bad events will be sent to clients. We'll
need to track each button separately to handle this in wlroots.

Fixes https://github.com/swaywm/sway/issues/3538
Fixes https://github.com/swaywm/wlroots/issues/1576


RedSoxFan:
Thanks!

p-hamann:
This commit did not solve the issue, but the last three did. Thanks a lot!

PR #1588, opened 2019-02-28
meson: enable more compiler warnings

emersion:


agx:
Looks good to me!

meson.build
```diff
@@ -16,16 +16,31 @@ project(
...
+	'-Wredundant-decls',
+	'-Wimplicit-fallthrough=2',
+	'-Wendif-labels',
+	'-Wstrict-aliasing=2',
+	'-Woverflow',
+
+	'-Wno-missing-braces',
+	'-Wno-missing-field-initializers',
+	'-Wno-unused-parameter',
+]), language: 'c')
```
ddevault: Can we summarize this as `-Wall -Wextra -Wpedantic`?

emersion: No, `-Wpedantic` gives a lot of these:

```
../types/wlr_output_layout.c: In function ‘wlr_output_layout_add’:
../types/wlr_output_layout.c:211:66: warning: ISO C99 requires at least one argument for the "..." in a variadic macro
    wlr_log(WLR_ERROR, "Failed to create wlr_output_layout_output");
```

And there's no way to disable this specific warning. There's also no easy way to fix `wlr_log`.

ascent12: Also, meson already control those as as `warning_level` option.

emersion: Yeah, the idea was to bump `warning_level` to 3 to enable `-Wpedantic`

PR #1589, opened 2019-03-01
seat: only store serial if pressing a button

emersion:
The grab serial can be used to start a pointer grab. A button pressed event
should be used for this purpose.

Thus, we should only save the grab serial if it's the first button pressed
event we send. This commit makes it so the serial is not saved if a button is
released while another button is still pressed.


ddevault:
Thanks!

PR #1591, opened 2019-03-01
seat: fix remaining wlr_button_state enum rename

ianyfan:
Fixes regression introduced by merging #1589 and #1590 together.


emersion:
Thanks!

Issue #1437, opened 2018-12-17
Add a no-op session


emersion:
This will be necessary for VKMS.

PR #1592, opened 2019-03-01
backend/session: add noop session

emersion:
This is the first step towards being able to run via DRM leasing and on render
nodes.

Test with:

    export WLR_BACKENDS=drm
    export WLR_SESSION=noop
    export WLR_DRM_DEVICES=/dev/dri/renderD128

Fixes https://github.com/swaywm/wlroots/issues/1437


ddevault:
Thanks!

Issue #1442, opened 2018-12-22
Support for multiple xwayland instances


emersion:
Would be better to let compositors start as many xwayland instances as they want.

I don't think a lot of things are blocking this. We probably just need to make the compositor set `DISPLAY` instead of setting it automatically. Since the compositor already sets `WAYLAND_DISPLAY` it pobably makes more sense anyway.

Timidger:
Out of curiosity what is the use case for this? Sandboxing?

emersion:
Yeah

Freyr666:
Could this also be used to fix the annoying blurry upscale for X11 windows?

emersion:
No.

PR #1596, opened 2019-03-02
xwayland: don't set DISPLAY

emersion:
Let the compositor set it. This allows for multiple Xwayland instances to run
at the same time.

Fixes https://github.com/swaywm/wlroots/issues/1442

* * *

Breaking change: compositors now need to set `DISPLAY` after initializing Xwayland. The value `DISPLAY` needs to be set to is stored in `wlr_xwayland.display_name`.

### DIFF
--- a/xwayland/xwayland.c
+++ b/xwayland/xwayland.c
@@ -165,11 +165,8 @@ static void xwayland_finish_display(struct wlr_xwayland *wlr_xwayland) {
 
 	unlink_display_sockets(wlr_xwayland->display);
 	wlr_xwayland->display = -1;
-	unsetenv("DISPLAY");
+	wlr_xwayland->display_name[0] = '\0';
 }
-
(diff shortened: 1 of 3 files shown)


xwayland/xwayland.c
```diff
@@ -165,12 +165,9 @@ static void xwayland_finish_display(struct wlr_xwayland *wlr_xwayland) {
 
 	unlink_display_sockets(wlr_xwayland->display);
 	wlr_xwayland->display = -1;
-	unsetenv("DISPLAY");
+	wlr_xwayland->display_name[0] = '\0';
```
ddevault: rootston needs to unset this now

emersion: This is not necessary. `xwayland_finish_display` is only called if `xwayland_start_display` fails and in `wlr_xwayland_destroy`. When this happens rootston will exit anyway.

ddevault:
Thanks!

PR #1598, opened 2019-03-03
[WIP] Implement old rendering API on top of new backend API

ascent12:
This is work split out from the renderer v6 pull request (#1355) so that it easier to review and can be merged earlier.

This implements the old renderer interface on top of the new backend API. This is done as a (hacky) shim inside of the `wlr_backend` and `wlr_output` functions. This shouldn't break the existing `wlr_output`, `wlr_backend` or `wlr_renderer` APIs, and the new functions (e.g. `wlr_backend_get_render_fd`) are not expected to be used yet.

This shim will be removed when the `wlr_renderer` part of v6 is merged.

Due to the way I cherry picked the commits and started work, the commit history is a bit of a mess and doesn't entirely make sense. I may fix this in the future.

Backend status:
- [ ] DRM
  - [x] Atomic (partial)
  - [ ] Legacy
  - [ ] Cursors
  - [ ] Multi-GPU
- [x] X11
- [x] Wayland (partial)
  - [ ] Cursors
- [ ] Headless
- [ ] No-op


ascent12:
While I'm still not finished, I can say I got a bit carried away with some of the changes.
I wanted to get rid of that mess which was `realloc_crtcs`/`realloc_planes` and actually make use of atomic modesetting a lot better. You can see my new hopefully clearer version of the function called `atomic_apply_state`. I still need to write a legacy version, though (I realy wish we could just remove legacy DRM).

I've also removed parts of the code which weren't being used and just making things a lot more complicated, such as overlay planes. I'll add proper overlay support in the future, but the design that was there would not have worked anyway.

panaman67:
@ascent12  "(I really wish we could just remove legacy DRM)."

+1

PR #1600, opened 2019-03-03
xwm: use min size as base size hint if it is missing and vice versa

ammen99:
This is what ICCCM states that a WM should do.

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -583,10 +583,22 @@ static void read_surface_normal_hints(struct wlr_xwm *xwm,
 	memcpy(xsurface->size_hints, &size_hints,
 		sizeof(struct wlr_xwayland_surface_size_hints));
 
-	if ((size_hints.flags & XCB_ICCCM_SIZE_HINT_P_MIN_SIZE) == 0) {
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

xwayland/xwm.c
```diff
@@ -583,10 +583,22 @@ static void read_surface_normal_hints(struct wlr_xwm *xwm,
 	memcpy(xsurface->size_hints, &size_hints,
 		sizeof(struct wlr_xwayland_surface_size_hints));
 
-	if ((size_hints.flags & XCB_ICCCM_SIZE_HINT_P_MIN_SIZE) == 0) {
+	bool has_min_size_hints = (size_hints.flags & XCB_ICCCM_SIZE_HINT_P_MIN_SIZE) == 0;
```
emersion: The condition has been inverted in the process. Should be `!= 0`.

ammen99: i wonder why it even works ...

ammen99: oh because I have inverted the conditions below as well. I should never do PRs while I'm on the verge of falling asleep, sorry.

emersion: This condition is not inverted, but the `if` condition has been inverted, so it probably doesn't work?

In any case, the variable name is wrong. Could you send a PR please? :)

PR #1601, opened 2019-03-04
xwm: fix typos in WM_NORMAL_HINTS handling

ammen99:
What I proposed in #1600 was pretty wrong, I'm sorry for that. I didn't notice the problem because of the double bug. This PR should hopefully make things right.

@emersion Please check again :)


emersion:
Thanks for the fix!

PR #1602, opened 2019-03-05
seat: add debug logs when validating grab serials

emersion:
Makes it easier to debug when something goes wrong, e.g. button_count stuck
to 2 because the compositor ate a button release event.


ddevault:
Thanks!

PR #1603, opened 2019-03-05
Allow setting a custom tty via WLR_TTY env variable

jnschulze:
Comes in handy when starting the application from an SSH session etc.

### DIFF
--- a/backend/session/direct.c
+++ b/backend/session/direct.c
@@ -155,9 +155,19 @@ static int vt_handler(int signo, void *data) {
 }
 
 static bool setup_tty(struct direct_session *session, struct wl_display *display) {
-	int fd = open("/dev/tty", O_RDWR | O_CLOEXEC);
+
+	bool default_tty = false;
+
+	const char *tty_path = getenv("WLR_DIRECT_TTY");
+
+	if (!tty_path) {
+		tty_path = "/dev/tty";
+		default_tty = true;
+	}
+
+	int fd = open(tty_path, O_RDWR | O_CLOEXEC);
 	if (fd == -1) {
-		wlr_log_errno(WLR_ERROR, "Cannot open /dev/tty");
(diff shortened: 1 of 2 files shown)


backend/session/direct.c
```diff
@@ -155,9 +155,19 @@ static int vt_handler(int signo, void *data) {
 }
 
 static bool setup_tty(struct direct_session *session, struct wl_display *display) {
-	int fd = open("/dev/tty", O_RDWR);
+
+	bool default_tty = false;
+
+	const char *tty_path = getenv("WLR_TTY");
+
+	if(!tty_path) {
```
emersion: Style: missing space between `if` and `(`

emersion:
Maybe we should name this `WLR_DIRECT_TTY`, since it only affects the direct session?

jnschulze:
I've updated the documentation accordingly.

@emersion I see your point but considering the number of env variables we currently have I'd just stick with `WLR_TTY`.

emersion:
Can you rebase against master and squash your changes, so that the change appears as one clean commit?

> considering the number of env variables we currently have

Not sure I understand why this makes `WLR_TTY` better?

jnschulze:
> > considering the number of env variables we currently have
> 
> Not sure I understand why this makes `WLR_TTY` better?

I think as we don't have that much variables, we don't need the verbosity of `WLR_DIRECT_TTY`. But I'm fine with your suggestion. Shall I just change it to `WLR_DIRECT_TTY`?

emersion:
Let's just see what others think.

ascent12:
Yeah, I can see the niceness of being somewhat explicit about it being associated with the direct session. I don't want people mistaking this for having any sort of effect on logind.

jnschulze:
It can be merged now as I've renamed the variable as desired.

emersion:
Thanks!

PR #1605, opened 2019-03-05
backend/drm: Don't fail on failing to find overlay format

ascent12:
Some hardware exists which doesn't support XRGB/ARGB overlays, and we aren't even using overlay planes, so don't fail on trying to find a format.

This code will be overhauled in the renderer v6 and related work, but we'll get a quick fix out now for people that are affected.


emersion:
Thanks!

PR #1609, opened 2019-03-08
wlr_output_layout_get_box: handle empty layout

RedSoxFan:
Fixes swaywm/sway#3734

If there were no outputs in the output layout,
wlr_output_layout_get_box would return the box:

```
{
  .x = INT_MIN,
  .y = INT_MIN,
  .width = INT_MIN - INT_MAX,
  .height = INT_MIN - INT_MAX
}
```

which results in an integer underflow for both the width and height.

This changes the logic to have the box be all zeroes, since an empty
output layout does not have a width or height and the location of
something without a size is irrelevant so this just uses the origin.

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -380,22 +380,24 @@ struct wlr_box *wlr_output_layout_get_box(
 		}
 	} else {
 		// layout extents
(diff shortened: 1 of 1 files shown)


emersion:
Thanks!

PR #1614, opened 2019-03-14
Add ability to distinguish between touchpads and mice.

lolzballs:
Currently, both touchpads and mice are `WLR_INPUT_DEVICE_POINTER`, with no ability to distinguish between touchpad and mice.

This is necessary for swaywm/sway#3784, which allows users to create default input configs per device type (i.e. natural scrolling for all touchpads, but no natural scrolling for all mice).

To test this change, rootston was updated to reflect this, printing out `mouse` or `touchpad` depending on the input device. On my machine (XPS 9560), it was confirmed to work:
```
2019-03-14 11:55:04 - [rootston/input.c:71] New input device: DLL07BE:01 06CB:7A13 Touchpad (1739:31251) touchpad seat:seat0
2019-03-14 11:55:04 - [rootston/input.c:71] New input device: Logitech USB Receiver (1133:50487) mouse seat:seat0
```

### DIFF
--- a/backend/libinput/pointer.c
+++ b/backend/libinput/pointer.c
@@ -16,7 +16,8 @@ struct wlr_pointer *create_libinput_pointer(
 		wlr_log(WLR_ERROR, "Unable to allocate wlr_pointer");
 		return NULL;
 	}
-	wlr_pointer_init(wlr_pointer, NULL);
+	wlr_pointer_init(wlr_pointer, NULL,
+		libinput_device_config_tap_get_finger_count(libinput_dev) > 0);
 	return wlr_pointer;
 }
 
--- a/backend/headless/input_device.c
+++ b/backend/headless/input_device.c
@@ -51,7 +51,7 @@ struct wlr_input_device *wlr_headless_add_input_device(
 			wlr_log(WLR_ERROR, "Unable to allocate wlr_pointer");
 			goto error;
 		}
-		wlr_pointer_init(wlr_device->pointer, NULL);
+		wlr_pointer_init(wlr_device->pointer, NULL, false);
 		break;
 	case WLR_INPUT_DEVICE_TOUCH:
 		wlr_device->touch = calloc(1, sizeof(struct wlr_touch));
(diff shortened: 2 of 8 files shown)


backend/libinput/pointer.c
```diff
@@ -16,7 +16,8 @@ struct wlr_pointer *create_libinput_pointer(
 		wlr_log(WLR_ERROR, "Unable to allocate wlr_pointer");
 		return NULL;
 	}
-	wlr_pointer_init(wlr_pointer, NULL);
+	wlr_pointer_init(wlr_pointer, NULL,
+		libinput_device_config_tap_get_finger_count(libinput_dev) > 0);
```
ddevault: I think I would rather fish this interface through than use it as the basis for heuristics.

ddevault: What about touchpads which support just one finger?

lolzballs: > I think I would rather fish this interface through than use it as the basis for heuristics.

What do you mean by fishing it through the interface? Add another argument to create_libinput_pointer?

> What about touchpads which support just one finger?

I used what [clutter uses](https://gitlab.gnome.org/GNOME/mutter/blob/master/clutter/clutter/evdev/clutter-input-device-evdev.c#L1459). I'm not too familiar with the workings of libinput, but if a touchpad only supports one finger, shouldn't `libinput_device_config_tap_get_finger_count` return 1?

ddevault: No, I mean add a wlr_pointer_get_finger_count or something similar.

I think I'd sooner just not support this feature. I don't like heursitic driven features like this.

lolzballs: Fair enough.

This was something that I wanted so that I could configure sway input devices by device type. I think it would be helpful to a lot of people who put their configs in a dotfile repo and would like to have one config work on multiple devices.

I also did some more research on what libinput did and it doesn't seem like there's any other way to determine if an input device is a touchpad currently.

emersion: Heuristics should live in the compositor. We do expose libinput objects.

PR #1615, opened 2019-03-15
backend/wayland: improve output number handling

RedSoxFan:
This improves the way the output numbers are handled for the wayland
backend. Instead of using the number of active outputs plus one, the
last used number is stored and new outputs will increment it. This
fixes the situation where you start with one output, create a second,
close the first, and create a third. Without this, both outputs will be
`WL-2`, which causes an issue since the identifier will also be
identical. With this, the last output is `WL-3` and the outputs can be
distinguished.

I'll submit a PR to do the same for the x11 backend after this gets merged


emersion:
Thanks!

emersion:
The headless backend also needs this.

PR #1616, opened 2019-03-15
backend/x11: improve output number handling

RedSoxFan:
_Same as #1615, but for the x11 backend_

This improves the way the output numbers are handled for the x11
backend. Instead of using the number of active outputs plus one, the
last used number is stored and new outputs will increment it. This
fixes the situation where you start with one output, create a second,
close the first, and create a third. Without this, both outputs will be
X11-2, which causes an issue since the identifier will also be
identical. With this, the last output is X11-3 and the outputs can be
distinguished.


emersion:
Thanks!

PR #1620, opened 2019-03-17
util/log: setup wayland log handler

emersion:
We already do this for the libinput logs.


emersion:
Fixed the FreeBSD build.

PR #1628, opened 2019-03-22
[WIP] Add support for hi-res scroll wheel events

### LINKED ISSUE #1627: Support libinput v120 API hi-res mouse scroll events
This is coming soonish, but after libinput 1.13 (kernel support has already landed). I've got the WIP code from Peter Hutterer and am working on implementing it in wlroots, so just adding an issue here so as not to duplicate effort.

tokyovigilante:
Will close #1627 

Requires a version of libinput built from
https://gitlab.freedesktop.org/whot/libinput/tree/wip/hi-res-scrolling, so probably not for merging
until this is released.

Replaces LIBINPUT_EVENT_POINTER_AXIS (which is now legacy) with
LIBINPUT_EVENT_POINTER_AXIS_WHEEL, providing access to hi-res scroll
events using libinput_event_pointer_get_axis_value_v120(). These are
described in the libinput docs, but basically report events as fractions
of 120, where a full 120 is a single wheel click. The old
libinput_event_pointer_get_axis_value reports 15 or 20 degrees per click
(dependent on HW, 15 being most common, so the increase in sensitivity is
between 6x (20/120) to 8x (15/120), hence the 0.125 scale factor (1/8).

This seems to require no changes to sway, as it simply takes the wlr_event.delta
value as a double, and provides buttery-smooth scrolling in Vivaldi, ST3,
and other apps I've tested.


backend/libinput/events.c
```diff
@@ -246,6 +246,8 @@ void handle_libinput_event(struct wlr_libinput_backend *backend,
 		handle_pointer_button(event, libinput_dev);
 		break;
 	case LIBINPUT_EVENT_POINTER_AXIS:
+		break;
```
emersion: Would be nice to add a comment about this.

tokyovigilante:
LIBINPUT_EVENT_POINTER_AXIS events are still used for other scroll types (TrackPoint, finger, two-finger touchpads etc) so shouldn't have been globally disabled; restoring those in the latest pull.

backend/libinput/pointer.c
```diff
@@ -105,6 +105,11 @@ void handle_pointer_axis(struct libinput_event *event,
 		usec_to_msec(libinput_event_pointer_get_time_usec(pevent));
 	switch (libinput_event_pointer_get_axis_source(pevent)) {
 	case LIBINPUT_POINTER_AXIS_SOURCE_WHEEL:
+		if (libinput_event_get_type(event) == LIBINPUT_EVENT_POINTER_AXIS) {
+			/* legacy API in libinput 1.13+, all wheel events now also sent
```
whot: ftr, hi-res scrolling was pushed back to libinput 1.14

meson.build
```diff
@@ -70,7 +70,7 @@ egl            = dependency('egl')
 glesv2         = dependency('glesv2')
 drm            = dependency('libdrm', version: '>=2.4.95')
 gbm            = dependency('gbm', version: '>=17.1.0')
-libinput       = dependency('libinput', version: '>=1.9.0')
+libinput       = dependency('libinput', version: '>=1.12.902')
```
whot: this will need to be updated for 1.14 as mentioned above.

backend/libinput/pointer.c
```diff
@@ -131,9 +136,13 @@ void handle_pointer_axis(struct libinput_event *event,
...
-			double hr_delta =
-					libinput_event_pointer_get_axis_value_v120(pevent, axes[i]);
-			wlr_event.delta = hr_delta * 0.125; // assuming 15 degree/detent wheel
+			if (wlr_event.source == WLR_AXIS_SOURCE_WHEEL) {
+				// send scaled hi-res value assuming 15 degree/detent wheel
+				wlr_event.delta =
+						libinput_event_pointer_get_axis_value_v120(pevent, axes[i]) * 0.125;
+			} else {
+				wlr_event.delta = libinput_event_pointer_get_axis_value(pevent, axes[i]);
+			}
```
whot: you can calculate the right amount of movement by comparing the v120 value with the degrees. if you get 5 degrees and a v120 of 30, you know the full click angle is 20. But that in itself is a problem for all the reasons I pointed out in that email I sent out. And which I publish as blog post soon. So for now, scaling to 15 is the best thing you can do, but I'd do this based on the click angle so you always get a value of 15 regardless of the physical click angle (instead of hardcoding 1/8).

tokyovigilante:
> ...I don't think we can get away _without_ a protocol extension. Biggest problem is the handling of discrete events that clients interpret in specific ways and I can't find a way around that that _doesn't_ break some clients.

What issues do you see with the discrete events? These don't seem to have changed.

whot:
Basic problem: right now, some clients expect that they get a discrete event for every wheel event, i.e. a sequence of ``wl_pointer.axis_source, axis_discrete, axis, frame``. The axis event is discarded, the discrete event is treated as one wheel scroll event and all is good.

If you start sending out event sequences ``wl_pointer.axis_source, axis, frame``, the client will now use the axis event as there is no discrete event. So you scroll by say 1/8 of a wheel. Repeat 6 times, you've now scrolled 7/8 of a wheel. Then the last sequence comes in as ``wl_pointer.axis_source, discrete, axis, frame``. The client sees the discrete, scrolls by a full wheel event and you have now scrolled by 15/8 of a wheel instead of 8/8 of a wheel.

Because we multiplex devices through a single ``wl_pointer``, the client *cannot* know if it was 7/8 on one physical mouse, then 1 wheel click on another physical mouse. And since they don't get the v120 value yet, they have no way of guessing it either.

If you skip discrete events altogether, this *may* be all fine, I haven't done sufficient testing for that bit yet though.

IIRC Xwayland had exactly the problem above.

tokyovigilante:
Hmm, I didn't realise I was effectively discarding the discrete events. Would it make sense to just send through discrete events from `LIBINPUT_EVENT_POINTER_AXIS` and the high-res deltas from `LIBINPUT_EVENT_POINTER_AXIS_WHEEL`, acknowledging the potential issues from non-15 degree mice?

whot:
No, we can't, see the event sequence in my comment above. Any time you send a discrete event, *that* event will be taken as one full scroll event and cause a full scroll, on top of the fractions you've already scrolled. I think the only solution around this is to add a new v120 event that clients can handle accordingly. I read through the protocol specs and it's unfortunately ambiguous enough that we can't just claim "they're doing it wrong" and move on. We have to avoid breaking clients here.

tokyovigilante:
I guess that's my point, what clients are using both discrete and axis degree values currently? I don't see how replacing the degree amounts with the v120 scaled amounts is sufficiently different to current behaviour. I realise with my suggestion they won't quite line up as they do currently but I can't understand why a client would be using both at the same time.

whot:
discrete events are optional, so you're not guaranteed to get them. You don't get 0 as value, you just don't get it. So if I give you a few axis events but no discrete event, you'll have to handle those. If I then give you an axis event with a discrete event - which one are you going to use? And what value are you going to scroll by for that last event?

Best example: xwayland currently does this.

Note that the value in `wl_pointer.axis` events is already defined badly by the protocol (imo). merely as "distance in screen coordinates" which implies that the compositor defines how far a wheel would scroll. libinput without the v120 doesn't necessarily give you enough information to calculate fractions of a wheel movement. the protocol without a v120 just gives you values but no reference to anything. The v120 has the simple advantage of being a known range, so you *know* that 60 is "half a logical scroll". You cannot get to that with any of the other values we currently expose - sure, there's heuristics but that can go wrong easily, especially on the protocol where you can't differ between devices.

tokyovigilante:
And I thought it would be easy... ;) Thanks for the detailed explanations. 

There doesn't seem to be an easy solution to suit all scenarios, so I wonder whether we should just pragmatically allow opting into hi-res events for those who hava a hi-res capable mouse and don't have applications where seeing hi-res events from the new API and discrete events from the old API would cause problems. (Anecdotally, my xwayland browser has no issues with the hi-res events enabled and disabling discrete events entirely, but obviously that won't be true in all cases).

Alternative would be a new protocol as you suggest, but then if clients are already broken, what are the odds they will opt into the new behaviour?

whot:
yeah, i *think* just dropping discrete events is the immediate best solution. All clients must handle axis events anyway, so you shouldn't see any negative effects from dropping them. It will cause some minor changes in behaviour because you're removing the information "this was one click". So if you have a toolkit/app that has a setting for "mouse wheel clicks scroll by N lines" that would break now.That's the only side-effect I can think of. Also, I'm not sure any toolkit does that anyway. Long-term, discrete events should be deprecated anyway because they don't contain anything over v120.

> opting into hi-res events for those who hava a hi-res capable

libinput *always* gives you the hi-res events, so you don't necessarily know if they are truly hi-res. Heuristics maybe, but they're unreliable. 

> if clients are already broken

client's are broken because the discrete event isn't specified well enough and so far we simply always got discrete events from wheels. So the case of a zero-value discrete event wasn't something we had to worry about (to the point where the xf86-input-libinput driver would cause a division by zero). So everyone relies on this to some degree, we can't easily change it. And we can't point the finger at the clients either because the protocol isn't specified well enough to provide enough information even with zero discrete events. It's just a matter of "we set this in stone X years ago and now things are different". Such is life :)

tokyovigilante:
Have just seen @whot’s latest blog on this, looks like great stuff! Will digest it and update this PR shortly.

https://who-t.blogspot.com/2020/04/high-resolution-wheel-scrolling-in.html

whot:
ftr, the [wayland mr](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/72) has links to the weston and mutter changes, they should provide easy templates for wlroots.

tokyovigilante:
Thanks Peter. As an aside, would you mind pulling in https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1620 to your GTK3 build? Fixes a 3.24.16 bug which breaks Waybar. I've had to roll back to FC32's 3.24.17 for now.

tokyovigilante:
OK, this should be a bit closer. I've split out handling of the two event types (`LIBINPUT_EVENT_POINTER_AXIS` and `LIBINPUT_EVENT_POINTER_AXIS_WHEEL`) in `backend/libinput/pointer.c` to another function (`handle_pointer_axis_wheel`) but the new wheel events could probably be handled in a single slightly refactored `handle_pointer_axis`.

Probably needs to stay WIP a bit longer, as updates to at least wayland 1.18 and libinput 1.16 as well as the GTK changes are needed. Also, I've taken the liberty of upgrading the `wl_seat` protocol version to 8 from 5, not sure what else this impacts but will be required to land this.

Seems to be working pretty well on FC32, although I've noted what is probably a separate bug - my MX Master has a separate horizontal scroll wheel, but scroll events from this wheel in `wev` are reported as vertical scroll stop events. Will have to look into this some more.

whot:
> Thanks Peter. As an aside, would you mind pulling in https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1620 to your GTK3 build? Fixes a 3.24.16 bug which breaks Waybar. I've had to roll back to FC32's 3.24.17 for now.

I submitted a build for the rebase on top of 3.24.17, should finish in a few minutes.

backend/libinput/pointer.c
```diff
@@ -141,6 +141,54 @@ void handle_pointer_axis(struct libinput_event *event,
...
+	if (!wlr_dev) {
+		wlr_log(WLR_DEBUG, "Got a pointer event for a device with no pointers?");
+		return;
+	}
+	struct libinput_event_pointer *pevent =
+		libinput_event_get_pointer_event(event);
+	struct wlr_event_pointer_axis wlr_event = { 0 };
+	wlr_event.device = wlr_dev;
+	wlr_event.time_msec =
+		usec_to_msec(libinput_event_pointer_get_time_usec(pevent));
```
whot, line 157: i don't know the full context here, but why get usec when you cast to msecs anyway? just use `foobar_get_time()` instead?

backend/libinput/pointer.c
```diff
@@ -141,6 +141,54 @@ void handle_pointer_axis(struct libinput_event *event,
...
+
+	if (libinput_event_pointer_get_axis_source(pevent) != LIBINPUT_POINTER_AXIS_SOURCE_WHEEL) {
```
whot, line 159: you can get a wheel event for `POINTER_AXIS_SOURCE_TILT`, I think. it's a bit of a grey area, adding tilt was a bit of a mistake and no device actually uses it. And I'm not 100% sure on the libinput change just yet whether tilt will use it. So just keep that in mind.

backend/libinput/pointer.c
```diff
@@ -141,6 +141,54 @@ void handle_pointer_axis(struct libinput_event *event,
...
+		wlr_log(WLR_DEBUG, "Got a wheel axis event for a non-wheel device");
+		return;
+	}
+	wlr_event.source = WLR_AXIS_SOURCE_WHEEL;
+
+	const enum libinput_pointer_axis axes[] = {
+		LIBINPUT_POINTER_AXIS_SCROLL_VERTICAL,
+		LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL,
+	};
+	for (size_t i = 0; i < sizeof(axes) / sizeof(axes[0]); ++i) {
```
whot, line 169: you don't have an `ARRAY_SIZE` or a macro like that? do you really open-code this everywhere?

backend/libinput/pointer.c
```diff
@@ -141,6 +141,54 @@ void handle_pointer_axis(struct libinput_event *event,
...
+			case LIBINPUT_POINTER_AXIS_SCROLL_VERTICAL:
+				wlr_event.orientation = WLR_AXIS_ORIENTATION_VERTICAL;
+				break;
+			case LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL:
+				wlr_event.orientation = WLR_AXIS_ORIENTATION_HORIZONTAL;
+				break;
+			}
+			double hr_delta =
+					libinput_event_pointer_get_axis_value_v120(pevent, axes[i]);
+			if (hr_delta != 0) {
```
whot, line 181: `hr_delta` should never be zero, you wouldn't get a `AXIS_WHEEL` event otherwise.

backend/libinput/pointer.c
emersion: We don't use macros.

backend/libinput/pointer.c
tokyovigilante: This function is 99% copied from the existing `handle_pointer_axis` code so I didn't want to rock the boat too much.

backend/libinput/pointer.c
tokyovigilante: I guess this is more of a potential problem than a real issue then.

backend/libinput/pointer.c
tokyovigilante: Should or must never? The check is just because it would lead to a divide-by-zero error if it ever was.

whot: for libinput at least, I don't see a legitimate case of sending those events with a delta of zero (one axis maybe, but not the other one). but feel free to leave it there, better safe than sorry I guess.

tokyovigilante: If you're happy I'll never see a zero there I'm be happy to remove it, one less line run on every input is fine with me!

emersion: Could just `assert(delta != 0)` instead

tokyovigilante: Good thinking, thanks.

tokyovigilante:
Very rough reimplementation based on the updated API at http://who-t.blogspot.com/2020/04/high-resolution-wheel-scrolling-in.html.

Seems to work well enough with my MX Master 2S but there is quite a bit of divergence in wlroots' handling of pointer input compared to Weston which I was basing this update on, so more for feedback than anything else currently, although it seems like this version could be merged before the libinput and Wayland changes land given the `wl_seat` protocol bump. 

I'm using @whot's Fedora 32 COPR to test which supplies the required libinput API, Wayland protocol updates etc.

tokyovigilante:
Closing in favour of #3082. Thanks @JoseExposito, getting close now!

PR #1629, opened 2019-03-22
meson.build: require libinput >= 1.9.0

ammen99:
We use LIBINPUT_SWITCH_TABLET_MODE, which is introduced in 1.9.0


emersion:
Thanks!

Issue #1631, opened 2019-03-22
layershell popups do not show


agx:
Something between 0.4.1 and 0.5.0 broke layer shell popups (at least in rootston). Using examples/layershell and right clicking in the square no longer pops up a new square. I did not yet get a chance to have a closer look.

PR #1637, opened 2019-03-24
rootston: Also iterate layer shell popups

agx:
Broken by 62fd03a7beb18da6c4a5929e38739cf2199dfa79

Closes: #1631


ddevault:
Thanks!

emersion:
We should probably use `wlr_layer_surface_v1_for_each_surface` instead here. This PR doesn't handle popups of popups.

PR #1638, opened 2019-03-25
wlr_xdg_toplevel{,_v6}: store pending fullscreen output

RedSoxFan:
Related to swaywm/sway#1813
Required for swaywm/sway#3966

Since the fullscreen request may be made before the toplevel's surface
is mapped, the requested fullscreen output needs to be stored so it
can be retrieved on map (along with the existing fullscreen property).

### DIFF
--- a/types/xdg_shell_v6/wlr_xdg_toplevel_v6.c
+++ b/types/xdg_shell_v6/wlr_xdg_toplevel_v6.c
@@ -19,6 +19,11 @@ void destroy_xdg_toplevel_v6(struct wlr_xdg_surface_v6 *surface) {
 	assert(surface->role == WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL);
 	unmap_xdg_surface_v6(surface);
 
+	if (surface->toplevel->client_pending.fullscreen_output) {
+		struct wlr_xdg_toplevel_v6_state *client_pending =
+			&surface->toplevel->client_pending;
+		wl_list_remove(&client_pending->fullscreen_output_destroy.link);
+	}
 	wl_resource_set_user_data(surface->toplevel->resource, NULL);
 	free(surface->toplevel);
 	surface->toplevel = NULL;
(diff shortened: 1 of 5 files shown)


emersion:
What happens if the output is destroyed before the toplevel is mapped?

(Maybe we could just make the compositor save the output pointer)

RedSoxFan:
> What happens if the output is destroyed before the toplevel is mapped?

That does need to be handled
 
> (Maybe we could just make the compositor save the output pointer)

For some reason, I was incorrectly thinking that wasn't possible. I think that might actually be better so I'm going to close this and I'll rework the Sway PR

RedSoxFan:
> > (Maybe we could just make the compositor save the output pointer)
> 
> For some reason, I was incorrectly thinking that wasn't possible. I think that might actually be better so I'm going to close this and I'll rework the Sway PR

Actually nevermind. mpv is sending the fullscreen request before sway even creates the view. It looks like this does require a wlroots change. I'll fix the output destruction issue for this PR

emersion:
Indeed, it seems you're right. We emit the `new_surface` event right after the first NULL commit.

RedSoxFan:
Well I updated it to handle the requested fullscreen output being destroyed. I wish this PR was as small as I originally thought, but I'm not really sure of a better way of doing this.

RedSoxFan:
Actually, I just realized how I can make this smaller. Should have a new version up in a couple minutes so I'm not going it back to `[WIP]`

RedSoxFan:
Well this is ready for review again

types/xdg_shell_v6/wlr_xdg_toplevel_v6.c
```diff
@@ -19,6 +19,11 @@ void destroy_xdg_toplevel_v6(struct wlr_xdg_surface_v6 *surface) {
 	assert(surface->role == WLR_XDG_SURFACE_V6_ROLE_TOPLEVEL);
 	unmap_xdg_surface_v6(surface);
 
+	if (surface->toplevel->client_pending.fullscreen_output) {
+		struct wlr_xdg_toplevel_v6_state *client_pending =
+			&surface->toplevel->client_pending;
+		wl_list_remove(&client_pending->fullscreen_output_destroy.link);
```
emersion: We should probably also reset `fullscreen_output` to NULL here.

RedSoxFan: I didn't set it to NULL since the toplevel is being freed right after the if anyway. I can add it if you want though.

emersion: Oh, no, you're right.

emersion:
Thanks!

Issue #193, opened 2017-10-02
Implement direct scan-out for fullscreen views


ddevault:

ascent12:
DRM overlay planes can be used for this, where available. We already have all of the shit there to detect them, but we currently do not expose them in the API.

emersion:
@ascent12 Can you elaborate how the API could look like? We need to extend both `wlr_output` and `wlr_drm_connector`. Concerning the `wlr_output` side, is adding a `wlr_output_set_fullscreen(struct wlr_output *, struct wlr_surface *)` enough? Same for `wlr_output_interface`? Or do we want a more complicated interface?

linkmauve:
You shouldn’t need any API change, you can promote any surface matching the criteria to a plane, as long as you still have some available.

emersion:
We need a wlr_output API change.

PR #1641, opened 2019-03-29
Add support for direct scan-out

emersion:
This adds support for direct scan-out. This is a required feature before working on planes.

This PR tries to to as little changes as possible to make it easy to review and avoid re-writing everything. The current DRM backend and interface are not great but refactoring can happen incrementally with future PRs.

A new `wlr_output_attach_buffer` function has been introduced.

TODO:

- [x] We need to hold client buffers for longer than usual. When compositing, we need to hold buffers until the buffer swap (and the kernel will take care of the rest with implicit synchronization). When scanning out, we need to hold buffers as long as they're visible on screen.
- [x] Handle transformed surfaces
- [x] Handle DMA-BUF flags
- [x] Make software cursors force composition
- [x] Make screen capture force composition
- [x] Update wlr_output_damage: when switching from direct scan-out to rendering, we need to re-render all damage accumulated while scanning out
- [x] Test as many clients as possible

Test plan: try those fullscreen

- `weston-simple-dmabuf-drm`
- `weston-simple-dmabuf-egl`
- `weston-simple-egl`
- mpv
- a glfw demo

Follow-up issues:
- Use atomic test mode to check whether a configuration works
- Support Y_INVERT DMA-BUFs by setting "rotation" = "reflect-y" on planes

Fixes https://github.com/swaywm/wlroots/issues/193

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -172,7 +172,8 @@ static bool atomic_crtc_set_cursor(struct wlr_drm_backend *drm,
 	atomic_begin(crtc, &atom);
 
 	if (bo) {
-		uint32_t fb_id = get_fb_for_bo(bo, plane->drm_format);
+		uint32_t fb_id =
+			get_fb_for_bo(bo, plane->drm_format, drm->addfb2_modifiers);
 		set_plane_props(&atom, plane, crtc->id, fb_id, false);
 	} else {
 		atomic_add(&atom, plane->id, plane->props.fb_id, 0);
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -859,6 +859,15 @@ static bool drm_connector_schedule_frame(struct wlr_output *output) {
 	return true;
 }
 
+static uint32_t strip_alpha_channel(uint32_t format) {
+	switch (format) {
+	case DRM_FORMAT_ARGB8888:
+		return DRM_FORMAT_XRGB8888;
+	default:
+		return DRM_FORMAT_INVALID;
+	}
+}
+
 static bool drm_connector_set_dmabuf(struct wlr_output *output,
 		struct wlr_dmabuf_attributes *attribs) {
 	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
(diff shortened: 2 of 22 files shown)


emersion:
Proposal:

```c
// Same as wlr_output_make_current
bool wlr_output_attach_render(struct wlr_output *output);
// Instead of rendering to a buffer, use an existing DMA-BUF
bool wlr_output_attach_dmabuf(struct wlr_output *output,
	struct wlr_dmabuf_attributes *attrs);
// Submit a frame
bool wlr_output_present(struct wlr_output *output); // or commit, see #1640
```

emersion:
This is now working pretty well. Apart from the API not being final yet, there is one more issue: synchronization.

When compositing, we can send `wl_buffer.release` as soon as we've swapped buffers, because implicit synchronization will take care of holding the buffers until we've finished drawing.

When scanning out, we need to wait for the buffer to stop being visible before sending `wl_buffer.release`. I've implemented this in https://github.com/swaywm/wlroots/pull/1641/commits/27ee4bd9110e2fb350f2ad9100ad13ca2c21c8df by making rootston reference the `wlr_buffer` when scanning out. It works, but I'm not a fan of this approach: rootston can only guess when the backend is done with the buffer. Here with the DRM backend it needs to hold the buffer for two pageflips (because we draw right after pageflips, so there is one whole pageflip during which the buffer has been drawn and submitted to DRM but it's not yet visible). For the Wayland backend it would need to wait differently (depending on the `wl_buffer.release` signal of the parent compositor).

To fix this, I think the most straightforward solution is to pass the `wlr_buffer` to the backend. Thoughts?

emersion:
All right, this is ready for review. Note that on Intel, I915_FORMAT_MOD_Y_TILED can't be scanned out and is used by default on my machine (so you need to disable it to test direct scan-out).

Here is a patch that adds debugging logs and disables I915_FORMAT_MOD_Y_TILED: https://sr.ht/nd2t.diff

emersion:
Ping

ddevault:
I'm no expert on the implementation details, but the API looks solid and there are no obvious mistakes.

ddevault:
Thanks!

PR #1643, opened 2019-03-31
send proximity_out in proximity in for tablet_v2

Ongy:
When the proximity_in event is sent for tablet_v2 and there's already a
surface that currently has tablet (tool) focus, it should be told that
this is no longer the case, so we need to send the proximity_out event.


emersion:
Thanks!

PR #1649, opened 2019-04-08
noop: implement setting a custom mode

ammen99:
I see no reason not to implement this, plus it makes noop outputs a bit more usable (I find it useful to add them to the output layout, so I need to set mode)


ddevault:
Thanks!

PR #1650, opened 2019-04-08
Unbreak -Werror on 32-bit platforms

jbeich:
`size_t` requires `%z`, not `%l` modifier. See error logs: [armv7](https://github.com/swaywm/wlroots/files/3055849/wlroots-0.5.0.19.log), [i386](https://github.com/swaywm/wlroots/files/3055850/wlroots-0.5.0.19.log).

Wayland backend also used `%lu` instead of `%ld` but there should be no difference given `size_t` is unsigned (unlike `ssize_t`).


jbeich:
Sorry for minor noise due to (re)formatting commit message. I've started from headless regression and worked until build succeeded.

ddevault:
Thanks!

ascent12:
`%zu` is the correct format specifier for `size_t`. `%zd` is for `ssize_t`.

PR #1652, opened 2019-04-13
rootston: remove disabled outputs from the output layout

ammen99:
Removing from the output layout also means that `wlr_output_layout_get_box()` may return `null`, and we need to be more careful with that. I checked all places, and added checks where it may happen that the output in question is disabled.


emersion:
Thanks!

PR #1653, opened 2019-04-14
output: do nothing in output_set_image if backend has no renderer

ammen99:
This is useful when using the noop backend for example.


emersion:
Thanks!

PR #1655, opened 2019-04-16
idle: enable the compositors to add custom idle timeouts

ammen99:
Is something like this desirable in wlroots? I want it so that I can set DPMS from inside Wayfire (Wayfire doesn't have swaymsg and so a plugin must turn off the outputs).

### DIFF
--- a/include/wlr/types/wlr_idle.h
+++ b/include/wlr/types/wlr_idle.h
@@ -50,6 +50,11 @@ struct wlr_idle_timeout {
 	struct wl_listener seat_destroy;
 
 	void *data;
+	struct {
+		struct wl_signal idle;
+		struct wl_signal resume;
+		struct wl_signal destroy;
+	} events;
 };
 
 struct wlr_idle *wlr_idle_create(struct wl_display *display);
@@ -68,4 +73,15 @@ void wlr_idle_notify_activity(struct wlr_idle *idle, struct wlr_seat *seat);
  */
 void wlr_idle_set_enabled(struct wlr_idle *idle, struct wlr_seat *seat,
 	bool enabled);
+
+/**
+ * Create a new timer on the given seat. The idle event will be called after
+ * the given amount of milliseconds of inactivity, and the resumed event will
+ * be sent at the first user activty after the fired event.
+ */
+struct wlr_idle_timeout *wlr_idle_timeout_create(struct wlr_idle *idle,
+	struct wlr_seat *seat, uint32_t timeout);
+
+void wlr_idle_timeout_destroy(struct wlr_idle_timeout *timeout);
+
 #endif
--- a/include/wlr/types/wlr_idle.h
+++ b/include/wlr/types/wlr_idle.h
@@ -46,15 +46,16 @@ struct wlr_idle_timeout {
 	bool enabled;
 	uint32_t timeout; // milliseconds
 
-	struct wl_listener input_listener;
-	struct wl_listener seat_destroy;
-
-	void *data;
 	struct {
 		struct wl_signal idle;
 		struct wl_signal resume;
 		struct wl_signal destroy;
 	} events;
+
+	struct wl_listener input_listener;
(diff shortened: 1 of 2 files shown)


types/wlr_idle.c
```diff
@@ -246,3 +216,65 @@ struct wlr_idle *wlr_idle_create(struct wl_display *display) {
 void wlr_idle_notify_activity(struct wlr_idle *idle, struct wlr_seat *seat) {
 	wlr_signal_emit_safe(&idle->events.activity_notify, seat);
 }
+
+struct wlr_idle_timeout *wlr_idle_timeout_create(struct wlr_idle *idle,
+	struct wlr_seat *seat, uint32_t timeout) {
```
ddevault: style: indent the second line by two indents

include/wlr/types/wlr_idle.h
```diff
@@ -68,4 +73,15 @@ void wlr_idle_notify_activity(struct wlr_idle *idle, struct wlr_seat *seat);
  */
 void wlr_idle_set_enabled(struct wlr_idle *idle, struct wlr_seat *seat,
 	bool enabled);
+
+/**
+ * Create a new timer on the given seat. The fired event will be called after
```
ddevault: Do you mean the idle event?

ddevault:
+1

include/wlr/types/wlr_idle.h
ammen99: Oh yeah, I initially called this `fired` then decided `idle` is a better name

ammen99:
Updated to address the comments from @ddevault

types/wlr_idle.c
```diff
@@ -246,3 +216,65 @@ struct wlr_idle *wlr_idle_create(struct wl_display *display) {
...
+		free(timer);
+
+		return NULL;
+	}
+
+	if (timer->enabled) {
+		// arm the timer
+		wl_event_source_timer_update(timer->idle_source, timer->timeout);
+		if (timer->timeout == 0) {
+			idle_notify(timer);
```
emersion: This won't work, `timer->resource` is NULL at this point.

I guess the solution will be to keep a private constructor with a `resource` argument, and expose a public constructor which provides `resource = NULL` to the private constructor.

include/wlr/types/wlr_idle.h
```diff
@@ -50,6 +50,11 @@ struct wlr_idle_timeout {
 	struct wl_listener seat_destroy;
 
 	void *data;
+	struct {
+		struct wl_signal idle;
+		struct wl_signal resume;
+		struct wl_signal destroy;
+	} events;
```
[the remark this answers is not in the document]
emersion: Hmm. Seemed pretty consistent to me. Do you have examples in mind?

The idea is to move the "internal" things at the end of the struct.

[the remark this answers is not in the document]
ammen99: Ah ok.
What I have seen is for example foreign-toplevel (granted, it was implemented by me :)
But anyway, data comes usually after the events.

ammen99:
@emersion Updated with a private constructor, hope I haven't missed anything this time.

types/wlr_idle.c
```diff
@@ -101,34 +100,44 @@ static void handle_input_notification(struct wl_listener *listener, void *data)
...
-		return;
+		if (idle_resource) {
+			wl_resource_post_no_memory(idle_resource);
+		}
+		return NULL;
+	}
+
+	if (idle_resource)
+	{
+		timer->resource = wl_resource_create(
```
emersion: Hmm. Can we just create the resource in the caller, and `wl_resource_set_user_data(resource, timer)` here?

ammen99: You're right, makes more sense. I totally forgot I can just check user data later.

emersion:
Thanks!

ammen99:
Thanks for the reviews!

PR #1658, opened 2019-04-19
data-device: ignore accept for selection offers

emersion:
It doesn't make sense for clients to send "accept" requests to offers that
aren't drag-and-drop. I discussed with Daniel Stone to make it a protocol
error [1] but too many clients send it (e.g. GTK+). Let's just log it for now.

[1]: https://gitlab.freedesktop.org/wayland/wayland/merge_requests/11#note_149710


ddevault:
Thanks!

PR #1659, opened 2019-04-21
Fix missing headers when building without X11

alyssais:
The deleted includes are redundant, because other headers will include the necessary files. Additionally, they cause build failures, because including EGL/egl.h or EGL/eglext.h directly, instead of through wlr/render/egl.h or wlr/render/interface.h, will mean that `MESA_EGL_NO_X11_HEADERS` will not have been defined, and so the EGL headers will attempt to pull in unnecessary X11 headers that may not exist on the system.

For the headers produced by glgen.sh, the includes couldn't simply be deleted, because no other header would include the EGL headers. Neither wlr/render/egl.h or wlr/render/interface.h felt appropriate to include, so I opted instead to copy the `MESA_EGL_NO_X11_HEADERS` definition before
the EGL includes.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -2,8 +2,6 @@
 #include <assert.h>
 #include <drm_fourcc.h>
 #include <drm_mode.h>
-#include <EGL/egl.h>
-#include <EGL/eglext.h>
 #include <errno.h>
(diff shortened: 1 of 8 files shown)


emersion:
Thanks!

PR #1660, opened 2019-04-22
output: refactor frame submission API

emersion:
This is necessary for direct scan-out and other upcoming features. This patch
changes the output API to look like the wl_surface API.

Outputs now have some double-buffered state: the frame to be submitted
(currently only wlr_renderer frames are supported) and the damaged region.
To attach a pending frame, use wlr_output_attach_render. To set the pending
damaged region, use wlr_output_set_damage.

To submit the pending state, call wlr_output_commit. This will submit the
pending frame to the backend.

To migrate from the old API to the new one:

- Replace wlr_output_make_current calls by wlr_output_attach_render
- Replace wlr_output_swap_buffers calls by wlr_output_set_damage and
  wlr_output_commit
- Replace wlr_output_damage_make_current calls by wlr_output_damage_attach_render
- Replace wlr_output_damage_swap_buffers with wlr_output_set_damage and wlr_output_commit
- Replace wlr_output.needs_swap with wlr_output.needs_frame
- Replace wlr_output.events.swap_buffers with wlr_output.events.precommit

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -238,14 +238,13 @@ static struct wlr_drm_connector *get_drm_connector_from_output(
 	return (struct wlr_drm_connector *)wlr_output;
 }
 
-static bool drm_connector_make_current(struct wlr_output *output,
+static bool drm_connector_attach_render(struct wlr_output *output,
 		int *buffer_age) {
 	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
 	return make_drm_surface_current(&conn->crtc->primary->surf, buffer_age);
 }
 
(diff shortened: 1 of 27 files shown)


ascent12:
Overall seems good to me.

agx:
Looks good to me too. Would it make sense to tag a 0.6.0 afterwards so compositors can pick that up?

emersion:
Updated the `wlr_output` backend interface and the `wlr_output_damage` interface. Ready for another review.

ddevault:
>Would it make sense to tag a 0.6.0 afterwards so compositors can pick that up

Only once one asks for it (which will happen no later than sway 1.1)

ddevault:
Thanks!

PR #1661, opened 2019-04-22
Add wlr_output_preferred_mode

emersion:
This is a little more robust than assuming the last mode is the preferred one. This also makes compositors' code clearer.

The second commit also fixes rootston setting the preferred mode and then changing it to the configured mode (which causes delays at startup).


ddevault:
Thanks!

PR #1663, opened 2019-04-23
render/egl: fix zero-length VLA

emersion:
UBSan says:

```
../render/egl.c:370:10: runtime error: variable length array bound evaluates to non-positive value 0
```


ddevault:
Thanks!

Issue #1197, opened 2018-08-22
Add support for EGL_KHR_partial_update


emersion:
This can be used instead of `EGL_EXT_buffer_age` for better performance.

Spec: https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_partial_update.txt

Full explanation here: https://gitlab.freedesktop.org/wayland/weston/issues/134

---

wlroots has migrated to gitlab.freedesktop.org. This issue has been moved to:

https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/1197

emersion:
The rendering API will need a refactoring to support this. In particular:

* We should query `EGL_BUFFER_AGE_KHR` (fallback to `EGL_EXT_buffer_age` if unsupported) when making a surface current
* We should call `eglSetDamageRegionKHR` with the back buffer damage before rendering anything
* We should still call `eglSwapBuffersWithDamageKHR` with the surface damage when swapping buffers

It would be nice to add sanity checks along the way, since `EGL_EXT_buffer_age` has quite a few undefined behaviour cases.

emersion:
It's not possible to support this extension with renderer v6, because we're no longer using `EGLSurface`. We'd need a new EGL extension which works on FBOs.

PR #1668, opened 2019-04-25
Don't submit too much damage

### LINKED ISSUE #1665: Too much damage is submitted when swapping buffers
According to the [EGL_EXT_swap_buffers_with_damage](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_swap_buffers_with_damage.txt) spec we should submit a:

>list of rectangles that define the region that has truly changed since the last frame 

However we submit too much: we accumulate damage from past buffers as well. In other words, we submit back buffer damage instead of surface damage. [EGL_KHR_partial_update](https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_partial_update.txt) has a nice diagram that illustrates this.

Note that we should still use back buffer damage to redraw: we need to ensure the entire back buffer is consistent.

emersion:
See the individual commits.

I wonder whether `wlr_output_damage_attach_render` should call `wlr_output_set_damage` for you. Compositors can always call it again to override it (or stop using `wlr_output_damage`, it's just a helper after all). @ammen99 Would this be something you'd want to override?

Fixes #1665

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -222,11 +222,16 @@ bool wlr_output_attach_render(struct wlr_output *output, int *buffer_age);
 bool wlr_output_preferred_read_format(struct wlr_output *output,
 	enum wl_shm_format *fmt);
 /**
- * Set the damage region for the frame to be submitted.
+ * Set the damage region for the frame to be submitted. This is the region of
+ * the screen that has changed since the last frame.
  *
  * Compositors implementing damage tracking should call this function with the
  * damaged region in output-buffer-local coordinates (ie. scaled and
  * transformed).
+ *
+ * This region is not to be confused with the renderer's buffer damage, ie. the
+ * region compositors need to repaint. Compositors usually need to repaint more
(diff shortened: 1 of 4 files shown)


ammen99:
@emersion I am a bit confused by what exactly you are asking, I'll need to look more closely into the commits. Anyway Wayfire doesn't always know what will be repainted at the very start of the frame (i.e on `attach_renderer`), so I need to set this damage at the end of the frame.

emersion:
I'm asking whether you take the damage from `wlr_output_damage` and pass it to `wlr_output` directly without modifying it. Apparently no?

ammen99:
No, sometimes I need to repaint more (even the whole output).

emersion:
No, this is ready to be merged. I'd like the release notes to indicate that rootston's behaviour was wrong (so that compositors using it as a reference can be fixed).

I'd also like to explore options to make it easier for compositors to get it right, I guess this should wait for https://github.com/swaywm/wlroots/issues/1197.

PR #1671, opened 2019-04-26
Drop fixup_epoll in .builds/freebsd.yml

jbeich:
`/latest` has `epoll-shim.pc` after https://github.com/freebsd/freebsd-ports/commit/68b90c9df610ee3d99c719360256fcacf2e83603

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -21,21 +21,6 @@ packages:
 sources:
 - https://github.com/swaywm/wlroots
 tasks:
-- fixup_epoll: |
-    cat << 'EOF' | sudo tee /usr/local/libdata/pkgconfig/epoll-shim.pc
-    prefix=/usr/local
-    exec_prefix=\$\{\$prefix\}
-    libdir=${prefix}/lib
-    sharedlibdir=${prefix}/lib
-    includedir=${prefix}/include/libepoll-shim
-    Name: epoll-shim
-    Description: epoll shim implemented using kevent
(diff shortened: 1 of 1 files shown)


emersion:
Gah, maybe we need to wait a few days

jbeich:
A few hours. [The build](http://beefy9.nyi.freebsd.org/build.html?mastername=112amd64-default&build=499924) has completed, so it's probably in the process of syncing to mirrors.

ddevault:
I can also freshen up the FreeBSD image on builds.sr.ht if necessary, lmk

emersion:
Since we update packages, we should be fine. I'll restart the build in a few hours :)

emersion:
@jbeich You fork is private or deleted, so I can't restart the build. Feel free to force push to restart the build whenever.

jbeich:
Sorry, I've deleted it. As repo name is no longer known to GitHub force-pushing doesn't update the PR even if forked again. Given https://github.com/Hjdskes/cage/pull/67 built fine it should be safe to merge this as is.

FWIW, I've asked GitHub for support in case I'm missing something but...
> It isn't currently possible to link a deleted branch back to its existing Pull Requests I'm afraid, but we'll pass your request onto the team to consider.
>
> I can't promise if or when it will be implemented, but your suggestion is definitely in the right hands.

ddevault:
We don't merge PRs on a hope that they'll pass CI later... you'll have to open a new one.

PR #1673, opened 2019-04-26
Implement support for foreign-toplevel fullscreening

ammen99:
This is the wlroots side of https://github.com/swaywm/wlr-protocols/pull/46

### DIFF
--- a/examples/foreign-toplevel.c
+++ b/examples/foreign-toplevel.c
@@ -7,6 +7,8 @@
 #include <wayland-client.h>
 #include "wlr-foreign-toplevel-management-unstable-v1-client-protocol.h"
 
+#define WLR_FOREIGN_TOPLEVEL_MANAGEMENT_VERSION 2
+
 /**
  * Usage:
  * 1. foreign-toplevel
@@ -29,10 +31,11 @@
  */
 
 enum toplevel_state_field {
-	TOPLEVEL_STATE_MAXIMIZED = 1,
-	TOPLEVEL_STATE_MINIMIZED = 2,
-	TOPLEVEL_STATE_ACTIVATED = 4,
-	TOPLEVEL_STATE_INVALID = 8,
+	TOPLEVEL_STATE_MAXIMIZED = (1 << 0),
+	TOPLEVEL_STATE_MINIMIZED = (1 << 1),
+	TOPLEVEL_STATE_ACTIVATED = (1 << 2),
+	TOPLEVEL_STATE_FULLSCREEN = (1 << 3),
+	TOPLEVEL_STATE_INVALID = (1 << 4),
 };
 
 struct toplevel_state {
@@ -102,6 +105,9 @@ static void print_toplevel_state(struct toplevel_v1 *toplevel, bool print_endl) 
 	} else {
 		printf(" inactive");
 	}
+	if (toplevel->current.state & TOPLEVEL_STATE_FULLSCREEN) {
+		printf(" fullscreen");
+	}
 
 	if (print_endl) {
 		printf("\n");
@@ -152,6 +158,8 @@ static uint32_t array_to_state(struct wl_array *array) {
 			state |= TOPLEVEL_STATE_MINIMIZED;
 		if (*entry == ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_ACTIVATED)
 			state |= TOPLEVEL_STATE_ACTIVATED;
+		if (*entry == ZWLR_FOREIGN_TOPLEVEL_HANDLE_V1_STATE_FULLSCREEN)
+			state |= TOPLEVEL_STATE_FULLSCREEN;
 	}
 
 	return state;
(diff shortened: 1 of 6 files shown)


examples/foreign-toplevel.c
```diff
@@ -29,10 +29,11 @@
...
 enum toplevel_state_field {
-	TOPLEVEL_STATE_MAXIMIZED = 1,
-	TOPLEVEL_STATE_MINIMIZED = 2,
-	TOPLEVEL_STATE_ACTIVATED = 4,
-	TOPLEVEL_STATE_INVALID = 8,
+	TOPLEVEL_STATE_MAXIMIZED  = 1,
+	TOPLEVEL_STATE_MINIMIZED  = 2,
+	TOPLEVEL_STATE_ACTIVATED  = 4,
+	TOPLEVEL_STATE_FULLSCREEN = 8,
+	TOPLEVEL_STATE_INVALID    = 16,
```
emersion: Can we use bitshifts for this?

ammen99: Yes. Should I change it in the wlroots header, too?

examples/foreign-toplevel.c
```diff
@@ -102,6 +103,9 @@ static void print_toplevel_state(struct toplevel_v1 *toplevel, bool print_endl)
 	} else {
 		printf(" inactive");
```
panaman67: Can this be:
```suggestion
		puts(" inactive")
```

examples/foreign-toplevel.c
```diff
@@ -102,6 +103,9 @@ static void print_toplevel_state(struct toplevel_v1 *toplevel, bool print_endl)
...
 	}
+	if (toplevel->current.state & TOPLEVEL_STATE_FULLSCREEN) {
+		printf(" fullscreen");
+	}
 
 	if (print_endl) {
 		printf("\n");
```
panaman67:
```suggestion
		putchar('\n');
```

emersion: Nah, the compiler will do it for you

examples/foreign-toplevel.c
emersion: Ditto. Plus missing the `;`

examples/foreign-toplevel.c
emersion: That would be better yeah

examples/foreign-toplevel.c
```diff
@@ -29,10 +29,11 @@
...
 enum toplevel_state_field {
-	TOPLEVEL_STATE_MAXIMIZED = 1,
-	TOPLEVEL_STATE_MINIMIZED = 2,
-	TOPLEVEL_STATE_ACTIVATED = 4,
-	TOPLEVEL_STATE_INVALID = 8,
+	TOPLEVEL_STATE_MAXIMIZED  = (1 << 0),
+	TOPLEVEL_STATE_MINIMIZED  = (1 << 1),
+	TOPLEVEL_STATE_ACTIVATED  = (1 << 2),
+	TOPLEVEL_STATE_FULLSCREEN = (1 << 3),
+	TOPLEVEL_STATE_INVALID    = (1 << 4),
```
emersion: Style: we don't align

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -84,6 +87,12 @@ struct wlr_foreign_toplevel_handle_v1_activated_event {
 	struct wlr_seat *seat;
 };
 
+struct wlr_foreign_toplevel_handle_v1_fullscreen_event {
+	struct wlr_foreign_toplevel_handle_v1 *toplevel;
+	bool fullscreen;
+	struct wl_resource *output;
```
emersion: Oh, can we expose the `wlr_output` instead of the `wl_resource`?

ammen99: Oh that is my bad, sorry. When I first wrote this, I forgot that there is a `wlr_output_from_resource`

ammen99:
Updated again, thanks for the reviews so far @emersion

examples/foreign-toplevel.c
```diff
@@ -236,7 +242,7 @@ static void handle_global(void *data, struct wl_registry *registry,
 	} else if (strcmp(interface,
 			zwlr_foreign_toplevel_manager_v1_interface.name) == 0) {
 		toplevel_manager = wl_registry_bind(registry, name,
-				&zwlr_foreign_toplevel_manager_v1_interface, 1);
+				&zwlr_foreign_toplevel_manager_v1_interface, version);
```
emersion: This is dangerous. For instance, if the compositor supports v3 and v3 happens to add a new event, this will make the client crash. Clients usually bind to `min(client_version, compositor_version)`.

ammen99: I did this here since the example client will always be updated to support the compositor version?

ammen99: But if we want to make this copy-paste-proof I get your point

emersion: Yeah, I'd rather hardcode it in the client if we want to only support the latest version

ammen99: well it requires the latest version, so I'll just hardcode 2 for now

emersion:
Thanks!

PR #1675, opened 2019-04-27
meson: fix wayland-server minimum required version

emersion:
Having 1.16 results in the following error when running the compositor:

  2019-04-27 17:30:50 - [wayland] wl_global_create: implemented version for 'wl_seat' higher than interface version (7 > 6)
  2019-04-27 17:30:50 - [sway/input/seat.c:428] seat_create:could not allocate seat

We require wayland-server >= 1.17 for wl_seat version 7.

Fixes: a671fc51d25c2adb723d2edf96c36011211f4b81

cc @ascent12

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -95,7 +95,7 @@ if cc.get_id() == 'clang'
 	add_project_arguments('-Wno-missing-braces', language: 'c')
 endif
 
-wayland_server = dependency('wayland-server', version: '>=1.16')
+wayland_server = dependency('wayland-server', version: '>=1.17')
 wayland_client = dependency('wayland-client')
 wayland_egl = dependency('wayland-egl')
 wayland_protos = dependency('wayland-protocols', version: '>=1.17')
--- a/types/seat/wlr_seat.c
+++ b/types/seat/wlr_seat.c
@@ -254,13 +254,8 @@ struct wlr_seat *wlr_seat_create(struct wl_display *display, const char *name) {
 	seat->touch_state.seat = seat;
 	wl_list_init(&seat->touch_state.touch_points);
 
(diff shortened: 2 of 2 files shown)


ddevault:
Sent a patch to Alpine for 1.17

emersion:
Hmm. This one is nasty because it'll cause a runtime error if you compile with 1.16 - not a compile-time error. Maybe we should revert @ascent12's patch until we get updated packages.

ddevault:
Can we detect at runtime which version is in use and adjust our wl_seat version accordingly?

emersion:
In this particular case we can. This should stay temporary though, we're just lucky that new ABI hasn't been introduced.

jbeich:
I've filed [FreeBSD bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237650) for 1.17.

emersion:
I'm starting to wonder whether we should hold off landing patches because one of our CI distributions have an outdated dependency. This isn't that of a big deal for this PR, but maybe we should consider building wayland or wayland-protocols from source for other PRs.

ddevault:
-1, I think we should be dilligent about pushing distros to keep their things up to date.

emersion:
Then let's block the DRM lease PR until released in wayland-protocols in all distros?

ddevault:
Not really opposed to (1) blocking until the next wayland-protocols release which includes it and (2) the landing of that release in each distro we CI against. We need to usher that protocol into stability anyway to land it in Vulkan.

emersion:
FreeBSD now has Wayland 1.17, so this PR is ready for review.

PR #1676, opened 2019-04-27
rootston: fix damage tracking debug mode

emersion:
We want to damage the whole output in this mode. However if we overwrite the
damaged region after it's useless.

Rename `output_damage` to `frame_damage` while I'm at it, it's a better name.

Fixes: 57d32d03a87ab4a74230294cefc5fa8b1e28c96d


ddevault:
Thanks!

PR #1677, opened 2019-04-27
seat: fallback to v6 if libwayland 1.17 isn't available

emersion:
See https://github.com/swaywm/wlroots/pull/1675#issuecomment-487300445


ddevault:
Thanks!

PR #1681, opened 2019-04-29
Drop fixup_epoll in .builds/freebsd.yml

jbeich:
`/latest` has `epoll-shim.pc` as it contains https://github.com/FreeBSDDesktop/epoll-shim/pull/1 after https://github.com/freebsd/freebsd-ports/commit/68b90c9df610ee3d99c719360256fcacf2e83603

Same as #1671.


emersion:
Thanks!

PR #1682, opened 2019-04-29
ci: enable Meson auto features

emersion:
This makes it so we don't miss build failures because the CI misses a
dependency.


ddevault:
Thanks!

Issue #894, opened 2018-04-23
DRM Explicit fencing


ascent12:
Another one of these new shiny graphics-related features I eventually want to get around to.
Basically, the purpose is to stop slow clients from stalling the graphics pipeline by controlling synchronization ourselves. It may also help with multi-GPU.

I'm not sure what impact this is going to have on the codebase, and how it may apply to non-DRM backends, if at all.

The `zwp_linux_explicit_synchronization_v1` protocol is used for this, although has not been merged at the time of writing.
[EGL_ANDROID_native_fence_sync](https://www.khronos.org/registry/EGL/extensions/ANDROID/EGL_ANDROID_native_fence_sync.txt) is related.
It's also related to Vulkan Fences/Semaphores.

Extra links:
https://www.collabora.com/news-and-blog/blog/2016/09/13/mainline-explicit-fencing-part-1/
https://www.collabora.com/news-and-blog/blog/2016/10/18/mainline-explicit-fencing-part-2/
https://www.collabora.com/news-and-blog/blog/2017/01/26/mainline-explicit-fencing-part-3/

emersion:
Here's the protocol proposal: https://patchwork.freedesktop.org/patch/177866/

ascent12:
After a lot more investigation to how this would be implemented, it turns out the nature of this problem is a lot different than I thought it was.
The way this would ideally be implemented, it's far less of a renderer issue and more of a state tracking issue.

Just so everyone is on the same page, a dma-fence is a kernel object which can signal the completion of asynchronous GPU operations. They are exposed to userspace using "sync files" which can be made up of multiple fences. Fortunately some kernel documentation for this exists:
https://github.com/torvalds/linux/blob/master/Documentation/sync_file.txt
Userspace API header:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/sync_file.h
An important thing to note is that these can be used will poll(2)-like APIs.

Now when it comes to implementing `zwp_linux_explicit_synchronisation`, there are two parts to the issue.

# Passing a fence to the client

This part isn't that tricky, but you still need to be a little careful.
When using atomic DRM, there is a property called `OUT_FENCE_PTR`, which will return a sync-file for us to indicate when the pageflip is complete. This just needs to be sent to the client.

However, you need to be careful if you're using the client's buffer multiple times (e.g. spanning 2 outputs) or if you're going to hold onto the client's buffer longer than one frame. sync-files can be merged together, so it's not really that hard to do.

[...]

PR #1685, opened 2019-05-03
Groundwork for wp_linux_explicit_synchronization

### LINKED ISSUE #1546: Proposal: wlr_surface_state improvements
Right now, if an extension protocol wants to have any state tied to a particular wl_surface commit, it's basically forced to implement its own tracking of different commits, or forces the burden onto the user, and there is often no value in making them do it.

Instead, I think it'd be worth creating a more general system for this, where extension protocols (and even non-protocols) can add their own state directly to a wlr_surface and can be managed properly over multiple commits. One thing in particular I think we need in the future (for explicit sync) is tracking of more than just a current + pending commit, and I think that this would make this easier.

It will allow us to easily take "snapshots" of the entirety of a client's state, instead of just their buffer. This is somewhat important for things like `wp_viewporter` and other extensions that affect rendering state.

```c
struct wlr_compositor {
    // ...
    struct {
        struct wl_signal new_surface;
        struct wl_signal new_state;
    } events;
};

struct wlr_compositor_new_state_args {
    // Will be NULL if this is the first state.
    // If the protocol requires it, you should copy your old state from here
    struct wlr_surface_state *old;
    struct wlr_surface_state *new;
};

struct wlr_compositor_create(struct wl_display); // As normal
// Returns a unique id which can be used for the following functions.
// Is essentially 'return ++comp->ids;'
uint32_t wlr_compositor_register(struct wlr_compositor *comp);

struct wlr_surface_state {
    size_t ref_cnt;
    // ...
    struct {
        struct wl_signal commit;
        // Called when reference count is 0, and state is superseded by a new one
        struct wl_signal destroy;
    } events;

    // Something implementing what is essentially a (uint32_t -> void *) hashmap
};

void wlr_surface_state_add(struct wlr_surface_state *st, uint32_t id, void *data);
void *wlr_surface_state_get(struct wlr_surface_state *st, uint32_t id);
void wlr_surface_state_remove(struct wlr_surface_state *st, uint32_t id);

// Instead of users probing wlr_surface directly, they should use this function when it
// comes time to render. +1 to reference count.
struct wlr_surface_state *wlr_surface_get_state(struct wlr_surface *s);
struct wlr_surface_state *wlr_surface_get_pending(struct wlr_surface *s);
void wlr_surface_state_unref(struct wlr_surface_state *st);
```
## Examples

Here are some examples of what I think the usage would look like. I've skipped over some of the details, but it should gets the point across.

```c
// Example of something late-binding and isn't preserved over commits

struct wlr_presentation_feedback {
    struct wl_resource *resource;
    // ...

    struct wl_listener state_destroy;
};

static void feedback_state_destroy(struct wl_listener *l, void *data)
{
    struct wlr_presentation_feedback *fb = wl_container_of(l, fb, state_destroy);

    // Feedback still exists, so the wlroots user didn't use this buffer
    if (fb->resource) {
        wp_presentation_feedback_send_discarded(fb->resource);
        wl_resource_destroy(fb->resource);
    }

    // Other cleanup
    free(fb);
}

static void presentation_handle_feedback(struct wl_client *client,
        struct wl_resource *presentation_res, struct wl_resource *surface_res,
        uint32_t id) {
    struct wlr_presentation *presentation = wl_resource_get_user_data(presentation_res);
    struct wlr_surface *surface = wl_resource_get_user_data(surface_res);
    struct wlr_surface_state *st = wlr_surface_get_pending(surface);

    if (wlr_surface_state_get_presentation_feedback(st)) {
        // error: already exists
    }

    struct wlr_presentation_feedback *fb = ...;
    fb->state_destroy.notify = feedback_state_destroy;
    wl_signal_add(&st->events.destroy, &fb->state_destroy);

    wlr_surface_state_add(st, presentation->id, fb);
}

struct wlr_presentation *wlr_presentation_create(struct wlr_compositor *comp) {
    // ...
    present->id = wlr_compositor_register(comp);
}

struct wlr_presentation_feedback *wlr_surface_state_get_presentation_feedback(
        struct wlr_presentation *presentation, struct wlr_surface_state *st) {
    assert(presentation && st);
    return wlr_surface_state_get(st, presentation->id);
}
```
```c
// Example of something early-binding and preserved over commits

struct wlr_viewport {
    struct wl_resource *resource;
    // ...

    struct wl_listener state_commit;
    struct wl_listener state_destroy;
};

static void viewport_handle_set_source(...) {
    struct wlr_surface_state *st = ...;
    struct wlr_viewport *vp = wlr_surface_state_get_viewport(viewporter, st);

    vp->source = ...;
}

static void viewport_state_commit(struct wl_listener *l, void *data) {
    struct wlr_viewport *vp = wl_container_of(l, vp, state_commit);
    struct wlr_surface_state *st = data;

    if (/* viewport source larger than buffer */) {
        // protocol error
    }
}

static void viewport_state_destroy(struct wl_listener *l, void *data) {
    struct wlr_viewport *vp = wl_container_of(l, vp, state_destroy);
    // Other cleanup
    free(vp);
}

static void new_state(struct wl_listener *l, void *data) {
    struct wlr_viewporter *viewporter = wl_container_of(l, viewporter, new_state);
    struct wlr_compositor_new_state_args *args = data;

    struct wlr_viewport *vp = ...;
    if (args->old) {
        struct wlr_viewport *old = wlr_surface_state_get_viewport(args->old);
        // Copy values over

        // So new calls affect the new viewport
        wl_resource_set_user_data(vp->resource, vp);
    } else {
        // Set defaults
    }

    vp->state_destroy.notify = viewport_state_commit;
    wl_signal_add(&args->new->events.commit);

    vp->state_destroy.notify = viewport_state_destroy;
    wl_signal_add(&args->new->events.destroy);

    wl_surface_state_add(args->new, viewporter->id, vp);
}

struct wlr_viewporter *wlr_viewporter_create(struct wlr_compositor *comp) {
    // ...
    viewporter->id = wlr_compositor_register(comp);
    viewporter->new_state.notify = new_state;
    wl_signal_add(&comp->events.new_state, &viewporter->new_state);
}

struct wlr_viewport *wlr_surface_state_get_viewport(
        struct wlr_viewporter *viewporter, struct wlr_surface_state *st) {
    assert(viewporter && st);
    return wlr_surface_state_get(st, viewporter->id);
}
```

## Other notes

- This doesn't have to specifically relate to a protocol. `wlr_renderer` could use this, and add `wlr_surface_state_get_texture`.
- Maybe we could rename it to something like `wlr_commit`. `wlr_surface_state` is kind of a long name.
- The existing usage of surface roles could integrate into this, but I haven't really thought much about it.
- There may be some extra functions I want to add some `wp_linux_explicit_synchronization` will work with this, but I haven't fully thought of the details. It would revolve around "locking" a state so that it won't be returned by `wlr_surface_get_state`.
- It may be possible to allow reusing allocations more, but it may not be worth the effort.

ascent12:
_Another entry in my series of massive PRs that I take too long to finish._

Groundwork needed for #894 
Closes #1546 

Due to the very large amount of API breakage this introduces, I'd like to give people a heads up and get any feedback on the new design.

To see a full explanation for why I'm doing this, see the issues linked above. But to sum it up quickly:  
Our current state-keeping method won't work for explicit-synchronization, since a surface being committed now doesn't really mean it's ready to be shown. Instead of just "pending" and "current", we have to keep a queue of commits, so that we always have a valid buffer handy.

The core part of this is `wlr_commit`. This represents a "coherent presentation" of the client's contents, and should contain all of the information to render a client correctly. Every bit of state synchronized by wl_surface.commit should be changed to use this.

Todo:
- [x] wlr_compositor
- [x] wlr_subcompositor
- [ ] wlr_xdg_shell and related (WIP)
- [ ] wlr_xdg_shell_v6 and related (There have been talks of removing this anyway)
- [ ] wlr_fullscreen_shell_v1
- [ ] wlr_layer_shell_v1
- [ ] xwm
- [ ] wlr_presentation_time
- [ ] wlr_linux_explicit_synchronization (Maybe a different PR)
- [ ] wlr_buffer
- [ ] wlr_renderer
- [ ] wlr_seat and related (They set wlr_surface roles)
- [ ] Other trivial uses of wlr_surface (e.g. wlr_output)
- [ ] tinywl
- [ ] rootston
- [ ] More comments and documentation

Extra notes:
- All wlr_surface functions are being merged into wlr_compositor. I haven't removed the old code yet, so I'm using a _2 suffix to just prevent some conflicts. That'll be removed eventually.
- wlr_region functions were merged into wlr_compositor.
- wlr_subcompositor was split out from wlr_compositor, technically making this an optional interface (but you'd want to use it)
- wlr_subcompositor is definitely the most hardest part of this, because it's simply a complicated specification. Changing the other protocols will be much easier by comparison.
- I don't want wlr_surface managing any real buffer operations
  - wl_shm should be added to the renderer (using wlr_shm from another of my PRs)
  - Some callback functions to get information for a wlr_buffer would be needed, as they're needed for protocol compliance checks.
- I haven't done any __real__ testing on a lot of the current code; more protocols need to be done before I can look at doing tinywl and rootston and giving it a real go. I've done some basic hack code to test the bare essentials, though.


emersion:
>Instead of just "pending" and "current", we have to keep a queue of commits, so that we always have a valid buffer handy.

Why is that? If we need a valid buffer, can't we keep just one? (And why keep the whole surface state instead of just buffers?)

>Our current state-keeping method won't work for explicit-synchronization, since a surface being committed now doesn't really mean it's ready to be shown.

explitcit-synchronization release events don't need to be sent when the buffer is really released. They need to be sent with `wl_buffer.release`.

>wlr_subcompositor was split out from wlr_compositor, technically making this an optional interface (but you'd want to use it)

+1

>I don't want wlr_surface managing any real buffer operations 

Also +1

ascent12:
>Why is that? If we need a valid buffer, can't we keep just one? (And why keep the whole surface state instead of just buffers?)

Some extensions affect rendering, such as wp_viewport (which I eventually want to add support for) and the colour management protocol that's being worked on. We also store any wlr_commits of synchronized subsurfaces. We can't correctly render saved buffers unless we save this information too. 

Others bits of information perhaps aren't as important to stash with the commit (e.g. wp_presentation_feedback), but it doesn't really hurt to have them there, and may make them easier to correctly manage for users.

The queue needs to exist for the situation that a client has multiple commits with slow to signal fences.

>explicit-synchronization release events don't need to be sent when the buffer is really released. They need to be sent with wl_buffer.release.

Yeah, the release_fence can be sent earlier, and I was thinking of adding something like
```c
void wlr_commit_send_release_fence(struct wlr_commit *commit, int fence_fd);
```
which sends the fence immediately to all of the `zwp_linux_buffer_release_v1` objects associated with this commit. Calling it twice would be an error or maybe just a no-op.
And if this function has not been called by the time the commit is destroyed, it will send "immediate_release" automatically.

A similar design could be done with things like `wp_presentation_feedback.{presented,discarded}`.

emersion:
>Some extensions affect rendering, such as wp_viewport (which I eventually want to add support for) and the colour management protocol that's being worked on. We also store any wlr_commits of synchronized subsurfaces. We can't correctly render saved buffers unless we save this information too.
>
>Others bits of information perhaps aren't as important to stash with the commit (e.g. wp_presentation_feedback), but it doesn't really hurt to have them there, and may make them easier to correctly manage for users.
>
>The queue needs to exist for the situation that a client has multiple commits with slow to signal fences.

Okay, this last bit is important. If I understand correctly:

- With explicit-synchronization, we may want for a buffer to be ready before attempting to render/display it (ie. wait on its in-fence FD).
- Because of this, we need a queue of states. All states in the queue (except maybe the front one) are being waited on. Once a state becomes ready, we discard older states.
- We need to be able to use and render any committed-but-not-released state. We need the whole state to be able to render (buffer + metadata).

>A similar design could be done with things like wp_presentation_feedback.{presented,discarded}.

Note to self: we need to keep the state around for longer because of this.

Some other notes:

- I think this PR does too many things at once. Things like moving stuff in different files, removing the resource list can be done separately and are not really related to the redesign.
- I don't like the name `wlr_commit`, especially now that we have `wlr_output_commit`. I'd prefer to keep `wlr_surface_state`.

ascent12:
Just some notes from when I was updating `wlr_seat`-related things. I won't be fixing these as part of this PR to try and keep some of the API breakages down, and will probably make a new issue for it.

- We aren't handling the x/y arguments gives to `wl_surface.attach` properly in a lot of cases. For example, with `wl_pointer.set_cursor`, those arguments are supposed to update the hotspot.
- `wl_pointer.set_cursor` and `zwp_tablet_tool_v1.set_cursor` just punts the hotspot handling off to the user. We could create extensions to `wlr_commit` to handle these properly.
- `wlr_drag_icon` seems largely unnecessary. It's just a container for unmap/map events on a surface, but we can just listen to events on the surface directly. This could be removed, unless it's kept around for managing `wl_surface.attach`'s x/y.

emersion:
>We aren't handling the x/y arguments gives to wl_surface.attach properly in a lot of cases. For example, with wl_pointer.set_cursor, those arguments are supposed to update the hotspot.

We do update the hotspot, this is handled in `wlr_output`. It wouldn't make sense to do this in `wlr_surface` since this is a cursor-specific behaviour.

>wl_pointer.set_cursor and zwp_tablet_tool_v1.set_cursor just punts the hotspot handling off to the user. We could create extensions to wlr_commit to handle these properly.

I'm not sure this is a good idea. Do you have an API proposal?

>wlr_drag_icon seems largely unnecessary

We'd need to add map/unmap events to wlr_surface. Not sure it's a good idea either, but it could be.

ascent12:
>It wouldn't make sense to do this in wlr_surface since this is a cursor-specific behaviour.

It'd all be "extensions" to `wlr_commit` done by whatever the surface role is. There is no reason we couldn't add anything role-specific to it, and it was my intention to do that for things like xdg-shell anyway.

>I'm not sure this is a good idea. Do you have an API proposal?

Something along the lines of
```c
// false if not pointer role
bool wlr_commit_get_pointer_hotspot(struct wlr_commit *commit, int32_t *x, int32_t *y);
```
The hotspot is something tied to the "correct" presentation of a surface, so it makes sense for the `wlr_commit` to handle it.

>We'd need to add map/unmap events to wlr_surface

No, you'd just listen for a commit, and check whether there is a buffer or not.

emersion:
>wlr_commit_get_pointer_hotspot

Not a fan of this TBH.

>No, you'd just listen for a commit, and check whether there is a buffer or not.

We need to do a bunch of stuff on map/unmap. For instance damaging the whole surface.

ascent12:
>Not a fan of this TBH.

I'm not going to get too much into this now, since it's not something I'm going to do in this PR, but I don't see any advantage to NOT putting in with the surface/commit.

>We do update the hotspot, this is handled in wlr_output.

Our cursor-related APIs need some serious thought put into them (#1363), and I consider `wlr_output_cursor_set_surface` to be broken, as it's completely incapable of handling subsurfaces or anything else that is going to have an effect on rendering. That work needs to be handled by the user, and they're going to know the role of the surface.

>We need to do a bunch of stuff on map/unmap. For instance damaging the whole surface.

There is nothing stopping you from doing that. Just looking at the rootston code specifically: https://github.com/swaywm/wlroots/blob/master/rootston/seat.c#L570-L589
You've got 3 handlers for that, but you can do with just the commit handler and move the `roots_drag_icon_damage_whole(icon);` into there, but since `roots_drag_icon_update_position(icon);` does that, even that wouldn't be needed.

emersion:
>Our cursor-related APIs need some serious thought put into them (#1363), and I consider wlr_output_cursor_set_surface to be broken, 

Agreed.

> as it's completely incapable of handling subsurfaces or anything else that is going to have an effect on rendering.

I don't worry a lot about subsurfaces. But yeah, it does have design issues.

> That work needs to be handled by the user, and they're going to know the role of the surface.

I think there should be a thin `wlr_surface` wrapper for cursors, with all the cursor-specific API.

>There is nothing stopping you from doing that.

But then you need to check whether the last commit was NULL and this one wasn't, and the other way around, and make sure to damage on destroy, etc.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1685

PR #1687, opened 2019-05-03
Enable RDP in .builds/freebsd.yml

jbeich:
FreeBSD is currently the only non-Linux platform covered by automation. Let's keep RDP code portable as well.

See [downstream build log](https://github.com/swaywm/wlroots/files/3142831/wlroots-0.6.0.log).


emersion:
Thanks!

PR #1691, opened 2019-05-06
backend/wayland: use xdg-decoration-unstable-v1

emersion:
This allows the toplevel to have proper decorations on compositors that support
xdg-decoration-unstable-v1.


ddevault:
Thanks!

PR #1694, opened 2019-05-10
Tablet v2 client

Ongy:
This adds tablet support to the wlroots wayland backend.

I've only tested this against rootston itself as parent compositor and some of the math might be off, converting from [0,65536] to [0,1] and back.

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -21,10 +21,6 @@
 #include "xdg-shell-client-protocol.h"
 #include "tablet-unstable-v2-client-protocol.h"
 
-struct wlr_wl_tablet_seat *wlr_wl_add_tablet_seat(
-		struct zwp_tablet_manager_v2 *manager,
-		struct wl_seat *seat, struct wlr_wl_backend *backend);
-
 struct wlr_wl_backend *get_wl_backend_from_backend(struct wlr_backend *backend) {
 	assert(wlr_backend_is_wl(backend));
 	return (struct wlr_wl_backend *)backend;
@@ -118,7 +114,7 @@ static bool backend_start(struct wlr_backend *backend) {
 	}
 
 	if (wl->tablet_manager && wl->seat) {
-		wlr_wl_add_tablet_seat(wl->tablet_manager,
+		wl_add_tablet_seat(wl->tablet_manager,
 			wl->seat, wl);
 	}
 
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -19,6 +19,11 @@
 #include "xdg-decoration-unstable-v1-client-protocol.h"
 #include "pointer-gestures-unstable-v1-client-protocol.h"
 #include "xdg-shell-client-protocol.h"
+#include "tablet-unstable-v2-client-protocol.h"
+
+struct wlr_wl_tablet_seat *wlr_wl_add_tablet_seat(
+		struct zwp_tablet_manager_v2 *manager,
+		struct wl_seat *seat, struct wlr_wl_backend *backend);
 
 struct wlr_wl_backend *get_wl_backend_from_backend(struct wlr_backend *backend) {
 	assert(wlr_backend_is_wl(backend));
@@ -81,6 +86,9 @@ static void registry_global(void *data, struct wl_registry *registry,
 	} else if (strcmp(iface, zwp_pointer_gestures_v1_interface.name) == 0) {
 		wl->zwp_pointer_gestures_v1 = wl_registry_bind(registry, name,
 			&zwp_pointer_gestures_v1_interface, 1);
+	} else if (strcmp(iface, zwp_tablet_manager_v2_interface.name) == 0) {
+		wl->tablet_manager = wl_registry_bind(registry, name,
+			&zwp_tablet_manager_v2_interface, 1);
 	}
 }
 
@@ -107,6 +115,11 @@ static bool backend_start(struct wlr_backend *backend) {
 
 	if (wl->keyboard) {
 		create_wl_keyboard(wl->keyboard, wl);
+	}
+
+	if (wl->tablet_manager && wl->seat) {
+		wlr_wl_add_tablet_seat(wl->tablet_manager,
+			wl->seat, wl);
 	}
 
 	for (size_t i = 0; i < wl->requested_outputs; ++i) {
--- a/backend/wayland/tablet_v2.c
+++ b/backend/wayland/tablet_v2.c
@@ -1,0 +1,941 @@
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+#endif
+
+#include <wayland-util.h>
+#include <assert.h>
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wlr/interfaces/wlr_tablet_pad.h>
+#include <wlr/interfaces/wlr_tablet_tool.h>
+#include <wlr/types/wlr_input_device.h>
+#include <wlr/interfaces/wlr_input_device.h>
+
+#include "util/signal.h"
+#include "wlr/util/log.h"
+#include "tablet-unstable-v2-client-protocol.h"
+
+#include <backend/wayland.h>
+
+struct wlr_wl_tablet_seat {
+	struct zwp_tablet_seat_v2 *tablet_seat;
+	struct wl_list tablets;
+	struct wl_list tools;
+//	struct wl_list pads; // wlr_wl_tablet_pad::link
+};
+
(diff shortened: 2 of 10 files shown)


backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+		free(ret);
+		return NULL;
+	}
+
+	zwp_tablet_seat_v2_add_listener(ret->tablet_seat,
+		&tablet_seat_listener, backend);
+
+	wl_list_init(&ret->tablets);
+	wl_list_init(&ret->tools);
+	//wl_list_init(&ret->pads);
```
emersion: Is this a TODO?

backend/wayland/backend.c
```diff
@@ -18,6 +18,11 @@
 #include "util/signal.h"
 #include "xdg-decoration-unstable-v1-client-protocol.h"
 #include "xdg-shell-client-protocol.h"
+#include "tablet-unstable-v2-client-protocol.h"
+
+struct wlr_wl_tablet_seat *wlr_wl_add_tablet_seat(
```
emersion: 1. This isn't a public function, so it shouldn't have the `wlr_` prefix. `add_wl_tablet_seat` would be a better name.
2. This should be declared in the private header.

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+#include <wlr/interfaces/wlr_tablet_pad.h>
+#include <wlr/interfaces/wlr_tablet_tool.h>
+#include <wlr/types/wlr_input_device.h>
+#include <wlr/interfaces/wlr_input_device.h>
+
+#include "util/signal.h"
+#include "wlr/util/log.h"
+#include "tablet-unstable-v2-client-protocol.h"
+
+#include <backend/wayland.h>
```
emersion: This is a private header, so it should be imported with double-quotes instead of angle-brackets.

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+
+struct wlr_wl_tablet_seat {
+	struct zwp_tablet_seat_v2 *tablet_seat;
+	struct wl_list tablets;
+	struct wl_list tools;
+//	struct wl_list pads; // wlr_wl_tablet_pad::link
```
emersion: Is this a TODO?

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+	struct wl_list strips; // wlr_wl_tablet_pad_strips::link
+};
+
+static uint32_t get_current_time_msec(void) {
+	struct timespec now;
+	clock_gettime(CLOCK_MONOTONIC, &now);
+	return now.tv_nsec / (1000 * 1000) + now.tv_sec * 1000;
+}
+
+struct wlr_wl_tablet_seat *wlr_wl_add_tablet_seat(
```
emersion: This should be moved to the private header

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+	.stop = handle_tablet_pad_strip_stop,
+	.frame = handle_tablet_pad_strip_frame,
+};
+
+static void handle_tablet_pad_group_buttons(void *data,
+		struct zwp_tablet_pad_group_v2 *pad_group,
+		struct wl_array *buttons) {
+	struct wlr_wl_tablet_pad_group *group = data;
+
+	group->group.buttons = calloc(1, buttons->size);
```
emersion: Where is this free'ed? Should we free the previous value before overriding it?

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+		struct zwp_tablet_seat_v2 *zwp_tablet_seat_v2,
+		struct zwp_tablet_pad_v2 *id) {
+	wlr_log(WLR_DEBUG, "New tablet pad");
+	struct wlr_wl_backend *backend = data;
+	struct wlr_wl_input_device *dev = create_wl_input_device(
+		backend, WLR_INPUT_DEVICE_TABLET_PAD);
+	if (!dev) {
+		/* This leaks a couple of server-sent resource ids. iirc this
+		 * shouldn't ever be a problem, but it isn't exactly nice
+		 * either. */
```
emersion: What do you mean? What is leaked?

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+	wlr_tablet_pad_init(wlr_dev->tablet_pad, NULL);
+	zwp_tablet_pad_v2_add_listener(id, &tablet_pad_listener, dev);
+}
+
+static void handle_tablet_tool_done(void *data,
+		struct zwp_tablet_tool_v2 *id) {
+	/* empty */
+}
+static enum wlr_tablet_tool_type tablet_type_to_wlr_type(enum zwp_tablet_tool_v2_type type) {
+
```
emersion: Style: extra newline here

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+		return WLR_TABLET_TOOL_TYPE_BRUSH;
+	case ZWP_TABLET_TOOL_V2_TYPE_PENCIL:
+		return WLR_TABLET_TOOL_TYPE_PENCIL;
+	case ZWP_TABLET_TOOL_V2_TYPE_AIRBRUSH:
+		return WLR_TABLET_TOOL_TYPE_AIRBRUSH;
+	case ZWP_TABLET_TOOL_V2_TYPE_MOUSE:
+		return WLR_TABLET_TOOL_TYPE_MOUSE;
+	case ZWP_TABLET_TOOL_V2_TYPE_LENS:
+		return WLR_TABLET_TOOL_TYPE_LENS;
+	default:
```
emersion: Possible to remove the default case?

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+}
+
+static void handle_tablet_tool_capability(void *data,
+		struct zwp_tablet_tool_v2 *id,
+		uint32_t capability) {
+	struct wlr_wl_tablet_tool *tool = data;
+
+	enum zwp_tablet_tool_v2_capability cap = capability;
+	
+	switch(cap) {
```
emersion: Style: missing space after `switch` keyword

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+	clear_tablet_tool_values(tool);
+	wl_signal_init(&tool->wlr_tool.events.destroy);
+	zwp_tablet_tool_v2_add_listener(id, &tablet_tool_listener, tool);
+}
+
+static void handle_tablet_name(void *data, struct zwp_tablet_v2 *zwp_tablet_v2,
+		const char *name) {
+	struct wlr_wl_input_device *dev = data;
+	struct wlr_tablet *tablet = dev->wlr_input_device.tablet;
+	tablet->name = strdup(name);
```
emersion: Is the "name" event guaranteed to be sent at most once? Where is this free'ed?

backend/wayland/tablet_v2.c
```diff
@@ -0,0 +1,941 @@
...
+};
+
+static void handle_tab_added(void *data,
+		struct zwp_tablet_seat_v2 *zwp_tablet_seat_v2,
+		struct zwp_tablet_v2 *id) {
+	wlr_log(WLR_DEBUG, "New tablet");
+	struct wlr_wl_backend *backend = data;
+	struct wlr_wl_input_device *dev = create_wl_input_device(
+		backend, WLR_INPUT_DEVICE_TABLET_TOOL);
+	dev->resource = id;
```
emersion: This should be done after the `if (!dev)` check

Ongy:
Ok, either I'm going crazy, or github is messing up the updated review comments.

The comments about name (and I think something in group) free-ing should probably be addressed via a separate PR, those should be in the general tablet code.

The comment about server-sent IDs is there because we just remove an object that will later on receive more ids from the server (groups->rings/strips iirc).
From what I know (might be wrong/out of date) this will "leak" those ids, because they are never handled properly.
This shouldn't be more of a problem than we already have, if the allocation fails, but it's not optimal either.

DarkDefender:
I've tried this out and it seems to work well in rootston.
However, the tablet doesn't seem to show up in sway under xwayland.
I does in rootston. I've recompiled both wlroots and sway but I'm guessing that there are some bits missing in sway? I didn't see anything obviously rootston specific in this patch so I don't know what is missing.

emersion:
Sway is missing tablet support.

PR #1695, opened 2019-05-13
output: clear output->damage on successful commit

ammen99:
Otherwise the damage is accumulated on subsequent frames.


emersion:
Thanks, nice catch!

PR #1704, opened 2019-05-22
session/logind: specify seat exactly

rindeal:
"/org/freedesktop/login1/seat/self" path triggers seat-finding code path in logind,
which currently relies on getting the session based on caller's PID.
This behaviour is deprecated in logind as it doesn't work eg. with systemd user units,
which run outside of user session.

We check for "seat0" in logind_change_vt() already as introduced in 47985d2dc56a6af469ac9375e7548136765aff16,
so hard-coding it here is not a problem, otherwise sd_session_get_seat() could be used.


ascent12:
Thanks!

PR #1705, opened 2019-05-22
session/logind: check for XDG_SESSION_ID first

rindeal:
In order to support compositors running as systemd user units without display manager,
a mechanism for specifying session ID exactly must exist.

Checking for `XDG_SESSION_ID` mimics loginctl behaviour https://github.com/systemd/systemd/blob/e95be7def26c6c5feaf08a4135aa4f50c53263a8/src/login/loginctl.c#L856.


ascent12:
Thanks!

PR #1713, opened 2019-06-02
drm: legacy: issue a NULL modeset on disable

### LINKED ISSUE #1706: Do a NULL modeset on connector unplug
I've learned today that if userspace doesn't do a NULL modeset when a connector changes its state from connected to disconnected, the pipe on the kernel side will remain active and will continue to send pixels (to /dev/null, wasting energy). This is apparently a "feature" for userspace that doesn't support hotplug events (so that after unplugging-replugging the screen receives again an image).

tl;dr we need to do a NULL modeset.

Emantor:
The DRM subsystem needs a NULL modeset for connectors which disappear
from the system to disable the hardware pipes, otherwise the pixels get
rendered but are sent nowhere.

The atomic backend does the equivalent by removing the properties and
issuing a commit.

Fixes #1706

Tested by using `WLR_DRM_NO_ATOMIC=1 sway` and toggling my laptops internal eDP-1 display.


ascent12:
Thanks

Issue #1716, opened 2019-06-07
button_count is corrupted when no pointer release is encountered


norcalli:
I am writing a program and I had it use right click as a means of exiting the event loop. I was trying to validate my serial to initiate a drag and drop and noticed that the serial in sway debug logs weren't being validated, but the button_count also seemed incorrect. It also kept incrementing.

I realized that by using right click to exit the program, `button_count` wasn't being decremented.

Tracking increments and decrements seems like it is a difficult balance act and could easily get corrupted if we encounter a state which is unaccounted for, so it seems like there needs to be an appropriate point at which we should *reset* `button_count` so as to at least have a catch-all in case we screw up.

In this case, if the surface originating a pointer event is destroyed, then it makes sense to reset the state.

norcalli:
potentially related to  swaywm/sway#4168

emersion:
Related: https://github.com/swaywm/wlroots/issues/1593

Issue #1593, opened 2019-03-02
Button state and pointer focus change


emersion:
It's not clear what wlroots should do when the compositor decides to change the pointer focus and there are some pressed buttons.

Clients receiving `leave` will assume buttons have been released. Clients receiving `enter` won't be aware that some buttons are being pressed. Additionally, the wlroots internal state will still assume some buttons are being pressed.

While `wl_keyboard.enter` contains an array of currently pressed keys, this isn't the case for `wl_pointer.enter`.

Related issues:

1. Dragging a window with its titlebar: `pointer.button(pressed)` gets sent to the client. Client sends a `xdg_toplevel.move` request. This starts a grab in the compositor. `pointer.leave` is sent to the client, the pointer focus is unset. Later on the grab ends. At this point the compositor _must_ send `pointer.button(released)` so that wlroots state is updated, even if the event isn't sent to any client (NULL pointer focus).
2. Pointer bindings: `pointer.button(pressed)` gets sent to the client. Later on the compositor decides this is a pointer binding so it sends `pointer.leave`. At this point the compositor _must_ send `pointer.button(released)` so that wlroots state is updated, even if the event isn't sent to any client.

Possible solutions:

1. Implicitly release all buttons when pointer focus changes.
2. Implicitly send pressed events to the new surface for all currently pressed buttons when pointer focus changes.

In any case, this requires to track which buttons are pressed in wlroots. We already do this for keyboard keys in `wlr_keyboard` but not in `wlr_seat_keyboard`. Not sure tracking buttons in `wlr_seat_pointer` is a good idea. Or should we track keys in `wlr_seat_keyboard` instead of `wlr_keyboard`?

ammen99:
> * Pointer bindings: `pointer.button(pressed)` gets sent to the client. Later on the compositor decides this is a pointer binding so it sends `pointer.leave`. At this point the compositor _must_ send `pointer.button(released)` so that wlroots state is updated, even if the event isn't sent to any client.

This isn't the only way. In Wayfire, I never send the pressed event, and the pointer focus never changes, as this is a one-time action which doesn't change anything.

PR #1717, opened 2019-06-08
Use a set to track pointer button state.

norcalli:
In addition to `button_count`, we keep track of the current buttons
pressed just as in `wlr_keyboard`.

Add `set_add` and `set_remove` to assist with this. These functions can
only be used with values greater than 0 (such as the button/key masks
for keyboards and pointers).

Partially addresses:
- https://github.com/swaywm/wlroots/issues/1716
- https://github.com/swaywm/wlroots/issues/1593

This doesn't make it so that button_count is incorruptible, and so some of the solutions described in #1593 might still be necessary (or another alternative), but this makes it possible to recover from errors upon triggering pointer events on another surface.

### DIFF
--- a/include/util/array.h
+++ b/include/util/array.h
@@ -3,7 +3,22 @@
 
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdbool.h>
 
 size_t push_zeroes_to_end(uint32_t arr[], size_t n);
+
+/**
+ * Add `target` to `values` if it doesn't exist
+ * "set"s should only be modified with set_* functions
+ * Values MUST be greater than 0
+ */
+bool set_add(uint32_t values[], size_t *len, size_t cap, uint32_t target);
+
+/**
+ * Remove `target` from `values` if it exists
+ * "set"s should only be modified with set_* functions
+ * Values MUST be greater than 0
+ */
+bool set_remove(uint32_t values[], size_t *len, size_t cap, uint32_t target);
 
 #endif
--- a/types/seat/wlr_seat_pointer.c
+++ b/types/seat/wlr_seat_pointer.c
@@ -8,6 +8,7 @@
 #include <wlr/util/log.h>
 #include "types/wlr_seat.h"
 #include "util/signal.h"
+#include "util/array.h"
 
 static void default_pointer_enter(struct wlr_seat_pointer_grab *grab,
 		struct wlr_surface *surface, double sx, double sy) {
@@ -339,26 +340,32 @@ void wlr_seat_pointer_notify_motion(struct wlr_seat *wlr_seat, uint32_t time,
 uint32_t wlr_seat_pointer_notify_button(struct wlr_seat *wlr_seat,
 		uint32_t time, uint32_t button, enum wlr_button_state state) {
 	clock_gettime(CLOCK_MONOTONIC, &wlr_seat->last_event);
+
+	struct wlr_seat_pointer_state* pointer_state = &wlr_seat->pointer_state;
+
 	if (state == WLR_BUTTON_PRESSED) {
-		if (wlr_seat->pointer_state.button_count == 0) {
-			wlr_seat->pointer_state.grab_button = button;
-			wlr_seat->pointer_state.grab_time = time;
+		if (pointer_state->button_count == 0) {
+			pointer_state->grab_button = button;
+			pointer_state->grab_time = time;
 		}
-		wlr_seat->pointer_state.button_count++;
+		set_add(pointer_state->buttons, &pointer_state->button_count,
+			WLR_POINTER_BUTTONS_CAP, button);
 	} else {
-		if (wlr_seat->pointer_state.button_count == 0) {
-			wlr_log(WLR_ERROR, "Corrupted seat button count");
-		} else {
-			wlr_seat->pointer_state.button_count--;
-		}
+		set_remove(pointer_state->buttons, &pointer_state->button_count,
+			WLR_POINTER_BUTTONS_CAP, button);
 	}
 
-	struct wlr_seat_pointer_grab *grab = wlr_seat->pointer_state.grab;
+
+	struct wlr_seat_pointer_grab *grab = pointer_state->grab;
 	uint32_t serial = grab->interface->button(grab, time, button, state);
 
-	if (serial && wlr_seat->pointer_state.button_count == 1 &&
+	wlr_log(WLR_DEBUG, "button_count=%zu grab_serial=%"PRIu32" serial=%"PRIu32"",
+		pointer_state->button_count,
+		pointer_state->grab_serial, serial);
+
+	if (serial && pointer_state->button_count == 1 &&
 			state == WLR_BUTTON_PRESSED) {
-		wlr_seat->pointer_state.grab_serial = serial;
+		pointer_state->grab_serial = serial;
 	}
 
 	return serial;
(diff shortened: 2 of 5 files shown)


norcalli:
Oh I forgot to describe how set_add and set_remove work. They maintain the invariant that empty values can only be at the end of the array and 0 represents an empty value.

set_add: only appends to the end
set_remove: if an element to remove is found, swap it with the value at the end and set the end position to 0.

Ongy:
Hey, there's similar (though slightly more complex) here already https://github.com/swaywm/wlroots/blob/master/types/tablet_v2/wlr_tablet_v2_tool.c#L260

The main difference (on a high level) is that it keeps 2 arrays in sync, one for the button itself, the other for the serial sent with the button.
Not sure if it's worth trying to keep this in a single function, though it might be interesting to have the same serial tracking in wlr_pointer.
Thoughts?

include/wlr/types/wlr_keyboard.h
```diff
@@ -58,7 +58,7 @@ struct wlr_keyboard {
 	xkb_mod_index_t mod_indexes[WLR_MODIFIER_COUNT];
 
 	uint32_t keycodes[WLR_KEYBOARD_KEYS_CAP];
-	size_t num_keycodes;
+	uint32_t num_keycodes;
```
emersion: This should stay `size_t`

include/util/array.h
```diff
@@ -3,7 +3,10 @@
 
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdbool.h>
 
 size_t push_zeroes_to_end(uint32_t arr[], size_t n);
+bool set_add(uint32_t values[], uint32_t* len, uint32_t cap, uint32_t target);
```
emersion: `len` and `cap` should be `size_t`

Style: `*` on the right

types/seat/wlr_seat_pointer.c
```diff
@@ -339,26 +340,33 @@ void wlr_seat_pointer_notify_motion(struct wlr_seat *wlr_seat, uint32_t time,
...
-			wlr_seat->pointer_state.grab_button = button;
-			wlr_seat->pointer_state.grab_time = time;
+		if (pointer_state->button_count == 0) {
+			pointer_state->grab_button = button;
+			pointer_state->grab_time = time;
 		}
-		wlr_seat->pointer_state.button_count++;
+		set_add(pointer_state->buttons, &pointer_state->button_count,
+			WLR_POINTER_BUTTONS_CAP, button);
+	/* } else if (state == WLR_BUTTON_RELEASED) { */
```
emersion: Remnant from a debugging session?

types/seat/wlr_seat_pointer.c
```diff
@@ -339,26 +340,33 @@ void wlr_seat_pointer_notify_motion(struct wlr_seat *wlr_seat, uint32_t time,
...
+			WLR_POINTER_BUTTONS_CAP, button);
 	}
 
-	struct wlr_seat_pointer_grab *grab = wlr_seat->pointer_state.grab;
+
+	struct wlr_seat_pointer_grab *grab = pointer_state->grab;
 	uint32_t serial = grab->interface->button(grab, time, button, state);
 
-	if (serial && wlr_seat->pointer_state.button_count == 1 &&
+	wlr_log(WLR_DEBUG, "button_count=%"PRIu32" grab_serial=%"PRIu32" serial=%"PRIu32"",
```
emersion: I'm not sure we want to keep this one, it'll be pretty verbose

util/array.c
```diff
@@ -19,3 +19,34 @@ size_t push_zeroes_to_end(uint32_t arr[], size_t n) {
...
+	for (uint32_t i = 0, n = *len; i < n; ++i) {
+		if (values[i] == target) {
+			return false;
+		}
+	}
+	values[(*len)++] = target;
+	return false;
+}
+
+// Values MUST be greater than 0
```
emersion: This should be documented in the header file, in a doc comment

util/array.c
```diff
@@ -19,3 +19,34 @@ size_t push_zeroes_to_end(uint32_t arr[], size_t n) {
...
+bool set_remove(uint32_t values[], uint32_t* len, uint32_t cap, uint32_t target) {
+	if (*len == 0) {
```
emersion: We can drop this check, since the loop will be a no-op in this case

util/array.c
```diff
@@ -19,3 +19,34 @@ size_t push_zeroes_to_end(uint32_t arr[], size_t n) {
...
+		return false;
+	}
+	for (uint32_t i = 0, n = *len; i < n; ++i) {
```
emersion: Why do we need `n`?

util/array.c
```diff
@@ -19,3 +19,34 @@ size_t push_zeroes_to_end(uint32_t arr[], size_t n) {
...
+		if (values[i] == target) {
+			uint8_t last_elem_pos = --(*len);
```
emersion: This should be `size_t`

norcalli: oh whoops. remnant from the small set version.

types/seat/wlr_seat_pointer.c
norcalli: Yeah, I also couldn't decide which version was more correct, not that it should matter

types/seat/wlr_seat_pointer.c
norcalli: Yeah, maybe. The average person clicks around 5000 times a day. However, I found it fairly useful to be able to have a user action I could use to track. For me, the IPC debug logs greatly overpowered these logs (as it stands, I have to rotate my debug logs constantly or they grow to > 1.5GB a day).

The main reason I chose to leave them in was due to this being a somewhat open problem still as far as corruption goes, and I thought it could be useful in debugging future issues.

include/wlr/types/wlr_keyboard.h
norcalli: Then I should make `button_count` `size_t` to mirror that.

norcalli:
@Ongy it feels like tracking multiple serials is only useful for tablets since they have multitouch where each finger acts like a separate pointer with only 1 button (finger pressed or not pressed). So I'm not sure how it would be helpful here (unless I'm misunderstanding the code).

norcalli:
@emersion I've addressed the comments. If it looks ok, I'll squashpush

norcalli:
@emersion I've addressed the comments. If it looks ok, I'll squashpush

Ongy:
The idea is to track every serial sent by wlroots in some automated fashion
I had to do this for the feature of entering a surface with pressed buttons (which is well defined for tablet)
This might be relevant for the pointer, there's a discussion in another issue about how to handle these things iirc

arandomhuman:
Suggestion: instead of a bool return type in `set_add` and `set_remove`, return the index at which the target was found/added. If -1, illegal arguments (*len >= cap, target == 0). If (index + 1) != len in `set_add`, the target was not added (and an error could be printed?). Similarly, if index == len in `set_remove`, the target was not found and removed.
Additionally, allow optionally specifying an index to `set_remove` which, along with the above change, allows keeping two sets in sync (e.g. buttons and serials, as @Ongy said).

emersion:
>Suggestion: instead of a bool return type in set_add and set_remove, return the index at which the target was found/added. If -1, illegal arguments (*len >= cap, target == 0). If (index + 1) != len in set_add, the target was not added (and an error could be printed?). Similarly, if index == len in set_remove, the target was not found and removed.
Additionally, allow optionally specifying an index to set_remove which, along with the above change, allows keeping two sets in sync (e.g. buttons and serials, as @Ongy said).

This can be done when a user of that function actually needs it.

emersion:
Nice! Can you squash everything?

PR #1719, opened 2019-06-11
Remove WLR_DRM_NO_ATOMIC_GAMMA workaround

ascent12:
This is [fixed on amdgpu](https://github.com/swaywm/sway/issues/2700#issuecomment-500236109), so we don't need this any more.


emersion:
Thanks!

PR #1720, opened 2019-06-11
[WIP] foreign-toplevel: support geometry event

adlocode:
Add an event for when a window's position or dimensions change.


emersion:
This needs to go through wlr-protocols review process first.

PR #1724, opened 2019-06-15
Implement input-timestamps-unstable-v1

### LINKED ISSUE #1514: Add support for the input-timestamps protocol
wlroots events currently have a `uint32_t time_msec` field. It will be necessary to either change the field (`struct timespec`?) or add a new one (`uint32_t time_nsec`?).

arandomhuman:
This PR implements the input-timestamps-unstable-v1 protocol. Brief description of commits in chronological order:

`- gitignore: add vim .swp files`
`- seat: expose more from_resource() functions`

Small miscellaneous commits for convenience.

`- input_timestamps: initial build support`
`- input_timestamps: basic framework for header and source`
`- rootston: add input_timestamps manager`
`- input_timestamps: complete manager implementation`
`- input_timestamps: add public function to send timestamps`

Complete implementation of the server side input-timestamps protocol,
subdivided in a (hopefully) clean manner. The protocol specific code is only in
the last two commits: the former being manager logic, the latter generating the
client event.

`- keyboard, pointer, touch: add nsec field to relevant events`
`- rootston: send input timestamps`

Add a time_frac_nsec field to all keyboard, (wl_)pointer, and touch events. The
field name has a 'frac' to distinguish it from 'time_msec' which has the
complete time in msec.
_Note:_ this field hasn't been added to other events, like pointer gestures,
tablet tool, switches, and so on; even though, for example, tablets are treated
the same as pointers in rootston (afaics).

`- backend/libinput: set the nsec field of relevant events`
`- backend/wayland: use input-timestamps client protocol to get nsec values`
`- backend/x11, backend/rdp, virtual_keyboard: set the nsec field of relevant events`

Add support for setting the time_frac_nsec field in the backends generating
events. Libinput gives usec precision, while x11 and rdp (and the virtual
keyboard protocol) only give msec precision. If possible, we use the
input-timestamps-client protocol to get more precise events on the wayland
backend, otherwise we get timestamps in msecs.

`- examples: add input-timestamp-v1 example`

Add a small example that just prints keyboard and pointer events along with
more precise timestamps.

Closes #1514.


ascent12:
>- gitignore: add vim .swp files

We don't do editor-specific gitignores. Add it to your personal gitignore instead.

arandomhuman:
Alright. I'll remove that commit after dealing with code reviews.

include/wlr/types/wlr_input_timestamps_v1.h
```diff
@@ -9,4 +9,81 @@
 #ifndef WLR_TYPES_WLR_INPUT_TIMESTAMPS_V1_H
 #define WLR_TYPES_WLR_INPUT_TIMESTAMPS_V1_H
 
+#include <wayland-server.h>
+
+/**
+ * This protocol specifies a way for a client to request and receive
+ * high-resolution timestamps for input events.
+ */
```
ddevault: I would put this comment at the top of the file

include/wlr/types/wlr_input_timestamps_v1.h
```diff
@@ -9,4 +9,81 @@
...
+
+
+/**
+ * A global interface used for requesting high-resolution timestamps for
+ * input events.
+ */
+
+struct wlr_input_timestamps_manager_v1 {
```
ddevault: nit: no newline between doc comment and thing it documents

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,219 @@
...
+
+#define INPUT_TIMESTAMPS_MANAGER_VERSION 1
+
+static const struct zwp_input_timestamps_manager_v1_interface manager_impl;
+static const struct zwp_input_timestamps_v1_interface impl;
+
+
+/**
+ * helper functions
+ */
```
ddevault: same feedback here

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,219 @@
...
+#include <wlr/util/log.h>
+#include "input-timestamps-unstable-v1-protocol.h"
+
+#define INPUT_TIMESTAMPS_MANAGER_VERSION 1
+
+static const struct zwp_input_timestamps_manager_v1_interface manager_impl;
+static const struct zwp_input_timestamps_v1_interface impl;
+
+
+/**
```
ddevault: And only one newline between globals and shit

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,219 @@
...
+#include <stdlib.h>
+
+#include "wayland-server.h"
+#include "wayland-util.h"
+
+#include <util/signal.h>
+#include <wlr/types/wlr_input_timestamps_v1.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/util/log.h>
+#include "input-timestamps-unstable-v1-protocol.h"
```
ddevault: All external includes with `<>`, then all internal includes with `""`, alpha-sorted within their respective groups

types/wlr_input_timestamps_v1.c
ddevault: actually, this and the following comment should just be removed

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,219 @@
...
+
+static void manager_handle_resource_destroy(struct wl_resource *resource) {
+	wl_list_remove(wl_resource_get_link(resource));
+}
+
+static void manager_handle_destroy(struct wl_client *client, struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+
+	wlr_log(WLR_DEBUG, "input_timestamps_v1 manager unbound from client %p",
+		client);
```
ddevault:
```c
wlr_log(WLR_DEBUG,
    "input_timestamps_v1 manager unbound from client %p", client);
```

has better balance

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,219 @@
...
+}
+
+static void manager_handle_get_keyboard_timestamps(struct wl_client *client,
+		struct wl_resource *resource, uint32_t id, struct wl_resource *keyboard) {
```
ddevault: micro-nit: these ought to have a `/* TODO */`

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,219 @@
...
+};
+
+static const struct zwp_input_timestamps_v1_interface impl = {
+	.destroy = handle_destroy,
+};
+
+
+/**
+ * Public functions
+ */
```
ddevault: Yeah all of these comments can go

types/wlr_input_timestamps_v1.c
```diff
@@ -313,3 +313,79 @@ void wlr_input_timestamps_manager_v1_destroy(struct wlr_input_timestamps_manager
...
+
+	struct wlr_seat_client *seat_client = NULL;
+	struct wlr_input_timestamps_v1 *input_timestamps;
+	wl_list_for_each(input_timestamps, &manager->input_timestamps, link) {
+		if (input_timestamps_type != input_timestamps->input_timestamps_type) {
+			continue;
+		}
+
+		switch (input_timestamps_type) {
+			case WLR_INPUT_TIMESTAMPS_KEYBOARD:;
```
ddevault: no semi-colons after case clauses, align case with switch indentwise

include/wlr/types/wlr_keyboard.h
```diff
@@ -96,6 +96,7 @@ enum wlr_key_state {
 
 struct wlr_event_keyboard_key {
 	uint32_t time_msec;
+	uint32_t time_frac_nsec;
```
ddevault: For naming I think this would be cleaner as `time_nsec` with a comment explaining. If not, prefer `time_nsec_part` over `time_frac_nsec`.

rootston/cursor.c
```diff
@@ -373,17 +379,21 @@ void roots_cursor_handle_motion_absolute(struct roots_cursor *cursor,
...
-		event->button, event->state, cursor->cursor->x, cursor->cursor->y);
+		event->time_frac_nsec, event->button, event->state,
+		cursor->cursor->x, cursor->cursor->y);
 }
 
 void roots_cursor_handle_axis(struct roots_cursor *cursor,
 		struct wlr_event_pointer_axis *event) {
+	wlr_input_timestamps_manager_v1_send_pointer_timestamp(
+		cursor->seat->input->server->desktop->input_timestamps_manager,
+		cursor->seat->seat, (event->time_msec / 1000.0), event->time_frac_nsec);
```
ddevault: Unnecessary parenthesis here

backend/wayland/wl_seat.c
```diff
@@ -77,10 +78,19 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 
+	uint32_t time_frac_nsec;
+	if ((backend->input_timestamps.manager) &&
+		(backend->input_timestamps.p.tv_sec == (time / 1000))) {
```
ddevault: Same comment about parens here

ddevault: (only more so)

examples/input-timestamps-unstable-v1.c
```diff
@@ -0,0 +1,565 @@
...
+
+static const struct wl_registry_listener registry_listener = {
+	.global = registry_handle_global,
+	.global_remove = registry_handle_global_remove,
+};
+
+
+/**
+ * wl_seat handling
+ */
```
ddevault: same with all of these comments

examples/input-timestamps-unstable-v1.c
```diff
@@ -0,0 +1,565 @@
...
+
+static void seat_handle_capabilities(void *data, struct wl_seat *seat,
+		enum wl_seat_capability caps) {
+	struct client *c = (struct client *) data;
+
+	/* Setup the pointer */
```
ddevault: and these, don't over-comment

examples/input-timestamps-unstable-v1.c
```diff
@@ -0,0 +1,565 @@
...
+		return -1;
+	}
+
+	if (ftruncate(fd, size) < 0) {
+		close(fd);
+		return -1;
+	}
+
+	return fd;
+}
```
ddevault: We have a helper elsewhere for this stuff. It would probably also be simpler/shorter to use a DRM surface instead of an SHM one.

ddevault:
Overall this looks quite good. Thanks for breaking it into easily reviewable chunks.

I think we should squash this into the following commits, in this order:

1. Add high-resolution timestamps to internal wlr input events
2. Add input-timestamps protocol implementation
3. Rig it up to rootston
4. Add example client

examples/input-timestamps-unstable-v1.c
emersion: You probably mean GL surface

include/wlr/types/wlr_input_timestamps_v1.h
arandomhuman: All (except one) headers follow this pattern in this directory, if they have any top level comment. E.g., wlr_cursor, wlr_idle, wlr_idle_inhibit, etc.

include/wlr/types/wlr_keyboard.h
arandomhuman: I would have preferred `time_nsec` too if it wasn't begging to be confused for `time_msec`, both semantically and spelling-wise. I chose `frac` because the fractional part of a floating point number has a very well established meaning.

arandomhuman:
Sorry, busy week!

@ddevault: I've incorporated most of the suggestions, except the ones I commented on, and removing the comments: consider a strong NAK from my side. In the source file, they help logically segregating the functions; and the example is supposed to be slightly hand-hold-y. Making it easy to parse code never hurts.

emersion:
>removing the comments: consider a strong NAK from my side. In the source file, they help logically segregating the functions; and the example is supposed to be slightly hand-hold-y. Making it easy to parse code never hurts.

FWIW, I'm also not a fan of these comments.

include/wlr/types/wlr_input_timestamps_v1.h
emersion: Yeah, I agree with @arandomhuman here. If we want to move these somewhere else this should be done for all files.

.gitignore
```diff
@@ -1,6 +1,7 @@
 .clang_complete
 *.o
 *.a
+.*.sw?
```
emersion: This should be removed.

include/wlr/types/wlr_keyboard.h
emersion: TBH I'm not a fan of splitting the timestamp like this. Using a single `uint64_t time_nsec` would be cleaner.

backend/rdp/peer.c
```diff
@@ -105,6 +105,7 @@ static int xf_input_mouse_event(rdpInput *input,
 		struct wlr_event_pointer_motion_absolute event = { 0 };
 		event.device = wlr_device;
 		event.time_msec = timespec_to_msec(&now);
+		event.time_frac_nsec = now.tv_nsec;
```
emersion: No modulo here?

backend/wayland/backend.c
```diff
@@ -227,6 +231,11 @@ struct wlr_backend *wlr_wl_backend_create(struct wl_display *display,
 		goto error_registry;
 	}
 
+	if (!wl->input_timestamps.manager) {
+		wlr_log(WLR_ERROR,
```
emersion: This can just be INFO or DEBUG

.gitignore
arandomhuman: @ascent12 already pointed that out. I'll remove the .gitignore commit when I'm squashing.

backend/wayland/wl_seat.c
```diff
@@ -78,10 +79,19 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 
+	uint32_t time_frac_nsec;
+	if (backend->input_timestamps.manager &&
+		backend->input_timestamps.p.tv_sec == time / 1000) {
```
emersion: Style: one more indent to tell the difference between the condition and the body

backend/wayland/wl_seat.c
```diff
@@ -96,11 +106,20 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 
+	uint32_t time_frac_nsec;
+	if ((backend->input_timestamps.manager) &&
```
emersion: Style: unnecessary parentheses here (applies to the rest of the file too)

.gitignore
emersion: Oh, sorry about that. Leaving this discussion opened so that we don't forget about it.

backend/wayland/wl_seat.c
```diff
@@ -195,24 +232,21 @@ static void keyboard_handle_keymap(void *data, struct wl_keyboard *wl_keyboard,
...
-	struct timespec now;
-	clock_gettime(CLOCK_MONOTONIC, &now);
-	return now.tv_nsec / 1000;
-}
-
 static void keyboard_handle_enter(void *data, struct wl_keyboard *wl_keyboard,
 		uint32_t serial, struct wl_surface *surface, struct wl_array *keys) {
-	struct wlr_input_device *dev = data;
+	struct wlr_input_device *dev =
+		&((struct wlr_wl_input_device *) data)->wlr_input_device;
```
emersion: Hmm, please do this in two steps

backend/wayland/wl_seat.c
```diff
@@ -244,22 +281,34 @@ static void keyboard_handle_leave(void *data, struct wl_keyboard *wl_keyboard,
 
 static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
 		uint32_t serial, uint32_t time, uint32_t key, uint32_t state) {
-	struct wlr_input_device *dev = data;
-	assert(dev && dev->keyboard);
+	struct wlr_wl_input_device *dev = data;
+	struct wlr_input_device *wlr_dev = &(dev->wlr_input_device);
```
emersion: Style: unnecessary parentheses (applies to the rest of the file too)

backend/wayland/wl_seat.c
```diff
@@ -521,11 +570,24 @@ void create_wl_keyboard(struct wl_keyboard *wl_keyboard, struct wlr_wl_backend *
...
 	dev->resource = wl_keyboard;
 	wlr_signal_emit_safe(&wl->backend.events.new_input, wlr_dev);
 }
 
+static void input_timestamps_handle_timestamp(void *data,
+		struct zwp_input_timestamps_v1 *zwp_input_timestamps_v1,
+		uint32_t tv_sec_hi, uint32_t tv_sec_lo, uint32_t tv_nsec) {
+	struct timespec *timestamp = data;
+
+	timestamp->tv_sec = ((uint64_t) tv_sec_hi << 32) + tv_sec_lo;
```
emersion: Style: no space between cast and variable name

```c
(uint64_t)tv_sec_hi
```

backend/rdp/peer.c
arandomhuman: Valid timespec objects should have the correct range [0-999999999] in the tv_nsec field already, unless I'm wrong.

backend/x11/input_device.c
```diff
@@ -20,6 +20,7 @@ static void send_key_event(struct wlr_x11_backend *x11, uint32_t key,
 		enum wlr_key_state st, xcb_timestamp_t time) {
 	struct wlr_event_keyboard_key ev = {
 		.time_msec = time,
+		.time_frac_nsec = (time % 1000) * 1000000,
```
emersion: Hmm. This doesn't seem right.

include/backend/wayland.h
```diff
@@ -39,6 +39,11 @@ struct wlr_wl_backend {
 	struct wl_keyboard *keyboard;
 	struct wlr_wl_pointer *current_pointer;
 	char *seat_name;
+	struct input_timestamps {
+		struct zwp_input_timestamps_manager_v1 *manager;
+		struct timespec kb;
+		struct timespec p;
```
emersion: It wouldn't hurt to spell `pointer` here

include/backend/wayland.h
```diff
@@ -39,6 +39,11 @@ struct wlr_wl_backend {
 	struct wl_keyboard *keyboard;
 	struct wlr_wl_pointer *current_pointer;
 	char *seat_name;
+	struct input_timestamps {
```
emersion: No need to name this struct

include/rootston/cursor.h
```diff
@@ -104,7 +104,7 @@ void roots_cursor_handle_focus_change(struct roots_cursor *cursor,
 void roots_cursor_handle_constraint_commit(struct roots_cursor *cursor);
 
 void roots_cursor_update_position(struct roots_cursor *cursor,
-	uint32_t time);
+	uint32_t time, uint32_t time_frac_nsec);
```
emersion: The first argument should probably be renamed to `time_msec`

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,383 @@
...
+	struct wlr_seat_client *focused_client = NULL;
+	switch (input_timestamps_type) {
+	case WLR_INPUT_TIMESTAMPS_KEYBOARD:;
+		focused_client = seat->keyboard_state.focused_client;
+		break;
+	case WLR_INPUT_TIMESTAMPS_POINTER:;
+		focused_client = seat->pointer_state.focused_client;
+		break;
+	case WLR_INPUT_TIMESTAMPS_TOUCH:;
+		// TODO wlr_seat_touch_state has no focused_client member
```
emersion: Hmm, this is because there can be multiple touches, each on a different surface. Could be left as a follow-up issue.

I'd prefer doing `focused_client = NULL` instead of sending events to an unrelated client though.

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,383 @@
...
+			seat_client = wlr_seat_client_from_touch_resource(
+				input_timestamps->input_resource);
+			break;
+		}
+		if (seat != input_timestamps->seat || focused_client != seat_client) {
+			continue;
+		}
+
+		zwp_input_timestamps_v1_send_timestamp(input_timestamps->resource,
+			(uint32_t) (tv_sec >> 32), (uint32_t) tv_sec, (uint32_t) tv_nsec);
```
emersion: Style: no space between cast and expression.

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,383 @@
...
+	wl_resource_add_destroy_listener(input_timestamps->input_resource,
+			&input_timestamps->input_destroy);
+	input_timestamps->input_destroy.notify = handle_input_destroy;
+
+	wl_resource_set_implementation(input_timestamps_resource, &impl,
+		input_timestamps, handle_resource_destroy);
+
+	switch (input_timestamps_type) {
+	case WLR_INPUT_TIMESTAMPS_KEYBOARD:
+		wlr_signal_emit_safe(&manager->events.new_keyboard_timestamps,
```
emersion: Maybe we could have a single signal, instead of 3 separate signals?

types/wlr_input_timestamps_v1.c
```diff
@@ -0,0 +1,383 @@
...
+
+static const struct zwp_input_timestamps_manager_v1_interface manager_impl;
+static const struct zwp_input_timestamps_v1_interface impl;
+
+/**
+ * helper functions
+ */
+
+static struct wlr_input_timestamps_v1 *
+wlr_input_timestamps_v1_from_resource(struct wl_resource *resource) {
```
emersion: Static functions shouldn't have the `wlr_` prefix. This is reserved for exported functions.

backend/rdp/peer.c
emersion: Oh right, indeed

backend/wayland/backend.c
arandomhuman: Made it INFO.

types/wlr_input_timestamps_v1.c
arandomhuman: I wouldn't prefer users checking against the type enum on their own, but upto you.

include/wlr/types/wlr_keyboard.h
arandomhuman: Considering most event timestamps are in millisecond precision, that would add a lot of unnecessary arithmetic IMO.

backend/x11/input_device.c
arandomhuman: `(time % 1000)` has a value within 10^3, so we multiple by 10^6 to get nanosecond (10^9) precision.

types/wlr_input_timestamps_v1.c
arandomhuman: Setting `focused_client = NULL` makes it ignore this call for touch events because of the following check. It should ideally be send to all clients receiving the corresponding event; let me check how this stuff is handled elsewhere.

arandomhuman:
Lots of lazy formatting issues; I'll do a second pass to see if I fixed them all.

types/wlr_input_timestamps_v1.c
emersion: Well, why do we need these signals in the first place? I don't think a new input timestamps object requires any action from the compositor. I'd rather leave this out until there is a clear use-case.

types/wlr_input_timestamps_v1.c
emersion: >Setting focused_client = NULL makes it ignore this call for touch events because of the following check.

Yes, that's intended.

>It should ideally be send to all clients receiving the corresponding event; let me check how this stuff is handled elsewhere.

No, only to the client receiving the touch point. A new `touch_point` argument is likely required.

backend/x11/input_device.c
emersion: Isn't `time_frac_nsec` between 0 and 999999?

types/wlr_input_timestamps_v1.c
arandomhuman: Honestly, I can't think of any use-cases except contrived ones like logging or unit testing. They're there mostly to be on par with other protocols.

backend/x11/input_device.c
arandomhuman: No, between 0 and 999999999 (10^10 - 1).

types/wlr_input_timestamps_v1.c
emersion: Then let's just remove them

backend/x11/input_device.c
emersion: Wait, so some information is duplicated between `time_msec` and `time_frac_nsec`? This is pretty confusing.

include/wlr/types/wlr_keyboard.h
ddevault: Let's go with two fields:

```c
uint32_t time_msec;
uint64_t time_nsec;
```

Which both represent the same timestamps, and don't need to be combined to get the full picture. I think this is the cleanest approach for this API, and is backwards compatible.

types/wlr_input_timestamps_v1.c
ddevault: bump

examples/input-timestamps-unstable-v1.c
```diff
@@ -0,0 +1,513 @@
...
+	struct zwp_input_timestamps_v1 *pointer_input_timestamps;
+	struct zwp_input_timestamps_v1 *keyboard_input_timestamps;
+
+	struct timespec pointer_timestamp;
+	struct timespec keyboard_timestamp;
+};
+
+/**
+ * global declarations
+ */
```
ddevault: similar comments in this file can go, too

backend/wayland/wl_seat.c
```diff
@@ -521,11 +587,27 @@ void create_wl_keyboard(struct wl_keyboard *wl_keyboard, struct wlr_wl_backend *
...
 	wlr_signal_emit_safe(&wl->backend.events.new_input, wlr_dev);
 }
 
+static void input_timestamps_handle_timestamp(void *data,
+		struct zwp_input_timestamps_v1 *zwp_input_timestamps_v1,
+		uint32_t tv_sec_hi, uint32_t tv_sec_lo, uint32_t tv_nsec) {
+	struct timespec *timestamp = data;
+
+	uint64_t tv_sec = ((uint64_t)tv_sec_hi << 32) + tv_sec_lo;
+	assert(tv_sec < (uint64_t)LONG_MAX && "input-timestamps value is too big");
```
emersion: Why check this?

ddevault:
bump?

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1724

PR #1726, opened 2019-06-16
Minor output cleanup

emersion:
See the individual commits.

Breaking change: `wlr_output_impl.transform` has been removed.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -678,11 +678,6 @@ bool wlr_drm_connector_add_mode(struct wlr_output *output,
 	return true;
 }
 
-static void drm_connector_transform(struct wlr_output *output,
-		enum wl_output_transform transform) {
-	output->transform = transform;
-}
-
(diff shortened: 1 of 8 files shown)


ddevault:
At some point weren't we thinking of implementing output transforms for DRM?

ascent12:
It's really hard to make use of properly (e.g. it only works correctly on intel with certain modifiers), and most drivers don't support a lot of transforms anyway, especially not useful ones like 90 or 270.
I'm fine with removing it.

ddevault:
Thanks!

emersion:
>At some point weren't we thinking of implementing output transforms for DRM?

Re-posting the commit message in case you missed it:

>We could add support for hardware transforms, but:
>
> - This would require a different field (something like hardware_transform)
> - Not all combinations are possible because there often are hardware
>   limitations
> - The Wayland protocol isn't ready for this (in particular xdg-output, see [1])
>
> This belongs to a different patch series anyway.

[1]: https://patchwork.freedesktop.org/series/52324/

PR #1727, opened 2019-06-17
x11: Send a frame event on the pointer after button events

dcz-purism:
Without the immediate frame event, the button event would not be processed in time: https://source.puri.sm/Librem5/phoc/issues/


emersion:
Thanks!

dcz-purism:
Thank you!

Issue #1723, opened 2019-06-13
DRM leasing


ddevault:
I'm working on implementing DRM leasing (and with it, VR support) and wanted to start the discussion here on how it should shake out. For those unaware, DRM leasing allows us to "lease" some DRM resources (connector, crtc, encoder) to another process, allowing us to share the DRM master. This is used by VR applications to have exclusive and low-latency control over the HMD (head mounted display). I think a comprehensive solution for VR on Wayland (and wlroots) involves the following broad steps, in no particular order:

- Expose the "non-desktop" flag to downstream compositors, so they can decide how to use non-desktop outputs (HMDs generally have this flag set, and are generally unsuitable for inclusion in the desktop layout). A typical compositor will not use this output at all, and will make it available for leasing. A VR compositor might seek out this output specifically to run its VR interface on. etc.
- Clean up and get merged the [DRM leasing Wayland protocol extension](https://patchwork.freedesktop.org/patch/247095/). I have a [cleaned up draft here](https://git.sr.ht/~sircmpwn/wayland-protocols/tree/master/unstable/drm-lease/drm-lease-unstable-v1.xml) for your consideration.
- Implement `wlr_drm_lease_manager_v1`. We could do this at a low-level or a high-level. The low-level API would involve expanding the public DRM backend API so that its users can enumerate our DRM resources and offer fine-grained leases of connectors, crtcs, and encoders separately. Until someone comes up with a use-case that demands this, I'd prefer a high-level approach: the downstream compositor just tells `wlr_drm_lease_manager_v1` which `wlr_outputs` it would like to offer for leasing. This will still require some changes to the DRM backend to allow the lease manager to ask the backend to stop using certain resources, but it could just use the existing DRM backend headers to dig into DRM internals and since it all lives inside of wlroots (rather than downstream compositors), it simplifies the maintenance burden significantly.
    - In the high-level model, when a client takes you up on the lease offer, the backend will emit an `output_remove` event for that `wlr_output`, so you stop using it. It'll reappear if you revoke the lease, and you can still access it (though it'll be a programming error to mutate it) by browsing the list of leases.

[...]

PR #1730, opened 2019-06-20
Introduce wlr_drm_lease_v1

ddevault:
Still deep in the trenches on this patch, but opening it in case anyone is curious or has early feedback.

- [x] meson rigging
- [x] Protocol state management
- [x] Create & transfer the lease
- [x] DRM backend updates
- [x] Example client
- [x] Rebase
- [x] Withdraw connector offers when a lease is issued
- [x] Reissue connector offers when lease terminates
- [x] Address known issues
- [x] Rootston support
- [ ] Consensus on protocol

Related ticket: #1723 

Test client: https://git.sr.ht/~sircmpwn/kmscube (`./kmscube -l`)

I don't intend to merge this until this protocol makes it through review on `wayland-devel`:

https://lists.freedesktop.org/archives/wayland-devel/2019-June/040672.html

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1349,6 +1349,12 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 
 			get_drm_connector_props(drm->fd, wlr_conn->id, &wlr_conn->props);
 
+			uint64_t non_desktop;
+			if (get_drm_prop(drm->fd, wlr_conn->id,
+						wlr_conn->props.non_desktop, &non_desktop)) {
+				wlr_conn->output.non_desktop = non_desktop;
+			}
+
 			size_t edid_len = 0;
 			uint8_t *edid = get_drm_prop_blob(drm->fd,
 				wlr_conn->id, wlr_conn->props.edid, &edid_len);
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1254,6 +1254,10 @@ static uint32_t get_possible_crtcs(int fd, drmModeRes *res,
 
 static void drm_conn_init_wlr_output(struct wlr_drm_backend *drm,
 		struct wlr_drm_connector *wlr_conn, drmModeConnector *drm_conn) {
+	wlr_log(WLR_DEBUG, "Initializing wlr_output for drm connector %s-%d",
+			conn_get_name(drm_conn->connector_type),
+			drm_conn->connector_type_id);
+	memset(&wlr_conn->output, 0, sizeof(struct wlr_output));
 	wlr_output_init(&wlr_conn->output,
 			&drm->backend, &output_impl, drm->display);
 	snprintf(wlr_conn->output.name, sizeof(wlr_conn->output.name),
@@ -1378,7 +1382,7 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 			}
 			if (!found) {
 				/* Yes, it was */
-				wlr_log(WLR_DEBUG, "DRM lease %d terminated by kernel",
+				wlr_log(WLR_DEBUG, "scan_drm_connectors: DRM lease %d terminated",
 						wlr_conn->lessee_id);
 				wlr_conn->state = WLR_DRM_CONN_DISCONNECTED;
 				wlr_conn->lessee_id = 0;
@@ -1756,8 +1760,8 @@ int drm_terminate_lease(struct wlr_drm_backend *backend, uint32_t lessee_id) {
 	wl_list_for_each(conn, &backend->outputs, link) {
 		if (conn->state == WLR_DRM_CONN_LEASED
 				&& conn->lessee_id == lessee_id) {
-			conn->state = WLR_DRM_CONN_DISCONNECTED;
 			conn->lessee_id = 0;
+			/* Will be re-initialized in scan_drm_connectors */
 		}
 	}
 	for (size_t i = 0; i < backend->num_crtcs; ++i) {
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -270,6 +270,7 @@ bool init_drm_resources(struct wlr_drm_backend *drm) {
 		struct wlr_drm_crtc *crtc = &drm->crtcs[i];
 		crtc->id = res->crtcs[i];
 		crtc->legacy_crtc = drmModeGetCrtc(drm->fd, crtc->id);
+		crtc->lessee_id = 0;
 		get_drm_crtc_props(drm->fd, crtc->id, &crtc->props);
 	}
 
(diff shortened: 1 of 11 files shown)


include/wlr/types/wlr_drm_lease_v1.h
```diff
@@ -0,0 +1,141 @@
...
+		struct wl_signal lease_requested; // wlr_drm_lease_request_v1
+		/**
+		 * Upon receiving this signal, it is safe to re-use the leased
+		 * resources. After the signal is processed, the backend will re-signal
+		 * new_output events for each leased output.
+		 *
+		 * The lifetime of the lease reference after the signal handler returns
+		 * is not defined.
+		 */
+		struct wl_signal lease_destroyed; // wlr_drm_lease_v1
```
emersion: This should be an event in `struct wlr_drm_lease_v1`

mv0:
Hi,

Sorry to barge in, but any reason not posting the protocol on wayland-dev ML? I think there might be others which would like to take a look, and this shouldn't be (a) compositor specific protocol.

ddevault:
I prefer to have a working implementation to demonstrate the protocol before I submit it for others to consider. Writing an implementation often turns up defects in the protocol, which I would rather hear about before it's standardized than after.

ddevault:
Test client: https://git.sr.ht/~sircmpwn/kmscube

ddevault:
This is basically ready for review. Known issues:

- I hardcoded which output I want to lease for testing purposes, later this will be replaced with automatically offering non-desktop outputs and a `lease` option in the config
- Needs to be rebased

The DRM changes and lease manager implementation should be in good shape, though, and if you want to test it you can update the hardcoded output name to some other output for leasing.

include/wlr/types/wlr_drm_lease_v1.h
ddevault: I haven't forgotten this, either.

backend/drm/drm.c
```diff
@@ -649,6 +654,8 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 		enum wl_output_transform transform,
 		int32_t hotspot_x, int32_t hotspot_y, bool update_texture) {
 	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
+	assert((conn->state != WLR_DRM_CONN_LEASED || texture == NULL)
+			&& "Use of leased connector is a programming error");
```
ddevault: This `texture == NULL` exception is necessary because when the output is destroyed, its cursors are destroyed as well.

backend/drm/drm.c
```diff
@@ -1032,7 +1050,7 @@ static void realloc_crtcs(struct wlr_drm_backend *drm) {
 		connector_match[i] = -1;
 	}
 	for (size_t i = 0; i < drm->num_crtcs; ++i) {
-		if (new_match[i] != UNMATCHED) {
+		if (new_match[i] != UNMATCHED && new_match[i] != SKIP) {
```
ddevault: Without this sway would crash when the lease is terminated, but I'm not 100% sure this is the right fix rather than a band-aid. Probably a consequence of the logic in the previous hunk, which may also be incorrect. Seems to work but w/e. Paging @ascent12 for comment.

ddevault: Actually, this is definitely wrong.

ddevault: Found the issue, it was elsewhere

computer-whisperer:
This doesn't seem to work for me at the moment. 

Here is a log from starting sway, running kmscube -l, and immediately closing sway.
https://gist.github.com/57beb710b04f7800e560a78cd6ea5779

Sway config file:
https://gist.github.com/17b45e21e1f23086fbc9c147aa882041

Here is a relevant snippet starting at line 1833:
> 00:00:17.683 [DEBUG] [backend/drm/drm.c:1718] Issuing DRM lease with the 4 objects:
> 00:00:17.683 [DEBUG] [backend/drm/drm.c:1725] connector: 59 crtc: 57 primary plane: 40
> 00:00:17.683 [DEBUG] [backend/drm/drm.c:1728] cursor plane: 56
> 00:00:17.683 [ERROR] [types/wlr_drm_lease_v1.c:118] drm_create_lease failed: Invalid argument

System configuration info:
Arch Linux
Linux 5.7.2-2 w/ ck patches
libdrm 2.4.102-1
Discrete AMD gpu
Integrated Intel gpu
Valve Index as DP-2 (plugged into AMD gpu)

I am using the following patches:
https://github.com/swaywm/sway/pull/4289.diff
https://github.com/swaywm/wlroots/pull/1730.diff
https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/248.diff

ddevault:
No support whatsoever is provided for these patchsets, you'll have to investigate the problem yourself and send an updated patch.

optlink:
If anyone has an idea on where I should start hacking away here is where I am at:

I am trying to get DRM leasing with SteamVR and a Valve Index working. Apparently SteamVR tries to identify the HMD output by the available output modes. Running xrandr with these patches applied gives a "non-desktop" XWAYLAND1 output with no modes. SteamVR then ignores that output and carries on until it realizes there's no display for it to use. I can manually add modes using the xrandr tool but that's not quite enough. SteamVR requires the "preferred mode" to be exactly 2880x1600. I have not yet found a way set which mode is preferred externally. Does anyone know if this is something that should be handled by wlroots or sway? Shouldn't wlroots pass the supported modes onto xwayland even if it's a non-desktop output?

ddevault:
You need my Xwayland patches for this to work.

optlink:
[This](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/248) patch? I am running it already on master.

michaelnew:
@optlink I tried this out today with a similar setup (Valve Index, SteamVR, and in my case a Vega 64). I applied the xwayland patch to the xserver master branch, but for sway and wlroots I just build the drm-lease branches rather than apply them as patches to master.

If I run xrandr I see `XWAYLAND0` and `XWAYLAND1` (the Index), with about a dozen modes listed. If I start up Steam and launch VR (I'm using the Steam beta and the SteamVR beta) everything works great, but `XWAYLAND1` disappears from the xrandr output. Once I quit VR, `XWAYLAND1` is still gone, and I then see `XWAYLAND2`, but no modes listed. And from that point if I launch SteamVR it tries to output to my main display, and eventually quits. Exiting and relaunching Sway fixes things.

optlink:
@michaelnew The differences between our setups that I can see are:
1. RX 5700 XT vs Vega 64
2. I applied the wlroots patches to wlroots 0.11
3. I am using sway 1.5 with a slightly modified [patch](https://hastebin.com/izaqovexoc.cpp) because it did not apply cleanly

I suspect the problem is with 3 since I do not fully understand the impact of the changes that were made for sway 1.5.
The main symptom is that xrandr returns no modes at all for XWAYLAND1, even before taking a lease. Running SteamVR anyway fails pretty spectacularly.

optlink:
I rebuilt sway and wlroots from their respective drm-lease branches and I get the exact same result. There must be something else wrong.

optlink:
It turns out that the Xwayland patches are broken on master currently. Building from the merge request branch works but you will need a fix for the GLX issue caused by recent changes to libglvnd: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1032

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2929

PR #1731, opened 2019-06-20
Support pointer-gestures on Wayland backend

myfreeweb:
This is useful for developing compositors that have their own gestures.


ddevault:
Thank you!

PR #1732, opened 2019-06-21
backend/drm: Simplify object matching code

ascent12:
We originally used `match_obj` on planes, but this was largely unnecessary. Instead, this assigns planes statically at startup.

See https://github.com/swaywm/wlroots/issues/1723#issuecomment-504247814
It'll help if some people could test this as well (especially hotplugging), so it's not just "works for me".

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1,4 +1,4 @@
-#define _POSIX_C_SOURCE 200112L
+#define _XOPEN_SOURCE 700
 #include <assert.h>
 #include <drm_fourcc.h>
 #include <drm_mode.h>
@@ -10,6 +10,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 #include <time.h>
 #include <wayland-server.h>
 #include <wayland-util.h>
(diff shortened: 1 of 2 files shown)


backend/drm/drm.c
```diff
@@ -88,118 +164,68 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		 * logic. Primary and cursor planes should only work on a
+		 * single CRTC, and this should be perfectly adequate, but
+		 * overlay planes can potentially work with multiple CRTCs,
+		 * meaning this could return inefficent/skewed results.
+		 *
+		 * However, we don't really care about overlay planes, as we
+		 * don't support them yet. We only bother to keep basic
+		 * tracking of them from DRM lease clients.
+		 */
+		int crtc_i = ffs(plane->possible_crtcs);
```
emersion: Can we rename this to `crtc_pos` instead of `crtc_i`? It starts at 1, so using "index" is confusing.

ascent12: Makes sense. I went with `crtc_bit` and added more clarification as to what it means in the comment.

backend/drm/drm.c
```diff
@@ -1537,22 +1459,6 @@ static void drm_connector_cleanup(struct wlr_drm_connector *conn) {
 	switch (conn->state) {
 	case WLR_DRM_CONN_CONNECTED:
 	case WLR_DRM_CONN_CLEANUP:;
```
panaman67:
```suggestion
	case WLR_DRM_CONN_CLEANUP:
```
semicolon not needed for fall through. Should remove?

backend/drm/drm.c
```diff
@@ -88,118 +164,72 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		/*
+		 * This is a very naive implementation of the plane matching
+		 * logic. Primary and cursor planes should only work on a
+		 * single CRTC, and this should be perfectly adequate, but
+		 * overlay planes can potentially work with multiple CRTCs,
+		 * meaning this could return inefficent/skewed results.
+		 *
+		 * However, we don't really care about overlay planes, as we
+		 * don't support them yet. We only bother to keep basic
+		 * tracking of them from DRM lease clients.
```
ddevault: for*

ddevault:
Holds up in testing :+1:

backend/drm/drm.c
ascent12: Pushed a fix

backend/drm/drm.c
ascent12: Pushed a fix

emersion:
Thanks!

PR #1734, opened 2019-06-24
backend/drm: Exit-early if 0 crtcs

ascent12:
This fixes an assertion failure if we're using a device that has 0 crtcs
as a renderer.
This would happen on some laptops with discrete GPUs.


ascent12:
This fix works according to the reporter on IRC.

emersion:
Thanks!

PR #1736, opened 2019-06-24
virtual_keyboard: Require keymap before accepting keycodes

dcz-purism:
A simple check capturing the nonsensical case of sending keycodes without having defined a keymap before. Key presses would still go through in case of wlroots due to having a default keymap defined, but I can't see relying on this as anything but a source of mistakes. Therefore, the implementation matches the protocol now, and rejects key presses if there's no keymap.

### DIFF
--- a/types/wlr_virtual_keyboard_v1.c
+++ b/types/wlr_virtual_keyboard_v1.c
@@ -61,6 +61,7 @@ static void virtual_keyboard_keymap(struct wl_client *client,
 		goto keymap_fail;
 	}
 	wlr_keyboard_set_keymap(keyboard->input_device.keyboard, keymap);
+	keyboard->has_keymap = true;
 	xkb_keymap_unref(keymap);
 	xkb_context_unref(context);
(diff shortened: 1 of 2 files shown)


types/wlr_virtual_keyboard_v1.c
```diff
@@ -76,6 +77,11 @@ static void virtual_keyboard_key(struct wl_client *client,
 		uint32_t state) {
 	struct wlr_virtual_keyboard_v1 *keyboard =
 		virtual_keyboard_from_resource(resource);
+	if (!keyboard->has_keymap) {
+		wl_resource_post_error(resource,
+			ZWP_VIRTUAL_KEYBOARD_V1_ERROR_NO_KEYMAP,
+			"Cannot send a keypress before defining a keymap");
```
emersion: A `return` is missing after `wl_resource_post_error`

dcz-purism:
Thanks for catching that!

PR #1740, opened 2019-06-28
backend/wayland: add support for relative pointers

arandomhuman:
This branch is rebased on the input-timestamps PR (#1724) because of lots of tedious merge conflicts. Only the final two commits are new.

Description of commits:

`- backend/wayland: use a locked pointer when receiving relative motion events`
Adds a relative pointer to the backend which fires off relative motion events as it gets them. For now, motion events are mostly duplicated (this, plus absolute motion events in pointer motion handling). Odd consequence of this is that the cursor image and the actual cursor may not coincide.

` - backend/wayland: use relative-pointer to issue relative motion events`
If possible, lock the pointer and hide the server cursor, and use a software cursor exclusively. The server and client cursors are kept in sync with `set_cursor_position_hint`, and a motion event on pointer enter events.

Test: open sway/rootston inside another sway session, and run the relative-pointer example. Sway handles cursor position hints properly. It _sometimes_ sets the cursor to the middle of the surface when locking the pointer (usually when it's moving slow), but otherwise it's fine.


emersion:
Sorry about this, I forgot this existed and we merged another PR to add relative-pointer support to the Wayland backend.

>If possible, lock the pointer and hide the server cursor, and use a software cursor exclusively. The server and client cursors are kept in sync with set_cursor_position_hint, and a motion event on pointer enter events.

Hmm, can you explain what's the rationale for this?

Issue #755, opened 2018-03-27
Fix serial validation for wlr_shell grabs


SirCmpwn:
https://github.com/swaywm/wlroots/pull/731 contains a workaround that just disables validation

emersion:
Related: serials are also used for some requests like setting selection. Currently we either don't check it (see `wlr_seat_validate_grab_serial`) or do some weird stuff like:

```c
seat->serial - serial < UINT32_MAX / 2
```

mstoeckl:
Another issue with the seat's selection serial value is the serial used to perform the validation (now `seat->selection_serial`) is shared between all programs using the seat, so if one client sets its selection with a very large serial value, then copy-paste will fail for all other clients until that serial is reached by the compositor.

emersion:
>if one client sets its selection with a very large serial value

We should check that the serial sent by the client is smaller than the current display serial.

Weston has likely the same issue.

>Serial values thus should be per-client.

This isn't possible because it won't protect against races.

mstoeckl:
Right, I think I understand it now. The "correct" validation for primary/gtk selection serials is to verify that
1. The client provided a serial value that the compositor previously sent to it. (Otherwise, the client could try a large set of sequential serial values, and steal the role of being the selection source from another application by picking.)
2. The serial provided by the request comes after the seat's selection source's serial. (Otherwise, given applications A,B, input events iA, iB, and selection requests sA,sB, if the ordering is iA,iB,sB,sA , then application A would provide the copy-paste content even though the user most recently asked for it from B.)

Because serials are only 32-bit, "after" can only be correctly defined for serial values closer than 2^31.

Validating the second property is easy and has already been done. Fully validating the first is significantly more complicated: it looks like the only way to do this correctly is to record all serial values sent to a client. We can approximate the first property by ignoring all "future" serial values.

If wlroots would aim for full validation, the serial recording can be relaxed slightly and kept fast by using a fixed size ring buffer (say, using 32 KB of memory) and assuming that all sufficiently old serial values are valid (but would probably fail the second test.)

mstoeckl:
[...]

PR #1742, opened 2019-06-28
Implement serial validation for selection requests

mstoeckl:
See also commit message, and #755. In short, this change verifies, for copy selection requests, that the serial numbers used to determine which request corresponds to the most recent user input were actually sent to the client.

Right now the code registers serial numbers for the keyboard, pointer, and touch events; serial numbers produced by any other input source will fail validation when provided by a client. (Serial numbers generated by e.g. xwayland code have no associated seat or client, and are not checked.) I'm not familiar with all of wlroots' protocols, so I don't know if e.g. tablet events are associated with a seat and should be allowed to trigger copy requests. 

This change alters the size of `struct wlr_seat_client`, silently breaking ABI for compositors which use that struct inline.

To test this PR, you'll need a misbehaving client; either modify some small program that does copy+paste to send in bad serial values, or try e.g. this work-in-progress test program: mstoeckl/wleird@01920741bb8e19bbd5af2e9ce53d957122d37847 .

Edit: Please remind me to rebase once the changes are acceptable.

### DIFF
--- a/types/seat/wlr_seat.c
+++ b/types/seat/wlr_seat.c
@@ -370,3 +370,36 @@ bool wlr_seat_validate_grab_serial(struct wlr_seat *seat, uint32_t serial) {
 	//	serial == seat->touch_state.grab_serial;
 	return true;
 }
+
+void wlr_serial_add(struct wlr_serial_ringset *set, uint32_t serial) {
+	if (set->count == 0 || set->data[set->end].max_incl + 1 != serial) {
+		set->count++;
+		if (set->count > WLR_SERIAL_RINGSET_SIZE) {
+			set->count = WLR_SERIAL_RINGSET_SIZE;
+		}
+		set->end = (set->end + 1) % WLR_SERIAL_RINGSET_SIZE;
+		set->data[set->end].min_incl = serial;
+		set->data[set->end].max_incl = serial;
+	} else {
+		set->data[set->end].max_incl = serial;
+	}
+}
+
+bool wlr_serial_maybe_valid(struct wlr_serial_ringset *set, uint32_t serial) {
+	for (int i = 0; i < set->count; i++) {
+		int j = (set->end - i + WLR_SERIAL_RINGSET_SIZE) % WLR_SERIAL_RINGSET_SIZE;
+		if (set->data[j].max_incl - serial > UINT32_MAX / 2) {
+			return false;
+		}
+		if (serial - set->data[j].min_incl <= UINT32_MAX / 2) {
+			return true;
+		}
+	}
+	return true;
+}
+
+uint32_t wlr_seat_client_next_serial(struct wlr_seat_client *client) {
+	uint32_t serial = wl_display_next_serial(wl_client_get_display(client->client));
+	wlr_serial_add(&client->serials, serial);
+	return serial;
+}
--- a/types/seat/wlr_seat.c
+++ b/types/seat/wlr_seat.c
@@ -142,6 +142,9 @@ static void seat_handle_bind(struct wl_client *client, void *_wlr_seat,
 		wl_signal_init(&seat_client->events.destroy);
 
 		wl_list_insert(&wlr_seat->clients, &seat_client->link);
+
+		// ensure first entry will have index zero
+		seat_client->serials.end = WLR_SERIAL_RINGSET_SIZE - 1;
 	}
 
 	wl_resource_set_implementation(wl_resource, &seat_impl,
@@ -371,11 +374,13 @@ bool wlr_seat_validate_grab_serial(struct wlr_seat *seat, uint32_t serial) {
 	return true;
 }
 
-void wlr_serial_add(struct wlr_serial_ringset *set, uint32_t serial) {
+uint32_t wlr_seat_client_next_serial(struct wlr_seat_client *client) {
+	uint32_t serial = wl_display_next_serial(wl_client_get_display(client->client));
+	struct wlr_serial_ringset *set = &client->serials;
+
 	if (set->count == 0 || set->data[set->end].max_incl + 1 != serial) {
-		set->count++;
-		if (set->count > WLR_SERIAL_RINGSET_SIZE) {
-			set->count = WLR_SERIAL_RINGSET_SIZE;
+		if (set->count < WLR_SERIAL_RINGSET_SIZE) {
+			set->count++;
 		}
 		set->end = (set->end + 1) % WLR_SERIAL_RINGSET_SIZE;
 		set->data[set->end].min_incl = serial;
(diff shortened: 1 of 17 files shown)


types/seat/wlr_seat.c
```diff
@@ -370,3 +370,36 @@ bool wlr_seat_validate_grab_serial(struct wlr_seat *seat, uint32_t serial) {
 	//	serial == seat->touch_state.grab_serial;
 	return true;
 }
+
+void wlr_serial_add(struct wlr_serial_ringset *set, uint32_t serial) {
+	if (set->count == 0 || set->data[set->end].max_incl + 1 != serial) {
+		set->count++;
+		if (set->count > WLR_SERIAL_RINGSET_SIZE) {
```
arandomhuman:
```
if (set->count < WLR_SERIAL_RINGSET_SIZE) {
        set->count++;
}
```
is simpler, maybe?

types/seat/wlr_seat.c
```diff
@@ -370,3 +370,36 @@ bool wlr_seat_validate_grab_serial(struct wlr_seat *seat, uint32_t serial) {
...
+			set->count = WLR_SERIAL_RINGSET_SIZE;
+		}
+		set->end = (set->end + 1) % WLR_SERIAL_RINGSET_SIZE;
```
arandomhuman: Check for `set->count == 0` and don't increment, otherwise `set->data[0]` is unused on the first call.

types/seat/wlr_seat.c
```diff
@@ -370,3 +370,36 @@ bool wlr_seat_validate_grab_serial(struct wlr_seat *seat, uint32_t serial) {
...
+		set->data[set->end].max_incl = serial;
+	} else {
+		set->data[set->end].max_incl = serial;
+	}
+}
+
+bool wlr_serial_maybe_valid(struct wlr_serial_ringset *set, uint32_t serial) {
+	for (int i = 0; i < set->count; i++) {
+		int j = (set->end - i + WLR_SERIAL_RINGSET_SIZE) % WLR_SERIAL_RINGSET_SIZE;
+		if (set->data[j].max_incl - serial > UINT32_MAX / 2) {
```
arandomhuman: I don't get these checks; what's the problem with simply checking `min_incl <= serial <= max_incl`? (assuming no serial overflows)

As it stands right now, you can have false negatives (consider `data[0].min_incl = 1, data[0].max_incl = MAX/2 + 2, serial = 1` on L392) and false positives with very new values (consider `data[0].max_incl = 1, data[0].min_incl = 1, serial = MAX/2 + 3` on L398), which doesn't match the purpose of the function. If you want to count all serials received before the start as valid, maybe add an explicit check?

types/seat/wlr_seat.c
mstoeckl: Yes, fixed.

types/seat/wlr_seat.c
mstoeckl: Instead, I initialize set->end as `WLR_SERIAL_RINGSET_SIZE - 1`, which has the same effect.

types/seat/wlr_seat.c
mstoeckl: Serial overflows are still possible, so the code must be able to account for them. Also, as there were a few additional logic issues (sets without any entries admit all serials), I've replaced this function with something whose validation logic should not have any false positives.

mstoeckl:
I've added temporary commits to register event serials for pointer gestures and tablet actions. Could someone who uses the associated protocols check that the changes don't break anything?

types/seat/wlr_seat.c
arandomhuman: I would prefer this check, to be honest. If these functions are reused elsewhere, that tiny initialization step is liable to be missed. Plus, setting `end = SIZE - 1` has no real semantic sense. Deferred to devs.

mstoeckl: I've decided to split out the case `set->count == 0`, as doing so makes the logic a bit easier to follow.

mstoeckl:
It looks like [`wl-copy`](https://github.com/bugaevc/wl-clipboard) is broken with this patch, because the code sent by `wl_callback::done` after [`wl_display::sync`](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_display) is an event serial, and this PR doesn't register those at the moment.

emersion:
Wait, this kind of serial shouldn't be used for selection requests. Only keyboard enter serials can be used for this.

bugaevc:
> It looks like `wl-copy` is broken with this patch, because the code sent by `wl_callback::done` after `wl_display::sync` is the (current, not next) event serial, and this PR doesn't register those at the moment.

Note that in master wl-clipboard actually uses wlr-data-control and doesn't need this serial hack (when wlr-data-control is available, that is, and I still haven't gotten around to make it work for primary selection).

emersion:
Let's not support clients that rely on this hack.

ddevault:
Thank you!

emersion:
This implementation makes it possible to use _any_ seat serial for _any_ operation. This should be restricted, e.g. selection requests should only be accepted if the serial comes from a keyboard enter event.

ddevault:
Shit, I misinterpreted this:



Then I only gave it a brief review. Thanks, Github. However, this confuses me:

>selection requests should only be accepted if the serial comes from a keyboard enter event

Why?

emersion:
Whether a client has access to the clipboard depends on the keyboard focus.

ddevault:
Source?

emersion:
In the protocol:

>The selection event is sent to a client immediately before receiving keyboard focus and when a new selection is set while the client has keyboard focus. The data_offer is valid until a new data_offer or NULL is received or until the client loses keyboard focus.

Related: https://gitlab.freedesktop.org/wayland/wayland/issues/37

ddevault:
Okay. Do you want to revert or do it in a follow-up patch?

emersion:
This isn't that of a big issue, since we just want to add more constraints. Maybe this will require some refactoring. I'm fine with a follow-up issue/pull request, depending whether or not @mstoeckl is interested in continuing this work.

emersion:
Now that I think of it, we probably could only store the last keyboard enter serial, instead of having a ring buffer.

Other follow-up issue: do the same for drag-and-drop.

ddevault:
>we probably could only store the last keyboard enter serial

Is this racey

emersion:
Probably not. The only case in which a race could happen is if the user copies something and switches focus at the same time very quickly before the client has sent the `set_selection` request. But accepting such a `set_selection` request would also allow clients to keep control of the clipboard longer than they are supposed to.

Will think about it a little bit more anyway.

mstoeckl:
>  I'm fine with a follow-up issue/pull request, depending whether or not @mstoeckl is interested in continuing this work.

Not at the moment, as I don't see much benefit to restricting the right to make copy selections further, especially since the Wayland protocol issue you mentioned is still active and may lead to spec changes.

Also (reading over the protocol text for the seventh time), the quote from the protocol about wl_data_device::selection indicates when a client is allowed to paste/receive data; the core protocol says nothing about restrictions on copying, so it's up to the compositor to decide how that should be done.

(Side question: are there any window managers which allow mouse interactions with one window while the keyboard focus is kept on another?)

emersion:
Example of possible malicious client with the serial validation implemented in this PR: a client that controls the clipboard when the cursor enters on of its surfaces.

mstoeckl:
The malicious client example is a possibility, but things like hover-to-copy could be relatively useful features to have. Could be -- as with many things, it's hard to tell without trying it. (On the other hand, making a copy request without *any* user interaction probably isn't.)

The race scenario sounds pretty likely; applications that sometimes take 100msec to respond to events are still unfortunately common, and that's enough time to e.g. double-click and leave for a primary-selection copy, or have the pointer cross over a series of small floating windows. One could instead adjust the ring buffer to record start and end serials for the interval in which keyboard focus was held.

emersion:
>hover-to-copy

I'm not aware of anybody doing this, and I'm not sure its a good idea to support it.

>The race scenario sounds pretty likely

Not sure about this. In any case, a ring buffer is overkill, keeping the last serial is enough.

PR #1743, opened 2019-06-30
Reject copy-selection requests with future serials

mstoeckl:
See also commit message.

To test, try programs like termite (which makes multiple selection requests with the same serial value), or modify some simple client to make wl_data_device::set_selection requests with a serial number that is much larger than it should be.


ddevault:
Please rebase this and address merge conflicts

emersion:
Now that #1742 is merged, this one can probably be closed.

PR #1748, opened 2019-07-10
Handle stacks (xdg_toplevel::set_parent relationships)

### LINKED ISSUE #373: Transcient windows hidden under their parent
1. Open Gedit
2. Write something
3. Close Gedit, this shows a dialog that asks to save before closing
4. Click on the parent
5. You end up in a state where the parent is focused, but you can't move it because it has a child, and the child is hidden under the parent. Pointer events are not sent to the parent, so you can't do anything.

dos1:
So far, rootston simply ignored the parent-child relationships as set by `xdg_toplevel::set_parent`. This PR implements the behavior specified by the protocol, so the child surfaces stay on top of their parents.

I've renamed `children` member to `child_surfaces` in order to try to make it a bit less confusing, as it has nothing to do with a member called `parent`.

Closes #373

### DIFF
--- a/include/rootston/view.h
+++ b/include/rootston/view.h
@@ -38,6 +38,7 @@ struct roots_view {
 	const struct roots_view_interface *impl;
 	struct roots_desktop *desktop;
 	struct wl_list link; // roots_desktop::views
+	struct wl_list parent_link; // roots_view::stack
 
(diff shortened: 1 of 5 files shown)


ammen99:
I guess this should be closed since rootston isn't in this repo anymore?

PR #1753, opened 2019-07-12
xdg_shell(_v6): Take maximize/fullscreen state into account on view init

dos1:
set_maximized and set_fullscreen calls can come before the view is
constructed and before its signal handlers are registered.


emersion:
Thanks!

PR #1754, opened 2019-07-14
backend/wayland: Send a frame event on the pointer after pointer events

dos1:
X11 backend does it already and without it, the output has issues
with updating when it should.


dos1:
@emersion Hmm, as far as I can see, it's already done by `pointer_handle_frame`. I'm having severe window refresh issues with KWin though and this patch helps a bit. Any idea what could be wrong there?

emersion:
Maybe KWin isn't sending frame events?

dos1:
Hah, of course - it does seem to be the case when using a touch screen with a client that doesn't register to wl_touch! :)

PR #1760, opened 2019-07-18
Use -fmacro-prefix-map to strip build path, when available

mstoeckl:
This PR applies the same change as https://github.com/swaywm/sway/pull/4282.

I have tested this change with Clang (8.0.0) and GCC (9.1.0, has `-fmacro-prefix-map`), with `--buildtype release`, and with wlroots built both standalone and as a subproject of sway.

To check that `-fmacro-prefix-map` actually strips paths correctly, run `strings libwlroots.so | grep wlroots`.


ddevault:
Thanks!

PR #1761, opened 2019-07-18
xdg-output: add support for xdg-output-unstable-v1 version 3

emersion:
This adds support for xdg-output-unstable-v1 version 3, added in [1].

The xdg_output.done event is now deprecated and is replaced with
wl_output.done.

This PR has been tested with [2].

[1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/commit/962dd535372c8e4681374c23d2603cbe06cd7031
[2]: https://gitlab.freedesktop.org/xorg/xserver/merge_requests/234


ddevault:
Thanks!

Issue #1640, opened 2019-03-28
Atomic output updates


emersion:
This would let us:

* Enable and disable multiple outputs without the fear of hitting the CRTC limit
* Test configurations before applying them
* Rollback if something fails
* More cool stuff

https://github.com/swaywm/wlr-protocols/pull/38#issuecomment-470472049

emersion:
API proposal:

* Make `wlr_output_set_*` functions double-buffered
* Add `wlr_output_commit` to apply the pending state
* Add `wlr_backend_commit` to apply the pending state for all outputs. Backends can choose not to implement this, in which case the function will fallback to calling `wlr_output_commit` on each output

emersion:
Items 1, 2 and 4 are now done.

PR #1762, opened 2019-07-18
output: atomic mode, enabled, scale and transform

emersion:
This commit makes more output properties (mode, enabled, scale and transform)
atomic. This means that they are double-buffered and only applied on commit.

Compositors now need to call wlr_output_commit after setting any of those
properties.

Internally, backends still apply properties sequentially. The behaviour should
be exactly the same as before. Future commits will update some backends to take
advantage of the atomic interface. Some backends are non-atomic by design, e.g.
the X11 backend or the legacy DRM backend.

Updates: https://github.com/swaywm/wlroots/issues/1640

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -328,12 +328,9 @@ static bool drm_connector_attach_render(struct wlr_output *output,
 	return make_drm_surface_current(&conn->crtc->primary->surf, buffer_age);
 }
 
-static bool drm_connector_commit(struct wlr_output *output) {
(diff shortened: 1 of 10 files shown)


ddevault:
Do you have a corresponding sway patch? I'd like to test drive this for a while before merging it.

emersion:
Good idea. Implemented here: https://github.com/swaywm/sway/pull/4355

ddevault:
Thanks!

PR #1763, opened 2019-07-19
output: drop wlr_output_mode.flags

emersion:
AFAIK this was always set to zero. Instead, compute wl_output mode flags on the
fly.

Technically this is a breaking change, but I don't think anybody uses this
field.


ddevault:
Thanks!

PR #1771, opened 2019-07-26
Introduce wlr_keyboard_group

### LINKED ISSUE #1217: Merge multiple devices into one keyboard
Context: my keyboard has an n-key rollover feature which makes it register itself as multiple HID devices. Each of these is interpreted as an individual keyboard by wlroots compositors. 

Consequently, modifiers are only sent if they belong to the same input device as the key being pressed. Grouping multiple devices into one keyboard would allow sharing these modifiers (as well as keymap and repeat rate), 

The `libinput list-devices` command indicates that these devices all belong to the same group, which might be a good way to merge keyboards by default.
```
Device:           Keyed Up Labs ES-87
Kernel:           /dev/input/event5
Group:            5
...
Device:           Keyed Up Labs ES-87 Keyboard
Kernel:           /dev/input/event6
Group:            5
...
Device:           Keyed Up Labs ES-87 Consumer Control
Kernel:           /dev/input/event7
Group:            5
```

### LINKED ISSUE #1421: Separate keymap from wlr_keyboard
In PR #1420 I proposed to add a state parameter to the `wlr_keyboard_set_keymap` function, but I think a overall better change would be to separate the keymap from the keyboard.

I think this would be more future proof, if `wl_keyboard` starts to support more keymap formats and would allow to use the `wl_keyboard` `keymap_format.no_keymap` format for devices that should send raw keycodes to the client.

The compositor decides if a the keyboards share the keymap or create a keymap for each keyboard to keep the current behavior.
If keyboards share the same keymap the modifiers will be shared too.

```
struct wlr_keymap_impl {
        void (*get_leds)(struct wlr_keymap *keymap);
        uint32_t (*get_modifiers)(struct wlr_keymap *keymap);
        void (*key_update)(struct wlr_keymap *keymap,
                        struct wlr_event_keyboard_key *event);
        bool (*modifier_update)(struct wlr_keymap *keymap);
        void (*state_update)(struct wlr_keymap *keymap,
                        struct wlr_event_keyboard_key *event);
};

struct wlr_keymap {
        struct wlr_keymap_impl *impl;
        void *keymap_buf;
        size_t keymap_size;
        enum wl_keyboard_keymap_format keymap_format;
}

strutc wlr_keyboard {
        [...]
        wlr_keymap *keymap;
        [...]
}

struct wlr_keymap *keymap =  wlr_keymap_xkb_create(xkb_keymap);
keyboard_set_keymap(keyboard1, keymap);
keyboard_set_keymap(keyboard2, keymap);
```

RedSoxFan:
Closes #1217 
Supersedes/Closes #1421

Sway PR: swaywm/sway#4373

A wlr_keyboard_group allows for multiple keyboard devices to be
combined into one logical keyboard. Each keyboard device can only be
added to one keyboard group. This helps with the situation where one
physical keyboard is exposed as multiple keyboard devices. It is up to
the compositors on how they group keyboards together, if at all.

Since a wlr_keyboard_group is one logical keyboard, the keys are a set.
This means that if a key is pressed on multiple keyboard devices, the
key event will only be emitted once, but the internal state will count
the number of devices that the key is pressed on. Likewise, the key
release will not be emitted until the key is released from all devices.
If the compositor wants access to which keys are pressed and released
on each keyboard device, the events for those devices can be listened
to, as they currently are, in addition to the group keyboard's events.

Also, all keyboard devices in the group must share the same keymap. If
the keymap's differ, the keyboard device will not be able to be added
to the group. Once in the group, if the keymap or effective layout for
one keyboard device changes, it will be synced to all keyboard devices
in the group. The repeat info is also synced.

### DIFF
--- a/include/wlr/types/wlr_keyboard.h
+++ b/include/wlr/types/wlr_keyboard.h
@@ -49,6 +49,7 @@ struct wlr_keyboard_modifiers {
 
 struct wlr_keyboard {
 	const struct wlr_keyboard_impl *impl;
+	struct wlr_keyboard_group *group;
 
 	char *keymap_string;
 	size_t keymap_size;
@@ -84,6 +85,7 @@ struct wlr_keyboard {
 		struct wl_signal modifiers;
 		struct wl_signal keymap;
 		struct wl_signal repeat_info;
+		struct wl_signal destroy;
 	} events;
 
 	void *data;
--- a/include/wlr/types/wlr_keyboard_group.h
+++ b/include/wlr/types/wlr_keyboard_group.h
@@ -1,0 +1,37 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_KEYBOARD_GROUP_H
+#define WLR_TYPES_WLR_KEYBOARD_GROUP_H
+
+#include <wayland-server-core.h>
+#include "wlr/types/wlr_keyboard.h"
+#include "wlr/types/wlr_input_device.h"
+
+struct wlr_keyboard_group {
+	struct wlr_keyboard keyboard;
+	struct wlr_input_device *input_device;
+	struct wl_list devices; // keyboard_group_device::link
+	struct wl_list keys; // keyboard_group_key::link
+	void *data;
+};
+
+struct wlr_keyboard_group *wlr_keyboard_group_create(void);
+
+struct wlr_keyboard_group *wlr_keyboard_group_from_wlr_keyboard(
+		struct wlr_keyboard *keyboard);
+
+bool wlr_keyboard_group_add_keyboard(struct wlr_keyboard_group *group,
+		struct wlr_keyboard *keyboard);
+
+void wlr_keyboard_group_remove_keyboard(struct wlr_keyboard_group *group,
+		struct wlr_keyboard *keyboard);
+
+void wlr_keyboard_group_destroy(struct wlr_keyboard_group *group);
+
+#endif
--- a/types/wlr_keyboard_group.c
+++ b/types/wlr_keyboard_group.c
@@ -1,0 +1,317 @@
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <time.h>
+#include <wayland-server-core.h>
+#include <xkbcommon/xkbcommon.h>
+#include "wlr/interfaces/wlr_keyboard.h"
+#include "wlr/types/wlr_keyboard.h"
+#include "wlr/types/wlr_keyboard_group.h"
+#include "wlr/util/log.h"
+
+struct keyboard_group_device {
+	struct wlr_keyboard *keyboard;
+	struct wl_listener key;
+	struct wl_listener modifiers;
+	struct wl_listener keymap;
+	struct wl_listener repeat_info;
+	struct wl_listener destroy;
+	struct wl_list link; // wlr_keyboard_group::devices
+};
+
+struct keyboard_group_key {
+	uint32_t keycode;
+	size_t count;
+	struct wl_list link; // wlr_keyboard_group::keys
+};
+
+static void keyboard_set_leds(struct wlr_keyboard *kb, uint32_t leds) {
+	struct wlr_keyboard_group *group = wlr_keyboard_group_from_wlr_keyboard(kb);
+	struct keyboard_group_device *device;
+	wl_list_for_each(device, &group->devices, link) {
+		wlr_keyboard_led_update(device->keyboard, leds);
+	}
+}
+
+static void keyboard_destroy(struct wlr_keyboard *kb) {
+	// Just remove the event listeners. The keyboard will be freed as part of
+	// the wlr_keyboard_group in wlr_keyboard_group_destroy.
+	wl_list_remove(&kb->events.key.listener_list);
+	wl_list_remove(&kb->events.modifiers.listener_list);
+	wl_list_remove(&kb->events.keymap.listener_list);
+	wl_list_remove(&kb->events.repeat_info.listener_list);
+	wl_list_remove(&kb->events.destroy.listener_list);
+}
+
+static const struct wlr_keyboard_impl impl = {
+	.destroy = keyboard_destroy,
+	.led_update = keyboard_set_leds
+};
+
+struct wlr_keyboard_group *wlr_keyboard_group_create(void) {
+	struct wlr_keyboard_group *group =
+		calloc(1, sizeof(struct wlr_keyboard_group));
+	if (!group) {
+		wlr_log(WLR_ERROR, "Failed to allocate wlr_keyboard_group");
+		return NULL;
+	}
+
+	group->input_device = calloc(1, sizeof(struct wlr_input_device));
+	if (!group->input_device) {
+		wlr_log(WLR_ERROR, "Failed to allocate wlr_input_device for group");
+		free(group);
+		return NULL;
+	}
+	wl_signal_init(&group->input_device->events.destroy);
+	group->input_device->keyboard = &group->keyboard;
+
+	wlr_keyboard_init(&group->keyboard, &impl);
+	wl_list_init(&group->devices);
+	wl_list_init(&group->keys);
+
+	return group;
+}
+
+struct wlr_keyboard_group *wlr_keyboard_group_from_wlr_keyboard(
+		struct wlr_keyboard *keyboard) {
+	if (keyboard->impl != &impl) {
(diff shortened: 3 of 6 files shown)


emersion:
This approach won't cut it: some keyboards are split in two, one having modifiers and the other regular keys. If the xkb context isn't shared, modifiers won't work correctly.

RedSoxFan:
> This approach won't cut it: some keyboards are split in two, one having modifiers and the other regular keys. If the xkb context isn't shared, modifiers won't work correctly.

The keyboard group has a wlr_keyboard. A keymap has to be set for the group's keyboard before any keyboards can be added the group. If the keymap for any of the keyboards in the group change, it is synced to all keyboards in the group, including the group's keyboard. The group listens for events from the keyboards in the group and calls `wlr_keyboard_notify_{key,modifier}` functions to update it's own state.

This approach allows for keyboards to be used identical to how they currently are, for device specific bindings, as well as being viewed as one.

For example, my laptop has XF86MonBrightness{Down,Up} on `0:6:Video_Bus` and the modifiers on `1:1:AT_Translated_Set_2_keyboard`. With this PR and the sway one, bindings combos like `Shift+XF86MonBrightnessDown` and `Mod4+XF86MonBrightnessUp+Space`  works, whereas without they do not as they are on separate input devices.

RedSoxFan:
> The group listens for events from the keyboards in the group and calls `wlr_keyboard_notify_{key,modifier}` functions to update it's own state.

Actually, I just pushed a change to make it so the modifiers for the wlr_keyboard_group are derived from the keys in it's xkb_state instead of being an OR of the modifiers from each keyboard in the group. The effective layout (group modifier) is still being synced across keyboards in the group in `handle_keyboard_modifiers`. This is also propagated to the wlr_keyboard_group's keyboard since it cannot always be derived from the xkb_state (ie sway's `xkb_switch_layout`).

b0o:
@RedSoxFan Have you had a chance to look into this further? This is an issue I've had with sway for months now, I would love to see it fixed.

RedSoxFan:
@b0o This is just waiting for someone willing to review. I know that @emersion has made some comments on IRC about not really liking the implementation, but I'm not really sure of a better one. I think this provides the compositor with the most flexibility and doesn't break any existing functionality.

It looks like both this PR and the sway PR can be cleanly rebased on top of their respective master, if you wanted to use them now and provide feedback.

emersion:
After thinking about it, this approach makes sense to me in the end. It acts like a keyboard multiplexer. The only thing I'm not 100% happy about is that we have unused xkb contexts (in the physical keyboards).

include/wlr/types/wlr_keyboard_group.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <wayland-server-core.h>
+#include "wlr/types/wlr_keyboard.h"
+#include "wlr/types/wlr_input_device.h"
+
+struct wlr_keyboard_group {
+	struct wlr_keyboard keyboard;
+	struct wlr_input_device *input_device;
+	struct wl_list devices; // keyboard_group_device::link
+	struct wl_list keys; // keyboard_group_key::link
```
emersion: Hmm. We already have logic for this in the seat I think.

types/wlr_keyboard_group.c
```diff
@@ -0,0 +1,299 @@
...
+#include "wlr/types/wlr_keyboard.h"
+#include "wlr/types/wlr_keyboard_group.h"
+#include "wlr/util/log.h"
+
+struct keyboard_group_device {
+	struct wlr_keyboard *keyboard;
+	struct wl_listener key;
+	struct wl_listener modifiers;
+	struct wl_listener keymap;
+	struct wl_listener repeat_info;
```
emersion: We should add a destroy listener in there too.

include/wlr/types/wlr_keyboard_group.h
RedSoxFan: It's possible to have multiple keyboard groups per seat. That's up to the compositor. For example, in my sway PR, there is the ability to group by keymap. Say there are four keyboard devices than span across two keymaps, there would be two keyboard groups for the seat

RedSoxFan:
> The only thing I'm not 100% happy about is that we have unused xkb contexts (in the physical keyboards).

They would still be used for features like sway's `--input-device` bindings. Each keyboard still needs to retain it's own state regardless in case it gets removed from the group

b0o:
@RedSoxFan I tested your branches merged onto the latest sway/wlroots masters. Launching sway, my keyboard is unresponsive; upon unplugging the keyboard, sway crashes. See attached debug output.
[swayout.log](https://github.com/swaywm/wlroots/files/3746789/swayout.log)

RedSoxFan:
@b0o Should be fixed now

types/wlr_keyboard_group.c
RedSoxFan: Added. `wlr_keyboard` didn't have a `destroy` signal so I had to add it

ddevault:
@emersion?

include/wlr/types/wlr_keyboard_group.h
emersion: Yeah. My question is: why do we need this extra state? I doesn't seem used anywhere here.

types/wlr_keyboard_group.c
```diff
@@ -0,0 +1,325 @@
...
+	struct wlr_keyboard *keyboard = group_device->keyboard;
+
+	struct keyboard_group_device *device;
+	wl_list_for_each(device, &keyboard->group->devices, link) {
+		struct wlr_keyboard *devkb = device->keyboard;
+		if (devkb->repeat_info.rate != keyboard->repeat_info.rate ||
+				devkb->repeat_info.delay != keyboard->repeat_info.delay) {
+			wlr_keyboard_set_repeat_info(devkb, keyboard->repeat_info.rate,
+					keyboard->repeat_info.delay);
+			return;
```
emersion: Why don't we set the repeat info to all keyboards in the group? Why do we stop here?

types/wlr_keyboard_group.c
```diff
@@ -0,0 +1,325 @@
...
+static void handle_keyboard_repeat_info(struct wl_listener *listener,
+		void *data) {
+	struct keyboard_group_device *group_device =
+		wl_container_of(listener, group_device, repeat_info);
+	struct wlr_keyboard *keyboard = group_device->keyboard;
+
+	struct keyboard_group_device *device;
+	wl_list_for_each(device, &keyboard->group->devices, link) {
+		struct wlr_keyboard *devkb = device->keyboard;
+		if (devkb->repeat_info.rate != keyboard->repeat_info.rate ||
```
emersion: This check should probably be in `wlr_keyboard_set_repeat_info` instead.

include/wlr/types/wlr_keyboard.h
```diff
@@ -110,6 +112,8 @@ void wlr_keyboard_set_keymap(struct wlr_keyboard *kb,
 void wlr_keyboard_set_repeat_info(struct wlr_keyboard *kb, int32_t rate,
 	int32_t delay);
 void wlr_keyboard_led_update(struct wlr_keyboard *keyboard, uint32_t leds);
+void wlr_keyboard_led_update_from_xkb_state(struct wlr_keyboard *keyboard,
+		struct xkb_state *xkb_state);
```
emersion: Not a fan of this. Can't we sync the modifiers on all keyboards in the group instead? So that enabling Caps Lock on one keyboard will enable it on all others too?

types/wlr_keyboard_group.c
RedSoxFan: `wlr_keyboard_set_repeat_info` will emit the `repeat_info` event for the keyboard which has this function as a listener. So all keyboards will be changed recursively and then once they are all changed, the `wlr_keyboard_group`'s keyboard will be changed and its `repeat_info` event will be emitted

types/wlr_keyboard_group.c
RedSoxFan: This is used to sync the repeat info to the other keyboards in the group so I think it makes more sense here

include/wlr/types/wlr_keyboard.h
RedSoxFan: I originally had it that way, but changed it. I don't remember why. I'll look into it again

include/wlr/types/wlr_keyboard_group.h
RedSoxFan: I don't think `util/array.h` is sufficient because we need a counter for the number of keyboard where the key is currently pressed The key press for the group should occur on the first key press, but the key release should not occur until ALL keyboards have released the key. Simply adding the key code to a 1-dimensional uint32_t array/set does not allow for this to happen.

include/wlr/types/wlr_keyboard.h
RedSoxFan: Changed it to sync the modifiers

include/wlr/types/wlr_keyboard_group.h
emersion: Ah, you're right!

emersion:
Thanks!

PR #1779, opened 2019-08-02
backend/drm: destroy output immediately

### LINKED ISSUE #1739: Destroy DRM outputs immediately
We have some logic to only destroy an output if a pageflip isn't pending. Now that https://github.com/swaywm/wlroots/pull/1738 is merged we don't need it anymore.

Emantor:
Instead of waiting for the next pageflip, destroy the output immediately
since we can now handle flips for outputs which no longer exist.
Also demote the missing crtc on flip to debug.

Fixes #1739

Creating as draft since I have not tested this yet.


Emantor:
Tested and works.

emersion:
Thanks!

PR #1781, opened 2019-08-07
Revert "output: atomic mode"

Emantor:
This reverts commit ee5f98ad49fed0439f3313ec685307831d1d1d05.

This intoduced problems where outputs could not be turned off because
they had flips pending.


ddevault:
Thanks!

PR #1782, opened 2019-08-07
wlr_input_method_v2: Remove input method's resource from the list on destroy

dos1:
It's added to manager->input_methods list in manager_get_input_method, but wasn't removed anywhere, leading to possible use-after-free in wlr_input_method_manager_v2_destroy.


dcz-purism:
Looks ok, I'll be using it over the next few days to see if the bug is still triggered anywhere.

ddevault:
Thanks! And to @dcz-purism as well for testing.

Issue #933, opened 2018-04-30
xdg_popups with grab not closed due to touch events


agx:
I only reproduced this with an xdg_popup that's parent is  layer_surface so far:

When pressing the touch screen outside the popup and the layer_surface the popup is supposed to close, this does not happen (while clicking with a mouse outside the xdg_popup correctly closes it).

Issue #897, opened 2018-04-24
layer-shell popups are not closed when clicking on other layer-surface (e.g. background layer) but should


agx:
If you have e.g. a top panel that uses popups (with grab) for menus and you click outside the popus and panel the menu closes (as it should). This does not work if one clicks on another layer surface (e.g. ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND) that is being clicked on (which makes using layer-shell for backgrounds and panels at the same time impossible). 

I have yet to figure out if we want to make ZWLR_LAYER_SHELL_V1_LAYER_BACKGROUND special or only always close popups when clicked outside the parent layer-shell and the popup.

This can e.g. be checked with phosh's layer-surfaces branch (https://code.puri.sm/guido.gunther/phosh/src/layer-surfaces). Comment out create_background() in phosh_shell_constructed() and closing of popups works as expected otherwise popups won't close since the background spans the whole output.

Opening a new issue instead of cluttering https://github.com/swaywm/wlroots/issues/889

SirCmpwn:
I don't fully understand the problem. If you click on another surface, background or not, we should close the grab.

agx:
...and that's not happening. I'll have  look.

agx:
The reason is that the pointer grab is client based so if the pointer moves out of the popup onto the background layer-surface ```wlr_seat_pointer_send_button()``` will determine that the same client is still focues and therefore not close the popup. If one moves the pointer of the surface of another client the popup is correctly closed.

SirCmpwn:
You're elaborating on the causes of the bug, rather than on why you believe it's the intended behavior, right?

agx:
Right.

PR #1783, opened 2019-08-08
Input grab fixes for xdg_popups

dos1:
Fixes for #897 and #933


dos1:
Added the same changes to xdg_shell_v6.

ddevault:
LGTM, thanks!

PR #1784, opened 2019-08-08
rootston: cursor: call pointer_notify_enter before pointer_notify_button

dos1:
This fixes issues with pointer hover not being registered after exiting xdg_popup grab - without this, the window underneath doesn't notice the pointer (and button clicks) until it gets slightly moved.


ammen99:
I think we should rather update the cursor focus when a surface gets unmapped or mapped.

ddevault:
Rootston is being moved here:

https://github.com/swaywm/rootston

PR #1785, opened 2019-08-08
desktop_surface_at: Set a "surface" argument in case of fullscreen surface

dos1:


ddevault:
Rootston is being moved here:

https://github.com/swaywm/rootston

PR #1787, opened 2019-08-10
Patches

MisterDA:
Hello,
This is a bunch of patches for bugs reported by [PVS-Studio](https://www.viva64.com/en/pvs-studio/).

If realloc fails, the whole array is lost.
cd1e96e8: xcursor: avoid leak and loss of all cursors if cursors realloc fails
b894c55c: render/drm: keep old drm_format if realloc fails

Remove dead code and trivial fixes.
cd91349c: backend/session: non-void function should return a value
364c7913: xdg_shell: remove variable self-assignment
af02f97a: wlr_surface: condition is always false
a1c2bc22: Simplify check
96fa3c2e: Fix memory leak

There are also around 20 reports for `calloc`, `strdup` calls that are not checked for possible null returns. I don’t know how serious you are about enforcing this kind of checks, so I did not make a patch for it.

See inline for comments on the other commits.

PVS-Studio also reports two errors that I'm not sure how to fix:
- There are identical sub-expressions  `(unsigned) current.preedit.cursor_begin > strlen(preedit_text)` to  the left and to the right of the `||` operator.
  https://github.com/swaywm/wlroots/blob/58b2584863201f32d1a4770a1d5cde5e57b300c6/examples/text-input.c#L140-L141
  *Edit: fixed in baffb90: examples: remove duplicated condition*
 
- The `surface` pointer was utilized before it was verified against nullptr. Check lines: 272, 280.
  https://github.com/swaywm/wlroots/blob/58b2584863201f32d1a4770a1d5cde5e57b300c6/types/seat/wlr_seat_keyboard.c#L272

Thanks for the great work!

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -7,6 +7,7 @@
 #include <GLES2/gl2.h>
 #include <GLES2/gl2ext.h>
 #include <inttypes.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -1427,7 +1428,7 @@ int handle_drm_event(int fd, uint32_t mask, void *data) {
 }
 
 void restore_drm_outputs(struct wlr_drm_backend *drm) {
-	uint64_t to_close = (1L << wl_list_length(&drm->outputs)) - 1;
+	uint64_t to_close = (UINT64_C(1) << wl_list_length(&drm->outputs)) - 1;
 
 	struct wlr_drm_connector *conn;
 	wl_list_for_each(conn, &drm->outputs, link) {
@@ -1444,7 +1445,7 @@ void restore_drm_outputs(struct wlr_drm_backend *drm) {
 		struct wlr_drm_connector *conn;
 		wl_list_for_each(conn, &drm->outputs, link) {
 			if (conn->state != WLR_DRM_CONN_CLEANUP || !conn->pageflip_pending) {
-				to_close &= ~(1 << i);
+				to_close &= ~(UINT64_C(1) << i);
 			}
 			i++;
 		}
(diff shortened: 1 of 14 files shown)


Emantor:
>     * The `surface` pointer was utilized before it was verified against nullptr. Check lines: 272, 280.
>       https://github.com/swaywm/wlroots/blob/58b2584863201f32d1a4770a1d5cde5e57b300c6/types/seat/wlr_seat_keyboard.c#L272

`client` is only set if `surface` is set, this is indeed checked indirectly.

Emantor:
>     * There are identical sub-expressions  `(unsigned) current.preedit.cursor_begin > strlen(preedit_text)` to  the left and to the right of the `||` operator.
>       https://github.com/swaywm/wlroots/blob/58b2584863201f32d1a4770a1d5cde5e57b300c6/examples/text-input.c#L140-L141
The solution is to remove one of the sub expressions. Git says that has been there from the beginning, should be fine to clean up.

MisterDA:
> The assert is hit before this return.

Not if asserts are disabled at compile-time with the `NDEBUG` macro.

dos1:
> > ```
> > * There are identical sub-expressions  `(unsigned) current.preedit.cursor_begin > strlen(preedit_text)` to  the left and to the right of the `||` operator.
> >   https://github.com/swaywm/wlroots/blob/58b2584863201f32d1a4770a1d5cde5e57b300c6/examples/text-input.c#L140-L141
> > ```
> 
> The solution is to remove one of the sub expressions. Git says that has been there from the beginning, should be fine to clean up.

I believe one of these expressions should check `cursor_end` instead of `cursor_begin`. @dcz-purism?

ddevault:
Thank you!

PR #1792, opened 2019-08-14
screencopy: Implement copy_with_damage

any1:


ddevault:
Can you rebase this against master?

types/wlr_screencopy_v1.c
```diff
@@ -10,8 +10,81 @@
...
+		if (damage->output == output) {
+			return damage;
+		}
+	}
+
+	return NULL;
+}
+
+static void accumulate_damage(struct screencopy_damage *damage)
+{
```
ddevault: style

types/wlr_screencopy_v1.c
```diff
@@ -10,8 +10,81 @@
...
+	pixman_region32_intersect_rect(region, region, 0, 0, output->width,
+		output->height);
+}
+
+static void damage_handle_output_precommit(struct wl_listener *listener,
+		void *data) {
+	struct screencopy_damage *damage =
+		wl_container_of(listener, damage, output_precommit);
+	if (!damage->is_handled)
+		accumulate_damage(damage);
```
ddevault: style

types/wlr_screencopy_v1.c
```diff
@@ -10,8 +10,81 @@
...
+}
+
+static struct screencopy_damage *new_output_damage(
+		struct wlr_screencopy_manager_v1 *manager,
+		struct wlr_output *output) {
+	struct screencopy_damage *damage = calloc(1, sizeof(*damage));
+	if (!damage)
+		return NULL;
```
ddevault: style

types/wlr_screencopy_v1.c
```diff
@@ -10,8 +10,81 @@
...
+	struct screencopy_damage *damage =
+		wl_container_of(listener, damage, output_precommit);
+	if (!damage->is_handled)
+		accumulate_damage(damage);
+}
+
+static struct screencopy_damage *new_output_damage(
+		struct wlr_screencopy_manager_v1 *manager,
+		struct wlr_output *output) {
+	struct screencopy_damage *damage = calloc(1, sizeof(*damage));
```
ddevault: We prefer `sizeof(struct screencopy_damage)`

types/wlr_screencopy_v1.c
```diff
@@ -10,8 +10,81 @@
...
+
+static struct screencopy_damage *get_screencopy_damage(
+		struct wlr_screencopy_manager_v1 *manager,
+		struct wlr_output *output) {
+	struct screencopy_damage *damage = find_output_damage(manager, output);
+	return damage ? damage : new_output_damage(manager, output);
+}
+
+static void destroy_screencopy_damage(struct screencopy_damage *damage)
+{
```
ddevault: style

types/wlr_screencopy_v1.c
```diff
@@ -52,8 +125,25 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
...
+
+		/* In case precommit hook is called before the damage precommit
+		 * hook, the damage precommit hook must know that this has been
+		 * handled. This variable is reset on the commit event.
+		 */
+		accumulate_damage(damage);
+		damage->is_handled = true;
+
+		if (!pixman_region32_not_empty(&damage->damage))
+			return;
```
ddevault: style

types/wlr_screencopy_v1.c
```diff
@@ -90,6 +197,19 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
...
+static void frame_handle_output_commit(struct wl_listener *listener,
+		void *data) {
+	struct wlr_screencopy_frame_v1 *frame =
+		wl_container_of(listener, frame, output_commit);
+
+	if (frame->with_damage) {
+		struct screencopy_damage *damage =
+			find_output_damage(frame->manager, frame->output);
+		if (damage)
+			damage->is_handled = false;
```
ddevault: style

ddevault:
LGTM

types/wlr_screencopy_v1.c
```diff
@@ -10,8 +10,82 @@
...
+static void damage_handle_output_precommit(struct wl_listener *listener,
+		void *data) {
+	struct screencopy_damage *damage =
+		wl_container_of(listener, damage, output_precommit);
+	if (!damage->is_handled) {
+		accumulate_damage(damage);
+	}
+}
+
+static struct screencopy_damage *new_output_damage(
```
emersion: This should be named `screencopy_damage_create`

types/wlr_screencopy_v1.c
```diff
@@ -52,8 +126,26 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
 		return;
 	}
 
+	struct screencopy_damage *damage = NULL;
+	if (frame->with_damage) {
+		damage = get_screencopy_damage(frame->manager, output);
```
emersion: This creates one global `screencopy_damage` per output. This breaks when there are multiple clients capturing the same output.

types/wlr_screencopy_v1.c
emersion: One `screencopy_damage` is created when a client starts capturing, but isn't free'ed until the compositor shuts down. This leaks resources.

any1: It is freed `in wlr_screencopy_manager_v1_destroy()`. Isn't that called when the client destroys its manager object or disconnects?
This is used for the per-client damage accumulation.

emersion: No, `wlr_screencopy_manager` is not per-client. It's created once by the compositor for its whole lifetime.

any1: Ahh, right, it needs to go into `manager_handle_resource_destroy()`.

any1:
I noticed that the manager keeps a list of frames so that they don't fall between ship and harbour. Wouldn't it make sense to move that list into the newly created connection object so that they're sure to get destroyed along with the client that requested them? I could do that in a separate PR if you like.

emersion:
>Wouldn't it make sense to move that list into the newly created connection object so that they're sure to get destroyed along with the client that requested them? I could do that in a separate PR if you like.

Err, I think a per-client struct is overkill. Just having an extra `wl_resource` field which contains the manager resource would be enough.

any1:
> Hmm, I think a per-client struct is overkill. Just having an extra wl_resource field which contains the manager resource would be enough.

I see, the frames are managed my the resource system.

include/wlr/types/wlr_screencopy_v1.h
```diff
@@ -27,9 +27,14 @@ struct wlr_screencopy_manager_v1 {
 	void *data;
 };
 
+struct wlr_screencopy_connection_v1 {
```
emersion: Rename to `wlr_screencopy_v1_client` for consistency with other places in wlroots where we do this.

include/wlr/types/wlr_screencopy_v1.h
```diff
@@ -27,9 +27,14 @@ struct wlr_screencopy_manager_v1 {
...
+	struct wlr_screencopy_manager_v1* manager;
```
emersion: Styke: `*` on the right

types/wlr_screencopy_v1.c
```diff
@@ -8,10 +8,85 @@
...
 #include "util/signal.h"
 
-#define SCREENCOPY_MANAGER_VERSION 1
+#define SCREENCOPY_MANAGER_VERSION 2
+
+struct screencopy_damage {
+	struct wl_list link;
+	struct wlr_output *output;
+	struct pixman_region32 damage;
+	struct wl_listener output_precommit;
```
emersion: We need an output destroy listener here as well, in case the output is removed while recording.

types/wlr_screencopy_v1.c
```diff
@@ -8,10 +8,85 @@
...
+	wl_list_insert(&conn->damages, &damage->link);
+
+	wl_signal_add(&output->events.precommit, &damage->output_precommit);
+	damage->output_precommit.notify =
+		screencopy_damage_handle_output_precommit;
+
+	return damage;
+}
+
+static struct screencopy_damage *screencopy_damage_get(
```
emersion: Nit: I'd rename this to `screencopy_damage_get_or_create` to make a clear difference with the function that just gets the damage object.

types/wlr_screencopy_v1.c
```diff
@@ -52,8 +128,26 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
 		return;
 	}
 
+	struct screencopy_damage *damage = NULL;
+	if (frame->with_damage) {
+		damage = screencopy_damage_get(frame->connection, output);
```
emersion: We should check for NULL here in case an allocation failed.

types/wlr_screencopy_v1.c
```diff
@@ -90,6 +201,20 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
 	frame_destroy(frame);
 }
 
+static void frame_handle_output_commit(struct wl_listener *listener,
```
emersion: So this handler is just required to handle `screencopy_damage_accumulate` being called both from the damage precommit handler and from the frame precommit handler. Actually, can't we just always accumulate from both without this check? This will just add damage twice, which is the same as adding damage once.

include/wlr/types/wlr_screencopy_v1.h
```diff
@@ -27,9 +27,14 @@ struct wlr_screencopy_manager_v1 {
...
+	struct wl_list damages;
+};
+
 struct wlr_screencopy_frame_v1 {
 	struct wl_resource *resource;
-	struct wlr_screencopy_manager_v1 *manager;
+	struct wlr_screencopy_connection_v1 *connection;
```
emersion: The frame can outlive the client/connection object. We need to handle this somehow.

types/wlr_screencopy_v1.c
any1: The accumulated damage is cleared at the end of the frame precommit callback. If the damage precommit callback happens after that, it will add damage to the accumulator that has already been cleared and sent to the client. This takes care of such an unfortunate sequence of events.

emersion: Ah, right, there's this case too, you're right.

It's pretty annoying to be honest. We're just "lucky" to have two events (precommit then commit) here, and this `is_handled` flag is pretty complicated. Maybe we can find another solution? For instance, in the damage precommit handler do nothing if a frame object for the same output and client/connection exists?

any1: Yeah, I agree that it's a little complicated.
What about adding a new callback that reports damage that's always called before precommit? It would be three lines in a different part of the system versus a few lines plus added complexity in this part.

include/wlr/types/wlr_screencopy_v1.h
```diff
@@ -27,9 +26,14 @@ struct wlr_screencopy_manager_v1 {
...
 
+struct wlr_screencopy_v1_client {
+	struct wl_list frames; // wlr_screencopy_frame_v1::link
+	struct wl_list damages;
+};
+
 struct wlr_screencopy_frame_v1 {
 	struct wl_resource *resource;
-	struct wlr_screencopy_manager_v1 *manager;
+	struct wlr_screencopy_v1_client *connection;
```
emersion: All of these should be renamed to `client` as well.

types/wlr_screencopy_v1.c
```diff
@@ -8,10 +8,103 @@
...
+}
+
+static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
+	struct pixman_region32 *region = &damage->damage;
+	struct wlr_output *output = damage->output;
+
+	/* This check is done so damage that has been added and cleared in the
+	 * frame precommit handler is not added again after it has been handled.
+	 */
+	if (damage->last_commit_seq == output->commit_seq)
```
emersion: Style: always use braces

include/wlr/types/wlr_output.h
```diff
@@ -114,6 +114,8 @@ struct wlr_output {
 
 	struct wlr_output_state pending;
 
+	uint32_t commit_seq;
```
emersion: Can we add docs for this field? Something among the lines of:

> Commit sequence number. Incremented on each commit, may overflow.

types/wlr_screencopy_v1.c
```diff
@@ -324,23 +456,52 @@ static const struct zwlr_screencopy_manager_v1_interface manager_impl = {
 };
 
 void manager_handle_resource_destroy(struct wl_resource *resource) {
+	struct wlr_screencopy_v1_client *conn =
+		connection_from_resource(resource);
+
+	struct wlr_screencopy_frame_v1 *frame, *tmp_frame;
+	wl_list_for_each_safe(frame, tmp_frame, &conn->frames, link) {
+		wl_resource_destroy(frame->resource);
```
emersion: This isn't in line with the protocol's destroy request description:

> All objects created by the manager will still remain valid, until their appropriate destroy request has been called.

include/wlr/types/wlr_screencopy_v1.h
any1: This member can be changed to "client", but arguments and/or variable clash with wl_client parameters in some places. I could rename those wl_client.

include/wlr/types/wlr_output.h
```diff
@@ -114,6 +114,9 @@ struct wlr_output {
 
 	struct wlr_output_state pending;
 
+        // Commit sequence number. Incremented on each commit, may overflow.
```
emersion: Style: we use tabs, not spaces

types/wlr_screencopy_v1.c
emersion: Bump: no newline between `)` and `{`

types/wlr_screencopy_v1.c
```diff
@@ -8,10 +8,118 @@
...
+static struct screencopy_damage *screencopy_damage_get_or_create(
+		struct wlr_screencopy_v1_client *client,
+		struct wlr_output *output) {
+	struct screencopy_damage *damage = screencopy_damage_find(client, output);
+	return damage ? damage : screencopy_damage_create(client, output);
+}
+
+static void client_unref(struct wlr_screencopy_v1_client *client)
+{
+	if (--client->ref != 0) {
```
emersion: Would be nice to add `assert(client->ref > 0)` right before

types/wlr_screencopy_v1.c
```diff
@@ -81,6 +201,21 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
 
 	zwlr_screencopy_frame_v1_send_flags(frame->resource, flags);
 
+	if (damage) {
+		struct pixman_box32 *damage_box =
```
emersion: Please add a `// TODO: send fine-grained damage events`

ddevault:
+1 from me once @emersion's comments are addressed.

ddevault:
Thanks!

PR #1793, opened 2019-08-14
backend/drm: fix missing plane formats

emersion:
The loop populating the format list was exiting early if ARGB8888 was found.


ascent12:
Thanks

PR #1794, opened 2019-08-15
wlr_xdg_toplevel: reparent on parent unmap

RedSoxFan:
Fixes  swaywm/sway#4421

[From the xdg-shell specification:](https://github.com/wayland-project/wayland-protocols/blob/630fb089103a1d0eab1a684b853ab5c4d2b252aa/stable/xdg-shell/xdg-shell.xml#L555)

	If the parent is unmapped then its children are managed as
	though the parent of the now-unmapped parent has become the
	parent of this surface. If no parent exists for the now-unmapped
	parent then the children are managed as though they have no
	parent surface.

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -117,8 +117,10 @@ struct wlr_xdg_toplevel_state {
 struct wlr_xdg_toplevel {
 	struct wl_resource *resource;
 	struct wlr_xdg_surface *base;
-	struct wlr_xdg_surface *parent;
 	bool added;
+
+	struct wlr_xdg_surface *parent;
+	struct wl_listener parent_unmap;
 
 	struct wlr_xdg_toplevel_state client_pending;
 	struct wlr_xdg_toplevel_state server_pending;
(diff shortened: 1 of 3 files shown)


emersion:
We need to change the parent back if any intermediate toplevel re-maps.

I'm not sure this should be handled in wlroots TBH.

RedSoxFan:
> We need to change the parent back if any intermediate toplevel re-maps.

Is that implicitly implied? I'm not seeing that explicitly stated anywhere and that was not my understanding from reading the specification.

> I'm not sure this should be handled in wlroots TBH.

I'm not sure why this wouldn't be handled in wlroots.

emersion:
>Is that implicitly implied? I'm not seeing that explicitly stated anywhere and that was not my understanding from reading the specification.

You are right. This patch does enough, unmapping resets all state.

```
10:33 <jadahl> emersion: if the parent is mapped again, any previous relationships are gone. spec says that unmapping (attach null) effectively puts it in the same state as if it was just created (thus no parents/children)
10:33 <jadahl> unmappign by destroying the role objects should be considered the same. no state is kept.
```

Issue #1780, opened 2019-08-05
Failing modeset on MST devices


Emantor:
When attaching my latptop to the dock, one monitor works while the other stays enabled but black:
```
Output DP-2 'Dell Inc. DELL U2410 F525M9AK0MML' (focused)
  Current mode: 1920x1200 @ 59.950001 Hz
  Position: 1920,0
  Scale factor: 1.000000
  Subpixel hinting: unknown
  Transform: normal
  Workspace: web
  Available modes:
    720x400 @ 70.082001 Hz
    640x480 @ 59.939999 Hz
    640x480 @ 59.939999 Hz
    640x480 @ 60.000000 Hz
    640x480 @ 75.000000 Hz
    720x480 @ 59.939999 Hz
    720x480 @ 59.939999 Hz
    720x480 @ 59.939999 Hz
    720x480 @ 60.000000 Hz
    720x480 @ 60.000000 Hz
    720x576 @ 50.000000 Hz
    720x576 @ 50.000000 Hz
    800x600 @ 60.317001 Hz
    800x600 @ 75.000000 Hz
    1024x768 @ 60.004002 Hz
    1024x768 @ 75.028999 Hz
    1280x720 @ 50.000000 Hz
    1280x720 @ 59.939999 Hz
    1280x720 @ 60.000000 Hz
    1280x720 @ 60.000000 Hz
    1152x864 @ 75.000000 Hz
    1280x1024 @ 60.020000 Hz
    1280x1024 @ 75.025002 Hz
    1600x1200 @ 60.000000 Hz
    1920x1080 @ 23.976000 Hz
    1920x1080 @ 24.000000 Hz
    1920x1080 @ 50.000000 Hz
    1920x1080 @ 59.939999 Hz
    1920x1080 @ 60.000000 Hz
    1920x1080 @ 60.000000 Hz
    1920x1200 @ 59.950001 Hz

Output DP-4 'Dell Inc. DELL U2212HM Y9KCH21D0KAL' (inactive)
  Available modes:
    720x400 @ 70.082001 Hz
    640x480 @ 59.939999 Hz
    640x480 @ 75.000000 Hz
    800x600 @ 60.317001 Hz
    800x600 @ 75.000000 Hz
    1024x768 @ 60.004002 Hz
    1024x768 @ 75.028999 Hz
    1152x864 @ 75.000000 Hz
    1280x1024 @ 60.020000 Hz
    1280x1024 @ 75.025002 Hz
[...]

PR #1797, opened 2019-08-16
output: re-introduce atomic mode, enabled, scale and transform

emersion:
This reverts commit 01f903874b7e27539488fad7f31476d5bcbc6ac9 and re-applies
commit ee5f98ad49fed0439f3313ec685307831d1d1d05.

Updates: https://github.com/swaywm/wlroots/issues/1640 (Atomic output updates issue)
See also: https://github.com/swaywm/wlroots/pull/1762 (Atomic output updates original PR)
See also: https://github.com/swaywm/wlroots/issues/1780 (Issue caused by atomic output updates)
See also: https://github.com/swaywm/sway/issues/4419 (Issue caused by atomic output updates)
See also: https://github.com/swaywm/wlroots/pull/1781 (Revert PR)

Sway PR: https://github.com/swaywm/sway/pull/4845

* * *

Breaking changes:

* Compositors now need to call `wlr_output_commit` after `wlr_output_enable`, `wlr_output_set_mode`, `wlr_output_set_custom_mode`, `wlr_output_set_transform` or `wlr_output_set_scale`
* `wlr_output_set_mode` no longer implicitly enables the output, compositors need to call `wlr_output_enable` as well
* Compositors now need to call `wlr_output_rollback` after `wlr_output_attach_render` if they decide not to render

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -422,12 +422,6 @@ static bool drm_connector_commit(struct wlr_output *output) {
 		return false;
 	}
 
-	if (output->pending.committed & WLR_OUTPUT_STATE_ENABLED) {
-		if (!enable_drm_connector(output, output->pending.enabled)) {
-			return false;
-		}
-	}
-
 	if (output->pending.committed & WLR_OUTPUT_STATE_MODE) {
 		switch (output->pending.mode_type) {
 		case WLR_OUTPUT_STATE_MODE_FIXED:
@@ -443,6 +437,12 @@ static bool drm_connector_commit(struct wlr_output *output) {
 				return false;
 			}
 			break;
+		}
+	}
+
+	if (output->pending.committed & WLR_OUTPUT_STATE_ENABLED) {
+		if (!enable_drm_connector(output, output->pending.enabled)) {
+			return false;
 		}
 	}
 
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -446,7 +446,9 @@ static bool drm_connector_commit(struct wlr_output *output) {
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


ddevault:
Can you point out the changes to this which fix the bug we reverted this for in the first place?

escalade:
Sway fails to build with this:

```
../sway/config/output.c: In function 'set_mode':
../sway/config/output.c:209:10: error: void value not ignored as it ought to be
  209 |   return wlr_output_set_custom_mode(output, width, height, mhz);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../sway/config/output.c:229:9: error: void value not ignored as it ought to be
  229 |  return wlr_output_set_mode(output, best);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../sway/config/output.c: In function 'apply_output_config':
../sway/config/output.c:270:19: error: void value not ignored as it ought to be
  270 |   modeset_success = wlr_output_set_mode(wlr_output, mode);
      |                   ^
../sway/config/output.c: In function 'set_mode':
../sway/config/output.c:230:1: warning: control reaches end of non-void function [-Wreturn-type]
  230 | }
      | ^
```

emersion:
This is now ready for review.

The previous failures were due to `wlr_output_attach_render` being called without `wlr_output_commit` in Sway. A buffer was registered in the pending state, submitted in the next commit. The leftover buffer was making the commit fail (because all buffers were busy).

To fix this:

* A new function, `wlr_output_rollback`, allows to discard the pending state and cancel a `wlr_output_attach_render` call
* `wlr_output` is now more strict and refuses to commit a buffer or modeset a disabled output

ddevault:
Thanks!

PR #1798, opened 2019-08-18
presentation_feedback: add the sampled state

YaLTeR:
cc https://github.com/swaywm/sway/issues/4451

The idea is that feedback goes from nothing to committed to either sampled (and subsequently presented) or discarded.

Sway patch: https://github.com/swaywm/sway/pull/4465

* * *

Compositor maintainers: call `wlr_presentation_surface_sampled` when rendering a surface.


ddevault:
I wonder if this is causing us to look bad in input latency benchmarks

emersion:
>I wonder if this is causing us to look bad in input latency benchmarks

Unlikely, these don't use this protocol.

PR #1802, opened 2019-08-24
wlr-layer-shell-v1: destroy xdg popups on unmap

RedSoxFan:
Mirrors #1530 (xdg-shell) for wlr-layer-shell-v1
Supersedes/Closes #1418 (#1530 + this is basically #1418)
Fixes swaywm/sway#4477

This destroys the xdg popups associated with a layer surface when the
layer surface is unmapped. It does not make sense to keep the popups
open when unmapped.


ddevault:
Thanks!

Issue #1699, opened 2019-05-18
RDP backend weirdness and crash


aemino:
When using the `rdp` backend and attempting to connect, with either `xfreerdp ... --rfx` or `xfreerdp ... --nsc`, my status bar (waybar) does not render properly, there are artifacts when moving the mouse, and a few seconds after moving the mouse, sway crashes. I'm not sure what useful information I can provide about this other then the following crash log. I included a few non-Sway messages just in case this issue is entirely unrelated to `wlroots`.

```xkbcommon: ERROR: Unrecognized RMLVO variant "nodeadkeys" was ignored
xkbcommon: ERROR: Unrecognized RMLVO option "grp:alt_toggle" was ignored
[19:25:25:717] [2620:2620] [INFO][com.freerdp.core.listener] - Listening on 0.0.0.0:3389
...
glamor: No eglstream capable devices found
glamor: 'wl_drm' not supported
Missing Wayland requirements for glamor GBM backend
Missing Wayland requirements for glamor EGLStream backend
Failed to initialize glamor, falling back to sw
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Unsupported high keycode 372 for name <I372> ignored
>                   X11 cannot support keycodes above 255.
>                   This warning only shows for the first high keycode.
> Internal error:   Could not resolve keysym XF86MonBrightnessCycle
Errors from xkbcomp are not fatal to the X server
Resources files: /home/aemino/.config/waybar/config, /home/aemino/.config/waybar/style.css
Unknown option -b
Created log file at "/tmp/Alacritty-2632.log"
[2019-05-17 19:25] [ERROR] Alacritty encountered an unrecoverable error:

	Error creating GL context; No monitors were detected.

[19:25:40:382] [2620:2620] [INFO][com.freerdp.core.connection] - Client Security: NLA:0 TLS:1 RDP:0
[19:25:40:383] [2620:2620] [INFO][com.freerdp.core.connection] - Server Security: NLA:0 TLS:1 RDP:1
[19:25:40:383] [2620:2620] [INFO][com.freerdp.core.connection] - Negotiated Security: NLA:0 TLS:1 RDP:0
[19:25:40:423] [2620:2620] [INFO][com.freerdp.core.connection] - Accepted client: NUS11180-10-roh
[19:25:40:423] [2620:2620] [INFO][com.freerdp.core.connection] - Accepted channels:
[...]

PR #1809, opened 2019-08-31
Fix calculation for partial gles2 pixel read

filips:
I believe there is a calculation error in the gles2_read_pixels() function, causing issue #1699 .

The calculation is OK for the first readout of the entire viewport, but once a partial area roughly below the middle of the viewport is changed, the data offset will make the glReadPixels() overflow the buffer, causing the seg fault and strange display behaviour discussed in the bug report.

### DIFF
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -317,8 +317,8 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
 	} else {
 		// Unfortunately GLES2 doesn't support GL_PACK_*, so we have to read
 		// the lines out row by row
-		for (size_t i = src_y; i < src_y + height; ++i) {
-			glReadPixels(src_x, src_y + height - i - 1, width, 1, fmt->gl_format,
+		for (size_t i = 0; i < height; ++i) {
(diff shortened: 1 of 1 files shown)


render/gles2/renderer.c
```diff
@@ -317,8 +317,8 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
 	} else {
 		// Unfortunately GLES2 doesn't support GL_PACK_*, so we have to read
 		// the lines out row by row
-		for (size_t i = src_y; i < src_y + height; ++i) {
-			glReadPixels(src_x, src_y + height - i - 1, width, 1, fmt->gl_format,
+		for (size_t i = 0; i < height; ++i) {
+	        glReadPixels(src_x, renderer->viewport_height - src_y - i, width, 1, fmt->gl_format,
```
ddevault: style error, use tabs for indentation

emersion: If `src_y = 0` and `i = 0`, then `renderer->viewport_height - src_y - i == renderer->viewport_height`. This is off-by-one for a y coordinate.

PR #1810, opened 2019-09-01
backend/drm: refresh connector props on scan

### LINKED ISSUE #1807: DRM connector properties not updated on hotplug
A DRM hotplug event means "refresh all your state". However we only look at connector status and link-status, and we don't update anything else.

This is an issue for e.g. an EDID change during suspend where the connector stays connected but its EDID is updated (user suspends, unplugs, replugs, resumes).

Emantor:
Instead of refreshing the properties only on DISCONNECT/CONNECT, always
refresh them during a connector scan.

Fixes #1807

Signed-off-by: Rouven Czerwinski <rouven@czerwinskis.de>


Emantor:
Needs testing, but I don't have a hotpluggable display on the train :smile: Will test once I arrive home.

backend/drm/drm.c
```diff
@@ -1255,6 +1255,15 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 			}
 		}
 
+		memset(&wlr_conn->props, 0, sizeof(wlr_conn->props));
```
emersion: We can do all of this only if `drm_conn->connection == DRM_MODE_CONNECTED`. There's no point in doing it for disconnected connectors.

backend/drm/drm.c
```diff
@@ -1255,6 +1255,15 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
...
+		get_drm_connector_props(drm->fd, wlr_conn->id, &wlr_conn->props);
+
+		size_t edid_len = 0;
+		uint8_t *edid = get_drm_prop_blob(drm->fd,
+			wlr_conn->id, wlr_conn->props.edid, &edid_len);
+		parse_edid(&wlr_conn->output, edid_len, edid);
```
emersion: We need to signal the fact that the EDID changed, if it changed. I think the easiest way to do it is by removing and re-adding the connector.

Emantor: Wouldn't this also mean that the optimization to not modeset existing outputs is lost? If I remove and re-add the connector, a new crtc will get allocated and the compositor has to restart rendering on the new output.

emersion: When the EDID changes, it means the monitor has been swapped, ie. supported resolutions changed. Maybe we can skip modesets if `memcmp(cur_mode, new_mode) == 0`?

emersion:
Also need to reload the list of modes.

backend/drm/drm.c
emersion: Note: we only need to signal if something has changed.

Emantor:
I have thought about this and its probably easiest to retrieve edid for the current drm connectors, remove the connectors with mismatched edid and let the function allocate new ones. This way we keep the mode and crtc and only remove mismatched connectors. IDK if edid is the correct blob to match on, I'd be happy if there is another way.

Emantor:
This now parses the edid and compares to the existing make, model and serial.

backend/drm/drm.c
```diff
@@ -1185,11 +1186,27 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
...
 			index++;
 			if (c->id == drm_conn->connector_id) {
-				wlr_conn = c;
+				union wlr_drm_connector_props drm_props;
+				struct wlr_output drm_output;
+				get_drm_connector_props(drm->fd, drm_conn->connector_id, &drm_props);
+				edid = get_drm_prop_blob(drm->fd,
+					drm_conn->connector_id, drm_props.edid, &edid_len);
+				parse_edid(&drm_output, edid_len, edid);
+				if (strcmp(wlr_conn->output.make, drm_output.make) == 0 &&
```
emersion: Instead of comparing the make/model/serial, I'd compare the blob ID. Some manufacturers don't fill make/model/serial correctly: two different screens (ie. different modes, size, etc) can have the same make/model/serial.

backend/drm/drm.c
```diff
@@ -1191,7 +1191,19 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
...
+				union wlr_drm_connector_props cur_props, new_props;
+				uint64_t cur_blob, new_blob;
+				wlr_log(WLR_DEBUG, "c->id: %d, drm_conn->id: %d", c->id, drm_conn->connector_id);
+				get_drm_connector_props(drm->fd, c->id, &cur_props);
+				get_drm_connector_props(drm->fd, drm_conn->connector_id, &new_props);
+				get_drm_prop(drm->fd, c->id, cur_props.edid, &cur_blob);
+				get_drm_prop(drm->fd, drm_conn->connector_id, new_props.edid, &new_blob);
 				wlr_conn = c;
+				if (cur_blob != new_blob)
+				{
```
ddevault: style, move { to same line as )

Emantor:
FWIW: this currently also requires a patches kernel with the series in https://patchwork.freedesktop.org/series/62816/ applied. I'll see if I can built a proper kernel to test today.

Emantor:
The required patch is now contained in kernels `>5.9-rc1`.

Emantor:
I tested this with kernel 5.10.48, however my kernel no longer seems to send a DRM change event. However this is still testable using the session active mechanism, i.e. by switching to a different vtty and back.

backend/drm/drm.c
```diff
@@ -1284,7 +1285,22 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
...
+				struct wlr_output edid_holder;
+				size_t edid_len = 0;
+				uint8_t *edid = get_drm_prop_blob(drm->fd, c->id,
+					c->props.edid, &edid_len);
+
+				parse_edid(&edid_holder, edid_len, edid);
+				free(edid);
+
+				if (strcmp(c->output.make, edid_holder.make) == 0
+					&& strcmp(c->output.model, edid_holder.make) == 0
```
progandy, line 1297: That ```make``` seems to be a typo.

Also, you compare make/model/serial again and not the edid blob id as mentioned in the commit message.

Emantor: Yep, I also fixed up the commit message, thanks for the review :+1:

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1810

PR #1814, opened 2019-09-03
backend/drm: check for mst: in path property

### LINKED ISSUE #1813: MST detection checks for PATH != 0
Instead we should read the property and check for the `mst:` prefix.

https://github.com/swaywm/wlroots/blob/master/backend/drm/drm.c#L1308

Otherwise this patch will break it: https://lists.freedesktop.org/archives/dri-devel/2019-June/221904.html

Emantor:
Instead of checking that the path property is not 0 to determine if the
connector is an MST connector, check if the path contains the mst:
string.

Fixes #1813

TODO:
- [x] needs testing

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1304,8 +1304,17 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 				wl_list_insert(&wlr_conn->output.modes, &mode->wlr_mode.link);
 			}
 
+			size_t path_len;
+			bool is_mst = false;
+			char *path = get_drm_prop_blob(drm->fd, wlr_conn->id,
+				wlr_conn->props.path, &path_len);
+			if (path_len > 4 && path && strncmp(path, "mst:", 4) == 0) {
+				is_mst = true;
+			}
+			free(path);
+
 			wlr_conn->possible_crtc = get_possible_crtcs(drm->fd, res, drm_conn,
-				wlr_conn->props.path != 0);
+				is_mst);
 			if (wlr_conn->possible_crtc == 0) {
 				wlr_log(WLR_ERROR, "No CRTC possible for connector '%s'",
 					wlr_conn->output.name);


backend/drm/drm.c
```diff
@@ -1304,8 +1304,12 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 				wl_list_insert(&wlr_conn->output.modes, &mode->wlr_mode.link);
 			}
 
+			bool is_mst = false;
+			if (strstr(wlr_conn->props.path, "mst:"))
```
emersion: `wlr_conn->props.path` is an `uint32_t`, need to `drmModePropertyGetBlob` it :P

Emantor: Damn! I thought I could get away cheap :D

backend/drm/drm.c
```diff
@@ -1304,8 +1304,14 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 				wl_list_insert(&wlr_conn->output.modes, &mode->wlr_mode.link);
 			}
 
+			bool is_mst = false;
+			drmModePropertyBlobRes *blob = drmModeGetPropertyBlob(drm->fd, wlr_conn->props.path);
+			if (blob && blob->data && strstr(blob->data, "mst:"))
```
emersion: Maybe we can check `strstr(…) == blob->data` to make sure it's a prefix?

emersion: Style: brackets are mandatory :)

ascent12: `strncmp` seems like it would be more appropriate.

backend/drm/drm.c
```diff
@@ -1304,8 +1304,14 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 				wl_list_insert(&wlr_conn->output.modes, &mode->wlr_mode.link);
 			}
 
+			bool is_mst = false;
+			drmModePropertyBlobRes *blob = drmModeGetPropertyBlob(drm->fd, wlr_conn->props.path);
```
ascent12: I had a helper function for reading blobs into a `malloc`ed buffer, and handled `prop == 0`: `get_drm_prop_blob`.

Emantor: Nice, I'll switch to that.

backend/drm/drm.c
Emantor: Too much OP-TEE and kernel code. Done :)

Emantor: Switched to strncmp  since we are checking only the first 4 bytes anyway

Emantor:
Tested this on my laptop with MST HUB (and an extra log message to verify), works correctly.

backend/drm/drm.c
```diff
@@ -1304,8 +1304,17 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 				wl_list_insert(&wlr_conn->output.modes, &mode->wlr_mode.link);
 			}
 
+			size_t path_len;
+			bool is_mst = false;
+			char *path = get_drm_prop_blob(drm->fd, wlr_conn->id,
+				wlr_conn->props.path, &path_len);
+			if(path_len > 4 && path && strncmp(path, "mst:", 4) == 0) {
+				is_mst = true;
+				free(path);
```
ascent12: Put the free outside of the if statement. For example, if PATH was ever to change and not include `mst:`, this would be a memory leak.
`free(NULL);` is a no-op, so you don't need to worry about `get_drm_prop_blob` failing.

backend/drm/drm.c
```diff
@@ -1304,8 +1304,17 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 				wl_list_insert(&wlr_conn->output.modes, &mode->wlr_mode.link);
 			}
 
+			size_t path_len;
+			bool is_mst = false;
+			char *path = get_drm_prop_blob(drm->fd, wlr_conn->id,
+				wlr_conn->props.path, &path_len);
+			if(path_len > 4 && path && strncmp(path, "mst:", 4) == 0) {
```
ascent12: Style: space after the `if`.

backend/drm/drm.c
Emantor: Absolutely, amended.

Emantor:
Thanks for the reviews @emersion & @ascent12!

PR #1815, opened 2019-09-04
Call pthread_sigmask() as early as possible so Xwayland SIGUSR1 is not caught by GPU driver threads

a-m-joseph:
Supersedes/replaces https://github.com/swaywm/sway/pull/4537

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -2,6 +2,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <libinput.h>
+#include <signal.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


emersion:
Hmm, this isn't a good place for this piece of code.

Can't the Mali driver ignore SIGUSR1?

a-m-joseph:
> Hmm, this isn't a good place for this piece of code.
> 
> Can't the Mali driver ignore SIGUSR1?

No.

emersion:
>No.

Why not?

a-m-joseph:
> > No.
> 
> Why not?

This patch is how you make the Mali driver ignore SIGUSR1.

When a thread receives a signal it wasn't expecting, the process terminates. The way you indicate that the signal is not unexpected is by calling pthread_sigmask() before the thread is created, which is what both this patch and Weston do.

emersion:
If there's no way to move this inside `wlr_xwayland`, I'd prefer this patch to be in sway's `main` function.

agx:
> If there's no way to move this inside wlr_xwayland, I'd prefer this patch to be in sway's main function.

I agree. Messing with signal masks in the library will only cause more problems. We're   doing this in phoc

https://source.puri.sm/Librem5/phoc/blob/master/src/main.c#L116

to avoid having the wrong thread receive the signal.

a-m-joseph:
> https://source.puri.sm/Librem5/phoc/blob/master/src/main.c#L116

In your code:

```
  sigprocmask(SIG_BLOCK, &mask, NULL);
```

$ man 2 sigprocmask
...
The use of **sigprocmask() is unspecified in a multithreaded process**; see pthread_sigmask(3). 
...

ddevault:
Mali is in the wrong here. They should not be doing this.

ascent12:
Yeah, I don't see why Mali would have issues with this, yet many of the Mesa drivers (which are multithreaded) don't.
We've have next to no interest in supporting proprietary drivers. It's fine if they happen to work with standard interfaces or with things like GBM, but going out of our way to have workarounds for them is not something I want to carry. That's not even something I would want to do with an open driver.

ddevault:
>We've have next to no interest in supporting proprietary drivers

I would say we have *less* than zero interest.

a-m-joseph:
> They should not be doing this.

Not be doing what, exactly?

PR #1818, opened 2019-09-12
Remove software cursor support

ascent12:
Closes https://github.com/swaywm/wlroots/issues/1363  
See https://github.com/swaywm/wlroots/issues/1363#issuecomment-529786169

~Hardware cursor support hasn't been added back yet, see: https://github.com/swaywm/wlroots/issues/1363#issuecomment-530752843~

I've updated examples/pointer.c to draw its own cursor. I haven't updated examples/multi-pointer.c yet.
They are very similar examples, and I wonder if we really need both.

### DIFF
--- a/include/wlr/interfaces/wlr_output.h
+++ b/include/wlr/interfaces/wlr_output.h
@@ -19,10 +19,6 @@ struct wlr_output_impl {
 	bool (*set_mode)(struct wlr_output *output, struct wlr_output_mode *mode);
 	bool (*set_custom_mode)(struct wlr_output *output, int32_t width,
 		int32_t height, int32_t refresh);
-	bool (*set_cursor)(struct wlr_output *output, struct wlr_texture *texture,
-		int32_t scale, enum wl_output_transform transform,
-		int32_t hotspot_x, int32_t hotspot_y, bool update_texture);
-	bool (*move_cursor)(struct wlr_output *output, int x, int y);
 	void (*destroy)(struct wlr_output *output);
 	bool (*attach_render)(struct wlr_output *output, int *buffer_age);
 	bool (*cursor_try_set_size)(struct wlr_output *output, int *x, int *y);
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -302,9 +302,6 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 void wlr_output_init(struct wlr_output *output, struct wlr_backend *backend,
 		const struct wlr_output_impl *impl, struct wl_display *display) {
 	assert(impl->attach_render && impl->commit);
-	if (impl->set_cursor || impl->move_cursor) {
-		assert(impl->set_cursor && impl->move_cursor);
-	}
 	output->backend = backend;
 	output->impl = impl;
 	output->display = display;


ascent12:
I'm reasonably happy with how this has turned out so far, at least for the backend code/API. I think this is ready for an initial round of review.

With `drm_connector_set_cursor` gone, it opens up a lot of opportunities to simplify many aspects of the DRM backend. I was tempted to do some of that now, but haven't; it'll just be 500+ lines of unrelated work muddying up this PR.

I've adding optionally caching textures inside of `wlr_xcursor_manager`.

`examples/pointer` works on both the DRM and Wayland backends. I added a white background to the cursor to show off the size of the cursor plane.

What's probably needed now is an implementation in a real compositor that handles output rotations (i.e. sway),

---

Now that the compositor has a lot more control over how the cursor is rendered, I realised that it opens the possibility for the compositor to also draw the drag-icon on it too. On AMD, you've got 128x128 to work with, which is enough most of the time. IIRC, Intel's ones were 256x256.

This would be a slight efficiency boost, because you're not damaging the main framebuffer as much.

---

An important point I forgot to bring up before in https://github.com/swaywm/wlroots/issues/1363#issuecomment-529786169 was the handling of subsurfaces, which is actually the main reason for wanting to do this. Right now, we do not correctly handle them or even display them at all, which is a violation of the protocol. Future work such as `wp_viewporter` and any extension that will affect rendering (colour management, alpha compositing) also can not be handled.

The main reason I do not want any sort of helper API is because it needs to be aware of all of these things, be able to adapt to even more things being added in the future, and is just going to be extremely complex. I'd consider something like this done "properly" which can hide all of this to be on the level of WLC, in terms of how high level it needs to be.

Yet the compositor will already have its own ways of drawing subsurface trees, handling rotations, damage tracking, and related things, which I think could just be used without many changes onto the cursor surface or onto the main framebuffer.

Awhile ago, I threw together an insane test client that uses cursor subsurfaces. I'll eventually clean it up and submit it to [wleird](https://github.com/emersion/wleird). Ideally, I want it to render entirely correctly in sway when this work is complete.

ddevault:
>What's probably needed now is an implementation in a real compositor that handles output rotations (i.e. sway)

Agreed, I'd like to see a sway implementation. Does tinywl need updating as well?

> The main reason I do not want any sort of helper API is because it needs to be aware of all of these things, be able to adapt to even more things being added in the future, and is just going to be extremely complex. I'd consider something like this done "properly" which can hide all of this to be on the level of WLC, in terms of how high level it needs to be.

Bleh, that really sucks. I've been dreading things like `wp_viewporter` because it represents offloading a *huge* amount of work onto downstream compositors. Something as simple as "display a surface correctly" is going to be pushed up into the thousands of lines of code. I *really* want to figure out a good way of letting wlroots optionally shoulder this burden.

> Yet the compositor will already have its own ways of drawing subsurface trees, handling rotations, damage tracking, and related things, which I think could just be used without many changes onto the cursor surface or onto the main framebuffer.

I hope so. This is why I want to see how sway et al is affected by these changes.

include/wlr/interfaces/wlr_output.h
```diff
@@ -19,12 +19,11 @@ struct wlr_output_impl {
...
 		int32_t height, int32_t refresh);
-	bool (*set_cursor)(struct wlr_output *output, struct wlr_texture *texture,
-		int32_t scale, enum wl_output_transform transform,
-		int32_t hotspot_x, int32_t hotspot_y, bool update_texture);
-	bool (*move_cursor)(struct wlr_output *output, int x, int y);
 	void (*destroy)(struct wlr_output *output);
 	bool (*attach_render)(struct wlr_output *output, int *buffer_age);
+	bool (*cursor_try_set_size)(struct wlr_output *output, int *x, int *y);
+	bool (*cursor_attach_render)(struct wlr_output *output, int *buffer_age);
+	bool (*cursor_commit)(struct wlr_output *output);
```
ddevault: One thing that bothers me in terms of future-proofing is that we're trying to remove cursors from this interface, yet we're still hardcoding the idea that cursors are special into it. The idea that a cursor plane is special seems rife with legacy potential - I don't see any reason why future graphics architectures would make this call instead of just using generalized planes. What are we to do in that case? Steal a plane and make it the "cursor" plane? Abort and make the compositor show it in software? The idea of cursors at all probably doesn't belong at this level of abstraction. I imagine that ideally the cursorness of a plane becomes an enum value for a bitfield in a general planes API, so that the downstream compositor can search through planes and pick it out as the cursor plane themselves (or this hypothetical cursor helper could do the same). But that's another breaking change we'll face later, so we're designing an API today which we know we'll break later, which is just annoying.

ddevault:
On the subject of the viability of a "helper", wlroots is designed as a whole to meet the compositor at two levels: low-level abstractions they can do whatever they want with (directly managing planes), and high-level abstractions that leverage these to do the high-level task they want (display a cursor on screen). The compositor should be able to say "display a cursor on screen" and wlroots will take the best approach, using hardware or software cursors as necessary. If this requires connecting disaparate systems, then the compositor will provide the connection. If this requires sacrificing some compositor autonomy, then the compositor will either give it up or will peel off the abstraction when they need to.

Adding a "helper" may have to wait until the renderer redesign as a whole is complete, and we have some kind of interface we can pass to this helper which will allow it to render software cursors correctly within the context of the downstream compositor's rendering constraints.

ddevault:
While I'm doing an API design brain dump, another thing which bugs me is the increasingly low-level nature of a *subset* of the `wlr_output` API. Every downstream compositor will have to deal with the high-level parts of wlr_output, but as parts of it become lower-level, the more confusing it becomes in lockstep. Don't have a solution.

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
  */
-void wlr_output_render_software_cursors(struct wlr_output *output,
-	pixman_region32_t *damage);
 
+/**
+ * Sets the size of the hardware cursor buffer to the values pointed to by
+ * x and y.
+ *
+ * Some backends have strict requirements for the cursor size and may not be
+ * able to use the size you requested exactly. The actual value used will be
```
ddevault: Are these strict requirements a min/max value, or do they get more complex (e.g. must be even)? If the former, I'd rather expose the min/max and let the compositor munge the numbers themselves, then fail opaquely if they don't meet the constraints.

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
+ * left in x and y. If this is not large enough for the cursor you wish to
+ * draw, you should fall back to using software cursors.
+ *
+ * This returns false if the backend does not support hardware cursors or from
+ * some other internal error. Either way, you should fall back to using
+ * software cursors.
+ */
+bool wlr_output_cursor_try_set_size(struct wlr_output *output, int *x, int *y);
```
ddevault: Let's also remove `_try` from the name, we don't have any precedent of doing this for things which might fail.

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
+ *
+ * This is in output buffer coordinates. Negative values and values that would
+ * place the cursor partially or entirely off the main framebuffer are allowed.
+ *
+ * The hotspot is the point on the cursor buffer (from the top left) where the
+ * cursor actually points. For example, with a cursor representing a hand, the
+ * hotspot may be over the index finger, rather than the top left corner.
+ *
+ * Some backends may not be able to move the cursor freely, but may still be
+ * able to make use of the hotspot to correctly show the cursor.
```
ddevault: This sounds better to me:

`Some backends (e.g. Wayland) don't control where the cursor goes, but may be able to suggest the hotspot to the system which does.`

The hotspot is not "shown", it informs placement.

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
+ * Apply pending cursor state.
+ *
+ * This may or may not be synchronized with the output, and would be applied on
+ * `wlr_output_commit`, depending on the backend limitations. However, it's
+ * advised to assume that they are sychronized and do any cursor operations at
+ * the same time you render to the output.
+ *
+ * If you have called `wlr_output_cursor_attach_render`, you must call this
+ * function before `wlr_output_attach_render`, otherwise you risk losing or
+ * corrupting rendering state.
```
ddevault: Can we make this an assertion instead of a "risk"?

include/wlr/types/wlr_xcursor_manager.h
```diff
@@ -58,12 +63,19 @@ struct wlr_xcursor *wlr_xcursor_manager_get_xcursor(
...
+ *
+ * The returned texture is owned by the xcursor manager. You should not save or
+ * destroy it.
+ * The manager makes use of the wlr_xcursor_image.userdata field. You must not
+ * modify this field yourself.
  */
-void wlr_xcursor_manager_set_cursor_image(struct wlr_xcursor_manager *manager,
-	const char *name, struct wlr_cursor *cursor);
+struct wlr_texture *wlr_xcursor_manager_get_texture(
+	struct wlr_xcursor_manager *manager, struct wlr_xcursor_image *image);
```
ddevault: Why are we passing in the `wlr_xcursor_image` instead of the name?

include/wlr/types/wlr_xcursor_manager.h
```diff
@@ -58,12 +63,19 @@ struct wlr_xcursor *wlr_xcursor_manager_get_xcursor(
...
+ * wlr_xcursor_manager.
+ *
+ * The wlr_xcursor_manager must have been created with a wlr_renderer and the
+ * wlr_xcursor_image must have come from a wlr_xcursor returned from
+ * wlr_xcursor_manager_get_xcursor.
+ *
+ * The returned texture is owned by the xcursor manager. You should not save or
+ * destroy it.
+ * The manager makes use of the wlr_xcursor_image.userdata field. You must not
+ * modify this field yourself.
```
ddevault: Maybe we can rejigger things a bit so that this field is in a private struct.

include/wlr/interfaces/wlr_output.h
emersion: I think this new API is designed to be temporary, and will be removed when we start exposing universal planes.

ddevault: Yeah, I know. I just want to be cautious and deliberate when we cause churn for downstream compositors.

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
  */
-bool wlr_output_cursor_set_image(struct wlr_output_cursor *cursor,
-	const uint8_t *pixels, int32_t stride, uint32_t width, uint32_t height,
-	int32_t hotspot_x, int32_t hotspot_y);
-void wlr_output_cursor_set_surface(struct wlr_output_cursor *cursor,
-	struct wlr_surface *surface, int32_t hotspot_x, int32_t hotspot_y);
-bool wlr_output_cursor_move(struct wlr_output_cursor *cursor,
-	double x, double y);
-void wlr_output_cursor_destroy(struct wlr_output_cursor *cursor);
+bool wlr_output_cursor_commit(struct wlr_output *output);
```
emersion: Any reason why we need an explicit function for this, instead of doing the commit with `wlr_output_commit`?

include/wlr/types/wlr_output.h
emersion: On DRM, this indicates a width/height that works. This doesn't mean that it's the min/max: for instance, sizes too small might fail. There's no way to get the min/max that i know of.

Compositors should just use this size if the backend has such limitation.

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
+ *
+ * Not every backend/output will support hardware cursors, so you must always
+ * be prepared to fall back to rendering the cursor to the main framebuffer
+ * yourself.
+ *
+ * Hardware cursors are always in output buffer coordinates, and does not take
+ * scale nor rotation into account. You are responsible for doing this
+ * yourself.
+ *
+ * Hardware cursors are always in the ARGB8888 format.
```
emersion: This is DRM-specific, and I don't think this actually matters?

include/wlr/types/wlr_output.h
```diff
@@ -303,26 +294,90 @@ void wlr_output_lock_attach_render(struct wlr_output *output, bool lock);
...
+bool wlr_output_cursor_try_set_size(struct wlr_output *output, int *x, int *y);
+
+/**
+ * Attaches the renderer to the cursor buffer.
+ *
+ * You must successfully call `wlr_output_cursor_try_set_size` at least once
+ * before using this. Any rendering done will be applied on
+ * `wlr_output_cursor_commit`.
+ */
+bool wlr_output_cursor_attach_render(struct wlr_output *output, int *buffer_age);
```
emersion: There's no damage tracking support for cursors. I'd just remove the `buffer_age` parameter.

include/wlr/types/wlr_output.h
ascent12: Originally this is what I was I was going to do, but this caused issues with the rendering state.
```c
eglMakeCurrent(egl, cursor_surface, ...);
// render stuff
eglMakeCurrent(egl, primary_surface, ...);
// render stuff
eglSwapBuffers(egl, cursor_surface, ...);
eglSwapBuffers(egl, primary_surface, ...);
```
doesn't really work, even if you add additional eglMakeCurrent calls before eglSwapBuffers.

include/wlr/types/wlr_xcursor_manager.h
ascent12: Because a user is still expected to need the `wlr_cursor_image` for hotspot information. Doing it this way just prevents needing to do two lookups.

include/wlr/types/wlr_output.h
ascent12: It becomes quite a pain to check reliably. Since the current EGL context/surface bound is global state, it could be bound to any output, even on a different backend.

include/wlr/interfaces/wlr_output.h
ascent12: Yeah, API churn is definitely an issue. But the old cursor code was extremely detrimental towards other work I was doing, particularly `wp_linux_explicit_synchronization` and the backend part of renderer v6. I would've had to hack up the interface even further, and I wasn't keen on adding knowingly broken code.
I feel this is a necessary intermediate step for moving forward.

---

For the DRM backend, universal planes really would make this interface pointless. I personally didn't plan on adding planes support for DRM until after all of this other rendering shit is finished, which at my pace would be *way* off in time. But with emersion doing libliftoff, that could come sooner than I anticipate.

But while I was implementing the Wayland backend part of this, I realised that for it, cursors really are special. They're completely unsynchronized with the main surface (unlike a plane would be), you don't get to move it, and if we wanted to add drag-and-drop support for the Wayland backend, we could even expose drag-icons for it.

We could do something planes-like with `wl_subsurfaces`, but cursors don't fit into that.

Basically, this needs more thought.

include/wlr/types/wlr_output.h
ascent12: I was actually wondering at one point whether it was worth adding damage to this (it's not like it's a lot of work), but I guess I just forgot about it and left this half-way solution.

include/wlr/types/wlr_xcursor_manager.h
ascent12: I didn't really want to make an entire copy of the `wlr_xcursor_theme` "tree". I figured a userdata field would be the simplest and most elegant solution, and could help people not using `wlr_renderer` when that truly becomes viable.

ddevault: Why does moving this to a private struct necessitate a copy of the `wlr_xcursor_theme` "tree"? Unsure of what "tree" means here, either.

include/wlr/types/wlr_output.h
ddevault: I wonder if we should move to `begin/end` semantics, rather than `current/commit`.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1818

PR #1819, opened 2019-09-13
layer-shell: ignore ack_configure() on closed surfaces

ammen99:
When the surface is closed, we destroy all pending serials waiting to be accepted. This means we need to ignore any future ack events, because we can have the following events:

1. -> configure()
2. -> close()
3. <- ack_configure()

At point 3, wlroots will error the client because of invalid serial,
however the client hasn't processed close() yet.


ammen99:
Quite fast, thanks :)

PR #1820, opened 2019-09-14
wlr_seat_touch: Destroy the touchpoint on client destroy

### LINKED ISSUE #1788: Destroying surface with active touchpoints leads to crashes
Reproduced with sway, phoc and rootston.

Steps to reproduce:
1. Launch any app, say, weston-terminal
2. Touch it and don't release.
3. Kill weston-terminal (for instance via ssh)
4. Move or release the finger.

Result: events are being sent on a surface that has been already destroyed, leading to a crash.

dos1:
Since e26217c51e3a5e1d7dfc95a8a76299e056497981, touchpoints can outlive
surfaces. This works fine as long as the client stays around, but fails
horribly otherwise; therefore we have to make sure that touchpoints don't
outlive their clients.

Fixes #1788

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -64,6 +64,7 @@ struct wlr_touch_point {
 
 	struct wl_listener surface_destroy;
 	struct wl_listener focus_surface_destroy;
+	struct wl_listener client_destroy;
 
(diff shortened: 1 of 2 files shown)


emersion:
Thanks both of you!

PR #1824, opened 2019-09-17
xdg-output: send wl_output.done after xdg_output created

versusvoid:
xdg-output version 3 requires to send wl_output.done
"after all xdg_output properties have been sent when the object is created"

### DIFF
--- a/types/wlr_xdg_output_v1.c
+++ b/types/wlr_xdg_output_v1.c
@@ -132,6 +132,8 @@ static void output_manager_handle_get_xdg_output(struct wl_client *client,
 	}
 
 	output_send_details(xdg_output, xdg_output_resource);
+
+	wl_output_send_done(output_resource);
 }
 
 static const struct zxdg_output_manager_v1_interface


types/wlr_xdg_output_v1.c
```diff
@@ -132,6 +132,10 @@ static void output_manager_handle_get_xdg_output(struct wl_client *client,
 	}
 
 	output_send_details(xdg_output, xdg_output_resource);
+
+	if (version => OUTPUT_DONE_DEPRECATED_SINCE_VERSION) {
```
emersion: Hmm, I have no idea how we missed that.

Let's schedule a `wl_output.done` unconditionally, just like in `output_update`.

emersion: Hmm, actually one issue is that this will broadcast the done event for all clients. We only need to schedule it for `output_resource`.

emersion: Maybe we could just immediately send `wl_output.done` and leave the atomicity as a TODO.

Issue #1830, opened 2019-09-24
Direct scan-out support in the Wayland backend


emersion:
This can be useful for testing and for Cage.

Issue #1768, opened 2019-07-23
backend/wayland: spurious eglSwapBuffers failures


emersion:
Sometimes `eglSwapBuffers` fails on the Wayland backend. Maybe we're running out of buffers.

```
2019-07-23 19:12:39 - [EGL] eglSwapInterval: eglSwapInterval
2019-07-23 19:12:39 - [EGL] eglSwapBuffers: eglSwapBuffers
2019-07-23 19:12:39 - [render/egl.c:400] eglSwapBuffers failed
```

RedSoxFan:
It looks like this effects the DRM backend as well: swaywm/sway#4419 and #1780

Emantor:
Probably "fixed" by #1781

emersion:
Can still reproduce.

aemino:
I also encounter this with my toy compositor using the Wayland backend. It tends to happen occasionally when I move my mouse in and out of the compositor window.

ascent12:
I recently updated the EGL log message to give the error code, now making it:
```
2019-09-20 15:30:31 - [EGL] command: eglSwapInterval, error: 0x300d, message: "eglSwapInterval"
2019-09-20 15:30:31 - [EGL] command: eglSwapBuffers, error: 0x300d, message: "eglSwapBuffers"
2019-09-20 15:30:31 - [render/egl.c:402] eglSwapBuffers failed
```

The reliable way to trigger it on sway is to simply have a client open, unfocus then focus the window (happens on both), which I assume is triggering an `xdg_surface.configure` and `wl_surface.frame`. The X11 backend is not affected.

0x300d is `EGL_BAD_SURFACE`.

emersion:
Seems like the error code has changed. It's now:

```
00:01:01.199 [EGL] command: eglSwapInterval, error: 0x3006, message: "eglSwapInterval"
00:01:01.199 [EGL] command: eglSwapBuffers, error: 0x300d, message: "eglSwapBuffers"
00:01:01.199 [render/egl.c:449] eglSwapBuffers failed
```

Seems related to the cursor. Maybe we swap too fast.

PR #1831, opened 2019-09-24
backend/wayland: add support for direct scan-out

emersion:
Closes: https://github.com/swaywm/wlroots/issues/1830

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -5,6 +5,7 @@
 
 #include <wlr/config.h>
 
+#include <drm_fourcc.h>
 #include <wayland-server-core.h>
 
 #include <wlr/backend/interface.h>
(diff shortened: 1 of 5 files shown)


emersion:
The second commit fixes https://github.com/swaywm/wlroots/issues/1768

emersion:
Gah, it doesn't fix it. It seems to me that #1768 is caused by the cursor surface.

backend/wayland/backend.c
```diff
@@ -59,6 +60,25 @@ static const struct xdg_wm_base_listener xdg_wm_base_listener = {
 	xdg_wm_base_handle_ping,
 };
 
+static void linux_dmabuf_v1_handle_format(void *data,
+		struct zwp_linux_dmabuf_v1 *linux_dmabuf_v1, uint32_t format) {
+	// This event is deprecated, we use the modifier event instead
```
ascent12: Do we know if all compositors always use the modifier event if we bind the latest version? E.g. will someone keep using format on AMDGPU without modifiers?
I think it would be worth just implementing this anyway, since it's just a single call to `wlr_drm_format_set_add(&formats, format, DRM_FORMAT_MOD_INVALID)`.

emersion: >will someone keep using format on AMDGPU without modifiers?

Yeah, the compositor should send a `modifier` event with `DRM_FORMAT_MOD_INVALID`.

I still added the call as suggested, just in case.

ddevault:
Nice, thanks!

Issue #655, opened 2018-02-20
Repaint scheduling


emersion:
The idea is to send a frame event to clients, give them time to render a frame, and then render + swap buffers.

See https://www.collabora.com/about-us/blog/2015/02/12/weston-repaint-scheduling/  
See https://github.com/swaywm/wlroots/issues/644#issuecomment-366547068

SirCmpwn:
I feel like we already had an issue for this somewhere...

emersion:
#617 is different but related.

emersion:
```
15:25 <MrCooper> pq: I guess ideally, the repaint window should be adjusted dynamically in response to hitting or missing the target
15:27 <pq> MrCooper, yes, more preferably by looking at the composite task fence completion timestamps.
15:27 <MrCooper> right
```

YaLTeR:
I'm looking into this. I added timeline logging based on the linked Weston patch. Here are my observations so far:
- The reported DRM presentation timestamp can be before and after a `CLOCK_MONOTONIC` measurement inside `page_flip_handler()`, depending on the monitor.
- The minimum time from the return of rootston's `output_render()` to a vblank (basically how far `output_render()` can be delayed) which doesn't drop a frame varies heavily depending on the size of the updating region. With `glxgears` and `weston-simple-egl` at the default small size I am able to delay `output_render()` all the way to 2 milliseconds between `output_render()` returning and vblank without dropped frames, but at 1920×1080 fullscreen that required time increases to 5 milliseconds with 4 milliseconds already causing dropped frames about half of the time.

With that said, with a naive implementation of repaint delay (delays `output_render()` similarly to the Weston post) and frame delay (moves the frame callback from the end of `output_render()` to the specified number of milliseconds after a vblank), and additionally blocking all idle frames between a vblank and an `output_render()`, I was able to achieve sub-frame latency on rootston for clients relying on the frame event (like vsynced XWayland clients):

![image](https://user-images.githubusercontent.com/1794388/61168557-ddbb4280-a504-11e9-9077-2c12e144f12b.png)

Here the frame delay is set to 1 millisecond and repaint delay is set to 2 milliseconds, so the client `[id:1]` has about a little more than 1 millisecond to render, which it successfully does.

emersion:
This has been implemented in swaywm/sway#4588.

PR #1832, opened 2019-09-25
Additions for repaint scheduling

YaLTeR:
For sway PR: https://github.com/swaywm/sway/pull/4588.

Idle frame block: when delaying the output render after a presentation, idle frames need to be blocked because they interfere (and a render is scheduled anyway).

Frame done timer: used for delayed frame callbacks.

cc #655

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -145,6 +145,8 @@ struct wlr_output {
 	struct wl_listener display_destroy;
 
 	void *data;
+
+	bool block_idle_frame;
 };
 
 struct wlr_output_event_precommit {
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -541,7 +541,8 @@ void wlr_output_send_frame(struct wlr_output *output) {
 static void schedule_frame_handle_idle_timer(void *data) {
 	struct wlr_output *output = data;
 	output->idle_frame = NULL;
-	if (!output->frame_pending && output->impl->schedule_frame) {
+	if (!output->frame_pending && output->impl->schedule_frame
+			&& !output->block_idle_frame) {
 		// Ask the backend to send a frame event when appropriate
 		if (output->impl->schedule_frame(output)) {
 			output->frame_pending = true;


include/wlr/types/wlr_surface.h
```diff
@@ -112,6 +112,8 @@ struct wlr_surface {
 	struct wl_listener renderer_destroy;
 
 	void *data;
+
+	struct wl_event_source *frame_done_timer;
```
ddevault: Can you add a comment documenting this?

emersion:
Hmm. Can you elaborate why `block_idle_frame` is needed? I would've guess you would just ignore frame events if the backend provides a presentation event with a predicted next page-flip timestamp.

I'm not a fan of `frame_done_timer` since this can very easily be implemented in the compositor.

Does this still work with the Wayland backend? What about Wayland backend plus nested in Weston (which already has frame scheduling)?

YaLTeR:
> Hmm. Can you elaborate why `block_idle_frame` is needed? I would've guess you would just ignore frame events if the backend provides a presentation event with a predicted next page-flip timestamp.

The issue I had is that as soon as `output_render` is put on a timer, an idle frame gets rendered (there's usually always at least one of those), which prevents the actual `output_render` from working properly once it's executed.

include/wlr/types/wlr_output.h
```diff
@@ -145,6 +145,8 @@ struct wlr_output {
 	struct wl_listener display_destroy;
 
 	void *data;
+
+	bool block_idle_frame;
```
ddevault: Docs here as well, please.

YaLTeR:
> I'm not a fan of `frame_done_timer` since this can very easily be implemented in the compositor.

Moved that into the sway PR.

ddevault:
Thanks!

emersion:
Ugh. I don't think we should have `block_idle_frame`.

ddevault:
Gah. What do you want to do about it?

emersion:
Let's keep it that way for now.

I don't like it because the frame API is already too complicated. I don't think the repaint scheduling stuff should listen to frame events at all.

I'll follow up in the Sway PR, and remove it here if we decide it's a better idea not to have it.

PR #1833, opened 2019-09-26
Fixes crash of compositor when unvalidated keycode 0 is passed

### LINKED ISSUE #1821: Whole sway crashes when using squeekboard (unprotected handler)
Sway crashes completely (with all childs...) with segfault when key "q" is pressed on squeekboard (https://source.puri.sm/Librem5/squeekboard). Otherwise it's working very nice, btw.

Actual reason is that keycode 0 is arriving here: https://github.com/swaywm/wlroots/blob/master/types/wlr_virtual_keyboard_v1.c#L75

It eventually crashes in set_add due assert.

Root cause for this keycode 0 is irrelevant in this case, in my humble opinion, because the thing shouldn't crash because of it in any case - this function is given to Wayland server as handler, so it can give it anything.

Simple if(keycode == 0) ... would be nice there.

amingin:
Fixes #1821

### DIFF
--- a/types/wlr_virtual_keyboard_v1.c
+++ b/types/wlr_virtual_keyboard_v1.c
@@ -75,7 +75,7 @@ context_fail:
 static void virtual_keyboard_key(struct wl_client *client,
 		struct wl_resource *resource, uint32_t time, uint32_t key,
 		uint32_t state) {
-	if(key == 0) {
+	if (key == 0) {
 		return;
 	}
 	struct wlr_virtual_keyboard_v1 *keyboard =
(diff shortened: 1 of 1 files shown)


types/wlr_virtual_keyboard_v1.c
```diff
@@ -75,6 +75,9 @@ static void virtual_keyboard_keymap(struct wl_client *client,
 static void virtual_keyboard_key(struct wl_client *client,
 		struct wl_resource *resource, uint32_t time, uint32_t key,
 		uint32_t state) {
+	if(key == 0) {
```
emersion: Style: space between `if` and `(`

amingin: Yup, done

PR #1841, opened 2019-10-03
xdg-foreign: add implementation

### LINKED ISSUE #1702: Add xdg-foreign support
This is used by flatpak applications, and potentially others.

cyclopsian:
Closes #1702

Sway branch for testing is here: [cyclopsian/sway:xdg-foreign](https://github.com/cyclopsian/sway/tree/xdg-foreign)

To test, run [portal-test](https://github.com/matthiasclasen/portal-test) or some other flatpak application that uses the desktop portal. Make the application fullscreen in sway and open a file dialog. Before these patches, the file dialogs would appear behind the fullscreen surface since toplevel->parent was not being set.

Both versions are supported because Mutter/GTK uses v1 and KWayland uses v2.

Possible changes that should be made (feedback requested):

- Combine the v1 and v2 implementations instead of copying the file? Both implementations are exactly the same. 
- Use an actually secure way of generating handles, maybe use libuuid?
- Store the handles in a hash table?

### DIFF
--- a/.builds/freebsd.yml
+++ b/.builds/freebsd.yml
@@ -10,6 +10,7 @@ packages:
 - graphics/png
 - graphics/wayland
 - graphics/wayland-protocols
+- misc/e2fsprogs-libuuid
 - multimedia/ffmpeg
 - x11/libX11
 - x11/libinput
--- a/include/wlr/types/wlr_xdg_foreign_v1.h
+++ b/include/wlr/types/wlr_xdg_foreign_v1.h
@@ -1,0 +1,86 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_XDG_FOREIGN_V1_H
+#define WLR_TYPES_WLR_XDG_FOREIGN_V1_H
+
+#include <wayland-server-core.h>
+
+#define WLR_XDG_FOREIGN_V1_HANDLE_SIZE 37
+
+struct wlr_xdg_foreign_v1 {
+	struct {
+		struct wl_global *global;
+		struct wl_list resources;
+		struct wl_list clients;
+	} exporter, importer;
+
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_xdg_exporter_v1 {
+	struct wlr_xdg_foreign_v1 *foreign;
+	struct wl_resource *resource;
+
+	struct wl_list link; // wlr_xdg_foreign_v1::exporter::clients
+
+	struct wl_list exports;
+};
+
+struct wlr_xdg_importer_v1 {
+	struct wlr_xdg_foreign_v1 *foreign;
+	struct wl_resource *resource;
+
+	struct wl_list link; // wlr_xdg_foreign_v1::importer::clients
+
+	struct wl_list imports;
+};
+
+struct wlr_xdg_exported_v1 {
+	struct wlr_surface *surface;
+	struct wl_resource *resource;
+
+	struct wl_list link; // wlr_xdg_exporter_v1::exports
+
+	struct wl_listener xdg_surface_unmap;
+
+	struct wl_list imports;
+	char handle[WLR_XDG_FOREIGN_V1_HANDLE_SIZE];
+};
+
+struct wlr_xdg_imported_v1 {
+	struct wlr_xdg_exported_v1 *exported;
+	struct wl_resource *resource;
+
+	struct wl_list export_link; // wlr_xdg_exported_v1::imports
+	struct wl_list link; // wlr_xdg_importer_v1::imports
+
+	struct wl_list children;
+};
+
+struct wlr_xdg_imported_child_v1 {
+	struct wlr_xdg_imported_v1 *imported;
+	struct wlr_surface *surface;
+
+	struct wl_list link; // wlr_xdg_imported_v1::children
+
+	struct wl_listener xdg_surface_unmap;
+	struct wl_listener xdg_toplevel_set_parent;
+};
+
+struct wlr_xdg_foreign_v1 *wlr_xdg_foreign_v1_create(
+		struct wl_display *display);
+void wlr_xdg_foreign_v1_destroy(struct wlr_xdg_foreign_v1 *foreign);
+
+#endif
--- a/meson.build
+++ b/meson.build
@@ -87,6 +87,7 @@ conf_data.set10('WLR_HAS_X11_BACKEND', false)
 conf_data.set10('WLR_HAS_XWAYLAND', false)
 conf_data.set10('WLR_HAS_XCB_ERRORS', false)
 conf_data.set10('WLR_HAS_XCB_ICCCM', false)
+conf_data.set10('WLR_HAS_XDG_FOREIGN', false)
 
 # Clang complains about some zeroed initializer lists (= {0}), even though they
 # are valid
@@ -107,6 +108,8 @@ libinput = dependency('libinput', version: '>=1.9.0')
 xkbcommon = dependency('xkbcommon')
 udev = dependency('libudev')
 pixman = dependency('pixman-1')
+uuid = dependency('uuid', required: false)
+uuid_create = cc.has_function('uuid_create')
 math = cc.find_library('m')
 rt = cc.find_library('rt')
 
@@ -179,6 +182,7 @@ summary = [
 	' x11_backend: @0@'.format(conf_data.get('WLR_HAS_X11_BACKEND', false)),
 	'   xcb-icccm: @0@'.format(conf_data.get('WLR_HAS_XCB_ICCCM', false)),
 	'  xcb-errors: @0@'.format(conf_data.get('WLR_HAS_XCB_ERRORS', false)),
+	' xdg-foreign: @0@'.format(conf_data.get('WLR_HAS_XDG_FOREIGN', false)),
 	'----------------',
 	''
 ]
--- a/types/meson.build
+++ b/types/meson.build
@@ -68,3 +68,10 @@ wlr_files += files(
 	'wlr_xdg_decoration_v1.c',
 	'wlr_xdg_output_v1.c',
 )
+if uuid.found() or uuid_create
+	conf_data.set10('WLR_HAS_XDG_FOREIGN', true)
+	wlr_files += files(
+		'wlr_xdg_foreign_v1.c',
+		'wlr_xdg_foreign_v2.c',
+	)
+endif
--- a/types/wlr_xdg_foreign_v1.c
+++ b/types/wlr_xdg_foreign_v1.c
@@ -1,0 +1,491 @@
+#define _POSIX_C_SOURCE 200809L
+
+#include <assert.h>
+#include <string.h>
+#include <stdlib.h>
+#include <wlr/types/wlr_xdg_foreign_v1.h>
+#include <wlr/types/wlr_xdg_shell.h>
+#include <wlr/types/wlr_xdg_shell_v6.h>
+#include <wlr/util/log.h>
+#include "util/signal.h"
+#include "util/uuid.h"
+#include "xdg-foreign-unstable-v1-protocol.h"
+
+#define FOREIGN_V1_VERSION 1
+
+static const struct zxdg_exported_v1_interface xdg_exported_impl;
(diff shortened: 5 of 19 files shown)


include/wlr/types/wlr_xdg_foreign_v1.h
```diff
@@ -0,0 +1,75 @@
...
+struct wlr_xdg_imported_v1 {
+	struct wlr_xdg_exported_v1 *exported;
+	struct wl_resource *resource;
+
+	struct wl_list export_link; // wlr_xdg_exported_v1::imports
+	struct wl_list link; // wlr_xdg_importer_v1::imports
+
+	struct wl_listener child_unmap;
+
+	struct wlr_surface *child;
```
ammen99: To my understanding, the protocol allows multiple `set_parent_of` calls on the same `xdg_imported`, however here you only keep track of one child.

To clarify, I think the following is possible:

1. Create dialogues A and B
2. Import surface C
3. Set C as parent of A
4. Set C as parent of B

And now  both A and B are children of C.

types/wlr_xdg_foreign_v1.c
```diff
@@ -0,0 +1,437 @@
...
+	if (exported == NULL) {
+		wl_client_post_no_memory(wl_client);
+		return;
+	}
+
+	exported->surface = surface;
+	exported->resource = wl_resource_create(wl_client,
+		&zxdg_exported_v1_interface, wl_resource_get_version(exporter->resource),
+		id);
+	exported->handle = calloc(11, sizeof(char));
```
ammen99: Why not use `sizeof(exported) + 3`?

emersion:
>Combine the v1 and v2 implementations instead of copying the file? Both implementations are exactly the same.

Please keep them separate. It makes it easier to drop v1 when needed.

I wonder if it makes sense to have v1 at all. It would be nicer to update GTK to use v2.

>Use an actually secure way of generating handles, maybe use libuuid?

Yeah, sounds like a good idea.

>Store the handles in a hash table?

Let's keep this for later.

ddevault:
libuuid is, annoyingly, part of util-linux and not the most portable thing in the world. Can we instead set up a file full of platform-specific getrandom implementations, with a fallback that uses psuedorandom numbers?

types/wlr_xdg_foreign_v1.c
ddevault: `sizeof(char)` is guaranteed to be 1.

emersion:
I thought it was fine because FreeBSD has it: https://www.freshports.org/misc/e2fsprogs-libuuid/

ddevault:
Well, I don't want to make the porting story even more difficult for new platforms. Maybe we make it optional and just disable xdg-foreign if it's not supported?

emersion:
Fine by me

types/wlr_xdg_foreign_v1.c
ammen99: I was speaking about the 11

ddevault: I know, there are two problems on this line.

ammen99: oh right :)

cyclopsian:
The C libraries on [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=uuid_create), [OpenBSD](https://man.openbsd.org/uuid_create), [NetBSD](https://netbsd.gw.com/cgi-bin/man-cgi?uuid_create) all implement uuid generation functions, so libuuid is not necessary there. Is there anything else that needs to be supported?

types/wlr_xdg_foreign_v1.c
cyclopsian: Sorry I should have left a comment. This code is a placeholder and will be removed.

cyclopsian:
The alternative is to use getrandom on linux and arc4random on the BSDs. We probably should just do that to avoid any external dependencies.

ddevault:
Nah, I'd rather use a UUID. Let's add a shim somewhere, choose the appropriate implementation based on the platform, and fallback to removing xdg-foreign support on unsupported platforms.

include/wlr/types/wlr_xdg_foreign_v1.h
cyclopsian: This has been replaced with a list of children. It's necessary to reset the parents after the imported resource gets destroyed.

types/wlr_xdg_foreign_v1.c
cyclopsian: Replaced with the UUID generation shim.

cyclopsian:
This PR has been updated with UUIDs and with fixes for some bugs and crashes. Now the build will check for libuuid (preferred) or uuid_create. If neither is found, it disables xdg-foreign. This can be checked with  WLR_HAS_XDG_FOREIGN in wlr/config.h. The sway branch I made has been updated to reflect this.

For verification I've been using a [test program](https://gist.github.com/cyclopsian/22a8e5632f0be3bdd708d63e0033ba30) to try out destroying and unexporting the parent windows while they still have imported handles open. It's GTK so it tests with the v1 protocol. The code is exactly the same between v1 and v2 so this should be adequate. Eventually I may get around to sending patches to GTK and Mutter to support v2.

ddevault:
Nice. Can you make an issue for tracking the status of v1 support throughout the ecosystem so we can manage its deprecation?

types/wlr_xdg_foreign_v1.c
```diff
@@ -0,0 +1,491 @@
...
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource, &zxdg_exporter_v1_interface,
+		&xdg_exporter_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void disconnect_imported(struct wlr_xdg_imported_v1 *imported) {
+	if (imported->exported != NULL) {
+		imported->exported = NULL;
+		zxdg_imported_v1_send_destroyed(imported->resource);
```
ammen99: My understanding is that sending destroyed doesn't necessarily destroy the `imported`. So if the importing client delays the destruction of the `imported` resource, parent-child relationships won't be removed even though they should be.

cyclopsian: Sorry for the delay. This was refactored, please check to see if it still has the same issue.

cyclopsian:
@ddevault I am working on getting v2 support merged into Mutter and GTK, do you want me to make the issue now?

cyclopsian:
Don't merge this yet, the two versions of the same protocol need to refer to the same data structures internally. Otherwise things will just silently not work e.g. if a program inside a flatpak speaks v1 and the importing program speaks v2.

util/uuid.c
```diff
@@ -0,0 +1,32 @@
+#include "util/uuid.h"
+
+#include <uuid.h>
```
emersion: This should be guarded behind `HAS_LIBUUID`.

util/meson.build
```diff
@@ -5,3 +5,14 @@ wlr_files += files(
...
 )
+
+if uuid.found()
+	wlr_deps += uuid
+	add_project_arguments('-DHAS_LIBUUID', language: 'c')
+endif
+
+if conf_data.get('WLR_HAS_XDG_FOREIGN', 0) == 1
+	wlr_files += files(
+			'uuid.c',
```
emersion: Indentation is weird here. Can be on a single line.

types/meson.build
```diff
@@ -68,3 +68,10 @@ wlr_files += files(
 	'wlr_xdg_decoration_v1.c',
 	'wlr_xdg_output_v1.c',
 )
+if uuid.found() or uuid_create
```
emersion: Hmm. This means we'll sometimes ship the header files without actually shipping the implementation.

Maybe we can use `exclude_files` in `install_subdir` to prevent this?

meson.build
```diff
@@ -107,6 +108,8 @@ libinput = dependency('libinput', version: '>=1.9.0')
 xkbcommon = dependency('xkbcommon')
 udev = dependency('libudev')
 pixman = dependency('pixman-1')
+uuid = dependency('uuid', required: false)
+uuid_create = cc.has_function('uuid_create')
```
emersion: `uuid.h` needs to be included.

util/uuid.c
emersion: Oh, I guess both libuuid and FreeBSD's libc use `uuid.h`. Weird, please disregard this comment.

.builds/freebsd.yml
```diff
@@ -10,6 +10,7 @@ packages:
 - graphics/png
 - graphics/wayland
 - graphics/wayland-protocols
+- misc/e2fsprogs-libuuid
```
emersion: This shouldn't be required. FreeBSD ships `uuid_create` with the libc.

types/meson.build
emersion: Bleh, already done

emersion:
Gentle ping: do you plan to continue working on this PR anytime soon?

cyclopsian:
I'm honestly not sure if there is a clean way to do this. Supporting both versions really complicates the code for very little benefit (the only change between v1 and v2 is the name of some symbols) and it seems like the easiest thing to do is just say with v1.

emersion:
I'd prefer to only support v2. But this doesn't fix the root issue: we'll need an upgrade path for v3 or the stable version of the protocol.

Maybe it would make sense to make v1 lean on v2, ie. make `wlr_xdg_foreign_v1_create` take a `wlr_xdg_foreign_v2` as an argument and re-use its data structures? But then v2 will get to see v1 resources, not great. Except if we add a way to register new xdg_foreign_v2 entries without providing a resource (this increases complexity).

ammen99:
> Maybe it would make sense to make v1 lean on v2, ie. make `wlr_xdg_foreign_v1_create` take a `wlr_xdg_foreign_v2` as an argument and re-use its data structures?

What about using an `impl` struct which does the correct call to either v1 or v2? And then uniting the support for v1 and v2 in the same wlroots object(s). It is unlikely that a compositor wants to export v1 but not v2 or vice-versa.

> But then v2 will get to see v1 resources, not great.

Could this be resolved by adding a tag to each object, whether it belongs to v1, v2 or potentially v3/stable?
Then, if v2 client wants to set parent which uses v1, this could be denied.

zzag:
> I'd prefer to only support v2. But this doesn't fix the root issue: we'll need an upgrade path for v3 or the stable version of the protocol.
> 
> Maybe it would make sense to make v1 lean on v2, ie. make `wlr_xdg_foreign_v1_create` take a `wlr_xdg_foreign_v2` as an argument and re-use its data structures? But then v2 will get to see v1 resources, not great. Except if we add a way to register new xdg_foreign_v2 entries without providing a resource (this increases complexity).

Alternatively, you could introduce a new type that is basically a hash table (handle -> `wlr_surface`), e.g. `wlr_foreign_registry`, and make `wlr_xdg_foreign_v1_create()` and `wlr_xdg_foreign_v2_create()` take it.

However, it raises a philosophical question: do we actually want different versions of the same protocol have compatibility. Currently, both v1 and v2 are very similar, but it's not guaranteed with future versions (hopefully, there won't be that many).

ammen99:
> However, it raises a philosophical question: do we actually want different versions of the same protocol have compatibility. Currently, both v1 and v2 are very similar, but it's not guaranteed with future versions (hopefully, there won't be that many).

I've also thought about that, are there any compositors that support both v1 and v2? Do they allow this or not?

ammen99:
GNOME is going to let clients mix the two protocols, from IRC:

>10:40:58 <ammen99_> jadahl: do you plan to allow clients to use v1 and v2 mixed?
> 10:41:05 <ammen99_> i.e one exports v1 and other imports with v2 
> 10:41:21 <jadahl> ammen99_: yes

I suppose the use-case would be one older application from flatpak and a modern portal impl or something like that.

emersion:
If we really wanted to support both, we could for instance make v1 an "add-on" for v2: all of the data structures live in v2, and v1 is just a proxy to v2.

emersion:
Having v1 and v2 has two completely separate files makes it easy to remove v1 once everybody has migrated to v2. It's not unheard of for a compositor to drop v1 and keep v2, for instance Sway has removed xdg-shell-unstable-v6 support while wlroots still supports it.

emersion:
>Alternatively, you could introduce a new type that is basically a hash table (handle -> wlr_surface), e.g. wlr_foreign_registry, and make wlr_xdg_foreign_v1_create() and wlr_xdg_foreign_v2_create() take it.

Sounds like a reasonable idea as well, and more or less matches what we've done for e.g. primary selection.

@ammen99 @zzag If you want to move this forward, feel free to adopt this PR, I don't think @cyclopsian has plans to resume work in the near future.

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2487

PR #1843, opened 2019-10-05
xwayland: Expose configure mask

soreau:
This allows compositors to know which elements are valid. Without this,
it is impossible to tell which are invalid. This means if only position or size
are valid for instance, the compositor can discard the values of the rest.

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -179,6 +179,7 @@ struct wlr_xwayland_surface_configure_event {
 	struct wlr_xwayland_surface *surface;
 	int16_t x, y;
 	uint16_t width, height;
+	uint16_t mask; // xcb_config_window_t
 };
 
 // TODO: maybe add a seat to these
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -871,6 +871,7 @@ static void xwm_handle_configure_request(struct wlr_xwm *xwm,
 		.y = mask & XCB_CONFIG_WINDOW_Y ? ev->y : surface->y,
 		.width = mask & XCB_CONFIG_WINDOW_WIDTH ? ev->width : surface->width,
 		.height = mask & XCB_CONFIG_WINDOW_HEIGHT ? ev->height : surface->height,
+		.mask = mask,
 	};
 	wlr_log(WLR_DEBUG, "XCB_CONFIGURE_REQUEST (%u) [%ux%u+%d,%d]", ev->window,
 		wlr_event.width, wlr_event.height, wlr_event.x, wlr_event.y);


xwayland/xwm.c
```diff
@@ -1408,16 +1409,17 @@ void wlr_xwayland_surface_activate(struct wlr_xwayland_surface *xsurface,
 }
 
 void wlr_xwayland_surface_configure(struct wlr_xwayland_surface *xsurface,
-		int16_t x, int16_t y, uint16_t width, uint16_t height) {
-	xsurface->x = x;
-	xsurface->y = y;
-	xsurface->width = width;
-	xsurface->height = height;
+		int16_t x, int16_t y, uint16_t width, uint16_t height, uint32_t mask) {
+	if (mask & XCB_CONFIG_WINDOW_X)
```
ammen99: style issues (same below): add {

include/wlr/xwayland.h
```diff
@@ -212,7 +213,7 @@ void wlr_xwayland_surface_activate(struct wlr_xwayland_surface *surface,
 	bool activated);
 
 void wlr_xwayland_surface_configure(struct wlr_xwayland_surface *surface,
-	int16_t x, int16_t y, uint16_t width, uint16_t height);
+	int16_t x, int16_t y, uint16_t width, uint16_t height, uint32_t mask);
```
ammen99: Maybe make `mask` a `uint16_t` the XCB function takes the mask as `uint16_t`

xwayland/xwm.c
soreau: Fixed.

include/wlr/xwayland.h
soreau: Done.

emersion:
>Without this, it is impossible
to configure position and size independently.

Why is this needed?

soreau:
> > Without this, it is impossible
> > to configure position and size independently.
> 
> Why is this needed?

One reason is that there are times when the size shouldn't be propagated when the position is changed. This can cause some clients to start with an unintended size.

emersion:
Instead, what about adding each member that changed to the mask in `wlr_xwayland_surface_configure`?

soreau:
It isn't clear what you mean because the latter patch here adds the mask to `wlr_xwayland_surface_configure`. What are you suggesting exactly?

ammen99:
@soreau  I think the idea is to compare the function parameters to `xsurface->x/y/width/height` and generate a mask from this. Which kinda makes sense.

soreau:
@ammen99 Ah I see, that does make sense.

soreau:
But from my initial tests, it does not work. It still propagates a wrong size.

emersion:
Can you provide reproduction steps for this bug?

ammen99:
@emersion May I ask why not pass a flag?

ammen99:
> Using this branch with http://ix.io/1XWy in place of the second patch, checkout wayfire fix-307 and run xterm.

Actually we can add workarounds in Wayfire if wlroots tries to autodetect which values have changed, but still I see no reason to not include flags, it makes things clearer and reflects the underlying APIs better.

emersion:
Just trying to understand why we don't have this issue in sway or rootston.

soreau:
@ammen99 Found the real bug in wayfire, so the second patch isn't needed.

ammen99:
For the record, the underlying problem was that the compositor needs to respond to ConfigureRequest with a configure, even when the surface is unmapped.

emersion:
Okay.

So what are you using the mask for in Wayfire?

ammen99:
@emersion So that we know whether the client wants to configure its position or not.

emersion:
I'm worried about adding an internal XCB flag like this.

Can't you just compare `event->width` with `xwayland_surface->width`?

ammen99:
@emersion, how do I know the difference between the client wanting to be positioned at 0,0 and client which has configured only width and height? (I assume x/y are 0 initially)

ammen99:
Not to mention the whole wlr_xwayland_surface is full of fields with xcb-specific types.

include/wlr/xwayland.h
```diff
@@ -179,6 +179,7 @@ struct wlr_xwayland_surface_configure_event {
 	struct wlr_xwayland_surface *surface;
 	int16_t x, y;
 	uint16_t width, height;
+	uint16_t mask;
```
emersion: Can you add a little `xcb_config_window_t` comment to document where the bitfield values are coming from?

emersion:
Well, okay, I'm fine with that. I still don't get why you need this information though.

ammen99:
@emersion, I'll try to explain. When opening a new application, Wayfire has a plugin which will try to position it around the screen (there are different modes, like center application, cascade, or random).

However some applications (Xwayland ones) want to position themselves, for example Steam. And Wayfire needs to know whether the client has specified a position manually or not, in order to decide how to place the window.

emersion:
Oh, right, this makes sense. Thanks for the explanation.

include/wlr/xwayland.h
soreau: Done.

Issue #1840, opened 2019-10-03
Let the driver choose our GBM buffer modifier


emersion:
See https://gitlab.freedesktop.org/daniels/kms-quads/blob/master/egl-gles.c#L573

PR #1844, opened 2019-10-06
backend/drm: use modifiers for our GBM buffers

emersion:
On my Intel GPU, the primary plane ends up using X_TILED instead of LINEAR.

Closes: https://github.com/swaywm/wlroots/issues/1840


ascent12:
Thanks

PR #1845, opened 2019-10-06
render/egl: support formats with zero modifiers

emersion:
There's some ongoing work to advertise the modifiers EGL extension even if modifiers aren't supported so that we can use `eglQueryDmaBufFormatsEXT`. However we need to be prepared for `eglQueryDmaBufModifiersEXT` to return zero modifiers.


ascent12:
Thanks

PR #1847, opened 2019-10-08
build: simplify by using disabler deps

emersion:
See https://github.com/mesonbuild/meson/issues/6010#issuecomment-539580547


ddevault:
Thanks!

PR #1848, opened 2019-10-08
build: workaround for meson disabler object not working with if not

emersion:
Successful wlroots build in Sway CI: https://builds.sr.ht/~emersion/job/95358#task-wlroots

(The failure at the sway stage is fixed by https://github.com/swaywm/sway/pull/4615)


ddevault:
Thanks!

PR #1850, opened 2019-10-11
wlr_box: properly calculate closest point for non-positive area boxes

ammen99:
If box->width/height is <= 0, the box doesn't contain any points, and so there is no closest point. wlr_box_closest_point should return NAN in this case.

In addition, we need to handle empty boxes in a few other output-layout-related places, because outputs can have size 0x0 when they are created or destroyed.

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -328,7 +328,7 @@ void wlr_output_layout_closest_point(struct wlr_output_layout *layout,
 		return;
 	}
 
-	double min_x = DBL_MAX, min_y = DBL_MAX, min_distance = DBL_MAX;
+	double min_x = 0, min_y = 0, min_distance = DBL_MAX;
 	struct wlr_output_layout_output *l_output;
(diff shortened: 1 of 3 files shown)


types/wlr_output_layout.c
```diff
@@ -328,7 +328,7 @@ void wlr_output_layout_closest_point(struct wlr_output_layout *layout,
 		return;
 	}
 
-	double min_x = DBL_MAX, min_y = DBL_MAX, min_distance = DBL_MAX;
+	double min_x = 0, min_y = 0, min_distance = DBL_MAX;
```
emersion: What happens if the minimum x coord is > 0? This will incorrectly set it as zero, right?

ammen99: No, because you compare `min_distance` first, and `min_distance` is DBL_MAX initially. So any point will be closer to a rectangle than this.

emersion: Oh right, the min x/y are not used in the loop, only the min distance is used.

PR #1851, opened 2019-10-12
examples/gamma-control.c: add parameter to set different gamma for rgb

brandsimon:
I wanted to modify my gamma value for each rgb value, so I changed to code to enable this.
If there is only one value provided for r:g:b, it is backwards compatible.
I'm happy to adapt the PR, if there is anything needed to do for a merge.

### DIFF
--- a/examples/gamma-control.c
+++ b/examples/gamma-control.c
@@ -108,20 +108,29 @@ static const struct wl_registry_listener registry_listener = {
 	.global_remove = registry_handle_global_remove,
 };
 
+static uint16_t cell_value(double orig, double contrast,
+		double brightness, double gamma) {
+	double result = contrast * pow(orig, gamma) + (brightness - 1);
+	return (uint16_t)(fmax(fmin(result,
+	                            1.0),
+	                       0.0) * UINT16_MAX);
+}
+
 static void fill_gamma_table(uint16_t *table, uint32_t ramp_size,
-		double contrast, double brightness, double gamma) {
+		double contrast, double brightness,
+		double gamma_red, double gamma_green, double gamma_blue) {
 	uint16_t *r = table;
 	uint16_t *g = table + ramp_size;
 	uint16_t *b = table + 2 * ramp_size;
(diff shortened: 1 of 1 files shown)


brandsimon:
I forgot to mention, I tested the values and they are exactly the same as with `xrandr --gamma r:g:b` on my system.

ddevault:
The examples are just that: examples. Not meant for regular use. I would rather see this feature implemented in a more generic, standalone tool for adjusting gamma.

brandsimon:
If it is just an example, why not merge it to show people, that this is possible.
People told me, that this is not possible yet with sway, but obviously, it is.
I agree, a standalone tool would be better.

emersion:
This is an example for developers to test their compositor, not for end users. We don't want to add all possible features to examples.

brandsimon:
I want to test my compositor to handle different gamma values for rgb correctly. :D
Never mind, thank you.

brandsimon:
Hi,

I am creating a separate package for this tool and have two questions:
1. When I add a new display and enable it via sway, gamma is only changed on the old displays, not the newly connected one. Is this a sway issue or do I need to listen to this event and handle it? If so, is there an example which demonstrates how to listen to this event?

2. Needs the wlr-gamma-control-unstable-v1.xml be included in every tool or is there a package, where this should be included? Either sway nor wlroots are installing this file on my system. (Arch Linux)

emersion:
This tool isn't supposed to get installed/packaged. It's just an example, it won't handle hotplug.

brandsimon:
@emersion 
Exactly that's what I am doing, I found the solution for the first question.

progandy:
https://github.com/swaywm/wlr-protocols
Since it isn't packaged, you could either copy it, use a git submodule or a git subtree.

This project might interest you: https://github.com/mherzberg/wlr-brightness

emersion:
Ah, sorry, I misunderstood. Then yeah you need to listen for `wl_output` globals being added on the fly. And @progandy replied to the second question.

PR #1859, opened 2019-10-16
Add new define EGL_NO_X11 for newer mesa library

DanySpin97:
Define both MESA_EGL_NO_X11_HEADERS and EGL_NO_X11 for
backward combatibility.

Newer versions of mesa now define -DEGL_NO_X11 instead of -DMESA_EGL_NO_X11_HEADERS.

### DIFF
--- a/glgen.sh
+++ b/glgen.sh
@@ -64,8 +64,13 @@ cat > "$OUTDIR/$BASE.h" << EOF
 #include <stdbool.h>
 #include <wlr/config.h>
 
-#if !WLR_HAS_X11_BACKEND && !WLR_HAS_XWAYLAND && !defined MESA_EGL_NO_X11_HEADERS
+#if !WLR_HAS_X11_BACKEND && !WLR_HAS_XWAYLAND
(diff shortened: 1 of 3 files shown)


glgen.sh
```diff
@@ -64,9 +64,14 @@ cat > "$OUTDIR/$BASE.h" << EOF
 #include <stdbool.h>
 #include <wlr/config.h>
 
-#if !WLR_HAS_X11_BACKEND && !WLR_HAS_XWAYLAND && !defined MESA_EGL_NO_X11_HEADERS
+#if !WLR_HAS_X11_BACKEND && !WLR_HAS_XWAYLAND
+#ifndef MESA_EGL_NO_X11_HEADERS
```
emersion: Why do we need these `#ifndef`s?

emersion: Ah, because multiple of these headers might be included at te same time.

emersion:
Thanks!

glgen.sh
DanySpin97: Otherwise redefining one of them would make the build fails.

PR #1860, opened 2019-10-17
protocol: sync layer-shell with upstream

emersion:
Fixes this warning:

    ../protocol/wlr-layer-shell-unstable-v1.xml:241: warning: since version not increasing

cc @Timidger


ddevault:
Whoops, missed that.

PR #1864, opened 2019-10-19
Implement input-method keyboard grab

xdavidwu:
I'm interested in implementing CJK input method in wayland so I try to implement this.
I've tested with my local client code but it's not suitable for a test example.
Should I extend examples/input-method.c for this or write a new one

### DIFF
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -170,6 +170,13 @@ clients = {
 		'dep': wlroots,
 		'proto': ['wlr-virtual-pointer-unstable-v1'],
 	},
+	'input-method-keyboard-grab': {
+		'src': 'input-method-keyboard-grab.c',
+		'dep': xkbcommon,
+		'proto': [
+			'input-method-unstable-v2',
+		],
+	},
 }
 
 foreach name, info : compositors
--- a/include/wlr/types/wlr_input_method_v2.h
+++ b/include/wlr/types/wlr_input_method_v2.h
@@ -34,6 +34,7 @@ struct wlr_input_method_v2 {
 	struct wl_resource *resource;
 
 	struct wlr_seat *seat;
+	struct wlr_seat_client *seat_client;
 
 	struct wlr_input_method_v2_state pending;
 	struct wlr_input_method_v2_state current;
@@ -41,13 +42,30 @@ struct wlr_input_method_v2 {
 	bool client_active; // state known to the client
 	uint32_t current_serial; // received in last commit call
 
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab;
+
 	struct wl_list link;
 
-	struct wl_listener seat_destroy;
+	struct wl_listener seat_client_destroy;
 
 	struct {
 		struct wl_signal commit; // (struct wlr_input_method_v2*)
+		struct wl_signal grab_keyboard; // (struct wlr_input_method_keyboard_grab_v2*)
 		struct wl_signal destroy; // (struct wlr_input_method_v2*)
+	} events;
+};
+
+struct wlr_input_method_keyboard_grab_v2 {
+	struct wl_resource *resource;
+	struct wlr_input_method_v2 *input_method;
+	struct wlr_keyboard *keyboard;
+
+	struct wl_listener keyboard_keymap;
+	struct wl_listener keyboard_repeat_info;
+	struct wl_listener keyboard_destroy;
+
+	struct {
+		struct wl_signal destroy; // (struct wlr_input_method_keyboard_grab_v2*)
 	} events;
 };
 
@@ -81,5 +99,17 @@ void wlr_input_method_v2_send_text_change_cause(
 void wlr_input_method_v2_send_done(struct wlr_input_method_v2 *input_method);
 void wlr_input_method_v2_send_unavailable(
 	struct wlr_input_method_v2 *input_method);
+
+void wlr_input_method_keyboard_grab_v2_send_key(
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+	uint32_t time, uint32_t key, uint32_t state);
+void wlr_input_method_keyboard_grab_v2_send_modifiers(
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+	struct wlr_keyboard_modifiers *modifiers);
+void wlr_input_method_keyboard_grab_v2_set_keyboard(
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+	struct wlr_keyboard *keyboard);
+void wlr_input_method_keyboard_grab_v2_destroy(
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab);
 
 #endif
--- a/types/wlr_input_method_v2.c
+++ b/types/wlr_input_method_v2.c
@@ -3,15 +3,19 @@
 #endif
 #include <assert.h>
 #include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
 #include <wayland-util.h>
 #include <wlr/types/wlr_input_method_v2.h>
 #include <wlr/types/wlr_surface.h>
 #include <wlr/util/log.h>
 #include <xkbcommon/xkbcommon.h>
 #include "input-method-unstable-v2-protocol.h"
+#include "util/shm.h"
 #include "util/signal.h"
 
 static const struct zwp_input_method_v2_interface input_method_impl;
+static const struct zwp_input_method_keyboard_grab_v2_interface keyboard_grab_impl;
 
 static struct wlr_input_method_v2 *input_method_from_resource(
 		struct wl_resource *resource) {
@@ -20,10 +24,19 @@ static struct wlr_input_method_v2 *input_method_from_resource(
 	return wl_resource_get_user_data(resource);
 }
 
+static struct wlr_input_method_keyboard_grab_v2 *keyboard_grab_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource,
+		&zwp_input_method_keyboard_grab_v2_interface,
+		&keyboard_grab_impl));
+	return wl_resource_get_user_data(resource);
+}
+
 static void input_method_destroy(struct wlr_input_method_v2 *input_method) {
 	wlr_signal_emit_safe(&input_method->events.destroy, input_method);
 	wl_list_remove(wl_resource_get_link(input_method->resource));
-	wl_list_remove(&input_method->seat_destroy.link);
+	wl_list_remove(&input_method->seat_client_destroy.link);
+	wlr_input_method_keyboard_grab_v2_destroy(input_method->keyboard_grab);
 	free(input_method->pending.commit_text);
 	free(input_method->pending.preedit.text);
 	free(input_method->current.commit_text);
@@ -101,10 +114,189 @@ static void im_get_input_popup_surface(struct wl_client *client,
 	wlr_log(WLR_INFO, "Stub: zwp_input_method_v2::get_input_popup_surface");
 }
 
+void wlr_input_method_keyboard_grab_v2_destroy(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab) {
+	if (!keyboard_grab) {
+		return;
+	}
+	wlr_signal_emit_safe(&keyboard_grab->events.destroy, keyboard_grab);
+	keyboard_grab->input_method->keyboard_grab = NULL;
+	if (keyboard_grab->keyboard) {
+		wl_list_remove(&keyboard_grab->keyboard_keymap.link);
+		wl_list_remove(&keyboard_grab->keyboard_repeat_info.link);
+		wl_list_remove(&keyboard_grab->keyboard_destroy.link);
+	}
+	wl_resource_set_user_data(keyboard_grab->resource, NULL);
+	free(keyboard_grab);
+}
+
+static void keyboard_grab_resource_destroy(struct wl_resource *resource) {
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+		keyboard_grab_from_resource(resource);
+	wlr_input_method_keyboard_grab_v2_destroy(keyboard_grab);
+}
+
+static void keyboard_grab_release(struct wl_client *client,
+		struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static const struct zwp_input_method_keyboard_grab_v2_interface keyboard_grab_impl = {
+	.release = keyboard_grab_release,
+};
+
+void wlr_input_method_keyboard_grab_v2_send_key(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+		uint32_t time, uint32_t key, uint32_t state) {
+	zwp_input_method_keyboard_grab_v2_send_key(
+		keyboard_grab->resource,
+		wlr_seat_client_next_serial(keyboard_grab->input_method->seat_client),
+		time, key, state);
+}
+
+void wlr_input_method_keyboard_grab_v2_send_modifiers(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+		struct wlr_keyboard_modifiers *modifiers) {
+	zwp_input_method_keyboard_grab_v2_send_modifiers(
+		keyboard_grab->resource,
+		wlr_seat_client_next_serial(keyboard_grab->input_method->seat_client),
+		modifiers->depressed, modifiers->latched,
+		modifiers->locked, modifiers->group);
+}
+
+static bool keyboard_grab_send_keymap(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+		struct wlr_keyboard *keyboard) {
+	int keymap_fd = allocate_shm_file(keyboard->keymap_size);
+	if (keymap_fd < 0) {
+		wlr_log(WLR_ERROR, "creating a keymap file for %zu bytes failed",
+			keyboard->keymap_size);
+		return false;
+	}
+
+	void *ptr = mmap(NULL, keyboard->keymap_size, PROT_READ | PROT_WRITE,
+		MAP_SHARED, keymap_fd, 0);
+	if (ptr == MAP_FAILED) {
+		wlr_log(WLR_ERROR, "failed to mmap() %zu bytes",
+			keyboard->keymap_size);
+		close(keymap_fd);
+		return false;
+	}
+
+	memcpy(ptr, keyboard->keymap_string, keyboard->keymap_size);
+	munmap(ptr, keyboard->keymap_size);
+
+	zwp_input_method_keyboard_grab_v2_send_keymap(keyboard_grab->resource,
+		WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1, keymap_fd,
+		keyboard->keymap_size);
+
+	close(keymap_fd);
+	return true;
+}
+
+static void keyboard_grab_send_repeat_info(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+		struct wlr_keyboard *keyboard) {
+	zwp_input_method_keyboard_grab_v2_send_repeat_info(
+		keyboard_grab->resource, keyboard->repeat_info.rate,
+		keyboard->repeat_info.delay);
+}
+
+static void handle_keyboard_keymap(struct wl_listener *listener, void *data) {
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+		wl_container_of(listener, keyboard_grab, keyboard_keymap);
+	keyboard_grab_send_keymap(keyboard_grab, data);
+}
+
+static void handle_keyboard_repeat_info(struct wl_listener *listener,
+		void *data) {
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+		wl_container_of(listener, keyboard_grab, keyboard_repeat_info);
+	keyboard_grab_send_repeat_info(keyboard_grab, data);
+}
+
+static void handle_keyboard_destroy(struct wl_listener *listener,
+		void *data) {
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+		wl_container_of(listener, keyboard_grab, keyboard_destroy);
+	wlr_input_method_keyboard_grab_v2_set_keyboard(keyboard_grab, NULL);
+}
+
+void wlr_input_method_keyboard_grab_v2_set_keyboard(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab,
+		struct wlr_keyboard *keyboard) {
+	if (keyboard == keyboard_grab->keyboard) {
+		return;
+	}
+
+	if (keyboard_grab->keyboard) {
+		wl_list_remove(&keyboard_grab->keyboard_keymap.link);
+		wl_list_remove(&keyboard_grab->keyboard_repeat_info.link);
+		wl_list_remove(&keyboard_grab->keyboard_destroy.link);
+	}
+
+	if (keyboard) {
+		if (keyboard_grab->keyboard == NULL ||
+				strcmp(keyboard_grab->keyboard->keymap_string,
+				keyboard->keymap_string) != 0) {
+			// send keymap only if it is changed, or if input method is not
+			// aware that it did not change and blindly send it back with
+			// virtual keyboard, it may cause an infinite recursion.
+			if (!keyboard_grab_send_keymap(keyboard_grab, keyboard)) {
+				wlr_log(WLR_ERROR, "Failed to send keymap for input-method keyboard grab");
+				return;
+			}
+		}
+		keyboard_grab_send_repeat_info(keyboard_grab, keyboard);
+		keyboard_grab->keyboard_keymap.notify = handle_keyboard_keymap;
+		wl_signal_add(&keyboard->events.keymap,
+			&keyboard_grab->keyboard_keymap);
+		keyboard_grab->keyboard_repeat_info.notify =
+			handle_keyboard_repeat_info;
+		wl_signal_add(&keyboard->events.repeat_info,
+			&keyboard_grab->keyboard_repeat_info);
+		keyboard_grab->keyboard_destroy.notify =
+			handle_keyboard_destroy;
+		wl_signal_add(&keyboard->events.destroy,
+			&keyboard_grab->keyboard_destroy);
+	}
+
+	keyboard_grab->keyboard = keyboard;
+};
 
 static void im_grab_keyboard(struct wl_client *client,
 		struct wl_resource *resource, uint32_t keyboard) {
-	wlr_log(WLR_INFO, "Stub: zwp_input_method_v2::grab_keyboard");
+	struct wlr_input_method_v2 *input_method =
+		input_method_from_resource(resource);
+	if (!input_method) {
+		return;
+	}
+	if (input_method->keyboard_grab) {
+		// Already grabbed
+		return;
+	}
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+		calloc(1, sizeof(struct wlr_input_method_keyboard_grab_v2));
+	if (!keyboard_grab) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+	struct wl_resource *keyboard_grab_resource = wl_resource_create(
+		client, &zwp_input_method_keyboard_grab_v2_interface,
+		wl_resource_get_version(resource), keyboard);
+	if (keyboard_grab_resource == NULL) {
+		free(keyboard_grab);
+		wl_client_post_no_memory(client);
+		return;
+	}
+	wl_resource_set_implementation(keyboard_grab_resource,
+		&keyboard_grab_impl, keyboard_grab,
+		keyboard_grab_resource_destroy);
+	keyboard_grab->resource = keyboard_grab_resource;
+	keyboard_grab->input_method = input_method;
+	input_method->keyboard_grab = keyboard_grab;
+	wl_signal_init(&keyboard_grab->events.destroy);
+	wlr_signal_emit_safe(&input_method->events.grab_keyboard, keyboard_grab);
 }
 
 static const struct zwp_input_method_v2_interface input_method_impl = {
@@ -176,10 +368,10 @@ static struct wlr_input_method_manager_v2 *input_method_manager_from_resource(
 	return wl_resource_get_user_data(resource);
 }
 
-static void input_method_handle_seat_destroy(struct wl_listener *listener,
+static void input_method_handle_seat_client_destroy(struct wl_listener *listener,
 		void *data) {
 	struct wlr_input_method_v2 *input_method = wl_container_of(listener,
-		input_method, seat_destroy);
+		input_method, seat_client_destroy);
 	wlr_input_method_v2_send_unavailable(input_method);
 }
 
@@ -196,6 +388,7 @@ static void manager_get_input_method(struct wl_client *client,
 		return;
 	}
 	wl_signal_init(&input_method->events.commit);
+	wl_signal_init(&input_method->events.grab_keyboard);
 	wl_signal_init(&input_method->events.destroy);
 	int version = wl_resource_get_version(resource);
 	struct wl_resource *im_resource = wl_resource_create(client,
@@ -208,14 +401,14 @@ static void manager_get_input_method(struct wl_client *client,
 	wl_resource_set_implementation(im_resource, &input_method_impl,
 		input_method, input_method_resource_destroy);
 
-	struct wlr_seat_client *seat_client = wlr_seat_client_from_resource(seat);
-	wl_signal_add(&seat_client->events.destroy,
-		&input_method->seat_destroy);
-	input_method->seat_destroy.notify =
-		input_method_handle_seat_destroy;
+	input_method->seat_client = wlr_seat_client_from_resource(seat);
+	input_method->seat = input_method->seat_client->seat;
+	wl_signal_add(&input_method->seat_client->events.destroy,
+		&input_method->seat_client_destroy);
+	input_method->seat_client_destroy.notify =
+		input_method_handle_seat_client_destroy;
 
 	input_method->resource = im_resource;
-	input_method->seat = seat_client->seat;
 	wl_list_insert(&im_manager->input_methods,
 		wl_resource_get_link(input_method->resource));
 	wlr_signal_emit_safe(&im_manager->events.input_method, input_method);
--- a/examples/input-method-keyboard-grab.c
+++ b/examples/input-method-keyboard-grab.c
@@ -1,0 +1,217 @@
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+#include <wayland-client.h>
+#include <xkbcommon/xkbcommon.h>
+#include "input-method-unstable-v2-client-protocol.h"
+
+static struct wl_display *display = NULL;
+static struct wl_seat *seat = NULL;
+static struct zwp_input_method_manager_v2 *input_method_manager = NULL;
+static struct zwp_input_method_v2 *input_method = NULL;
+static struct zwp_input_method_keyboard_grab_v2 *kb_grab = NULL;
+
+static bool active = false;
+static bool pending_active = false;
+
+static struct xkb_context *xkb_context = NULL;
+static struct xkb_keymap *keymap = NULL;
+static struct xkb_state *xkb_state = NULL;
+
+static void handle_key(void *data,
+		struct zwp_input_method_keyboard_grab_v2 *im_keyboard_grab,
+		uint32_t serial, uint32_t time, uint32_t key, uint32_t state) {
+	printf("handle_key %u %u %u %u\n", serial, time, key, state);
+	xkb_keysym_t keysym = xkb_state_key_get_one_sym(xkb_state, key + 8);
+	char keysym_name[64];
+	xkb_keysym_get_name(keysym, keysym_name, sizeof(keysym_name));
+	printf("xkb translated to %s\n", keysym_name);
+	if (state == WL_KEYBOARD_KEY_STATE_PRESSED) {
+		if (keysym == XKB_KEY_KP_Enter || keysym == XKB_KEY_Return) {
+			printf("Stopping grab\n");
+			zwp_input_method_keyboard_grab_v2_release(kb_grab);
+			kb_grab = NULL;
+		}
+	}
+
+}
+
+static void handle_modifiers(void *data,
+		struct zwp_input_method_keyboard_grab_v2 *im_keyboard_grab,
+		uint32_t serial, uint32_t mods_depressed, uint32_t mods_latched,
+		uint32_t mods_locked, uint32_t group) {
+	printf("handle_modifiers %u %u %u %u %u\n", serial, mods_depressed,
+		mods_latched, mods_locked, group);
+	xkb_state_update_mask(xkb_state, mods_depressed, mods_latched,
+		mods_locked, 0, 0, group);
+}
+
+static void handle_keymap(void *data,
+		struct zwp_input_method_keyboard_grab_v2 *im_keyboard_grab,
+		uint32_t format, int32_t fd, uint32_t size) {
+	printf("handle_keymap\n");
+	char *keymap_string = mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0);
+	xkb_keymap_unref(keymap);
+	keymap = xkb_keymap_new_from_string(xkb_context, keymap_string,
+		XKB_KEYMAP_FORMAT_TEXT_V1, XKB_KEYMAP_COMPILE_NO_FLAGS);
+	munmap(keymap_string, size);
+	close(fd);
+	xkb_state_unref(xkb_state);
+	xkb_state = xkb_state_new(keymap);
+}
+
+static void handle_repeat_info(void *data,
+		struct zwp_input_method_keyboard_grab_v2 *im_keyboard_grab,
+		int32_t rate, int32_t delay) {
+	printf("handle_repeat_info %d %d", rate, delay);
+}
+
+
+static const struct zwp_input_method_keyboard_grab_v2_listener grab_listener = {
+	.key = handle_key,
+	.modifiers = handle_modifiers,
+	.keymap = handle_keymap,
+	.repeat_info = handle_repeat_info,
+};
+
+static void handle_activate(void *data,
+		struct zwp_input_method_v2 *zwp_input_method_v2) {
+	pending_active = true;
+}
+
+static void handle_deactivate(void *data,
+		struct zwp_input_method_v2 *zwp_input_method_v2) {
+	pending_active = false;
+}
+
+static void handle_unavailable(void *data,
+		struct zwp_input_method_v2 *zwp_input_method_v2) {
+	printf("IM unavailable\n");
+	zwp_input_method_v2_destroy(zwp_input_method_v2);
+	input_method = NULL;
+}
+
+static void im_activate(void *data,
+		struct zwp_input_method_v2 *id) {
+	kb_grab = zwp_input_method_v2_grab_keyboard(input_method);
+	if (kb_grab == NULL) {
+		fprintf(stderr, "Failed to grab\n");
+		exit(EXIT_FAILURE);
+	}
+	zwp_input_method_keyboard_grab_v2_add_listener(kb_grab, &grab_listener,
+		NULL);
+	printf("Started grab, press enter to stop grab\n");
+}
+
+static void im_deactivate(void *data,
+		struct zwp_input_method_v2 *context) {
+	if (kb_grab != NULL) {
+		zwp_input_method_keyboard_grab_v2_release(kb_grab);
+		kb_grab = NULL;
+	}
+}
+
+static void handle_done(void *data,
+		struct zwp_input_method_v2 *zwp_input_method_v2) {
+	bool prev_active = active;
+	if (active != pending_active) {
+		printf("Now %s\n", pending_active ? "active" : "inactive");
+	}
+	active = pending_active;
+	if (active && !prev_active) {
+		im_activate(data, zwp_input_method_v2);
+	} else if (!active && prev_active) {
+		im_deactivate(data, zwp_input_method_v2);
+	}
+}
+
+static void handle_surrounding_text(void *data,
+		struct zwp_input_method_v2 *zwp_input_method_v2,
+		const char *text, uint32_t cursor, uint32_t anchor) {
+	// not for this test
+}
+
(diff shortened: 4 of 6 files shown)


ddevault:
This looks like a great start. Can you rebase it against master? Whether or not to extend examples/input-method.c or write a new example is at your discretion.

ddevault:
Is there a particular IME software you're looking to implement this for? ibus?

xdavidwu:
I plan to implement a new IME with libchewing, not using existing frameworks like ibus or fcitx.

ddevault:
Oh, nice. Hoping for anthy support too :smile:

xdavidwu:
I found that `virtual-keyboard-unstable-v1` states that it can `accompany the input method protocol`, so that means in input-method keyboard grab, we should not intercept events generated by virtual-keyboard, right?

ddevault:
You should intercept those events, I think. Virtual keyboards behave just like any other kind.

xdavidwu:
Then I think there is no way for a input method to send unhandled key events back to the surface. Should we add an interface to input-method protocol for this?

ddevault:
I'm not sure I fully understand the problem. Can you explain it in more detail?

xdavidwu:
I have been writing a experimental [input-method client](https://github.com/xdavidwu/wlchewing) that grabs, and I found that I need to send key events that are not handled by input-method back to the application for things like moving the cursor to already committed text or some keyboard shortcuts like ctrl+s. If I use virtual-keyboard for that, the events would be caught by the input-method again.

ddevault:
Is the key forwarding event in input-method not sufficient?

xdavidwu:
There is only event that send key to input method, but no request method to send it back (when unhandled) to the surface that requested text input.

ddevault:
https://gitlab.freedesktop.org/wayland/wayland-protocols/blob/master/unstable/input-method/input-method-unstable-v1.xml#L145-196

xdavidwu:
Oh I didn't know that v1 has it. In input-method-unstable-v2 bundled in protocols/ it is missing.

ddevault:
Ah, I see. I think we did split that off, now that you mention it. Let's say that virtual keyboards should bypass input-method, then. Virtual keyboards which want to do i18n input should probably use input-method themselves, I think that was the idea.

xhebox:
@xdavidwu Good work. I am trying to modify wlchewing, or almost rewrite a new one based libpinyin probably. But I found a bug with the current implementation. The workaround patch first:
```
--- wlroots-0.10.0/types/wlr_keyboard_group.c   2020-01-23 02:22:04.000000000 +0800
+++ wlroots-0.10.0/types/wlr_keyboard_group.c   2020-01-23 02:22:04.000000000 +0800
@@ -288,6 +288,9 @@
                                group_kb->repeat_info.delay);
        }
 
+       if (keyboard->virtual_keyboard != NULL)
+               group->keyboard.virtual_keyboard = keyboard->virtual_keyboard;
+
        refresh_state(device, WLR_KEY_PRESSED);
        return true;
 }
```

I am not familiar with the internal mechanics of wlroots, I can not tell the full picture. But, this is basically what happened: I have a native keyboard, and a virtual keyboard. So sway is using keyboard_group to multiplex events. Thus, what we grab is the fake wlr_keyboard in keybaord_group object. But that fake wlr_keyboard was not with virtual_keyboard pointer. So your sway grab code will not work at all.

**BTW**: I do not think the current implementation is "good" enough.
1. It seems that, when you grab the keyboard, keymap events are not ignored like key/modifiers. I still need a check to avoid infinite recursion. And the protocol xml file talks nothing about that. It will be good if all events avoid recursion automatically.
2. Grab implementation in sway is not consistent on the behavior of ignoring. Key events will check the resource while modifiers events are not. Check https://github.com/swaywm/sway/pull/4932/commits/7471b43fdcdd7161c746fba188a3581c79c07344#diff-d7aa39a56641abe2c1362ca922ba299bR453-R455 and https://github.com/swaywm/sway/pull/4932/commits/7471b43fdcdd7161c746fba188a3581c79c07344#diff-d7aa39a56641abe2c1362ca922ba299bR556.
3. I think grab implementation should be at the side of wlroots. It is weird that you expose the pointer of virtual_keyboard. Though with the patch, it now works, there must be a better way to implement, IMO.

**EDIT**: I can confirm that grab interface and ime interface are working in a pure wayland environment(no x11/GLX at all). I can type things in a patched pure-wayland(cairo-gtk3-wayland) firefox. I will continue my work this minimal IME and exam popup interface. Because I really can not find a working native IME.

xdavidwu:
wlr_keyboard_group is a helper to reduce keyboard events. For example, assuming there is a keyboard group g1 consisting of two keyboards k1 and k2, and the original events sequence look like this: (with the same key) k1 pressed, k2 pressed, k1 released, k2 released.
Without helps from wlr_keyboard_group, compositor will send all events to client.
With wlr_keyboard_group, compositor would only send events at the first pressed and the last released.
When keyboard group is enabled, sway listens on both individual and grouped signals to handle bindings but only grouped signal will be sent to clients.

Keyboards with same keymap can be grouped.

In keyboard grab + virtual keyboard combo, both virtual and the original keyboard device will likely have the same keymap and currently will be grouped. But the events here should not be reduced. In the above example, replace k1 with original keyboard and k2 with virtual keyboard, k1 should be sent to input-method grab and k2 should be sent to text-input client.

I think the ideal solution is not to group the virtual one when it is a keyboard grab + virtual keyboard combo. I will find a way to do this.

A temporarily workaround would be turn off grouping in sway config, by `seat * keyboard_grouping none`.

As for the keymap issue, I think we should avoid sending identical keymaps to client in `wlr_seat_set_keyboard`. This can be a seperated PR.

> Grab implementation in sway is not consistent on the behavior of ignoring. Key events will check the resource while modifiers events are not.

Nice catch, I will fix it soon.

I implemented the grab in wlroots initially but decided to move it to sway to make use of pressed event sent state tracking to avoid cut off a press-release pair.

xhebox:
> I think the ideal solution is not to group the virtual one when it is a keyboard grab + virtual keyboard combo. I will find a way to do this.
I agreed. Better keep virtual_keyboard away from the keyboard_group.

But I think the missing events on sway side matters. It maybe reveals some hidden logic errors on keyboard_group.

**EDIT**: Actually with the sway-side implementation, you bypass the original keyboard_start_grab API, what do you plan with that then? I think the original API is more elegant. IMO, it is more important to keep things simple and clean than easier coding in a project of such scale.

xdavidwu:
wlr_keyboard_grab is used on implementing grab like xdg popup grab. I think input-method grab should take precedence over that kind of grab and events forwarded with virtual-keyboard should be sent to that kind of grab to make text-input on popups work properly. This situation make it 2 levels and cannot be simply implemented with just wlr_keyboard_grab.

xhebox:
I am not aware of that. Yes, IME should have higher precedence over popup grab. Thanks for informing.

And good luck to your debugging. I will stick at the no-groupping workaround for now, but any test requests are welcome :). I will post comments here if I found anything wrong with IME things.

xhebox:
OK, here my test IME is: https://github.com/xhebox/wlpinyin. Anyone who wants to test could compile it with libpinyin, which only needs sqlite and Berkely DB or so.

It is deadly simple, no GUI, but with basic function. It supports candidates selection(just show as pre-edit string), context guessing are not added yet, though. So here's my report on this draft:

1. keyboard_group hell needs to be solved
2. popup surface implementation, it maybe hard, since compositor need to find the position
3. automatically avoid recursion of keymap event
4. protocol is good and convenient to use. it's practical, nice work

All these texts are printed through wlpinyin = 这是一段wlpinyin打出来的话.

xhebox:
Keymap recursion is not solved.

Check https://github.com/swaywm/wlroots/blob/fc98e64d46fea0d27ab0c1bb2c939fcaf40a4f33/types/wlr_input_method_v2.c#L214.

Maybe move the recursion detection into `grab_send_keymap`.

xhebox:
@xdavidwu You have said that 2-level makes it hard. I've come up a new way to make nested grab possible. I applied my patch with commits before https://github.com/swaywm/wlroots/pull/1864/commits/c8ed6a89229af050442d14fc895685bd3d33fa21, and https://github.com/swaywm/sway/pull/4740.

No sway side implementation is needed. A keymap recursion patch is also included, do not mind it.

```
--- copy/include/wlr/types/wlr_seat.h	2020-01-23 02:22:04.000000000 +0800
+++ wlroots-0.10.0/include/wlr/types/wlr_seat.h	2020-02-18 20:12:34.207565389 +0800
@@ -134,6 +134,8 @@
  */
 struct wlr_seat_keyboard_grab {
 	const struct wlr_keyboard_grab_interface *interface;
+	bool ime_grab;
+	struct wlr_seat_keyboard_grab* nested_grab;
 	struct wlr_seat *seat;
 	void *data;
 };
diff '--color=auto' -ru copy/types/seat/wlr_seat.c wlroots-0.10.0/types/seat/wlr_seat.c
--- copy/types/seat/wlr_seat.c	2020-01-23 02:22:04.000000000 +0800
+++ wlroots-0.10.0/types/seat/wlr_seat.c	2020-02-18 20:13:46.674266565 +0800
@@ -227,6 +227,7 @@
 		free(seat);
 		return NULL;
 	}
+	keyboard_grab->nested_grab = NULL;
 	keyboard_grab->interface = &default_keyboard_grab_impl;
 	keyboard_grab->seat = seat;
 	seat->keyboard_state.default_grab = keyboard_grab;
--- copy/types/seat/wlr_seat_keyboard.c	2020-01-23 02:22:04.000000000 +0800
+++ wlroots-0.10.0/types/seat/wlr_seat_keyboard.c	2020-02-18 20:15:33.324277137 +0800
@@ -167,6 +167,12 @@
 
 void wlr_seat_keyboard_start_grab(struct wlr_seat *wlr_seat,
 		struct wlr_seat_keyboard_grab *grab) {
+	if (wlr_seat->keyboard_state.grab->ime_grab) { // switch position, so ime_grab is always the top one
+		struct wlr_seat_keyboard_grab *newgrab = grab;
+		grab = wlr_seat->keyboard_state.grab;
+		wlr_seat->keyboard_state.grab = grab;
+	}
+	grab->nested_grab = wlr_seat->keyboard_state.grab;
 	grab->seat = wlr_seat;
 	wlr_seat->keyboard_state.grab = grab;
 
@@ -177,7 +178,7 @@
 	struct wlr_seat_keyboard_grab *grab = wlr_seat->keyboard_state.grab;
 
 	if (grab != wlr_seat->keyboard_state.default_grab) {
-		wlr_seat->keyboard_state.grab = wlr_seat->keyboard_state.default_grab;
+		wlr_seat->keyboard_state.grab = wlr_seat->keyboard_state.grab->nested_grab;
 		wlr_signal_emit_safe(&wlr_seat->events.keyboard_grab_end, grab);
 		if (grab->interface->cancel) {
 			grab->interface->cancel(grab);
--- copy/types/wlr_input_method_v2.c	2020-02-18 20:04:44.793169946 +0800
+++ wlroots-0.10.0/types/wlr_input_method_v2.c	2020-02-18 20:21:47.407379121 +0800
@@ -159,7 +159,7 @@
 	struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab = grab->data;
 	struct wlr_keyboard *keyboard = grab->seat->keyboard_state.keyboard;
 	if (keyboard->virtual_keyboard) {
-		wlr_seat_keyboard_send_key(grab->seat, time, key, state);
+		grab->nested_grab->interface->key(grab, time, key, state);
 	} else {
 		zwp_input_method_keyboard_grab_v2_send_key(
 			im_keyboard_grab->resource, im_keyboard_grab->serial++,
@@ -172,7 +172,7 @@
 	struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab = grab->data;
 	struct wlr_keyboard *keyboard = grab->seat->keyboard_state.keyboard;
 	if (keyboard->virtual_keyboard) {
-		wlr_seat_keyboard_send_modifiers(grab->seat, modifiers);
+		grab->nested_grab->interface->modifiers(grab, modifiers);
 	} else {
 		zwp_input_method_keyboard_grab_v2_send_modifiers(
 			im_keyboard_grab->resource, im_keyboard_grab->serial++,
--- a/include/wlr/types/wlr_input_method_v2.h	2020-02-18 20:27:47.748570408 +0800
+++ b/include/wlr/types/wlr_input_method_v2.h	2020-02-18 20:27:47.748570408 +0800
@@ -59,6 +59,7 @@
 	struct wl_resource *resource;
 	struct wlr_input_method_v2 *input_method;
 
 	struct wlr_seat_keyboard_grab *grab;
 	bool grabbed;
 
--- a/types/wlr_input_method_v2.c	2020-02-18 20:27:47.751903782 +0800
+++ b/types/wlr_input_method_v2.c	2020-02-18 20:27:47.751903782 +0800
@@ -203,6 +203,11 @@
 static bool im_keyboard_grab_send_keymap(
 		struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab,
 		struct wlr_keyboard *keyboard) {
+		if (im_keyboard_grab->keyboard != NULL &&
+			strcmp(im_keyboard_grab->keyboard->keymap_string,
+			keyboard->keymap_string) == 0) {
+			return true;
+		}
 	int keymap_fd = allocate_shm_file(keyboard->keymap_size);
 	if (keymap_fd < 0) {
 		wlr_log(WLR_ERROR, "creating a keymap file for %zu bytes failed",
@@ -305,6 +305,8 @@
 		wl_resource_destroy(im_keyboard_grab_resource);
 		return;
 	}
+	keyboard_grab->ime_grab = true;
+	im_keyboard_grab->keyboard = state_keyboard;
 	im_keyboard_grab_send_repeat_info(im_keyboard_grab, state_keyboard);
 	wlr_seat_keyboard_start_grab(input_method->seat, keyboard_grab);
 	wlr_seat_keyboard_notify_modifiers(input_method->seat,
```

Tested that, it can forward keys normally, with `groupping none` trick. 

And here's my concern to this new implementation:

1. It integrates with the original API pretty well, no need to bother sway.
2. I worry that if IME exited abnormally, `seat_end_grab` will not be called. Thus grab is never released. It that possible? Or because it's a wayland resource, it will be released automatically?
3. I do not have a great pop up surfaces generator to test the behavior of grab upon a pop up surface.

Does this new implementation look good to you? I could make a PR to your fork repo if you think it's effective.

xdavidwu:
@xhebox your patch make wlr_seat_keyboard_grab able to grow into unlimited number of layers, grabs that are not from input-method are able to be nested. This breaks the original behavior where when a grab is started, the old grab is just gone and won't be able to come back. I am afraid that it may break something.

Also, comparing to my sway approach, your code does not comply to [wayland input implicit grab rule](https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Input). When an input method start a grab while a key is already pressed, the following release event of that key should be sent to the original client, not the input method grab.

> I worry that if IME exited abnormally, seat_end_grab will not be called. Thus grab is never released. It that possible? Or because it's a wayland resource, it will be released automatically?

When you call wl_resource_set_implemetation, you pass a destroyer function pointer and it will be called when the resource is destroyed. You need to make sure the grab is ended in the destroyer.

xdavidwu:
@xhebox your patch make wlr_seat_keyboard_grab able to grow into unlimited number of layers, grabs that are not from input-method are able to be nested. This breaks the original behavior where when a grab is started, the old grab is just gone and won't be able to come back. I am afraid that it may break something.

Also, comparing to my sway approach, your code does not comply to [wayland input implicit grab rule](https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Input). When an input method start a grab while a key is already pressed, the following release event of that key should be sent to the original client, not the input method grab.

> I worry that if IME exited abnormally, seat_end_grab will not be called. Thus grab is never released. It that possible? Or because it's a wayland resource, it will be released automatically?

When you call wl_resource_set_implemetation, you pass a destroyer function pointer and it will be called when the resource is destroyed. You need to make sure the grab is ended in the destroyer.

> Keymap recursion is not solved.
> 
> Check
> 
> https://github.com/swaywm/wlroots/blob/fc98e64d46fea0d27ab0c1bb2c939fcaf40a4f33/types/wlr_input_method_v2.c#L214
> .
> 
> Maybe move the recursion detection into `grab_send_keymap`.

It should be already solved. The line is the in handler of keymap signal of currently tracked wlr_keyboard. That wlr_keyboard should never be set to virtual one after I fixed the modifier event handling in sway PR. Did you pull the sway-side changes?

xhebox:
@xdavidwu Thanks for your patient reply :). It's a quick patch, and more like a proof of concept.

> your patch make wlr_seat_keyboard_grab able to grow into unlimited number of layers, grabs that are not from input-method are able to be nested.

That can be fixed easily. We can only nest IME grab, and just keep other grab behaviors still. Thanks to point out.

> Also, comparing to my sway approach, your code does not comply to wayland input implicit grab rule. 

I think I now know what you mean by `avoid cut off a press-release pair`. But can these members provide a change to do that? 

```
struct wl_keyboard { ...
        size_t num_keycodes;
        struct wlr_keyboard_modifiers modifiers;
...};

static void keyboard_key_update(struct wlr_keyboard *keyboard,
                struct wlr_event_keyboard_key *event) {
        if (event->state == WLR_KEY_PRESSED) {
                set_add(keyboard->keycodes, &keyboard->num_keycodes,
                        WLR_KEYBOARD_KEYS_CAP, event->keycode);
        }
        if (event->state == WLR_KEY_RELEASED) {
                set_remove(keyboard->keycodes, &keyboard->num_keycodes,
                        WLR_KEYBOARD_KEYS_CAP, event->keycode);
        }

        assert(keyboard->num_keycodes <= WLR_KEYBOARD_KEYS_CAP);
}
```

> That wlr_keyboard should never be set to virtual one after I fixed the modifier event handling in sway PR. Did you pull the sway-side changes?

I can not really remember that... I will give a try now.

xhebox:
I confirm that, I applied `wlroots/pull/1864.diff` and `sway/pull/4932.diff` just now:

```
/home/xhe/.bash_history:proxychains curl -o ime2.patch -L "https://patch-diff.githubusercontent.com/raw/swaywm/sway/pull/4932.diff"
/home/xhe/.bash_history:proxychains curl -o ime.patch -L "https://patch-diff.githubusercontent.com/raw/swaywm/wlroots/pull/1864.diff"
```

No, it does not work somehow.

xdavidwu:
> I confirm that, I applied `wlroots/pull/1864.diff` and `sway/pull/4932.diff` just now:
> 
> ```
> /home/xhe/.bash_history:proxychains curl -o ime2.patch -L "https://patch-diff.githubusercontent.com/raw/swaywm/sway/pull/4932.diff"
> /home/xhe/.bash_history:proxychains curl -o ime.patch -L "https://patch-diff.githubusercontent.com/raw/swaywm/wlroots/pull/1864.diff"
> ```
> 
> No, it does not work somehow.

I cannot reproduce. Could you add an assert to make sure the keyboard is not virtual keyboard in that handler and try again?

xdavidwu:
> But can these members provide a change to do that?

I think you will need to take snapshots on all wlr_keyboard's when input-method grab starts to utilize that. Also that tracks events that is received from backends, not events that are sent out. I am not sure if it fits the needs here. Imagining the case that there are keyboard groups, I think it is not easy to find out what events are really sent.

xhebox:
> Could you add an assert to make sure the keyboard is not virtual keyboard in that handler and try again?

If you mean this:

```
 static void handle_keyboard_keymap(struct wl_listener *listener, void *data) {
        struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab =
                wl_container_of(listener, im_keyboard_grab, keymap_listener);
+       assert(((struct wlr_keyboard*)data)->virtual_keyboard != NULL); 
        im_keyboard_grab_send_keymap(im_keyboard_grab, data);
 }
```

I tried, sway did not get aborted. It's not a virtual keyboard. But I still get infinite keymap events, once I commented zwp_virtual_keyboard_send_keymap, only one keymap event is there. BTW, I am using that `groupping none` configuration.

> I think you will need to take snapshots on all wlr_keyboard's when input-method grab starts to utilize that.

I think we can check send release events when a new wlr_keyboard is the first time being grabbed. 

> Also that tracks events that is received from backends,

Ouch, that is a big problem.

xdavidwu:
Is your virtual-keyboard and input-method under the same wayland connection?

xhebox:
I am passing the same `state->seat` and I have only one compositor. So I think yes? Can you give a way to verify that? My terminal is indeed not spawned by sway... But I set an explicit `SWAYSOCK`, no helps.

xdavidwu:
Try
```
assert(wl_resource_get_client(((struct wlr_keyboard *)data)->virtual_keyboard->resource) == wl_resource_get_client(im_keyboard_grab->resource));
```

xhebox:
Not aborted, werid. This is what my events like:
```
[../im.c:230] deactivate
[../im.c:83] surround: 0,0, 
[../im.c:222] activate
[../im.c:83] surround: 0,0, 
[../im.c:230] deactivate
[../im.c:83] surround: 0,0, 
[../im.c:222] activate
[../im.c:83] surround: 0,0, 
[../im.c:230] deactivate
[../im.c:83] surround: 0,0, 
[../im.c:222] activate
[../im.c:83] surround: 0,0, 
[../im.c:94] keymap 1, 18, 47298
[../im.c:167] repeat_info 25, 600
[../im.c:94] keymap 1, 19, 47298
[../im.c:94] keymap 1, 20, 47298
[../im.c:94] keymap 1, 21, 47298
[../im.c:94] keymap 1, 22, 47298
[../im.c:94] keymap 1, 23, 47298
```

xdavidwu:
If you press a key, will it stop?

xhebox:
Yes, and will start infinite loops again after some deactivation/activation.

xdavidwu:
I think the commit I just pushed should fix it.

xhebox:
Confimed

xhebox:
@xdavidwu Made a patch based on the idea of that quick patch:

https://github.com/xhebox/wlroots/commit/d5698c998a611d5f4a2510821ed289c7c7c2d8d9

Some points about mine or your implementation:

1. Mine does not check if virtual keyboard and IME are with the same client. Maybe add it later.

2. You are using the shortcut model to avoid cut off press-release pairs. That means, when a release event comes, you check if it's in the pressed keys(recorded by sway). If it is, then just forward it.

And mine is recording what is sent to the grabber, so when a release event comes, I check it's sent to the grabber. If it is not, then forward it the original client.

These two methods should be identical, except when duplicate keys are allowed. If two consecutive `pressed f` events followed by a `release f`, nor yours or mine could know to which press event the release event corresponds. Also, if IME sent the same key twice, mine will have a problem, as I am using `set_add`. That can be solved by using normal array or so...

And another difference is that, you bypass the shortcut model if it's sent to the grabber. That should not matter much, IMO.

3. I think I got why I will miss all release key events with keyboard_group. Keyboard_group is working by counting. If `native press f`, `virtual press f`, `native release f` is emitted, then it counts as `+1 +1 -1 = 1`. And release events are only sent when the count is 0.

If wlchewing did not miss key release events, then real keyboard events and virtual keyboard events must not overlap. You probably send virtual keyboard events after the real key is released.

So if a virtual keyboard from IME is in a keyboard group, then it must not count, but directly deliver the events. Otherwise, we should promise they are not grouped into the same group.

Or we could restrict IME: just do not send before the real key is released.

xdavidwu:
> So if a virtual keyboard from IME is in a keyboard group, then it must not count, but directly deliver the events. Otherwise, we should promise they are not grouped into the same group.
>
> Or we could restrict IME: just do not send before the real key is released.

The simple way is just don't group virtual keyboard with others when it is from the same client as input method. I think this should be done in compositor, not the library. We group keyboards in compositor so compositor should handle that.

We must not restrict IMEs. Forwarding press before real release does not violate any protocols. Also IMEs should forward as soon as possible. Imaging doing a Ctrl+c in terminal, user will expect sigint sent right on pressed not on released.

examples/meson.build
```diff
@@ -149,6 +149,14 @@ clients = {
 		'dep': wlroots,
 		'proto': ['wlr-virtual-pointer-unstable-v1'],
 	},
+	'input-method-keyboard-grab': {
+		'src': 'input-method-keyboard-grab.c',
+		'dep': xkbcommon,
+		'proto': [
+			'input-method-unstable-v2',
+			'text-input-unstable-v3',
```
emersion: `text-input-unstable-v3` shouldn't be required.

include/wlr/types/wlr_input_method_v2.h
```diff
@@ -41,16 +43,29 @@ struct wlr_input_method_v2 {
 	bool client_active; // state known to the client
 	uint32_t current_serial; // received in last commit call
 
+	struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab;
```
emersion: Can rename to just `keyboard_grab`, since it's already a member of `struct wlr_input_method_v2` it's unambiguous.

types/wlr_input_method_v2.c
```diff
@@ -20,10 +24,24 @@ static struct wlr_input_method_v2 *input_method_from_resource(
...
+
+static void im_keyboard_grab_destroy(
+		struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab);
+
 static void input_method_destroy(struct wlr_input_method_v2 *input_method) {
 	wlr_signal_emit_safe(&input_method->events.destroy, input_method);
 	wl_list_remove(wl_resource_get_link(input_method->resource));
 	wl_list_remove(&input_method->seat_destroy.link);
+	if (input_method->im_keyboard_grab != NULL) {
+		wl_resource_destroy(input_method->im_keyboard_grab->resource);
```
emersion: The compositor cannot destroy the client's resources, except in the `destroy` request handler. Instead, the compositor can make the resource inert.

For more info, see `CONTRIBUTING.md`.

types/wlr_input_method_v2.c
```diff
@@ -20,10 +24,24 @@ static struct wlr_input_method_v2 *input_method_from_resource(
...
 
+static struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource,
+		&zwp_input_method_keyboard_grab_v2_interface,
+		&im_keyboard_grab_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void im_keyboard_grab_destroy(
```
emersion: This doesn't seem used.

include/wlr/types/wlr_keyboard.h
```diff
@@ -47,6 +47,8 @@ struct wlr_keyboard_modifiers {
 	xkb_mod_mask_t group;
 };
 
+struct wlr_virtual_keyboard_v1;
```
emersion: This forward declaration is unnecessary.

include/wlr/types/wlr_input_method_v2.h
```diff
@@ -30,6 +30,8 @@ struct wlr_input_method_v2_state {
 	struct wlr_input_method_v2_delete_surrounding_text delete;
 };
 
+struct wlr_input_method_keyboard_grab_v2;
```
emersion: This forward declaration is unnecessary.

include/wlr/types/wlr_keyboard.h
```diff
@@ -67,6 +69,8 @@ struct wlr_keyboard {
 		int32_t delay;
 	} repeat_info;
 
+	struct wlr_virtual_keyboard_v1 *virtual_keyboard;
```
emersion: Instead of adding a new field, we should add a way to get a `wlr_virtual_keyboard_v1 directly from a `wlr_keyboard` or a `wlr_input_device`. Ideally this should be done in a standalone commit. See e.g. `get_libinput_device_from_device`.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +119,177 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
 static void im_grab_keyboard(struct wl_client *client,
 		struct wl_resource *resource, uint32_t keyboard) {
-	wlr_log(WLR_INFO, "Stub: zwp_input_method_v2::grab_keyboard");
+	struct wlr_input_method_v2 *input_method =
+		input_method_from_resource(resource);
+	if (!input_method) {
+		return;
+	}
+	if (input_method->im_keyboard_grab) {
+		// Already grabbed
```
emersion: Should we send a client error here? If it's missing, we should add it to the protocol.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +119,177 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+		free(im_keyboard_grab);
+		wl_client_post_no_memory(client);
+		return;
+	}
+	wl_resource_set_implementation(im_keyboard_grab_resource,
+		&im_keyboard_grab_impl, im_keyboard_grab,
+		im_keyboard_grab_resource_destroy);
+	im_keyboard_grab->resource = im_keyboard_grab_resource;
+	im_keyboard_grab->input_method = input_method;
+	im_keyboard_grab->seat_client = wlr_seat_client_for_wl_client(
```
emersion: The `input_method` is already tied to a seat. Why do we need this `seat_client` field?

types/wlr_input_method_v2.c
```diff
@@ -101,10 +119,177 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+	void *ptr = mmap(NULL, keyboard->keymap_size, PROT_READ | PROT_WRITE,
+		MAP_SHARED, keymap_fd, 0);
+	if (ptr == MAP_FAILED) {
+		wlr_log(WLR_ERROR, "failed to mmap() %zu bytes",
+			keyboard->keymap_size);
+		close(keymap_fd);
+		return false;
+	}
+
+	strcpy(ptr, keyboard->keymap_string);
```
emersion: I'd prefer to use `memcpy` here, since it has a size argument.

include/wlr/types/wlr_input_method_v2.h
```diff
@@ -82,4 +97,14 @@ void wlr_input_method_v2_send_done(struct wlr_input_method_v2 *input_method);
 void wlr_input_method_v2_send_unavailable(
 	struct wlr_input_method_v2 *input_method);
 
+void wlr_input_method_keyboard_grab_v2_send_key(
+	struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab,
+	uint32_t time, uint32_t key, uint32_t state);
+void wlr_input_method_keyboard_grab_v2_send_modifiers(
+	struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab,
+	struct wlr_keyboard_modifiers *modifiers);
+void wlr_input_method_keyboard_grab_v2_set_keyboard(
```
emersion: We should add a way for the compositor to refuse to grab (`wlr_input_method_keyboard_grab_v2_destroy`?).

types/wlr_input_method_v2.c
```diff
@@ -101,10 +119,177 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+		// aware that it did not change and blindly send it back with
+		// virtual keyboard, it may cause an infinite recursion.
+		if (!im_keyboard_grab_send_keymap(im_keyboard_grab, keyboard)) {
+			wlr_log(WLR_ERROR, "Failed to send keymap for input-method keyboard grab");
+			return;
+		}
+	}
+	im_keyboard_grab_send_repeat_info(im_keyboard_grab, keyboard);
+
+	im_keyboard_grab->keyboard = keyboard;
```
emersion: We need a keyboard destroy listener.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +119,177 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
 
+static void im_keyboard_grab_destroy(
+		struct wlr_input_method_keyboard_grab_v2 *im_keyboard_grab) {
+	im_keyboard_grab->input_method->im_keyboard_grab = NULL;
+	if (im_keyboard_grab->keyboard) {
+		wl_list_remove(&im_keyboard_grab->keymap_listener.link);
+		wl_list_remove(&im_keyboard_grab->repeat_info_listener.link);
+
+		// send modifier state to original client
+		wlr_seat_keyboard_notify_modifiers(im_keyboard_grab->input_method->seat,
```
emersion: Hmm. I wonder if the compositor should be responsible for this.

examples/meson.build
xdavidwu: Removed, done.

include/wlr/types/wlr_input_method_v2.h
xdavidwu: I have renamed all im_keyboard_grab* to keyboard_grab*.

types/wlr_input_method_v2.c
xdavidwu: Done.

types/wlr_input_method_v2.c
xdavidwu: It becomes used now after I resolved previous conversation about inert design and resource destory.

include/wlr/types/wlr_keyboard.h
xdavidwu: Removed, done.

include/wlr/types/wlr_input_method_v2.h
xdavidwu: Removed, done.

include/wlr/types/wlr_keyboard.h
xdavidwu: I have add a `wlr_input_device_get_virtual_keyboard` to get from `wlr_input_device`.

types/wlr_input_method_v2.c
xdavidwu: Done.

emersion:
Is this ready for another review round?

include/wlr/types/wlr_input_method_v2.h
xdavidwu: Added.

types/wlr_input_method_v2.c
xdavidwu: Done. It now listens to and handles keyboard destroy.

emersion:
> Let's say that virtual keyboards should bypass input-method, then.

That means a VNC session will bypass the input-method, too. I wonder if that's desirable.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+		}
+		keyboard_grab_send_repeat_info(keyboard_grab, keyboard);
+		keyboard_grab->keymap_listener.notify = handle_keyboard_keymap;
+		wl_signal_add(&keyboard->events.keymap,
+			&keyboard_grab->keymap_listener);
+		keyboard_grab->repeat_info_listener.notify =
+			handle_keyboard_repeat_info;
+		wl_signal_add(&keyboard->events.repeat_info,
+			&keyboard_grab->repeat_info_listener);
+		keyboard_grab->destroy_listener.notify =
```
emersion, line 269: `wl_signal_add` is not called for this listener.

include/wlr/types/wlr_input_method_v2.h
```diff
@@ -41,16 +41,34 @@ struct wlr_input_method_v2 {
...
 	} events;
 };
 
+struct wlr_input_method_keyboard_grab_v2 {
+	struct wl_resource *resource;
+	struct wlr_input_method_v2 *input_method;
+	struct wlr_seat_client *seat_client;
+	struct wlr_keyboard *keyboard;
+
+	struct wl_listener keymap_listener;
```
emersion, line 63: So far the naming convention for listeners has been `<object>_<event>`. For consistency we should rename those to `keyboard_keymap`, `keyboard_repeat_info` and `keyboard_destroy`.

types/wlr_input_method_v2.c
```diff
@@ -20,10 +24,24 @@ static struct wlr_input_method_v2 *input_method_from_resource(
...
+
+static void keyboard_grab_destroy(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab);
+
 static void input_method_destroy(struct wlr_input_method_v2 *input_method) {
 	wlr_signal_emit_safe(&input_method->events.destroy, input_method);
 	wl_list_remove(wl_resource_get_link(input_method->resource));
 	wl_list_remove(&input_method->seat_destroy.link);
+	if (input_method->keyboard_grab != NULL) {
+		keyboard_grab_destroy(input_method->keyboard_grab);
```
emersion, line 43: wlroots destructors are no-op when called with NULL. We should move this NULL check in `keyboard_grab_destroy`.

types/wlr_input_method_v2.c
```diff
@@ -37,6 +55,9 @@ static void input_method_resource_destroy(struct wl_resource *resource) {
 	if (!input_method) {
 		return;
 	}
+	if (input_method->keyboard_grab != NULL) {
+		wl_resource_destroy(input_method->keyboard_grab->resource);
```
emersion, line 59: We can't destroy resources when the client hasn't requested it. We need to make the resource inert instead. See CONTRIBUTING.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
 	wlr_log(WLR_INFO, "Stub: zwp_input_method_v2::get_input_popup_surface");
 }
 
+static void keyboard_grab_destroy(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab) {
+	wlr_signal_emit_safe(&keyboard_grab->events.destroy, keyboard_grab);
+	keyboard_grab->input_method->keyboard_grab = NULL;
+	if (keyboard_grab->keyboard) {
+		wl_list_remove(&keyboard_grab->keymap_listener.link);
+		wl_list_remove(&keyboard_grab->repeat_info_listener.link);
```
emersion, line 131: Missing the keyboard destroy listener

types/wlr_input_method_v2.c
emersion: Making the resource inert is already done in `keyboard_grab_destroy` (called in `input_method_destroy`). So we should be able to just remove this `wl_resource_destroy` call.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+
+		// send modifier state to original client
+		wlr_seat_keyboard_notify_modifiers(keyboard_grab->input_method->seat,
+			&keyboard_grab->keyboard->modifiers);
```
emersion, line 135: Hmm, this does things behind the compositor's back. I'd rather not.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+	}
+	wl_resource_set_user_data(keyboard_grab->resource, NULL);
+	free(keyboard_grab);
+}
+
+static void keyboard_grab_resource_destroy(struct wl_resource *resource) {
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
+		keyboard_grab_from_resource(resource);
+	if (!keyboard_grab) {
```
emersion, line 144: Ditto, this NULL check should be in `keyboard_grab_destroy`.

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+		wl_list_remove(&keyboard_grab->destroy_listener.link);
+	}
+
+	if (keyboard) {
+		if (keyboard_grab->keyboard == NULL ||
+				strcmp(keyboard_grab->keyboard->keymap_string,
+				keyboard->keymap_string) != 0) {
+			// send keymap only if it is changed, or if input method is not
+			// aware that it did not change and blindly send it back with
+			// virtual keyboard, it may cause an infinite recursion.
```
emersion, line 255: What do you mean?

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+		keyboard_grab->destroy_listener.notify =
+			handle_keyboard_destroy;
+	}
+
+	keyboard_grab->keyboard = keyboard;
+};
+
+void wlr_input_method_keyboard_grab_v2_destroy(
+		struct wlr_input_method_keyboard_grab_v2 *keyboard_grab) {
+	keyboard_grab_destroy(keyboard_grab);
```
emersion, line 278: We could just merge these two functions if they do the same thing

types/wlr_input_method_v2.c
```diff
@@ -101,10 +122,196 @@ static void im_get_input_popup_surface(struct wl_client *client,
...
+		free(keyboard_grab);
+		wl_client_post_no_memory(client);
+		return;
+	}
+	wl_resource_set_implementation(keyboard_grab_resource,
+		&keyboard_grab_impl, keyboard_grab,
+		keyboard_grab_resource_destroy);
+	keyboard_grab->resource = keyboard_grab_resource;
+	keyboard_grab->input_method = input_method;
+	keyboard_grab->seat_client = wlr_seat_client_for_wl_client(
```
emersion, line 310: What if the seat client is destroyed? We should setup a destroy listener.

types/wlr_input_method_v2.c
xhebox: I can send the keymap state received from input_method back to the compositor by virtual_keyboard, then input_method will notify me the "new" keymap state - the one i just sent by virtual_keyboard.

types/wlr_input_method_v2.c
xdavidwu: Fixed.

include/wlr/types/wlr_input_method_v2.h
xdavidwu: Fixed.

types/wlr_input_method_v2.c
xdavidwu: Fixed.

types/wlr_input_method_v2.c
xdavidwu: Fixed.

types/wlr_input_method_v2.c
xdavidwu: Fixed.

types/wlr_input_method_v2.c
xdavidwu: Fixed.

types/wlr_input_method_v2.c
xdavidwu: I've removed it but I found another problem. The protocol specified that `destroy` request on zwp_input_method_v2 should also destroy associated child like zwp_input_method_keyboard_grab_v2. If input method is already inert, how can this be done? How can I find resource of grab to call wl_resource_destroy? Should I add a layer to input method user data to track child object resource?

types/wlr_input_method_v2.c
xdavidwu: I've moved this to compositor side.

types/wlr_input_method_v2.c
xdavidwu: I've moved this to compositor side.

types/wlr_input_method_v2.c
xdavidwu: This is already handled by input method and will make it (and the grab, if exists) inert.

xdavidwu:
> > Let's say that virtual keyboards should bypass input-method, then.
> 
> That means a VNC session will bypass the input-method, too. I wonder if that's desirable.

I've make only virtual keyboard from the same client as grab bypass input-method grab on sway side.

include/wlr/types/wlr_input_method_v2.h
```diff
@@ -41,16 +41,34 @@ struct wlr_input_method_v2 {
...
 
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab;
+
 	struct wl_list link;
 
 	struct wl_listener seat_destroy;
 
 	struct {
 		struct wl_signal commit; // (struct wlr_input_method_v2*)
+		struct wl_signal grab_keyboard; // (struct wlr_input_method_v2*)
```
emersion, line 52: The data pointer is a `struct wlr_input_method_keyboard_grab_v2 *`

emersion:
>I've make only virtual keyboard from the same client as grab bypass input-method grab on sway side.

This isn't going to work in all cases. For instance a user using Waypipe would have all Wayland objects belonging to the same client. On the long run I think we'll want to add a way for clients to tie a virtual-keyboard to an input-method.

types/wlr_input_method_v2.c
emersion: Hmm, that is troublesome. I think right before the `wl_resource_destroy` call for the zwp_input_method_v2 resource we'll want to also call `wl_resource_destroy` on the zwp_input_method_keyboard_grab_v2 resource, if any.

types/wlr_input_method_v2.c
emersion: Looks good

types/wlr_input_method_v2.c
emersion: Ah, I suppose this is to get seat serials.

types/wlr_input_method_v2.c
emersion: The seat client may be destroyed before the seat, so I still think a destroy listener is necessary to avoid invalid pointers.

emersion:
With these remaining comments fixed, this LGTM!

include/wlr/types/wlr_input_method_v2.h
xdavidwu: Fixed.

types/wlr_input_method_v2.c
xdavidwu: The `seat_destroy` listener in `wlr_input_method_v2` actually listens to seat client destroy instead of the seat.

types/wlr_input_method_v2.c
xdavidwu: If input method is inert, struct `wlr_input_method_v2` is already freed. We need another way to find wl_resource of the grab.

emersion: If the input method is inert, it means we've already free'ed the `wlr_input_method_v2` (and the grab with it). So it should be fine.

types/wlr_input_method_v2.c
emersion: Ah, okay. Can we change `wlr_input_method_v2` to have a `struct wlr_seat_client *seat_client` field instead of a `seat` field, and rename `seat_destroy` to `seat_client_destroy`?

emersion: Gentle ping :)

xdavidwu: Done. Changes for sway side for seat_client instead of seat has also been added to sway side PR.

emersion:
>This isn't going to work in all cases. For instance a user using Waypipe would have all Wayland objects belonging to the same client. On the long run I think we'll want to add a way for clients to tie a virtual-keyboard to an input-method.

Created a follow-up issue: https://github.com/swaywm/wlroots/issues/2322

emersion:
I've squashed all of your commits into three logical commits. Also made a minor change to avoid breaking the API: added back `wlr_input_method_v2.seat`.

Issue #1852, opened 2019-10-13
Sway starts with black screen, only cursor visible


johnae:
I know there's a similar issue posted already, however, in my case this started happening when compiling sway against wlroots after commit https://github.com/swaywm/wlroots/commit/3b4824a2fe933a829b9710da069f8e067cf7e1a5 (which is only three commits). So this is very new.

I don't know which one is causing this yet. The sway commit compiled against wlroots is https://github.com/swaywm/sway/commit/533acbe679b2eb1abfcac31aa6bfadd5cffe265d (so latest commit as of right now).

emersion:
Needs debug logs.

johnae:
This is the commit in wlroots that causes it anyway: https://github.com/swaywm/wlroots/commit/2bdd1d0896cce69f2ecfbc49f5d28abcc6e302cb

johnae:
Here's a debug log (I killed sway from other VT after a while fyi): 
[sway.log](https://github.com/swaywm/wlroots/files/3721578/sway.log)

emersion:
```
2019-10-13 11:06:19 - [backend/drm/atomic.c:57] eDP-1: Atomic commit failed (modeset): Invalid argument
```

Then followed by:

```
2019-10-13 11:06:24 - [backend/drm/drm.c:952] eDP-1: Retrying pageflip
2019-10-13 11:06:24 - [backend/drm/drm.c:503] Starting renderer on output 'eDP-1'
2019-10-13 11:06:24 - [backend/drm/atomic.c:57] eDP-1: Atomic commit failed (modeset): Invalid argument
```

Can you do this:

```
sudo dmesg -C
echo 0xFE | sudo tee /sys/module/drm/parameters/debug
sway
# exit sway
echo 0x00 | sudo tee /sys/module/drm/parameters/debug
dmesg >dmesg.log
```

And upload `dmesg.log`?

johnae:
If it matters, this is a Dell XPS 13 9370. Haven't had any issues like this with sway/wlroots otherwise. Will get back to you with the requested log in a while.

emersion:
If you could also try to use Weston and see if it does the same thing, that would be helpful.

johnae:
[dmesg.log](https://github.com/swaywm/wlroots/files/3721628/dmesg.log)

emersion:
```
[13286.136997] [drm:drm_atomic_check_only [drm]] checking 000000006fec0bb4
[...]

PR #1870, opened 2019-10-22
backend/drm: retry without modifiers for the primary plane

emersion:
On some Intel cards using modifiers can fill the FIFO and prevent
hotplugged outputs from being properly enabled.

Add a fallback without modifiers.

References: https://github.com/swaywm/wlroots/issues/1840
Closes: https://github.com/swaywm/wlroots/issues/1852


emersion:
Confirmed to [work](https://github.com/swaywm/wlroots/issues/1852#issuecomment-545270438).

ascent12:
Thanks

PR #1872, opened 2019-10-22
backend/drm: track gbm_bo during direct scan-out

emersion:
We need to destroy the gbm_bo we imported and drmModeRmFb.

Closes: https://github.com/swaywm/sway/issues/4662


tokyovigilante:
Fixed for me with mpv@e82bf5f:
```
$ sudo cat /sys/kernel/debug/dri/0/framebuffer | grep sway | wc -l
5
```

PR #1875, opened 2019-10-25
Implement wlr-output-power-management

agx:
~~only got some light testing therefore wip~~

See https://github.com/swaywm/wlr-protocols/pull/61

### DIFF
--- a/examples/output-power-management.c
+++ b/examples/output-power-management.c
@@ -1,0 +1,134 @@
+#define _POSIX_C_SOURCE 200809L
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <wayland-client-protocol.h>
+#include <wayland-client.h>
+#include "wlr-output-power-management-unstable-v1-client-protocol.h"
+
+struct output {
+	struct wl_output *wl_output;
+	struct zwlr_output_power_v1 *output_power;
+	struct wl_list link;
+};
+
+static struct wl_list outputs;
+static struct zwlr_output_power_manager_v1 *output_power_manager = NULL;
+
+static void output_power_handle_mode(void *data,
+		struct zwlr_output_power_v1 *output_power,
+		enum zwlr_output_power_v1_mode mode) {
+	struct output *output = data;
+
+	switch (mode) {
+	case ZWLR_OUTPUT_POWER_V1_MODE_OFF:
+		printf("Output %p disabled\n", output);
+		break;
+	case ZWLR_OUTPUT_POWER_V1_MODE_ON:
+		printf("Output %p enabled\n", output);
+		break;
+	}
+}
+
+static void output_power_handle_failed(void *data,
+		struct zwlr_output_power_v1 *output_power) {
+	fprintf(stderr, "failed to set output power mode\n");
+	exit(EXIT_FAILURE);
+}
+
+static const struct zwlr_output_power_v1_listener output_power_listener = {
+	.mode = output_power_handle_mode,
+	.failed = output_power_handle_failed,
+};
+
+static void registry_handle_global(void *data, struct wl_registry *registry,
+		uint32_t name, const char *interface, uint32_t version) {
+	if (strcmp(interface, wl_output_interface.name) == 0) {
+		struct output *output = calloc(1, sizeof(struct output));
+		output->wl_output = wl_registry_bind(registry, name,
+			&wl_output_interface, 1);
+		wl_list_insert(&outputs, &output->link);
+	} else if (strcmp(interface,
+			zwlr_output_power_manager_v1_interface.name) == 0) {
+		output_power_manager = wl_registry_bind(registry, name,
+			&zwlr_output_power_manager_v1_interface, 1);
+	}
+}
+
+static void registry_handle_global_remove(void *data,
+		struct wl_registry *registry, uint32_t name) {
+	// Who cares?
+}
+
+static const struct wl_registry_listener registry_listener = {
+	.global = registry_handle_global,
+	.global_remove = registry_handle_global_remove,
+};
+
+static const char usage[] = "usage: output-power-management [options...]\n"
+	"  -h: show this help message\n"
+	"  -e: turn outputs on\n"
+	"  -d: turn outputs off\n";
+
+int main(int argc, char *argv[]) {
+	wl_list_init(&outputs);
+
+	int opt;
+	enum zwlr_output_power_v1_mode mode =
+		ZWLR_OUTPUT_POWER_V1_MODE_ON;
+	while ((opt = getopt(argc, argv, "edh")) != -1) {
+		switch (opt) {
+		case 'e':
+			mode = ZWLR_OUTPUT_POWER_V1_MODE_ON;
+			break;
+		case 'd':
+			mode = ZWLR_OUTPUT_POWER_V1_MODE_OFF;
+			break;
+		case 'h':
+		default:
+			fprintf(stderr, usage);
+			return opt == 'h' ? EXIT_SUCCESS : EXIT_FAILURE;
+		}
+	}
+
+	struct wl_display *display = wl_display_connect(NULL);
+	if (display == NULL) {
+		fprintf(stderr, "failed to create display\n");
+		return -1;
+	}
+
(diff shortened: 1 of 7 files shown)


agx:
Dropping the wip here, tested against this branch: https://source.puri.sm/guido.gunther/phoc/tree/wlr-output-power-management plust the workaround from https://github.com/swaywm/wlroots/issues/1889

types/wlr_output_power_management_v1.c
```diff
@@ -0,0 +1,244 @@
...
+		return;
+	}
+
+	switch (mode) {
+	case ZWLR_OUTPUT_POWER_V1_MODE_OFF:
+	case ZWLR_OUTPUT_POWER_V1_MODE_ON:
+		break;
+	default:
+		wlr_log(WLR_ERROR, "Invalid power mode %d", mode);
+		zwlr_output_power_v1_send_failed(output_power->resource);
```
emersion: We should add a `return` right after to prevent `wlr_output_enable` from being called.

But actually sending an invalid enum value should be a protocol error (not because the output doesn't support it, but because it's an invalid value).

types/wlr_output_power_management_v1.c
```diff
@@ -0,0 +1,244 @@
...
+		output_power_manager_from_resource(manager_resource);
+	struct wlr_output *output = wlr_output_from_resource(output_resource);
+
+	struct wlr_output_power_v1 *output_power =
+		calloc(1, sizeof(struct wlr_output_power_v1));
+	if (output_power == NULL) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+	output_power->output = output;
```
emersion: `output` can be NULL at this point, in case it's inert.

types/wlr_output_power_management_v1.c
```diff
@@ -0,0 +1,244 @@
...
+}
+
+static void output_power_handle_output_destroy(struct wl_listener *listener,
+		void *data) {
+	struct wlr_output_power_v1 *output_power =
+		wl_container_of(listener, output_power, output_destroy_listener);
+	output_power_destroy(output_power);
+}
+
+static void wlr_output_power_v1_send_mode(struct wlr_output_power_v1 *output_power) {
```
emersion: Our internal functions shouldn't have the `wlr_` prefix, for consistency with the version script we have in place.

types/wlr_output_power_management_v1.c
agx: i've introduced an error for that and updated the protocol

types/wlr_output_power_management_v1.c
agx: modified to send
```
			zwlr_output_power_v1_send_failed(mgmt->resource);
```

too in this case. should we mention this in the protocol?

types/wlr_output_power_management_v1.c
agx: good point, fixed.

agx:
@emersion thanks for having a look, should hopefully be good now.

agx:
Anything i can do to move this forward?

types/wlr_output_power_management_v1.c
```diff
@@ -0,0 +1,247 @@
...
+		break;
+	default:
+		wlr_log(WLR_ERROR, "Invalid power mode %d", mode);
+		wl_resource_post_error(output_power_resource,
+			ZWLR_OUTPUT_POWER_V1_ERROR_INVALID_MODE,
+			"Invalid power mode");
+		return;
+	}
+
+	bool ok = wlr_output_enable(output_power->output, mode);
```
emersion: I'd prefer to have a signal for this instead of enabling/disabling the output behind the compositor's back. For instance, in Sway we have an internal field for the DPMS status, we'd need to update it there. This would also allow a compositor to decide whether or not to apply the client's request (for instance, if an output has a idle-inhibit client on it, the compositor might want to keep it on).

agx: @emersion so this 

a) shouldn't enable/disable the connector (and just send a signal so the compositor can do it) or 
b) should it do the enabling/disabling *and* send a signal so the compositor gets informed about it

i assume a) ?

emersion: > shouldn't enable/disable the connector (and just send a signal so the compositor can do it)

This :P

emersion: Clarification: we need to provide a way for the compositor not to obey to the client, so (b) won't do.

examples/output-power-management.c
```diff
@@ -0,0 +1,134 @@
...
+	// Who cares?
+}
+
+static const struct wl_registry_listener registry_listener = {
+	.global = registry_handle_global,
+	.global_remove = registry_handle_global_remove,
+};
+
+static const char usage[] = "usage: wlr-output-management [-e|-d|-h]\n"
+	"  -h          show this help message\n"
```
ddevault: This indent is kind of weird

emersion:
Added a `set_mode` event, rebased against master.

include/wlr/types/meson.build
```diff
@@ -0,0 +1,53 @@
...
+	'wlr_text_input_v3.h',
+	'wlr_touch.h',
+	'wlr_virtual_keyboard_v1.h',
+	'wlr_xcursor_manager.h',
+	'wlr_xdg_decoration_v1.h',
+	'wlr_xdg_output_v1.h',
+	'wlr_xdg_shell_v6.h',
+	'wlr_xdg_shell.h',
+	subdir: 'wlr/types',
+)
```
ddevault: ??

emersion: Oops, rebase issue. Fixed.

agx:
@emersion looks good to me, thanks a lot for fixing this up!

agx:
Can i do anything to get this applied? It would feel odd to merge this myself now that @emersion fixed my omissions.

protocol/wlr-output-power-management-unstable-v1.xml
```diff
@@ -0,0 +1,128 @@
...
+    <request name="set_mode">
+      <description summary="Set an outputs power save mode">
+        Set an output's power save mode to the given mode. The mode change
+        is effective immediately. If the output does not support the given
+        mode a failed event is sent.
+      </description>
+      <arg name="mode" type="uint" enum="mode" summary="the power save mode to set"/>
+    </request>
+
+    <event name="mode">
```
ddevault: It occurs to me that this is late in the review process, and I apologise - but this caught my eye and it doesn't really sit well for me. The language here overlaps with *mode*setting. Can we call this "state", perhaps?

agx: that would give

```
 <request name="set_state">
      <description summary="Set an outputs power save state">
        Set an output's power save state to the given state. The state change
        is effective immediately. If the output does not support the given
        state a failed event is sent.
      </description>
      <arg name="state" type="uint" enum="state" summary="the power save state to set"/>
    </request>
```

which also reads odd to me. Also dpms talks about `modes`. Maybe someone has an idea besides `mode` or `state`? Otherwise

ddevault: Do you have any documentation from display or GPU vendors which describes this feature? Maybe we can ape their vocabulary, or at least lend some legitimacy to the "mode" conflict.

ddevault: Perhaps "status" would work?

agx: I looked at https://en.wikipedia.org/wiki/VESA_Display_Power_Management_Signaling : `DPMS defines four modes; normal, standby, suspended and off` and https://tldp.org/HOWTO/Battery-Powered/displaytypes.html but the wikipedia one is loke base on the tldp one. X besides modes also talks about 'power levels' (which is longer but seems to fit). My order of preference (best to worst is)

1. mode (used elsewhere)
2. power level (used elsewhere but longer)
3. state 
4. status (odd plural form)

I'm not really hooked to any of these though so please just pick one and i'll brush up the protocol.

emersion: I'm fine with sticking with "mode" because the symbols are prefixed with "power".

ddevault: Alright, mode it is.

ddevault:
Thanks!

PR #1878, opened 2019-10-26
output-management-v1: add assertion as a safety net

emersion:
Makes it easier to figure out when the compositor submits an invalid
output state.

References: https://github.com/swaywm/sway/pull/4673


ddevault:
Thanks!

PR #1879, opened 2019-10-26
virtual-pointer: Add support for wlr-virtual-pointer-unstable-v1

atx:
This is mostly copy of the virtual-keyboard code.

I am sort of unsure if this is the correct way to handle the axis_start/axis_stop calls. At the moment it looks like they get automatically generated inside wlroots (should they even be in the virtual-pointer protocol?).

### DIFF
--- a/types/wlr_virtual_pointer_v1.c
+++ b/types/wlr_virtual_pointer_v1.c
@@ -1,0 +1,353 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_virtual_pointer_v1.h>
+#include <wlr/types/wlr_pointer.h>
+#include <wlr/util/log.h>
+#include "util/signal.h"
+#include "wlr-virtual-pointer-unstable-v1-protocol.h"
+
+static void input_device_destroy(struct wlr_input_device *dev) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		(struct wlr_virtual_pointer_v1 *)dev;
+	wl_resource_set_user_data(pointer->resource, NULL);
+	wlr_signal_emit_safe(&pointer->events.destroy, pointer);
+	wl_list_remove(&pointer->link);
+	free(pointer);
+}
+
+static const struct wlr_input_device_impl input_device_impl = {
+	.destroy = input_device_destroy
+};
+
+static const struct zwlr_virtual_pointer_v1_interface virtual_pointer_impl;
+
+static struct wlr_virtual_pointer_v1 *virtual_pointer_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource,
+		&zwlr_virtual_pointer_v1_interface, &virtual_pointer_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void virtual_pointer_motion(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time,
+		wl_fixed_t dx, wl_fixed_t dy) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	struct wlr_event_pointer_motion event = {
+		.device = wlr_dev,
+		.time_msec = time,
+		.delta_x = wl_fixed_to_double(dx),
+		.delta_y = wl_fixed_to_double(dy),
+		.unaccel_dx = wl_fixed_to_double(dx),
+		.unaccel_dy = wl_fixed_to_double(dy),
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.motion, &event);
+}
+
+static void virtual_pointer_motion_absolute(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time, uint32_t x, uint32_t y,
+		uint32_t x_extent, uint32_t y_extent) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	if (x_extent == 0 || y_extent == 0) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	struct wlr_event_pointer_motion_absolute event = {
+		.device = wlr_dev,
+		.time_msec = time,
+		.x = (double)x / x_extent,
+		.y = (double)y / y_extent,
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.motion_absolute, &event);
+}
+
+static void virtual_pointer_button(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time, uint32_t button,
+		uint32_t state) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	struct wlr_event_pointer_button event = {
+		.device = wlr_dev,
+		.time_msec = time,
+		.button = button,
+		.state = state ? WLR_BUTTON_PRESSED : WLR_BUTTON_RELEASED
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.button, &event);
+}
+
+static void virtual_pointer_axis(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time, uint32_t axis,
+		wl_fixed_t value) {
+	if (axis > WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
+		wl_resource_post_error(resource,
+				ZWLR_VIRTUAL_POINTER_V1_ERROR_INVALID_AXIS,
+				"Invalid enumeration value %d", axis);
+		return;
+	}
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	pointer->axis_valid[axis] = true;
+	pointer->axis_event[pointer->axis].device = wlr_dev;
+	pointer->axis_event[pointer->axis].time_msec = time;
+	pointer->axis_event[pointer->axis].orientation = axis;
+	pointer->axis_event[pointer->axis].delta = wl_fixed_to_double(value);
+}
+
+static void virtual_pointer_frame(struct wl_client *client,
+		struct wl_resource *resource) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+
+	for (size_t i = 0;
+			i < sizeof(pointer->axis_valid) / sizeof(pointer->axis_valid[0]);
+			++i) {
+		if (pointer->axis_valid[i]) {
+			/* Deliver pending axis event */
+			wlr_signal_emit_safe(&wlr_dev->pointer->events.axis,
+					&pointer->axis_event[i]);
+			memset(&pointer->axis_event[i], 0, sizeof(pointer->axis_event[i]));
+			pointer->axis_valid[i] = false;
+		}
+	}
+
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.frame, wlr_dev->pointer);
+}
+
+static void virtual_pointer_axis_source(struct wl_client *client,
+		struct wl_resource *resource, uint32_t source) {
+	if (source > WL_POINTER_AXIS_SOURCE_WHEEL_TILT) {
+		wl_resource_post_error(resource,
+				ZWLR_VIRTUAL_POINTER_V1_ERROR_INVALID_AXIS_SOURCE,
+				"Invalid enumeration value %d", source);
+		return;
+	}
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	pointer->axis_event[pointer->axis].device = wlr_dev;
+	pointer->axis_event[pointer->axis].source = source;
+}
+
+static void virtual_pointer_axis_stop(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time, uint32_t axis) {
+	if (axis > WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
+		wl_resource_post_error(resource,
+				ZWLR_VIRTUAL_POINTER_V1_ERROR_INVALID_AXIS,
+				"Invalid enumeration value %d", axis);
+		return;
+	}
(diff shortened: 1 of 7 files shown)


types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,284 @@
+
```
ddevault: newline?

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,284 @@
...
+#include <stdlib.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_virtual_pointer_v1.h>
+#include <wlr/types/wlr_pointer.h>
+#include <wlr/util/log.h>
+#include "util/signal.h"
+#include "wlr-virtual-pointer-unstable-v1-protocol.h"
+
+static void input_device_destroy(struct wlr_input_device *dev) {
+}
```
ddevault: Needs `/* This space deliberately left blank */` if deliberate, or an implementation if not

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,284 @@
...
+
+static void handle_display_destroy(struct wl_listener *listener, void *data) {
+	struct wlr_virtual_pointer_manager_v1 *manager =
+		wl_container_of(listener, manager, display_destroy);
+	wlr_virtual_pointer_manager_v1_destroy(manager);
+}
+
+struct wlr_virtual_pointer_manager_v1* wlr_virtual_pointer_manager_v1_create(
+		struct wl_display *display)
+{
```
ddevault: braces go on the same line

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,284 @@
...
+	wl_signal_init(&manager->events.destroy);
+	manager->global = wl_global_create(display,
+		&zwlr_virtual_pointer_manager_v1_interface, 1, manager,
+		virtual_pointer_manager_bind);
+	return manager;
+}
+
+void wlr_virtual_pointer_manager_v1_destroy(
+	struct wlr_virtual_pointer_manager_v1 *manager)
+{
```
ddevault: ditto

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,284 @@
...
+	wlr_signal_emit_safe(&manager->events.destroy, manager);
+	wl_list_remove(&manager->display_destroy.link);
+	wl_global_destroy(manager->global);
+	struct wl_resource *resource, *resource_tmp;
+	wl_resource_for_each_safe(resource, resource_tmp, &manager->resources) {
+		wl_resource_destroy(resource);
+	}
+	struct wlr_virtual_pointer_v1 *pointer, *pointer_tmp;
+	wl_list_for_each_safe(pointer, pointer_tmp, &manager->virtual_pointers,
+			link) {
```
ddevault: this would be better balanced if you put the newline before the &

ddevault:
Looking good, nice work.

atx:
Updated

emersion:
>I am sort of unsure if this is the correct way to handle the axis_start/axis_stop calls. At the moment it looks like they get automatically generated inside wlroots (should they even be in the virtual-pointer protocol?).

`axis_stop` is generated when the backend sends an `axis` event with a zero value. `axis_source` is part of the `axis` event as well, just like `value_discrete`. I don't see any `axis_start` though.

atx:
>  I don't see any axis_start though.

Eh, I meant `axis_source`

> axis_stop is generated when the backend sends an axis event with a zero value.

Right, I am going to wire up axis_stop to sending a zero value axis event then.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,291 @@
...
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.axis, &event);
+}
+
+static void virtual_pointer_destroy_resource(struct wl_resource *resource) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	wlr_signal_emit_safe(&pointer->events.destroy, pointer);
+	wl_list_remove(&pointer->link);
+	wlr_input_device_destroy(&pointer->input_device);
+	free(pointer);
```
emersion: The compositor can destroy the device too. We should move this logic into `input_device_destroy` and make the resource inert from there.

More info about inert resources in `CONTRIBUTING`. Feel free to ask questions about all of this on IRC, too.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,291 @@
...
+
+	manager->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &manager->display_destroy);
+
+	wl_list_init(&manager->resources);
+	wl_list_init(&manager->virtual_pointers);
+
+	wl_signal_init(&manager->events.new_virtual_pointer);
+	wl_signal_init(&manager->events.destroy);
+	manager->global = wl_global_create(display,
```
emersion: This is missing a NULL check.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	wl_resource_set_implementation(pointer_resource, &virtual_pointer_impl,
+		virtual_pointer, virtual_pointer_destroy_resource);
+
+	wlr_input_device_init(&virtual_pointer->input_device,
+		WLR_INPUT_DEVICE_POINTER, &input_device_impl, "virtual pointer",
+		0x0, 0x0);
+
+	struct wlr_seat_client *seat_client = wlr_seat_client_from_resource(seat);
```
emersion: This function returns NULL if the seat is inert (destroyed by the compositor, but the client isn't aware of this yet). We'll probably want to create an inert virtual pointer in this case.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,291 @@
...
+		.x = (double)x / x_extent,
+		.y = (double)y / y_extent,
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.motion_absolute, &event);
+}
+
+static void virtual_pointer_button(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time, uint32_t button,
+		uint32_t state) {
+	if (button == 0) {
```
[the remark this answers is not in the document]
ddevault: Removing this but doesn't seem to need a protocol error. The client should be able to deal with unknown buttons gracefully, maybe the user has a funky mouse.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,291 @@
...
+		wl_fixed_t value) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	struct wlr_event_pointer_axis event = {
+		.device = wlr_dev,
+		.time_msec = time,
+		.source = source,
+		.delta = wl_fixed_to_double(value),
+		.delta_discrete = wl_fixed_to_int(value),
```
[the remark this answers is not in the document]
ddevault: After a reading of the wlr_seat code, I think this should be set to zero here and the rest of the code is correct.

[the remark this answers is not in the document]
emersion: Won't work for the other case, where `delta` would be zero and `delta_discrete` non-zero.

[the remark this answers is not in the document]
ddevault: Ah, good point. This wlr_seat code is kind of poorly designed.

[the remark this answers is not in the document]
ddevault: We also have to send the pending event in the `axis_stop` handler. I have a patch which does tihs.

ddevault:
Pushed a fixup commit, will squash once we like it. @emersion can you re-review this?

ddevault:
@atx is there a client implementation of this available anywhere?

atx:
@ddevault At the moment there is only the rudimentary client in `examples/virtual-pointer.c`

ddevault:
Do you have plans to implement it for wvnc?

atx:
> Do you have plans to implement it for wvnc?

Yes (once I get enough free time...).

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,327 @@
...
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	pointer->axis_event.device = wlr_dev;
+	pointer->axis_event.time_msec = time;
+	pointer->axis_event.source = source;
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.axis, &pointer->axis_event);
```
emersion: I wonder if the protocol should be stricter. It currently allows for things which don't make sense like:

```
axis_stop(time=42, axis=vertical)
axis(time=42, axis=vertical, value=69)
frame()
```

Maybe we should only allow for one axis/axis_stop event per axis (vertical/horizontal) in each frame.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,327 @@
...
+		.device = wlr_dev,
+		.time_msec = time,
+		.button = button,
+		.state = state ? WLR_BUTTON_PRESSED : WLR_BUTTON_RELEASED
+	};
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.button, &event);
+}
+
+static void virtual_pointer_axis(struct wl_client *client,
+		struct wl_resource *resource, uint32_t time, uint32_t source,
```
emersion: Ideally, we need to make sure `source` is a valid enum value (note, it's named `axis` in the protocol). Maybe add a protocol error for that.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,327 @@
...
+		wl_fixed_t value) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	pointer->axis_event.device = wlr_dev;
+	pointer->axis_event.time_msec = time;
+	pointer->axis_event.source = source;
```
emersion: This logic doesn't handle this sequence of events properly:

```
axis(time=42, axis=vertical, value=69)
axis(time=42, axis=horizontal, value=69)
frame()
```

It should send two axis events, yet it overwrites the vertical event with the horizontal one. I'd suggest making `pointer->axis_event` an array with as many elements as there are axes (currently 2).

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,327 @@
...
+
+struct wlr_virtual_pointer_manager_v1* wlr_virtual_pointer_manager_v1_create(
+		struct wl_display *display) {
+	struct wlr_virtual_pointer_manager_v1 *manager = calloc(1,
+		sizeof(struct wlr_virtual_pointer_manager_v1));
+	if (!manager) {
+		return NULL;
+	}
+
+	manager->display_destroy.notify = handle_display_destroy;
```
emersion: We should move this block at the end of the function: in case the global can't be allocated, we don't want to have a free'd display destroy listener (will cause a use-after-free when a new display destroy listener is added or when the signal is emitted).

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,327 @@
...
+	virtual_pointer->input_device.pointer = pointer;
+	virtual_pointer->resource = pointer_resource;
+	wl_signal_init(&virtual_pointer->events.destroy);
+
+	wl_list_insert(&manager->virtual_pointers, &virtual_pointer->link);
+	wlr_signal_emit_safe(&manager->events.new_virtual_pointer, &event);
+}
+
+static const struct zwlr_virtual_pointer_manager_v1_interface manager_impl = {
+	.create_virtual_pointer = virtual_pointer_manager_create_virtual_pointer
```
emersion: We're missing a destroy request handler here.

types/wlr_virtual_pointer_v1.c
emersion: Actually, there's some mixup between `source` and `axis` here: the `axis_source` event handler populates the same field. I believe this function shouldn't touch `source` at all.

types/wlr_virtual_pointer_v1.c
ddevault: Doesn't libwayland do that for us once we fill out the enum field in the XML?

ddevault: No, it doesn't. That's lame.

emersion: https://gitlab.freedesktop.org/wayland/wayland/issues/104

types/wlr_virtual_pointer_v1.c
ddevault: There is no destroy request.

emersion: Hmm, we should definitely add one.

ddevault:
I think this is in good shape.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,346 @@
...
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	struct wlr_event_pointer_motion_absolute event = {
+		.device = wlr_dev,
+		.time_msec = time,
+		.x = (double)x / x_extent,
```
emersion: We should check here that `x_extent > 0 && y_extent > 0`, or we can be killed by a division by zero.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,346 @@
...
+static void virtual_pointer_frame(struct wl_client *client,
+		struct wl_resource *resource) {
+	struct wlr_virtual_pointer_v1 *pointer =
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+
+	if (pointer->axis_event[pointer->axis].device) {
```
emersion: Inside a single frame, there can be an event on each axis (diagonal move):

```
axis(time=42, axis=vertical, value=69)
axis(time=42, axis=horizontal, value=69)
frame()
```

In this case we need to trigger both events, not just horizontal.

types/wlr_virtual_pointer_v1.c
```diff
@@ -0,0 +1,346 @@
...
+		virtual_pointer_from_resource(resource);
+	if (pointer == NULL) {
+		return;
+	}
+	struct wlr_input_device *wlr_dev = &pointer->input_device;
+	pointer->axis = axis;
+	pointer->axis_event[pointer->axis].device = wlr_dev;
+	pointer->axis_event[pointer->axis].time_msec = time;
+	pointer->axis_event[pointer->axis].orientation = axis;
+	wlr_signal_emit_safe(&wlr_dev->pointer->events.axis,
```
emersion: I think we should set `delta`/`delta_discrete` to zero and let the `frame` request handler send the event. This will prevent two `axis` events for the same axis from being sent during the same frame.

ddevault:
Rebased and pushed a new set of updates. Note that wayvnc now supports this protocol:

https://github.com/any1/wayvnc

PR #1881, opened 2019-10-26
backend/drm: add support for custom modes

emersion:
Use the CVT algorithm to create a drmModeModeInfo.

* * *

Breaking change: `wlr_drm_connector_add_mode` now returns the new mode instead of a `bool`.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -23,6 +23,7 @@
 #include <wlr/util/log.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
+#include "backend/drm/cvt.h"
 #include "backend/drm/drm.h"
 #include "backend/drm/iface.h"
 #include "backend/drm/util.h"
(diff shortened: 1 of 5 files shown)


backend/drm/drm.c
```diff
@@ -672,25 +673,39 @@ bool drm_connector_set_mode(struct wlr_output *output,
 	return true;
 }
 
-bool wlr_drm_connector_add_mode(struct wlr_output *output,
+static bool drm_connector_set_custom_mode(struct wlr_output *output,
+		int32_t width, int32_t height, int32_t refresh) {
+	drmModeModeInfo mode = {0};
+	generate_cvt_mode(&mode, width, height, (float)refresh / 1000, false, false);
+	mode.flags |= DRM_MODE_TYPE_USERDEF;
```
RedSoxFan: _I know this is listed as WIP, but I tried it anyway_
This needs a `s/flags/type/` to avoid the early return in `wlr_drm_connector_add_mode`

emersion: Indeed! Fixed

ddevault:
Thanks!

montvid:
@emersion just wanted to comment that using CVT I get 69.xx hz and using GTF I get 70hz in Gnome monitor settings. So I am using GTF to get exact 70hz as I read before all other output settings are used for CRT monitors only and not needed for LCD. I would suggest to compare what command would serve better overall as I had to choose GTF.
```
gtf 1920 1080 70
 1920x1080 @ 70.00 Hz (GTF) hsync: 78.68 kHz; pclk: 205.20 MHz
  Modeline "1920x1080_70.00"  205.20  1920 2056 2264 2608  1080 1081 1084 1124  -HSync 
+Vsync
```
```
cvt 1920 1080 70
1920x1080 69.92 Hz (CVT) hsync: 78.80 kHz; pclk: 204.25 MHz
Modeline "1920x1080_70.00"  204.25  1920 2056 2256 2592  1080 1083 1088 1127 -hsync +vsync
```

montvid:
CVT gives exotic Hz rates that are not there by default in waymsg -t get_outputs. I bet GTF works better and gives 60Hz instead of 59.xx. I tried 60Hz for example:
https://github.com/swaywm/sway/pull/4675#issuecomment-602075296

PR #1882, opened 2019-10-27
Fix heap-use-after-free in wlr_send_tablet_v2_tablet_pad_leave

kuruczgy:
See swaywm/sway#4660

This change does fix the crash, but I am not sure if this is the proper way to do it, so I want someone with knowledge about wlroots memory management to review this.

cc @emersion @Ongy

### DIFF
--- a/types/tablet_v2/wlr_tablet_v2_pad.c
+++ b/types/tablet_v2/wlr_tablet_v2_pad.c
@@ -171,6 +171,9 @@ void destroy_tablet_pad_v2(struct wl_resource *resource) {
 	}
 	free(pad->strips);
 
+	if (pad->pad->current_client == pad) {
+		pad->pad->current_client = NULL;
+	}
 	free(pad);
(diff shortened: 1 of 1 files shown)


types/tablet_v2/wlr_tablet_v2_pad.c
```diff
@@ -171,6 +171,7 @@ void destroy_tablet_pad_v2(struct wl_resource *resource) {
 	}
 	free(pad->strips);
 
+	if (pad->pad->current_client == pad) pad->pad->current_client = NULL;
```
ddevault: Style: braces are mandatory.

ddevault:
Will need to rebase those commits together before the final merge

kuruczgy:
> Will need to rebase those commits together before the final merge

Do I also need to squash them together, or will that be done automatically? (Sorry, this is my first time sending a PR on github.)

ddevault:
Squash, yes please.

https://git-rebase.io

PR #1886, opened 2019-10-29
Fix geo_box bug in cursor resizing

coderkalyan:
While trying out the tinywl code, I found that the resize mode was behaving weirdly ... so I looked into code. Turns out the `begin_interactive` method stores the cursor position plus the geo_box position; however, `process_cursor_resize` wasn't taking this into account, causing windows to jump down in size unexpectedly when resized and lose alignment with the cursor. To fix this, I simply added a member to the `tinywl_server` struct that stores the geo_box when the mouse enters grab mode, and I referenced that data in the resize method. I considered polling for this data every time instead of storing it in the server struct, but 1) since changes in this value are not relevant and 2) it could potentially decrease performance (I don't know enough about wlroots to know how much) I decided to just store it. I can change this if desired.

Hope this makes tinywl a little smoother!

### DIFF
--- a/tinywl/tinywl.c
+++ b/tinywl/tinywl.c
@@ -54,6 +54,7 @@ struct tinywl_server {
 	enum tinywl_cursor_mode cursor_mode;
 	struct tinywl_view *grabbed_view;
 	double grab_x, grab_y;
+	struct wlr_box grab_geo_box;
 	int grab_width, grab_height;
 	uint32_t resize_edges;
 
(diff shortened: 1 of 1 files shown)


tinywl/tinywl.c
```diff
@@ -712,17 +713,17 @@ static void begin_interactive(struct tinywl_view *view,
 	}
 	server->grabbed_view = view;
 	server->cursor_mode = mode;
-	struct wlr_box geo_box;
-	wlr_xdg_surface_get_geometry(view->xdg_surface, &geo_box);
+	//struct wlr_box geo_box;
```
ddevault: Drop this line

ddevault:
LGTM

coderkalyan:
@ddevault Completely forgot about this! I just resubmitted the changes in a new PR: https://github.com/swaywm/wlroots/pull/2132 to avoid any merge conflicts. I also removed the commented line :)

coderkalyan:
@ddevault Just rebased and force pushed, here you go!

Issue #1889, opened 2019-10-30
drm output disable fails with dsi panel


agx:
For a LCD panel connected via dsi (nwl dsi host controller + (mxsfb of  dcss display controller)) a  
`wlr_output_enable (output->wlr_output, false);` fails like 

```
Oct 30 17:02:19 pureos phoc[390]: [subprojects/wlroots/backend/drm/atomic.c:55] DSI-1: Atomic commit failed (modeset): Invalid argument
```

this with `drm.debug=0x3f`

```
[  258.912733] [drm:drm_ioctl] pid=1175, dev=0xe201, auth=1, DRM_IOCTL_MODE_ATOMIC
[  258.912759] [drm:drm_atomic_state_init] Allocated atomic state 00000000d3edbb8c
[  258.912783] [drm:drm_mode_object_get] OBJ ID: 37 (2)
[  258.912797] [drm:drm_atomic_get_crtc_state] Added [CRTC:32:crtc-0] 00000000963a1f74 state to 00000000d3edbb8c
[  258.912809] [drm:drm_mode_object_put.part.3] OBJ ID: 37 (3)
[  258.912824] [drm:drm_atomic_set_mode_prop_for_crtc] Set [NOMODE] for [CRTC:32:crtc-0] state 00000000963a1f74
[  258.912843] [drm:drm_mode_object_get] OBJ ID: 34 (4)
[  258.912852] [drm:drm_mode_object_get] OBJ ID: 34 (5)
[  258.912864] [drm:drm_atomic_get_connector_state] Added [CONNECTOR:34:DSI-1] 00000000c425d48f state to 00000000d3edbb8c
[  258.912875] [drm:drm_mode_object_put.part.3] OBJ ID: 34 (6)
[  258.912888] [drm:drm_atomic_set_crtc_for_connector] Link [CONNECTOR:34:DSI-1] state 00000000c425d48f to [NOCRTC]
[  258.912897] [drm:drm_mode_object_put.part.3] OBJ ID: 34 (5)
[  258.912909] [drm:drm_atomic_check_only] checking 00000000d3edbb8c
[  258.912927] [drm:drm_atomic_helper_check_modeset] [CRTC:32:crtc-0] mode changed
[  258.912937] [drm:drm_atomic_helper_check_modeset] [CRTC:32:crtc-0] enable changed
[  258.912947] [drm:drm_atomic_helper_check_modeset] [CRTC:32:crtc-0] active changed
[...]

PR #1891, opened 2019-11-01
backend/drm: Retry disable without unsetting mode and crtc id

agx:
otherwise we can't blank/unblank on DSI with fixed pipeline.

Closes: #1889

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -101,6 +101,7 @@ static bool atomic_crtc_pageflip(struct wlr_drm_backend *drm,
 		struct wlr_drm_connector *conn,
 		struct wlr_drm_crtc *crtc,
 		uint32_t fb_id, drmModeModeInfo *mode) {
+	uint32_t flags = DRM_MODE_PAGE_FLIP_EVENT;
 	if (mode != NULL) {
 		if (crtc->mode_id != 0) {
 			drmModeDestroyPropertyBlob(drm->fd, crtc->mode_id);
@@ -111,10 +112,6 @@ static bool atomic_crtc_pageflip(struct wlr_drm_backend *drm,
 			wlr_log_errno(WLR_ERROR, "Unable to create property blob");
 			return false;
 		}
-	}
-
-	uint32_t flags = DRM_MODE_PAGE_FLIP_EVENT;
-	if (mode != NULL) {
 		flags |= DRM_MODE_ATOMIC_ALLOW_MODESET;
 	} else {
 		flags |= DRM_MODE_ATOMIC_NONBLOCK;
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -117,6 +117,11 @@ static bool atomic_crtc_pageflip(struct wlr_drm_backend *drm,
 		flags |= DRM_MODE_ATOMIC_NONBLOCK;
 	}
 
+	if (crtc->allow_mode_set) {
+		flags |= DRM_MODE_ATOMIC_ALLOW_MODESET;
+		crtc->allow_mode_set = false;
+	}
+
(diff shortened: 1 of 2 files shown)


emersion:
We really want to unset the CRTC, otherwise it keeps the pipeline on and we can't use it for another connector.

agx:
@ascent12 can you point me to the uapi parts that state it should work? With the tracing i did in the kernel it seemed to not even hit the display driver (let alone lower layers like panel or dsi host controller)  upon tht request but rejects things in generic drm already, see also the debug output in #1889

agx:
this looks pretty different now and works for me on arm64 (mxsfb) and intel.

backend/drm/atomic.c
```diff
@@ -140,16 +140,19 @@ static bool atomic_conn_enable(struct wlr_drm_backend *drm,
 		return !enable;
 	}
 
+	/* crtc will be enabled on next page flip */
+	if (enable) {
+		crtc->needs_modeset = true;
```
emersion: This seems to set `DRM_MODE_ATOMIC_ALLOW_MODESET`. However we already set it in the `atomic_commit` call, so I don't think it's necessary.

backend/drm/atomic.c
```diff
@@ -140,16 +140,19 @@ static bool atomic_conn_enable(struct wlr_drm_backend *drm,
...
 	atomic_begin(crtc, &atom);
 	atomic_add(&atom, crtc->id, crtc->props.active, enable);
-	if (enable) {
-		atomic_add(&atom, conn->id, conn->props.crtc_id, crtc->id);
-		atomic_add(&atom, crtc->id, crtc->props.mode_id, crtc->mode_id);
-	} else {
-		atomic_add(&atom, conn->id, conn->props.crtc_id, 0);
-		atomic_add(&atom, crtc->id, crtc->props.mode_id, 0);
-	}
+	atomic_add(&atom, conn->id, conn->props.crtc_id, 0);
```
emersion: This disables the plane even if `enable == true`.

agx: no, we return in early in the `enable` case, see above.

emersion: Hmm. So we don't commit anymore. If we don't commit, we don't know whether we can enable the output (bandwidth limits).

backend/drm/atomic.c
agx: it's for the case where we `enable` the crtc but `atomic_crtc_pageflip()` does not get passed a mode.

backend/drm/atomic.c
agx: but isn't this only really useful if we also assign plane props, etc. which happens on pageflip (which is where it happens now)?

agx:
Seems kmscube and weston use same pattern of putting the next commit instead of handling it separateley as wlroots currently does:

https://gitlab.freedesktop.org/wayland/weston/blob/master/libweston%2Fbackend-drm%2Fkms.c#L972

This allows to have this together with the fb properties which is needed if we unassign that on disable (which seems needed for drivers that use `drm_simple_kms_crtc_check` (see #1889).

backend/drm/atomic.c
agx: Is there another solution i'm missing? it'd be nice to have these drm drivers also work out of the box with wlroots.

emersion:
This should be fixed by https://github.com/swaywm/wlroots/pull/2169 as well.

emersion:
@agx Is this PR still required with the new DRM backend changes that landed in master?

agx:
ah...did not see that ping back in may. I'll put checking that on my todo list (currently basing on 0.10.1 here)

emersion:
Closing because I think it's fixed.

PR #1894, opened 2019-11-02
backend/x11: Give X11 a real rendering loop

ascent12:
This is some work split out of one of my renderer v6 PRs.

Makes use of the present extension to get notified of vsync, and not require any stupid timer hacks. Also make use of the present version of ConfigureNotify, because why not?

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -5,6 +5,7 @@
 #include <string.h>
 
 #include <xcb/xcb.h>
+#include <xcb/present.h>
 #include <xcb/xinput.h>
 
 #include <wlr/interfaces/wlr_output.h>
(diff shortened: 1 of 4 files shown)


backend/x11/output.c
```diff
@@ -217,22 +187,61 @@ struct wlr_output *wlr_x11_output_create(struct wlr_backend *backend) {
 	wlr_signal_emit_safe(&x11->backend.events.new_input, &output->pointer_dev);
 	wlr_signal_emit_safe(&x11->backend.events.new_input, &output->touch_dev);
 
+	// Start the rendering loop
+	wlr_output_send_frame(&output->wlr_output);
```
emersion: I don't think we need to do this. The compositor will damage the output and call `wlr_output_schedule_frame`.

emersion:
This sounds like a good idea.

I get a black screen when running this on Sway with `WLR_BACKENDS=x11`.

ascent12:
>I get a black screen when running this on Sway with WLR_BACKENDS=x11.

Weird. All of the wlroots examples I tried were working fine.
I'll have to look into that more.

emersion:
FWIW, I get a black screen with both my updated rootston fork and sway.

ascent12:
I figured it out. I was forgetting to call `wlr_output_update_enabled`, which `wlr_output_damage` pays attention to, but the naive examples do not.

backend/x11/output.c
ascent12: Done

emersion:
Thanks!

PR #1895, opened 2019-11-02
session/logind: support CanGraphical property

### LINKED ISSUE #1424: Add logind CanGraphical support
See https://gitlab.gnome.org/GNOME/gdm/issues/103

RPigott:
Closes #1424.

A number of sway users have been reporting that sway fails to start on systemd 243 in swaywm/sway#4672 and elsewhere. I'm guessing this could be fixed by waiting for the logind CanGraphical property on the seat before trying TakeDevice.

Unfortunately, it seems that systemd-logind has been broken for some time and doesn't actually emit PropertiesChanged when it should for CanGraphical, see systemd/systemd#13811. So users with broken sway will need both patches for the fix.

The issue was tagged easy so I thought I'd give it a shot, but this is my first attempt with wlroots code, so there may be some bad assumptions or silly mistakes.

I'm not sure what should happen, if anything, when CanGraphical changes true->false. I'm not sure under what circumstances that would happen. I didn't add any behavior for that in this patch.

Also, I put the wait for CanGraphical in session_create but it only really needs to happen before TakeDevice I think. I didn't have a reference to the event loop in take_device though. I'm actually not even sure if it's appropriate to wait with wl_event_loop_dispatch anyway, everything I know about wayland I've basically just learned from ddevault's (excellent but unfinished) book. Is that right? Help?

Anyway, it compiles and seems to work for me. PTAL.

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -35,7 +35,9 @@ struct logind_session {
 
 	char *id;
 	char *path;
+	char *seat_path;
 
+	bool can_graphical;
 	// specifies whether a drm device was taken
 	// if so, the session will be (de)activated with the drm fd,
 	// otherwise with the dbus PropertiesChanged on "active" signal
@@ -175,6 +177,34 @@ out:
 	return ret >= 0;
 }
 
+static bool find_seat_path(struct logind_session *session) {
+	int ret;
+	sd_bus_message *msg = NULL;
+	sd_bus_error error = SD_BUS_ERROR_NULL;
+
+	ret = sd_bus_call_method(session->bus, "org.freedesktop.login1",
+			"/org/freedesktop/login1", "org.freedesktop.login1.Manager",
+			"GetSeat", &error, &msg, "s", session->base.seat);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to get seat path: %s", error.message);
+		goto out;
+	}
+
+	const char *path;
+	ret = sd_bus_message_read(msg, "o", &path);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Could not parse seat path: %s", error.message);
+		goto out;
+	}
+	session->seat_path = strdup(path);
+
+out:
+	sd_bus_error_free(&error);
+	sd_bus_message_unref(msg);
+
+	return ret >= 0;
+}
+
 static bool session_activate(struct logind_session *session) {
 	int ret;
 	sd_bus_message *msg = NULL;
@@ -236,6 +266,7 @@ static void logind_session_destroy(struct wlr_session *base) {
 	sd_bus_unref(session->bus);
 	free(session->id);
 	free(session->path);
+	free(session->seat_path);
 	free(session);
 }
 
@@ -328,7 +359,7 @@ error:
 	return 0;
 }
 
-static int properties_changed(sd_bus_message *msg, void *userdata,
+static int session_properties_changed(sd_bus_message *msg, void *userdata,
 		sd_bus_error *ret_error) {
 	struct logind_session *session = userdata;
 	int ret = 0;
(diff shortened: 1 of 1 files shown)


backend/session/logind.c
```diff
@@ -672,6 +816,27 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
+		sd_bus_error_free(&error);
+		goto error_bus;
+	}
+
+	if (!session->can_graphical) {
+		wlr_log(WLR_INFO, "Waiting for 'CanGraphical' on seat %s", session->base.seat);
+	}
+
+	while (!session->can_graphical) {
+		wl_event_loop_dispatch(event_loop, -1);
```
emersion: This function can return an error. Need to exit when it's the case, instead of entering in a busy loop.

backend/session/logind.c
```diff
@@ -437,6 +468,105 @@ static int properties_changed(sd_bus_message *msg, void *userdata,
...
+	int ret = 0;
+
+	// if we have a drm fd we don't depend on this
+	if (session->has_drm) {
+		return 0;
+	}
+
+	// PropertiesChanged arg 1: interface
+	const char *interface;
+	ret = sd_bus_message_read_basic(msg, 's', &interface); // skip path
```
ascent12: I'm not sure what this comment is supposed to mean.

backend/session/logind.c
```diff
@@ -672,6 +816,27 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
...
+	ret = sd_bus_get_property_trivial(session->bus,
+		"org.freedesktop.login1", session->seat_path,
+		"org.freedesktop.login1.Seat", "CanGraphical", &error,
+		'b', &session->can_graphical);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to get 'CanGraphical' property: %s",
+			error.message);
+		sd_bus_error_free(&error);
+		goto error_bus;
+	}
```
ascent12: Wouldn't it be better to use `sd_seat_can_graphical` instead of calling out to dbus here?

backend/session/logind.c
```diff
@@ -437,6 +468,105 @@ static int properties_changed(sd_bus_message *msg, void *userdata,
...
+			sd_bus_error error = SD_BUS_ERROR_NULL;
+			ret = sd_bus_get_property_trivial(session->bus,
+				"org.freedesktop.login1", session->seat_path,
+				"org.freedesktop.login1.Seat", "CanGraphical", &error,
+				'b', &session->can_graphical);
+			if (ret < 0) {
+				wlr_log(WLR_ERROR, "Failed to get 'CanGraphical' property: %s",
+					error.message);
+				return 0;
+			}
```
ascent12: Also `sd_seat_can_graphical`.

ascent12:
>LGTM overall, I just wonder whether we should really block like this. The alternative would be to introduce a signal, but it's likely not worth the effort.

Yeah, I don't think it's really worth it right now either. Blocking is the only sane thing to do, unless we rejig all of the initialisation stuff.

backend/session/logind.c
RPigott: Makes sense, thanks. I added an error handling bit.

backend/session/logind.c
RPigott: I'm not actually sure either. I copied/modified this bit from the original (session_)properties_changed function, and the same comment remains there too.

It sorta seems like the comment thinks the reader might otherwise expect PropertiesChanged to return the object path, but the signature is always 'sa{sv}as' beginning with the interface name, so I'm not sure why the reader would expect that...

backend/session/logind.c
RPigott: Yes that sounds better. Changed.

ascent12:
Thanks!

PR #1896, opened 2019-11-03
session/logind: Clean up add_signal_matches

ascent12:
The original signal matching was using the old interface before sd_bus_match_signal was added, which the new code uses. Change them all to use it now.


emersion:
Thanks!

Issue #1770, opened 2019-07-25
[request] nearest neighbour scaling on xwayland apps


luispabon:
Currently, xwayland apps look pretty blurry on outputs set to scale 2, which is problematic in terms of eye fatigue (most major IDEs for instance aren't wayland native). 

Since https://github.com/swaywm/wlroots/issues/1635 depends on xwayland changes that could take some time to get merged & released (if ever), would it be possible to change the scaling mechanism to use nearest neighbour instead?

emersion:
I'd be interested to see:

- How simple it would be to change the algorithm
- How much of a difference this makes for LoDPI apps
- How much GPU time it uses

dos1:
I suppose we're talking about rootston there.

- `glTexParameteri(target, GL_TEXTURE_MAG_FILTER, GL_NEAREST);`
- it makes a huge difference (as in, it's clearly noticeable), although I'm not exactly sure that one is always better than the other - leaving it to a config option would be the best idea IMO
- none. Nearest neighbor interpolation is actually mathematically simpler than linear, so it would potentially be even faster, however, with the way modern GPUs work there shouldn't be any difference whatsoever.

emersion:
@ascent12 Thoughts?

emersion:
Well it guess it would be `GL_TEXTURE_MIN_FILTER`, since we're displaying LoDPI apps on HiDPI screens.

ascent12:
It'll probably look terrible with fractional scaling, but I'd rather not let that bad idea get in the way of good integer scaling.
I wouldn't oppose it if someone opened a PR. Some comparison screenshots (especially involving text) would be good too.

luispabon:
@dos1 thanks for pointing out where the magic was happening. I tried all the filters I could find on the gl headers:
```c
GL_NEAREST
GL_LINEAR
GL_NEAREST_MIPMAP_NEAREST
GL_LINEAR_MIPMAP_NEAREST
GL_NEAREST_MIPMAP_LINEAR
GL_LINEAR_MIPMAP_LINEAR
```
Screenshots: [scaling_shots.zip](https://github.com/swaywm/wlroots/files/3435356/scaling_shots.zip)

dos1:
[...]

PR #1898, opened 2019-11-05
renderer: use GL_NEAREST with integer scale factor

colemickens:
Fixes: #1770

This was based on a suggestion offered in IRC regarding implementing this feature without breaking the existing API. In this case, I've added getters/setters to the renderer/renderer_impl so the output can stash the scale inside the renderer's state so the renderer knows whether to use GL_NEAREST or GL_LINEAR.

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -58,6 +58,7 @@ struct wlr_gles2_renderer {
 	} shaders;
 
 	uint32_t viewport_width, viewport_height;
+	float scale;
 };
 
 enum wlr_gles2_texture_type {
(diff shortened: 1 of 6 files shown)


progandy:
I wonder if you could choose GL_NEAREST depending on whether the matrix contains only integer values? Is that too expensive or is there a risk of errors introduced through floating point arithmetic?

colemickens:
@progandy I also wondered about that, but linear algebra was ages ago and I don't know how these GL transforms work, etc.

ddevault:
I agree with @emersion, this isn't really the best place to do this imo.

Issue #1826, opened 2019-09-19
High-level scene API


ascent12:
Over time, I've been pushing to make many of our API lower-level and efficient as they could reasonably be. Unfortunately, this does make things much harder to use, especially for people without specialised knowledge, making the barrier to entry still very high for using wlroots.

Basically, I think we need a restricted "easy mode" API that can help people still do everything as efficiently as possible (all handled internally). This honestly wouldn't be that different than WLC or a parts of libweston.

Instead of the user handling rendering with damage tracking, planes, and all that shit we want to add, they just say "the window goes here".

I don't know how high-level this really should be. Should this handle shells, drag-and-drop, or other complex operations? Does this even belong in wlroots itself or should it exist as a layered library in its own repository?

The name I was thinking for it is `wlr_desktop`.

AdrianVovk:
I think a separate library would be the best place for this. The should also be a way to hook into wlroots itself if the user chooses to do something a bit more custom sometimes

ascent12:
I was leaning towards external too. Adding some ways to "go lower" would be nice, but we need to be careful to not make doing that really complex, defeating the purpose of the API.

ddevault:
I think we would need a more detailed design doc written up before we could judge whether to include it in wlroots or not.

Rather than `wlr_desktop`, which implies a focus on the desktop use-case, I'd prefer `wlr_scene`.

emersion:
+1 for `wlr_scene`. I think such a helper would make planes a lot less annoying to use. It could just be a bunch of buffers to draw at certain positions (ala [libliftoff](https://github.com/emersion/libliftoff)).

Bl4ckb0ne:
Hi there, I'm working on [https://git.sr.ht/~bl4ckb0ne/wltrunk](wltrunk), a library based on wlroots that aims to bring a higher level API. I'll gladly discuss anything on either irc (i'm on #sway-devel with the same handle) or by mail.

ddevault:
I don't think I would like to see the scene API end up in wltrunk.

emersion:
Note that `wlr_scene` could simplify damage tracking and could allow us not to draw regions of buffers hidden behind other opaque buffers.

ascent12:
I've seriously been thinking about implementing this.
[...]

PR #1900, opened 2019-11-06
render/gles: Simplify textures a bit

ascent12:
While I was writing https://github.com/swaywm/wlroots/issues/1826#issuecomment-550171172 and looking through our code, I realised it's not as simple as it should be.

We don't need our own enum for types. Instead we just use GL_TEXTURE_{2D,EXTERNAL_OES}, which already describes usage.

Also fixes a situation where we were using GL_TEXTURE_2D in a situation we should not have. wl_drm buffers are always GL_TEXTURE_EXTERNAL_OES, no matter if they're RGB or any other format.

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -60,31 +60,24 @@ struct wlr_gles2_renderer {
 	uint32_t viewport_width, viewport_height;
 };
 
-enum wlr_gles2_texture_type {
-	WLR_GLES2_TEXTURE_GLTEX,
-	WLR_GLES2_TEXTURE_WL_DRM_GL,
-	WLR_GLES2_TEXTURE_WL_DRM_EXT,
-	WLR_GLES2_TEXTURE_DMABUF,
-};
-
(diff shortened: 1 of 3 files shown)


include/render/gles2.h
```diff
@@ -60,31 +60,24 @@ struct wlr_gles2_renderer {
...
-	int width, height;
-	bool has_alpha;
-	enum wl_shm_format wl_format; // used to interpret upload data
-	bool inverted_y;
 
-	// Not set if WLR_GLES2_TEXTURE_GLTEX
+	// Basically:
+	//   GL_TEXTURE_2D == mutable
+	//   GL_TEXTURE_EXTERNAL_OES == immutable
+	EGLint target;
```
emersion: Nit: I guess this should be `GLint`

ascent12: You're right. I don't know why I used EGL there.
GLenum is an even more correct type.

PR #1901, opened 2019-11-06
render/gles2: provide public API to access GL texture

emersion:
Prior to this commit, compositors needed to render the texture to an
intermediate off-screen buffer using wlr_renderer APIs if they wanted to
use a custom rendering path (e.g. render to a 3D scene).

A new wlr_gles2_texture_get_attribs exposes the GL texture target and ID
so that compositors can render wlr_textures with their own shaders. An
example of a compositor doing so is available at [1].

[1]: https://git.sr.ht/~sircmpwn/wxrc/tree/3db905b7842ac42cf1878876e647005b41f00a52/src/render.c#L227

### DIFF
--- a/include/wlr/render/gles2.h
+++ b/include/wlr/render/gles2.h
@@ -9,6 +9,7 @@
 #ifndef WLR_RENDER_GLES2_H
 #define WLR_RENDER_GLES2_H
 
+#include <GLES2/gl2.h>
 #include <wlr/backend.h>
 #include <wlr/render/wlr_renderer.h>
 
@@ -23,5 +24,17 @@ struct wlr_texture *wlr_gles2_texture_from_wl_drm(struct wlr_egl *egl,
 	struct wl_resource *data);
 struct wlr_texture *wlr_gles2_texture_from_dmabuf(struct wlr_egl *egl,
 	struct wlr_dmabuf_attributes *attribs);
+
+struct wlr_gles2_texture_attribs {
+	GLenum target; /* either GL_TEXTURE_2D or GL_TEXTURE_EXTERNAL_OES */
+	GLuint tex;
+
+	bool inverted_y;
+	bool has_alpha;
+};
+
+bool wlr_texture_is_gles2(struct wlr_texture *texture);
+void wlr_gles2_texture_get_attribs(struct wlr_texture *texture,
+	struct wlr_gles2_texture_attribs *attribs);
 
 #endif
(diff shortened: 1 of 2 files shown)


include/wlr/render/gles2.h
```diff
@@ -24,4 +25,16 @@ struct wlr_texture *wlr_gles2_texture_from_wl_drm(struct wlr_egl *egl,
 struct wlr_texture *wlr_gles2_texture_from_dmabuf(struct wlr_egl *egl,
 	struct wlr_dmabuf_attributes *attribs);
 
+struct wlr_gles2_texture_attribs {
+	GLenum target;
```
ammen99: Can we somehow let the library user know which targets to expect?

ascent12: Probably just a comment mentioning `GL_TEXTURE_2D` and `GL_TEXTURE_EXTERNAL_OES`. It'll never be anything other than that, even in the future, since all of the dmabuf type stuff is inherently 2D.

include/wlr/render/gles2.h
```diff
@@ -24,4 +25,16 @@ struct wlr_texture *wlr_gles2_texture_from_wl_drm(struct wlr_egl *egl,
...
+	GLuint tex;
+
+	bool inverted_y;
+	bool has_alpha;
+};
```
ascent12: This design stops us from ever supporting multi-planar formats (excluding the ones the driver does for us), because multiple OpenGL textures may be involved.

emersion: I think it will be reasonable to break this API when we get support for multiplanar formats.

include/wlr/render/gles2.h
emersion: That's a good point.

emersion:
Actually, I've been thinking about a way to remove some complexity from compositor fragment shaders. It turns out it's perfectly valid to attach multiple fragment shaders to a single program as long as only one shader has a `main` function. So we could expose a fragment shader defining a `wlr_texture_get(x, y)` GLSL function that compositors could link to and use in their own programs. This would work nicely with multi-planar textures and color management.

ddevault:
Can we do that separately?

emersion:
Oh, yeah, of course. It'll require a lot more changes, so it's a better idea to do it separately. Let's open a new issue for this idea.

ddevault:
Thanks!

PR #1903, opened 2019-11-07
Introduce wlr_renderer_get_egl

ddevault:


include/wlr/render/interface.h
```diff
@@ -68,6 +68,7 @@ struct wlr_renderer_impl {
 	void (*destroy)(struct wlr_renderer *renderer);
 	void (*init_wl_display)(struct wlr_renderer *renderer,
 		struct wl_display *wl_display);
+	struct wlr_egl *(*get_egl)(struct wlr_renderer *renderer);
```
emersion: Why do we need this to be in the `wlr_renderer` interface? I'd prefer to make this GLES2-specific, with just a new `wlr_gles2_renderer_get_egl` function in `include/wlr/render/gles2.h`.

ddevault: Why should it be EGL specific? All sorts of renderers use EGL. In fact, the renderer create function accepts a wlr_egl, so EGL is brought to the table even earlier than the renderer's initialization.

emersion: Yeah, but we really didn't have the choice, and it's definitely something we want to remove (and find a better solution for). Vulkan and software renderers wouldn't be tied to EGL at all, and even OpenGL-based renderers shouldn't need to use `wlr_egl` (we want it to become just an optional utility).

ddevault: Yeah, but all of this is going to be thrown away for renderer v6 anyway

emersion: Renderer v6 won't happen anytime soon (if it happens at all). I don't think it's a good reason to do whatever we want with the current interface.

We used to have `wlr_backend_get_egl` and finally got rid of it, I'd be sad if we re-introduced something similar like this.

@ascent12 What do you think?

ammen99:
The idea gets +1 from me, Wayfire already does hacks to get this: https://github.com/WayfireWM/wayfire/blob/master/src/main.cpp#L61

include/wlr/render/interface.h
ddevault: Acknowledging the design of the current interface is *why* I designed it this way.

emersion: See https://github.com/swaywm/wlroots/issues/1352#issuecomment-436804658 paragraph wlr_egl:

>This has been "demoted" to a pure utility type for writing OpenGL renderers. If you're not writing exactly that, you would not be touching this type.
>
>[…]
>
>Since we're dealing with non-default framebuffers in OpenGL, people writing their own renderers are free to attach their own depth or stencil buffers, making wlr_renderer_create_func_t unnecessary.

ddevault: Alright, I'll move it to the gles code.

PR #1909, opened 2019-11-10
Refactor EGL/GL API loading

emersion:
Remove glapi.sh code generation, replace it with hand-written loading
code that checks extension strings before calling eglGetProcAddress.

The GLES2 renderer still uses global state because of:

- {PUSH,POP}_GLES2_DEBUG macros
- wlr_gles2_texture_from_* taking a wlr_egl instead of the renderer


render/egl.c
```diff
@@ -125,19 +135,95 @@ static void init_dmabuf_formats(struct wlr_egl *egl) {
...
+			return false;
+		}
+	}
+	if (check_egl_ext(exts_str, "EGL_KHR_swap_buffers_with_damage")) {
+		egl->exts.swap_buffers_with_damage_khr = true;
+		if (!load_egl_proc(&egl->procs.eglSwapBuffersWithDamageKHR,
+				"eglSwapBuffersWithDamageKHR")) {
+			return false;
+		}
+	}
```
ascent12: We don't need to save both of these. I'd check for KHR and only use EXT if that fails.
They have exactly the same function signature and semantics.

emersion: Done

render/egl.c
```diff
@@ -77,6 +76,16 @@ static bool check_egl_ext(const char *exts, const char *ext) {
 	return false;
 }
 
+static bool load_egl_proc(void *proc_ptr, const char *name) {
+	void *proc = (void *)eglGetProcAddress(name);
+	if (proc == NULL) {
+		wlr_log(WLR_ERROR, "eglGetProcAddress(%s) failed", name);
+		return false;
```
ascent12: I wonder if it's worth replacing this with `abort()`.
A driver failing to give us a function after advertising the extension sounds like a serious driver bug, or our bug because we spelt the function name wrong.

I don't think this is worth trying to recover from, and greatly simplifies some of the conditions later.
Weston doesn't even bother to check the return, probably for the same reasons.

render/egl.c
ascent12: Was this lost in a rebase or something? It's still here.

include/wlr/render/egl.h
```diff
@@ -47,6 +47,23 @@ struct wlr_egl {
...
+		PFNEGLBINDWAYLANDDISPLAYWL eglBindWaylandDisplayWL;
+		PFNEGLUNBINDWAYLANDDISPLAYWL eglUnbindWaylandDisplayWL;
+		PFNEGLSWAPBUFFERSWITHDAMAGEEXTPROC eglSwapBuffersWithDamageEXT;
+		PFNEGLSWAPBUFFERSWITHDAMAGEKHRPROC eglSwapBuffersWithDamageKHR;
+		PFNEGLQUERYDMABUFFORMATSEXTPROC eglQueryDmaBufFormatsEXT;
+		PFNEGLQUERYDMABUFMODIFIERSEXTPROC eglQueryDmaBufModifiersEXT;
+		PFNEGLEXPORTDMABUFIMAGEQUERYMESAPROC eglExportDMABUFImageQueryMESA;
+		PFNEGLEXPORTDMABUFIMAGEMESAPROC eglExportDMABUFImageMESA;
+		PFNEGLDEBUGMESSAGECONTROLKHRPROC eglDebugMessageControlKHR;
+	} procs;
```
ascent12: Do we want to take the opportunity to change these to more aesthetically pleasing names?
e.g. `eglGetPlatformDisplayEXT -> get_platform_display`.

ddevault: -1, I think we ought to stick with the names we'd normally find in the header. I hate that Weston does this.

render/egl.c
emersion: Hmm, yeah I probably overwrote the fix during a rebase. Fixed again.

render/egl.c
emersion: Yeah, I agree. Done.

emersion:
Ping

ascent12:
Thanks

PR #1910, opened 2019-11-11
Add wlr_buffer interface

ddevault:
This allows downstream compositors to create their own wlr_buffer
implementations.

This is a little bit smelly because wl_drm throws a wrench into things.
In the future I would like to introduce wl_resource_get_implementation
to libwayland, which would allow us to add a proxy around mesa's wl_drm
implementation so we can track our own state alongside it. With this
change, we'd be able to swap out the wl_resource parameters given to
each wlr_buffer function with wlr_buffer references, and drop the
special-cased codepaths for wl_drm.

### DIFF
--- a/include/wlr/interfaces/wlr_buffer.h
+++ b/include/wlr/interfaces/wlr_buffer.h
@@ -1,0 +1,75 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_IFACES_WLR_BUFFER_H
+#define WLR_IFACES_WLR_BUFFER_H
+#include <stdbool.h>
+#include <pixman.h>
+#include <wayland-server-core.h>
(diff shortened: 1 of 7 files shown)


include/wlr/interfaces/wlr_buffer.h
```diff
@@ -0,0 +1,68 @@
...
+struct wlr_buffer_impl {
+	/**
+	 * Return true if this wl_resource is an instance of a wl_buffer using this
+	 * API.
+	 *
+	 * Note: downstream compositors should not call this. We have to
+	 * special-case some buffer types and it will not necessarily work the way
+	 * you expect.
+	 */
+	bool (*is_instance)(struct wl_resource *resource);
```
ddevault: This whole field can be removed after #1911 is done.

ascent12:
I mentioned this on #wayland, but I think we should try and get devices added to linux-dmabuf and just drop wl_drm. The complexity and special cases it adds are just annoying, and I have no interest in supporting anything which can't do linux-dmabuf.

emersion:
https://gitlab.freedesktop.org/mesa/mesa/merge_requests/2798

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/3046

PR #1912, opened 2019-11-12
backend/wayland: add support for relative-pointer-unstable-v1

emersion:
We just send relative motion events alongside absolute motion events.
Compositors can figure out how absolute and relative events are related
(e.g. whether they have been triggered by the same logical event) with
the frame event.


ddevault:
Holding off on merge until we have a compositor which takes advantage of it (wxrc support has been discussed OOB)

emersion:
Note, nested Wayland compositors can already take advantage of this.

ddevault:
Thanks!

PR #1913, opened 2019-11-12
backend/wayland: expose remote objects

emersion:
Expose the remote wl_display, wl_surface and wl_seat used by the Wayland
backend.

This allows compositors to customize the Wayland backend and to have
more freedom. For instance a compositor might want to handle clipboard
and drag-and-drop from the remote Wayland compositor. Another compositor
might want to setup pointer constraints.


ddevault:
Thanks!

PR #1915, opened 2019-11-16
Simplify globals implementation by removing destructors

emersion:
Some globals are static and it doesn't make sense to destroy them before
the wl_display. For instance, wl_compositor should be created before the
display is started and shouldn't be destroyed.

For these globals, we can simplify the code by removing the destructor
and stop keeping track of wl_resources (these will be destroyed with the
wl_display by libwayland).

* * *

Breaking change: destructors and resource lists for static globals have been removed.


emersion:
This is now ready for review.

ddevault:
+206 −836 is an excellent patch. Thanks!

Issue #1917, opened 2019-11-16
Presentation feedback API doesn't identify the frame submission


emersion:
The presentation feedback API in `wlr_output` doesn't let backends specify for which frame submission the event is. So presentation feedback doesn't work properly if the delay is greater than one frame.

In practice, the DRM backend guarantees that the delay will be < 1 frame. However that's not the case for other backends.

We should add a way to track frame submissions (with a sequence number? with an object?). We already have `output->commit_seq`, we can probably re-use this for presentation feedback.

PR #1919, opened 2019-11-16
Improve presentation-time API

emersion:
The wlr_presentation_feedback struct now tracks presentation feedback
for multiple resources (but still a single surface content update). This
allows the compositor to properly send presentation events even when
there is more than one frame of latency or when it references a
surface's buffer.

Most of the time, compositors just display the surface's current buffer
on an output. Add an helper to make it easy to support presentation-time
in this case.

References: https://github.com/swaywm/sway/issues/4572
Closes: https://github.com/swaywm/wlroots/issues/1917

cc @YaLTeR 

* * *

Breaking change: instead of calling `wlr_presentation_surface_sampled` at render time and `wlr_presentation_send_surface_presented` at presentation time, most compositors now just need to call `wlr_presentation_surface_sampled_on_output` at render time. Compositors that call `wlr_buffer_ref` need to call `wlr_presentation_surface_sampled` when they ref the buffer, `wlr_presentation_feedback_send_presented` when this buffer has been presented for the first time and then `wlr_presentation_feedback_destroy`.

Sway PR: https://github.com/swaywm/sway/pull/4731

### DIFF
--- a/types/wlr_presentation_time.c
+++ b/types/wlr_presentation_time.c
@@ -9,31 +9,19 @@
 
 #define PRESENTATION_VERSION 1
 
-static struct wlr_presentation_feedback *presentation_feedback_from_resource(
-		struct wl_resource *resource) {
-	assert(wl_resource_instance_of(resource,
-		&wp_presentation_feedback_interface, NULL));
-	return wl_resource_get_user_data(resource);
-}
-
 static void feedback_handle_resource_destroy(struct wl_resource *resource) {
-	struct wlr_presentation_feedback *feedback =
-		presentation_feedback_from_resource(resource);
-	wl_list_remove(&feedback->surface_commit.link);
-	wl_list_remove(&feedback->surface_destroy.link);
-	wl_list_remove(&feedback->link);
-	free(feedback);
+	wl_list_remove(wl_resource_get_link(resource));
 }
 
-// Destroys the feedback
-static void feedback_send_presented(struct wlr_presentation_feedback *feedback,
+static void feedback_resource_send_presented(
+		struct wl_resource *feedback_resource,
 		struct wlr_presentation_event *event) {
-	struct wl_client *client = wl_resource_get_client(feedback->resource);
-	struct wl_resource *resource;
-	wl_resource_for_each(resource, &event->output->resources) {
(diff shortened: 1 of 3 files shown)


types/wlr_presentation_time.c
```diff
@@ -63,18 +51,38 @@ static void feedback_handle_surface_commit(struct wl_listener *listener,
...
 static void feedback_handle_surface_destroy(struct wl_listener *listener,
 		void *data) {
 	struct wlr_presentation_feedback *feedback =
 		wl_container_of(listener, feedback, surface_destroy);
-	feedback_send_discarded(feedback);
+	if (feedback->sampled) {
+		// The compositor might have a handle on this feedback
+		struct wl_resource *resource, *tmp;
+		wl_resource_for_each_safe(resource, tmp, &feedback->resources) {
+			feedback_resource_send_discarded(resource);
```
YaLTeR: Since this feedback was already sampled, isn't it possible for the compositor to present the sampled contents even after the surface has been destroyed, necessitating a `presented` rather than `discarded`?

emersion: Per the protocol, `discarded` should be sent if the surface is destroyed:

> or, the user did not see the content update because it was superseded or its surface destroyed, and the content update is discarded.

Anyway, if the surface is destroyed, it doesn't really matter, we just want to cleanup the feedback objects properly.

YaLTeR: > Per the protocol, `discarded` should be sent if the surface is destroyed:
> 
> > or, the user did not see the content update because it was superseded or its surface destroyed, and the content update is discarded.

This reads to me as "the user did not see the content update because \<for example\> its surface was destroyed \<before the compositor had a chance to obtain its contents\>", but I might be mistaken.

> Anyway, if the surface is destroyed, it doesn't really matter, we just want to cleanup the feedback objects properly.

Alright; if it's fine by wlroots internals then it works.

emersion: Ah, I see, you're right. It's not really more work to send `presented`, so I've just updated the PR to let the compositor do that instead.

ddevault:
Thanks!

PR #1924, opened 2019-11-19
build: Pass lib_wlr as first argument to pkgconfig.generate()

soreau:
Eliminates this warning when building wlroots as a subproject:

subprojects/wlroots/meson.build:216: DEPRECATION: Library wlroots
was passed to the libraries keyword argument of a previous call
to generate() method instead of first positional argument. Adding
wlroots to Requires field, but this is a deprecated behaviour
that will change in a future version of Meson. Please report the
issue if this warning cannot be avoided in your case.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -213,8 +213,7 @@ message('\n'.join(summary))
 subdir('examples')
 
 pkgconfig = import('pkgconfig')
-pkgconfig.generate(
-	libraries: lib_wlr,
+pkgconfig.generate(lib_wlr,
 	version: meson.project_version(),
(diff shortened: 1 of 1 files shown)


soreau:
> Ref, the Meson PR that introduced this behaviour: [mesonbuild/meson#4440](https://github.com/mesonbuild/meson/pull/4440)
> 
> (Interestingly I can't reproduce this warning in Sway w/ wlroots as a subproject, don't know why.)

I think since wf-config was also building as a subproject with wlroots as a dependency, that might have triggered it.

> Thanks!

No problem.

PR #1926, opened 2019-11-19
backend/x11: Revert usage of present extension

ascent12:
This reverts commit 3317134adff0ed179e0ecaea6d778dbd0684f771.
This reverts commit a3c3b928a3814f1a44babf487b1508415958c721.

There are some serious issues when running this on a real X server, as
opposed to running this on Xwayland, where this was tested.

More investigation needs to be done into why these issues happen and if
our usage of the present extension is correct.

### DIFF
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -12,7 +12,6 @@
 #include <X11/Xlib-xcb.h>
 #include <wayland-server-core.h>
 #include <xcb/xcb.h>
-#include <xcb/present.h>
 #include <xcb/xfixes.h>
 #include <xcb/xinput.h>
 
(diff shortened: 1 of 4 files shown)


emersion:
Seems like Present is broken by design:

```
10:46 <zzag> MrCooper: sigh, and it seems like there is no any way to differentiate between my events and Mesa's event :(
10:46 <zzag> events*
10:46 <emersion> and still receive both events, with eventid set to the one you generated?
10:46 <emersion> mesa's eventid should be different
10:48 <MrCooper> each event delivery target receives all events with its eventid
10:48 <zzag> emersion: that's correct! I receive two events for the same eventid (one that I requested, and the one that Mesa requested)
10:49 <emersion> that's weird
10:49 <emersion> when we tried to use present events in wlroots a while ago we ran into the same issue iirc
10:49 <emersion> we just assumed we used Present wrong
10:49 <MrCooper> the semantics might be like this accidentally, but there are clients relying on them, so we can't change them (I tried) :(
10:50 <emersion> iirc this completely breaks mesa, because mesa doesn't receive present events anymore?
10:50 <MrCooper> only the Gallium nine frontend
10:51 <MrCooper> which uses multiple display connections, and relies on events from Present requests sent on one of them to be delivered to the other one as well
10:52 <emersion> wew
```

PR #1928, opened 2019-11-19
Add -Wmissing-prototypes

emersion:
This requires functions without a prototype definition to be static.
This allows to detect dead code, export less symbols and put shared
functions in headers.


ascent12:
Thanks.

PR #1929, opened 2019-11-21
Unbreak examples/screencopy on FreeBSD

jbeich:
Similar to https://github.com/cyclopsian/wdisplays/pull/6
wlroots already uses `shm_open` correctly elsewhere:
https://github.com/swaywm/wlroots/blob/16e5e9541b3de49e397a3d2caa3212db25487648/util/shm.c#L29


RedSoxFan:
Thanks!

PR #1931, opened 2019-11-22
meson: Various improvements

ascent12:
Bumps minimum version to 0.51.0

- Remove all intermediate static libraries.
  They serve no purpose and are just add a bunch of boilerplate for
  managing dependencies and options. It's now managed as a list of
  files which are compiled into libwlroots directly.

- Use install_subdir instead of installing headers individually.
  I've changed my mind since I did that. Listing them out is annoying as
  hell, and it's easy to forget to do it.

- Add not_found_message for all of our optional dependencies that have a
  meson option. It gives some hints about what option to pass and what
  the optional dependency is for.

- Move all backend subdirectories into their own meson.build. This
  keeps some of the backend-specific build logic (especially rdp and
  session) more neatly separated off.

- Don't overlink example clients with code they're not using.
  This was done by merging the protocol dictionaries and setting some
  variables containing the code and client header file.
  Example clients now explicitly mention what extension protocols they
  want to link to.

- Split compositor example logic from client example logic.

- Minor formatting changes


ddevault:
>Remove all intermediate static libraries.

-1, I liked these for organizational purposes

emersion:
>organizational purposes

Can you elaborate?

ddevault:
By putting together static libraries and meson dependencies for our internal libraries, we can treat them no differently from any other dependency. I like that.

emersion:
The static libraries are not like any other dependency currently. They are added to the `wlr_parts` list, which is separate from the `wlr_deps` one. Adding files to a `wlr_files` list wouldn't be very different.

I prefer @ascent12's take because it doesn't link intermediary static libraries which are just thrown away after the final library is built.

ascent12:
>organizational purposes

The directory structure already is our organization.

If you went all-in with proper internal dependencies (which we weren't), there would be even more boilerplate.
Our code is more inter-dependent than not, and I'm sure it would lead to a bunch of circular dependencies.

The static_library approach was just acting as a container for some files, which isn't necessary at all.

emersion:
Needs rebase.

protocol/meson.build
```diff
@@ -1,103 +1,147 @@
...
+			output: '@BASENAME@-client-protocol.h',
+			command: [wayland_scanner, 'client-header', '@INPUT@', '@OUTPUT@'],
+		)
 
-wlr_protos = declare_dependency(
-	link_with: lib_wl_protos,
-	sources: wl_protos_headers,
-)
+		set_variable(name.underscorify() + '_c', code)
+		set_variable(name.underscorify() + '_h', header)
```
emersion: I'm not a very big fan of `set_variable` overall, because it makes it difficult to understand where they come from. I'd prefer to have a dict instead (e.g. `protocols_code.get(name)` would return the file for the generated code for instance).

protocol/meson.build
```diff
@@ -1,103 +1,147 @@
...
-	'wlr-input-inhibitor-unstable-v1.xml',
-	'wlr-layer-shell-unstable-v1.xml',
-	'wlr-output-management-unstable-v1.xml',
-	'wlr-screencopy-unstable-v1.xml',
-]
+protocols = {
+	# Stable upstream protocols
+	'xdg-shell': {
+		'path': wl_protocol_dir / 'stable/xdg-shell/xdg-shell.xml',
+		'client': true,
```
emersion: I wonder if it's worth it to worry about generating client code unnecessarily. As I understand it, the target will only get built if something else depends on it (even if it would get built unconditionally it wouldn't be a big deal).

We could remove the `client` field.

meson.build
```diff
@@ -102,41 +100,25 @@ wayland_client = dependency('wayland-client')
...
 drm            = dependency('libdrm', version: '>=2.4.95')
 gbm            = dependency('gbm', version: '>=17.1.0')
 libinput       = dependency('libinput', version: '>=1.9.0')
 xkbcommon      = dependency('xkbcommon')
 udev           = dependency('libudev')
 pixman         = dependency('pixman-1')
-libcap         = dependency('libcap', required: get_option('libcap'))
-logind         = dependency('lib' + get_option('logind-provider'), required: get_option('logind'), version: '>=237')
 math           = cc.find_library('m')
 rt             = cc.find_library('rt')
```
emersion: Minor style nit: while at it, can we stop cargo-culting the dependencies alignment here? We don't align in C, so I don't think we should align in Meson files.

backend/meson.build
```diff
@@ -32,51 +31,63 @@ backend_files = files(
...
-	drm,
-	egl,
-	gbm,
-	libinput,
-	pixman,
-	xkbcommon,
-	wayland_server,
-	wlr_protos,
-	wlr_render,
+# Client protocols used by wayland backend
```
emersion: Maybe we can move this to `backend/wayland/meson.build`?

emersion: Actually, I think it would be nicer to have exactly one Meson file per backend.

protocol/meson.build
ascent12: >I'd prefer to have a dict instead (e.g. protocols_code.get(name) would return the file for the generated code for instance).

That's actually a pretty good idea. Unfortunately meson (0.52.1) can't use a non string literal for a dictionary key:
```python
protocol_code = {}
foreach proto, path : protocols
    # ...
    protocol_code += { proto: code } # This is invalid
endforeach
```
And due to the way dictionaries work (i.e. they're pure) and some dictionary-of-dictionaries type stuff, I can't really think of a way to work around this (e.g. putting it in the protocols dictionary).

But this will be fixed in 0.53.0: https://mesonbuild.com/Syntax.html#dictionaries
It has been left as set_variable/get_variable currently, but changed to use an array of protocol strings in the wayland backend.
I think this should be revisited when the next meson version is out and we eventually target it for whatever reason.

protocol/meson.build
ascent12: Done. I just used build_by_default: false

meson.build
ascent12: Done

backend/meson.build
ascent12: Done

ascent12:
I'd also like to point out the second commit I've added just now. See the commit message for justification.

emersion:
Big +1 from me for the 2nd commit.

protocol/meson.build
emersion: I guess it's already the default since we require Meson >= 0.51. It doesn't hurt to explicitly set it though.

> the default value is false (changed in 0.50)

protocol/meson.build
emersion: Makes sense. Created a new issue so that we don't forget about it.

PR #1932, opened 2019-11-23
Amend typos

ammgws:


ammgws:
@ammen99 there is also one in `protocol/wlr-gamma-control-unstable-v1.xml`, should I add that too or leave it be?

ammen99:
I think it won't be a problem to add it to this PR.

ammen99:
@ammgws Can you also send a PR for the protocols repo? https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-gamma-control-unstable-v1.xml#L113

ammgws:
@ammen99 Done

ddevault:
Thanks!

PR #1935, opened 2019-11-26
wlr_keyboard_group: reference count destruction

RedSoxFan:
Fixes swaywm/sway#4744

This introduces a simple reference counter for destruction of
`wlr_keyboard_group` objects. This allows for the current operation to
complete (ex. `wlr_keyboard_notify_key` for the group's keyboard), before
the `wlr_keyboard_group`, and thus it's keyboard, are completely
destroyed.

### DIFF
--- a/include/wlr/types/wlr_keyboard_group.h
+++ b/include/wlr/types/wlr_keyboard_group.h
@@ -18,6 +18,8 @@ struct wlr_keyboard_group {
 	struct wlr_input_device *input_device;
 	struct wl_list devices; // keyboard_group_device::link
 	struct wl_list keys; // keyboard_group_key::link
+	bool destroying;
+	uint32_t nrefs;
 	void *data;
 };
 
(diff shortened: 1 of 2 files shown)


include/wlr/types/wlr_keyboard_group.h
```diff
@@ -18,6 +18,8 @@ struct wlr_keyboard_group {
 	struct wlr_input_device *input_device;
 	struct wl_list devices; // keyboard_group_device::link
 	struct wl_list keys; // keyboard_group_key::link
+	bool destroying;
```
emersion: Instead of having a `destroying` field, can we initiialize `nrefs` to 1 and make `wlr_keyboard_group_destroy` call `keyboard_group_unref`?

emersion:
Would be nice to add a note to `wlr_keyboard_group_destroy` to say it may not be destroyed immediately.

emersion:
I wonder if this is the right fix. I don't think any wlroots object supports being destroyed from one of its event handlers. Doing this for all objects (what if a reload is triggered from a mouse button?) would introduce complexity.

An alternative fix would be to use `wl_event_loop_add_idle` to schedule the Sway reload from the keyboard key handler.

PR #1937, opened 2019-11-27
Update environment variable documentation to include more backends

n3rdopolis:
Testing with exporting WLR_BACKENDS=drm worked, and it wasn't documented. Checking the backends folder, and it also mentions an RDP backend as well


ddevault:
Thanks!

PR #1940, opened 2019-11-30
keyboard: emit key events without keymap

emersion:
Sometimes compositors don't need an XKB keymap at all, they just handle
raw keycodes. Emit key events even if no keymap is set.


ddevault:
Thanks!

PR #1942, opened 2019-12-01
Revert "output: add block_idle_frame"

emersion:
This reverts commit cbb2781fed7944dae680a66a35443f5b1a678ec7.

In [1], we found issues with block_idle_frame and replaced it with
frame_pending. block_idle_frame is now unused.

[1]: https://github.com/swaywm/sway/pull/4772


ddevault:
Thanks!

Issue #1874, opened 2019-10-25
Segfault in `drm_connector_set_mode`


KenMacD:
When switching from another virtual console back I get the following segfault:

```
#0  0x00007fd3e6920c49 in drm_connector_set_mode (output=output@entry=0x55a34d3b7d00, wlr_mode=0x0)
    at ../backend/drm/drm.c:652
        conn = 0x55a34d3b7d00
        drm = <optimized out>
        mode = <optimized out>
#1  0x00007fd3e691ec70 in session_signal (listener=0x55a34d1ed7c0, data=<optimized out>) at ../backend/drm/backend.c:96
        plane = <optimized out>
        conn = 0x55a34d3b7d00
        drm = 0x55a34d1ed730
        session = <optimized out>
#2  0x00007fd3e696264e in wlr_signal_emit_safe (signal=signal@entry=0x55a34d1d6408, data=data@entry=0x55a34d1d6400)
    at ../util/signal.c:29
        pos = 0x55a34d1ed7c0
        l = 0x55a34d1ed7c0
        cursor = {link = {prev = 0x55a34d1ed7c0, next = 0x7fffa160d550}, notify = 0x7fd3e69625c0 <handle_noop>}
        end = {link = {prev = 0x7fffa160d530, next = 0x55a34d1d6408}, notify = 0x7fd3e69625c0 <handle_noop>}
#3  0x00007fd3e692f081 in resume_device (msg=<optimized out>, userdata=0x55a34d1d6400, ret_error=<optimized out>)
    at ../backend/session/logind.c:323
        dev = 0x55a34d1dd380
        session = 0x55a34d1d6400
        ret = <optimized out>
        fd = 30
        major = 226
        minor = 0
#4  0x00007fd3e5d9d642 in  () at /usr/lib/libsystemd.so.0
#5  0x00007fd3e5d9d184 in  () at /usr/lib/libsystemd.so.0
#6  0x00007fd3e5d9d15d in  () at /usr/lib/libsystemd.so.0
#7  0x00007fd3e5d9d184 in  () at /usr/lib/libsystemd.so.0
#8  0x00007fd3e5d9d15d in  () at /usr/lib/libsystemd.so.0
#9  0x00007fd3e5d9d184 in  () at /usr/lib/libsystemd.so.0
#10 0x00007fd3e5d9d15d in  () at /usr/lib/libsystemd.so.0
#11 0x00007fd3e5d9d184 in  () at /usr/lib/libsystemd.so.0
[...]

Issue #1868, opened 2019-10-22
Segfault in swap_drm_surface_buffers


emersion:
Happens when docking with more outputs than CRTCs.

```
#0  0x0000000000000000 in  ()
#1  0x00007feb60838863 in swap_drm_surface_buffers (surf=surf@entry=0x55598a395958, damage=damage@entry=0x0) at ../subprojects/wlroots/backend/drm/renderer.c:154
#2  0x00007feb60838904 in get_drm_surface_front (surf=0x55598a395958) at ../subprojects/wlroots/backend/drm/renderer.c:168
        renderer = 0x55598a65a4c0
#3  0x00007feb60834b24 in drm_connector_start_renderer (conn=0x55598a6afa40) at ../subprojects/wlroots/backend/drm/drm.c:513
        drm = 0x55598a36eea0
        crtc = 0x55598a3855f8
        plane = 0x55598a395950
        bo = <optimized out>
        fb_id = <optimized out>
        mode = <optimized out>
#4  0x00007feb60836198 in drm_connector_start_renderer (conn=0x55598a6afa40) at ../subprojects/wlroots/backend/drm/drm.c:600
        conn = 0x55598a6afa40
        drm = <optimized out>
#5  0x00007feb60836198 in drm_connector_set_mode (output=0x55598a6afa40, mode=0x55598aad1190) at ../subprojects/wlroots/backend/drm/drm.c:609
        conn = 0x55598a6afa40
        drm = <optimized out>
#6  0x000055598880ed1b in set_mode (refresh_rate=<optimized out>, height=<optimized out>, width=1680, output=0x55598a6afa40) at ../sway/config/output.c:229
        mhz = <optimized out>
        mode = <optimized out>
        best = <optimized out>
        wlr_output = 0x55598a6afa40
        modeset_success = <optimized out>
        output_box = <optimized out>
#7  0x000055598880ed1b in apply_output_config (oc=oc@entry=0x55598aad0bf0, output=output@entry=0x55598ac360f0) at ../sway/config/output.c:266
        wlr_output = 0x55598a6afa40
        modeset_success = <optimized out>
        output_box = <optimized out>
#8  0x000055598882d5cc in output_enable (output=output@entry=0x55598ac360f0, oc=0x55598aad0bf0) at ../sway/tree/output.c:118
        __PRETTY_FUNCTION__ = "output_enable"
        wlr_output = 0x55598a6afa40
        len = 4
        ws = <optimized out>
[...]

PR #1944, opened 2019-12-05
backend/drm: fix segfault in init_drm_surface

emersion:
When surf->gbm was previously set, we destroy it without setting it to
NULL. Later on, we only create the GBM surface if surf->gbm is NULL.
This result in a use-after-free when we start using surf->gbm.

Closes: https://github.com/swaywm/wlroots/issues/1868
Closes: https://github.com/swaywm/wlroots/issues/1874
Closes: https://github.com/swaywm/sway/issues/4785
Closes: https://github.com/swaywm/sway/issues/4717
Closes: https://github.com/swaywm/sway/issues/4730
Fixes: 2bdd1d0896cc ("backend/drm: use modifiers for our GBM buffers")

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -81,6 +81,7 @@ bool init_drm_surface(struct wlr_drm_surface *surf,
 			surf->back = NULL;
 		}
 		gbm_surface_destroy(surf->gbm);
+		surf->gbm = NULL;
 	}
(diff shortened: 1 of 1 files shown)


ddevault:
Badass. Thanks!

agx:
this looks like a fix for a crash i've also been seeing, great!

Vorsplummi:
I can change the resolution on my secondary monitor without crashing now. Thank you.

Issue #1854, opened 2019-10-14
There is no /dev/ttyv10 on FreeBSD


lantw44:
In file `backend/session/direct-freebsd.c`, function `setup_tty`, it constructs the path to vty with:
```c
snprintf(tty_path, sizeof(tty_path), "/dev/ttyv%d", tty - 1);
```
It fails when `tty - 1` is 10 or greater because there is no /dev/ttyv10 on the system. FreeBSD uses /dev/ttyva and /dev/ttyvb instead.
```
$ ls -1 /dev/ttyv*
/dev/ttyv0
/dev/ttyv1
/dev/ttyv2
/dev/ttyv3
/dev/ttyv4
/dev/ttyv5
/dev/ttyv6
/dev/ttyv7
/dev/ttyv8
/dev/ttyv9
/dev/ttyva
/dev/ttyvb
```
This causes sway to fail with 'Cannot open tty: No such file or directory.'

ascent12:
Patches are welcome. We don't really have a prominent FreeBSD user/maintainer to test things like this.

myfreeweb:
Did you launch a bunch of direct sessions simultaneously? :) I've never really used more than one display server, and I use consolekit anyway so my wayfire is on `/dev/ttyv0` all the time.

This seems trivial, try changing that line to use `%x` instead of `%d`.

lantw44:
> Did you launch a bunch of direct sessions simultaneously? :)

Yes!

/dev/ttyv8 ran GDM (installed from FreeBSD ports) on Xorg.
/dev/ttyv9 ran GNOME (installed from upstream git) on Xorg.

The next vty available for use was /dev/ttyva, but wlroots chose /dev/ttyv10 and failed.

> I've never really used more than one display server, and I use consolekit anyway so my wayfire is on `/dev/ttyv0` all the time.

I know there is an option called `CONSOLEKIT` in x11-toolkits/wlroots port, but I don't know how to use it. I was just finding an easy way to run Wayland natively on FreeBSD. I used to run mutter in nested mode (`mutter --wayland --nested --no-x11`), but I guess applications and toolkits may behave differently if I run them natively.

This bug was found during testing a GTK4 bug: https://gitlab.gnome.org/GNOME/gtk/issues/2201.

> This seems trivial, try changing that line to use `%x` instead of `%d`.

I already did it on my local installation, but I haven't checked how the kernel will pick the name if the number of consoles is greater tha 16.

lantw44:
[...]

PR #1945, opened 2019-12-07
backend/session/freebsd: Fix the way to get TTY path

lantw44:
Previously, the path of TTY is generated using snprintf with %d format.
It works with TTY 1 to 10, but fails with TTY with greater number
because the number used in the name is in base 32 instead of base 10.
Since there is no standard function to convert a number to a string with
a custom base, this commit adds a function to do it.

Fixes: https://github.com/swaywm/wlroots/issues/1854

### DIFF
--- a/backend/session/direct-freebsd.c
+++ b/backend/session/direct-freebsd.c
@@ -139,6 +139,51 @@ static int vt_handler(int signo, void *data) {
 	return 1;
 }
 
+static bool get_tty_path(int tty, char path[static 11], size_t len) {
+	assert(tty > 0);
+
+	const char prefix[] = "/dev/ttyv";
+	static_assert(sizeof(prefix) + 1 <= 11, "TTY path prefix is too long");
+
+	const size_t prefix_len = sizeof(prefix) - 1;
+	strcpy(path, prefix);
+
+	size_t offset = prefix_len;
+	const int num = tty - 1;
+	if (num == 0) {
+		path[offset++] = '0';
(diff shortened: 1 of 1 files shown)


backend/session/direct-freebsd.c
```diff
@@ -139,6 +139,50 @@ static int vt_handler(int signo, void *data) {
...
+	int num = tty - 1;
+	if (num == 0) {
+		path[offset++] = '0';
+		path[offset++] = '\0';
+		return true;
+	}
+
+	const int base = 32;
+	size_t num_len = 0;
+	for (int remaning = num; remaning > 0; remaning /= base, num_len++);
```
emersion: Instead we can just set this to a reasonable value.

backend/session/direct-freebsd.c
```diff
@@ -139,6 +139,50 @@ static int vt_handler(int signo, void *data) {
...
+
+	// Return early if the buffer is too small.
+	if (prefix_len + num_len + 1 > len) {
+		return false;
+	}
+
+	char num_buf[num_len];
```
emersion: No need for this intermediary buffer

backend/session/direct-freebsd.c
lantw44: Fixed by moving the length check into the loop.

backend/session/direct-freebsd.c
lantw44: Fixed by swapping bytes in-place.

PR #1950, opened 2019-12-12
buffer: improve error handling

emersion:
In case the texture can't be imported, release the buffer so that the
client can submit another one. In case the allocation fails, disconnect
the client.


ddevault:
Thanks!

PR #1951, opened 2019-12-15
output: fix wlr_output_preferred_mode fallback

emersion:
`mode` points to an invalid pointer (head of the list) when the loop
stops.

Closes: https://github.com/swaywm/sway/issues/4717


ascent12:
Thanks

Issue #1955, opened 2019-12-20
The performance/responsiveness of EGL apps is broken on the sway with the lastest wlroots


kchibisov:
After 515679e4fed1c7c762a8c269fca09cbb195a4323 some OpenGL/EGL apps are barely usable(e.g. alacritty). Alacritty performance degraded by ~25%(on our benchmarks) by the commit I've mentioned, input became jerky and so on. Firefox on Wayland also seems to be affected. Now it start's like 5-10+ seconds to just open a window and also feels very slow.

cc @emersion ?

emersion:
Your bug report is missing debug logs. Please add a link to the full debug log file.

kchibisov:
From broken commit: [broken_sway.log](https://github.com/swaywm/wlroots/files/3988988/broken_sway.log)

From the one before it: [good_sway.log](https://github.com/swaywm/wlroots/files/3988990/good_sway.log)

GreyXor:
Same here, 
Seen before with @emersion : 

> DMA-BUF import extension not present
> early eglGetString(EGL_EXTENSIONS) doesnt advertise everything

emersion:
Thanks for reporting this issue. Can you try https://github.com/swaywm/wlroots/pull/1956?

PR #1956, opened 2019-12-20
render: fix EGL extensions not loaded

emersion:
Some extensions are only advertised by the EGL implementation with a
non-zero EGLDisplay. That's the case when the extension can only be
enabled when the hardware/driver supports it for instance.

Instead of checking for all extensions without a display, check only for
EGL_EXT_platform_base and EGL_KHR_debug which are used before
eglGetDisplay. Check for all other extensions when we have a display.

Closes: https://github.com/swaywm/wlroots/issues/1955
Closes: https://github.com/swaywm/sway/issues/4828

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -133,6 +133,9 @@ out:
 
 bool wlr_egl_init(struct wlr_egl *egl, EGLenum platform, void *remote_display,
 		EGLint *config_attribs, EGLint visual_id) {
+	// Check for EGL_EXT_platform_base before creating a display, because we
+	// actually use this extension to create displays. Check for EGL_KHR_debug
+	// before creating display to get EGL logs as soon as possible.
 	const char *exts_str = eglQueryString(EGL_NO_DISPLAY, EGL_EXTENSIONS);
 	if (exts_str == NULL) {
 		wlr_log(WLR_ERROR, "Failed to query EGL extensions");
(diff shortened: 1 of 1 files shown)


kchibisov:
Seems like it's working fine. At least the beginning of the debug logs from sway are ~identical now.

log after your patch.
[sway_patched.log](https://github.com/swaywm/wlroots/files/3989073/sway_patched.log)

ddevault:
That arch failure is weird.

emersion:
It's an arch packaging issue:

```
export PKG_CONFIG_DEBUG_SPEW=1
pkg-config x11-xcb
Package xproto was not found in the pkg-config search path.
Perhaps you should add the directory containing `xproto.pc'
to the PKG_CONFIG_PATH environment variable
Package 'xproto', required by 'x11', not found
Package 'kbproto', required by 'x11', not found
```

emersion:
Reported here: https://bugs.archlinux.org/task/64914

JonnyMako:
I can confirm that this branch fixes https://github.com/swaywm/sway/issues/4828

ppascher:
This also fixes vulkan not working since commit 515679e4fed1c7c762a8c269fca09cbb195a4323. Various extensions were not detected, swapchain could not be created (steam, vkcube).

ddevault:
Can you rebase this?

emersion:
Done

ddevault:
Thanks!

PR #1958, opened 2019-12-21
add WLR_USE_XDG_DIR option

sd-hd:
This patch add a option to toggle between XDG_RUNTIME_DIR and "/dev/shm".

If wlroots run inside a strict snap, they didn't have access to "/dev/shm".

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -62,3 +62,4 @@ done like so:
 * *WAYLAND_DISPLAY*, *_WAYLAND_DISPLAY*, *WAYLAND_SOCKET*: if set probe Wayland
   backend in *wlr_backend_autocreate*
 * *XCURSOR_PATH*: directory where xcursors are located
+* *XDG_RUNTIME_DIR*: used to store non persistent temporary files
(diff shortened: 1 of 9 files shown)


emersion:
I'd prefer not to merge this. Instead, I'd suggest you add /dev/shm to a whitelist.

progandy:
> I'd prefer not to merge this. Instead, I'd suggest you add /dev/shm to a whitelist.

Or alternatively snap should have options to bindmount/symlink another directory or mount a tmpfs for that path.

sd-hd:
Mounting a tmpfs of "/dev/shm" didn't work at the current snapd implementation:
`error: cannot validate snap "sway": layout "/dev/shm" in an off-limits area`

A whitelist would be very wlroots-specific and also requires some patches for the snap daemon.

It would be best if you could configure the temporary directory in wlroots. E. g. with Environments Variables.

emersion:
I consider this being a snap bug.

ddevault:
Who even runs a Wayland compositor in a snap in the first place?

sd-hd:
@ddevault 
It's for a kiosk system.

ddevault:
And?

PR #1959, opened 2019-12-21
ci: add xorgproto dep to Arch build

emersion:
Arch Linux maintainers are still figuring out whether they should ask
people to explicitely install it or make it a dependency of xcb again.
In the meantime, add it as an explicit dependency. I'll revert this
patch if they decide otherwise.

[1]: https://bugs.archlinux.org/task/64914


ddevault:
Thanks!

PR #1961, opened 2019-12-22
Revert "ci: add xorgproto dep to Arch build"

emersion:
This reverts commit 35bc3e662a34fe92a3de4e3768dc976f8ac2c242.

Per [1], the dependency has been re-added and we shouldn't need to
explicitly install it anymore.

[1]: https://bugs.archlinux.org/task/64914


ddevault:
Thanks!

Issue #775, opened 2018-03-29
Make it easy to write a custom renderer


emersion:
The only way to create a custom renderer right now is to fork `render/gles`. But IMHO there's an opportunity to share code in a better way between the GLES2 renderer, maybe a future GLES3 renderer and compositor's custom renderers.

- [x] Share code in wlr_egl
- [x] Untie wlr_gles2_renderer and wlr_gles2_texture
- [ ] Make wlr_gles2_texture re-usable in external renderers
- [ ] Make wlr_gles2_renderer re-usable in external renderers

## Share code in wlr_egl

Some EGL code is in the renderer. We can share more by moving some functions (those which create EGLSurfaces for instance) into wlr_egl.

## Make wlr_texture a reusable type

Right now wlr_texture is tied to the wlr_renderer. But it would be possible to untie it from the renderer, making it reusable in another renderer. For a custom GLES2 renderer (e.g. a 3D one) this would be helpful.

## Export wlr_gles2_* functions

This will allow custom renderers to choose what to re-use, and to build a wlr_gles2_renderer wrapper.

emersion:
This proposal is outdated, closing.

PR #1962, opened 2019-12-22
add destroy listener for textures

### LINKED ISSUE #1949: Use-after-free when destroying textures after renderer
```
==67892==ERROR: AddressSanitizer: heap-use-after-free on address 0x60b000006620 at pc 0x7f553256f1b5 bp 0x7ffcbc1ef9c0 sp 0x7ffcbc1ef9b0
READ of size 8 at 0x60b000006620 thread T0
    #0 0x7f553256f1b4 in wlr_egl_is_current ../subprojects/wlroots/render/egl.c:347
    #1 0x7f553257eab2 in gles2_texture_destroy ../subprojects/wlroots/render/gles2/texture.c:122
    #2 0x7f5532583b49 in wlr_texture_destroy ../subprojects/wlroots/render/wlr_texture.c:16
    #3 0x7f55325fcc96 in wlr_buffer_unref ../subprojects/wlroots/types/wlr_buffer.c:139
    #4 0x561fd5c1be7d in buffer_handle_release ../wlr_buffer.c:19
    #5 0x561fd5c00d7e in wl_signal_emit /usr/include/wayland-server-core.h:472
    #6 0x561fd5c02065 in glider_buffer_unlock ../allocator.c:88
    #7 0x561fd5c0ad4c in unlock_drm_buffer ../backend/connector.c:458
    #8 0x561fd5c11479 in destroy_drm_buffer ../backend/device.c:356
    #9 0x561fd5c0e12a in finish_drm_device ../backend/device.c:190
    #10 0x561fd5c0279a in backend_destroy ../backend/backend.c:33
    #11 0x561fd5c02927 in handle_session_destroy ../backend/backend.c:52
    #12 0x7f5532697e37 in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #13 0x7f55325227b3 in wlr_session_destroy ../subprojects/wlroots/backend/session/session.c:149
    #14 0x7f55325218f7 in handle_display_destroy ../subprojects/wlroots/backend/session/session.c:65
    #15 0x7f55320dbcde  (/lib64/libwayland-server.so.0+0x8cde)
    #16 0x7f55320dc466 in wl_display_destroy (/lib64/libwayland-server.so.0+0x9466)
    #17 0x561fd5c1648d in main ../main.c:114
    #18 0x7f55315ca152 in __libc_start_main (/lib64/libc.so.6+0x27152)
    #19 0x561fd5c00aed in _start (/home/simon/src/glider/build/glider+0x34aed)

0x60b000006620 is located 32 bytes inside of 112-byte region [0x60b000006600,0x60b000006670)
freed by thread T0 here:
    #0 0x7f5532ca06b0 in __interceptor_free /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:122
    #1 0x561fd5c194c8 in glider_renderer_destroy ../renderer.c:55
    #2 0x561fd5c16427 in main ../main.c:112
    #3 0x7f55315ca152 in __libc_start_main (/lib64/libc.so.6+0x27152)

    #0 0x7f5532ca0cd8 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x561fd5c18d0a in glider_gbm_renderer_create ../renderer.c:10
    #2 0x561fd5c15cf6 in main ../main.c:69
    #3 0x7f55315ca152 in __libc_start_main (/lib64/libc.so.6+0x27152)

SUMMARY: AddressSanitizer: heap-use-after-free ../subprojects/wlroots/render/egl.c:347 in wlr_egl_is_current
  0x0c167fff8c70: 00 00 00 00 00 00 fa fa fa fa fa fa fa fa 00 00
  0x0c167fff8c80: 00 00 00 00 00 00 00 00 00 00 00 00 fa fa fa fa
  0x0c167fff8c90: fa fa fa fa 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c167fff8ca0: 00 00 fa fa fa fa fa fa fa fa 00 00 00 00 00 00
  0x0c167fff8cb0: 00 00 00 00 00 00 00 05 fa fa fa fa fa fa fa fa
=>0x0c167fff8cc0: fd fd fd fd[fd]fd fd fd fd fd fd fd fd fd fa fa
  0x0c167fff8cd0: fa fa fa fa fa fa fd fd fd fd fd fd fd fd fd fd
  0x0c167fff8ce0: fd fd fd fd fa fa fa fa fa fa fa fa 00 00 00 00
  0x0c167fff8cf0: 00 00 00 00 00 00 00 00 01 fa fa fa fa fa fa fa
  0x0c167fff8d00: fa fa 00 00 00 00 00 00 00 00 00 00 00 00 00 02
  0x0c167fff8d10: fa fa fa fa fa fa fa fa 00 00 00 00 00 00 00 00
  Addressable:           00
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==67892==ABORTING
```

Emantor:
Add a destroy listener for textures, which destroys the gles2 textures
when the renderer is destroyed.

Fixes #1949

----

I am not entirely sure about the API changes, since the `wlr_renderer` is now passed directly to the texture creation functions. I am all ears for possible improvements.

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -97,6 +97,8 @@ const enum wl_shm_format *get_gles2_wl_formats(size_t *len);
 
 struct wlr_gles2_texture *gles2_get_texture(
 	struct wlr_texture *wlr_texture);
+struct wlr_gles2_renderer *gles2_get_renderer(
+		struct wlr_renderer *wlr_renderer);
 
 void push_gles2_marker(const char *file, const char *func);
 void pop_gles2_marker(void);
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -100,6 +100,14 @@ struct wlr_gles2_texture *gles2_get_texture(
 struct wlr_gles2_renderer *gles2_get_renderer(
 		struct wlr_renderer *wlr_renderer);
 
+struct wlr_texture *gles2_texture_from_pixels(struct wlr_renderer *renderer,
+	enum wl_shm_format wl_fmt, uint32_t stride, uint32_t width, uint32_t height,
+	const void *data);
+struct wlr_texture *gles2_texture_from_wl_drm(struct wlr_renderer *renderer,
+	struct wl_resource *data);
+struct wlr_texture *gles2_texture_from_dmabuf(struct wlr_renderer *renderer,
+	struct wlr_dmabuf_attributes *attribs);
+
 void push_gles2_marker(const char *file, const char *func);
 void pop_gles2_marker(void);
 #define PUSH_GLES2_DEBUG push_gles2_marker(_WLR_FILENAME, __func__)
--- a/include/wlr/render/wlr_texture.h
+++ b/include/wlr/render/wlr_texture.h
@@ -18,6 +18,8 @@ struct wlr_texture_impl;
 
 struct wlr_texture {
 	const struct wlr_texture_impl *impl;
+
+	struct wl_listener renderer_destroy;
 };
 
 /**
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -3,7 +3,7 @@
 #include <wlr/interfaces/wlr_input_device.h>
 #include <wlr/interfaces/wlr_output.h>
 #include <wlr/render/egl.h>
-#include <wlr/render/gles2.h>
+#include <wlr/render/wlr_renderer.h>
 #include <wlr/util/log.h>
 #include "backend/headless.h"
 #include "util/signal.h"
--- a/examples/multi-pointer.c
+++ b/examples/multi-pointer.c
@@ -10,7 +10,6 @@
 #include <wayland-server-core.h>
 #include <wlr/backend.h>
 #include <wlr/backend/session.h>
-#include <wlr/render/gles2.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_keyboard.h>
--- a/examples/pointer.c
+++ b/examples/pointer.c
@@ -9,7 +9,6 @@
 #include <wayland-server-core.h>
 #include <wlr/backend.h>
 #include <wlr/backend/session.h>
-#include <wlr/render/gles2.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_keyboard.h>
--- a/include/wlr/render/gles2.h
+++ b/include/wlr/render/gles2.h
@@ -19,14 +19,6 @@ struct wlr_renderer *wlr_gles2_renderer_create(struct wlr_egl *egl);
 
 struct wlr_egl *wlr_gles2_renderer_get_egl(struct wlr_renderer *renderer);
 
-struct wlr_texture *wlr_gles2_texture_from_pixels(struct wlr_renderer *renderer,
-	enum wl_shm_format wl_fmt, uint32_t stride, uint32_t width, uint32_t height,
-	const void *data);
-struct wlr_texture *wlr_gles2_texture_from_wl_drm(struct wlr_renderer *renderer,
-	struct wl_resource *data);
-struct wlr_texture *wlr_gles2_texture_from_dmabuf(struct wlr_renderer *renderer,
-	struct wlr_dmabuf_attributes *attribs);
-
 struct wlr_gles2_texture_attribs {
 	GLenum target; /* either GL_TEXTURE_2D or GL_TEXTURE_EXTERNAL_OES */
 	GLuint tex;
--- a/include/wlr/render/gles2.h
+++ b/include/wlr/render/gles2.h
@@ -19,12 +19,12 @@ struct wlr_renderer *wlr_gles2_renderer_create(struct wlr_egl *egl);
 
 struct wlr_egl *wlr_gles2_renderer_get_egl(struct wlr_renderer *renderer);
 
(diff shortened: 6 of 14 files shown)


emersion:
The intent with the current API was to allow compositors to create `wlr_texture`s without having to use `wlr_renderer`. However it doesn't seem like this is used by anyone (@ammen99?), since `wlr_surface` requires `wlr_renderer` anyway.

emersion:
See https://github.com/swaywm/wlroots/issues/775

An alternative would be to add a `destroy` signal to `wlr_egl`.

Emantor:
Pushed a new version which moves the destruction function to `wlr_texture` instead of being gles2 specific. I can also separate out the gles2 texture creation function into its own commit if required.

emersion:
This is still GLES2 specific, as the renderer passed to the texture creation functions take a `wlr_renderer` which is asserted to be the GLES2 renderer.

Emantor:
Yes, since there is no other renderer. But the listener is now registered for all `wlr_textures` in the init function. The texture creation functions obv. need to assert since they are gles2 specific. Only the texture destruction part should now work for all `wlr_texture` parts, if we had more than gles2.

ammen99:
> However it doesn't seem like this is used by anyone (@ammen99?), since `wlr_surface` requires `wlr_renderer` anyway.

I am not using any of the functions which are touched here, so I am not against this change. Renderers are supposed to be redesigned sometime in the future anyway IIRC.

emersion:
Maybe we should just hide these functions from the public API for now.

Emantor:
> Maybe we should just hide these functions from the public API for now.

Should this be done in this PR? Is there anything else blocking this PR?

emersion:
Yeah, since this PR contains breaking changes, can we make these functions private?

Emantor:
Done, I also cleaned up some `gles2.h` imports.

emersion:
Are `wlr_texture` users prepared for the texture to be implicitly destroyed?

include/render/gles2.h
```diff
@@ -97,6 +97,16 @@ const enum wl_shm_format *get_gles2_wl_formats(size_t *len);
 
 struct wlr_gles2_texture *gles2_get_texture(
 	struct wlr_texture *wlr_texture);
+struct wlr_gles2_renderer *gles2_get_renderer(
+		struct wlr_renderer *wlr_renderer);
+
+struct wlr_texture *wlr_gles2_texture_from_pixels(struct wlr_renderer *renderer,
```
emersion: Private functions need to have their `wlr_` prefix stripped, otherwise they will be visible in the resulting dynamic library.

Emantor: done.

Emantor:
> Are `wlr_texture` users prepared for the texture to be implicitly destroyed?

Not sure how to verify this short of verifying all callers, I have been running this PR on top of master and tested output hotplug so far.

emersion:
Grepping for `wlr_texture`:

- `backend/drm/renderer.c`: since the backend has ownership over the renderer it should be fine
- `types/wlr_output.c`: used for cursors, would be safer to add a listener
- `types/wlr_buffer.c`: should add a listener

Emantor:
> Grepping for `wlr_texture`:
> 
>     * `backend/drm/renderer.c`: since the backend has ownership over the renderer it should be fine
> 
>     * `types/wlr_output.c`: used for cursors, would be safer to add a listener

Looks straightforward.

>     * `types/wlr_buffer.c`: should add a listener

Hm, how do I handle the reference counting? Just ignore, send a release for the buffer and subsequently destroy?

emersion:
>Hm, how do I handle the reference counting? Just ignore, send a release for the buffer and subsequently destroy?

We shouldn't release the buffer. We could set `texture` to NULL instead (this would be a breaking change I think? However most callers should already be prepared for it via `wlr_surface_get_texture`).

Emantor:
This should be ready for review.

include/wlr/render/wlr_texture.h
```diff
@@ -18,6 +18,8 @@ struct wlr_texture_impl;
 
 struct wlr_texture {
 	const struct wlr_texture_impl *impl;
+
+	struct wl_listener renderer_destroy;
```
emersion, line 22: This listener may run before or after other renderer destroy listeners (listener call order is undefined). This can cause a use-after-free if the texture listener is called first, then the buffer/output listener is called.

We should probably add a `destroy` signal to `wlr_texture` to prevent this.

Emantor:
Will be superseded by usage of `wlr_buffer`

emersion:
Hm, I don't think so. `wlr_texture` will remain as the abstraction for pixels ready to be textured from by the renderer.

Emantor:
> Hm, I don't think so. `wlr_texture` will remain as the abstraction for pixels ready to be textured from by the renderer.

Yep, but I remember the last place where this was lacking to be the cursor code having no nice place to hook into. And with your reworks in flight currently (and this PR not having been rebased in a while), I'll see how things turn out in master. I intend to pick this up again, but it will be a while, so closing this for now.

emersion:
I see! No problem, thanks for working on this :)

PR #1964, opened 2019-12-24
xdg-shell: fix inert xdg_surface handling

emersion:
Closes: https://github.com/swaywm/sway/issues/4834
Closes: https://github.com/swaywm/wlroots/issues/1890

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -244,6 +244,11 @@ struct wlr_xdg_toplevel_show_window_menu_event {
 
 struct wlr_xdg_shell *wlr_xdg_shell_create(struct wl_display *display);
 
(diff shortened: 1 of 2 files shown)


ammen99:
It is worth mentioning that this patch makes all applications that use `gtk-layer-shell` crash as soon as I try to open a popup. This happens in both Sway and Wayfire, with gtk-layer-demo, mate-panel and wf-panel. Popups of native GTK apps work.

ammen99:
Oh, and without these patches, with mate-panel's popups, I can trigger an assertion inside wlroots (on Wayfire). So there is something fishy here.

emersion:
Ah, indeed, I was rejecting a NULL `parent` arg in the `get_popup` handler, which layer-shell relies on. Can you try this updated version?

ammen99:
@emersion Yes, it seems to work now.

ddevault:
Thanks!

PR #1966, opened 2019-12-25
[WIP] Introduce new scene-graph API

emersion:
A new wlr_scene API has been added, following the design ideas from [1].
The new API contains the minimal set of features required to make the
API useful. The goal is to design a solid fundation and add more
features in the future.

[1]: https://github.com/swaywm/wlroots/issues/1826#issuecomment-564601757


emersion:
This is now ready for review. A compositor using wlr_scene is available here: https://lists.sr.ht/~sircmpwn/sedna/patches/9403

ddevault:
Initial impressions from testing with sedna:

Looks like there are sometimes issues with double buffering the main display framebuffer on the Wayland backend, it can start flickering under certain conditions. One way I was able to trigger it was by opening gnome-calculator, then running mpv on a largeish (1080p) video, then closing mpv.

types/wlr_scene.c
```diff
@@ -0,0 +1,309 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <wlr/backend.h>
+#include <wlr/render/wlr_renderer.h>
+#include <wlr/types/wlr_matrix.h>
+#include <wlr/types/wlr_scene.h>
+#include <wlr/types/wlr_surface.h>
+#include "util/signal.h"
+
+static struct wlr_scene *scene_node_get_root(struct wlr_scene_node *node) {
```
ddevault: nit: this reads like it'd walk up the node tree until it found the root node. Perhaps better to call it `scene_from_root_node` or something.

ddevault:
On the whole this looks fine, but I'm not convinced that it demonstrates much value yet. How would this translate into plane allocation? Is there a roadmap to libliftoff integration? Are we sure that if we ship this API it'll be reasonably easy to carry into this future?

emersion:
>Looks like there are sometimes issues with double buffering the main display framebuffer on the Wayland backend, it can start flickering under certain conditions. One way I was able to trigger it was by opening gnome-calculator, then running mpv on a largeish (1080p) video, then closing mpv.

Weird, there's no damage tracking going on and the rendering logic is the same. I'll try to reproduce and fix it.

>On the whole this looks fine, but I'm not convinced that it demonstrates much value yet.

Yes, this is intended. I want to break down the work in separate pull requests.

> How would this translate into plane allocation? Is there a roadmap to libliftoff integration?

This should be pretty straightforward. See glider for an example of how libliftoff could be wired up:

* Attaching the client's buffer to libliftoff: https://github.com/emersion/glider/blob/b9c19866d6f80c700b43e8a331d8e9ba34495dac/xdg_shell.c#L49
* Rendering if the buffer didn't make it into a KMS plane: https://github.com/emersion/glider/blob/b9c19866d6f80c700b43e8a331d8e9ba34495dac/output.c#L63

However, the plan is to have abstract libliftoff behind a wlroots API so that all backends can implement such optimizations. This would be orthogonal to `wlr_scene`: compositor can choose not to use `wlr_scene` if they want to use planes.

> Are we sure that if we ship this API it'll be reasonably easy to carry into this future?

I'd be happy to keep this PR unmerged until I can demonstrate more features.

emersion:
>Looks like there are sometimes issues with double buffering the main display framebuffer on the Wayland backend, it can start flickering under certain conditions. One way I was able to trigger it was by opening gnome-calculator, then running mpv on a largeish (1080p) video, then closing mpv.

Caused by the renderer scissor not properly unset when done. Should be fixed now.

ddevault:
Confirmed fixed.

emersion:
The double-buffering has been removed for now since there was no use-case for it.

I've written a Cage patch which uses this PR: https://github.com/Hjdskes/cage/pull/197

emersion:
This is pretty much ready for review.

Some open questions, not necessarily blocking:

- Should `wlr_scene_surface` handle subsurfaces automatically? Right now this is the case, but compositors might want to customize their rendering (?).
- Should `wlr_scene_render_output` take a `wlr_renderer` as argument? This would allow compositors to use a custom renderer instead of the backend's. Should we add a function to render the scene which doesn't take a `wlr_output`?
- Should the node coords use `double` values instead of `int`? This would allow nodes to be placed in-between two logical pixels (but still aligned to a physical pixel) on HiDPI outputs. Maybe desirable for drag icons and cursors. There's an impedence mismatch with wlroots' surface iterators.

djpohly:
A couple of naming thoughts from a random compositor author:
- `WLR_SCENE_NODE_ROOT`: if the intent is for these to be nestable, would "container" be a better name than "root" for the node type?  It could make sense to use "root" as the name of the field inside `wlr_scene` instead.
- `wlr_scene_node_toggle()`: the name "toggle" sounds like it only switches from enabled to disabled or vice versa - maybe `set_enabled` or `set_visible` is a more intuitive name?

emersion:
Thanks for the feedback!

> WLR_SCENE_NODE_ROOT: if the intent is for these to be nestable, would "container" be a better name than "root" for the node type? It could make sense to use "root" as the name of the field inside wlr_scene instead.

Right, the intent is for ROOT to not be nestable. The [original issue](https://github.com/swaywm/wlroots/issues/1826#issuecomment-564601757) has a separate TREE type for this. Maybe ROOT can just be a TREE, not sure yet. I guess it depends how outputs are integrated in the design.

> wlr_scene_node_toggle(): the name "toggle" sounds like it only switches from enabled to disabled or vice versa - maybe set_enabled or set_visible is a more intuitive name?

Hm, I've chosen toggle() because I wanted a verb for consistency with move(). `wlr_output` has a enable() function that takes a bool but that's even more confusing. Maybe set_position()/set_enabled() would be better.

djpohly:
Out of curiosity sparked by some of the discussion here and in #1826, I hacked up [a quick proof-of-concept](/djpohly/wlroots/tree/scenegraph-rect) to see how much/little change it would take from the current PR to accomplish two things:

1. Implementing a RECT node type (solid-colored rectangle for window borders or such) as an example of a node which renders something that isn't a wlr_surface. This required render_output() to iterate over nodes rather than surfaces, but after that refactoring, it was easy to add.
2. Allowing ROOT to be used as a nestable container node.  This required very little change, since all the necessary functionality was already implemented and needed only to be exposed.

Both are demonstrated in the scene-graph example by placing a small bar atop each stacked surface and grouping the two together with a nested ROOT-type node.  It's not visually pretty or anything, but it does work.

To be clear, I don't mean to propose merging these changes or even implementing these specific features necessarily. I just thought I'd share it in case there are any ideas here that prove helpful as the API continues to develop.

sdilts:
> Should wlr_scene_surface handle subsurfaces automatically?

As a potential user of this API, I would say yes. I would prefer the "easy mode" to remain as easy as possible, so unless there is a no-pain way to allow custom implementations, I would think it would be better to keep it simple. A cursory look shows that this only uses public APIs, so it should be possible for someone to pull this code out into their own project and modify it to suit their needs.

kennylevinsen:
> As a potential user of this API, I would say yes. I would prefer the "easy mode" to remain as easy as possible, so unless there is a no-pain way to allow custom implementations, I would think it would be better to keep it simple.

While I am not sure which way we should go, this would be my exact argument for "no".

If we are not sure if the vast majority of our compositors want the logic this way, especially if it is not easy to sidestep this logic, then the best option would normally be to keep it simple and *not* add the logic at all. Especially if it's very little work for downstream, and if it's easy to add in later.

emersion:
> Out of curiosity sparked by some of the discussion here and in #1826, I hacked up a quick proof-of-concept

Very nice! I'd be interested in these patches to be submitted as pull requests once the basic scene-graph lands.

> I would prefer the "easy mode" to remain as easy as possible, so unless there is a no-pain way to allow custom implementations, I would think it would be better to keep it simple.

I'd agree with @kennylevinsen here, especially if we start adding more stuff later like damage tracking. Subsurfaces effectively create sub-trees in the scene-graph. It wouldn't be too hard to add a small optional `wlr_scene_surface_create_with_subsurfaces` helper which automatically creates surface nodes for all subsurfaces.

emersion:
I'm also wondering how to plumb things like `wlr_output_damage` and `wlr_output_layout`. Should the `wlr_scene` take a `wlr_output_damage`/`wlr_output_layout` as an optional argument, or should it be the other way around? Or something else entirely?

emersion:
Dropped `wlr_scene_commit_output`, because it doesn't allow the scen-graph to have the necessary per-output state tracking structs for things like damage tracking.

emersion:
I've written https://github.com/swaywm/wlroots/pull/3117 to see how damage tracking could be handled. Seems to work pretty well. Some notes:

- I'm increasingly tempted to drop sub-surfaces from `wlr_scene_surface` and turn sub-surface handling in a helper. Having exactly 1 surface per `wlr_scene_surface` simplifies things.
- I'm considering making new child nodes disabled by default. This allows callers to setup the node's position (potentially other things in the future like children) and *then* enable it when they're done. That way, we don't unnecessarily accumulate too much damage.

types/wlr_scene.c
```diff
@@ -0,0 +1,306 @@
...
+	if (!node->state.enabled) {
+		return NULL;
+	}
+
+	// TODO: optimize by storing a bounding box in each node?
+	lx -= node->state.x;
+	ly -= node->state.y;
+
+	struct wlr_scene_node *child;
+	wl_list_for_each(child, &node->state.children, state.link) {
```
djpohly, line 173: Should this be `wl_list_for_each_reverse`? It looks like child nodes are stored in rendering order (bottom to top), and though the documentation comment doesn't specify, I think it's natural to assume this function would return the topmost matching surface.

emersion: Good catch! Fixed.

emersion:
I've removed sub-surface support from this PR, and moved it to https://github.com/swaywm/wlroots/pull/3128 where explicit scene-graph nodes are created.

Issue #1099, opened 2018-06-29
Only send current mode via wl_output


emersion:
Modes were a protocol design mistake. If a client uses this information, it's probably broken.

GNOME does this.

SirCmpwn:
I don't see any reason *not* to send modes. It's easy and in the protocol, and it's not dangerous if the client knows what other modes are available. I find it useful to enumerate modes with weston-info at times, too.

emersion:
Will discuss this with the rest of the Wayland community. If possible, I'd like those to get deprecated.

emersion:
I'd like to bump this again for two reasons:

- Modes as exposed in the core protocol have many issues: https://gitlab.freedesktop.org/wayland/wayland/issues/92
- We now support a specialized protocol for output management, which exposes modes in a better way.

PR #1968, opened 2019-12-25
output: only advertise current mode

emersion:
- Regular clients shouldn't care about modes
- Modes exposed are missing metadata such as aspect-ratio, interleaved, etc
- Modes exposed cannot be pruned [1]
- wlr-output-management provides a better API for privileged clients

[1]: https://gitlab.freedesktop.org/wayland/wayland/issues/92

Closes: https://github.com/swaywm/wlroots/issues/1099


ddevault:
Thanks!

Issue #1623, opened 2019-03-18
Don't expose make/model via wl_output


emersion:
Similar to https://github.com/swaywm/wlroots/issues/1610, but for `wl_output.{make,model}`.

* These two are [legacy `wl_output` properties](https://gitlab.freedesktop.org/wayland/wayland/blob/master/protocol/wayland.xml#L2415)
* `xdg-output.{name,description}` are much better
* It's not clear what make/model/serial is for a virtual output
* Clients shouldn't rely on it

ascent12:
These properties are somewhat benign compared to the layout position.
But I don't really feel strongly either way.

emersion:
Well some people are thinking about using those with the wlr-output-management protocol to get the make/model.

agx:
In order to superseed these we should use a format in xdg-output.description that allow to extract that information.

agx:
>>> Why would this be needed?

The current description (e.g. `Dell Inc. DELL U2717D J0XYN71G742L (HDMI-A-1)`) is a bit too verbose to present to the user so being able to spit out the parts like vendor & model is nice (and gradually displaying more information in case of identical vendor & model so the user can tell them apart.

emersion:
xdg-output.description will be used by other clients displaying information about outputs (e.g. screen sharing apps, screen shooters, etc). So instead of having workarounds to re-format the title in the shell, I think it would be better to have wlroots or the compositor produce higher quality xdg-output.description.

agx:
What workarounds do you refer to?

agx:
`wl_output` hat distinct fields to extract information, `xdg-output.descripton` is free form text. In some situations you would have the above look like:

`Dell Inc. DELL U2717D` 

while in others you might want the user to see

`Dell Inc. DELL U2717D on HDMI-A-1`

but I don't see how the compositor could tell what is needed so end users be able to make sense of the format. Since `xdg-output` is unlikely to grow structured data about the output I see these alternatives:

1.) make wlr-output-management provide that information
2.)  keep the information in wl_output
3.) Cram everything into `xdg-output.description` in a paresable form
4.) As a band aid  parse xdg-output.description client side as is

1.) would likely be best and it would even match the protocol's intent. We could  Does this make sense?

emersion:
>wl_output hat distinct fields to extract information

These are legacy. wl_output shouldn't have had those fields.

>In some situations

What are the situations you'd want to have the first and what are the situations in which you'd want to have the second form?

[...]

PR #1969, opened 2019-12-26
output: add description

emersion:
wlr_output.description is a string containing a human-readable string
identifying the output. Compositors can customise it via
wlr_output_set_description, for instance to make the name more
user-friendly.
    
References: https://github.com/swaywm/wlroots/issues/1623

cc @agx


RedSoxFan:
Thanks!

PR #1970, opened 2019-12-28
render: guard rendering operations between begin() and end()

emersion:
Add a wlr_renderer.rendering bool, set it to true between
wlr_renderer_begin() and wlr_renderer_end(). Assert we're rendering when
calling functions that render.

### DIFF
--- a/include/wlr/render/wlr_renderer.h
+++ b/include/wlr/render/wlr_renderer.h
@@ -25,6 +25,8 @@ struct wlr_drm_format_set;
 struct wlr_renderer {
 	const struct wlr_renderer_impl *impl;
 
+	bool rendering;
+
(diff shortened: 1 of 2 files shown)


ammen99:
Shouldn't read_pixels also be guarded by begin/end?

emersion:
`read_pixels` can happen after `renderer_end` actually, as long as the buffers haven't been swapped. This is used by the screencopy implementation actually: it calls `read_pixels` right before the buffer swap.

Currently, I don't have a good idea to safe-guard `read_pixels` calls, because it would kind of need the renderer to know about the output that's currently being rendered, or the renderer to know when buffers are swapped.

ddevault:
Thanks!

PR #1971, opened 2019-12-29
backend/drm: don't modeset with a NULL mode after TTY switch

emersion:
This fixes a segfault in drm_connector_set_mode (mode = NULL).

This happens because we set wlr_output.enabled to true if the connector
is attached to the CRTC. When the user disables an output in the
wlroots-based compositor, switches to another VT (enabling the output),
then switches back, wlroots sets wlr_output.enabled to true but
wlr_output.current_mode is NULL.

We should consider not reading properties from KMS after a TTY switch,
disabling all connectors. However this may result in flickering (outputs
being disabled then re-enabled).

Closes: https://github.com/swaywm/wlroots/issues/1874


ddevault:
Thanks!

PR #1973, opened 2019-12-29
wlr_keyboard_key_event: add group_update property

RedSoxFan:
When a keyboard is added to or removed from a wlr_keyboard_group, a key
event will be emitted for all keys that only pressed for that keyboard
in the group. This allows for compositors to update their internal
state. However because these keys were already down, only the state
should be updated. Compositors should not execute bindings and surfaces
should not receive the key event

### DIFF
--- a/backend/wayland/wl_seat.c
+++ b/backend/wayland/wl_seat.c
@@ -215,6 +215,7 @@ static void keyboard_handle_enter(void *data, struct wl_keyboard *wl_keyboard,
 			.state = WLR_KEY_PRESSED,
 			.time_msec = time,
 			.update_state = false,
+			.group_update = false,
 		};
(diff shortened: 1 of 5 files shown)


emersion:
Not a fan of this solution: the keyboard group logic bleeds in the common code, and compositors need to add special-cases.

emersion:
Can't we just ignore keypresses during Sway reload?

RedSoxFan:
> Not a fan of this solution

Would you prefer removing the static modifier from the wlr_keyboard's state handling functions and adding a signal to wlr_keyboard_group that the compositor should listen on? I think that solution results in duplicated code.

> Can't we just ignore keypresses during Sway reload?

This isn't restricted to reload. Changing the keymap or repeat info in sway for a keyboard using a binding could result in the same issue.

PR #1977, opened 2020-01-02
backend/wayland: fix frame callback not registered

emersion:
This got removed in [1]. I probably messed up the rebase.

[1]: https://github.com/swaywm/wlroots/pull/1797/files#diff-3065f86e6de87d143d4a7673a8ee3a2d

Fixes: 5d1ba0f44687 ("output: re-introduce atomic mode, enabled, scale and transform")


ddevault:
Thanks!

PR #1978, opened 2020-01-03
virtual-pointer: Actually use the value passed to axis_discrete

any1:
It turns out that scrolling doesn't work unless this value is set somewhere.


ddevault:
Thanks!

PR #1979, opened 2020-01-05
Allow focus change

soreau:
This fixes clicks on menu items in steam client.


emersion:
You can just copy-pasta the XWM instead of maintaining a fork.

Issue #1645, opened 2019-04-02
Steam under XWayland is not interpreting mouse clicks correctly


HoshiYamazaki:
Two monitors, one FullHD, second with scale 1.75, Steam from Arch repo, newest packages.

HoshiYamazaki:
So the problem is, when I get dropdown menu for example in library and try to click something, it doesn't work (for example Utilities - I wanted to install SteamVR to test WMR under OpenHMD).

RedSoxFan:
Related swaywm/sway#1865

emersion:
I actually think it's a duplicate.

PR #1980, opened 2020-01-05
xwm: allow applications to change focus between their own surfaces

ammen99:
Although currently this problem is present in only Steam, and it is actually a client bug.

Relevant discussion:  https://github.com/swaywm/sway/issues/1865 and #1645 

What I am not sure: should we update `xwm->focus_surface` ?


soreau:
This can have my Tested-by. The steam exit menu button closes the client as expected with this patch.

emersion:
>What I am not sure: should we update xwm->focus_surface ?

Yeah, I think so.

ammen99:
> Yeah, I think so.

Alright, done.

emersion:
Thanks!

PR #1982, opened 2020-01-08
Allow WLR_RDP_PORT to be any valid TCP/UDP port number

### LINKED ISSUE #1883: WLR_RDP_PORT defaults to 3389, but user values are clamped to 0<n<1024...?
https://github.com/swaywm/wlroots/blob/ca45f4490ccce64bf7aa0985951319646b55d258/backend/backend.c#L161

I don't understand why `WLR_RDP_PORT` is restricted to <1024 when it defaults to 3389, and this restriction makes it harder to run Sway RDP as a regular user.

ammgws:
Fixes #1883

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -157,9 +157,9 @@ static struct wlr_backend *attempt_rdp_backend(struct wl_display *display,
 	if (_port) {
 		char *endptr;
 		int port = strtol(_port, &endptr, 10);
-		if (*endptr || port <= 0 || port >= 1024) {
+		if (*endptr || port <= 0 || port > 65535) {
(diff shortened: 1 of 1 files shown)


backend/backend.c
```diff
@@ -157,9 +157,9 @@ static struct wlr_backend *attempt_rdp_backend(struct wl_display *display,
 	if (_port) {
 		char *endptr;
 		int port = strtol(_port, &endptr, 10);
-		if (*endptr || port <= 0 || port >= 1024) {
+		if (*endptr || port <= 0 || port >= 65535) {
```
emersion: 65535 is a valid port

ammgws: Indeed.

PR #1983, opened 2020-01-09
backend/wayland: handle display errors more gracefully

emersion:
Previously, an error on the remote Wayland display would result in an
infinite loop priting:

    2020-01-09 13:39:03 - [wayland] Source dispatch function returned negative value!
    2020-01-09 13:39:03 - [wayland] This would previously accidentally suppress a follow-up dispatch

This happens when the remote compositor disconnects the client because
of a protocol error, for instance.

Handle wl_display_dispatch and wl_display_dispatch_pending returning -1
by terminating the local display and printing an error.


ddevault:
Thanks!

PR #1984, opened 2020-01-09
backend/wayland: listen to wl_buffer.release events

emersion:
Previously, we just assumed submitting a new frame would make the
compositor release the current one. This isn't always the case, for
instance Sway retains old buffers when a transaction is pending. This
resulted in synchronization issues with clients writing in
front-buffers.

Fix this by un-referencing a wlr_buffer when the parent compositor sends
wl_buffer.release.

Tested by running a fullscreen mpv instance in Sway with the Wayland
backend.


ddevault:
Thanks!

PR #1985, opened 2020-01-09
Introduce wlr_output_layer

emersion:
This series add a new `wlr_output_layer` API, an implementation of this API in the Wayland backend and an example compositor.

My plan is to make [glider](https://github.com/emersion/glider) switch to this API to demonstrate how it can be implemented under DRM.

## Introduce wlr_output_layer

This new API allows compositors to display buffers without needing to
perform rendering operations. This API can be implemented on Wayland
using subsurfaces and on DRM using KMS planes.
    
The goal is to make use of this API in a future scene-graph API.

## backend/wayland: implement the output layer API

The output layer API is implemented using subsurfaces. I chose to
implement this API in the Wayland backend before doing so in the DRM
backend, because it's way easier on Wayland. On DRM, one needs to figure
out how buffers can be mapped to KMS planes (libliftoff can help) and
perform atomic test-only commits (our current DRM backend isn't ready
for this).

## examples: add output-layers example

This new example demonstrates how to use the wlr_output_layer API. It's
a compositor that displays all client surfaces using wlr_output_layer.
To test, one can for instance run:
    
    build/examples/output-layers -s 'weston-simple-dmabuf-egl & weston-simple-egl'
    
Under the Wayland backend (where layers work as long as clients use
DMA-BUFs), thsi should display two surfaces.


ddevault:
It seems like this is designed so that `wlr_output_layer` is entirely siloed and responsible for its own state. This is going to lead to copying down the line. Why not make it an interface that each backend can implement? DRM with libliftoff, for example, which looks very similar. DRY up the sources of truth. Thoughts?

emersion:
>It seems like this is designed so that wlr_output_layer is entirely siloed and responsible for its own state. This is going to lead to copying down the line. Why not make it an interface that each backend can implement? DRM with libliftoff, for example, which looks very similar. DRY up the sources of truth. Thoughts?

I agree state duplication is not great, but there are several issues with this approach:

- A need a guarantee that nothing happens before `wlr_output_commit`. In particular, it should always be possible to cancel all pending changes with `wlr_output_rollback` (in case the commit fails, for instance). Rolling back is important because `wlr_output_commit` doesn't explain why a commit has failed, so the compositor wouldn't know what to do after a failed commit to perform a new non-failing commit. Neither Wayland nor libliftoff allow to rollback.
- Ordering between layers needs to be tracked. On the DRM backend, the zpos property will be used. The backends code will be much simpler if the common code takes care of the `wl_list` stuff (backends can just iterate over the list of layers to update their state when the ordering changes). Once the ordering is tracked, it's very little work to keep track of other things like position or buffer.
- The Wayland backend would need to track everything the common code currently tracks. One thing I think is important is making sure compositors can repeatedly change the pending state without any cost. It's pretty handy to be able to set the position of a layer at each frame, or to set the buffer to NULL then non-NULL in case importing it fails.

Let me know if any of the above is unclear or needs more explanation.

ascent12:
Hmm, the example doesn't seem to be showing anything for me other than the grey background. I can see the log messages from weston-simple-egl.

emersion:
I've investigated a little, and there are two things that could get in your way when testing this patch:

* Mesa uses wl_drm instead of linux-dmabuf if modifiers aren't supported, as it's the case on AMD: https://gitlab.freedesktop.org/mesa/mesa/blob/4ef2c53755686a22563dc2be4592d53b1cc96dc7/src/egl/drivers/dri2/platform_wayland.c#L920
  This prevents us from getting a dmabuf. Try with Intel instead?
* `weston-simple-dmabuf-egl` from stable Weston doesn't support xdg-shell. Try the one from Weston master.

Does it help?

emersion:
Rebased against master to fix the Alpine build failure.

emersion:
Thinking about this again, this plan is probably not a good idea. Output layers are not like the rest of the output state. An output commit can succeed even if all layers are rejected by the backend, and this just means the compositor needs to draw these layers manually.

I think a better approach is *not* to make output layers part of the output state. This means `wlr_output_rollback` would not affect layers.

emersion:
>I think a better approach is not to make output layers part of the output state.

Hmm, this is likely a bad idea in the end. We still want output layer changes to be applied on `wlr_output_commit` to ensure atomicity. If we start introducing some special cases ("a commit applies pending output state, plus output layers" and "a rollback reverts pending state, except output layers state"), then the API will become hairy pretty quickly.

include/wlr/types/wlr_output_layer.h
```diff
@@ -0,0 +1,83 @@
...
+	uint32_t committed; // enum wlr_output_layer_state_field
+	struct wl_list link;
+	struct wlr_buffer *buffer; // only valid if WLR_OUTPUT_LAYER_STATE_BUFFER
+	int x, y; // only valid if WLR_OUTPUT_LAYER_STATE_POSITION
+};
+
+struct wlr_output_layer {
+	struct wlr_output *output;
+	/** If true, the backend has accepted to display the layer. If false, the
+	 * compositor needs to manually render the layer. After each output commit,
```
kennylevinsen, line 40: - Should manual rendering be through a large layer that was initialized earlier, or through a buffer attached directly on the wlr_output?
- How do layers interact with buffers attached directly to a wlr_output? E.g. ordering - place above/below only operate on layers after all.

emersion: That's a good question.

Right now, the output's primary buffer (set via `wlr_output_attach_render` or `wlr_output_attach_buffer`) is always used, cannot be NULL, and is assumed to be the fallback in case it's not possible to display all output layers (ie, the compositor will paint layers that can't be displayed onto the primary buffer). The primary buffer is always below all output layers.

We should at least document all of that. We should also wonder whether this is a good design.

I think long-term we'll want to have the primary buffer be a regular layer just like the other layers. The main motivation is that both KMS and Wayland allow the primary buffer to be placed below layers and to set the same scaling/cropping/etc properties on both primary and layers.

However, we probably want to keep the old `wlr_output_attach_render`/`wlr_output_attach_buffer` functions working, at least for some time. If backends can take the primary buffer in a regular layer, these functions can probably just internally create a layer.

So I _think_ it's fine to keep the status quo in this PR as a first step, then upgrade to the long-term plan described above.

kennylevinsen: > Right now, the output's primary buffer (set via `wlr_output_attach_render` or `wlr_output_attach_buffer`) is always used, cannot be NULL, and is assumed to be the fallback in case it's not possible to display all output layers (ie, the compositor will paint layers that can't be displayed onto the primary buffer). The primary buffer is always below all output layers.

Hmm, what if the declined layer has a z-ordering above other opaque layers, so that it cannot be painted from below? We'd have to bail all intersecting layers as well in that case, or carve a transparent section into them.

emersion: That is correct. libliftoff should handle this case already. There are quite a few test cases, this one should cover the case you describe:

https://github.com/emersion/libliftoff/blob/799dddd6c449d7b34bedb24e413b4d754f05a9ab/test/test_alloc.c#L141

kennylevinsen: Ah, so the idea is that the implementation of wlr_output_layer guarantees that intersecting layers are also declined? I was worried the compositor would need to handle this. Such guarantee would be nice to mention here.

I'm fine with having the regular output buffer act as composition buffer, but that would also be nice to have documented as it not obvious.

emersion: Added some documentation about this.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/1985

Issue #1406, opened 2018-12-03
Variable refresh rate support


emersion:
e.g. FreeSync

Upcoming DRM feature: https://lists.freedesktop.org/archives/dri-devel/2018-October/192874.html

```
11:13 <pq> Shibe, SardemFF7, emersion, DRM freesync support as currently designed will require compositors to implement support for it, and clients to opt-in to it. Mesa will intergrate some rules about which GL/Vulkan apps will automatically opt in. Additionally, it will only be used for fullscreen windows, never for compositor itself.
11:15 <emersion> what does it mean to "opt in"? is the blacklist a way to opt-out?
11:16 <emersion> for fullscreen windows, what should the compositor do?
11:18 <pq> emersion, yes, the Mesa black-list is to avoid opt-inning apps that should never get freesync enabled, that includes all X11 composite managers.
11:18 <emersion> oh, i see
11:19 <pq> emersion, the compositor needs to set the VRR DRM property to allow VRR, when it detects the window that opted in being on direct scanout.
11:19 <emersion> ah, so the "detect the window has opted in" part is the _VARIABLE_REFRESH X11 property?
11:20 <pq> the net effet of all this is that all hardware-rendering GL/Vulkan apps that show fullscreen will likely get VRR enabled automatically, without any changes to the app.
11:20 <pq> emersion, yes
11:20 <emersion> okay, so we'd also need a protocol for Wayland, right?
11:21 <pq> yes
11:21 <emersion> (used by mesa)
11:21 <pq> correct, AFAIU
11:21 <ascent12> When you say "never for the compositor itself", are you stictly talking about x11 compositors or wayland ones too?
11:21 <emersion> i think it's just for X11 ones
11:22 <pq> ascent12, the assumption is that all compositors, both Wayland and X11, have too unpredictable timings to be used with VRR. Wildly changing timings may cause "luminance flicker" on the VRR-supporting (poor) monitors.
11:22 <emersion> wayland compositors won't enable VRR on the DRM, so they should be fine
11:22 <emersion> *DRM side
11:23 <pq> yeah, a Wayland compositor will just know to not shoot itself in the foot, while X11 compositors are X11 clients to they depend on the blacklist.
11:23 <pq> *so they
11:25 <Deuxis> someone needs to make an API / framework version of http://www.toodarkpark.org/computers/humor/shoot-self-in-foot.html
[...]

PR #1987, opened 2020-01-10
Variable Refresh Rate

emersion:
- [ ] Should we make the output commit fail if VRR can't be enabled? We don't know when VRR is indeed enabled, but in some cases we know when it fails. We still don't have support for test-only commits, so dealing with failures is going to be annoying for compositors.
- [x] Sway implementation

Fixes: https://github.com/swaywm/wlroots/issues/1406
~~Depends on: https://github.com/swaywm/wlroots/pull/2046~~

Future work: Wayland protocol to opt-in to VRR.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -412,6 +412,43 @@ static bool drm_connector_commit_buffer(struct wlr_output *output) {
 	return true;
 }
 
+static void drm_connector_enable_adaptive_sync(struct wlr_output *output,
+		bool enabled) {
+	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
+	struct wlr_drm_backend *drm = get_drm_backend_from_backend(output->backend);
+
+	struct wlr_drm_crtc *crtc = conn->crtc;
+	if (!crtc) {
+		return;
+	}
+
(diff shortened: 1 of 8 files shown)


emersion:
This is ready for review.

emersion:
Rebased against master.

ddevault:
Can we bubble up whether or not the output supports adaptive sync, and add that to the swaymsg -t get_outputs response?

emersion:
I've decided against bubbling up this information, because other backends (X11 and Wayland) won't be able to know.

ddevault:
`{SUPPORTED,UNSUPPORTED,UNKNOWN}`

emersion:
>Can we bubble up whether or not the output supports adaptive sync, and add that to the swaymsg -t get_outputs response?

Done, but decided to report the current status instead, it's more reliable.

ddevault:
Thanks!

ruabmbua:
Wow this is amazing. Just tested it with my new setup, and I will never go back to windows, even for gaming.
Currently playing native shadow of the tomb raider on AMD graphics in sway, and I never had a better experience anywhere else.

Buttersmooth 160fps without any artifacts, and super low latency.

PR #1988, opened 2020-01-10
Drop RDP backend

ddevault:
Users interested in remote access to wlroots compositors should use
wayvnc:

https://github.com/any1/wayvnc


emersion:
Thanks!

PR #1989, opened 2020-01-10
Improve numerical precision of fractional scale factor.

jameswalmsley:
Relates to: https://github.com/swaywm/sway/issues/4896

Setting a fractional scale using wlr-output-management-v1 does not cause the same output settings
as using e.g. swaymsg.

### DIFF
--- a/protocol/wlr-output-management-unstable-v1.xml
+++ b/protocol/wlr-output-management-unstable-v1.xml
@@ -479,5 +479,12 @@
       </description>
       <arg name="scale" type="fixed"/>
     </request>
+
+    <request name="set_scale_by_target_width">
+      <description summary="set the scaled target width">
+        This request sets the head's scale.
+      </description>
+      <arg name="target_width" type="int" summary="scaled width"/>
+    </request>
   </interface>
 </protocol>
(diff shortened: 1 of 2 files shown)


jameswalmsley:
Updated the idea to extend the protocol in a compatible way.

ddevault:
Honestly, I don't find this change especially compelling. The behavior you observed (1 pixel space) seems likely to be unrelated, this just hides the problem.

jameswalmsley:
Hi @ddevault, @emersion  maybe this is just an issue that doesn't need solving.

Anyway I've pushed one final suggestion that provides a method for setting the scale by specifying the desired target width.

My thoughts were maybe some applications might have the need to set a precise scale, maybe not for desktop stuff but e.g. multi-panel wayland compositors or specialised display systems.

In any case this gave me an good excuse to dive in and learn some stuff about wayland and wlroots.

Super grateful for all the hard work you are putting into your open-source projects, and the great work of the whole sway community. I guess I'll have to buy your wayland book and learn more.

Best 

James

ddevault:
Yeah, I don't think that's going to be the solution, either. I appreciate your putting in the work and time to research possible solutions, though.

PR #1990, opened 2020-01-11
render/egl: remove SURFACELESS_MESA special case

emersion:
Users can just pass EGL_DEFAULT_DISPLAY themselves.


ddevault:
Thanks!

PR #1992, opened 2020-01-13
render/gles2: unbind textures after use

emersion:
Keeping textures bound results in hard-to-debug situations where some GL
operations incorrectly affect the texture.


ddevault:
Thanks!

PR #1995, opened 2020-01-14
Bump meson version to 0.51.2

### LINKED ISSUE #1994: Fails to configure with no ffmpeg dependencies
When there are no ffmpeg dependencies installed, meson fails to configure wlroots on `dmabuf-capture` example regardless of whether examples are enabled or not.

```
examples/meson.build:170:1: ERROR: Argument is of an unacceptable type 'Disabler'.
Must be either an external dependency (returned by find_library() or dependency()) or an internal dependency (returned by declare_dependency()).
```

Meson 0.51.1

ascent12:
Closes #1994 

There was an issue in 0.51.0 and earlier, where lists of dependencies
and disablers weren't acting like they should. Instead of disabling a
build, it would error out instead.

Changing this logic to work around it is annoying, so just bump the
version instead.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
 	'c',
 	version: '0.9.1',
 	license: 'MIT',
-	meson_version: '>=0.51.0',
+	meson_version: '>=0.51.2',
 	default_options: [
 		'c_std=c11',
 		'warning_level=2',


ascent12:
I didn't even realize there was a 0.51.2. Yeah, definitely makes sense to specify that instead.

PR #2001, opened 2020-01-17
backend/drm: prevent outputs from being destroyed on commit

emersion:
This would happen if initializing the renderer fails. Instead, we just
mark the output as disabled.

References: https://github.com/swaywm/sway/pull/4917


ddevault:
Thanks!

PR #2002, opened 2020-01-17
keyboard-group: two NULL keymaps are equal

emersion:
Previously, creating a keyboard group without any keymap set would
result in an error:

    Device keymap does not match keyboard group's


RedSoxFan:
Thanks!

PR #2007, opened 2020-01-24
meson: Fix protocol includes for compositor examples

ascent12:
This should fix the alpine build, due to underspecifying the example dependencies.


examples/meson.build
```diff
@@ -151,9 +152,14 @@ clients = {
 }
 
 foreach name, info : compositors
+	extra_src = []
+	foreach p : info.get('proto', [])
+		extra_src += get_variable(p.underscorify() + '_server_h')
```
emersion: We probably need to add the '_c' file too right?

PR #2006, opened 2020-01-24
fix build errors when using -fno-common

thermitegod:
GCC 10 will enable -fno-common by default and wlroots currently fails to compile when using this flag.

PR #2013, opened 2020-01-28
Declare wlr_seat globals as extern

khardix:
When compiling wlroots with GCC 10, the linking fails with `multiple definition of 'default_pointer_grab_impl'` and similar errors for `default_keyboard_grab_impl` and `default_touch_grab_impl`. Full error: [gcc-10-ld.err.txt](https://github.com/swaywm/wlroots/files/4122843/gcc-10-ld.err.txt)

If I recall my C correctly, these should be marked `extern` – they just declare that the objects exist somewhere in the linked library, not necessarily in the current translation unit. Adding the `extern` qualifier causes the linking to pass successfully.

### DIFF
--- a/include/types/wlr_seat.h
+++ b/include/types/wlr_seat.h
@@ -4,9 +4,9 @@
 #include <wayland-server-core.h>
 #include <wlr/types/wlr_seat.h>
 
-const struct wlr_pointer_grab_interface default_pointer_grab_impl;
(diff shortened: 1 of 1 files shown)


khardix:
Ah, seems that this is a duplicate of #2006 – feel free to close.

emersion:
Hmm. The other PR was not rebased, so merging this one instead.

PR #2016, opened 2020-01-31
Remove minor keyword incompatibilities with C++

nitaigao:
Hi wlroots Team :wave: 

When working on my compositor I have noticed there are a small number of statements that cause some incompatibilities with C++.
 
- A few `static` array statements are peppered throughout the headers. 
- The `xwayland` struct has a member named `class` which is a reserved keyword.

I humbly raise this PR to make the adjustments to fix up these issues. Other than these pretty much everything else works great with C++.

Thanks for all the work on such a great framework, i hope this doesn't inconvenience too much.

### DIFF
--- a/backend/drm/util.c
+++ b/backend/drm/util.c
@@ -221,7 +221,7 @@ uint32_t get_fb_for_bo(struct gbm_bo *bo, uint32_t drm_format,
 	return id;
 }
 
-static inline bool is_taken(size_t n, const uint32_t arr[static n], uint32_t key) {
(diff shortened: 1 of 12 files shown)


ddevault:
Hiya, this has been brought up before but we're not interested. You should forward declare things yourself or some similar workaround. We don't subscribe to the line of thinking which presumes C and C++ are the same language, to us C++ is no different from any other FFI.

nitaigao:
ok thanks for the info, will do.

PR #2017, opened 2020-02-03
xwm: correctly send focus event to new window

Emantor:
Even if we allow the focus change to the new window, we have to send the
focus explicitly. Otherwise the focus may get lost when we close a
window belonging to the same process, but do not send focus to the
remaining window.

Fixes swaywm/sway#4926

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -1245,6 +1245,7 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
 		xwm_send_focus_window(xwm, xwm->focus_surface);
 	} else {
 		xwm->focus_surface = requested_focus;
+		xwm_send_focus_window(xwm, xwm->focus_surface);
 	}
 }
 


emberfade:
Unfortunately this did not fix the issue with firefox for me.

Emantor:
Indeed, it is *keyboard* input which breaks, I was only testing *mouse* input. I'll need to dig into this some more.

ammen99:
IIRC I tried to do what this PR does, but it caused Xwayland lockups. Could have been a bug in Wayfire though.

choopm:
Feedback for your changes:
This patch doesn't fix my issue with rxvt-unicode in daemon mode (empty workspace, create two urxvtc terminals, close one using ctrl+d -> focus doesn't move to prior focussed window) but makes the cursor go crazy blinking (I'm using hide cursor while typing) when hovering over some terminals, looks inconsistent and makes things worse for me.

PR #2019, opened 2020-02-05
virtual-pointer: Add request for mapping to specific output

any1:


emersion:
Can you open a wlr-protocols pull request for the protocol changes too?

any1:
Can do. The wayvnc branch for this is here: https://github.com/any1/wayvnc/tree/input-output-mapping

include/wlr/types/wlr_virtual_pointer_v1.h
```diff
@@ -45,6 +46,7 @@ struct wlr_virtual_pointer_v1_new_pointer_event {
 	struct wlr_virtual_pointer_v1 *new_pointer;
 	/** Suggested by client; may be NULL. */
 	struct wlr_seat *suggested_seat;
+	struct wlr_output *selected_output;
```
emersion: Maybe name this `suggested_output`, since we already have `suggested_seat`.

Also a comment indicating that it may be NULL would be nice.

emersion:
Overall looks good. Can you update this patch to use the latest protocol proposal?

emersion:
Thanks!

PR #2023, opened 2020-02-09
Fix wl_output_transform rotation direction, and composition

mstoeckl:
See https://gitlab.freedesktop.org/wayland/weston/issues/99 for motivation. I also found and fixed a small issue in `wlr_output_transform_compose` when composing transforms with flips and rotations.

This change breaks forward compatibility for past versions of wlroots; for example, upgrading wlroots with a corresponding patch, without upgrading sway, could change monitor orientations for anyone using sway's https://github.com/swaywm/sway/blob/master/sway/commands/output/transform.c .

I have tested these changes using ~the wlroots example `rotation` compositor~ and `swaymsg output eDP-1 transform flipped-90 clockwise`, and by patching an image display program to call wl_surface::set_buffer_transform() with a given value.

* * *

Breaking change: output transformations have been aligned with the Wayland protocol's definition

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -1108,7 +1108,15 @@ enum wl_output_transform wlr_output_transform_invert(
 enum wl_output_transform wlr_output_transform_compose(
 		enum wl_output_transform tr_a, enum wl_output_transform tr_b) {
 	uint32_t flipped = (tr_a ^ tr_b) & WL_OUTPUT_TRANSFORM_FLIPPED;
-	uint32_t rotated =
-		(tr_a + tr_b) & (WL_OUTPUT_TRANSFORM_90 | WL_OUTPUT_TRANSFORM_180);
+	uint32_t rotation_mask = WL_OUTPUT_TRANSFORM_90 | WL_OUTPUT_TRANSFORM_180;
+	uint32_t rotated;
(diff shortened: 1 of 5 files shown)


types/wlr_output.c
```diff
@@ -1108,7 +1108,15 @@ enum wl_output_transform wlr_output_transform_invert(
...
 	uint32_t flipped = (tr_a ^ tr_b) & WL_OUTPUT_TRANSFORM_FLIPPED;
-	uint32_t rotated =
-		(tr_a + tr_b) & (WL_OUTPUT_TRANSFORM_90 | WL_OUTPUT_TRANSFORM_180);
+	uint32_t rotation_mask = WL_OUTPUT_TRANSFORM_90 | WL_OUTPUT_TRANSFORM_180;
+	uint32_t rotated;
+	if (tr_b & WL_OUTPUT_TRANSFORM_FLIPPED) {
+		// When a rotation of k degrees is followed by a flip, the
+		// equivalent transform is a flip followed by a rotation of
+		// -k degrees.
+		rotated = (tr_b - tr_a) & rotation_mask;
```
emersion: Does this need an update? Since 90 and 270 are just swapped, I would've guessed not?

emersion: Oh, my bad, the commit messages explains it.

ddevault:
Thanks!

PR #2026, opened 2020-02-15
Keyboard shortcuts inhibit

michaelweiser:
The keyboard shortcuts inhibitor protocol is useful for remote desktop
and virtualization software in order to request all keyboard events to
be passed to it and (almost) none being resonded to by the compositor.
This allows the session at the other end of the remote desktop
connection or inside the virtual machine to be interacted with as usual
(e.g. Alt+Tab to switch windows on the remote system instead of
locally).

Add the wayland protocol to the meson build files.

Copy'n'search'n'replace the very similar idle inhibit protocol
implementation. This already provides all the basic functionality:
- creating and destroying inhibitors upon request by a client,
- destruction in reaction to destruction of surfaces or displays,
- a list of inhibitors to search through for existing ones as well as
- a signal to be sent to the compositor upon registration of a new
  inhibitor.

Beyond that we add the active and inactive events to be sent to the
client and wire those to activate and deactivate functions for the
compositor to call in confirmation of activation of a new inhibitor or
(un-)suspending of an existing inhibitor e.g. in response to a special
key combination entered by the user as suggested by the protocol.

As mandated by the protocol, we check the existance of an inhibitor for
a given surface and seat upon creation and return the error provided by
the protocol for that purpose.

~~As an optimisation we add a list of per-surface inhibitors to
wlr_surface and a helper that looks up the inhibitor for a given seat
and surface based on the focused surface being tracked in the seat's
keyboard. This is done on the assumption that most compositors will want
to know if the surface keyboard input would currently be sent to if it
wasn't interpreted as compositor shortcuts has an inhibitor attached.
This also avoids passing the inhibitor manager around to all the places
such a check would be needed and then searching the list of all
registered inhibitors.~~

Again, copy'n'search'n'replace the idle inhibit example to become a
simple keyboard shortcuts inhibit example, adding the active and
inactive events.

### DIFF
--- a/include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
+++ b/include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
@@ -1,0 +1,85 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_KEYBOARD_SHORTCUTS_INHIBIT_V1_H
+#define WLR_TYPES_WLR_KEYBOARD_SHORTCUTS_INHIBIT_V1_H
+
+#include <wayland-server-core.h>
+#include <wlr/types/wlr_seat.h>
+
+/* This interface permits clients to inhibit keyboard shortcut processing by
+ * the compositor.
+ *
+ * This allows clients to pass them on to e.g. remote desktops or virtual
+ * machine guests.
+ *
+ * Inhibitors are created for surfaces and seats. They should only be in effect
+ * while this surface has focus.
+ */
+
+struct wlr_keyboard_shortcuts_inhibit_manager_v1 {
+	// wlr_keyboard_shortcuts_inhibitor_v1::link
+	struct wl_list inhibitors;
+	struct wl_global *global;
+
+	struct wl_listener display_destroy;
+
+	struct {
+		struct wl_signal new_inhibitor;	// wlr_keyboard_shortcuts_inhibitor_v1
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_keyboard_shortcuts_inhibitor_v1 {
+	struct wlr_surface *surface;
+	struct wlr_seat *seat;
+	bool active;
+	struct wl_resource *resource;
+
+	struct wl_listener surface_destroy;
+	struct wl_listener seat_destroy;
+
+	// wlr_keyboard_shortcuts_inhibit_manager_v1::inhibitors
+	struct wl_list link;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+/*
+ * A compositor creating a manager will handle the new_inhibitor event and call
+ * wlr_keyboard_shortcuts_inhibitor_v1_activate() if it decides to honour the
+ * inhibitor. This will send the active event to the client, confirming
+ * activation of the inhibitor. From then on the compositor should respect the
+ * inhibitor until it calls wlr_keyboard_shortcuts_inhibitor_v1_deactivate() to
+ * suspend the inhibitor with an inactive event to the client or receives the
+ * destroy signal from wlroots, telling it that the inhibitor has been
+ * destroyed.
+ *
+ * Not sending the active event to the client is the only way under the
+ * protocol to let the client know that the compositor will not be honouring an
+ * inhibitor. It's the client's job to somehow deal with not receiving the
+ * event, i.e. not assume that shortcuts are inhibited and maybe destroy the
+ * pending and request a new inhibitor after a timeout.
+ */
+
+struct wlr_keyboard_shortcuts_inhibit_manager_v1 *
+wlr_keyboard_shortcuts_inhibit_v1_create(struct wl_display *display);
+
+void wlr_keyboard_shortcuts_inhibitor_v1_activate(
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor);
+
+void wlr_keyboard_shortcuts_inhibitor_v1_deactivate(
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor);
+
+#endif
--- a/types/wlr_keyboard_shortcuts_inhibit_v1.c
+++ b/types/wlr_keyboard_shortcuts_inhibit_v1.c
@@ -1,0 +1,227 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <util/signal.h>
+#include <wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h>
+#include "keyboard-shortcuts-inhibit-unstable-v1-protocol.h"
+
+static const struct zwp_keyboard_shortcuts_inhibit_manager_v1_interface
+	keyboard_shortcuts_inhibit_impl;
+
+static const struct zwp_keyboard_shortcuts_inhibitor_v1_interface
+	keyboard_shortcuts_inhibitor_impl;
+
+static struct wlr_keyboard_shortcuts_inhibit_manager_v1 *
+wlr_keyboard_shortcuts_inhibit_manager_v1_from_resource(
+		struct wl_resource *manager_resource) {
+	assert(wl_resource_instance_of(manager_resource,
+		&zwp_keyboard_shortcuts_inhibit_manager_v1_interface,
+		&keyboard_shortcuts_inhibit_impl));
+	return wl_resource_get_user_data(manager_resource);
+}
+
+static struct wlr_keyboard_shortcuts_inhibitor_v1 *
+wlr_keyboard_shortcuts_inhibitor_v1_from_resource(
+		struct wl_resource *inhibitor_resource) {
+	assert(wl_resource_instance_of(inhibitor_resource,
+		&zwp_keyboard_shortcuts_inhibitor_v1_interface,
+		&keyboard_shortcuts_inhibitor_impl));
+	return wl_resource_get_user_data(inhibitor_resource);
+}
+
+static void keyboard_shortcuts_inhibitor_v1_destroy(
+		struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor) {
+	if (!inhibitor) {
+		return;
+	}
+
+	wlr_signal_emit_safe(&inhibitor->events.destroy, inhibitor);
+
+	wl_resource_set_user_data(inhibitor->resource, NULL);
+	wl_list_remove(&inhibitor->link);
+	wl_list_remove(&inhibitor->surface_destroy.link);
+	wl_list_remove(&inhibitor->seat_destroy.link);
+	free(inhibitor);
+}
+
+static void keyboard_shortcuts_inhibitor_v1_handle_resource_destroy(
+		struct wl_resource *inhibitor_resource) {
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor =
+		wlr_keyboard_shortcuts_inhibitor_v1_from_resource(
+				inhibitor_resource);
+	keyboard_shortcuts_inhibitor_v1_destroy(inhibitor);
+}
+
+static void keyboard_shortcuts_inhibitor_handle_surface_destroy(
+		struct wl_listener *listener, void *data) {
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor =
+		wl_container_of(listener, inhibitor, surface_destroy);
+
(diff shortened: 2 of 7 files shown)


michaelweiser:
Ongoing sway implementation is at https://github.com/michaelweiser/sway/tree/keyboard-shortcuts-inhibit.

include/wlr/types/wlr_surface.h
```diff
@@ -109,6 +109,9 @@ struct wlr_surface {
 	// wlr_subsurface::parent_pending_link
 	struct wl_list subsurface_pending_list;
 
+	// wlr_keyboard_shortcuts_inhibitor_v1
+	struct wl_list keyboard_shortcuts_inhibitors;
```
emersion: We try not to leak protocol implementation details in `wlr_surface`. Moreover, the protocol is unstable, so new versions will change how this works.

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,228 @@
...
+
+	wl_resource_set_implementation(wl_resource,
+		&keyboard_shortcuts_inhibit_impl,
+		keyboard_shortcuts_inhibit, NULL);
+}
+
+struct wlr_keyboard_shortcuts_inhibit_manager_v1 *
+wlr_keyboard_shortcuts_inhibit_v1_create(struct wl_display *display) {
+	struct wlr_keyboard_shortcuts_inhibit_manager_v1
+		*keyboard_shortcuts_inhibit = calloc(1,
```
emersion: Style: we generally break this kind of lines right after the `=` sign

emersion: This variable should probably be named `manager`, to have a clear difference with inhibitors.

include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
```diff
@@ -0,0 +1,103 @@
...
+
+	void *data;
+};
+
+struct wlr_keyboard_shortcuts_inhibitor_v1 {
+	struct wlr_surface *surface;
+	struct wlr_seat_client *seat_client;
+	bool active;
+	struct wl_resource *resource;
+	struct wl_listener surface_destroy;
```
emersion: We probably also need a listener of the seat client destruction?

include/wlr/types/wlr_surface.h
michaelweiser: I expected as much but it was the one way I could think of to avoid bookkeeping by and passing around of the manager and inhibitors within the compositor. Can you recommend another way of  implementing similar properties? Or is this the way to go: Hand the manager and inhibitors to the compositor and tell it to get that information to where it needs it its own way?

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Done (assuming I caught them all).

michaelweiser: Done.

include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
michaelweiser: Right and done.

Aside: Is seat client a good choice here? It was the one seat type object I could find a function for to turn up a reference on from a wayland seat resource. The inhibitor could just as well just remember the seat which would be closer to the protocol spec and likely still provide the compositor with all it needs to know. I've got no idea of the repercussions either way.

include/wlr/types/wlr_surface.h
ddevault: The latter might work. It could also be possible to store the surface on the inhibitor and provide a convenience function which answers "does this surface have an inhibitor?" by simply iterating through all of the inhibitors.

ddevault:
LGTM, nice work.

include/wlr/types/wlr_surface.h
michaelweiser: This is what I ended up doing in the updated sway implementation: I've effectively moved the per-surface backlink and convenience function into the compositor. I've put that up for discussion as swaywm/sway#5021.

michaelweiser:
Let me know whether and when I should squash all the delta up into the initial two commits of adding the protocol and the example.

ddevault:
Now would be a good time to do that.

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	manager->display_destroy.notify = handle_display_destroy;
+	wl_display_add_destroy_listener(display, &manager->display_destroy);
+
+	return manager;
+}
+
+void wlr_keyboard_shortcuts_inhibitor_v1_activate(
+		struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor) {
```
emersion: It would be nice to make this a no-op if the inhibitor is already active.

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,212 @@
...
+	zwp_keyboard_shortcuts_inhibitor_v1_send_active(inhibitor->resource);
+	inhibitor->active = true;
+}
+
+void wlr_keyboard_shortcuts_inhibitor_v1_deactivate(
+		struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor) {
```
emersion: Ditto

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,212 @@
...
+
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor =
+		calloc(1, sizeof(struct wlr_keyboard_shortcuts_inhibitor_v1));
+	if (!inhibitor) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	struct wl_resource *wl_resource = wl_resource_create(client,
+		&zwp_keyboard_shortcuts_inhibitor_v1_interface, 1, id);
```
emersion: The version must be taken from the manager resource:

```c
uint32_t version = wl_resource_get_version(resource);
```

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,212 @@
...
+		struct wl_listener *listener, void *data) {
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor =
+		wl_container_of(listener, inhibitor, seat_client_destroy);
+	keyboard_shortcuts_inhibitor_v1_destroy(inhibitor);
+}
+
+static void keyboard_shortcuts_inhibitor_v1_handle_destroy(
+		struct wl_client *client,
+		struct wl_resource *manager_resource) {
+	wl_resource_destroy(manager_resource);
```
emersion: This name is misleading, since it's not a manager resource, it's an inhibitor resource. We should rename this variable to `resource` or `inhibitor_resource`.

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor =
+		calloc(1, sizeof(struct wlr_keyboard_shortcuts_inhibitor_v1));
+	if (!inhibitor) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+
+	struct wl_resource *wl_resource = wl_resource_create(client,
```
emersion: We have both `wl_resource` (the inhibitor) and `resource` (the manager). It would be nice to rename those to make them less confusing.

types/wlr_keyboard_shortcuts_inhibit_v1.c
```diff
@@ -0,0 +1,212 @@
...
+
+static void keyboard_shortcuts_inhibitor_v1_destroy(
+		struct wlr_keyboard_shortcuts_inhibitor_v1 *inhibitor) {
+	if (!inhibitor) {
+		return;
+	}
+
+	wlr_signal_emit_safe(&inhibitor->events.destroy, inhibitor);
+
+	wl_resource_set_user_data(inhibitor->resource, NULL);
```
emersion: We should probably implicitly send a deactivate event here.

include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
```diff
@@ -0,0 +1,85 @@
...
+	struct wlr_surface *surface;
+	struct wlr_seat_client *seat_client;
+	bool active;
+	struct wl_resource *resource;
+
+	struct wl_listener surface_destroy;
+	struct wl_listener seat_client_destroy;
+
+	// wlr_keyboard_shortcuts_inhibit_manager_v1::inhibitors
+	struct wl_list manager_link;
```
emersion: This can be renamed to `link` now that the other type of link is gone.

include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
emersion: Yeah, I think using the seat would be better, since it'll still work if the client destroys all of its `wl_seat` resources.

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Mhmm, isn't destroying an inhibitor in its strictest sense something different than deactivating it? My understanding is that a client creates or destroys an inhibitor while the server is only able to (de-)activate but *not* destroy it. Based on that, wouldn't the client obviously assume the inhibitor to become irrelevant/inactive after being destroyed? Couldn't it even be confusing to a client state machine to receive an inactive event while it's destroying it?

As prior art, pointer contraint doesn't send unconfined/unlocked events on destroy either.

The protocol has no wording on it either way.

include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
michaelweiser: Done.

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Done.

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Done.

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Done. Since all of these originated in idle-inhibit I also threw in a patch that backports them there for your consideration.

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Done.

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: Done.

include/wlr/types/wlr_keyboard_shortcuts_inhibit_v1.h
michaelweiser: Done.

types/wlr_keyboard_shortcuts_inhibit_v1.c
emersion: I'm worried about the case where the seat or surface gets destroyed, and this implicitly makes the inhibitor inert. In this case the client won't know that the inhibitor is no longer active, yet the inhibitor object will still be alive.

michaelweiser: Ah, then we could send the notification from the respective resource destroy handlers only: https://github.com/swaywm/wlroots/pull/2026/commits/44619800af73169f2bae283f8e252cd13edb3352.

Should pointer constraints do this as well then?

emersion: Probably yeah!

emersion:
This PR LGTM. Can you squash the keyboard-shortcuts-inhibit changes together?

types/wlr_keyboard_shortcuts_inhibit_v1.c
michaelweiser: I had a look but it's not as simple since constraints can be updated (set_region). So tracking if it's active and a notification should be sent is not as straightforward. Should I even attempt to muck about in there? :)

emersion: No worries, I've created an issue for this: https://github.com/swaywm/wlroots/issues/2038

michaelweiser:
All done. sway implementation updated as well.

PR #2031, opened 2020-02-18
output-management: add current_configuration_dirty

ammen99:
Previously, if the current configuration contains an output X which is destroyed, its head is automatically removed. If the compositor submits the new configuration after X was removed, the current output configuration is incorrectly detected to be the same as the previous one, and no done event is sent. To prevent this, we can just keep track of whether the current configuration is dirty, i.e whether we have sent a done event for it.


ammen99:
Not really sure what is going on with the alpine build, I don't think it is caused by my PR.

ascent12:
>Not really sure what is going on with the alpine build, I don't think it is caused by my PR.

Rebase your branch onto master. This is something I've already fixed.

PR #2036, opened 2020-02-19
text_input_v3: Note features supported by the text field

dcz-purism:
With this information, consumers can realize they will never retrieve some state, and adjust their strategy.

---

The protocol specifies that features are either supported or unsupported, so while the API user is able to determine this by the received calls, this functionality will be required by pretty much every API user in exactly the same form, so it makes sense to make it available here.


ddevault:
Thanks!

PR #2037, opened 2020-02-19
renderer: Add an ability to render to texture

dos1:
This can be useful for features like window thumbnails or correct
alpha blending of windows with subsurfaces.

This introduces a new wlr_renderer_begin argument - target texture.
Passing NULL makes it stay at current behavior (render to screen).


emersion:
I think this should live in the compositor, not in wlroots.

dos1:
This is easy to do when you just assume that wlr_renderer is backed by GLES2 - is this a sensible assumption to make when using wlroots? (from API intention perspective of course, I can already see the code ;))

emersion:
>is this a sensible assumption to make when using wlroots?

One can call `wlr_gles2_renderer_create` instead of `wlr_renderer_autocreate`.

PR #2042, opened 2020-02-27
wlr_output_power_management_v1: Init output_power->link

agx:
This makes sure the `wl_list_remove(&output_power->link)` in
`output_power_destroy()` does not crash even when the output_power never
got added to a list. This can e.g. happen in the `mgmt->output ==
output` error path of `output_power_manager_get_output_power`.


ddevault:
Thanks!

PR #2043, opened 2020-02-28
Introduce wlr_client_buffer

emersion:
Split out the client/resource handling out of wlr_buffer by introducing
wlr_client_buffer. Make wlr_buffer an interface so that compositors can
create their own wlr_buffers (e.g. backed by GBM, like glider [1]).

[1]: https://github.com/emersion/glider/blob/c66847dd1cf8ae5e68666ce7cb3be41427b38dc7/include/gbm_allocator.h#L7

Should we move `wlr_client_buffer` to its own header file?

* * *

This is a breaking change. Compositors that don't directly use `wlr_buffer` shouldn't need an update. For compositors which use it directly:

* `wlr_buffer` has been renamed to `wlr_client_buffer`
* `wlr_buffer_{ref,unref,get_dmabuf}` still take a `wlr_buffer`, compositors need to use `&wlr_client_buffer->base`
* `wlr_buffer_create` and `wlr_buffer_apply_damage` have been renamed to `wlr_client_buffer_create` and `wlr_client_buffer_apply_damage`
* `wlr_buffer_get_resource_size` has been renamed to `wlr_resource_get_buffer_size`

Sway PR: https://github.com/swaywm/sway/pull/5049

### DIFF
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -13,14 +13,52 @@
 #include <wayland-server-core.h>
 #include <wlr/render/dmabuf.h>
 
+struct wlr_buffer;
+
+struct wlr_buffer_impl {
+	void (*destroy)(struct wlr_buffer *buffer);
+	bool (*get_dmabuf)(struct wlr_buffer *buffer,
+		struct wlr_dmabuf_attributes *attribs);
+};
+
+struct wlr_buffer {
+	const struct wlr_buffer_impl *impl;
+
+	size_t n_refs;
+};
+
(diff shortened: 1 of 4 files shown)


ddevault:
Does this make https://github.com/swaywm/wlroots/pull/1910 obsolete?

emersion:
No. #1910 allows compositors to register additional client buffer types, this PR is about compositor buffer types (such as buffers directly allocated via GBM). So these two PRs are orthogonal.

With this PR, #1910 should probably define `wlr_client_buffer_impl` instead of `wlr_buffer_impl`.

ddevault:
Thanks! Nice work.

PR #2044, opened 2020-02-28
Augment wlr_buffer to support swapchains

emersion:
~~Depends on https://github.com/swaywm/wlroots/pull/2043~~

This PR adds the required infrastructure to `wlr_buffer` to support swapchains. glider has been updated to use it: https://github.com/emersion/glider/pull/7.

Once this PR is merged, it will be possible to add DRM support to https://github.com/swaywm/wlroots/pull/1985 via libliftoff.

* * *

`wlr_buffer` now has a concept of "released", "producer" and "consumer". A buffer is created by a producer (the allocator or the client). It's then referenced by consumers (e.g. the backend or the compositor). Once all consumers are done with the buffer (ie. once they have un-referenced the buffer), the buffer is released and can be re-used (by the swapchain or the client).

This is inspired from glider's buffer: https://github.com/emersion/glider/blob/c66847dd1cf8ae5e68666ce7cb3be41427b38dc7/include/allocator.h#L17. ~~One difference is that this PR is less explicit: both consumers and producers call `wlr_buffer_unref`. On glider, there's one API for producers (`glider_buffer_init`, `glider_buffer_unref`) and a separate API for consumers (`glider_buffer_lock`, `glider_buffer_unlock`). This makes it more obvious when the buffer is released. Is it worth it to make this breaking change?~~ Renamed.

glider's buffer is itself inspired from the renderer v6's `wlr_image` type: https://github.com/swaywm/wlroots/pull/1355/files#diff-6105fb2a3a3620a8a1aabaa7b11beb84R16

* * *

Breaking changes:

- Compositors that want to keep a reference to a `wlr_buffer`  to read it later should call `wlr_buffer_lock` and `wlr_buffer_unlock` instead of `wlr_buffer_ref` and `wlr_buffer_unref`. Same applies to backends.
- `wlr_client_buffer_create` has been renamed to `wlr_client_buffer_import`. Callers need to `wlr_buffer_unlock` the resulting buffer once they're done with it.
- Users who create buffers need to call `wlr_buffer_drop` instead of `wlr_buffer_unref`.

Sway PR: https://github.com/swaywm/sway/pull/5088

### DIFF
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -32,6 +32,8 @@ struct wlr_buffer_impl {
 struct wlr_buffer {
 	const struct wlr_buffer_impl *impl;
 
+	int width, height;
+
 	bool dropped;
 	size_t n_locks;
 
@@ -47,7 +49,7 @@ struct wlr_buffer {
  * they should call wlr_buffer_drop.
  */
 void wlr_buffer_init(struct wlr_buffer *buffer,
-	const struct wlr_buffer_impl *impl);
+	const struct wlr_buffer_impl *impl, int width, int height);
 /**
  * Unreference the buffer. This function should be called by producers when
  * they are done with the buffer.
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -25,6 +25,10 @@ struct wlr_buffer {
 	const struct wlr_buffer_impl *impl;
 
 	size_t n_refs;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
 };
 
 void wlr_buffer_init(struct wlr_buffer *buffer,
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -21,27 +21,49 @@ struct wlr_buffer_impl {
 		struct wlr_dmabuf_attributes *attribs);
 };
 
+/**
+ * A buffer containing pixel data.
+ *
+ * A buffer has a single producer (the party who created the buffer) and
+ * multiple consumers (parties reading the buffer). When all consumers are done
+ * with the buffer, it gets released and can be re-used by the producer. When
+ * the producer and all consumers are done with the buffer, it gets destroyed.
+ */
 struct wlr_buffer {
 	const struct wlr_buffer_impl *impl;
 
-	size_t n_refs;
+	bool dropped;
+	size_t n_locks;
 
 	struct {
 		struct wl_signal destroy;
+		struct wl_signal release;
 	} events;
 };
 
+/**
+ * Initialize a buffer. This function should be called by producers. The
+ * initialized buffer is referenced: once the producer is done with the buffer
+ * they should call wlr_buffer_drop.
+ */
 void wlr_buffer_init(struct wlr_buffer *buffer,
 	const struct wlr_buffer_impl *impl);
 /**
- * Reference the buffer.
+ * Unreference the buffer. This function should be called by producers when
+ * they are done with the buffer.
  */
-struct wlr_buffer *wlr_buffer_ref(struct wlr_buffer *buffer);
+void wlr_buffer_drop(struct wlr_buffer *buffer);
 /**
- * Unreference the buffer. After this call, `buffer` may not be accessed
- * anymore.
+ * Lock the buffer. This function should be called by consumers to make
+ * sure the buffer can be safely read from. Once the consumer is done with the
+ * buffer, they should call wlr_buffer_unlock.
  */
-void wlr_buffer_unref(struct wlr_buffer *buffer);
+struct wlr_buffer *wlr_buffer_lock(struct wlr_buffer *buffer);
+/**
+ * Unlock the buffer. This function should be called by consumers once they are
+ * done with the buffer.
+ */
+void wlr_buffer_unlock(struct wlr_buffer *buffer);
 /**
  * Reads the DMA-BUF attributes of the buffer. If this buffer isn't a DMA-BUF,
  * returns false.
@@ -70,6 +92,7 @@ struct wlr_client_buffer {
 	struct wlr_texture *texture;
 
 	struct wl_listener resource_destroy;
+	struct wl_listener release;
 };
 
 struct wlr_renderer;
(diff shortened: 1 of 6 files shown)


emersion:
This is now ready for review.

types/wlr_surface.c
```diff
@@ -310,6 +310,11 @@ static void surface_apply_damage(struct wlr_surface *surface) {
 		wlr_buffer_unref(&surface->buffer->base);
 	}
 	surface->buffer = buffer;
+
+	// wlr_surface is a buffer consumer: the buffer should only be released
+	// when another one replaces it
+	wlr_buffer_ref(&buffer->base);
+	wlr_buffer_unref(&buffer->base);
```
ascent12: Why is this code referencing and immediately unreferencing the buffer?

emersion: This is a consequence of the naming decisions explained in the PR description ("One difference is that this PR is less explicit: both consumers and producers call wlr_buffer_unref").

`wlr_surface` is a little special: it produces `wlr_buffer` objects from buffers coming from clients, but it also allows the compositor to access `wlr_surface.buffer.texture` at any time for rendering. This means `wlr_surface` is also a consumer of the buffer (because it exposes an API to read from the buffer). This ref+unref sequence allows to transition from being a producer to being a consumer.

If the naming was more explicit, this would read less weird:

```c
wlr_buffer_lock(&buffer->base); // lock the buffer to ensure we can read from this at any time
wlr_buffer_unref(&buffer->base); // we don't need a write-only reference to the buffer anymore
```

(And we would call `wlr_buffer_unlock` when the buffer gets replaced by another one.)

Do we want to change the naming? Do you have other ideas?

ddevault: I would prefer that language, yeah.

include/wlr/types/wlr_buffer.h
```diff
@@ -21,23 +21,51 @@ struct wlr_buffer_impl {
...
 void wlr_buffer_init(struct wlr_buffer *buffer,
-	const struct wlr_buffer_impl *impl);
+	const struct wlr_buffer_impl *impl, int width, int height);
 /**
- * Reference the buffer.
+ * Unreference the buffer. This function should be called by producers when
+ * they are done with the buffer.
  */
-struct wlr_buffer *wlr_buffer_ref(struct wlr_buffer *buffer);
+void wlr_buffer_drop(struct wlr_buffer *buffer);
```
ddevault: Release?

emersion: Conflicts with the concept of "releasing" a buffer when all consumers are done with it (wording used in the Wayland protocol: `wl_buffer.release`).

ddevault: Eh. Don't have any better ideas.

emersion: But maybe we should s/lock/acquire/ and s/unlock/release/?

emersion: Also, chose to use "drop" instead of keeping "unref" so that old `wlr_buffer_unref` users don't forget to update their code to use `wlr_buffer_unlock`.

ddevault: Hm, okay.

ascent12:
I don't really want to turn this into bikeshedding the function names, but would something a bit more explicit about "reading" be clearer?
- `wlr_buffer_lock` => `wlr_buffer_read_lock`
- `wlr_buffer_unlock` => `wlr_buffer_read_unlock`
or something vaguely similar.

`drop` and `unlock` don't really have enough context on their own to say who is supposed to use what.

emersion:
Actually, locking is necessary for write access too. In the case of a swapchain:

- The swapchain holds the buffers, and needs to know when the buffers can be re-used
- The renderer writes to the buffers
- The buffers should be locked while being rendered to, so that they don't get re-used

ascent12:
Yeah, that makes sense.

include/wlr/types/wlr_buffer.h
ascent12: Is this function really necessary? There are only 2 potential users of this function, which is wlr_client_buffer, and a swapchain.
wlr_client_buffer doesn't need it, because it can just destroy the buffer on the release signal.
The swapchain can just implement this itself, taking the release event into account.

emersion: Do you mean that we can replace `wlr_buffer_drop` with `wlr_buffer_destroy`?

This is necessary when e.g. resizing a swapchain: the old buffers (with the old size) can be dropped by the swapchain while still being used by e.g. DRM for scanout.

emersion: And in `wlr_surface`, we'd need to have a list of buffers the compositor is still using and a bunch of release listeners to only destroy buffers when we can.

emersion:
@ascent12 Ping? Any better naming ideas? (e.g. lock → acquire / unlock → release?)

include/wlr/types/wlr_buffer.h
ascent12: >This is necessary when e.g. resizing a swapchain: the old buffers (with the old size) can be dropped by the swapchain while still being used by e.g. DRM for scanout.

My point is that you'd still keep roughly the same logic, but it would be implement in the swapchain rather than in `wlr_buffer`.

>And in wlr_surface, we'd need to have a list of buffers the compositor is still using and a bunch of release listeners to only destroy buffers when we can.

Isn't that just done by the reference counting?

emersion: >My point is that you'd still keep roughly the same logic, but it would be implement in the swapchain rather than in wlr_buffer.

If the swapchain has 3 slots and is resized, there are 3 "dangling buffers" still used by DRM that will need to be destroyed when released.

>Isn't that just done by the reference counting?

What do you suggest actually? I realize I may have misunderstood.

Do you want to remove `wlr_buffer_drop` and replace it with a `wlr_buffer_destroy` function which just immediately destroys the buffer? If that's the case, someone needs to keep track of the release event and destroy the buffer when it happens. I don't want to duplicate this logic in `wlr_surface` and in the swapchain.

emersion:
Bump

emersion:
So… What do we want to do with this pull request? Are there any actionable tasks to unblock it?

Issue #1925, opened 2019-11-19
wlr_output's design doesn't work well with repaint scheduling


emersion:
`wlr_output` has a baked-in assumption that the compositor will render right away when the `frame` event is triggered.

If the compositor doesn't do this, then `wlr_output` assumes the compositor stopped its rendering loop. Some wlroots utilities (`wlr_output_damage`, `wlr_screencopy_v1`) need to resume the rendering loop, and will call `wlr_output_schedule_frame` to do so. This function will re-submit the previous frame (perform a page-flip on DRM) to request a new `frame` event. We don't resume the rendering loop immediately to make sure we're correctly aligned with vsync.

All of this will prevent the compositor from submitting a new frame some time after the `frame` event.

To implement repaint scheduling in Sway, `wlr_output.block_idle_frame` has been introduced. It's basically a way for the compositor to say "i'll submit a frame soon, so please ignore any `wlr_output_schedule_frame` call".

I'm not a fan of the current situation because the `wlr_output` interface has become too complicated and opinionated. Removing `wlr_output.block_idle_frame` is not trivial, see https://github.com/swaywm/sway/pull/4588#discussion_r347147171.

I'll try to think of a better solution, with repaint scheduling + variable refresh rate in mind.

ascent12:
I wonder if it's worth going with the `wl_surface` model for this. We've already gone towards that with most things in `wlr_output`, but it could be frame events too.
Instead of the backend dictating the user's rendering loop, the user would have a more active role in driving it. I honestly think we should drop all of this  `needs_frame`, `frame_pending`, `idle_frame` crap.

Although one foreseeable issue is restarting the rendering loop on VT-switching, but that has always been awkward.

emersion:
One issue is that it's not possible for DRM to implement a `wl_surface` model, because once you've submitted a frame you need to wait vsync before submitting another one. That's why `frame_pending` exists: to prevent users from pushing multiple frames per vblank so that all backends work the same way.

We could work-around this by making the DRM backend wait on `wlr_output_commit` and submit the frame right before vblank, however I'm not sure how this would work with variable refresh rate and how to make sure we don't miss a frame.

`needs_frame` allows the output to request a frame event if there's no damage (so the compositor would skip rendering normally). This is useful in a handful of situations, from VT switching to screen capture. I don't have an idea to remove it (yet).

`idle_frame` is tied to `wlr_output_schedule_frame`, so it would go away with this function.

ascent12:
[...]

PR #2046, opened 2020-03-02
output: remove wlr_output_impl.schedule_frame

emersion:
This function allowed backends to provide a custom function for frame
scheduling. Before resuming the rendering loop, the DRM and Wayland
backends would wait for vsync.

There isn't a clear benefit of doing this. The only upside is that we
get more stable timings: the delay between two repaints doesn't change too
much and is close to a mutliple of the refresh rate.

However this introduces latency, especially when a client misses a
frame. For instance a fullscreen game missing vblank will need to wait
more than a whole frame before being able to display new content. This
worst case scenario happens as follows:

- Client is still rendering its frame and cannot submit it in time
- Deadline is reached
- Compositor decides to stop the rendering loop since nothing changed on
  screen
- Client finally manages to render its frame, submits it
- Compositor calls wlr_output_schedule_frame
- DRM backend waits for next vblank
- The wlr_output frame event is fired, compositor draws new content on screen
- On the second next vblank, the new content reaches the screen

With this patch, the wlr_output frame event is fired immediately when
the client submits its late frame.

This change also makes it easier to support variable refresh rate, since
VRR is all about being able to present too-late frames earlier.

References: https://github.com/swaywm/wlroots/issues/1925

cc @ascent12 @YaLTeR

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -960,47 +960,6 @@ static bool drm_connector_move_cursor(struct wlr_output *output,
 	return ok;
 }
 
-static bool drm_connector_schedule_frame(struct wlr_output *output) {
-	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
-	struct wlr_drm_backend *drm = get_drm_backend_from_backend(output->backend);
-	if (!drm->session->active) {
-		return false;
-	}
-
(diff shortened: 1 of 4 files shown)


ddevault:
Thanks!

PR #2048, opened 2020-03-02
backend/wayland: fix seat caps handling

emersion:
Previously, each time a wl_seat.capabilities event was received the
Wayland backend created new input devices. It now only does so the first
time.

Input devices are now destroyed when the cap is removed.

Closes: https://github.com/swaywm/sway/issues/5055


emersion:
Confirmed working: https://github.com/swaywm/sway/issues/5055#issuecomment-593869276

ddevault:
Needs conflicts resolved

PR #2049, opened 2020-03-04
xcursor: un-hardcode /usr/share in default XCURSORPATH

jbeich:
Without this change Sway won't apply cursor theme on empty workspaces on FreeBSD (uses `/usr/local` by default). Xwayland and Gtk3 know where to find icons (for cursors and not), so hovering over applications will often use client-side cursor, hiding the issue. CC @raichoo

A similar bug exists in libwayland but not libXcursor.

### DIFF
--- a/xcursor/meson.build
+++ b/xcursor/meson.build
@@ -1,3 +1,8 @@
+prefix = get_option('prefix')
+datadir = get_option('datadir')
+
+add_project_arguments('-DDATADIR="@0@"'.format(join_paths(prefix, datadir)), language : 'c')
+
 wlr_files += files(
 	'wlr_xcursor.c',
 	'xcursor.c',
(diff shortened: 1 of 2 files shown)


jbeich:
Oops, I didn't read libwayland code carefully. It seems this can be fixed by applying https://gitlab.freedesktop.org/wayland/wayland/commit/dd8891be36ec

jbeich:
@ddevault, why did you close without leaving me ability to fix my PR?

ddevault:
Maybe I misunderstood you? You said that the bug can be fixed in libwayland-cursor.

jbeich:
IIUC, wlroots has forked XCURSORPATH code from libwayland which in turn was forked from libXcursor. libwayland and libXcursor work fine by default due to appending prefix-specific paths via ICONDIR.

jbeich:
f206c8db3ef9 is an example backport but it doesn't show up here because you've closed the PR.

ddevault:
I see. I don't seem to be able to re-open the PR, though. Did you delete your branch? You may have to open a new one.

jbeich:
> Did you delete your branch?

No.

> You may have to open a new one.

OK.

PR #2052, opened 2020-03-05
util/log: improve time prefix

emersion:
This is useful when debugging a rendering loop.

### DIFF
--- a/util/log.c
+++ b/util/log.c
@@ -11,6 +11,7 @@
 
 static bool colored = true;
 static enum wlr_log_importance log_importance = WLR_ERROR;
+static struct timespec start_time = {-1};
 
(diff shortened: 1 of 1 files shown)


emersion:
BTW, maybe we should consider dropping the current date (the YYYY-MM-DD part) to get a smaller log prefix.

ddevault:
>BTW, maybe we should consider dropping the current date (the YYYY-MM-DD part) to get a smaller log prefix.

Agreed. Maybe we should log a monotonic clock in terms of seconds since the compositor began?

emersion:
Good idea, changed the commit to do that.

ddevault:
Thanks!

Issue #2051, opened 2020-03-05
Busy loop with clients requesting frame callbacks without rendering


emersion:
Because of https://github.com/swaywm/wlroots/pull/2046 clients requesting frame callbacks without rendering at all cause a busy-loop in the compositor:

1. Client requests frame callback without rendering anything
2. Compositor resumes the rendering loop immediately
3. Compositor notices nothing changed on screen, skips the page-flip
4. Compositor sends frame callback event to client
5.  Goto (1)

This can be reproduced with the so-called "last good mpv commit" 8a83430a:

```
mpv --force-window --vid=1 --external-file=<path to image> <path to audio file>
```

Firefox also uses this behavior.

I'm putting together a test client right now.

ascent12:
I think we should just make compositors always perform `wlr_output_commit` every frame, even if it contains no state changes or new rendering. i.e. `wlr_output_commit` is the thing that will trigger the frame callback to happen.

emersion:
No, this will keep the rendering loop active even if nothing is happening (no clients requesting the frame callback).

emersion:
Test client: https://github.com/emersion/wleird/blob/master/frame-callback.c

PR #2053, opened 2020-03-05
output: make wlr_output_schedule_frame set output->needs_frame

emersion:
This way, wlr_output_schedule_frame will always be followed by a
wlr_output_commit. This forces the compositor to render an extra
frame before stopping the rendering loop.

To test, run wleird's frame-callback [1], make sure it's the only
visible client on the output and check the interval between frame
events is the output's refresh period instead of zero.

[1]: https://github.com/emersion/wleird/blob/master/frame-callback.c

Closes: https://github.com/swaywm/wlroots/issues/2051


ddevault:
Thanks!

PR #2055, opened 2020-03-06
xwayland: improve XWM atom map

emersion:
### xwayland: use explicit indexes when initializing atom_map

It's very easy to break the mapping between the atom_name enum and the
atom_map array. Use explicit indexes to prevent issues.

### xwayland: remove underscore prefix from atom names
    
Previously, some atoms had a leading underscore, others didn't. Be more
consistent and never use a leading underscore (symbols with a leading
underscore followed by an upper-case letter are reserved).


ddevault:
Thanks!

PR #2057, opened 2020-03-06
xdg_surface: send error on no role commit

### LINKED ISSUE #2056: Crash with chromium ozone
I am using sway 1.4 with wlroots 0.10.0 and I notice that wlroots crashes when chromium (v80) ozone creates too many popups.

The issue can be reproduced by creating nested bookmarks (ie: bookmark folders inside the bookmarks toolbar) and then navigating to the nested bookmark and clicking on the bookmark. This bug is not 100% reproducible (sometimes chromium will crash and sway will still work). Chromium will be running in the background after sway crashes.

Here is the [debug log](https://gist.github.com/travankor/885ffa5a938f50cf534c661d5f1a46d4) of sway when it crashes (using a vanilla config).

<details>

I apologize for this, but I cannot create a coredump of sway for some reason :/. I have coredumps of processes like redshift (minus7/redshift) and swaybg (note: these crashes happen after starting sway the second time after the initial crash so they're probably unrelated), but linux does not create a coredump of sway/wlroots (not using systemd but `kernel.core_pattern=$HOME/crash/core.%u.%e.%p` and `fs.suid_dumpable=2`.

I reproduced this bug three times and get the exact message in dmesg:
```
sway[XXXX]: segfault at 10 ip 00007f7802dda2fc sp 00007fff43855f30 error 4 in libwlroots.so.5.9.1[7f7802db4000+49000]
Code: 28 00 0f 84 e6 00 00 00 80 78 20 00 0f 84 ac 00 00 00 80 7d 48 00 74 38 80 7d 49 00 75 4f 5d c3 66 0f 1f 44 00 00 48 8b 45 30 <80> 78 10 00 0f 84 a2 00 00 00 48 8b 50 48 80 7d 48 00 48 89 90 c8
```
What [stack exchange](https://stackoverflow.com/questions/2549214/interpreting-segfault-messages) says.
```

Emantor:
Send an error to the client if it wants to commit a surface without a
role instead of crashing the compositor.

Fixes #2056

Set to draft because I have not quite figured out how this call is reached.

### DIFF
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -357,7 +357,10 @@ void handle_xdg_surface_commit(struct wlr_surface *wlr_surface) {
 
 	switch (surface->role) {
 	case WLR_XDG_SURFACE_ROLE_NONE:
-		assert(false);
+		wl_resource_post_error(surface->resource,
+			XDG_SURFACE_ERROR_NOT_CONSTRUCTED,
+			"xdg_surface must have a role");
+		return;
 	case WLR_XDG_SURFACE_ROLE_TOPLEVEL:
 		handle_xdg_surface_toplevel_committed(surface);
 		break;


emersion:
Yeah, understanding what's going on would be best.

PR #2059, opened 2020-03-09
output: always signal needs_frame if damage may have increased [was: backend/drm: damage all buffers after VT switch]

thejh:
After commit 94fa6c88f ("output: don't emit the mode event if it hasn't changed"), the entire output is directly damaged by drm_connector_set_mode() after each VT switch.
However, wlr_output_damage_whole() only damages the next buffer, not all buffers.

Because of this, after switching to a text VT and back to the one running sway, all areas that have not independently been damaged afterwards stay black in the second buffer, effectively causing those areas to flicker between the correct content and black.

Fix it by adding a new signal invalidate_buffers that drm_connector_set_mode() can use to tell the damage tracking logic that it should apply full damage to all buffers.

I have verified that this fixes the flickering I have observed on my PC; however, I'm not really familiar with the internals of things like VT switching or sway, so it might well be that the way in which I'm fixing it doesn't make much sense.

This issue reproducibly happens on my workstation, which has two displays driven by Intel integrated graphics.

Closes: https://github.com/swaywm/sway/issues/5086

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -555,7 +555,7 @@ bool set_drm_connector_gamma(struct wlr_output *output, size_t size,
 
 	bool ok = drm->iface->crtc_set_gamma(drm, conn->crtc, size, _r, _g, _b);
 	if (ok) {
-		wlr_output_update_needs_frame(output);
+		wlr_output_update_needs_frame(output, true);
 
 		free(conn->crtc->gamma_table);
 		conn->crtc->gamma_table = gamma_table;
@@ -900,7 +900,7 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 			return false;
 		}
 
-		wlr_output_update_needs_frame(output);
+		wlr_output_update_needs_frame(output, true);
 	}
 
 	if (!update_texture) {
@@ -960,7 +960,7 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 	}
 	bool ok = drm->iface->crtc_set_cursor(drm, crtc, bo);
 	if (ok) {
-		wlr_output_update_needs_frame(output);
+		wlr_output_update_needs_frame(output, true);
 	}
 	return ok;
 }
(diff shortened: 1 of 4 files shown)


emersion:
`output_handle_needs_frame` in `types/wlr_output_damage.c` should already take care of this.

thejh:
Okay... so, `wlr_output_damage_whole()` first adds the entire transformed width and height to `output->damage`, then calls `wlr_output_update_needs_frame()`. That thing is then supposed to send a signal to the damage tracking logic that will then grab the output's damage and copy it into the damage tracking logic.

But `wlr_output_update_needs_frame()` has an optimization that was added very recently in commit 7bce056f1d9490cf04166d3eb0183c9eec2204ff ("output: don't send a needs_frame event if already sent") to avoid emitting a signal again if `->needs_frame` was already set. And according to my debugging, wlr_output_schedule_frame() already called `wlr_output_update_needs_frame()` before  `wlr_output_damage_whole()` added damage to the output. So that optimization is wrong.

This bug is triggered by the code added in commit a4c7c6db097c8f493a88a1f0ef5c6f0c78f70e01 ("output: make wlr_output_schedule_frame set output->needs_frame"), which adds the call from `wlr_output_schedule_frame()` to `wlr_output_update_needs_frame()`.

So I guess the right fix might be to add a parameter to wlr_output_update_needs_frame() that tells it whether we only want to schedule output or actually added damage? I'll replace the patch in this pull request with one that does that - WDYT?

emersion:
Ah, I see. Maybe we could add a `damage` event to `wlr_output` that would be emitted when damage is added to `wlr_output.damage`? Or we could just revert the commit with the optimization, and add a comment to make sure it doesn't happen again.

thejh:
If you simply revert the optimization, you get infinite recursion, right?

But since it sounds like you have an idea of how you'd prefer the fix to look, do you want to implement it?

emersion:
>I think you get infinite recursion?

`needs_frame` triggers `wlr_output_schedule_frame` in `wlr_output_damage`, but not the other way around? I agree it's a bit confusing, hence the `damage` signal suggestion.

>Since it sounds like you have an idea of how you'd prefer the fix to look, do you want to implement it?

Honestly I'm fine with either solution and we always prefer contributors to send a patch. :)

PR #2060, opened 2020-03-09
render/gles2: remove duplicated format list

emersion:


ddevault:
NACK. Incurs a greater runtime cost for trivial benefits. Not worth the code churn.

emersion:
`get_gles2_wl_formats` is called once to initialize `wl_shm`. I did this because adding new formats is annoying.

ddevault:
Adding a single line to a list is annoying? Hardly.

emersion:
Causes issues when you want to make a format depend on the availability of a GL extension. Also error-prone, I added the format but it wouldn't show up in `weston-info`.

ddevault:
Eh, fair enough then.

ddevault:
Thanks!

PR #2063, opened 2020-03-10
backend/headless: use FBOs instead of pbuffers

emersion:
Allows the headless backend to use another backend's renderer.

### DIFF
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -62,7 +62,9 @@ static void backend_destroy(struct wlr_backend *wlr_backend) {
 	wlr_signal_emit_safe(&wlr_backend->events.destroy, backend);
 
 	wlr_renderer_destroy(backend->renderer);
-	wlr_egl_finish(&backend->egl);
+	if (backend->egl == &backend->priv_egl) {
+		wlr_egl_finish(&backend->priv_egl);
+	}
 	free(backend);
 }
 
@@ -85,41 +87,15 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	backend_destroy(&backend->backend);
 }
 
-struct wlr_backend *wlr_headless_backend_create(struct wl_display *display,
-		wlr_renderer_create_func_t create_renderer_func) {
-	wlr_log(WLR_INFO, "Creating headless backend");
-
-	struct wlr_headless_backend *backend =
-		calloc(1, sizeof(struct wlr_headless_backend));
-	if (!backend) {
-		wlr_log(WLR_ERROR, "Failed to allocate wlr_headless_backend");
-		return NULL;
-	}
+static bool backend_init(struct wlr_headless_backend *backend,
+		struct wl_display *display, struct wlr_renderer *renderer) {
 	wlr_backend_init(&backend->backend, &backend_impl);
 	backend->display = display;
 	wl_list_init(&backend->outputs);
 	wl_list_init(&backend->input_devices);
 
-	static const EGLint config_attribs[] = {
-		EGL_SURFACE_TYPE, 0,
-		EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
-		EGL_BLUE_SIZE, 1,
-		EGL_GREEN_SIZE, 1,
-		EGL_RED_SIZE, 1,
-		EGL_NONE,
-	};
-
-	if (!create_renderer_func) {
-		create_renderer_func = wlr_renderer_autocreate;
-	}
-
-	backend->renderer = create_renderer_func(&backend->egl,
-		EGL_PLATFORM_SURFACELESS_MESA, EGL_DEFAULT_DISPLAY,
-		(EGLint*)config_attribs, 0);
-	if (!backend->renderer) {
-		wlr_log(WLR_ERROR, "Failed to create renderer");
-		goto error_backend;
-	}
+	backend->renderer = renderer;
+	backend->egl = wlr_gles2_renderer_get_egl(renderer);
 
 	if (wlr_gles2_renderer_check_ext(backend->renderer, "GL_OES_rgb8_rgba8") ||
 			wlr_gles2_renderer_check_ext(backend->renderer,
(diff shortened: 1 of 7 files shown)


ammen99:
Shouldn't wlroots also offer a way to get the FB to render to? For example, Wayfire also switches between various FBs when rendering. Until now, I have always tried to reset to FB 0 to render the the output, which seems to not be the case anymore.

ammen99:
Oh and by the way, doesn't this make the headless backend dependent on a GLES renderer? From my understanding, this wasn't the case before, but I am not very familiar with EGL.

emersion:
>Shouldn't wlroots also offer a way to get the FB to render to? For example, Wayfire also switches between various FBs when rendering. Until now, I have always tried to reset to FB 0 to render the the output, which seems to not be the case anymore.

If you switch between FBs, you should save the previous value with `glGet(GL_FRAMEBUFFER_BINDING)` and restore it afterwards.

>Oh and by the way, doesn't this make the headless backend dependent on a GLES renderer? From my understanding, this wasn't the case before, but I am not very familiar with EGL.

Yes, that's true. Could be fixed by using `eglGetProcAddress` for all GL functions.

ammen99:
> Yes, that's true. Could be fixed by using `eglGetProcAddress` for all GL functions.

My understanding was that we want to make it possible to write a pixman-based renderer or something :P

emersion:
>My understanding was that we want to make it possible to write a pixman-based renderer or something :P (note I do not care about this case, just pointing it out since somebody does)

Yeah, I care about this, but I don't think it's related. Currently the headless backend depends on EGL, so this PR doesn't change a lot. With renderer v6 the compositor will manage swapchains directly (or indirectly via helpers) so the backend won't even care.

(My `eglGetProcAddress` comment was about desktop OpenGL.)

any1:
CPU load due to `glReadPixels()` is doubled when running `weston-simple-egl` in wayvnc after this change is applied. Maybe a different pixel format would give better performance?

emersion:
Can you try `GL_RGBA4` and `GL_RGBA8_OES` instead of `GL_RGB8_OES`?

any1:
Yes, using RGBA8_OES performs much better. GL_RGBA4 actually performs much worse.

emersion:
This is now ready for review.

(Note that this PR depends on another PR, so I'll keep it as a draft to prevent it from being merged.)

backend/headless/backend.c
```diff
@@ -84,6 +85,43 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
...
+	wl_list_init(&backend->input_devices);
+
+	backend->renderer = renderer;
+	backend->egl = wlr_gles2_renderer_get_egl(renderer);
+
+	if (!wlr_egl_make_current(backend->egl, EGL_NO_SURFACE, NULL)) {
+		wlr_log(WLR_ERROR, "wlr_egl_make_current failed");
+		return false;
+	}
+	if (!check_rgba8_ext()) {
```
any1, line 115: Maybe fall back on a less efficient pixel format if the "good one" doesn't work? It's worse for `glReadPixels()` to not have the right format, but it's even worse if it doesn't work at all.

any1:
I tested this. It works well and there are no obvious regressions.

emersion:
Rebased and added a fallback for the internal format.

any1:
The fallback is perhaps not terribly important, but might I suggest RGB565 instead?

backend/headless/backend.c
```diff
@@ -84,6 +87,44 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	backend_destroy(&backend->backend);
 }
 
+static bool check_rgba8_ext(void) {
+	const char *exts_str = (const char *)glGetString(GL_EXTENSIONS);
+	if (exts_str == NULL) {
+		wlr_log(WLR_ERROR, "glGetString(GL_EXTENSIONS) failed");
+		return false;
+	}
+	return strstr(exts_str, "GL_OES_rgb8_rgba8") != NULL;
```
ddevault, line 96: Is... is this really how we're checking for GL extensions now?

backend/headless/backend.c
```diff
@@ -84,6 +87,44 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
...
+	if (!wlr_egl_make_current(backend->egl, EGL_NO_SURFACE, NULL)) {
+		wlr_log(WLR_ERROR, "wlr_egl_make_current failed");
+		return false;
+	}
+
+	if (check_rgba8_ext()) {
+		backend->internal_format = GL_RGBA8_OES;
+	} else {
+		wlr_log(WLR_INFO, "GL_OES_rgb8_rgba8 not supported, "
+			"falling back to GL_RGBA4 internal format");
```
ddevault, line 118: WLR_INFO is user-visible, we should give them more information about what this means

ddevault:
Overall, this LGTM

backend/headless/backend.c
progandy: It looks like ```RGBA8_OES``` is mandatory when ```GL_OES_required_internalformat"``` is supported. ```OES_rgb8_rgba8``` only limits the availability of ```RGB8```.

https://www.khronos.org/registry/OpenGL/extensions/OES/OES_required_internalformat.txt

emersion: >Is... is this really how we're checking for GL extensions now?

I guess I'll add a `wlr_gles2` helper function for this.

>It looks like RGBA8_OES is mandatory when GL_OES_required_internalformat" is supported. OES_rgb8_rgba8 only limits the availability of RGB8.

I don't get it. We only care about RGBA8_OES, which is defined in OES_rgb8_rgba8.

progandy: > I don't get it. We only care about RGBA8_OES, which is defined in OES_rgb8_rgba8.

OES_required_internalformat defines ```RGBA8_OES``` as well, so it could be usable even if OES_rgb8_rgba8 is not supported.

emersion: Good point. Added all of these.

emersion:
Comments addressed.

ddevault:
Thanks!

PR #2065, opened 2020-03-12
output: replace wlr_output.damage with a damage event

emersion:
This patch disambiguates the needs_frame event by uncoupling it from
damage. A new separate damage event is emitted when the backend damages
the output (this happens e.g. VT is changed or software cursors are
used). The event specifies the damaged region.

The wlr_output.damage field is removed. wlr_output is no longer
responsible for tracking its own damage, this is wlr_output_damage's
job.

This is a breaking change, but wlr_output_damage users shouldn't need an
update.

Bugs fixed:

- Screen flashes on VT switch
- Cursor damage issues on the X11 and headless backends

Closes: https://github.com/swaywm/sway/issues/5086
References: https://github.com/swaywm/wlroots/pull/2059

cc @soreau @any1 @thejh @kchibisov 

* * *

Breaking change: `wlr_output.damage` is removed, listen to `wlr_output.events.damage` instead.

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -138,7 +138,6 @@ struct wlr_output {
 
 	bool needs_frame;
 	// damage for cursors and fullscreen surface, in output-local coordinates
-	pixman_region32_t damage;
 	bool frame_pending;
 	float transform_matrix[9];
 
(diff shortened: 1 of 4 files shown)


soreau:
@emersion I tested this and it does indeed fix cursor problems on x11 backend. I also briefly tested drm and wayland backends, and didn't notice any problems. Thanks for the quick patch and cc.

any1:
This fixes the issue on the headless backend.

kchibisov:
Seems to fix swaywm/sway#5086.

ddevault:
Thanks!

PR #2066, opened 2020-03-12
output: add comment about needs_frame in wlr_output_schedule_frame

emersion:
Add a comment to not forget why this call is necessary.

References: https://github.com/swaywm/wlroots/pull/2053


ddevault:
Thanks!

PR #2067, opened 2020-03-13
Send keyboard enter/leave on capability change

ifreund:
This is more correct according to the protocol and fixes issues with
clients that wait for an enter event before processing key events.

This can be tested by focusing alacritty, switching to a tty and back, then typing.  Without this commit, input is ignored as the client is waiting for an enter event.

Fixes https://github.com/swaywm/sway/issues/5085 (though it was closed as a duplicate already)

A second PR doing the same thing for pointers is in the works.

### DIFF
--- a/types/seat/wlr_seat_keyboard.c
+++ b/types/seat/wlr_seat_keyboard.c
@@ -218,6 +218,18 @@ void wlr_seat_keyboard_send_modifiers(struct wlr_seat *seat,
 	}
 }
 
+void seat_client_send_keyboard_leave_raw(struct wlr_seat_client *seat_client,
+		struct wlr_surface *surface) {
+	uint32_t serial = wlr_seat_client_next_serial(seat_client);
+	struct wl_resource *resource;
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

types/seat/wlr_seat_keyboard.c
```diff
@@ -409,8 +414,28 @@ void seat_client_create_keyboard(struct wlr_seat_client *seat_client,
...
-	// right away
+	struct wlr_seat_client *focused_client =
+		seat_client->seat->keyboard_state.focused_client;
+	struct wlr_surface *focused_surface =
+		seat_client->seat->keyboard_state.focused_surface;
+
+	// Send an enter event if there is a focused client/surface stored
+	if (focused_client != NULL && focused_surface != NULL) {
+		struct wl_array keys;
+		wl_array_init(&keys);
```
emersion: The keys array should be populated from the keyboard state.

ifreund: Whoops, should be fixed by #2068

PR #2070, opened 2020-03-15
linux-explicit-synchronization-v1: new protocol implementation

emersion:
Depends on https://github.com/swaywm/wlroots/pull/2044 for `wlr_buffer.events.release`.

This patch adds support for the
linux-explicit-synchronization-unstable-v1 protocol. No support has been
added for sync fences yet. The protocol is already useful as-is because
it offers per-commit release events (wl_buffer.release is global state).

To test, run weston-simple-dmabuf-egl.

* * *

Since `wlr_client_buffer` are created at `wl_surface.commit` time, there's no need for any additional per-commit `wlr_surface` state.

My plan for sync fences is to add something like this to `wlr_buffer`:

```c
void wlr_buffer_set_in_fence(struct wlr_buffer *buffer, int fd);
int wlr_buffer_get_out_fence(struct wlr_buffer *buffer);
```

This allows for chaining operations:

* `zwp_linux_surface_synchronization_v1.set_acquire_fence` would call `wlr_buffer_set_in_fence`
* At render time, the compositor would wait for `wlr_buffer_get_out_fence`. After rendering, it would call `wlr_buffer_set_in_fence`
* The DRM backend would set the `IN_FENCE_FD` property to `wlr_buffer_get_out_fence`. After commit, it would call `wlr_buffer_set_in_fence` with `OUT_FENCE_PTR`
* On release, the protocol implementation would call `wlr_buffer_get_out_fence` and send `zwp_linux_buffer_release.fenced_release`.


emersion:
Yeah, see the updated PR description.

emersion:
This PR has been updated and is now functional. I'd like to hear some feedback before tackling the last remaining TODOs.

backend/drm/atomic.c
```diff
@@ -129,6 +133,12 @@ static bool atomic_crtc_pageflip(struct wlr_drm_backend *drm,
 	}
 	atomic_add(&atom, crtc->id, crtc->props.mode_id, crtc->mode_id);
 	atomic_add(&atom, crtc->id, crtc->props.active, 1);
+	if (crtc->props.out_fence_ptr) {
```
ascent12: I believe this property is guranteed for atomic drivers.

backend/drm/drm.c
```diff
@@ -390,9 +401,19 @@ static bool drm_connector_commit_buffer(struct wlr_output *output) {
 			wlr_log(WLR_ERROR, "get_fb_for_bo failed");
 			return false;
 		}
+
+		if (output->pending.buffer->in_fence_fd >= 0) {
+			in_fence_fd = dup(output->pending.buffer->in_fence_fd);
```
ascent12: Use `fcntl(fd, F_DUPFD_CLOEXEC)` instead.

emersion:
>I would prefer to go with the "better" way of implementing this, but that's probably not going to happen in the near future.

The goal of this PR is to implement the "easy" way first. Then we can build on top of it to implement the "better" way. Doing all in one go would result in a massive PR.

emersion:
A possible issue with the approach taken in this PR is that we bolt the sync fences onto the `wlr_buffer`. Maybe would be better to require compositors to *explicitly* wire up the sync fences.

emersion:
Apparently `sync_file` is old already and `drm_syncobj` is the new hotness. We'll probably want to wait for linux-explicit-synchronization-v2 before resuming this work.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2070

PR #2071, opened 2020-03-15
Fix uint32 overflow in fill_empty_gamma_table on Icelake platform

filips:
Because of the larger gamma table on Icelake platforms, there is a 32 bit overflow leading to the issue https://github.com/swaywm/sway/issues/4826

Closes https://github.com/swaywm/sway/issues/4826


MisterDA:
Hi, not sure if it is immensely relevant, but perhaps the [macro for integer constant expression](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdint.h.html#tag_13_48_03_04) `UINT64_C(0xffff)` from `stdint.h` is better style/suited than the cast `(uint64_t)0xffff`.

emersion:
Would would that macro be better than a cast?

PR #2072, opened 2020-03-16
backend/wayland: close keymap FD

emersion:
We don't actually need the keymap. We need to close the FD or we will
run out of FDs.

References: https://github.com/swaywm/sway/issues/5115


ddevault:
Thanks!

PR #2080, opened 2020-03-23
Return failure of wlr_renderer_init_wl_display()

ifreund:
This makes it easier for the user of this library to properly handle
failure of this function.

### DIFF
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -160,18 +160,18 @@ bool wlr_renderer_format_supported(struct wlr_renderer *r,
 	return r->impl->format_supported(r, fmt);
 }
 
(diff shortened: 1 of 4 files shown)


render/wlr_renderer.c
```diff
@@ -189,6 +189,8 @@ void wlr_renderer_init_wl_display(struct wlr_renderer *r,
 	if (r->impl->init_wl_display) {
 		r->impl->init_wl_display(r, wl_display);
```
emersion: We should probably also make `impl->init_wl_display` return a `bool`.

ifreund: Thanks, done in the new commit.

emersion:
Nice. Can you squash these commits together?

ifreund:
Squashed!

PR #2081, opened 2020-03-23
Return false on wlr_keyboard_set_keymap() failure

ifreund:
This allows users of the library to handle or ignore the error as they
see fit.


ifreund:
That was a quick merge, no problem!

PR #2082, opened 2020-03-24
build: Add 'auto' to logind-provider combo option

soreau:
The logind provider defaulted to systemd and in order to use elogind,
-Dlogin-provider=elogind was required. This adds 'auto' as a choice
for the login-provider option and sets it as default. Using 'auto',
the build will check for systemd first and if it's not found, try
to find and use elogind automatically.

### DIFF
--- a/backend/session/meson.build
+++ b/backend/session/meson.build
@@ -21,15 +21,46 @@ if not get_option('logind').disabled()
 	msg += 'You may need to pass "-Dlogind-provider=elogind" or "-Dlogind-provider=systemd" to ensure the correct library is detected.'
 endif
 
-logind = dependency('lib' + get_option('logind-provider'),
-	required: get_option('logind'),
-	not_found_message: '\n'.join(msg).format(get_option('logind-provider')),
-	version: '>=237',
-)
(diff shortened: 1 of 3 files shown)


backend/session/meson.build
```diff
@@ -21,15 +21,41 @@ if not get_option('logind').disabled()
...
-	not_found_message: '\n'.join(msg).format(get_option('logind-provider')),
-	version: '>=237',
-)
+logind_version = '>=237'
+
+if get_option('logind-provider') == 'auto'
+	assert(get_option('auto_features').auto(), '-Dlogind-provider must be set to systemd or elogind since auto_features != auto')
+	logind = dependency('libsystemd',
+		required: get_option('logind'),
+		not_found_message: 'libsystemd not found, trying libelogind',
```
ascent12: Can you change these `not_found_message`s to match the non-auto case? I wanted them to describe what arguments should be passed and what they're for.

soreau: Yes, they're added now.

ddevault:
Thanks!

Issue #2076, opened 2020-03-20
linux-dmabuf extension advertized when renderer doesn't support DMA-BUF


emersion:
We enable it as long as the renderer has the required hook, but we don't check whether the EGL extension is present.

https://github.com/swaywm/wlroots/blob/b614ded3fcb022c4567bbee932a6df3e83acaff6/render/wlr_renderer.c#L186

PR #2084, opened 2020-03-24
render: only expose linux-dmabuf if EGL extension is supported

emersion:
Only expose linux-dmabuf extension if EGL_EXT_image_dmabuf_import_ext is
supported.

Closes: https://github.com/swaywm/wlroots/issues/2076

### DIFF
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -355,9 +355,14 @@ static bool gles2_init_wl_display(struct wlr_renderer *wlr_renderer,
 		struct wl_display *wl_display) {
 	struct wlr_gles2_renderer *renderer =
 		gles2_get_renderer(wlr_renderer);
-	if (!wlr_egl_bind_display(renderer->egl, wl_display)) {
-		wlr_log(WLR_INFO, "failed to bind wl_display to EGL");
-		return false;
+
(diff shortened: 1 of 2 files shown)


render/gles2/renderer.c
```diff
@@ -355,10 +356,24 @@ static bool gles2_init_wl_display(struct wlr_renderer *wlr_renderer,
 		struct wl_display *wl_display) {
 	struct wlr_gles2_renderer *renderer =
 		gles2_get_renderer(wlr_renderer);
-	if (!wlr_egl_bind_display(renderer->egl, wl_display)) {
-		wlr_log(WLR_INFO, "failed to bind wl_display to EGL");
-		return false;
+
+	if (renderer->egl->exts.bind_wayland_display_wl) {
```
Emantor: Checking for BINd_WAYLAND_DISPLAY_WL is a separate change or should at least be mentioned in the commit message.

emersion: The check was performed in `wlr_egl_bind_display` anyway before this commit.

Emantor: And since both are public API, its probably best to leave the check there too.

emersion: Yes.

I've split the changes into two separate commits.

PR #2085, opened 2020-03-25
Prevent memory leak in copypaste of the screencopy example

rabits:
If someone want to use screencopy as is processing multiple screenshots - it could be hard to find this issue with SHM. It's better to make sure it's creating just once - because the SHM file will not be removed from the shared memory until the application is terminated.

### DIFF
--- a/examples/screencopy.c
+++ b/examples/screencopy.c
@@ -35,6 +35,7 @@
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <assert.h>
 #include <wayland-client-protocol.h>
 #include "wlr-screencopy-unstable-v1-client-protocol.h"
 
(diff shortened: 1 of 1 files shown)


progandy:
I don't think this condition makes much sense without code to handle differing width and height on reuse. A combination of ```wl_buffer_destroy(buffer->wl_buffer); munmap(buffer->data)``` should free the buffer.

emersion:
The screencopy example is one-off, so I don't think it makes sense to half-add support for multiple captures. If you want, we can add an `assert` instead.

ddevault:
+1 for an assert

rabits:
I see, the buffer is created for a specific resolution so it's really not a good idea if the capture parameters are changing, right. Ok, assert will clearly say that the buffer is already allocated and should be cleaned before allocating the new one.

PR #2090, opened 2020-03-29
drm: Use DRM_FORMAT_MOD_LINEAR when no other format modifiers are present

agx:
The absence of modifiers means "assume linear" not "don't use any
modifier". This prevents gallium from using plain resource_create but
rather resource_create_with_modifiers.

This unbreaks broken rendering on etnaviv as resource_create assumes
it can supertile (which makes sense since e.g. dcss and ipuv3 can
handle these formats). However lcdif/mxsfb can't handle this so
be explicit about the modifier.

Weston does the same, see drm_plane_populate_formats().

This is basically a safety net if the display controller does not
announce supported modifiers. We'll fix this as well in the kernel but having
good default is useful too.

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -90,6 +90,10 @@ bool init_drm_surface(struct wlr_drm_surface *surf,
 		if (drm_format != NULL) {
 			surf->gbm = gbm_surface_create_with_modifiers(renderer->gbm,
 				width, height, format, drm_format->modifiers, drm_format->len);
+		} else {
+			uint64_t formats[] = { DRM_FORMAT_MOD_LINEAR };
+			surf->gbm = gbm_surface_create_with_modifiers(renderer->gbm,
+				width, height, format, formats, 1);
 		}
 	}
 


emersion:
>The absence of modifiers means "assume linear" not "don't use any
modifier".

I dont think this is correct? Doesnt it mean "use an implicit modifier"?

emersion:
Related: https://lists.freedesktop.org/archives/dri-devel/2020-March/260542.html

agx:
> I dont think this is correct? Doesnt it mean "use an implicit modifier"?

Do you have a reference? If so how would the GPU figure out what that is so it uses a modifier suitable for the display controller?

 Weston assumes `DRM_MOD_LINEAR` and that looks like a conservative approach to guard against incomplete modifier/format information:

https://gitlab.freedesktop.org/wayland/weston/-/blob/master/libweston/backend-drm/kms.c#L490

I've also submitted https://lore.kernel.org/linux-arm-kernel/26877532e272c12a74c33188e2a72abafc9a2e1c.1584973664.git.agx@sigxcpu.org/

which doesn't fix the whole issue though since we assume ARGB8888 support as well which e.g. isn't true for mxsfb atm (although there's patches now to fix that too). Given that situation (which likely applies to other non PC platforms) i figured it'd be best to be conservative here.

emersion:
>If so how would the GPU figure out what that is so it uses a modifier suitable for the display controller?

The usage flags with `GBM_BO_USE_SCANOUT` are supposed to make this work.

agx:
> The usage flags with GBM_BO_USE_SCANOUT are supposed to make this work.

Good point, i think i can see where that gets lost now.

emersion:
Also, see this discussion which is about adding the `usage` flags to `gbm_bo_create_with_modifiers`: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/3197

emersion:
@agx Have you fixed this issue in Mesa instead of wlroots?

agx:
no, i did not get a chance to look at this again.

agx:
Sorry for the delay. 

I think https://lore.kernel.org/dri-devel/26877532e272c12a74c33188e2a72abafc9a2e1c.1584973664.git.agx@sigxcpu.org/ 

fixes that (totally forgot about that patch)

Issue #633, opened 2018-02-14
Add viewporter protocol


emersion:
Source: https://github.com/wayland-project/wayland-protocols/blob/master/stable/viewporter/viewporter.xml

Test clients:
* GLFW
* gstreamer

emersion:
Related: with this MR, xwayland will be able to render correctly games that rely on changing the video mode: https://gitlab.freedesktop.org/xorg/xserver/merge_requests/270

PR #2092, opened 2020-03-31
Implement the viewporter protocol

emersion:
The [viewporter protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/blob/master/stable/viewporter/viewporter.xml) allows clients to ask the compositor to crop and scale buffers during composition. This allows clients to avoid having to crop and scale in an intermediary buffer. The use-cases are:

- Xwayland games that expect to be able to perform modeset
- Video players that present DMA-BUFs coming out of VA-API
- Optimized scrolling in web browsers

TODO:

- [x] Apply viewport transforms to damage
- [x] Error out when src_width or src_height are not integers and destination size is not set
- [x] Compositor implementation: https://github.com/swaywm/sway/pull/5433

Test plan:

- [x] `weston-scaler -s`
- [x] `weston-scaler -d`
- [x] `weston-scaler -b`

Future plans: wire this up with direct scan-out. With my [wlr_output_layer proposal](https://github.com/swaywm/wlroots/pull/1985), I think we can extend [the API](https://github.com/swaywm/wlroots/pull/1985/files#diff-bc323577ae8471cb9563146f750574a7) to allow compositors to feed src/dst rectangles to the backend:

```c
/**
 * Set the layer's final size. If the size is different from the attached buffer's,
 * the buffer will be scaled.
 */
void wlr_output_layer_scale(struct wlr_output_layer *layer, int width, int height);
/**
 * Set the layer's source box. This defines what area of the buffer will be used
 * as a source. This can be used to crop the buffer.
 */
void wlr_output_layer_set_src_box(struct wlr_output_layer *layer,
        const struct wlr_fbox *box);
```

`wlr_output_layer_scale` maps directly to the `CRTC_W` and `CRTC_H` KMS properties. `wlr_output_layer_set_buffer_src_box` maps directly to the `SRC_*` KMS properties.

Closes: https://github.com/swaywm/wlroots/issues/633

* * *

Breaking change: renderer implementations need to be updated, `render_texture_with_matrix` has been replaced with `render_subtexture_with_matrix` which allows specifying a source rectangle.

### DIFF
--- a/include/wlr/render/wlr_renderer.h
+++ b/include/wlr/render/wlr_renderer.h
@@ -55,6 +55,13 @@ bool wlr_render_texture(struct wlr_renderer *r, struct wlr_texture *texture,
 bool wlr_render_texture_with_matrix(struct wlr_renderer *r,
 	struct wlr_texture *texture, const float matrix[static 9], float alpha);
 /**
+ * Renders the requested texture using the provided matrix, after cropping it
+ * to the provided rectangle.
+ */
+bool wlr_render_subtexture_with_matrix(struct wlr_renderer *r,
+	struct wlr_texture *texture, const struct wlr_fbox *box,
+	const float matrix[static 9], float alpha);
+/**
  * Renders a solid rectangle in the specified color.
  */
 void wlr_render_rect(struct wlr_renderer *r, const struct wlr_box *box,
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -139,6 +139,30 @@ static void surface_set_input_region(struct wl_client *client,
 	}
 }
 
+/**
+ * Computes the surface viewport source size, ie. the size after applying the
+ * surface's scale, transform and cropping (via the viewport's source
+ * rectangle) but before applying the viewport scaling (via the viewport's
+ * destination rectangle).
+ */
+static void surface_state_viewport_src_size(struct wlr_surface_state *state,
+		int *out_width, int *out_height) {
+	if (state->viewport.has_src) {
+		*out_width = state->viewport.src.width;
+		*out_height = state->viewport.src.height;
+	} else {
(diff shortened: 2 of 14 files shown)


types/wlr_surface.c
```diff
@@ -149,15 +149,20 @@ static void surface_state_finalize(struct wlr_surface *surface,
...
-	int width = state->buffer_width / state->scale;
-	int height = state->buffer_height / state->scale;
-	if ((state->transform & WL_OUTPUT_TRANSFORM_90) != 0) {
-		int tmp = width;
-		width = height;
-		height = tmp;
+	if (state->viewport.has_dst && state->buffer_resource != NULL) {
+		state->width = state->viewport.dst_width;
+		state->height = state->viewport.dst_height;
+	} else {
```
zzag, line 155: If the surface is only cropped, then the surface width and the surface height should be `state->viewport.src_width` and `state->viewport.src_height`, respectively.

include/wlr/render/wlr_renderer.h
```diff
@@ -32,6 +32,11 @@ struct wlr_renderer {
 	} events;
 };
 
+struct wlr_texture_box {
```
emersion, line 35: This isn't a very good name. Maybe something like `wlr_fbox` would be better?

emersion:
This is now ready for review.

ddevault:
On a first pass, this LGTM. Would like @ascent12's thoughts and pending some discussions in IRC.

emersion:
Removed the "breaking" label, this isn't a breaking change.

Update to use `wlr_fbox`.

emersion:
Ah, this is a breaking change for custom renderers. My bad.

emersion:
Ping @ascent12

ddevault:
Thanks!

PR #2097, opened 2020-04-02
output: introduce wlr_output_test

emersion:
This PR adds `wlr_output_test`, which can be used to check whether pending output changes would be accepted by the backend. This allows compositors to try different changes before attempting a commit.

Right now this function just performs the check previously done in `attach_buffer` and `commit`. This change allows test-only commits to be run in `wlr_output_test`, which is necessary for using libliftoff (thus necessary for DRM support of output layers introduced in https://github.com/swaywm/wlroots/pull/1985).

Compositors implementing wlr-output-management can now call this function instead of returning true unconditionally.

Should `wlr_output_test` be implicitly called in `wlr_output_commit`? This would add a guarantee for backends that the output changes are valid when their `commit` function is called.

Sway PR: https://github.com/swaywm/sway/pull/5188

* * *

Breaking changes:

* For compositors: `wlr_output_attach_buffer` doesn't return a `bool` anymore. The checks previously performed in this function are now performed in `wlr_output_commit` and the new `wlr_output_test` function.
* For backends: `wlr_output_impl.attach_buffer` is gone, buffer checks now need to be performed in `commit`. A new `test` function has been added.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -334,6 +334,10 @@ static bool drm_connector_attach_render(struct wlr_output *output,
 	return make_drm_surface_current(&conn->crtc->primary->surf, buffer_age);
 }
 
+static bool drm_connector_test(struct wlr_output *output) {
+	return true;
+}
+
 static bool drm_connector_commit_buffer(struct wlr_output *output) {
 	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
 	struct wlr_drm_backend *drm = get_drm_backend_from_backend(output->backend);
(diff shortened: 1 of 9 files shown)


ammen99:
This should also be used when implementing output-management, right?

emersion:
Indeed!

emersion:
Sway pull request: https://github.com/swaywm/sway/pull/5188

ddevault:
Still miffed that we have to brute force testing configurations to find the subset which works... but thanks!

Issue #1963, opened 2019-12-22
build: convert protocol variables to dicts


emersion:
>I'd prefer to have a dict instead (e.g. protocols_code.get(name) would return the file for the generated code for instance).

That's actually a pretty good idea. Unfortunately meson (0.52.1) can't use a non string literal for a dictionary key:
```python
protocol_code = {}
foreach proto, path : protocols
    # ...
    protocol_code += { proto: code } # This is invalid
endforeach
```
And due to the way dictionaries work (i.e. they're pure) and some dictionary-of-dictionaries type stuff, I can't really think of a way to work around this (e.g. putting it in the protocols dictionary).

But this will be fixed in 0.53.0: https://mesonbuild.com/Syntax.html#dictionaries
It has been left as set_variable/get_variable currently, but changed to use an array of protocol strings in the wayland backend.
I think this should be revisited when the next meson version is out and we eventually target it for whatever reason.

_Originally posted by @ascent12 in https://github.com/swaywm/wlroots/pull/1931_

emersion:
When we bump to 0.53, we'll also be able to use `summary` instead of `message`.

PR #2101, opened 2020-04-04
build: use meson.override_dependency

emersion:
When built as a subproject, this removes the need for the parent project
to know about the dependency variable name.

This requires Meson 0.54.0.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
 	'c',
 	version: '0.10.0',
 	license: 'MIT',
-	meson_version: '>=0.51.2',
+	meson_version: '>=0.54.0',
 	default_options: [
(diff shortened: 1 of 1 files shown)


jbeich:
I've [filed bug](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=245420) to update FreeBSD package.

jbeich:
Can you refresh build? https://pkg.freebsd.org/FreeBSD:12:amd64/latest/All/meson-0.54.0.txz

emersion:
When this is merged, we'll be able to do https://github.com/swaywm/wlroots/issues/1963 as well.

ddevault:
Thanks!

PR #2104, opened 2020-04-06
Reset EGL current surface after output commit/rollback

emersion:
Most of the pending output state is not forwarded to the backend prior
to an output commit. For instance, wlr_output_set_mode just stashes the
mode without calling any wlr_output_impl function.
wlr_output_impl.commit is responsible for applying the pending mode.

However, there are exceptions to this rule. The first one is
wlr_output_attach_render. It won't go away before renderer v6 is
complete, because it needs to set the current EGL surface.

The second one is wlr_output_attach_buffer.
wlr_output_impl.attach_buffer is removed in [1].

When wlr_output_rollback is called, all pending state is supposed to be
cleared. This works for all the state except the two exceptions
mentionned above. To fix this, introduce wlr_output_impl.rollback.

Right now, the backend resets the current EGL surface. This prevents GL
commands from affecting the output after wlr_output_rollback.

This patch is required for FBO-based outputs to work properly. The
compositor might be using FBOs for its own purposes [2], having leftover
FBO state can have bad consequences.

[1]: https://github.com/swaywm/wlroots/pull/2097
[2]: https://github.com/swaywm/wlroots/pull/2063#issuecomment-597614312

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -574,6 +574,11 @@ static bool drm_connector_commit(struct wlr_output *output) {
 	return true;
 }
 
+static void drm_connector_rollback(struct wlr_output *output) {
+	struct wlr_drm_backend *drm = get_drm_backend_from_backend(output->backend);
+	wlr_egl_make_current(&drm->renderer.egl, EGL_NO_SURFACE, NULL);
+}
+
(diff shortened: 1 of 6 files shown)


emersion:
Rebased against master.

ddevault:
Thanks!

PR #2107, opened 2020-04-10
output: fix blurred hw cursors with fractional scaling

emersion:
The scaling factor was being implicitly cast to an int.

Closes: https://github.com/swaywm/sway/issues/4927

* * *

Breaking change for backends: `wlr_output_impl.set_cursor` now takes a `float scale` instead of an `int32_t`.


ddevault:
Thanks!

PR #2108, opened 2020-04-10
output: misc. wlr_output_test fixes

emersion:
## output: check for buffer size compatibility in common code 

Instead of checking for buffer size compatibility in each backend,
centralize the check in wlr_output itself.

## backend/wayland: check scan-out buffer is compatible in output_test

If the buffer doesn't have a supported format/modifier, make the test
fail.


ddevault:
Thanks!

PR #2109, opened 2020-04-10
output: fix maybe-uninitialized warning

emersion:
GCC is complaining about a maybe-uninitialized variable when doing a
release build. Even if that can't actually happen because all enum
values are handled, add an abort call to silence the warning.


ddevault:
Thanks!

Issue #2073, opened 2020-03-19
Segfault with keyboard unplugged


aidanharris:
```
Core was generated by `sway -d'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  seat_client_create_keyboard (id=37, version=<optimized out>, seat_client=<optimized out>) at ../wlroots-9999/types/seat/wlr_seat_keyboard.c:425
425			size_t num_keycodes = keyboard->num_keycodes;
[Current thread is 1 (Thread 0x7fc5e4f574c0 (LWP 3704))]
(gdb) bt
#0  0x00007fc5e597f850 in seat_client_create_keyboard (id=37, version=<optimized out>, seat_client=<optimized out>) at ../wlroots-9999/types/seat/wlr_seat_keyboard.c:425
#1  0x00007fc5e597f850 in seat_handle_get_keyboard (client=<optimized out>, seat_resource=<optimized out>, id=37) at ../wlroots-9999/types/seat/wlr_seat.c:43
#2  0x00007fc5e5254bfd in ffi_call_unix64 () at /usr/src/debug/dev-libs/libffi-3.3-r1/libffi-3.3/src/x86/unix64.S:101
#3  0x00007fc5e525037a in ffi_call_int (cif=0x7ffdb6da47b0, fn=0x7fc5e597f6e0 <seat_handle_get_keyboard>, rvalue=<optimized out>, avalue=<optimized out>, closure=<optimized out>) at /usr/src/debug/dev-libs/libffi-3.3-r1/libffi-3.3/src/x86/ffi64.c:669
#4  0x00007fc5e5a07ea9 in wl_closure_invoke (closure=closure@entry=0x55ce1f4c8230, target=target@entry=0x55ce1f4cdcb0, opcode=opcode@entry=1, data=data@entry=0x55ce1f505670, flags=<optimized out>)
    at /usr/src/debug/dev-libs/wayland-9999/wayland-9999/src/connection.c:1018
#5  0x00007fc5e5a0c864 in wl_client_connection_data (fd=<optimized out>, mask=<optimized out>, data=0x55ce1f505670) at /usr/src/debug/dev-libs/wayland-9999/wayland-9999/src/wayland-server.c:432
#6  0x00007fc5e5a0d032 in wl_event_loop_dispatch (loop=0x55ce1e5fff00, timeout=timeout@entry=-1) at /usr/src/debug/dev-libs/wayland-9999/wayland-9999/src/event-loop.c:1027
[...]

PR #2110, opened 2020-04-12
Add check for NULL keyboard (fix sway #5205)

r-c-f:
Check for a NULL keyboard_state.keyboard value in
seat_client_create_keyboard() before trying to use it, as is done in
other functions like seat_client_send_repeat_info(). Prevents a segfault
in certain situations on keyboard removal, as seen in the sway issue.

Closes: https://github.com/swaywm/sway/issues/5205
Closes: https://github.com/swaywm/wlroots/issues/2073


types/seat/wlr_seat_keyboard.c
```diff
@@ -411,6 +411,8 @@ void seat_client_create_keyboard(struct wlr_seat_client *seat_client,
 	wl_list_insert(&seat_client->keyboards, wl_resource_get_link(resource));
 
 	struct wlr_keyboard *keyboard = seat_client->seat->keyboard_state.keyboard;
+	if (keyboard == NULL)
```
emersion, line 414: Style: brackets are mandatory

PR #2111, opened 2020-04-14
backend/multi: add backends at end of list

emersion:
This allows wlr_multi_for_each_backend to iterate on the backends in the
order where they have been added.


ddevault:
Thanks!

PR #2021, opened 2020-02-08
backend/drm: Introduce wlr_drm_fb

ascent12:
This is a type which manages gbm_surfaces and imported dmabufs in the same place, and makes the lifetime management between the two shared. It should lead to easier to understand code, and fewer special cases.

This also contains a fair bit of refactoring to start using this new type.

---

I was never able to reproduce @kennylevinsen's issue with weird colors/tearing during direct scanout with mpv, so I have no idea whether this actually fixes it or not. I was only really going off of a hunch. Even if it doesn't fix it, this is still a good bit of refactoring that makes the code simpler.

This PR does need more testing before it's merged. I don't really have a super reliable way to do direct scanout, and I haven't gotten around to multi-GPU testing yet.

PR #2114, opened 2020-04-14
render/drm_format_set: disallow DRM_FORMAT_INVALID

emersion:
It doesn't make sense to add DRM_FORMAT_INVALID to a format set. Adding
an assertion allows us to safely query the format set with
DRM_FORMAT_INVALID. See [1].

[1]: https://github.com/swaywm/wlroots/pull/2021#discussion_r385839668


ddevault:
Thanks!

PR #2118, opened 2020-04-15
backend/drm: introduce wlr_drm_fb

emersion:
Same as https://github.com/swaywm/wlroots/pull/2021 but rebased, review comments addressed and:

- "backend/drm: Don't allow legacy to use direct scanout" has been dropped
- FB state fields have been reworked: the pending FB is empty if no new buffer has been submitted by the compositor

My changes are in the last commit. I'll integrate them into previous commits before merging.

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -69,10 +69,27 @@ static void atomic_add(struct atomic *atom, uint32_t id, uint32_t prop, uint64_t
 	}
 }
 
-static void set_plane_props(struct atomic *atom, struct wlr_drm_plane *plane,
-		uint32_t crtc_id, uint32_t fb_id, bool set_crtc_xy) {
+static void plane_disable(struct atomic *atom, struct wlr_drm_plane *plane) {
 	uint32_t id = plane->id;
 	const union wlr_drm_plane_props *props = &plane->props;
+	atomic_add(atom, id, props->fb_id, 0);
+	atomic_add(atom, id, props->crtc_id, 0);
+}
+
+static void set_plane_props(struct atomic *atom, struct wlr_drm_backend *drm,
+		struct wlr_drm_plane *plane, uint32_t crtc_id, int32_t x, int32_t y) {
+	uint32_t id = plane->id;
+	const union wlr_drm_plane_props *props = &plane->props;
+	struct wlr_drm_fb *fb = plane_get_next_fb(plane);
+	struct gbm_bo *bo = drm_fb_acquire(fb, drm, &plane->mgpu_surf);
+	if (!bo) {
+		goto error;
+	}
+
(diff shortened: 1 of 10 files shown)


kennylevinsen:
Confirmed to fix the tearing issue mentioned in #2021.

My original mpv repro no longer works on any revision of wlroots, but a few fullscreen qutebrowser instances with native buffers enabled playing youtube videos seem to do the trick to recrate the issue with weird tearing on cursor movement.

If anyone would like to try, the magic qutebrowser incantation is: `qutebrowser --qt-flag ignore-gpu-blacklist --qt-flag enable-gpu-rasterization --qt-flag enable-native-gpu-memory-buffers --qt-flag num-raster-thread=4 URL_TO_YT_VIDEO_OF_YOUR_CHOICE`.

emersion:
Ping.

I've been using this for a week as a daily driver.

backend/drm/atomic.c
```diff
@@ -55,32 +55,42 @@ static bool atomic_commit(int drm_fd, struct atomic *atom,
...
 	if (!atom->failed && drmModeAtomicAddProperty(atom->req, id, prop, val) < 0) {
 		wlr_log_errno(WLR_ERROR, "Failed to add atomic DRM property");
 		atom->failed = true;
 	}
 }
 
-static void set_plane_props(struct atomic *atom, struct wlr_drm_plane *plane,
-		uint32_t crtc_id, uint32_t fb_id, bool set_crtc_xy) {
+static void plane_disable(struct atomic *atom, struct wlr_drm_plane *plane)
+{
```
ddevault, line 73: style nit

ascent12:
>"backend/drm: Don't allow legacy to use direct scanout" has been dropped

I still think it's a good idea to keep this. In the event that someone tries that, and it'll probably fail, it's just going to mess everything up.

emersion:
>it'll probably fail, it's just going to mess everything up.

It'll fail if the format/modifier is different. However wlroots compositors retry without direct scan-out in this case, so it should be fine.

I agree it's not perfect, because now that we have `wlr_output_test`, the DRM backend should perform a test-only commit and allow the compositor to not even try a regular commit. We don't do that yet.

ascent12:
>retry without direct scan-out in this case, so it should be fine.

There is potentially a visual glitch that happens here.

With legacy in general, I think it's just better to keep it down the the extremely basic rendering loop and not a lot else. It's very annoying trying to maintain legacy assumptions when you want to do more advanced features, because you risk visual glitches or completely unsynchronizing your state with the kernel.

emersion:
You're right. Added the commit back.

emersion:
Squashed the changes into previous commits. Thanks for the reviews.

PR #2120, opened 2020-04-15
logind: Avoid unnecessary ReleaseDevice calls

kennylevinsen:
Libinput releases its devices one at a time on shutdown. When using the
logind session implementation, each device closure leads to a
ReleaseDevice call. With the 28 event devices on my machine, that adds
up to ~1.4 seconds of waiting.

As all devices requested via TakeControl are automatically released when
ReleaseControl is called or when the dbus connection is terminated, we
do not need to spend time calling ReleaseDevice.

Implement a new session shutdown flag that informs the logind session
implementation that we are not interested in ReleaseDevice calls.

### DIFF
--- a/backend/multi/backend.c
+++ b/backend/multi/backend.c
@@ -49,6 +49,8 @@ static void multi_backend_destroy(struct wlr_backend *wlr_backend) {
 
 	wl_list_remove(&backend->display_destroy.link);
 
+	wlr_session_shutdown(backend->session);
+
 	struct subbackend_state *sub, *next;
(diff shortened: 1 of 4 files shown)


emersion:
Hmm, this API isn't great.

Also, maybe we can open a systemd issue to understand what's taking so long before adding this workaround?

kennylevinsen:
Superseded by https://github.com/swaywm/wlroots/pull/2121

I found that if you move the close up before ReleaseDevice, then ReleaseDevice ends up taking barely any time at all, sparing us from this API change. In fact, it's even faster than what we are doing here.

This is also more alike what Weston does.

PR #2121, opened 2020-04-15
logind: Close fd before releasing device

kennylevinsen:
This speeds up shutdown significantly, and is in line with how Weston
does it.

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -107,8 +107,10 @@ static void logind_release_device(struct wlr_session *base, int fd) {
 	if (fstat(fd, &st) < 0) {
 		wlr_log(WLR_ERROR, "Failed to stat device '%d': %s", fd,
 			strerror(errno));
+		close(fd);
 		return;
 	}
+	close(fd);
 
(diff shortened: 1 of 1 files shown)


backend/session/logind.c
```diff
@@ -104,15 +104,17 @@ static void logind_release_device(struct wlr_session *base, int fd) {
 	struct logind_session *session = logind_session_from_session(base);
 
 	struct stat st;
-	if (fstat(fd, &st) < 0) {
+	int ret = fstat(fd, &st);
+	close(fd);
+	if (ret < 0) {
 		wlr_log(WLR_ERROR, "Failed to stat device '%d': %s", fd,
 			strerror(errno));
```
emersion, line 111: Can we move the `strerror` call before `close`, to prevent it from showing an erroneous value?

kennylevinsen: Good point. Wetson has the same bug: https://gitlab.freedesktop.org/wayland/weston/-/blob/master/libweston/launcher-logind.c#L229.

PR #2133, opened 2020-04-22
Screencopy dmabuf

any1:


types/wlr_screencopy_v1.c
```diff
@@ -432,6 +467,9 @@ static void capture_output(struct wl_client *wl_client,
 	frame->box = buffer_box;
 	frame->stride = 4 * buffer_box.width; // TODO: depends on read format
 
+	// TODO: Stride is decided by the user for dmabuf, send 0 or something
+	// instead
+
```
cyanreg, line 472: Need to let the compositor know about the buffer modifiers.

any1: > Need to let the compositor know about the buffer modifiers.

Doesn't that happen in `linux-dmabuf-v1-unstable`? I assumed that modifiers would be fully abstracted away at this stage.

cyanreg: Actually it does, so its fine.
So its just a matter of changing the wording of the protocol to make stride == 0 "pick whatever".

any1: Well, the wording of the protocol is currently "The client **should** then create a buffer with the provided attributes", so it doesn't really _need_ to change, but perhaps we ought to change it for clarification. :)

render/gles2/renderer.c
```diff
@@ -331,6 +335,63 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	glGenFramebuffers(1, &fbo);
+	glBindFramebuffer(GL_FRAMEBUFFER, fbo);
+	glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
+			GL_RENDERBUFFER, rbo);
+
+	GLenum status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
+	if (status != GL_FRAMEBUFFER_COMPLETE)
+		goto failure;
+
+	float mat[9];
```
emersion, line 377: Can we add a `// TODO: use ANGLE_framebuffer_blit if available`?

include/wlr/render/interface.h
```diff
@@ -64,6 +64,9 @@ struct wlr_renderer_impl {
 	void (*destroy)(struct wlr_renderer *renderer);
 	bool (*init_wl_display)(struct wlr_renderer *renderer,
 		struct wl_display *wl_display);
+	bool (*clone_dmabuf)(struct wlr_renderer *renderer,
```
emersion, line 67: Can we rename this to `blit_dmabuf`, to be consistent with OpenGL's wording? "Clone" doesn't make it clear that this isn't about `dup`'ing the DMA-BUF FDs.

include/wlr/types/wlr_screencopy_v1.h
```diff
@@ -13,6 +13,8 @@
 #include <wayland-server-core.h>
 #include <wlr/types/wlr_box.h>
 
+struct wlr_dmabuf_v1_buffer;
```
emersion, line 16: This forward declaration shouldn't be necessary

render/gles2/renderer.c
```diff
@@ -14,6 +14,10 @@
 #include <wlr/util/log.h>
 #include "render/gles2.h"
 
+#ifndef GL_RENDERBUFFER_OES
+#define GL_RENDERBUFFER_OES 0x8D41
```
emersion, line 18: We shouldn't do this

render/gles2/renderer.c
```diff
@@ -331,6 +335,63 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
 	return glGetError() == GL_NO_ERROR;
 }
 
+static bool gles2_clone_dmabuf(struct wlr_renderer *wlr_renderer,
+		struct wlr_dmabuf_attributes *dst_attr,
+		struct wlr_dmabuf_attributes *src_attr)
+{
```
emersion, line 341: Style: `{` on the same line as the previous one

render/gles2/renderer.c
```diff
@@ -331,6 +335,63 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	glBindRenderbuffer(GL_RENDERBUFFER, 0);
+
+	GLuint fbo = 0;
+	glGenFramebuffers(1, &fbo);
+	glBindFramebuffer(GL_FRAMEBUFFER, fbo);
+	glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
+			GL_RENDERBUFFER, rbo);
+
+	GLenum status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
+	if (status != GL_FRAMEBUFFER_COMPLETE)
```
emersion, line 374: Style: braces are mandatory

render/gles2/renderer.c
```diff
@@ -603,6 +665,13 @@ struct wlr_renderer *wlr_gles2_renderer_create(struct wlr_egl *egl) {
 			"glEGLImageTargetTexture2DOES");
 	}
 
+	if (check_gl_ext(exts_str, "GL_OES_EGL_image")) {
+		renderer->exts.egl_image_oes = true;
+		load_gl_proc(&gles2_procs.glEGLImageTargetRenderbufferStorageOES,
+			"glEGLImageTargetRenderbufferStorageOES");
+
```
emersion, line 672: Style: stray newline

render/wlr_renderer.c
```diff
@@ -154,6 +154,16 @@ bool wlr_renderer_read_pixels(struct wlr_renderer *r, enum wl_shm_format fmt,
 		src_x, src_y, dst_x, dst_y, data);
 }
 
+bool wlr_renderer_clone_dmabuf(struct wlr_renderer *r,
+		struct wlr_dmabuf_attributes *dst,
+		struct wlr_dmabuf_attributes *src)
+{
```
emersion, line 160: Style: `{` should be on the same line as the previous one

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +282,39 @@ static void frame_handle_copy(struct wl_client *wl_client,
...
+		fmt = wl_shm_buffer_get_format(shm_buffer);
+		width = wl_shm_buffer_get_width(shm_buffer);
+		height = wl_shm_buffer_get_height(shm_buffer);
+		stride = wl_shm_buffer_get_stride(shm_buffer);
+	} else if(dma_buffer) {
+		width = dma_buffer->attributes.width;
+		height = dma_buffer->attributes.height;
+
+		// Doesn't matter for dmabuf
+		fmt = frame->format;
```
emersion, line 314: Should we allow the client to choose an arbitrary format here? We don't want the compositor to do arbitrary format conversions.

any1: Doesn't the format conversion happen on the GPU when it's sampling the input texture and writing it into the dmabuf, given that the GPU supports the output format?

This assignment is perhaps a bit misleading, I did this to make the `fmt == frame->format` check below always true for the dmabuf case.

emersion: The format conversion may be costly.

render/gles2/renderer.c
any1: You know where the header for this is?

ascent12: `<GLES2/gl2ext.h>`

emersion: Should be in `GLES2/gl2ext.h`

types/wlr_screencopy_v1.c
cyanreg: I don't think we should be allowing clients to choose formats, at least not without providing them with a hint and some info on what conversions are allowed. I'd still rather not let them choose their formats.
In Vulkan image copies and downloads are generally restricted to only allow same formats.

render/gles2/renderer.c
any1: Odd, I don't have it in mine. Are you guys running arch too?

types/wlr_screencopy_v1.c
any1: Unless we want to make a backwards incompatible change to the protocol the format still needs to be `wl_shm_format`, even for dmabuf. Is that ok?

render/gles2/renderer.c
ascent12: Oh, in this case, `GL_RENDERBUFFER` is part of core GLES2. You just need to use that.

any1: Oh, I see. I guess that only applies the GL ES 1.0 then. I was reading this: https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image.txt

types/wlr_screencopy_v1.c
cyanreg: I think that's the idea.

any1: Well, I pushed two commits to that effect yesterday, so the format is enforced now.

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+// wl_shm_format describes little-endian formats, libpng uses big-endian
+// formats (so Wayland's ABGR is libpng's RGBA).
+static const struct format formats[] = {
+	{WL_SHM_FORMAT_XRGB8888, true},
+	{WL_SHM_FORMAT_ARGB8888, true},
+	{WL_SHM_FORMAT_XBGR8888, false},
+	{WL_SHM_FORMAT_ABGR8888, false},
+};
+
+static int find_render_node(char* node, size_t maxlen) {
```
emersion, line 78: Style: `*` on the right

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+	int r = -1;
+	drmDevice* devices[64];
+
+	int n = drmGetDevices2(0, devices, sizeof(devices) / sizeof(devices[0]));
+	for (int i = 0; i < n; ++i) {
+		drmDevice* dev = devices[i];
+		if (!(dev->available_nodes & (1 << DRM_NODE_RENDER)))
```
emersion, line 85: Style: braces are mandatory

examples/screencopy-dmabuf.c
emersion, line 78: wlroots has the convention to return a `bool` when a function can fail

protocol/wlr-screencopy-unstable-v1.xml
```diff
@@ -102,6 +102,12 @@
 
         The client should then create a buffer with the provided attributes, and
         send a "copy" request.
+
+        When the buffer is a dmabuf, the stride parameter may be ignored.
```
emersion, line 106: These changes need to be sent to wlr-protocols

render/gles2/renderer.c
```diff
@@ -331,6 +331,72 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
 	return glGetError() == GL_NO_ERROR;
 }
 
+static bool gles2_blit_dmabuf(struct wlr_renderer *wlr_renderer,
+		struct wlr_dmabuf_attributes *dst_attr,
+		struct wlr_dmabuf_attributes *src_attr) {
+	bool r = false;
+
+	if (!gles2_procs.glEGLImageTargetRenderbufferStorageOES) {
+		return 0;
```
emersion, line 340: s/0/false/

render/gles2/renderer.c
```diff
@@ -331,6 +331,72 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	struct wlr_texture *src_tex =
+		wlr_texture_from_dmabuf(wlr_renderer, src_attr);
+
+	// This is to take into account y-inversion on both buffers rather than
+	// just the source buffer.
+	bool src_inverted_y =
+		!!(src_attr->flags & WLR_DMABUF_ATTRIBUTES_FLAGS_Y_INVERT);
+	bool dst_inverted_y =
+		!!(dst_attr->flags & WLR_DMABUF_ATTRIBUTES_FLAGS_Y_INVERT);
+	struct wlr_gles2_texture* gles2_src_tex = gles2_get_texture(src_tex);
```
emersion, line 357: Style: `*` on the right

render/gles2/renderer.c
```diff
@@ -331,6 +331,72 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	float mat[9];
+	wlr_matrix_projection(mat, 1, 1, WL_OUTPUT_TRANSFORM_NORMAL);
+
+	wlr_renderer_begin(wlr_renderer, dst_attr->width, dst_attr->height);
+	wlr_renderer_clear(wlr_renderer, (float[]){ 0.0, 0.0, 0.0, 0.0 });
+	wlr_render_texture_with_matrix(wlr_renderer, src_tex, mat, 1.0f);
+	wlr_renderer_end(wlr_renderer);
+
+	r = true;
+failure:
```
emersion, line 391: This should be named "out" rather than "failure"

render/wlr_renderer.c
```diff
@@ -154,6 +154,15 @@ bool wlr_renderer_read_pixels(struct wlr_renderer *r, enum wl_shm_format fmt,
 		src_x, src_y, dst_x, dst_y, data);
 }
 
+bool wlr_renderer_blit_dmabuf(struct wlr_renderer *r,
```
emersion, line 157: This function should assert that we're not in the middle of a rendering operation.

protocol/wlr-screencopy-unstable-v1.xml
```diff
@@ -102,6 +102,12 @@
...
+
+        The format parameter shall be interpreted according to
```
emersion, line 108: Hmm, this isn't great. I wonder if we could do something better, like separate wl_shm and dmabuf hints…

protocol/wlr-screencopy-unstable-v1.xml
emersion, line 106: It would be better to say that the compositor can send a zero stride to mean "you can choose whatever".

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+		return EXIT_FAILURE;
+	}
+
+	gbm_device = gbm_create_device(drm_fd);
+	if (!gbm_device) {
+		fprintf(stderr, "Failed to create gbm device\n");
+		return EXIT_FAILURE;
+	}
+
+	struct wl_display * display = wl_display_connect(NULL);
```
emersion, line 301: Style: extra space

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+	{WL_SHM_FORMAT_ABGR8888, false},
+};
+
+static int find_render_node(char* node, size_t maxlen) {
+	int r = -1;
+	drmDevice* devices[64];
+
+	int n = drmGetDevices2(0, devices, sizeof(devices) / sizeof(devices[0]));
+	for (int i = 0; i < n; ++i) {
+		drmDevice* dev = devices[i];
```
emersion, line 84: Style: `*` on the right

protocol/wlr-screencopy-unstable-v1.xml
emersion, line 106: Should we also specify the modifier to the client?

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +293,39 @@ static void frame_handle_copy(struct wl_client *wl_client,
 		return;
 	}
 
-	struct wl_shm_buffer *buffer = wl_shm_buffer_get(buffer_resource);
-	if (buffer == NULL) {
+	struct wlr_dmabuf_v1_buffer* dma_buffer = NULL;
```
emersion, line 296: Style: `*` on the right

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+static const struct format formats[] = {
+	{WL_SHM_FORMAT_XRGB8888, true},
+	{WL_SHM_FORMAT_ARGB8888, true},
+	{WL_SHM_FORMAT_XBGR8888, false},
+	{WL_SHM_FORMAT_ABGR8888, false},
+};
+
+static int find_render_node(char* node, size_t maxlen) {
+	int r = -1;
+	drmDevice* devices[64];
```
emersion, line 80: Style: `*` on the right

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+static uint32_t fourcc_from_wl_shm_format(enum wl_shm_format fmt) {
+	switch (fmt) {
+	case WL_SHM_FORMAT_ARGB8888: return DRM_FORMAT_ARGB8888;
+	case WL_SHM_FORMAT_XRGB8888: return DRM_FORMAT_XRGB8888;
+	default:;
+	}
+	return fmt;
+}
+
+static void dmabuf_created(void* data,
```
emersion, line 107: Style: `*` on the right

render/gles2/renderer.c
```diff
@@ -331,6 +331,72 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+static bool gles2_blit_dmabuf(struct wlr_renderer *wlr_renderer,
+		struct wlr_dmabuf_attributes *dst_attr,
+		struct wlr_dmabuf_attributes *src_attr) {
+	bool r = false;
+
+	if (!gles2_procs.glEGLImageTargetRenderbufferStorageOES) {
+		return 0;
+	}
+
+	struct wlr_egl* egl = wlr_gles2_renderer_get_egl(wlr_renderer);
```
emersion, line 343: Style: `*` on the right

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +293,39 @@ static void frame_handle_copy(struct wl_client *wl_client,
...
+	struct wl_shm_buffer *shm_buffer = wl_shm_buffer_get(buffer_resource);
+	if (shm_buffer == NULL) {
+		dma_buffer =
+			wlr_dmabuf_v1_buffer_from_buffer_resource(buffer_resource);
+	}
+
+	if (shm_buffer == NULL && dma_buffer == NULL ) {
```
emersion, line 303: Style: extra space

protocol/wlr-screencopy-unstable-v1.xml
any1: Yeah, this is a bit lame but the alternative is to have the user implement an almost identical callback with slightly different parameters. In the end, I believe this to be the more pragmatic approach.

protocol/wlr-screencopy-unstable-v1.xml
any1: How about this: "A stride parameter of 0 means that the stride is inconsequential to the operation of the compositor and may be chosen by the client."

any1: Are there any obvious benefits from doing so? What are the drawbacks of not allowing the client to choose its modifiers?

emersion: I was worried about some renderers not being able to perform tiling conversions. However it seems like `vkCmdCopyImage` doesn't care about tiling, so it should be fine.

any1: Oh, yeah, I forgot: We don't actually know that the buffer is going to be a dmabuf at this point, so the stride must contain a valid value.

any1:
Rebased and updated based on review.

include/wlr/render/wlr_renderer.h
```diff
@@ -104,6 +104,12 @@ const struct wlr_drm_format_set *wlr_renderer_get_dmabuf_formats(
 bool wlr_renderer_read_pixels(struct wlr_renderer *r, enum wl_shm_format fmt,
 	uint32_t *flags, uint32_t stride, uint32_t width, uint32_t height,
 	uint32_t src_x, uint32_t src_y, uint32_t dst_x, uint32_t dst_y, void *data);
+
+/**
+ * Blits the dmabuf in src onto the one in dst.
+ */
+bool wlr_renderer_blit_dmabuf(struct wlr_renderer *r,
```
progandy, line 111: Should this function also take a region so that ```zwlr_screencopy_capture_output_region``` can work as well?

any1: Could do that, but it could also be added later.

progandy: Ah right, wlr_renderer is an unstable API, so breaking changes are not an issue.
Maybe a todo entry here as well?

protocol/wlr-screencopy-unstable-v1.xml
any1: https://github.com/swaywm/wlr-protocols/pull/80

include/wlr/render/wlr_renderer.h
any1: Well, I could also add the arguments and just leave them unimplemented. That might be "nicer" thing to do.

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +293,39 @@ static void frame_handle_copy(struct wl_client *wl_client,
 		return;
 	}
 
-	struct wl_shm_buffer *buffer = wl_shm_buffer_get(buffer_resource);
-	if (buffer == NULL) {
+	struct wlr_dmabuf_v1_buffer *dma_buffer = NULL;
+	struct wl_shm_buffer *shm_buffer = wl_shm_buffer_get(buffer_resource);
+	if (shm_buffer == NULL) {
+		dma_buffer =
+			wlr_dmabuf_v1_buffer_from_buffer_resource(buffer_resource);
```
emersion, line 300: This function aborts if the resource isn't a dmabuf. We need to use `wlr_dmabuf_v1_resource_is_buffer` to check before calling it.

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +293,39 @@ static void frame_handle_copy(struct wl_client *wl_client,
...
+	int32_t width = 0;
+	int32_t height = 0;
+	int32_t stride = 0;
+
+	if (shm_buffer) {
+		fmt = wl_shm_buffer_get_format(shm_buffer);
+		width = wl_shm_buffer_get_width(shm_buffer);
+		height = wl_shm_buffer_get_height(shm_buffer);
+		stride = wl_shm_buffer_get_stride(shm_buffer);
+	} else if(dma_buffer) {
```
emersion, line 320: Style nit: missing space after `if`

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +293,39 @@ static void frame_handle_copy(struct wl_client *wl_client,
...
+		fmt = fourcc_to_wl_shm_format(dma_buffer->attributes.format);
+		width = dma_buffer->attributes.width;
+		height = dma_buffer->attributes.height;
+
+		// Doesn't matter for dmabuf
+		stride = frame->stride;
+	}
```
emersion, line 327:
```c
} else {
	abort();
}
```

Just to make sure we handle all cases

types/wlr_screencopy_v1.c
```diff
@@ -132,7 +134,8 @@ static void frame_destroy(struct wlr_screencopy_frame_v1 *frame) {
 	if (frame == NULL) {
 		return;
 	}
-	if (frame->output != NULL && frame->buffer != NULL) {
+	if (frame->output != NULL && frame->shm_buffer != NULL &&
+			frame->dma_buffer) {
```
emersion, line 138: This condition will never be true. Maybe the end of it should be changed to `(frame->shm_buffer != NULL || frame->dma_buffer != NULL)`?

render/gles2/renderer.c
```diff
@@ -331,6 +331,74 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+
+	if (!gles2_procs.glEGLImageTargetRenderbufferStorageOES) {
+		return false;
+	}
+
+	bool r = false;
+
+	struct wlr_egl *egl = wlr_gles2_renderer_get_egl(wlr_renderer);
+	if (!wlr_egl_is_current(egl)) {
+		wlr_egl_make_current(egl, EGL_NO_SURFACE, NULL);
```
emersion, line 347: The renderer is now expected to reset the EGL context when done, via `wlr_egl_unset_current`

types/wlr_screencopy_v1.c
any1: Yes, that is most likely what I meant to say.

include/wlr/render/wlr_renderer.h
any1: @emersion, you have an opinion on this?

emersion: Hmm, I'd prefer to do a breaking change:

1. I don't think a lot of compositors will be using this function, so this would be a "cheap" breaking change
2. It's hard to get the API right when you haven't actually implemented the feature.

any1:
Rebased and updated according to review.

render/gles2/renderer.c
```diff
@@ -331,6 +331,74 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	wlr_renderer_end(wlr_renderer);
+
+	r = true;
+out:
+	glBindFramebuffer(GL_FRAMEBUFFER, 0);
+	glDeleteFramebuffers(1, &fbo);
+	glDeleteRenderbuffers(1, &rbo);
+	wlr_egl_destroy_image(egl, image);
+	wlr_texture_destroy(src_tex);
+	wlr_egl_unset_current(egl);
```
emersion, line 398: As discussed on IRC, this causes issues because this function is called in wlr_output's precommit event handler. Unsetting the current EGL context makes `wlr_egl_swap_buffers` fail.

A workaround would be to leave the EGL context behind here (by _not_ calling `wlr_egl_unset_current`). A proper solution would involve restoring the previous context or ref-counting the number of times `wlr_egl_make_current` is called (this will probably mess up the current EGL surface).

any1: In order to restore the context, I would have to save it before calling `blit_dmabuf()` and then restore it after. Is there an established pattern for that?

any1: Maybe something like this?
```
diff --git a/render/gles2/renderer.c b/render/gles2/renderer.c
index 8c5651db..a3ac3ec0 100644
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -340,6 +340,9 @@ static bool gles2_blit_dmabuf(struct wlr_renderer *wlr_renderer,
                return false;
        }

+       EGLContext old_ctx = eglGetCurrentContext();
+       EGLSurface old_surf = eglGetCurrentSurface(EGL_DRAW);
+
        bool r = false;
        struct wlr_texture *src_tex =
                wlr_texture_from_dmabuf(wlr_renderer, src_attr);
@@ -395,7 +398,7 @@ out:
        glDeleteRenderbuffers(1, &rbo);
        wlr_egl_destroy_image(egl, image);
        wlr_texture_destroy(src_tex);
-       wlr_egl_unset_current(egl);
+       wlr_egl_make_current(egl, old_ctx, old_surf);
        return r;
 }
```

emersion: >Is there an established pattern for that?

No, because we overwrite the previous context everywhere right now.

>Maybe something like this?

Something like that should work, but maybe also extract the `EGL_READ` surface separately to avoid surprises. And also use `eglGetCurrentDisplay` in case we're juggling between multiple displays (like we do in the DRM backend).

any1: Maybe this should be a pair of functions called `wlr_egl_save_current(struct wlr_egl_context*)` and `wlr_egl_restore_current(struct wlr_egl_context*)`?

danshick: Push and pop?

any1: push/pop would be fancy, but I believe the call stack will suffice.

any1: I have now added the utility functions as discussed on IRC and applied them to `blit_dmabuf()`.

types/wlr_screencopy_v1.c
```diff
@@ -269,18 +293,42 @@ static void frame_handle_copy(struct wl_client *wl_client,
...
-	if (buffer == NULL) {
+	struct wlr_dmabuf_v1_buffer *dma_buffer = NULL;
+	struct wl_shm_buffer *shm_buffer = wl_shm_buffer_get(buffer_resource);
+	if (shm_buffer == NULL &&
+			wlr_dmabuf_v1_resource_is_buffer(buffer_resource)) {
+		dma_buffer =
+			wlr_dmabuf_v1_buffer_from_buffer_resource(buffer_resource);
+	}
+
+	if (shm_buffer == NULL && dma_buffer == NULL) {
```
emersion, line 304: Hmm, so, the only way for a client to know whether a linux-dmabuf wl_buffer is supported by screencopy is to try and get a protocol error. Something more graceful would be nicer.

any1: As a client developer, that doesn't really bother me. The capability check would still need to be implemented on the client side and it wouldn't be much simpler than just trying to send a dmabuf before settling on an shm buffer if dmabuf fails the first time.

emersion: This is a protocol error, this means that you get disconnected from the compositor. Also protocol errors should only be used for client bugs.

any1: Hmm, yes, I see. If a client tries to use dmabuf with an older compositor, they will get a protocol error. So, the only way to make things compatible is to extend the protocol.

emersion: Right, bumping the protocol version is required here I think.

any1: Done

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,353 @@
...
+static void frame_handle_linux_dmabuf(void *data,
+		struct zwlr_screencopy_frame_v1 *frame, uint32_t fourcc,
+		uint32_t width, uint32_t height) {
+	buffer.width = width;
+	buffer.height = height;
+	buffer.format = fourcc;
+}
+
+static void frame_handle_buffer_done(void *data,
+		struct zwlr_screencopy_frame_v1 *frame) {
```
emersion, line 133: Should error out if the "linux_dmabuf" event hasn't been received.

render/egl.c
```diff
@@ -394,6 +394,20 @@ bool wlr_egl_is_current(struct wlr_egl *egl) {
 	return eglGetCurrentContext() == egl->context;
 }
 
+void wlr_egl_save_context(struct wlr_egl_context *context)
+{
```
emersion, line 398: Style: `{` on the previous line

render/egl.c
```diff
@@ -394,6 +394,20 @@ bool wlr_egl_is_current(struct wlr_egl *egl) {
...
+	context->display = eglGetCurrentDisplay();
+	context->context = eglGetCurrentContext();
+	context->draw_surface = eglGetCurrentSurface(EGL_DRAW);
+	context->read_surface = eglGetCurrentSurface(EGL_READ);
+}
+
+bool wlr_egl_restore_context(struct wlr_egl_context *context)
+{
```
emersion, line 406: Style: `{` on the previous line

render/gles2/renderer.c
```diff
@@ -331,6 +331,77 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	if (!gles2_procs.glEGLImageTargetRenderbufferStorageOES) {
+		return false;
+	}
+
+	struct wlr_egl_context old_context;
+	wlr_egl_save_context(&old_context);
+
+	bool r = false;
+	struct wlr_texture *src_tex =
+		wlr_texture_from_dmabuf(wlr_renderer, src_attr);
```
emersion, line 348: This function returns NULL when the import fails.

render/gles2/renderer.c
```diff
@@ -331,6 +331,77 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
...
+	struct wlr_gles2_texture *gles2_src_tex = gles2_get_texture(src_tex);
+	gles2_src_tex->inverted_y = src_inverted_y ^ dst_inverted_y;
+
+	GLuint rbo = 0;
+	glGenRenderbuffers(1, &rbo);
+	glBindRenderbuffer(GL_RENDERBUFFER, rbo);
+
+	// TODO: The imported buffer should be checked with
+	// eglQueryDmaBufModifiersEXT to see if it may be modified.
+	EGLImageKHR image = wlr_egl_create_image_from_dmabuf(egl, dst_attr);
```
emersion, line 370: We should also check that this isn't `EGL_NO_IMAGE_KHR`

types/wlr_screencopy_v1.c
```diff
@@ -253,6 +269,23 @@ static void frame_handle_buffer_destroy(struct wl_listener *listener,
 	frame_destroy(frame);
 }
 
+static enum wl_shm_format fourcc_to_wl_shm_format(uint32_t fourcc) {
```
emersion, line 272: Hmm. Can't we just compare formats without converting them? We shouldn't need these functions now.

types/wlr_screencopy_v1.c
```diff
@@ -434,6 +492,15 @@ static void capture_output(struct wl_client *wl_client,
 
 	zwlr_screencopy_frame_v1_send_buffer(frame->resource, frame->format,
 		buffer_box.width, buffer_box.height, frame->stride);
+
+	if (version >= 3) {
+		uint32_t fourcc = fourcc_from_wl_shm_format(frame->format);
```
emersion, line 497: The preferred OpenGL read format isn't the same as the format used by the output buffer.

any1: But `wlr_output_preferred_read_format()` yields `enum wl_shm_format`

types/wlr_screencopy_v1.c
any1: This one is needed to check if the supplied dmabuf format is correct. The other one is needed to convert from `frame->format` to fourcc.

Well, actually we can use the other one to check the format, but the way its done now is nicer, imo.

types/wlr_screencopy_v1.c
emersion: `wlr_output_preferred_read_format` is only applicable to wl_shm, not to DMA-BUFs.

any1: This is why I convert the wl_shm format to fourcc. Do you mean that the conversion fourcc -> gl read format -> wl_shm_format -> fourcc might not be accurate? Does there exist some way to read the fourcc format of the dmabuf directly?

emersion: What I mean is that the format returned by `wlr_output_preferred_read_format` will not necessarily be the same as the one used by `wlr_output_export_dmabuf` (modulo conversion). `wlr_output_preferred_read_format` is about `wlr_renderer_read_pixels` only, nothing else.

escalade:
Fails to compile with #2255:

```
../render/gles2/renderer.c: In function 'gles2_blit_dmabuf':
../render/gles2/renderer.c:370:22: error: too few arguments to function 'wlr_egl_create_image_from_dmabuf'
  370 |  EGLImageKHR image = wlr_egl_create_image_from_dmabuf(egl, dst_attr);
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

emersion:
>Fails to compile with #2255

This one hasn't been merged, so it's expected.

types/wlr_screencopy_v1.c
any1: Oh, right, I found it. It's in the `struct wlr_dmabuf_attributes`.

examples/screencopy-dmabuf.c
```diff
@@ -0,0 +1,362 @@
...
+	buffer.format = fourcc;
+	have_linux_dmabuf = true;
+}
+
+static void frame_handle_buffer_done(void *data,
+		struct zwlr_screencopy_frame_v1 *frame) {
+	assert(!buffer.bo);
+
+	if (!have_linux_dmabuf) {
+		fprintf(stderr, "\n");
```
emersion, line 139: Maybe print a useful error message here?

render/gles2/renderer.c
```diff
@@ -331,6 +331,86 @@ static bool gles2_read_pixels(struct wlr_renderer *wlr_renderer,
 	return glGetError() == GL_NO_ERROR;
 }
 
+static bool gles2_blit_dmabuf(struct wlr_renderer *wlr_renderer,
+		struct wlr_dmabuf_attributes *dst_attr,
+		struct wlr_dmabuf_attributes *src_attr) {
+	assert(!wlr_renderer->rendering);
```
emersion, line 337: This assert should be moved in the common code, in `wlr_renderer_blit_dmabuf`, so that each renderer doesn't need to add it.

emersion:
Can you also squash this PR into logical commits? ie. squash "examples: screencopy-dmabuf: Use screencopy v3" into "examples: Add screencopy-dmabuf example" and "screencopy: Implement screencopy v3" into "screencopy: Implement copying into dmabuf". Would be nice to have the `wlr_renderer_blit_dmabuf` addition in a separate commit.

examples/screencopy-dmabuf.c
any1: Hah! Forgot to fill in that one.

any1:
Fixed the error message, the assert and squashed as requested.

PR #2135, opened 2020-04-26
tinywl: Fix window resize slightly off

greg904:
Previously, when dragging the left border of a window with the mouse in tinywl,
there was a bug where it snap the top level surface's geometry X coordinate
directly to the position of the mouse, as if you started the resize right on the
border. This also affected the other (right, top and bottom) borders.

I think that the previous resize code was hard to understand. Honestly I
have not spent a lot of time trying to understand why it didn't work and
I wrote another resize algorithm instead. I hope this is fine. Now, instead of working directly with widths and heights which are complicated, we work with the borders (left,
right, top, bottom). This is easier to understand IMO.

Note: I originally fixed this [in the waybox compositor](https://github.com/wizbright/waybox/pull/23) but
then I realized that the code was taken from tinywl and that it had the same
issues so I copied my fix for tinywl.

### DIFF
--- a/tinywl/tinywl.c
+++ b/tinywl/tinywl.c
@@ -54,8 +54,7 @@ struct tinywl_server {
 	enum tinywl_cursor_mode cursor_mode;
 	struct tinywl_view *grabbed_view;
 	double grab_x, grab_y;
-	struct wlr_box grab_geo_box;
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

PR #2136, opened 2020-04-26
export-dmabuf: Schedule output frame on request

any1:
A client requesting frames in the ready callback may miss frames that
happen while it is waiting to receive the event or sending the request.
If that happens, the client will have an outdated frame for an
indefinite period. A new frame might not be scheduled for a very long
time.

With this change, clients will receive new frames immediately upon
request.

### DIFF
--- a/types/wlr_export_dmabuf_v1.c
+++ b/types/wlr_export_dmabuf_v1.c
@@ -150,6 +150,8 @@ static void manager_handle_capture_output(struct wl_client *client,
 			attribs->fd[i], size, attribs->offset[i], attribs->stride[i], i);
 	}
 
+	wlr_output_schedule_frame(output);
+
 	wl_list_remove(&frame->output_precommit.link);
 	wl_signal_add(&output->events.precommit, &frame->output_precommit);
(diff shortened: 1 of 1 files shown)


types/wlr_export_dmabuf_v1.c
```diff
@@ -150,6 +150,10 @@ static void manager_handle_capture_output(struct wl_client *client,
 			attribs->fd[i], size, attribs->offset[i], attribs->stride[i], i);
 	}
 
+	// Schedule a buffer commit
+	output->needs_frame = true;
```
emersion, line 154: No need for this line, this is already performed by `wlr_output_schedule_frame` now

cyanreg:
This causes the compositor to render again, doesn't it?
Can you just submit old buffers at the display framerate unless an update happens?

emersion:
This PR will cause compositors which support damage tracking to just swap buffers without rendering anything.

cyanreg:
LGTM

emersion:
Thanks!

PR #2137, opened 2020-04-27
input/tablet: fix up updated axes after rotation

Xyene:
In the case that only one axis received an event, rotating the input can
cause the change to actually happen on the other axis, as far as clients
are concerned.

This commit updates the axes flags to be consistent post-rotation.

Fixes swaywm/sway#4776.

### DIFF
--- a/types/wlr_cursor.c
+++ b/types/wlr_cursor.c
@@ -539,11 +539,33 @@ static void handle_tablet_tool_axis(struct wl_listener *listener, void *data) {
 	struct wlr_cursor_device *device;
 	device = wl_container_of(listener, device, tablet_tool_axis);
 
-	struct wlr_output *output =
-		get_mapped_output(device);
+	struct wlr_output *output = get_mapped_output(device);
 	if (output) {
-		apply_output_transform(&event->x, &event->y, output->transform);
+		// In the case that only one axis received an event, rotating the input can
+		// cause the change to actually happen on the other axis, as far as clients
+		// are concerned.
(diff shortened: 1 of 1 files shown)


types/wlr_cursor.c
```diff
@@ -539,11 +539,32 @@ static void handle_tablet_tool_axis(struct wl_listener *listener, void *data) {
...
+		//
+		// Here, we feed apply_output_transform NAN on the axis that didn't change,
+		// and remap the axes flags based on whether it returns NAN itself.
+		double x = event->updated_axes & WLR_TABLET_TOOL_AXIS_X ? event->x : NAN;
+		double y = event->updated_axes & WLR_TABLET_TOOL_AXIS_Y ? event->y : NAN;
+
+		apply_output_transform(&x, &y, output->transform);
+
+		event->updated_axes &= ~(WLR_TABLET_TOOL_AXIS_X | WLR_TABLET_TOOL_AXIS_Y);
+		event->x = isnan(x) ? 0 : x;
```
emersion, line 556: Can we improve slightly readability here by moving the coords assignment in the `if` branches? This would batch the `updated_axes` and `x`/`y` assignments . Something among those lines:

```c
event->x = event->y = 0;

if (!isnan(x)) {
	event->updated_axes |= WLR_TABLET_TOOL_AXIS_X;
	event->x = x;
}

…
```

Xyene: Done!

PR #2139, opened 2020-04-27
Texture improvements

emersion:
See individual commits.

* * *

Breaking changes for renderers:

* `wlr_texture_init` now takes the size of the texture
* `wlr_texture_impl.get_size` has been removed


ddevault:
Thanks!

PR #2140, opened 2020-04-28
Remove libcap support

ascent12:
This is simply a false sense of security, and is worse than just using
setuid. CAP_SYS_ADMIN is an extremely serious capability that is
effectively as powerful as root.

It also required users to be in the input group, which allows any
process to keylog the entire system.


emersion:
Thanks!

PR #2141, opened 2020-04-28
backend/drm: strip alpha channel if necessary

emersion:
Some primary planes don't support ARGB8888, they only support XRGB8888
(for instance on older Intel hardware). The DRM backend would fail to
initialize.

When that's the case, try to allocate buffers without an alpha channel.


agx:
LGTM, some embedded device fb also only have non-alpha planes

JonnyMako:
I've been seeing some weirdness with firefox-nightly menus having some issues. The first time the menu drops down it is visible but the cursor is unable to interact with the menu. I wonder if this will also solve that problem. I'll give it a try.

PR #2142, opened 2020-04-28
input/tablet: clear focused surface on surface destroy

Xyene:
Otherwise, we can end up left with a dangling pointer to a
previously-focused, now-destroyed surface.

Fixes swaywm/sway#5264.


types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -226,11 +226,11 @@ struct wlr_tablet_v2_tablet_tool *wlr_tablet_tool_create(
 
 	tool->wlr_tool = wlr_tool;
 	wl_list_init(&tool->clients);
+	wl_list_init(&tool->surface_destroy.link);
```
Xyene, line 229: Note for reviewer: this field already existed in the struct, but as far as I can tell was never used.

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -322,6 +322,15 @@ static void queue_tool_frame(struct wlr_tablet_tool_client_v2 *tool) {
 	}
 }
 
+static void handle_tablet_tool_surface_destroy(struct wl_listener *listener,
+		void *data) {
+	struct wlr_tablet_v2_tablet_tool *tool =
+		wl_container_of(listener, tool, surface_destroy);
+	wl_list_remove(&tool->surface_destroy.link);
+	wl_list_init(&tool->surface_destroy.link);
+	tool->focused_surface = NULL;
```
Xyene, line 331: Note for reviewer: I'm not really sure if this is enough, but it does fix the linked issue. In particular, `wlr_send_tablet_v2_tablet_tool_proximity_out` also clears `tool->current_client`, but I'm not sure if we should be doing that here, too.

types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -366,6 +375,12 @@ void wlr_send_tablet_v2_tablet_tool_proximity_in(
 		return;
 	}
 
+	// Reinitialize the focus destroy events
+	wl_list_remove(&tool->surface_destroy.link);
+	wl_list_init(&tool->surface_destroy.link);
```
emersion, line 380: No need to `wl_list_init` if `wl_signal_add` is called right after.

The reason we use `wl_list_init` is to be able to use `wl_list_remove` to unregister the listener unconditionally, without having to check whether we're listening to a signal or not.

types/tablet_v2/wlr_tablet_v2_tool.c
emersion: Hmm, yeah I think we should clear `current_client` too.

Should we just call `wlr_send_tablet_v2_tablet_tool_proximity_out`?

emersion:
We should also remove the listener in `wlr_send_tablet_v2_tablet_tool_proximity_out`.

types/tablet_v2/wlr_tablet_v2_tool.c
Xyene: Yeah, if we're also clearing `current_client` calling `proximity_out` makes sense to me too, it's already guarded by `current_client != NULL`.

types/tablet_v2/wlr_tablet_v2_tool.c
Xyene: Makes sense, done :+1:

Xyene:
> We should also remove the listener in `wlr_send_tablet_v2_tablet_tool_proximity_out`.

Good catch, done.

emersion:
One last thing: we should remove the listener in `handle_wlr_tablet_tool_destroy`, otherwise the surface's listener list will contain a pointer to a free'ed `wl_listener`.

minus7:
I can confirm this fixes https://github.com/swaywm/sway/issues/5264

PR #2143, opened 2020-04-30
examples/dmabuf-capture: use getopt

emersion:
This makes it a little bit less annoying to provide the right arguments.
All options have reasonable defaults.


ddevault:
Thanks!

PR #2148, opened 2020-05-01
backend/drm: make page-flip error scarier

emersion:
The logs don't currently display the importance of a line. It's easy to
read "skipping page-flip" as a debug message instead of an error
message.

Change the error message to make it more explicit.

References: https://github.com/swaywm/wlroots/pull/2147

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -341,7 +341,8 @@ static bool drm_crtc_page_flip(struct wlr_drm_connector *conn,
 	drmModeModeInfo *drm_mode = mode ? &mode->drm_mode : NULL;
 
 	if (conn->pageflip_pending) {
-		wlr_log(WLR_ERROR, "Skipping pageflip on output '%s'", conn->output.name);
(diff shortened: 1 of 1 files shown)


kennylevinsen:
Much better! :)

emersion:
Note, the Arch build fails because of:

```
> clang
clang: error while loading shared libraries: libicuuc.so.67: cannot open shared object file: No such file or directory
```

ddevault:
Has that been reported to Arch upstream?

ddevault:
Thanks!

emersion:
Still investigating. On my system I have icu 67.1-1, on builds.sr.ht we have 65.1-3. Both are up-to-date, clang works locally. clang depends on icu through libxml2. Both my local machine and builds.sr.ht have libxml2 2.9.10-2.

tl;dr probably an issue due to a partial update on one of Arch package mirrors.

ddevault:
...so, report your findings upstream?

emersion:
Yeah, definitely. Still collecting more info for now.

ddevault:
Thanks :)

PR #2149, opened 2020-05-01
util/log: write log importance

emersion:
When colors aren't used, write the log importance to stderr. This makes
it easier to grep for errors and avoids mistaking error messages for
debug messages.


ddevault:
That arch problem looks legit, can you investigate?

emersion:
[Yes](https://github.com/swaywm/wlroots/pull/2148#issuecomment-622405622)

PR #2151, opened 2020-05-02
Add wlr_surface_accepts_touch

David96:
This is needed to simulate pointer events when a surface hasn't bound to touch events.


types/wlr_touch.c
```diff
@@ -20,3 +21,12 @@ void wlr_touch_destroy(struct wlr_touch *touch) {
 		free(touch);
 	}
 }
+
+bool wlr_surface_accepts_touch(struct wlr_seat *wlr_seat, struct wlr_surface *surface) {
```
emersion, line 25: This should be moved in `types/seat/wlr_seat_touch.c` since this is a seat-related function.

PR #2155, opened 2020-05-04
backend/libinput: clamp tablet event coordinates to [0, 1]

### LINKED ISSUE #2128: Out-of-bound tablet tool coordinate handling
The tablet backend uses `libinput_event_tablet_tool_get_(x|y)_transformed` to get the position of the tablet tool in the range [0, 1]. At least one part of the code explicitly comments this assumption.

https://github.com/swaywm/wlroots/blob/ebeef0fbe8b15fceb618ee5bc6947d1c3edf950f/include/wlr/types/wlr_tablet_tool.h#L118-L119

However, [libinput can return values outside of this range](https://wayland.freedesktop.org/libinput/doc/1.4.0/tablet-support.html#tablet-bounds) on some tablets.

> ... the resulting value may be less than 0 or larger than the upper range provided. It is up to the caller to test for this and handle or ignore these events accordingly.

I'm not sure if this can cause issues down the line, but there being at least one comment based on that assumption, I felt it was worthwhile to at least open a ticket for further discussion.

Xyene:
libinput can return events that are out of the range [0, 1] on some
tablets where the actual sensor area is larger than the
kernel-advertised area \[1\]. To prevent confusing consumers of tablet
events with out-of-range coordinates, clamp them to [0, 1] here.

From a cursory look around, this seems to be what GNOME does too[2].

(Note that I'm not sure if this fixes any real-world problems since I don't
own such a tablet, but it *seems* like the type of issue that'd cause
impossible-to-remotely-debug issues.)

\[1\]: https://wayland.freedesktop.org/libinput/doc/1.13.0/tablet-support.html#out-of-bounds-motion-events
\[2\]: https://gitlab.gnome.org/GNOME/mutter/-/blob/master/src/backends/native/meta-input-device-native.c#L1485

Closes #2128.

### DIFF
--- a/backend/libinput/tablet_tool.c
+++ b/backend/libinput/tablet_tool.c
@@ -188,6 +188,15 @@ static void ensure_tool_reference(struct wlr_libinput_tablet_tool *tool,
 	++tool->pad_refs;
 }
 
+static void clamp_tool_event_axes(double *x, double *y) {
+	// libinput can return events that are out of the range [0, 1] on some
+	// tablets where the actual sensor area is larger than the kernel-advertised
+	// area. To prevent confusing consumers of tablet events with out-of-range
+	// coordinates, clamp them to [0, 1] here.
+	*x = fmin(fmax(*x, 0), 1);
+	*y = fmin(fmax(*y, 0), 1);
+}
+
 void handle_tablet_tool_axis(struct libinput_event *event,
 		struct libinput_device *libinput_dev) {
 	struct wlr_input_device *wlr_dev =
@@ -246,6 +255,8 @@ void handle_tablet_tool_axis(struct libinput_event *event,
 		wlr_event.updated_axes |= WLR_TABLET_TOOL_AXIS_WHEEL;
 		wlr_event.wheel_delta = libinput_event_tablet_tool_get_wheel_delta(tevent);
 	}
+
+	clamp_tool_event_axes(&wlr_event.x, &wlr_event.y);
 	wlr_signal_emit_safe(&wlr_dev->tablet->events.axis, &wlr_event);
 }
 
@@ -271,6 +282,7 @@ void handle_tablet_tool_proximity(struct libinput_event *event,
 		usec_to_msec(libinput_event_tablet_tool_get_time_usec(tevent));
 	wlr_event.x = libinput_event_tablet_tool_get_x_transformed(tevent, 1);
 	wlr_event.y = libinput_event_tablet_tool_get_y_transformed(tevent, 1);
+	clamp_tool_event_axes(&wlr_event.x, &wlr_event.y);
 
 	switch (libinput_event_tablet_tool_get_proximity_state(tevent)) {
 	case LIBINPUT_TABLET_TOOL_PROXIMITY_STATE_OUT:
@@ -335,6 +347,7 @@ void handle_tablet_tool_tip(struct libinput_event *event,
 		usec_to_msec(libinput_event_tablet_tool_get_time_usec(tevent));
 	wlr_event.x = libinput_event_tablet_tool_get_x_transformed(tevent, 1);
 	wlr_event.y = libinput_event_tablet_tool_get_y_transformed(tevent, 1);
+	clamp_tool_event_axes(&wlr_event.x, &wlr_event.y);
 
 	switch (libinput_event_tablet_tool_get_tip_state(tevent)) {
 	case LIBINPUT_TABLET_TOOL_TIP_UP:


emersion:
Hmm. I wonder if that's a good place to clamp.

The Wayland backend may also yield these values in case the button is pressed and then the tool is moved outside the surface bounds.

Maybe we should do the clamping in `wlr_cursor`?

Xyene:
I think it might be best to fix this separately in the Wayland backend (possibly re-using this logic). My concern is if we leave it to `wlr_cursor`, we rely on the compositor to have already dealt with the negative values in a sensible manner.

I'm mostly concerned about feeding sway negative coordinates through non-trivial math for handling monitor rotations, scaling etc. &mdash; I'm not sure how that code would respond to negative inputs in the present, much less in the future. It's probably tolerant of them, but it still seems a bit "scary".

emersion:
Hmm. Another thing is that some compositors may *want* to react to out-of-bounds tablet events. For instance a compositor might want to fade-out or apply an effect on the cursor depending on the distance to the edge. I'm not sure it's a good idea for wlroots to decide values should be clamped.

Xyene:
To see if there was actually anything wrong with how sway would handle tablet coordinates coming outside the range [0, 1] (rather than a theoretical possibility) without owning such a tablet myself, I switched the "clamping" function to this:

```c
static void clamp_tool_event_axes(double *x, double *y) {
	*x = -0.25 + fmin(fmax(*x, 0), 1) * 1.5;
	*y = -0.25 + fmin(fmax(*y, 0), 1) * 1.5;
}
```

...such that my events are now simulated as coming in the range [-0.25, 1.25]. Then I loaded up sway and moved the pen around, and sway crashed when I put the tip down while it was in one of the out-of-bounds edges.

<details>
<summary>ASan trace</summary>

```
==19727==ERROR: AddressSanitizer: SEGV on unknown address 0x0000000001a0 (pc 0x5601d3d74d25 bp 0x7ffd351f7290 sp 0x7ffd351f7270 T0)
==19727==The signal is caused by a READ memory access.                                                                  
==19727==Hint: address points to the zero page.                                                                         
  ┊ #0 0x5601d3d74d24 in container_raise_floating ../sway/tree/container.c:1528                                         
  ┊ #1 0x5601d3d0a39c in seatop_begin_down ../sway/input/seatop_down.c:110                                              
  ┊ #2 0x5601d3d057f6 in handle_tablet_tool_tip ../sway/input/seatop_default.c:217                                      
  ┊ #3 0x5601d3d03e5b in seatop_tablet_tool_tip ../sway/input/seat.c:1470                                               
  ┊ #4 0x5601d3ced32f in handle_tool_tip ../sway/input/cursor.c:668                                                     
  ┊ #5 0x7f93565b6dec in wlr_signal_emit_safe ../util/signal.c:29                                                       
  ┊ #6 0x7f935656a0d8 in handle_tablet_tool_tip ../types/wlr_cursor.c:534                                               
  ┊ #7 0x7f93565b6dec in wlr_signal_emit_safe ../util/signal.c:29                                                       
  ┊ #8 0x7f9356507257 in handle_tablet_tool_tip ../backend/libinput/tablet_tool.c:363                                   
  ┊ #9 0x7f9356500f9a in handle_libinput_event ../backend/libinput/events.c:273                                         
  ┊ #10 0x7f93564ff3e3 in handle_libinput_readable ../backend/libinput/backend.c:41                                     
  ┊ #11 0x7f9356680401 in wl_event_loop_dispatch (/lib/x86_64-linux-gnu/libwayland-server.so.0+0xb401)                  
  ┊ #12 0x7f935667e9e4 in wl_display_run (/lib/x86_64-linux-gnu/libwayland-server.so.0+0x99e4)                          
  ┊ #13 0x5601d3cbfb73 in server_run ../sway/server.c:225                                                               
  ┊ #14 0x5601d3cbe0ba in main ../sway/main.c:409                                                                       
  ┊ #15 0x7f9356254e0a in __libc_start_main ../csu/libc-start.c:308                                                     
  ┊ #16 0x5601d3c9d549 in _start (/usr/local/bin/sway+0x3f549)                                                          
                                                                                                                        
AddressSanitizer can not provide additional info.                                                                       
SUMMARY: AddressSanitizer: SEGV ../sway/tree/container.c:1528 in container_raise_floating                               
==19727==ABORTING 
```

</details>

At the very least, it seems sway _does_ currently crash when faced with one of these tablets, so there should be a fix somewhere, either in wlroots or in sway.

> For instance a compositor might want to fade-out or apply an effect on the cursor depending on the distance to the edge.

I think this is a fair point. But, it's unclear to me whether this is an actual "feature" of these tablets, or just a consequence of how they're built. My understanding is we're talking about (a handful of?) devices like this Wacom Cintiq 24HD,



...where some portion of the light gray border surrounding the screen has a sensor under it, presumably due to real-world building constraints. (At the very least, I couldn't find any mention of that area having a sensor under it in any manuals/promotional material I dug up, so I'd guess that area is ignored even if it does send events. I find this quirk interesting, so I'd love to find more info on it, but couldn't find much.)

This does seem a bit of an edge case, though. If wlroots didn't handle it, a compositor would see something like "I have a tablet event coming on this output that is 1920x1080, and the coordinates of the event are (2000, 900)" and then get confused because 2000 > 1920, and this doesn't ever happen for the more common/better-tested pointer/touch paths. For what it's worth, sway appears to correctly not allow the cursor to visually move past the edges, but crashes anyway.

I'll defer judgement on which side should fix this, but it does seem to me like something a compositor using wlroots is much more likely to get wrong than right, and is sufficiently obscure that it'd take a long time to run into. Conversely, it'd probably bite very few users in the first place, so it's not a huge deal.

progandy:
>  I think this is a fair point. But, it's unclear to me whether this is an actual "feature" of these tablets, or just a consequence of how they're built. My understanding is we're talking about (a handful of?) devices like this Wacom Cintiq 24HD,

It might be there to make the detection of edge-swipe gestures more reliable.

emersion:
Side note, another use-case is allowing compositors to define shortcuts when the event is out-of-bounds (e.g. scroll when dragging on the left side). This is still pretty niche.

jigpu:
linuxwacom maintainer over here hoping to clear (or muddy? :sweat_smile:) things up...

Almost every one of Wacom's display tablet products has this out-of-bounds area surrounding the edge of the display. It isn't simply the result of hardware or manufacturing limitations. A typical display will have the sensor extend 2mm beyond each edge, though some sensors (mostly integrated into tablet PCs, but also the rare branded product like the "[Wacom One](https://www.wacom.com/en-us/products/pen-displays/wacom-one)") have no such extension.

There are a few different purposes to the edge, but the main ones that stand out to me as applicable to compositors are user calibration and edge pressure. There might be other uses I'm unaware of outside of Linux; I'd have to check with the Windows and Mac developers.

User calibration corrects for parallax errors caused by the frontglass-display-sensor stack having nonzero thickness. Having some additional room around the edges gives more flexibility for dealing with a pen that appears to be inside the display area, but which is physically outside of it. Out-of-bounds coordinates would be needed by the compositor or (or sway?) whenever you are applying the calibration transformation to the raw libinput coordinates.

Edge pressure is used whenever you want to know that the user is intentionally trying to move the pointer off-screen for some reason. It is used to implement the kind of features that @progandy and @emersion describe. Edge-swipe in particular stands out to me as something which I've seen used before.

emersion:
Thanks for chiming in, @jigpu! Yeah, I agree saying that the coords can be out-of-bounds would probably be less opinionated and allow some of these features.

Xyene:
Thanks for taking the time to write up that detailed post, @jigpu! That clears up the purpose of those negative coordinates, then. (I just realized I deleted this branch while cleaning up without replying, oops.)

> Yeah, I agree saying that the coords can be out-of-bounds would probably be less opinionated and allow some of these features.

Likewise, in light of this information it seems best that we just fix the documentation in wlroots. Sway at least doesn't seem to break with negative tablet coordinates, so it's fine.

PR #2158, opened 2020-05-04
Fixups for tablet tool implicit grabs

Xyene:
Refs https://github.com/swaywm/sway/issues/5302.

wlroots counterpart to https://github.com/swaywm/sway/pull/5303.

### DIFF
--- a/types/tablet_v2/wlr_tablet_v2_tool.c
+++ b/types/tablet_v2/wlr_tablet_v2_tool.c
@@ -792,18 +792,6 @@ static void implicit_tool_up(struct wlr_tablet_tool_v2_grab *grab) {
 	check_and_release_implicit_grab(grab);
 }
 
-/* Only send the motion event, when we are over the surface for now */
-static void implicit_tool_motion(
-	struct wlr_tablet_tool_v2_grab *grab, double x, double y) {
-	struct implicit_grab_state *state = grab->data;
-	if (state->focused != state->original) {
-		return;
-	}
-
-	wlr_send_tablet_v2_tablet_tool_motion(grab->tool, x, y);
-}
-
-
 static void implicit_tool_button(
 	struct wlr_tablet_tool_v2_grab *grab, uint32_t button,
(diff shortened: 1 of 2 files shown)


types/tablet_v2/wlr_tablet_v2_tool.c
```diff
@@ -792,18 +792,6 @@ static void implicit_tool_up(struct wlr_tablet_tool_v2_grab *grab) {
 	check_and_release_implicit_grab(grab);
 }
 
-/* Only send the motion event, when we are over the surface for now */
-static void implicit_tool_motion(
-	struct wlr_tablet_tool_v2_grab *grab, double x, double y) {
-	struct implicit_grab_state *state = grab->data;
-	if (state->focused != state->original) {
```
emersion, line 799: Hmm. But using the default handler would send the event to the wrong surface then?

Xyene: Before this change pressure also used the default handler, and surfaces _did_ get the pressure events when implicitly grabbed (just not the motion ones, and there were some more issues addressed in swaywm/sway#5303).

emersion: Hmm, I still don't get why this change is required. If events are sent to the right surface, then `state->focused == state->original` should be true, right?

Xyene: I think `state->focused` is misleadingly named, it's not actually the surface that is focused (as far as sway knows), it's actually "which surface *would* have focus if we weren't in an implicit grab", i.e. the surface currently under the pen.

https://github.com/swaywm/wlroots/blob/973cb19ab643649bdb32a2aa5e09bbe727205fb6/types/tablet_v2/wlr_tablet_v2_tool.c#L767-L779

emersion: Ah, indeed! Thanks for clearing this up.

Issue #1506, opened 2019-01-26
Make grabs more useful


emersion:
Grabs are currently used only for drag-and-drop. They could be useful to a lot more things:

* Pointer constraints (https://github.com/swaywm/wlroots/issues/1501)
* Move/resize windows (those would be grab interfaces implemented in the compositor)
* Button grabs (while pressing a button, focus is locked)

Maybe current grab APIs could be extended for some of these use-cases. Maybe we need to introduce cursor grabs.

Laaas:
I do not think extending pointer grabs is doable easily. `wlr_seat` and `wlr_cursor` are fundamentally unrelated. However, making cursor grabs may entail much of the same work as just extending pointer grabs.

Laaas:
I thought about something: You can't have multiple pointer grabs, so if we just extended pointer grabs, then you wouldn't be able to have a constraint active while you have another grab active, which would be pretty unfortunate.

emersion:
>I do not think extending pointer grabs is doable easily.

Depends the use-case. Buttons grabs would be pointer grabs, not cursor grabs.

PR #262, opened 2017-10-11
wlr-data-device

acrisci:
In this PR, I'm going to make the clipboard actually work.

I'm going to go a different way with this interface to make drag and drop a bit easier to do later.

TODO:

* [x] basic clipboard
* [x] drag and drop
* [x] drag and drop icons
* [x] handle multiple data devices bound to a seat
* [ ] <s>xwayland clipboard sync</s> (needs another pr)
* [ ] <s>cleanup (refactor seat functions into seat?)</s> (needs another pr)
* [x] seat selection signal

PR #2162, opened 2020-05-06
Fix #2161 and Sway#5294

### LINKED ISSUE #2161: Crash when menu popup open and cursor gets out quickly
I got a consistent crash when open Thunar, open the menu and down some levels and move suddenly the cursor to the rightmost edge of the screen. I got the coredump and also managed to add a simple check on the code.

```
Core was generated by `sway'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f8a89e9cc62 in xdg_pointer_grab_enter (grab=0x55ecb135dbc8, surface=0x0, sx=0, sy=0)
    at ../types/xdg_shell/wlr_xdg_popup.c:21
21	../types/xdg_shell/wlr_xdg_popup.c: No existe el fichero o el directorio.
[Current thread is 1 (Thread 0x7f8a88f4a940 (LWP 9457))]
(gdb) bt
#0  0x00007f8a89e9cc62 in xdg_pointer_grab_enter (grab=0x55ecb135dbc8, surface=0x0, sx=0, sy=0)
    at ../types/xdg_shell/wlr_xdg_popup.c:21
#1  0x000055ecaf02b16b in handle_pointer_motion
    (seat=0x55ecb0ded2b0, time_msec=4570544, dx=<optimized out>, dy=<optimized out>)
    at ../sway/sway/input/seatop_default.c:463
#2  0x000055ecaf0234b8 in pointer_motion
    (cursor=0x55ecb0ded8d0, time_msec=4570544, device=0x55ecb111de80, dx=50, dy=0, dx_unaccel=25, dy_unaccel=0) at ../sway/sway/input/cursor.c:324
#3  0x000055ecaf02359e in handle_pointer_motion_relative
    (listener=<optimized out>, data=<optimized out>) at ../sway/sway/input/cursor.c:332
#4  0x00007f8a89ebbece in wlr_signal_emit_safe (signal=<optimized out>, data=0x7ffd32aba7d0)
    at ../util/signal.c:27
#5  0x00007f8a89ebbece in wlr_signal_emit_safe (signal=<optimized out>, data=data@entry=0x7ffd32aba7d0)
    at ../util/signal.c:27
#6  0x00007f8a89e844ba in handle_pointer_motion (event=<optimized out>, libinput_dev=<optimized out>)
    at ../backend/libinput/pointer.c:41
#7  0x00007f8a89e8326c in handle_libinput_readable
    (fd=<optimized out>, mask=<optimized out>, _backend=0x55ecb069d250)
    at ../backend/libinput/backend.c:41
#8  0x00007f8a89f09faa in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#9  0x00007f8a89f084e7 in wl_display_run () at /usr/lib/libwayland-server.so.0
#10 0x000055ecaf00c59d in main (argc=1, argv=0x7ffd32ababe8) at ../sway/sway/main.c:409
(gdb) 
```

I changed the line 21 to add a simple check 
Original line:
`	if (wl_resource_get_client(surface->resource) == popup_grab->client) {`
Change:
`	if (surface != NULL && wl_resource_get_client(surface->resource) == popup_grab->client) {`

With this change, I can not longer reproduce the crash.

alejor:
Add a check for surface != NULL inside xdg_pointer_grab_enter()

### DIFF
--- a/types/xdg_shell/wlr_xdg_popup.c
+++ b/types/xdg_shell/wlr_xdg_popup.c
@@ -18,7 +18,7 @@ static void xdg_popup_grab_end(struct wlr_xdg_popup_grab *popup_grab) {
 static void xdg_pointer_grab_enter(struct wlr_seat_pointer_grab *grab,
 		struct wlr_surface *surface, double sx, double sy) {
 	struct wlr_xdg_popup_grab *popup_grab = grab->data;
-	if (wl_resource_get_client(surface->resource) == popup_grab->client) {
+	if (surface != NULL && wl_resource_get_client(surface->resource) == popup_grab->client) {
 		wlr_seat_pointer_enter(grab->seat, surface, sx, sy);
 	} else {
 		wlr_seat_pointer_clear_focus(grab->seat);


tchebb:
I think this fix will work, but there's an underlying architectural issue that it doesn't address. Here's what's happened, as far as I can tell:

Prior to #183, compositors used two functions to tell wlroots about pointer movement between surfaces:
1. `wlr_seat_pointer_enter()`, called when the pointer moved over a new surface, caused wlroots to mark that surface as focused, notify it that the pointer had entered it, and notify the previously-focused surface (queried from internal state) that the pointer had left it.
2. `wlr_seat_pointer_clear_focus()`, called when the pointer left a surface but did not immediately enter a new one, caused wlroots to do an abridged version of the above: it notified the previously-focused surface that the pointer had left it and unmarked that surface as focused, but it did not send any entry notifications (since no surface had been entered).

`wlr_seat_pointer_clear_focus()` was, and still is, internally implemented by calling `wlr_seat_pointer_enter()` with a `NULL` surface argument. However, this seems intended to have been purely an implementation detail: `wlr_seat_pointer_enter()`'s documentation did not specify that you can pass a `NULL` surface, no external clients did, and the very existence of `wlr_seat_pointer_clear_focus()` indicates that `wlr_seat_pointer_enter()` was not intended to be used for clearing focus.

Eventually, #183 came along with commit 17b134e1783a ("wlr-seat: pointer grab interface"). In order for wlroots to support pointer grabs (used for xdg_shell popups and, since #262, drag-and-drop), this commit added an extra layer of indirection. Now, instead of calling `wlr_seat_pointer_enter()` directly, clients were supposed to call a new function `wlr_seat_pointer_notify_enter()`, which calls different `enter()` functions depending on the type of grab that's active. If no grab is active, it just delegates back to `wlr_seat_pointer_enter()`. Clients switched to the new function, and things seemed good.

However, there was an oversight: #183 added indirected versions of `wlr_seat_pointer_enter()` and several other functions, but `wlr_seat_pointer_clear_focus()` was not among them. Clients continued to call that function directly to clear focus regardless of whether a grab was active. What this meant was that grabs were not notified when the pointer left a surface and did not immediately enter a new one (or when focus was cleared due to special circumstances).

This brings us, finally, to the commit that @Xyene identified as the cause of swaywm/sway#5294. In swaywm/sway#5220, @nickdiego noticed weird behavior of drag-and-drop grabs caused by—surprise!—missing events when the pointer left a surface. In order to fix this, they wrote swaywm/sway#5222, adding a call to `wlr_seat_pointer_notify_enter()` with a `NULL` surface argument. This was the first time (at least in sway) the function was called that way, and although both `wlr_seat_pointer_enter()` (as discussed previously) and the drag-and-drop enter hook (`drag_handle_pointer_enter()`) can gracefully handle a `NULL` surface, the xdg_shell popup enter hook (`xdg_pointer_grab_enter()`) cannot and segfaults.

So, how should we fix it? One approach is the one taken by this commit, which is to ensure that all grab enter hooks can treat a `NULL` surface as a focus clear. I think this is a fine approach, but there are several other things I'd also want to see if we choose it:
1. The fact that `wlr_seat_pointer_enter()` can take a `NULL` surface should be explicitly documented in `wlr_seat.h`.
2. All `wlr_seat_pointer_clear_focus()` calls in clients should be replaced with `wlr_seat_pointer_notify_enter(wlr_seat, NULL, 0, 0)` calls. Note that swaywm/sway#5222 didn't do this: it added the latter but did not remove the former. This is wrong, IMO, since a grab may wish to fully override the default behavior.
3. For consistency, `wlr_seat_pointer_clear_focus()` should perhaps be deprecated entirely in favor of `wlr_seat_pointer_enter(wlr_seat, NULL, 0, 0)`.

The other approach we could take is to explicitly prohibit passing a `NULL` surface to `wlr_seat_pointer_notify_enter()` and add a new indirected function, `wlr_seat_pointer_notify_clear_focus()`. If we take this path, we'd need to
1. Plumb up `wlr_seat_pointer_notify_clear_focus()` like all the other indirected functions, and add implementations of it to existing grabs (i.e. xdg_shell popup and drag-and-drop).
2. Replace all calls to `wlr_seat_pointer_clear_focus()` in clients with calls to `wlr_seat_pointer_notify_clear_focus()`,
3. Add an `assert(surface)` call to `wlr_seat_pointer_notify_enter()` to catch misbehaving clients early.

I personally like the second approach a bit better, and I'll probably attempt to write a PR implementing it this weekend barring any objections. As a stopgap, I think we should revert swaywm/sway#5222, since a rare drag-and-drop bug is preferable to a segfault every time your cursor leaves a surface while a popup is up.

emersion:
I agree with @tchebb. I also slightly prefer the second approach because the x/y parameters don't make sense for leave events. Since we're going to break the API, it would be nice to rename `clear_focus` to `leave` to align the wording with the Wayland protocol.

emersion:
Also see https://github.com/swaywm/wlroots/issues/1506

alejor:
tchebb, I agree absolutely with you, Indeed I just try to fix it as sway is my daily driver and any issue that can crash it makes me feeling bad, but I don't know any clue about the architecture. I already known that was a ugly workaround, but any is better to segfault and loss the daily work. Thanks for your advice and hope you can figure out the best way to tackle this. Cheers!

nickdiego:
Wow! thanks for the in-depth analysis and sorry for not being able to look into this regression introduced by my change. Quite busy week here :/ TBH, I hesitated a bit before sending that patch without any kind of automated test / CI to "ensure" it was regressions-free.

> The other approach we could take is to explicitly prohibit passing a NULL surface to wlr_seat_pointer_notify_enter() and add a new indirected function, wlr_seat_pointer_notify_clear_focus().

IMHO, this sounds better, had even asked about such a function to @emersion as I was a bit surprised it didn't exist yet :P 

Thanks for looking into it. I still have several DND issues on the pipeline to report / try to fix :( hope to have some time for it in the coming weeks

tchebb:
I've posted #2217 and swaywm/sway#5368 to implement the second solution I proposed above.

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2217

Issue #2150, opened 2020-05-01
Re-enabling output sometimes fail


jbeich:
Regressed by https://github.com/swaywm/wlroots/pull/2118. See [debug log](https://github.com/swaywm/wlroots/files/4566180/sway.log).

### Steps to reproduce
1. Start Sway
2. Disable display
3. Enable display
4. Repeat (2) and (3) until black screen

### Actual result
Display wakes up but with no signal goes back to sleep. On console:
```
[main.c:168] Destroying output DP-1 (...)
[ERROR] [backend/drm/renderer.c:379] Tried to acquire an FB with a NULL BO
[ERROR] [backend/drm/renderer.c:379] Tried to acquire an FB with a NULL BO
[ERROR] [backend/drm/drm.c:754] Failed to initialize renderer on connector 'DP-1': initial page-flip failed
[ERROR] [backend/drm/drm.c:835] Failed to initialize renderer for plane
[sway/config/output.c:423] Failed to commit output DP-1
[ERROR] [backend/drm/renderer.c:379] Tried to acquire an FB with a NULL BO
```

### Expected result
Display works like before sleep

### Environment
```
$ cc --version
FreeBSD clang version 10.0.0 (git@github.com:llvm/llvm-project.git llvmorg-10.0.0-0-gd32170dbd5b)
Target: x86_64-unknown-freebsd13.0
Thread model: posix
InstalledDir: /usr/bin

$ sway --version
sway version 1.2-rc1-334-g5e5e5f2e
wlroots version 0.10.0-107-g61d6408f

$ cat /tmp/sway.conf
bindsym Ctrl+Alt+BackSpace exit
bindsym Down output * dpms off
bindsym Up output * dpms on

output * background #15c6ac solid_color
```

emersion:
Can you try https://github.com/swaywm/wlroots/pull/2168?

jbeich:
> Can you try #2168?

After applying I can't reproduce anymore. Thanks!

PR #2168, opened 2020-05-07
backend/drm: disable cursor in dealloc_crtc

emersion:
dealloc_crtc was destroying GBM surfaces, but the cursor_enabled flag
was left as-is. When re-enabling the output, atomic_crtc_pageflip would
try enabling the cursor plane, but would fail because no framebuffer is
available.

Closes: https://github.com/swaywm/wlroots/issues/2150


ddevault:
Thanks!

Issue #1899, opened 2019-11-05
Stop including EGL/eglmesaext.h


emersion:
This was a mistake on Mesa's side. We'll need to revert https://github.com/swaywm/wlroots/commit/d113e48a2a32542fe6e12f1759f07888364609bf.

See https://gitlab.freedesktop.org/wayland/weston/merge_requests/289#note_283447

ascent12:
https://github.com/KhronosGroup/EGL-Registry/pull/95 was merged. We should be able to drop it the next mesa release if they pull in the new EGL headers.

emersion:
* https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4953
* https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/225

emersion:
Also need to revert https://github.com/swaywm/wlroots/pull/2171

emersion:
Done in https://github.com/swaywm/wlroots/pull/2222.

PR #2171, opened 2020-05-07
Make EGL/eglmesaext.h header inclusion optional, only if present. Makes wlroots compile on any egl/gles driver.

jolivain:
This pull request makes the EGL/eglmesaext.h header included only if present. The EGL_WL_bind_wayland_display is now standard, and some EGL/GLES drivers are providing it without the eglmesaext.h file.

Those patches were tested in Buildroot, compiled with imx-gpu-viv gpu driver, on an i.MX8M Mini EVK board (case without eglmesaext.h).
They were also tested against a Mesa 20.0.6 (case with eglmesaext.h).

### DIFF
--- a/include/wlr/render/egl.h
+++ b/include/wlr/render/egl.h
@@ -16,10 +16,14 @@
 #define EGL_NO_X11
 #endif
 
+#include <wlr/config.h>
+
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
+#if WLR_HAS_EGLMESAEXT_H
 // TODO: remove eglmesaext.h
 #include <EGL/eglmesaext.h>
+#endif
 #include <pixman.h>
 #include <stdbool.h>
 #include <wayland-server-core.h>
--- a/include/wlr/config.h.in
+++ b/include/wlr/config.h.in
@@ -1,6 +1,8 @@
 #ifndef WLR_CONFIG_H
 #define WLR_CONFIG_H
 
+#mesondefine WLR_HAS_EGLMESAEXT_H
+
 #mesondefine WLR_HAS_LIBCAP
 
 #mesondefine WLR_HAS_SYSTEMD
--- a/meson.build
+++ b/meson.build
@@ -86,6 +86,7 @@ conf_data.set10('WLR_HAS_X11_BACKEND', false)
 conf_data.set10('WLR_HAS_XWAYLAND', false)
 conf_data.set10('WLR_HAS_XCB_ERRORS', false)
 conf_data.set10('WLR_HAS_XCB_ICCCM', false)
+conf_data.set10('WLR_HAS_EGLMESAEXT_H', false)
 
 # Clang complains about some zeroed initializer lists (= {0}), even though they
 # are valid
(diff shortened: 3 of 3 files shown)


include/wlr/render/egl.h
```diff
@@ -16,10 +16,13 @@
 #define EGL_NO_X11
 #endif
 
+#include <wlr/config.h>
+
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
-// TODO: remove eglmesaext.h
```
emersion, line 21: This TODO should be retained. Including `eglmesaext.h` is still a hack.

render/gles2/texture.c
```diff
@@ -214,7 +214,9 @@ struct wlr_texture *wlr_gles2_texture_from_wl_drm(struct wlr_egl *egl,
 		texture->has_alpha = false;
 		break;
 	case EGL_TEXTURE_RGBA:
+#if defined(EGL_WL_bind_wayland_display) && defined(EGL_TEXTURE_EXTERNAL_WL)
```
emersion, line 217: Since this is now in Khronos' registry, why do we need an `#ifdef` here?

include/wlr/render/egl.h
jolivain: I added it back. Thanks.

render/gles2/texture.c
jolivain: It's in Khronos registry, but it's now in all gpu drivers.  Mesa still publish this extension in eglmesaext.h.  It'll be updated at some point.  Each gpu driver updates Khronos headers at different speed. To make wlroots compile on a wider range of drivers (older extension version, older drivers, very recent driver with up-to-date khronos header, etc.), the define protect its usage, only when it's defined.

emersion:
More background on why we include `eglmesaext.h`: https://github.com/swaywm/wlroots/issues/1899

jolivain:
eglmesaext.h is meant for unofficial (non Khronos registered) extension.  Since it was included in some Mesa releases, it'll take some time before it can really be removed (e.g. Mesa in LTS distros, ...). This patch, including the file only if present will allow a smoother transition.

For info, the latest mesa-20.1.0-rc2 tag still expose the extension in eglmesaext.h:
https://gitlab.freedesktop.org/mesa/mesa/-/blob/mesa-20.1.0-rc2/include/EGL/eglmesaext.h#L38

escalade:
Fails to compile with libmali:

https://pastebin.com/7yHbxh2U

emersion:
>Fails to compile with libmali

Looks like unrelated GBM issues.

render/gles2/texture.c
emersion: But aren't there a lot more symbols that may not be defined, like `eglBindWaylandDisplayWL`?

I'd prefer not to clutter the code with these `#ifdef`s. If the drivers are outdated, they should get updated. If the drivers are closed-source, we won't do anything to support them.

jolivain: Runtime symbols like eglBindWaylandDisplayWL are normally protected by the dynamic resolution with eglGetProcAddress(), and scanning in EGL_EXTENSIONS string. Compilation time definitions (IDs, typedefs, etc.) can also be protected with ifdefs on the extension (e.g. #ifdef EGL_WL_bind_wayland_display). Here, the issue is that some drivers claim an older version of the extention, with a missing definition.

If wlroots wants to focus only on Mesa (and eventually only Khronos conformant closed-source drivers), then I agree with you this quirk can be removed. I'll update the pull-request by dropping this change, and I'll see if I can fix that issue directly with GPU driver vendors.

emersion: We want to focus on any open-source Khronos conformant GL implementation. We won't offer support for closed-source drivers because we can't go down the stack and fix issues in the GL implementation in case of a bug. Debugging a black box is counter-productive.

PR #2172, opened 2020-05-08
input/keyboard: expose keymap matching helper

Xyene:
sway needs this logic too, and currently ships a version that has fallen
behind in terms of bugfixes (b1a63bc). It might be reasonable for
wlroots to expose this functionality instead.


RedSoxFan:
Thanks!

PR #2175, opened 2020-05-08
xwayland: split server

emersion:
Split the server part of wlr_xwayland into wlr_xwayland_server. This
allows compositors to implement their own XWM when wlroots' isn't a good
fit.

~~Sway PR: https://github.com/swaywm/sway/pull/5314~~ A Sway PR isn't necessary for this change.

* * *

Breaking change: `wlr_xwayland.display_name` is now a `const char *` field. Most compositors should still be able to use it like before. Server-specific fields have been moved from `wlr_xwayland` to `wlr_xwayland_server`.

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -34,6 +34,7 @@ struct wlr_xwayland_server {
 	int x_fd[2];
 	struct wl_event_source *x_fd_read_event[2];
 	bool lazy;
+	bool enable_wm;
 
 	struct wl_display *wl_display;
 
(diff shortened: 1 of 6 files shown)


emersion:
This is now ready for review.

Question: should we keep `wlr_xwayland.display_name` and copy the value from `wlr_xwayland_server.display_name` instead of adding `wlr_xwayland_display_name`?

emersion:
Actually, changed my mind and made `wlr_xwayland.display` a `const char *`.

ddevault:
Thanks!

PR #2178, opened 2020-05-09
backend/drm: simplify atomic commit logic

emersion:
We don't need a per-CRTC atomic request anymore. Let's make the request
per-commit so that it's easier to debug.

This is also groundwork for supporting wlr_output_test properly.

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -9,57 +9,42 @@
 
 struct atomic {
 	drmModeAtomicReq *req;
-	int cursor;
 	bool failed;
 };
 
-static void atomic_begin(struct wlr_drm_crtc *crtc, struct atomic *atom) {
-	if (!crtc->atomic) {
-		crtc->atomic = drmModeAtomicAlloc();
-		if (!crtc->atomic) {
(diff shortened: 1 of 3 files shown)


backend/drm/atomic.c
```diff
@@ -9,57 +9,42 @@
...
-	atom->cursor = drmModeAtomicGetCursor(atom->req);
-	atom->failed = false;
-}
-
-static bool atomic_end(int drm_fd, uint32_t flags, struct atomic *atom) {
-	if (atom->failed) {
-		return false;
-	}
+static void atomic_begin(struct atomic *atom) {
+	memset(atom, 0, sizeof(*atom));
```
panaman67, line 16: Because struct atomic has the 2 fields, and you never access them before you set them, this can be removed and you can add `atom->failed = false` after the if right? I could be wrong....

emersion: Yes, but I prefer to memset the whole struct. It's too easy to add a new field in the struct and forget to initialize it everywhere. That's also the reason why we use calloc instead of malloc throughout the codebase.

panaman67: That makes complete sense.... my bad.

PR #2180, opened 2020-05-10
Null check surface in xdg_pointer_grab_enter

tadeokondrak:
Before this change, opening a popup then moving the cursor somewhere
without a surface would crash the compositor.


Xyene:
Thanks for the PR, and for looking into this! This fix is also being discussed in https://github.com/swaywm/wlroots/pull/2162, the reason it's not yet merged is there is some discussion as whether this is the best way to go about this or not.

tadeokondrak:
Ah, hadn't seen that one.

PR #2185, opened 2020-05-11
backend/drm: fix segfault in drm_crtc_page_flip

emersion:
When no cursor plane is available, drm_crtc_page_flip would segfault.


progandy:
Looks good.

Issue #2166, opened 2020-05-06
Plane 0 doesn't support format 0x34325258


zlandau:
While running from the lastest master versions of wlroot and sway, I noticed my logs are flooded with messages like this:

```
May 06 14:13:01 buffalo sway[429005]: 15:33:42.651 [ERROR] [backend/drm/renderer.c:215] Plane 0 doesn't support format 0x34325258
May 06 14:13:01 buffalo sway[429005]: 15:33:42.651 [ERROR] [backend/drm/drm.c:934] Cannot allocate cursor resources
```

The messages appear every maybe 6 times per second.

According to drm_info, it does support that format:

```
    ├───Plane 0
    │   ├───Object ID: 41
    │   ├───CRTCs: {0}
    │   ├───Formats:
    │   │   ├───XRGB8888 (0x34325258)
    │   │   └───ARGB8888 (0x34325241)
    │   └───Properties
    │       └───"type" (immutable): enum {Overlay, Primary, Cursor} = Primary
```

I don't notice any downstream behavior that isn't working due to these messages, but the flooding of the logs doesn't seem ideal, at the very least.

# System Information

wlroots revision: swaywm/wlroots@6357e166f9a31f0c828ea5bb0000627c15442505
swaywm/sway@9cda5a5bd62efe161217481147121d5de8a8dc20
Device: PCI 1002:6758 Advanced Micro Devices, Inc. [AMD/ATI] Turks XT [Radeon HD 6670/7670]
Driver: Driver: radeon (ATI Radeon) version 2.50.0 (20080528)
Full output of drm_info: https://gist.github.com/0a50bfb8c538997e5c8d7f989e611da7

emersion:
This happens because this device doesn't advertise a cursor plane.

superna9999:
@emersion does wlroots mandates a cursor plane ? AFAIK is was not the case a few months ago

emersion:
No, this is a bug.

L-as:
I can run both sway and the pointer example fine if I set WLR_NO_HARDWARE_CURSORS=1, but I assume this is supposed to be detected by wlroots and done automatically if no cursor plane exists?

L-as:
commit [8da9d9679e4cad640b0f0a518ca1529c8cdecbe0](https://github.com/swaywm/wlroots/commit/8da9d9679e4cad640b0f0a518ca1529c8cdecbe0) seems to have introduced this behaviour @emersion @ascent12

[...]

PR #2186, opened 2020-05-11
backend/drm: remove missing cursor plane workaround

emersion:
We have a workaround for legacy drivers that support drmModeSetCursor
without exposing a cursor plane. It doesn't work anymore now that we've
moved to a more atomic interface.

Let's just remove this workaround and fallback to software cursors.

Closes: https://github.com/swaywm/wlroots/issues/2166

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -941,14 +941,8 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 	}
 
 	struct wlr_drm_plane *plane = crtc->cursor;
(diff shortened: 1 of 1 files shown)


L-as:
I just tested this and it works great for me, but FYI commit [70883fd10beba4f78e3325eb6e38750360502f58](https://github.com/swaywm/wlroots/commit/70883fd10beba4f78e3325eb6e38750360502f58) causes my setup not to work again. I'll make a new issue for that.

PR #2191, opened 2020-05-12
input/pointer: notify compositor when constraint region changes

Xyene:
This allows a compositor to know when warping back into the region is
appropriate.

Refs swaywm/sway#5268.

### DIFF
--- a/types/wlr_pointer_constraints_v1.c
+++ b/types/wlr_pointer_constraints_v1.c
@@ -116,6 +116,7 @@ static void pointer_constraint_commit(
 	}
 	constraint->current.committed |= constraint->pending.committed;
 
+	bool updated_region = !!constraint->pending.committed;
 	constraint->pending.committed = 0;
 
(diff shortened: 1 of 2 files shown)


types/wlr_pointer_constraints_v1.c
```diff
@@ -126,6 +127,10 @@ static void pointer_constraint_commit(
 		pixman_region32_copy(&constraint->region,
 			&constraint->surface->input_region);
 	}
+
+	if (updated_region) {
+		wlr_signal_emit_safe(&constraint->events.set_region, constraint);
```
emersion, line 132: Can we set the signal data to NULL? This would allow us to change the data to an event struct in the future (e.g. with the `pending.committed` bitfield) without breaking the API.

Xyene: Hadn't thought of that, but that makes sense -- done :+1:

PR #2195, opened 2020-05-14
Make wlr_xcursor_manager_load() return a bool

ifreund:
This is currently inconsistent with the rest of the library and a bit of
a footgun for new compositors. However, this breaks the API in a very
unfortunate way for existing compositors.

* * *

Breaking change: `wlr_xcursor_manager_load` now returns a bool, making it consistent with other wlroots functions. Compositors that check the return value need to invert their check.

### DIFF
--- a/include/wlr/types/wlr_xcursor_manager.h
+++ b/include/wlr/types/wlr_xcursor_manager.h
@@ -46,7 +46,7 @@ void wlr_xcursor_manager_destroy(struct wlr_xcursor_manager *manager);
 /**
  * Ensures an xcursor theme at the given scale factor is loaded in the manager.
  */
-int wlr_xcursor_manager_load(struct wlr_xcursor_manager *manager,
+bool wlr_xcursor_manager_load(struct wlr_xcursor_manager *manager,
 	float scale);
 
 /**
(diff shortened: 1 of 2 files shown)


ifreund:
So with regard to breaking existing compositors:
- sway doesn't check the return value ([link](https://github.com/swaywm/sway/search?q=wlr_xcursor_manager_load&unscoped_q=wlr_xcursor_manager_load))
- hikari doesn't check the return value ([link](https://hub.darcs.net/raichoo/hikari/browse/src/server.c#461))
- wayfire doesn't check the return value ([link](https://github.com/WayfireWM/wayfire/search?q=wlr_xcursor_manager_load&unscoped_q=wlr_xcursor_manager_load))
- cardboard doesn't check the return value ([link](https://gitlab.com/search?utf8=%E2%9C%93&search=wlr_xcursor_manager_load&group_id=&project_id=16442283&search_code=true&repository_ref=master&nav_source=navbar))
- tinywl doesn't check the return value ([link](https://github.com/swaywm/wlroots/blob/42e485dcc306f386aa8d3b8907c77ea3316360d5/tinywl/tinywl.c#L900))

This is obviously not an exhaustive list of compositors using wlroots, but I think it shows that this might not be as severe of a breaking change in effect as it seems at first glance.

emersion:
Thanks!

PR #2196, opened 2020-05-14
output: make wlr_output_set_gamma atomic

emersion:
wlr_output_set_gamma is now double-buffered and applies the gamma LUT on
the next output commit.

* * *

Breaking changes:

* `wlr_output_set_gamma` is now double-buffered and applied on the next output commit
* `wlr_output_impl.set_gamma` is gone. Backends now need to read the pending output state on commit instead.

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -71,16 +71,11 @@ static bool create_mode_blob(struct wlr_drm_backend *drm,
 }
 
 static bool create_gamma_lut_blob(struct wlr_drm_backend *drm,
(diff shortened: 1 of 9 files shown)


ddevault:
Thanks!

betaboon:
@emersion now my systemlog is being spammed with `[ERROR] [backend/drm/atomic.c:177] setting gamma LUT 1024`.

emersion:
Yeah this is a remnant from debugging code I forgot to remove. Removed in https://github.com/swaywm/wlroots/pull/2200

betaboon:
thanks for the headsup :D

PR #2200, opened 2020-05-18
backend/drm: rollback atomic blobs

emersion:
If the atomic commit fails or is test-only, rollback
wlr_drm_crtc.{mode_id,gamma_lut} blob IDs to their previous value. This
prevents the next commits from failing or applying test-only state.


ddevault:
Thanks!

PR #2202, opened 2020-05-18
examples: make output-power-management oneshot by default

ammen99:
As discussed in IRC, with oneshot mode, we can use this example client as a keybinding or anything like that. Otherwise, the user will have to manually kill the client each time.

### DIFF
--- a/examples/output-power-management.c
+++ b/examples/output-power-management.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include <errno.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


examples/output-power-management.c
```diff
@@ -72,22 +73,27 @@ static const struct wl_registry_listener registry_listener = {
 static const char usage[] = "usage: output-power-management [options...]\n"
 	"  -h: show this help message\n"
 	"  -e: turn outputs on\n"
-	"  -d: turn outputs off\n";
+	"  -d: turn outputs off\n"
+	"  -m: turn monitor mode on\n";
```
emersion, line 77: "monitor mode" is a little bit confusing for a tool that configures outputs. Maybe we can change the description to "don't exit and monitor changes"?

ammen99: what about "don't exit and continuously print changes" ?

emersion: Sure, looks good!

agx: or '-w': continousely watch for power mgmt mode chnges. Apart from that looks good to me.

ammen99: I made it as @agx suggested (with -`w`)

ammen99:
> LGTM, can you squash these changes?

Yes, done :)

emersion:
Thanks!

Issue #2197, opened 2020-05-15
Don't leave bound EGL contexts behind


emersion:
In wlroots we often bind to an EGL context, do EGL operations, and then leave the context bound. Apart from making it harder to debug cases where we forget to bind the context, this also causes issues for multi-threaded compositors (since `eglMakeCurrent` fails when another thread has the context bound, even if correct synchronization is implemented).

PR #2206, opened 2020-05-19
render/egl: introduce wlr_egl_unset_current

emersion:
This function can be called after wlr_egl_make_current to cleanup the
EGL context. This avoids having lingering EGL contexts that make things
work by chance.

Closes: https://github.com/swaywm/wlroots/issues/2197


ddevault:
Thanks!

PR #2207, opened 2020-05-19
Fix -Wreturn-type warnings

emersion:
When calling assert(0) instead of returning a value, -Wreturn-type
warnings are triggered because assertions can be disabled. Replace these
assertions with abort().


ddevault:
Thanks!

Issue #2209, opened 2020-05-19
Segfault with latest egl context change


JonnyMako:
With https://github.com/swaywm/wlroots/commit/1edc42157bbb01b25ec27750f883fd84ee5908d6 I'm getting a segfault at startup.

[gdb.txt](https://github.com/swaywm/wlroots/files/4651358/gdb.txt)

emersion:
Can you try https://github.com/swaywm/wlroots/pull/2210?

JonnyMako:
That is hanging. I'll try to get info from log.

JonnyMako:
Log attached.
[sway.log](https://github.com/swaywm/wlroots/files/4652162/sway.log)

emersion:
```
00:00:02.470 [ERROR] [EGL] command: eglSwapBuffers, error: 0x300d, message: "eglSwapBuffers"
00:00:02.470 [ERROR] [render/egl.c:449] eglSwapBuffers failed
00:00:02.470 [ERROR] [backend/drm/renderer.c:276] Failed to swap buffers
00:00:02.470 [ERROR] [backend/drm/drm.c:727] Failed to initialize renderer on connector 'eDP-1': initial page-flip failed
00:00:02.470 [ERROR] [backend/drm/drm.c:816] Failed to initialize renderer for plane
00:00:02.471 [sway/config/output.c:423] Failed to commit output eDP-1
```

JonnyMako:
On exit, I get:
`[backend/drm/atomic.c:36] eDP-1: Atomic commit failed (modset): Permission deniedl page-flip failed`

emersion:
Abort fixed, but there's still this `eglSwapBuffers` failure to address.

tchebb:
I'm seeing the `eglSwapBuffers` failure and hang on latest master as well. Using a single GPU (Intel Haswell integrated graphics), kernel 5.6.13. Reverting commit d28a7da95d1a ("backend/drm: add missing wlr_egl_unset_current") gets rid of the problem for me.

emersion:
Can you try https://github.com/swaywm/wlroots/pull/2212?

GreyXor:
> Can you try #2212?

It's fixed here with #2212

emersion:
I've actually come up with a Better(tm) solution, can you try https://github.com/swaywm/wlroots/pull/2213?

hensur:
I just ran into this as well. I didn't try #2212, but I can report that #2213 fixes the issue for me.

[...]

PR #2212, opened 2020-05-20
backend/drm: don't unset EGL context before swapping buffers

emersion:
drm_fb_lock_surface swaps buffers. We need to make sure that the EGL
context is current before calling wlr_egl_swap_buffers.

Move the wlr_egl_unset_current call into drm_fb_lock_surface to avoid
other surprises like this.

Fixes: d28a7da95d1a ("backend/drm: add missing wlr_egl_unset_current")
Closes: https://github.com/swaywm/wlroots/issues/2209

* * *

This is ready for review, but let's wait for user feedback before merging it, in case I missed something again.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -556,8 +556,6 @@ static bool drm_connector_commit(struct wlr_output *output) {
 		}
 	}
 
-	wlr_egl_unset_current(&drm->renderer.egl);
-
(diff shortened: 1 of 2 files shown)


emersion:
Actually I think https://github.com/swaywm/wlroots/pull/2213 is a better alternative solution.

PR #2213, opened 2020-05-20
render/egl: unset current context after swapping buffers

emersion:
After swapping buffers, it doesn't make sense to perform more rendering
operations. Unset the context to reflect this.

This commit makes it so the context is always only current between
wlr_egl_make_current and wlr_egl_swap_buffers.

This is an alternative to [1].

[1]: https://github.com/swaywm/wlroots/pull/2212

Closes: #2209

* * *

This is ready for review, ~~but let's wait for user feedback before merging it, in case I missed something again~~.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -556,8 +556,6 @@ static bool drm_connector_commit(struct wlr_output *output) {
 		}
 	}
 
-	wlr_egl_unset_current(&drm->renderer.egl);
-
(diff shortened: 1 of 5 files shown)


emersion:
Confirmed it fixes the issue: https://github.com/swaywm/wlroots/issues/2209#issuecomment-631452634

ddevault:
Thanks!

PR #2217, opened 2020-05-21
Allow keyboard and pointer grabs to hook clear_focus()

tchebb:
This is necessary for some grabs, which currently have no way of knowing when the pointer/keyboard focus has left a surface. For example, without this, a drag-and-drop grab can erroneously drop into a window that the cursor is no longer over.

This is the plumbing needed to properly fix swaywm/sway#5220. The existing fix, swaywm/sway#5222, relies on every grab's `enter()` hook allowing a `NULL` surface. This is not guaranteed by the API and, in fact, is not the case for the xdg-shell popup grab and results in a crash when the cursor leaves a surface and does not immediately enter another one while a popup is open (#2161).

This fix also adds an assertion to wlr_seat_pointer_notify_enter() that ensures it's never called with a `NULL` surface. This will make Sway crash much more until it fixes its usage of the API, so we should land this at the same time as a fix in Sway (which I haven't posted yet).

Also clean up the header declarations a bit in separate commits.

Unanswered questions:
1. Should we rename the `clear_focus` functions to `leave` like suggested by @emersion in #2162? that seems like a more breaking change to me, and I'm not sure it's worth it.
2. Do `wlr_seat_touch_point_focus()` and `wlr_seat_touch_point_clear_focus()` also need the same treatment? I have no touch device to test with, and these functions aren't used by Sway or by tinywl. It looks like they were/are used by rootston, though.

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -391,19 +391,6 @@ void wlr_seat_pointer_send_axis(struct wlr_seat *wlr_seat, uint32_t time_msec,
 void wlr_seat_pointer_send_frame(struct wlr_seat *wlr_seat);
 
 /**
- * Start a grab of the pointer of this seat. The grabber is responsible for
- * handling all pointer events until the grab ends.
- */
-void wlr_seat_pointer_start_grab(struct wlr_seat *wlr_seat,
-		struct wlr_seat_pointer_grab *grab);
-
-/**
- * End the grab of the pointer of this seat. This reverts the grab back to the
- * default grab for the pointer.
- */
-void wlr_seat_pointer_end_grab(struct wlr_seat *wlr_seat);
-
-/**
  * Notify the seat of a pointer enter event to the given surface and request it
  * to be the focused surface for the pointer. Pass surface-local coordinates
  * where the enter occurred.
@@ -440,6 +427,19 @@ void wlr_seat_pointer_notify_axis(struct wlr_seat *wlr_seat, uint32_t time_msec,
 void wlr_seat_pointer_notify_frame(struct wlr_seat *wlr_seat);
 
 /**
+ * Start a grab of the pointer of this seat. The grabber is responsible for
+ * handling all pointer events until the grab ends.
+ */
+void wlr_seat_pointer_start_grab(struct wlr_seat *wlr_seat,
+		struct wlr_seat_pointer_grab *grab);
+
+/**
+ * End the grab of the pointer of this seat. This reverts the grab back to the
+ * default grab for the pointer.
+ */
+void wlr_seat_pointer_end_grab(struct wlr_seat *wlr_seat);
+
+/**
  * Whether or not the pointer has a grab other than the default grab.
  */
 bool wlr_seat_pointer_has_grab(struct wlr_seat *seat);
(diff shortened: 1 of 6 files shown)


tchebb:
cc @emersion, @alejor, @nickdiego, @Xyene

deprecated-dot:
For avoidance of doubt, is this the fix for segfaults when moving the cursor away from the context menu item without selecting anything? I look forward to it!

tchebb:
@deprecated-dot That's swaywm/sway#5368, technically, but it depends on this one and will likely land at the same time.

deprecated-dot:
@tchebb Thanks!

emersion:
>that seems like a more breaking change to me, and I'm not sure it's worth it.

Compositors need an update to use the new `notify` function instead right?

> Do wlr_seat_touch_point_focus() and wlr_seat_touch_point_clear_focus() also need the same treatment?

Hmm, yeah, probably. Let's keep this in mind, I'll open a follow-up issue once this PR is merged.

emersion:
Overall this PR's logic LGTM.

emersion:
I think I'd still prefer renaming `clear_focus` to `leave`, because `clear_focus` makes it sound like it unsets the surface focus. However with grabs that may not be the case. `leave` sounds more like an event.

tchebb:
Sorry for the late response here. A couple points regarding changing the name:
1. Are you proposing the name change only for the newly-added function, or for the existing `wlr_seat_{pointer,keyboard}_clear_focus()` as well? If the former, the inconsistency doesn't seem worth it; if the latter, that will be a more breaking change than this PR is currently, since it will break compilation for clients instead of just leaving them with incorrect grab behavior. (On the other hand, one could argue that big, obvious breakage is better than subtle breakage.)
2. I think that `leave` as a name isn't great when we already have an `enter` that sends both a leave and an enter event when called. `clear_focus` makes it more obvious that it only needs calling when no new surface has been entered. OTOH, I don't want to drag us into bikeshedding for this change, so I'll happily defer to whatever naming you prefer.

emersion:
>Are you proposing the name change only for the newly-added function, or for the existing wlr_seat_{pointer,keyboard}_clear_focus() as well?

The existing functions as well.

>I think that leave as a name isn't great when we already have an enter that sends both a leave and an enter event when called.

But then we'd have "enter" which sends more than just an enter event, and "clear_focus" that always just sends a leave event.

Or are you suggesting that we rename "enter" to something like set_focus?

tchebb:
> Or are you suggesting that we rename "enter" to something like set_focus?

If we're changing names, I think my preference would be to rename "enter" to "change_focus" and keep "clear_focus" named as is. My main question is whether the current names are bad enough to warrant breaking the build for every client, though.

emersion:
All right. Let's just merge this patch since it's consistent with the current naming and doesn't break the API. We can discuss about naming in a follow-up issue.

ammen99:
I am unsure why this isn't marked as a breaking change, things worked before, don't work anymore ... (Although from what I gathered the old behavior was not guaranteed to work?)

emersion:
What worked before? The old functions like `wlr_keyboard_clear_focus` haven't been touched and still behave the same as before.

ammen99:
I used to call `wlr_seat_pointer_notify_enter()` with `NULL`, from the PR I can read that I have been misusing it. However, it was accepting a NULL argument before, but doesn't do so anymore.

emersion:
I haven't marked the new asserts in `notify_enter` functions as a breaking change, because they were segfaulting under some circumstances.

ammen99:
Just saying this as a heads up to anybody who might have been doing the same mistake as I was :)

emersion:
But maybe a breaking change note would be nicer indeed. Will do.

deprecated-dot:
All seems well, it's nice being able to change focus after bringing up a context menu and not having sway crash. 

Thanks all, for all your work!

emersion:
For reference: this PR was based on discussions in https://github.com/swaywm/wlroots/pull/2162#issuecomment-625083795

PR #2219, opened 2020-05-23
xwayland: send focus change event unconditionally

Xyene:
This fixes issues with (at least) dialogs in Jetbrains IDEs becoming
unclickable if they ever lost focus (ref. swaywm/sway#5373). Prior to
this change, since `xwm->focus_surface` would be set prior to
`xwm_surface_activate` being called, the latter would short-circuit
immediately and not notify the application of the focus change.

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -1243,12 +1243,12 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
 	// Because of that, we allow changing focus between surfaces belonging to the
 	// same application.
 	struct wlr_xwayland_surface *requested_focus = lookup_surface(xwm, ev->event);
-	if (!xwm->focus_surface || !requested_focus ||
-			requested_focus->pid != xwm->focus_surface->pid) {
-		xwm_send_focus_window(xwm, xwm->focus_surface);
-	} else {
+	if (xwm->focus_surface && requested_focus &&
+			requested_focus->pid == xwm->focus_surface->pid) {
 		xwm->focus_surface = requested_focus;
 	}
+
+	xwm_send_focus_window(xwm, xwm->focus_surface);
 }
 
 static void xwm_handle_xcb_error(struct wlr_xwm *xwm, xcb_value_error_t *ev) {


emersion:
This basically means we let X11 apps override each other's focus.

Why does the condition fail in this case? Is the PID different?

Xyene:
Hmm, I don't think that's what this does unless I misunderstood this completely...

The original code was:

```c
	struct wlr_xwayland_surface *requested_focus = lookup_surface(xwm, ev->event);
	if (!xwm->focus_surface || !requested_focus ||
			requested_focus->pid != xwm->focus_surface->pid) {
		xwm_send_focus_window(xwm, xwm->focus_surface);
	} else {
		xwm->focus_surface = requested_focus;
	}
```

The `if` condition fails if the focus change shouldn't be allowed, in which case as per the comment, "Reset the focus to the old one if it changed."

The simplified fix for this issue is:

```c
	struct wlr_xwayland_surface *requested_focus = lookup_surface(xwm, ev->event);
	if (!xwm->focus_surface || !requested_focus ||
			requested_focus->pid != xwm->focus_surface->pid) {
		xwm_send_focus_window(xwm, xwm->focus_surface);
	} else {
		xwm->focus_surface = requested_focus;
		xwm_send_focus_window(xwm, xwm->focus_surface);
	}
```

That is, adding a call to `xwm_send_focus_window` in the case we _do_ allow the focus change. Then I just cleaned up the code by inverting the if condition and calling `xwm_send_focus_window` unconditionally.

progandy:
Maybe ```xwm_surface_activate``` should be called for a successful request instead, so that also ```xwm_set_net_active_window``` is run?

emersion:
>Hmm, I don't think that's what this does unless I misunderstood this completely...

Eh, indeed, I completely missed the inverted condition.

Xyene:
Re: `xwm_surface_activate`, I also noticed the `xwm_set_net_active_window` difference (and calling it instead works fine, of course), but I don't know enough about X11 to know dialogs interact with `NET_ACTIVE_WINDOW` (if at all specially)... so I just left the existing logic as-is.

Is the expected behavior under "real" X11 that `NET_ACTIVE_WINDOW` changes?

emersion:
Tried reading the i3 source code, but they re-send the whole state to the server in one go: https://github.com/i3/i3/blob/10646eb0021403c1f1448339603b24c50da0e604/src/handlers.c#L1054

emersion:
In any case, if the handler needs to send more messages to the X11 server, it should be done in a separate patch.

emersion:
Thanks!

PR #2222, opened 2020-05-25
render/egl: stop including eglmesaext.h

emersion:
This is a Mesa-specific header that was needed because some Wayland EGL
extensions were missing from the Khronos registry. Now that this has
been fixed [1] and Mesa [2] & glvnd [3] have sync'ed their headers, we
can drop this workaround.

[1]: https://github.com/KhronosGroup/EGL-Registry/pull/95
[2]: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/4953
[3]: https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/225

* * *

* [glvnd 1.3.2](https://gitlab.freedesktop.org/glvnd/libglvnd/-/tags/v1.3.2) contains the patch
* [Mesa 20.2.0](https://docs.mesa3d.org/relnotes/20.2.0.html) contains the patch

### DIFF
--- a/include/wlr/config.h.in
+++ b/include/wlr/config.h.in
@@ -1,8 +1,6 @@
 #ifndef WLR_CONFIG_H
 #define WLR_CONFIG_H
 
-#mesondefine WLR_HAS_EGLMESAEXT_H
-
 #mesondefine WLR_HAS_LIBCAP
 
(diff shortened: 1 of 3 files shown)


emersion:
This is now ready for review.

PR #2224, opened 2020-05-26
xwayland_socket_connected: avoid calling wl_event_source_remove on NULL

### LINKED ISSUE #2163: Crash in xwayland_socket_connected
I observed the following crash from sway:

```
(gdb) bt full
#0  0x00007f62defae059 in wl_event_source_remove () at /lib/x86_64-linux-gnu/libwayland-server.so.0
#1  0x00007f62def64c96 in xwayland_socket_connected (fd=<optimized out>, mask=<optimized out>, data=0x560c22144bd0) at ../xwayland/xwayland.c:269
        wlr_xwayland = 0x560c22144bd0
#2  0x00007f62defae36a in wl_event_loop_dispatch () at /lib/x86_64-linux-gnu/libwayland-server.so.0
#3  0x00007f62defac9ba in wl_display_run () at /lib/x86_64-linux-gnu/libwayland-server.so.0
#4  0x0000560c20208842 in  ()
#5  0x00007f62decdb09b in __libc_start_main (main=
    0x560c20208320, argc=2, argv=0x7ffe930029e8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7ffe930029d8) at ../csu/libc-start.c:308
        result = <optimized out>
        unwind_buf = 
              {cancel_jmp_buf = {{jmp_buf = {0, -4827057226722392500, 94610078599568, 140731364682208, 0, 0, -1231805636900370868, -1162169670741211572}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x7ffe93002a00, 0x7f62df614190}, data = {prev = 0x0, cleanup = 0x0, canceltype = -1828705792}}}
        not_first_call = <optimized out>
#6  0x0000560c202089ba in  ()
```

This happened while a floating popup window was opened from an X11 application. The X11 applicatio closed immediately (no core dump), and then sway crashed with the stacktrace above on the next keystroke.

I will update if I see this crash again.

quantum5:
This should fix #2163.

I was unable to obtain more details through debug prints, because the bug has not triggered in the past two weeks. However, I believe this fix should be sufficient to prevent future crashes.

### DIFF
--- a/xwayland/server.c
+++ b/xwayland/server.c
@@ -351,8 +351,12 @@ static bool server_start(struct wlr_xwayland_server *server) {
 static int xwayland_socket_connected(int fd, uint32_t mask, void *data) {
 	struct wlr_xwayland_server *server = data;
 
-	wl_event_source_remove(server->x_fd_read_event[0]);
-	wl_event_source_remove(server->x_fd_read_event[1]);
+	if (server->x_fd_read_event[0]) {
+		wl_event_source_remove(server->x_fd_read_event[0]);
+	}
+	if (server->x_fd_read_event[1]) {
+		wl_event_source_remove(server->x_fd_read_event[1]);
+	}
 	server->x_fd_read_event[0] = server->x_fd_read_event[1] = NULL;
 
 	server_start(server);


emersion:
I'd prefer to understand what goes wrong before merging a fix.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2224

Issue #2227, opened 2020-05-27
get_(pointer|keyboard|touch) does not correctly handle missing capabilities


kennylevinsen:
get_(pointer|keyboard|touch) must succeed if a seat has *ever* had the relevant capability, but wlroots does nothing but log an error if the seat does not *currently* have the relevant capability.

If a new capability is lost in between a client being notified of its creation and the client creating the respective wl_(pointer|keyboard|touch) object, then a resource is never allocated server-side for the object, and the upcoming attempt to release the object will lead to a protocol error.

PR #2228, opened 2020-05-27
seat: Create inert objects for missing capabilities

kennylevinsen:
We should throw a protocol error if the relevant capability has never
existed when get_(pointer|keyboard|touch) is called. Otherwise, it
should succeed, even if the capability is not currently present.

This follows the spec, and avoids possible races with the client when
capabilities are lost.

Closes: https://github.com/swaywm/wlroots/issues/2227

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -214,6 +214,7 @@ struct wlr_seat {
 
 	char *name;
 	uint32_t capabilities;
+	uint32_t accumulated_capabilities;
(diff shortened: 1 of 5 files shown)


kennylevinsen:
cc @emersion

kennylevinsen:
There we go, tested to not break anything locally. Would be nice to get the one alacritty user capable of consistent reproduction to verify.

emersion:
Ref https://gitlab.freedesktop.org/wayland/wayland/-/issues/165 for the missing error codes

PR #2229, opened 2020-05-27
wlr_keyboard_group: introduce enter and leave

RedSoxFan:
Related swaywm/sway#4975
Sway PR: swaywm/sway#5393

This introduces the enter and leave events for wlr_keyboard_group.

The enter event is emitted when a keyboard is added to the group while a
key is pressed that is not pressed by any other keyboard in the group.
The data is a wl_array of the pressed key codes unique to the keyboard
that should now be considered pressed.

Similarly the leave event is emitted when a keyboard is removed from the
group while at least one key is pressed that is not pressed by any other
keyboard in the group. The data is a wl_array of the pressed key codes
unique to the keyboard that should now be considered released.

The purpose of these events are to allow the compositor to update its
state to avoid corruption. Additionally, for the leave event, the
focused surface may have been notified of a key press for some or all of
the key codes and needs to be notified of a key release to avoid state
corruption.

These were previously emitted as normal key events, but they are not
normal key events. There is no actual key press or release associated
with the events. It's purely for state keeping purposes. Emitting them
as separate events allows the compositor to handle them differently.
Since these are purely for state keeping purposes and are not associated
with an actual key being pressed or released, bindings should not be
triggered as a result of these events.

### DIFF
--- a/include/wlr/types/wlr_keyboard_group.h
+++ b/include/wlr/types/wlr_keyboard_group.h
@@ -18,6 +18,30 @@ struct wlr_keyboard_group {
 	struct wlr_input_device *input_device;
 	struct wl_list devices; // keyboard_group_device::link
 	struct wl_list keys; // keyboard_group_key::link
+
+	struct {
+		/*
+		 * Sent when a keyboard has entered the group with keys currently
+		 * pressed that are not pressed by any other keyboard in the group. The
+		 * data for this signal will be a wl_array containing the key codes.
+		 * This should be used to update the compositor's internal state.
+		 * Bindings should not be triggered based off of these key codes and
+		 * they should also not notify any surfaces of the key press.
+		 */
+		struct wl_signal enter;
+
+		/*
+		 * Sent when a keyboard has left the group with keys currently pressed
+		 * that are not pressed by any other keyboard in the group. The data for
+		 * this signal will be a wl_array containing the key codes. This should
+		 * be used to update the compositor's internal state. Bindings should
+		 * not be triggered based off of these key codes. Additionally, surfaces
+		 * should only be notified if they received a corresponding key press
+		 * for the key code.
+		 */
+		struct wl_signal leave;
+	} events;
+
 	void *data;
 };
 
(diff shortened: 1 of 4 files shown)


include/wlr/types/wlr_keyboard_group.h
```diff
@@ -18,6 +18,30 @@ struct wlr_keyboard_group {
...
 	struct wl_list keys; // keyboard_group_key::link
+
+	struct {
+		/*
+		 * Sent when a keyboard has entered the group with keys currently
+		 * pressed that are not pressed by any other keyboard in the group. The
+		 * data for this signal will be a wl_array containing the key codes.
+		 * This should be used to update the compositor's internal state.
+		 * Bindings should not be triggered based off of these key codes and
+		 * they should also caused any surfaces to be notified.
```
Xyene, line 29: Small typo in this last sentence, I think.

RedSoxFan: Oops. Fixed

Xyene:
Thanks!

emersion:
What's the difference between enter and leave? Would a compositor need to handle each differently?

The name is a little misleading because this has nothing to do with the Wayland events with the same name.

emersion:
>What's the difference between enter and leave? Would a compositor need to handle each differently?

Ah, so for one of these the compositor needs to silently press them, for the other to silently release them.

emersion:
What happens if the compositor uses a tree of keyboard groups? Does this still work?

Maybe we could have an event that just updates the set of currently pressed keys.

emersion:
Note that these events could be used by the Wayland backend to properly forward wl_keyboard.enter/leave events.

RedSoxFan:
> The name is a little misleading because this has nothing to do with the Wayland events with the same name.

I'm open to renaming them if you have any suggestions. 

> What happens if the compositor uses a tree of keyboard groups? Does this still work?

I don't think I ever considered this use case. I don't think this works too well right now though. I would think it would have the same pitfall as swaywm/sway#4955. As for these events in particular, I could add wlr_keyboard_group_notify_* functions for them.

Issue #2208, opened 2020-05-19
Add support for ZWP_TABLET_TOOL_V2_TYPE_FINGER


emersion:
It's missing at least here:

https://github.com/swaywm/wlroots/blob/666498db015ad1439a8916bce609baf8fa042b5d/backend/wayland/tablet_v2.c#L470

emersion:
Actually it's missing from libinput: https://gitlab.freedesktop.org/libinput/libinput/-/issues/490

PR #2230, opened 2020-05-28
Add comments for missing tablet tool entries

emersion:
And stop using default cases, so that we know which parts of the code
need an update when adding a new enum entry.

Closes: https://github.com/swaywm/wlroots/issues/2208


ddevault:
Thanks!

PR #2234, opened 2020-05-29
wlr_cursor: expose a wlr_cursor_frame function

RedSoxFan:
Related to swaywm/sway#5361
Required by swaywm/sway#5400

This exposes a wlr_cursor_frame function for compositors to use when
simulating cursor events. In can be used in combination with
wlr_cursor_move and other related functions.

wlr_signal_emit_safe is not exposed to compositors so this wrapper
is needed.


RedSoxFan:
Oops. I missed wlr_seat_pointer_notify_frame. Closing

PR #2237, opened 2020-05-31
export-dmabuf: Keep frame->output = NULL until frame is ready

any1:
This fixes a crash that happens when a client requests a frame on a
backend that does not implement export_dmabuf.

An assertion fails with the message:
sway: types/wlr_output.c:777: wlr_output_lock_attach_render: Assertion
`output->attach_render_locks > 0' failed.


any1:
Oops. Github confuses me quite often with its UI. I didn't notice the "View changes" button above your comment until after I force pushed changes and now I can't see what your suggestion was. I hope it's what you requested.

Issue #2096, opened 2020-04-02
backend/drm: rework atomic support


emersion:
This one is probably going to be controversial.

Right now we have a single DRM backend using both the legacy API and the atomic API. Both are abstracted behind `wlr_drm_interface`, which mostly mimicks the legacy API. However, using the atomic API in the backend doesn't provide any benefit and costs us an interface and two different implementations.

To really take advantage of the atomic API, we need to either add a new atomic backend or completely rework `wlr_drm_interface`. I'm not sure it's a good idea to go down the `wlr_drm_interface` route, because a whole lot of logic will diverge between legacy and atomic, and a lot of features will be atomic-only.

Also note that the kernel automatically adds legacy support to atomic drivers. So it's not like we'd support less hardware by using only the legacy API.

ascent12:
I'll agree that `wlr_drm_interface` is crap, and is just atomic bolted on top of legacy. Most of the functions in it can be chucked out, and basically become `test_state` (no-op on legacy) and `apply_state`, instead of throwing atomic out altogether.
I actually started some work doing this, but lost motivation. I really should pick that back up.

emersion:
>Most of the functions in it can be chucked out, and basically become test_state (no-op on legacy) and apply_state, instead of throwing atomic out altogether.

That would be the "rework `wlr_drm_interface`" route. I'd be interested to see how that plays out, especially when we start adding more atomic-only features.

> I really should pick that back up.

Please do :)

emersion:
This issue is now mostly addressed. The main remaining piece is VRR.

PR #2238, opened 2020-06-01
backend/drm: atomic improvements

emersion:
This PR:

- Makes adaptive sync atomic
- Performs a blocking atomic commit if no modeset or buffer is pending

The second point fixes output state updates not applied when no modeset or page-flip is performed. For instance, `wlr_output_set_gamma` followed by `wlr_output_commit` was a no-op (but shouldn't be).

No Sway PR required, since `wlr_output_enable_adaptive_sync` is already followed by `wlr_output_commit`.

Closes: https://github.com/swaywm/wlroots/issues/2096

* * *

Breaking change: `wlr_output_enable_adaptive_sync` must now be followed by `wlr_output_commit` to apply the change.

### DIFF
--- a/backend/drm/atomic.c
+++ b/backend/drm/atomic.c
@@ -196,6 +196,14 @@ static bool atomic_crtc_commit(struct wlr_drm_backend *drm,
 		}
 	}
 
+	bool prev_vrr_enabled =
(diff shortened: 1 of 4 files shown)


ddevault:
Thank you!

PR #2245, opened 2020-06-02
backend/wayland: fix spurious eglSwapBuffers failures

emersion:
This was introduced in [1]. However after reverting that PR I still
can't reproduce the bug the PR aimed to fix [2].

Since there's no good explanation why we would need to swap buffers
before resizing, let's just revert the PR.

[1]: https://github.com/swaywm/wlroots/pull/1486
[2]: https://github.com/swaywm/wlroots/issues/1371

Closes: https://github.com/swaywm/wlroots/issues/1768


ddevault:
Thanks!

PR #2246, opened 2020-06-02
render: stop making EGL context current in wlr_egl_init

emersion:
This leaves an EGL context current behind. wlr_gles2_renderer_create was
assuming the EGL context was already current because of this (because it
called a GL function right off the bat).


ddevault:
Thanks!

PR #2252, opened 2020-06-04
Patches to address static analysis warnings

MisterDA:
Hi!
This is a collection of small patches fixing some warnings reported by PVS-Studio.
Happy to change things if I got the "bug fix" wrong.

I’m not so sure about my fix for 7d1d619 "types/wlr_output: fix pointer to stack variable escaping scope".

### DIFF
--- a/xwayland/xwayland.c
+++ b/xwayland/xwayland.c
@@ -89,7 +89,7 @@ struct wlr_xwayland *wlr_xwayland_create(struct wl_display *wl_display,
 	};
 	xwayland->server = wlr_xwayland_server_create(wl_display, &options);
 	if (xwayland->server == NULL) {
-		free(xwayland->server);
+		free(xwayland);
 		return NULL;
 	}
 
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -800,7 +800,7 @@ bool drm_connector_set_mode(struct wlr_drm_connector *conn,
 		return false;
 	}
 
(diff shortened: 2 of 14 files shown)


xwayland/xwayland.c
```diff
@@ -91,7 +91,6 @@ struct wlr_xwayland *wlr_xwayland_create(struct wl_display *wl_display,
 	};
 	xwayland->server = wlr_xwayland_server_create(wl_display, &options);
 	if (xwayland->server == NULL) {
-		free(xwayland->server);
```
emersion, line 94: Typo: this should free `xwayland` instead

render/drm_format_set.c
```diff
@@ -76,15 +76,16 @@ bool wlr_drm_format_set_add(struct wlr_drm_format_set *set, uint32_t format,
 
 		if (fmt->len == fmt->cap) {
 			size_t cap = fmt->cap ? fmt->cap * 2 : 4;
-
-			fmt = realloc(fmt, sizeof(*fmt) + sizeof(fmt->modifiers[0]) * cap);
```
emersion, line 80: I don't understand why this change is necessary.

MisterDA: `fmt` leaks memory if the reallocation fails, because it will be overwritten with `NULL`. This patch ensures that we keep the old data in place if the reallocation fails.

emersion: `fmt` is a local variable, so it doesn't matter right?

MisterDA: You’re right.

PR #2256, opened 2020-06-05
contributing: resource destruction clarifications

emersion:


subdiff:
Since we talked about it on IRC my feedback: the clarification makes sense to me. I would add the wp_presentation_feedback object from presentation-time protocol as another example for an object being destroyed server-side besides wl_callback since it's in comparison less generic but it's also ok without it.

ddevault:
Thanks!

Issue #2253, opened 2020-06-04
Infinite recursion when trying to disable display on wlroots master


YSelfTool:
Excerpt from the debug log:
```
00:00:24.933 [sway/config/output.c:343] Turning off output DP-5
00:00:24.933 [sway/config/output.c:418] Committing output DP-5
00:00:25.280 [DEBUG] [backend/drm/drm.c:1073] Reallocating CRTCs
00:00:25.280 [DEBUG] [backend/drm/drm.c:1084] State before reallocation:
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'eDP-1' crtc=0 state=3 desired_enabled=1
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-1' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'HDMI-A-1' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-2' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'HDMI-A-2' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-3' crtc=-1 state=1 desired_enabled=1
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-4' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-5' crtc=1 state=3 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-6' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-7' crtc=2 state=3 desired_enabled=1
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-8' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1090]   'DP-9' crtc=-1 state=0 desired_enabled=0
00:00:25.280 [DEBUG] [backend/drm/drm.c:1142] State after reallocation:
[...]

PR #2259, opened 2020-06-05
backend/drm: fix stack overflow in dealloc_crtc

emersion:
Call drm_crtc_commit directly instead of calling drm_connector_set_mode.
This restores the previous behaviour where conn_enable was called [1].

[1]: https://github.com/swaywm/wlroots/blob/0c7c562482575cacaecadcd7913ef25aeb21711f/backend/drm/drm.c#L1093

Closes: https://github.com/swaywm/wlroots/issues/2253


ddevault:
Thanks!

PR #2261, opened 2020-06-05
Fix `get_current_time_msec` returning microseconds, and refactor to time utility

Xyene:
Refs https://github.com/swaywm/sway/pull/5425.

### DIFF
--- a/util/time.c
+++ b/util/time.c
@@ -4,6 +4,8 @@
 
 #include "util/time.h"
 
+const long NSEC_PER_SEC = 1000000000;
+
 int64_t timespec_to_msec(const struct timespec *a) {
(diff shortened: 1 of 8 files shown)


util/time.c
```diff
@@ -0,0 +1,27 @@
+#define _POSIX_C_SOURCE 200809L
+#include <stdint.h>
+#include <time.h>
+
+#include "util/time.h"
+
+const long NSEC_PER_SEC = 1000000000;
+
+uint32_t timespec_to_msec(const struct timespec *a) {
```
emersion, line 9: I'd prefer to have this return an `int64_t`, in case it's called with the result of `timespec_sub`.

Xyene: Ah, that explains why it was signed in the first place -- done.

PR #2264, opened 2020-06-07
Add support for xwayland_minimize actions

### LINKED ISSUE #2251: xwayland request_minimize signal
some games have the option "Minimizes on focus loss",
this seems broken in non-tiling wlroots compositors,
as it seems to not be implemented in wlroots.

damianatorrpm:
as required by chrome/steam csd or games:
Fixes #2251
In the PR there is also a fix for a forgotten break statement
Minimize functionality tested with wayfire


xwayland/xwm.c
```diff
@@ -1215,7 +1233,21 @@ static void xwm_handle_client_message(struct wlr_xwm *xwm,
...
 	} else if (ev->type == xwm->atoms[NET_ACTIVE_WINDOW]) {
 		xwm_handle_net_active_window_message(xwm, ev);
-	} else if (!xwm_handle_selection_client_message(xwm, ev)) {
+	} else if (ev->type == xwm->atoms[WM_CHANGE_STATE]) {
+		struct wlr_xwayland_surface *xsurface = lookup_surface(xwm, ev->window);
+		uint32_t detail = ev->data.data32[0];
+
+		if (xsurface == NULL) {
+			return;
+		} else if (detail == ICCCM_ICONIC_STATE) {
```
emersion, line 1242: What is the relationship between `ICCCM_ICONIC_STATE` and `NET_WM_STATE_HIDDEN`?

xwayland/xwm.c
```diff
@@ -1792,6 +1825,25 @@ struct wlr_xwm *xwm_create(struct wlr_xwayland *xwayland, int wm_fd) {
 	return xwm;
 }
 
+void wlr_xwayland_surface_set_minimized(struct wlr_xwayland_surface *surface,
+		bool minimized) {
+
```
emersion, line 1830: Style: stray newline

xwayland/xwm.c
```diff
@@ -1215,7 +1233,21 @@ static void xwm_handle_client_message(struct wlr_xwm *xwm,
...
+			return;
+		} else if (detail == ICCCM_ICONIC_STATE) {
+		        wlr_xwayland_surface_set_minimized(xsurface, true);
+		} else if (detail == ICCCM_NORMAL_STATE) {
+		        wlr_xwayland_surface_set_minimized(xsurface, false);
+		} else {
+			wlr_log(WLR_DEBUG, "unhandled wm_change_state event");
+		}
+	} 
+	else if (!xwm_handle_selection_client_message(xwm, ev)) {
```
emersion, line 1250: Style: `else if` on the same line as `}`

emersion:
Thanks for your PR!

Please make sure tabs are used for indentation. Some lines use spaces in this patch.

xwayland/xwm.c
damianatorrpm: if a window is minimized it's both the NET_WM_STATE is hidden the WM_STATE is iconic

emersion: Hm, so why do we need to handle `ICCCM_ICONIC_STATE` changes?

damianatorrpm:
I think all is resolved

xwayland/xwm.c
```diff
@@ -664,7 +670,9 @@ static void read_surface_net_wm_state(struct wlr_xwm *xwm,
 			xsurface->maximized_vert = true;
 		} else if (atom[i] == xwm->atoms[NET_WM_STATE_MAXIMIZED_HORZ]) {
 			xsurface->maximized_horz = true;
-		}
+		} else if (atom[i] == xwm->atoms[NET_WM_STATE_HIDDEN]) {
+			xsurface->minimized = true;
+		}
```
emersion, line 675: Style: trailing whitespace

xwayland/xwm.c
```diff
@@ -1142,6 +1151,9 @@ static void xwm_handle_net_wm_state_message(struct wlr_xwm *xwm,
 		} else if (property == xwm->atoms[NET_WM_STATE_MAXIMIZED_HORZ] &&
 				update_state(action, &xsurface->maximized_horz)) {
 			xsurface_set_net_wm_state(xsurface);
+		} else if (property == xwm->atoms[NET_WM_STATE_HIDDEN] &&
+				update_state(action, &xsurface->minimized)) {
```
emersion, line 1155: Style: trailing whitespace

xwayland/xwm.c
```diff
@@ -1792,6 +1824,24 @@ struct wlr_xwm *xwm_create(struct wlr_xwayland *xwayland, int wm_fd) {
 	return xwm;
 }
 
+void wlr_xwayland_surface_set_minimized(struct wlr_xwayland_surface *surface,
+		bool minimized) {
+	const bool change = surface->minimized != minimized;
```
emersion, line 1829: Style: trailing whitespace

xwayland/xwm.c
emersion: Bump

damianatorrpm: This is the way minimize is being set - through WM_STATE which is iconic state - i3 does the same, net_wm_state_hidden mean window is not visible, iconic state means minimized

damianatorrpm:
Strange how do I remove the whitespace than, I edit using gedit and it's set to tabs...

TheAvidDev:
It _is_ a tab, it's just extra after the end of the line. For example, you have code:
```
...minimized != minimized;	
```
After the `;` is a `	` (Tab - Github doesn't seem to like rendering this) so probably just move your cursor after it and hit backspace.

emersion:
Yup. To be more explicit:

```c
const bool change = surface->minimized != minimized;	← this is a tab
```

damianatorrpm:
@TheAvidDev @emersion ah, now I see it, of course trailing like in a trailer...

xwayland/xwm.c
damianatorrpm: Additionally you can do the following to test what clients do,
run you compositor with debug, so WLR_DEBUG is printed.
Click minimize button on chrome or steam csd, you will see WM_CHANGE_STATE
event. The WM_STATE has 3 types (minimized - iconic - 3) (normal - normal - 1)
and withdrawn which is basically closed.

damianatorrpm: Also you can compare the status of steam for example in minimized and unminimized 
state, both things WM_STATE and _NET_WM_STATE are changed.
The change signal / request comes via WM_STATE.

damianatorrpm:
https://github.com/i3/i3/commit/136b3e345bfde637d3270eafaa2d8aa3555cc3d5 
see also i3 handling

xwayland/xwm.c
emersion: Please refrain from marking open discussions as resolved.

Okay, so I guess WM_CHANGE_STATE with ICCCM_ICONIC_STATE is a way for clients to request to be minimized. In this case, this should go through the `request_minimize` event.

xwayland/xwm.c
```diff
@@ -1215,6 +1233,19 @@ static void xwm_handle_client_message(struct wlr_xwm *xwm,
...
+		uint32_t detail = ev->data.data32[0];
+
+		if (xsurface == NULL) {
+			return;
+		} else if (detail == ICCCM_ICONIC_STATE) {
+		        wlr_xwayland_surface_set_minimized(xsurface, true);
+		} else if (detail == ICCCM_NORMAL_STATE) {
+		        wlr_xwayland_surface_set_minimized(xsurface, false);
+		} else {
+			wlr_log(WLR_DEBUG, "unhandled wm_change_state event");
```
emersion, line 1247: Please also log the `detail` value

xwayland/xwm.c
```diff
@@ -1164,6 +1176,12 @@ static void xwm_handle_net_wm_state_message(struct wlr_xwm *xwm,
 
 		wlr_signal_emit_safe(&xsurface->events.request_maximize, xsurface);
 	}
+
+	if (minimized != xsurface->minimized) {
+		xsurface->saved_width = xsurface->width;
```
emersion, line 1181: We should only do this if the surface is minimized.

xwayland/xwm.c
```diff
@@ -1215,6 +1233,19 @@ static void xwm_handle_client_message(struct wlr_xwm *xwm,
 		xwm_handle_wm_protocols_message(xwm, ev);
 	} else if (ev->type == xwm->atoms[NET_ACTIVE_WINDOW]) {
 		xwm_handle_net_active_window_message(xwm, ev);
+	} else if (ev->type == xwm->atoms[WM_CHANGE_STATE]) {
+		struct wlr_xwayland_surface *xsurface = lookup_surface(xwm, ev->window);
```
emersion, line 1237: Can we extract all of this into a `xwm_handle_wm_change_state_message` function?

xwayland/xwm.c
```diff
@@ -1792,6 +1824,24 @@ struct wlr_xwm *xwm_create(struct wlr_xwayland *xwayland, int wm_fd) {
...
+	if (minimized) {
+		xsurface_set_wm_state(surface, ICCCM_ICONIC_STATE);
+	} else {
+		xsurface_set_wm_state(surface, ICCCM_NORMAL_STATE);
+	}
+
+	xsurface_set_net_wm_state(surface);
+	xcb_flush(surface->xwm->xcb_conn);
+	if (change) {
+	    wlr_signal_emit_safe(&surface->events.request_minimize, surface);
```
emersion, line 1841: `wlr_xwayland_surface_set_minimized` should not emit a `request_minimize` event. `request_minimize` should only be used to signal that a client requests to be minimized.

xwayland/xwm.c
damianatorrpm: Not sure what your above statement means code-wise, 
or does it just mean 'resolved'.

xwayland/xwm.c
damianatorrpm: Please be more specific in your comments.
Do what emit signals? Safe state?
The handling is exactly the same as for maximized in this function. 

I suppose the below is what you think is right so I changed it to that:
`	if (minimized && minimized != xsurface->minimized) {
`

xwayland/xwm.c
damianatorrpm: So where shoult the request_minimize event go?

damianatorrpm:
Bump.

emersion:
Please bump every week. We don't have enough time to reply to 2-day bumps for patches that aren't bugfixes.

damianatorrpm:
It's a bug fix see description.

emersion:
No, it's a new feature.

damianatorrpm:
Cool stuff. Once you replied to the in-conversations let me know.

xwayland/xwm.c
```diff
@@ -306,6 +309,9 @@ static void xsurface_set_net_wm_state(struct wlr_xwayland_surface *xsurface) {
 	if (xsurface->maximized_horz) {
 		property[i++] = xwm->atoms[NET_WM_STATE_MAXIMIZED_HORZ];
 	}
+	if (xsurface->minimized) {
+		property[i++] = xwm->atoms[NET_WM_STATE_HIDDEN];
```
MisterDA, line 313: The `property` array must be expanded to size `5` if the properties are not mutually exclusive, otherwise there will be buffer overflow.

damianatorrpm: changed it

xwayland/xwm.c
emersion: No, we want to emit the signal even if the client requests to *not* be minimized.

Please have a look at the similar block of code above for maximized. We only set `saved_width` and `saved_height` if the surface becomes maximized. We should do the same here, only update the saved size if the surface wants to become minimized.

xwayland/xwm.c
emersion: There are two code-paths in this PR:

- Client wants to become minimized. It sends `NET_WM_STATE_HIDDEN`, and if I understand correctly, it can alternatively send `ICCCM_ICONIC_STATE`. In this case, wlroots should (1) not minimize the window and (2) emit a `request_minimize` event. (1) is important because we need to let the compositor handle the request. Maybe the compositor can't minimize windows. Maybe the compositor wants to do something else when a client requests to be minimized. The compositor must stay in control. wlroots shouldn't do anything behind the compositor's back.
- A compositor decides to minimize a window. This can be in response to a `request_minimize` event, or this can be because of something else (e.g. the user clicked "minimize" in a compositor menubar). The compositor will call `wlr_xwayland_surface_set_minimized`.

xwayland/xwm.c
emersion: See https://github.com/swaywm/wlroots/pull/2264/files#r443474523.

xwayland/xwm.c
ddevault: @damianatorrpm this discussion needs to be resolved - this is the main blocker for this patch.

wlroots isn't a window manager in its own right, its a toolkit for window managers. We just help information flow smoothly from client to compositor implementation, and the compositor implementation calls the shots. So, instead of having wlroots' Xwayland layer handle the minimize request, we need to add a new event which the compositor can register for, which is emitted when the window requests to be minimized. Then, the compositor can decide what to do - for example, calling `wlr_xwayland_surface_set_minimized` themselves.

xwayland/xwm.c
ddevault: This also needs to be addressed

ddevault:
@damianatorrpm you need to be more polite to the people kind enough to review your patch. There are multiple open problems in the discussion that you need to address; bumping the thread before then is not helpful.

Also, general review comment: when you have everything squared away, please [rebase](https://git-rebase.io) these commits into a single commit.

xwayland/xwm.c
damianatorrpm: This has been changed, see the updated PR.

damianatorrpm: This has been resolved, see updated PR

xwayland/xwm.c
damianatorrpm: This has been resolved, see updated PR

xwayland/xwm.c
emersion: Can you look into this?

xwayland/xwm.c
emersion: Can you look into this?

damianatorrpm:
> @damianatorrpm you need to be more polite to the people kind enough to review your patch. There are multiple open problems in the discussion that you need to address; bumping the thread before then is not helpful.
> 
> Also, general review comment: when you have everything squared away, please [rebase](https://git-rebase.io) these commits into a single commit.

@ddevault 
These bumps included the request to answer a follow up question which had not been answered,
I was asked via IRC by to bump every week.

The commits are in a single commit.

When contributors send pull request, reviewers should have the tactfulness to read related and linked documentation, to have an understanding technically what it is about, instead of arguing on documented properties, which shows a 'less than interested' attitude.

>@damianatorrpm this discussion needs to be resolved - this is the main blocker for this patch.
All things mentioned have been addressed to the best of my knowledge and the discussions 'resolved'.

xwayland/xwm.c
damianatorrpm: I don't see how this is wrong, maybe you suggest a proper way?

ddevault:
You have a bad attitude, and it's not worth further damaging our reviewer's mental health to deal with you. It is not us that lack tact in this exchange.

ddevault:
If someone else wants to adopt this work and bring it to us with a better attitude, we'd be interested in reviewing it again later.

PR #2266, opened 2020-06-08
render/gles2: use glGetAttribLocation instead of hardcoded indices

### LINKED ISSUE #2265: render/gles2: use glGetAttribLocation instead of harded indices
Right now we have magic indices for shader attribs. Instead, use `glGetAttribLocation` to retrieve the indices, so that we don't blow up when modifying the shader.

Ideally we would do one `glGetAttribLocation` call on init after compiling shaders and stash the result in the renderer struct, so that we don't need to query it each time we need to render a quad.

Bl4ckb0ne:
Fixes #2265

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -38,6 +38,8 @@ struct wlr_gles2_tex_shader {
 	GLint invert_y;
 	GLint tex;
 	GLint alpha;
+	GLint pos_attrib;
+	GLint tex_attrib;
 };
 
 struct wlr_gles2_renderer {
@@ -58,11 +60,14 @@ struct wlr_gles2_renderer {
 			GLuint program;
 			GLint proj;
 			GLint color;
+			GLint pos_attrib;
 		} quad;
 		struct {
 			GLuint program;
 			GLint proj;
 			GLint color;
+			GLint pos_attrib;
+			GLint tex_attrib;
 		} ellipse;
 		struct wlr_gles2_tex_shader tex_rgba;
 		struct wlr_gles2_tex_shader tex_rgbx;
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -14,6 +14,13 @@
 #include <wlr/util/log.h>
 #include "render/gles2.h"
 
+static const GLfloat verts[] = {
+	1, 0, // top right
+	0, 0, // top left
+	1, 1, // bottom right
+	0, 1, // bottom left
+};
+
 struct wlr_gles2_procs gles2_procs = {0};
 
 static const struct wlr_renderer_impl renderer_impl;
@@ -86,37 +93,6 @@ static void gles2_scissor(struct wlr_renderer *wlr_renderer,
 	POP_GLES2_DEBUG;
 }
 
-static void draw_quad_with_texcoord(GLfloat x1, GLfloat y1,
-		GLfloat x2, GLfloat y2) {
-	GLfloat verts[] = {
-		1, 0, // top right
-		0, 0, // top left
-		1, 1, // bottom right
-		0, 1, // bottom left
-	};
-	GLfloat texcoord[] = {
-		x2, y1, // top right
-		x1, y1, // top left
-		x2, y2, // bottom right
-		x1, y2, // bottom left
-	};
-
-	glVertexAttribPointer(0, 2, GL_FLOAT, GL_FALSE, 0, verts);
-	glVertexAttribPointer(1, 2, GL_FLOAT, GL_FALSE, 0, texcoord);
-
-	glEnableVertexAttribArray(0);
-	glEnableVertexAttribArray(1);
-
-	glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
-
-	glDisableVertexAttribArray(0);
-	glDisableVertexAttribArray(1);
-}
-
-static void draw_quad(void) {
-	draw_quad_with_texcoord(0, 0, 1, 1);
-}
-
 static bool gles2_render_subtexture_with_matrix(
 		struct wlr_renderer *wlr_renderer, struct wlr_texture *wlr_texture,
 		const struct wlr_fbox *box, const float matrix[static 9],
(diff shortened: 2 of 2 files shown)


render/gles2/renderer.c
```diff
@@ -205,7 +210,20 @@ static void gles2_render_ellipse_with_matrix(struct wlr_renderer *wlr_renderer,
...
-	draw_quad();
+
+
+	glVertexAttribPointer(renderer->shaders.ellipse.pos_loc, 2, GL_FLOAT,
+			GL_FALSE, 0, verts);
+	glVertexAttribPointer(renderer->shaders.ellipse.tex_loc, 2, GL_FLOAT,
+			GL_FALSE, 0, texcoord);
+
+	glEnableVertexAttribArray(0);
+	glEnableVertexAttribArray(1);
```
ascent12, line 221: Shouldn't these use the *_loc members too?

include/render/gles2.h
```diff
@@ -56,11 +58,15 @@ struct wlr_gles2_renderer {
...
 			GLint color;
+			GLint pos_loc;
+			GLint tex_loc;
 		} quad;
 		struct {
 			GLuint program;
 			GLint proj;
 			GLint color;
+			GLint pos_loc;
+			GLint tex_loc;
```
emersion, line 69: I don't think the ellipse shader uses texcoords

render/gles2/shaders.c
```diff
@@ -6,14 +6,14 @@ const GLchar quad_vertex_src[] =
 "uniform mat3 proj;\n"
 "uniform vec4 color;\n"
 "attribute vec2 pos;\n"
-"attribute vec2 texcoord;\n"
+//"attribute vec2 texcoord;\n"
```
emersion, line 9: We can probably get rid of this (and the rest of the texcoord stuff) for colored quads

render/gles2/renderer.c
Bl4ckb0ne: It could yes, do you want me to change it?

include/render/gles2.h
Bl4ckb0ne: The ellipse shader uses [`ellipse_fragment_src`](https://github.com/swaywm/wlroots/blob/master/render/gles2/renderer.c#L720), which uses the [`texcoord`](https://github.com/swaywm/wlroots/blob/master/render/gles2/shaders.c#L35) attribute

render/gles2/shaders.c
Bl4ckb0ne: Oops, this was from a debug session. But yes, the `pos` vertices and the `texcoord` vertices are the same, I think we could reuse the same array of vertices.

render/gles2/renderer.c
ascent12: Yes.

include/render/gles2.h
emersion: Ah, indeed.

render/gles2/shaders.c
Bl4ckb0ne: This has been put back.

render/gles2/renderer.c
Bl4ckb0ne: Done for all calls to `glEnableVertexAttribArray`

include/render/gles2.h
```diff
@@ -56,11 +58,15 @@ struct wlr_gles2_renderer {
...
```
emersion, line 62: This isn't used

render/gles2/renderer.c
```diff
@@ -205,7 +210,20 @@ static void gles2_render_ellipse_with_matrix(struct wlr_renderer *wlr_renderer,
 
 	glUniformMatrix3fv(renderer->shaders.ellipse.proj, 1, GL_FALSE, transposition);
 	glUniform4f(renderer->shaders.ellipse.color, color[0], color[1], color[2], color[3]);
-	draw_quad();
+
+
```
emersion, line 214: Style: stray newline here

render/gles2/renderer.c
```diff
@@ -86,32 +99,6 @@ static void gles2_scissor(struct wlr_renderer *wlr_renderer,
 	POP_GLES2_DEBUG;
 }
 
-static void draw_quad(void) {
```
emersion, line 89: I wonder if it's worth it to keep this function and make it take `pos_loc` and `tex_loc`.

emersion: Eh, probably not

render/gles2/renderer.c
```diff
@@ -656,6 +678,8 @@ struct wlr_renderer *wlr_gles2_renderer_create(struct wlr_egl *egl) {
 	renderer->shaders.tex_rgbx.invert_y = glGetUniformLocation(prog, "invert_y");
 	renderer->shaders.tex_rgbx.tex = glGetUniformLocation(prog, "tex");
 	renderer->shaders.tex_rgbx.alpha = glGetUniformLocation(prog, "alpha");
+	renderer->shaders.quad.pos_loc = glGetAttribLocation(prog, "pos");
```
emersion, line 681: s/quad/tex_rgbx/

include/render/gles2.h
```diff
@@ -37,6 +37,8 @@ struct wlr_gles2_tex_shader {
 	GLint invert_y;
 	GLint tex;
 	GLint alpha;
+	GLint pos_loc;
```
emersion, line 40: Nit: the existing members are locations too, so it's inconsistent to add a `_loc` suffix to the new members, we should drop it

Bl4ckb0ne: It's an attribute location compared to other that are uniform locations. `pos_attrib` maybe?

include/render/gles2.h
Bl4ckb0ne: Done

render/gles2/renderer.c
Bl4ckb0ne: Done

include/render/gles2.h
emersion: Works for me

Bl4ckb0ne: Done

render/gles2/renderer.c
Bl4ckb0ne: Done

emersion:
Needs rebase.

Bl4ckb0ne:
I rebased and squashed, but I'd like to have another review please, I had to make a few changes caused by the merge of https://github.com/swaywm/wlroots/pull/2092

PR #2267, opened 2020-06-08
render: Fix blit_dmabuf() breakage due to API change

any1:


emersion:
Hmm. So one issue with this is that `external_only` is unconditionally set to false when the modifiers extension isn't supported.

any1:
> Hmm. So one issue with this is that `external_only` is unconditionally set to false when the modifiers extension isn't supported.

Well, it's an improvement upon the situation that we had before. :)

emersion:
It means `blit_dmabuf` will always fail on AMD because of the check.

emersion:
Err. I mean unconditionally set to true.

https://github.com/swaywm/wlroots/blob/e0d4f7517282c4c31d0486f80974fec20fd118e8/render/egl.c#L529

any1:
You're making more sense now. Maybe set it to `false` instead if the attribute isn't supported?

any1:
I can do either of two things: Reinstate the TODO, or change the default to `false` when the mod isn't supported.

emersion:
> Maybe set it to false instead if the attribute isn't supported?

I don't think that's a good idea, I believe external will fail less often for sampling because that's what we were doing before (when we weren't checking for the hint).

Let's keep the TODO for now, and we can discuss about resolving it after this is merged.

any1:
Ok, done.

render/gles2/renderer.c
```diff
@@ -364,7 +364,7 @@ static bool gles2_blit_dmabuf(struct wlr_renderer *wlr_renderer,
 
 	// TODO: The imported buffer should be checked with
 	// eglQueryDmaBufModifiersEXT to see if it may be modified.
-	EGLImageKHR image = wlr_egl_create_image_from_dmabuf(egl, dst_attr);
+	EGLImageKHR image = wlr_egl_create_image_from_dmabuf(egl, dst_attr, NULL);
```
emersion, line 367: NULL will make the call crash

any1: Well, that was hasty of me.

PR #2240, opened 2020-06-01
Renderer v6, incremental version

emersion:
This is an alternative to https://github.com/swaywm/wlroots/pull/1355. Instead of re-working the whole API, this PR just uses the new infrastructure inside the DRM backend.

The DRM backend used `EGLSurface` and `gbm_surface` until now. Stop using them and switch to `wlr_allocator` and `wlr_swapchain`. These new interfaces are mostly copied over from [glider](https://github.com/emersion/glider).

This PR doesn't break the wlroots API. For now the new APIs are not exposed in public header, once we add APIs that allow compositors to use it directly, we can easily move the headers to the public dir.

TODO:

- [x] Fix flip & 180 degree rotation
- [x] Damage tracking support (buffer age)
- [x] Multi-GPU
- [x] Debug `EGL_BAD_PARAMETER` failures when calling `eglDestroyImageKHR` on shutdown
- [x] Make sure screencopy + export-dmabuf work fine
- [x] Testing
  - [x] Investigate glClear issue in iris: https://github.com/swaywm/wlroots/pull/2240#issuecomment-673517061 ([upstream bug](https://gitlab.freedesktop.org/mesa/mesa/-/issues/3425), [fix](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/7384), workaround: `INTEL_DEBUG=nofc` or `MESA_LOADER_DRIVER_OVERRIDE=i965`)
  - [x] Investigate crashes when switching resolution: https://github.com/swaywm/wlroots/pull/2240#issuecomment-674381722
  - [x] Fix inverted captured rect with screencopy

References: https://github.com/swaywm/wlroots/issues/1352

* * *

Breaking changes: some implicit assumptions about `wlr_renderer`'s EGL context no longer hold true. Compositors not using EGL directly shouldn't be affected.

- wlroots' framebuffer no longer is the EGL default framebuffer
- wlroots' framebuffer no longer has a depth attachment

PR #2268, opened 2020-06-08
[WIP] Headless export-dmabuf

any1:
This is very much work in progress, but it does work quite well.

Things I'd like some feedback on:
 * GBM buffer allocation. Would it be better to get the render node from DRM if the renderer belongs to DRM?
 * The way I added `glEGLImageTargetRenderbufferStorageOES` into the backend is rather ugly. Suggestions?

### DIFF
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -179,6 +179,14 @@ bool wlr_renderer_blit_dmabuf(struct wlr_renderer *r,
 	return r->impl->blit_dmabuf(r, dst, src);
 }
 
+GLuint wlr_renderer_renderbuffer_from_image(struct wlr_renderer *r,
+		EGLImageKHR image) {
+	if (!r->impl->renderbuffer_from_image) {
+		return false;
+	}
+	return r->impl->renderbuffer_from_image(image);
+}
+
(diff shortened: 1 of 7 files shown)


render/wlr_renderer.c
```diff
@@ -164,6 +164,14 @@ bool wlr_renderer_blit_dmabuf(struct wlr_renderer *r,
 	return r->impl->blit_dmabuf(r, dst, src);
 }
 
+GLuint wlr_renderer_renderbuffer_from_image(struct wlr_renderer *r,
```
emersion, line 167: This leaks EGL/GL implementation details into the generic renderer interface.

render/egl.c
```diff
@@ -419,6 +419,9 @@ void wlr_egl_save_context(struct wlr_egl_context *context) {
 }
 
 bool wlr_egl_restore_context(struct wlr_egl_context *context) {
+	if (!context->display)
```
emersion, line 422: This doesn't correctly restore `EGL_NO_DISPLAY`.

emersion:
I've adapted #2240 for the headless backend: https://github.com/swaywm/wlroots/pull/2374

PR #2270, opened 2020-06-09
Improve DMA-BUF lifetime docs

emersion:
`wlr_buffer_get_dmabuf`'s behaviour avoids exporting the buffer as a DMA-BUF over and over each time we want to import it (e.g. to KMS or EGL).

cc @any1


ddevault:
Thanks!

Issue #2273, opened 2020-06-11
Build fails from AUR "wlroots-git"


joachimneu:
When trying to install "wlroots-git" from AUR, the build fails with:
```
Build targets in project: 116

wlroots 0.10.0

        systemd: 1
        elogind: 0
       xwayland: 1
    x11_backend: 1
      xcb-icccm: 1
     xcb-errors: 1

Found ninja-1.10.0 at /usr/bin/ninja
ninja: Entering directory `build'
[296/326] Compiling C object 'examples/c590b3c@@screencopy-dmabuf@exe/screencopy-dmabuf.c.o'
FAILED: examples/c590b3c@@screencopy-dmabuf@exe/screencopy-dmabuf.c.o 
cc -Iexamples/c590b3c@@screencopy-dmabuf@exe -Iexamples -I../examples -Iprotocol -I/usr/lib/libffi-3.2.1/include -I/usr/include/libpng16 -I/usr/include/libdrm -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -g -DWLR_USE_UNSTABLE -Wundef -Wlogical-op -Wmissing-include-dirs -Wold-style-definition -Wpointer-arith -Winit-self -Wstrict-prototypes -Wimplicit-fallthrough=2 -Wendif-labels -Wstrict-aliasing=2 -Woverflow -Wmissing-prototypes -Wno-missing-braces -Wno-missing-field-initializers -Wno-unused-parameter -fmacro-prefix-map=../= -DLIBINPUT_MAJOR=1 -DLIBINPUT_MINOR=15 -DLIBINPUT_PATCH=4 '-DICONDIR="/usr/share/icons"' -march=x86-64 -mtune=generic -O2 -fno-plt -g -fvar-tracking-assignments -fdebug-prefix-map=/home/joachim/.cache/yay/wlroots-git/src=/usr/src/debug -D_FORTIFY_SOURCE=2 -MD -MQ 'examples/c590b3c@@screencopy-dmabuf@exe/screencopy-dmabuf.c.o' -MF 'examples/c590b3c@@screencopy-dmabuf@exe/screencopy-dmabuf.c.o.d' -o 'examples/c590b3c@@screencopy-dmabuf@exe/screencopy-dmabuf.c.o' -c ../examples/screencopy-dmabuf.c
In file included from /usr/include/string.h:495,
[...]

PR #2274, opened 2020-06-11
examples/screencopy-dmabuf: call strncpy with maxlen - 1

emersion:
The original code wasn't wrong since we were manually writing a null
byte anyway, but this makes GCC happy.

Closes: https://github.com/swaywm/wlroots/issues/2273


ddevault:
Thanks!

PR #2275, opened 2020-06-11
examples/fullscreen-shell: stop advertising linux-dmabuf unconditonally

emersion:
Remove the wlr_linux_dmabuf_v1_create call. wlr_renderer_init_wl_display
will take care of creating the linux-dmabuf global if the OpenGL
implementation supports it.


ddevault:
Thanks!

PR #2277, opened 2020-06-12
render/gbm_allocator: add support for gbm_bo_get_fd_for_plane

emersion:
This ensures we get the right FD for multi-planar buffers.

Depends on [1].

[1]: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/5442


emersion:
This upstream Mesa MR has been merged, so this is now ready for review.

PR #2279, opened 2020-06-13
wlr_seat: restore NULL behavior for notify_enter

djpohly:
Unifies the behavior of `wlr_seat_*_notify_enter` with that of `wlr_seat_*_enter` when NULL is passed for the surface parameter.  Provides a non-breaking alternative for the fix in #2217 by using the grab interface.


emersion:
We decided against this approach. See the discussion in https://github.com/swaywm/wlroots/pull/2162#issuecomment-625083795

djpohly:
Thanks - missed that part of the discussion.

emersion:
Yeah, it wasn't properly linked from the PR - sorry about that.

Issue #2281, opened 2020-06-15
wlr_renderer_rect() misrenders negative width and height


goleo108:
I was trying to find a bug in Wio's render_view_border() which causes it to render borders in wrong positions when you reflect view, it's more noticeable when you reflect view by both x and y coordinates:
![bug](https://user-images.githubusercontent.com/56087471/84708324-d3dd6300-af4f-11ea-8e5d-b3028189c056.png).

emersion:
We should probably `assert(width > 0 && height > 0)`

goleo108:
But then it's pointless to have them as signed integers? If you are ok breaking backwards compatibility, it's better to either make them unsigned integers or replace them with x2 and y2 almost like in Go: https://golang.org/pkg/image/#Rectangle

emersion:
1. Not worth it to break the API for this
2. Unsigned integers make it too easy to mess up substractions

goleo108:
Then I recommend to change x, y, width and height values because nobody likes crashes, something like (please note I didn't test this code):
```
void wlr_render_rect(struct wlr_renderer *r, const struct wlr_box *box,
		const float color[static 4], const float projection[static 9]) {
	float matrix[9];
	wlr_box tbox = *box;
	if (tbox.width < 0) {
		tbox.width *= -1;
		tbox.x -= width;
	}
	if (tbox.height < 0) {
		tbox.height *= -1;
		tbox.y -= height;
	}
	wlr_matrix_project_box(matrix, &tbox, WL_OUTPUT_TRANSFORM_NORMAL, 0,
		projection);

	wlr_render_quad_with_matrix(r, color, matrix);
}
```

goleo108:
I'll send a PR after testing.

ascent12:
Negative width/height fundamentally doesn't make sense. I don't think adding workarounds for broken user code is a good idea.

goleo108:
Ok, I already patched Wio to avoid negative width and height, I just don't like duplicating code.

PR #2282, opened 2020-06-16
render: assert sane values for box functions

Emantor:
Width and height should always be > 0 for render functions which take a
wlr_box.

References https://github.com/swaywm/wlroots/issues/2281


goleo108:
@emersion I think it's a mistake to crash when width/height is 0, render code should simply ignore it, I mean not render at all. Also if user must make sure width and height are never negative, maybe it's better to state so in documentation and remove assertion from library for little performance?

goleo108:
Anyway, if you want to leave it as is, I'll simply patch Wio to not call `wlr_renderer_rect()` when width/height is 0.

emersion:
>render code should simply ignore [zero]

Maybe. I'm not feeling strongly about it.

>maybe it's better to state so in documentation and remove assertion from library for little performance?

-1

PR #2283, opened 2020-06-17
[WIP] Remove attach_render in read format check

### LINKED ISSUE #2164: Rendering goes bonkers with headless and wdisplays
Steps to reproduce:
* `swaymsg create_output`
* Run `wdisplays`
* Move the headless display and "Apply"

any1:
This fixes #2164

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -664,6 +664,9 @@ static bool drm_connector_pageflip_renderer(struct wlr_drm_connector *conn) {
 	struct wlr_drm_plane *plane = crtc->primary;
 	if (plane_get_next_fb(plane)->type == WLR_DRM_FB_TYPE_NONE) {
 		drm_surface_render_black_frame(&plane->surf);
+		wlr_renderer_preferred_read_format(
+				plane->surf.renderer->wlr_rend,
+				&conn->output.read_format);
 		if (!drm_fb_lock_surface(&plane->pending_fb, &plane->surf)) {
 			return false;
 		}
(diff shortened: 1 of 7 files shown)


emersion:
While this does fix the issue, this is a step backwards regarding renderer v6. With renderer v6 backends won't have access to the renderer anymore, they will just handle buffers.

Since the read format is per-EGLSurface this issue is pretty complicated. The renderer needs to make the surface/FBO current to know the preferred read format, but the renderer doesn't know when a surface/FBO is created so can't cache the read format.

any1:
Know a better solution? :)

emersion:
On the long run, we'll want to remove `wlr_output_preferred_read_format` because the output won't be tied to a specific renderer anymore.

For now, I want to avoid introducing the `read_format` field, because it adds a new renderer-specific field to `wlr_output`, and we're trying to move away from that.

If we don't want to break the API, the least intrusive fix is probably to add this at the end of `wlr_output_preferred_read_format`:

```c
if (output->impl->rollback) {
    output->impl->rollback(output);
}
```

It doesn't fix the fact that the output EGLSurface is made current behind the compositor's back. But it should at least fix the bug.

As discussed on IRC it's tricky to come up with a solution that doesn't make the EGLSurface/FBO current because the preferred read format depends on the EGLSurface/FBO.

any1:
I have tried adding the rollback at the end, and it does work given that buffer state check is removed in the headless backend's implementation of the rollback callback. I do wonder if there is a possibility that a rollback at this junction will cause a frame to be rendered "out of context" after the rollback.

emersion:
>buffer state check is removed in the headless backend's implementation of the rollback callback

Hm, do you mean that the headless backend only resets the current FBO if the pending state says `attach_render` has been called?

In which case, replacing `impl->attach_render` by `wlr_output_attach_render` would probably fix it, but really makes it clear that `wlr_output_preferred_read_format` should go away -- it's not nice to change the output state behind the compositor's back.

any1:
> Hm, do you mean that the headless backend only resets the current FBO if the pending state says attach_render has been called?

Yes

PR #2288, opened 2020-06-18
surface: don't unset width and height in finalize

Emantor:
During surface finalization we may not have received a new buffer,
resetting width and height in this case is wrong since we display the
old buffer in this case.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -174,16 +174,16 @@ static void surface_state_finalize(struct wlr_surface *surface,
 		}
 	}
 
-	if (state->buffer_resource != NULL) {
-		if (state->viewport.has_dst) {
+	if (state->viewport.has_dst) {
+		if (state->buffer_width == 0 && state->buffer_height == 0) {
+			state->width = state->height = 0;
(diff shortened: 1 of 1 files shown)


types/wlr_surface.c
```diff
@@ -174,16 +174,12 @@ static void surface_state_finalize(struct wlr_surface *surface,
...
 
-	if (state->buffer_resource != NULL) {
-		if (state->viewport.has_dst) {
-			state->width = state->viewport.dst_width;
-			state->height = state->viewport.dst_height;
-		} else {
-			surface_state_viewport_src_size(state,
-				&state->width, &state->height);
-		}
+	if (state->viewport.has_dst) {
```
emersion, line 177: If the src/dst rects are set but there's no buffer, we still need to set width/height to zero.

We can probably add a `if (state->buffer_width == 0 && state->buffer_height == 0)` check?

Emantor: Added the check.

mntmn:
Yep, tested with slurp and swaylock, fixes the issues.

PR #2289, opened 2020-06-18
surface: ignore viewport src rect on NULL buffer

emersion:
According to the viewporter protocol:

> If the wl_buffer is NULL, the surface has no content and therefore no size.


ddevault:
Thanks!

PR #2292, opened 2020-06-19
output: fix dangling renderer context after wlr_output_preferred_read_format

emersion:
## output: rename impl->rollback to rollback_render

The output backend API is now mostly state-less thanks to the atomic
hooks (commit and test). There is one exception though: attach_render.
This function makes the rendering context current. However sometimes the
compositor might decide not to render after attach_render (e.g. when
there's nothing new to render to the back buffer). Thus
wlr_output_rollback has been introduced to revert the pending state.

Because the output backend API is mostly state-less, the only thing
wlr_output_impl.rollback needs to do is revert the current rendering
context. Rename the function to rollback_render to make this clear. Add
a check in the common wlr_output code to only call rollback_render when
attach_buffer has been previously called.

On the long term, we'll be able to remove attach_render and
rollback_render together.

## output: make rollback_render mandatory

If the output backend provides attach_render, assert it also provides a
way to revert it via rollback_render.

## output: fix dangling renderer context after wlr_output_preferred_read_format

attach_render was called without un-setting the current rendering
context afterwards.

Closes: #2164

* * * 

Breaking change for backends: `wlr_output_impl.rollback` has been renamed to `rollback_render` and now needs to unconditionally unset the renderer context. It'll only be called if `attach_render` has previously been called.

cc @any1

### DIFF
--- a/include/wlr/interfaces/wlr_output.h
+++ b/include/wlr/interfaces/wlr_output.h
@@ -23,7 +23,7 @@ struct wlr_output_impl {
 	bool (*attach_render)(struct wlr_output *output, int *buffer_age);
 	bool (*test)(struct wlr_output *output);
 	bool (*commit)(struct wlr_output *output);
-	void (*rollback)(struct wlr_output *output);
+	void (*rollback_render)(struct wlr_output *output);
 	size_t (*get_gamma_size)(struct wlr_output *output);
 	bool (*export_dmabuf)(struct wlr_output *output,
 		struct wlr_dmabuf_attributes *attribs);
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -314,7 +314,7 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 
 void wlr_output_init(struct wlr_output *output, struct wlr_backend *backend,
 		const struct wlr_output_impl *impl, struct wl_display *display) {
-	assert(impl->attach_render && impl->commit);
+	assert(impl->attach_render && impl->rollback_render && impl->commit);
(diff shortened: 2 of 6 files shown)


include/wlr/interfaces/wlr_output.h
```diff
@@ -23,7 +23,7 @@ struct wlr_output_impl {
 	bool (*attach_render)(struct wlr_output *output, int *buffer_age);
 	bool (*test)(struct wlr_output *output);
 	bool (*commit)(struct wlr_output *output);
-	void (*rollback)(struct wlr_output *output);
+	void (*rollback_render)(struct wlr_output *output);
```
ddevault, line 26: A bit out of scope for this change, but do you think you could go through and add doc comments to each of these fields?

emersion: Good idea. Done in https://github.com/swaywm/wlroots/pull/2293

types/wlr_output.c
```diff
@@ -448,15 +448,16 @@ bool wlr_output_attach_render(struct wlr_output *output, int *buffer_age) {
 
 bool wlr_output_preferred_read_format(struct wlr_output *output,
```
any1, line 449: Is this guaranteed to be called while no other output is attached? Could this cause problems when another output is attached?

emersion: Yeah, that's a problem, but already was before. I'll open a follow-up issue about this.

ddevault:
Thanks!

types/wlr_output.c
emersion: https://github.com/swaywm/wlroots/issues/2296

PR #2294, opened 2020-06-19
backend/noop: add missing rollback_render output impl

emersion:
3c5dbfd97c77 ("output: make rollback_render mandatory") makes a no-op
output init fail without this function.


ddevault:
Thanks!

PR #2295, opened 2020-06-19
tablet: add sentinel entry for tablet tool enum

Xyene:
Required by https://github.com/swaywm/sway/pull/5472; could be worked around but this seems cleanest.

This might be a breaking change if a compositor had a switch based off this enum, and is compiling with `-Wall`.


emersion:
In general I'm not a fan of these because it forces us to add a dummy case for it in all switch statements...

emersion:
(I can definitely see how this helps with the Sway PR though)

Xyene:
Fair enough, I've rewritten the sway side to not require this (at the expense of some nasty bits, but that was unavoidable either way).

Issue #2174, opened 2020-05-08
Xwayland hangs after restart


emersion:
1. Run `xeyes`
2. Kill Xwayland
3. Run `xeyes` again

Xwayland is correctly restarted, but `xeyes` hangs.

Tested with the lazy mode.

emersion:
@ammen99 bisected it to 27609ba0

ammen99:
I just realized that in Wayfire, a hang can happen even without restarting Xwayland - just opening a google chrome window on two outputs is enough (but doesn't seem to happen in Sway). I will try to investigate when I have time.

PR #2298, opened 2020-06-20
Fix Xwayland restart

soreau:
This fixes two problems on Xwayland restart:

- The xwayland ready signal now fires on restart
- xwm is destroyed instead of leaked

The latter caused compositors to consume all cpu once Xwayland was restarted.

Closes: https://github.com/swaywm/wlroots/issues/2174

### DIFF
--- a/xwayland/server.c
+++ b/xwayland/server.c
@@ -232,8 +232,6 @@ static int xserver_handle_ready(int signal_number, void *data) {
 		.wm_fd = server->wm_fd[0],
 	};
 	wlr_signal_emit_safe(&server->events.ready, &event);
-	/* ready is a one-shot signal, fire and forget */
-	wl_signal_init(&server->events.ready);
 
 	return 1; /* wayland event loop dispatcher's count */
 
--- a/xwayland/xwayland.c
+++ b/xwayland/xwayland.c
@@ -55,8 +55,6 @@ static void handle_server_ready(struct wl_listener *listener, void *data) {
 	}
 
 	wlr_signal_emit_safe(&xwayland->events.ready, NULL);
-	/* ready is a one-shot signal, fire and forget */
-	wl_signal_init(&xwayland->events.ready);
 }
 
 void wlr_xwayland_destroy(struct wlr_xwayland *xwayland) {
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -1312,6 +1312,11 @@ static int x11_event_handler(int fd, uint32_t mask, void *data) {
 	xcb_generic_event_t *event;
 	struct wlr_xwm *xwm = data;
 
+	if ((mask & WL_EVENT_HANGUP) || (mask & WL_EVENT_ERROR)) {
+		xwm_destroy(xwm);
+		return 0;
+	}
+
 	while ((event = xcb_poll_for_event(xwm->xcb_conn))) {
 		count++;
 
@@ -1493,6 +1498,7 @@ void xwm_destroy(struct wlr_xwm *xwm) {
 	wl_list_remove(&xwm->compositor_destroy.link);
 	xcb_disconnect(xwm->xcb_conn);
 
+	xwm->xwayland->xwm = NULL;
 	free(xwm);
 }
 


xwayland/xwayland.c
```diff
@@ -37,6 +37,9 @@ static void handle_server_ready(struct wl_listener *listener, void *data) {
 		wl_container_of(listener, xwayland, server_ready);
 	struct wlr_xwayland_server_ready_event *event = data;
 
+	if (xwayland->xwm) {
```
emersion, line 40: No need for this NULL check, `xwm_destroy` already does it

emersion:
For me, this fixes the issue for the first restart, but not for subsequent restarts. This fails:

```sh
xeyes &
pkill Xwayland
xeyes & # Xwayland restarts as expected
pkill Xwayland
xeyes & # Fails
```

soreau:
> For me, this fixes the issue for the first restart, but not for subsequent restarts. This fails:
> 
> ```shell
> xeyes &
> pkill Xwayland
> xeyes & # Xwayland restarts as expected
> pkill Xwayland
> xeyes & # Fails
> ```

Yes, I just found the same test fails here as well. Any idea why?

soreau:
Interestingly enough:
```
xcalc
killall Xwayland
xcalc # works
killall Xwayland
xcalc # hangs..
^C
killall xwayland
xcalc # works
```

xwayland/server.c
```diff
@@ -232,8 +232,6 @@ static int xserver_handle_ready(int signal_number, void *data) {
 		.wm_fd = server->wm_fd[0],
 	};
 	wlr_signal_emit_safe(&server->events.ready, &event);
-	/* ready is a one-shot signal, fire and forget */
-	wl_signal_init(&server->events.ready);
```
emersion, line 236: There's another place where this is done:

https://github.com/swaywm/wlroots/blob/c93016028680a0ce2bac2f2b8eeff5e5c1ea467e/xwayland/xwayland.c#L58

We should probably remove this one as well.

soreau:
@emersion Trying to destroy xwm on ready signal was too late, which was causing the intermittent failures. Now I've moved it to the more appropriate destroy function, and it seems to work reliably.

xwayland/xwayland.c
```diff
@@ -29,7 +29,9 @@ struct wlr_xwayland_cursor {
 static void handle_server_destroy(struct wl_listener *listener, void *data) {
 	struct wlr_xwayland *xwayland =
 		wl_container_of(listener, xwayland, server_destroy);
+
 	wlr_xwayland_destroy(xwayland);
+	xwm_destroy(xwayland->xwm);
```
emersion, line 34: Hm. This isn't run on restart. right?

soreau: You're right, it's not even being called..

soreau: Ok, new patch. I wanted to avoid adding a new signal but this seems to work for wayfire at least.. for some reason, sway still uses 100% cpu when restarting Xwayland.

soreau: And that is because my signal handler is only called under wayfire, not sway.. go figure.

soreau: Should be fixed now, I guess sway uses lazy and I wasn't emitting the signal in that case.

soreau:
If the `restarting` signal is acceptable, we probably should expose this to compositors as well, since they can now have their own xwm.

xwayland/server.c
```diff
@@ -146,7 +146,10 @@ static void server_finish_process(struct wlr_xwayland_server *server) {
 	/* We do not kill the Xwayland process, it dies to broken pipe
 	 * after we close our side of the wm/wl fds. This is more reliable
 	 * than trying to kill something that might no longer be Xwayland.
+	 * We do need to destroy xwm however, so send a signal to give an
+	 * opportunity.
 	 */
+	wlr_signal_emit_safe(&server->events.finish, NULL);
```
emersion, line 152: Hmm. What I think happens is that we close the X11 FD when stopping the server, but I don't think we handle this properly in the XWM.

When the X11 server dies, the XWM should cleanup and destroy itself, just like any other X11 client. Right now it seems we aren't handling FD failures properly in XWM. I think we're missing something like this in `x11_event_handler`:

```c
if ((mask & WL_EVENT_HANGUP) || (mask & WL_EVENT_ERROR)) {
	// Cleanup
}
```

xwayland/xwm.c
```diff
@@ -1312,6 +1312,11 @@ static int x11_event_handler(int fd, uint32_t mask, void *data) {
 	xcb_generic_event_t *event;
 	struct wlr_xwm *xwm = data;
 
+	if ((mask & WL_EVENT_HANGUP) || (mask & WL_EVENT_ERROR)) {
+		xwm_destroy(xwm);
```
emersion, line 1316: `wlr_xwayland.xwm` will still be set to the free'ed memory after this. I think we need to add a destroy event to `wlr_xwm` to let `wlr_xwayland` do its cleanup, wdyt?

emersion: Actually no need for this complexity. We have a `wlr_xwayland` field in `wlr_xwm`, we can just reset the pointer to NULL there.

PR #2304, opened 2020-06-25
Implement logind session SetType method to change session type to wayland

### LINKED ISSUE #2144: logind: call SetType on Session
New systemd feature that allows us to update the session type.

See https://github.com/systemd/systemd/pull/14925

tokyovigilante:
Sets systemd-logind session type to `wayland` so that systemd recognises a graphical session is running. Needed for logind idle inhibition to work. Relies on the logind session `SetType` command which is not yet released (in systemd master, tagged for v246).

Relevant systemd PR here - https://github.com/systemd/systemd/pull/14925

Will close #2144.

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -257,12 +257,16 @@ static bool set_type(struct logind_session *session) {
 	sd_bus_error_free(&error);
 	sd_bus_message_unref(msg);
 
+	if (ret < 0) {
+		return false;
+	}
+
 	ret = setenv("XDG_SESSION_TYPE", "wayland", 1);
 	if (ret < 0) {
 		wlr_log(WLR_ERROR, "Failed to set XDG_SESSION_TYPE for session");
+		return false;
 	}
-
-	return ret >= 0;
+	return true;
 }
 
 static void release_control(struct logind_session *session) {
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -250,12 +250,18 @@ static bool set_type(struct logind_session *session) {
 		session->path, "org.freedesktop.login1.Session", "SetType",
 		&error, &msg, "s", "wayland");
 	if (ret < 0) {
-		wlr_log(WLR_ERROR, "Failed to set session type for session: %s",
+		wlr_log(WLR_ERROR, "Failed to set logind session type for session: %s",
 			error.message);
 	}
 
 	sd_bus_error_free(&error);
 	sd_bus_message_unref(msg);
+
+	ret = setenv("XDG_SESSION_TYPE", "wayland", 1);
+	if (ret < 0) {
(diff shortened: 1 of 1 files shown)


emersion:
We need to set `XDG_SESSION_TYPE` as well.

tokyovigilante:
Thanks, noted. I'm not sure about the error handling, probably either failing isn't the end of the world, maybe just log and move on.

emersion:
We should definitely not fail when systemd is < v246.

backend/session/logind.c
```diff
@@ -241,6 +241,30 @@ static bool take_control(struct logind_session *session) {
...
+		&error, &msg, "s", "wayland");
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to set logind session type for session: %s",
+			error.message);
+	}
+
+	sd_bus_error_free(&error);
+	sd_bus_message_unref(msg);
+
+	ret = setenv("XDG_SESSION_TYPE", "wayland", 1);
```
emersion, line 260: We shouldn't set `XDG_SESSION_TYPE` if `SetType` failed, or the two sources of truth will become out-of-sync.

tokyovigilante: Makes sense definitely for v246, which I guess isn't far away, but on v245 is it worth setting `XDG_SESSION_TYPE` independently? If I'm understanding this right, the only reason `SetType` would realistically fail is if it isn't present on v245, but is it worth setting `XDG_SESSION_TYPE` then? If not then obviously just returning after the `SetType` call is fine.

emersion: > If not then obviously just returning after the SetType call is fine.

Yup!

tokyovigilante: OK cool.

craigbarnes:
> LGTM as well, but let's wait for v246 to be released before merging.

systemd v246 has been released.

backend/session/logind.c
```diff
@@ -819,6 +847,11 @@ static struct wlr_session *logind_session_create(struct wl_display *disp) {
 		}
 	}
 
+	if (!set_type(session)) {
+		// Not fatal
+		wlr_log(WLR_INFO, "Failed to set logind session type to wayland");
```
craigbarnes, line 852: This message isn't strictly accurate. If it was the `setenv()` call that caused the `false` return, the logind session type *was* set to `wayland`.

tokyovigilante: True. We already log both failure modes correctly in `set_type`, maybe this log is just redundant?

craigbarnes: > maybe this log is just redundant?

I'd say so, yes. Since it's just a "best effort" attempt to set the session type and not a fatal error, it seems like making the function a `void` return and just calling `set_type(session)` would be sufficient.

emersion:
Works with latest systemd. Thanks!

PR #2305, opened 2020-06-26
xcursor: add xorg-x11 and cursors path to XCURSORPATH

kchibisov:
This should match default XCURSORPATH, which is used by libwayland-cursor
and other xcursor loading libraries more closely.


emersion:
libwayland code: https://gitlab.freedesktop.org/wayland/wayland/-/blob/5d201df72f3d4f4cb8b8f75f980169b03507da38/cursor/xcursor.c#L620

PR #2306, opened 2020-06-28
wlr_drag: Destroy drag after releasing grabs

kennylevinsen:
wlr_drag sets up keyboard, pointer and touch grabs, which block 'enter'
events (and thus focus changes). For the compositor to be able to update
focus (e.g. to focus the drop target) from the destroy handler, the
grabs must be released before the destroy event is signalled.


kennylevinsen:
Needed by: https://github.com/swaywm/sway/pull/5489

PR #2307, opened 2020-06-28
Fix screencopy-dmabuf y inversion

any1:


render/gles2/renderer.c
```diff
@@ -384,7 +384,9 @@ static bool gles2_blit_dmabuf(struct wlr_renderer *wlr_renderer,
 	bool dst_inverted_y =
 		!!(dst_attr->flags & WLR_DMABUF_ATTRIBUTES_FLAGS_Y_INVERT);
 	struct wlr_gles2_texture *gles2_src_tex = gles2_get_texture(src_tex);
-	gles2_src_tex->inverted_y = src_inverted_y ^ dst_inverted_y;
+	// The result is negated because using a dmabuf as a colour attachment
+	// as is done below, also causes a y-inversion.
```
emersion, line 388: Can we add a note: "This is caused by wlr_matrix_projection."?

any1: Oh, so it's not because of the dmabuf colour attachment?

emersion: Yeah, the color attachment renders things normally.

`wlr_matrix_projection` however bakes a y-invert transform in the matrix. This is necessary when rendering to an EGLSurface, where the coordinates are inverted.

Things would be better if `wlr_matrix_projection` didn't do that, and the y-invert implementation detail was handled inside of `wlr_renderer`. It's too late for this because wlr_matrix is stable API.

any1: Ahh, ok, this makes more sense than what I assumed was the cause.

any1: Done.

I agree that it would be nicer to handle the y-inversion outside of `wlr_matrix_projection`. Maybe just add a `wlr_matrix_projection_v2` that doesn't do this?

emersion: I'm not sure how that would help, because compositors would still use the old function (or otherwise end up with an upside-down rendering).

Maybe we could at least use it inside `gles2_blit_dmabuf`. In https://github.com/swaywm/wlroots/pull/2240 I've just flipped the matrix given by `wlr_matrix_projection`.

any1: Well, gcc has `__attribute__((deprecated))` which would help to draw attention to the issue, but maybe it's not worth the hassle.

PR #2314, opened 2020-07-02
xwayland: check executable exists on init

emersion:
If the executable doesn't exist, fail early. This avoids setting an
invalid DISPLAY env variable when Xwayland isn't installed.

### DIFF
--- a/xwayland/server.c
+++ b/xwayland/server.c
@@ -1,6 +1,7 @@
 #define _POSIX_C_SOURCE 200809L
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


ddevault:
I don't like this patch. I would rather attempt to execute it and only set DISPLAY if it succeeds.

emersion:
Problem is, we execute Xwayland lazily. So if Xwayland doesn't exist, DISPLAY will be set, then apps will try to connect to it and hang forever.

ddevault:
Maybe we can try running `Xwayland -version` to acertain its existence/workitude?

A strict reading of POSIX allows execvp to run programs which are found outside of the PATH variable. I'm als not certain that this interprets PATH 100% to spec.

emersion:
>Maybe we can try running Xwayland -version to ascertain its existence/workitude?

`-version` is implemented but not yet released. `Xwayland -help` should work (with standard output/error set to /dev/null).

>A strict reading of POSIX allows execvp to run programs which are found outside of the PATH variable. I'm also not certain that this interprets PATH 100% to spec.

True. And execvp has implementation-defined behavior when PATH is unset.

emersion:
Updated the check to use `exec`. This has the downside of adding ~80ms to the startup time, which kind of negates the benefit of lazy xwayland I guess.

ddevault:
Meh. Maybe we should ditch lazy Xwayland, the benefits seem to be negligible.

emersion:
Hm, after thinking about it, non-lazy Xwayland wouldn't work either. When Xwayland is ready it'll send SIGUSR1. We wouldn't be able to know whether Xwayland is available before the signal, so we'd need to block for a potentially long time.

ddevault:
So what is the right approach?

ammen99:
> Hm, after thinking about it, non-lazy Xwayland wouldn't work either. When Xwayland is ready it'll send SIGUSR1. We wouldn't be able to know whether Xwayland is available before the signal, so we'd need to block for a potentially long time.

What is the problem exactly? Here, you seem to be testing whether Xwayland runs at all, why do we need to wait for SIGUSR1?

emersion:
>Here, you seem to be testing whether Xwayland runs at all, why do we need to wait for SIGUSR1?

My goal was to find the best way to detect whether Xwayland is available. The potential solutions were:

- Manually parse `$PATH` and try to find it: re-implements logic in `exec`, `exec` may behave differently depending on the libc implementation
- Execute `Xwayland -help`: slow
- Execute `Xwayland` normally and block until SIGUSR1: slow as well

None of those are great.

emersion:
Closing in favor of https://github.com/swaywm/wlroots/pull/2717.

Issue #2098, opened 2020-04-02
Choose more consistent commit behavior


emersion:
Right now we have two commit events: one for `wlr_surface`, one for `wlr_output`. Each of these structs have a state struct with a `pending` and `current` field.

When `commit` is emitted, I'd expect these assumptions to be true:

1. Users should have access to the committed state in the `current` state
2. Users should be able to figure out what changed in this commit (how?)
3. The `pending` state should be empty (it's just been applied)

Questions:

- Should we allow access to the previous state? Note that we don't want to retain locks on buffers longer than necessary.
- `pending.committed` is a bitmask of fields changed since last commit. What should `current.committed` mean? Right now it seems like it's the union of all commits' `pending.committed`.

Right now (1) is true, however:

* `wlr_output` clears the pending state after `commit` is emitted, breaking (3). The pending state is used by uers to figure out what changed (2). `wlr_output` provides no access to the previous state.
* `wlr_surface` has a `previous` state, however `previous.buffer` is an exception and is always set to NULL. (3) is true. Also, the previous state seems to be only used by `wlr_surface` itself. Users have no way to figure out what changed (2).

It would be nice to decide on a consistent behavior.

emersion:
Oh, and each shell (xdg-shell, layer-shell) also has similar state structs with similar issues.

emersion:
>Should we allow access to the previous state? Note that we don't want to retain locks on buffers longer than necessary.

Probably not, because users can easily cache any value if they need to.

emersion:
>Users should be able to figure out what changed in this commit (how?)

https://github.com/swaywm/wlroots/pull/2315 introduces a `wlr_output_event_commit` struct which contains a `committed` field.

emersion:
> wlr_output clears the pending state after commit is emitted, breaking (3). The pending state is used by users to figure out what changed (2). wlr_output provides no access to the previous state.

Addressed in https://github.com/swaywm/wlroots/pull/2630

emersion:
`wlr_surface` and `wlr_output` have a commit event, however things are more complicated for add-on protocols like xdg-shell, layer-shell and xdg-decoration. It'd be nice to be able to figure out what the newly committed state fields are in the `wl_surface` commit handler, to be able to have atomic logic that depends on the whole new state.

One solution would be to re-pupose the `current.committed` bitfield as the set of fields that changed since the last commit. Not sure it's a good idea.

[...]

PR #2315, opened 2020-07-02
Introduce wlr_output_event_commit

emersion:
## output: introduce wlr_output_event_commit
    
This event contains a `committed` bitfield, which allows callers to know
which output fields changed during the commit.
    
This allows users to setup a single atomic commit listener, instead of
setting up one listener for each event (mode, scale, transform, and so
on).
    
References: https://github.com/swaywm/wlroots/issues/2098

## output-power-management-v1: listen to output commit

* * *

Breaking change: the `wlr_output.events.commit` event now has a data argument of type `struct wlr_output_event_commit *` instead of `struct wlr_output *`

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -163,7 +163,7 @@ struct wlr_output {
 		// Emitted right before commit
 		struct wl_signal precommit; // wlr_output_event_precommit
 		// Emitted right after commit
-		struct wl_signal commit;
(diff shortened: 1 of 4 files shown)


ddevault:
Nice! Do you have a sway patch?

emersion:
Sway doesn't need to be updated for the breaking change, but here's a patch taking advantage of this PR: https://github.com/swaywm/sway/pull/5649

ddevault:
Thanks!

PR #2316, opened 2020-07-03
xwm: end transfers when the requestor is destroyed

jchv:
This improves the failure cases when incremental transfers fail to complete successfully for one reason or another. In particular it should vastly improve the failure case seen at https://github.com/swaywm/sway/issues/4007, especially because QXcb uses a special temporary window for its requestor, that gets always destroyed in the case that the error happens.


emersion:
>QXcb uses a special temporary window for its requestor

As a side note, we should probably do that in wlroots too, instead of having a queue of transfers.

PR #2317, opened 2020-07-03
input/pointer: add functions for modifying pointer state

Xyene:
These allow a compositor to do things like skip motion events on pointer constraint unlock, or not move the pointer during touch emulation.

Refs swaywm/sway#5431, swaywm/sway#5459.

---

Marked as draft until the Sway-side PRs are updated / we decide if this is how we want to do this.


emersion:
I extracted the `wlr_seat_pointer_wrap` function in a145430afac6.

PR #2319, opened 2020-07-05
examples: fix improper use of `free`

### LINKED ISSUE #2303: idle.c example calls free instead of wl_registry_destroy
I used this example code as a model for some code I was developing, and ran into segfaults as a result:

https://github.com/swaywm/wlroots/blob/master/examples/idle.c#L130

Is there any particular reason that `free` is used in place of `wl_registry_destroy` here?

From some basic debugging I did, issues occur when `wl_registry.global_remove` shows up (which seemed to be happening when DPMS turned the display off for 10+ seconds). From what I understand, the client and server probably both think the `wl_registry` is bound, otherwise I don't think the server would be sending events to it.

j-n-f:
Closes #2303

### DIFF
--- a/examples/idle.c
+++ b/examples/idle.c
@@ -127,7 +127,7 @@ int main(int argc, char *argv[]) {
 	wl_registry_add_listener(registry, &registry_listener, NULL);
 	wl_display_dispatch(display);
 	wl_display_roundtrip(display);
-	free(registry);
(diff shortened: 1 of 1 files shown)


j-n-f:
Looks like that build failure had nothing to do with the code.

emersion:
Yeah, this is a known FreeBSD bug.

PR #2328, opened 2020-07-13
render: egl: Use current display to restore NULL context

### LINKED ISSUE #2327: wlr_egl_restore_context() fails when context->display == NULL
When no context is current `wlr_egl_save_context()` will set `context->display = NULL` and `eglMakeCurrent()` doesn't like that. See https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglMakeCurrent.xhtml:

> As with other commands taking EGLDisplay parameters, if display is not a valid EGLDisplay handle, an EGL_BAD_DISPLAY error is generated. (Some implementations have chosen to allow EGL_NO_DISPLAY as a valid display parameter for eglMakeCurrent. This behavior is not portable to all EGL implementations, and should be considered as an undocumented vendor extension).

This results in an error:
```
[EGL] command: eglMakeCurrent, error: EGL_BAD_DISPLAY (0x3008), message: "Invalid display (nil)"
```

any1:
eglGetCurrentDisplay() returns NULL when there is no context current and
eglMakeCurrent() needs a display argument.

Fixes #2327

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -459,7 +459,20 @@ void wlr_egl_save_context(struct wlr_egl_context *context) {
 }
 
 bool wlr_egl_restore_context(struct wlr_egl_context *context) {
-	return eglMakeCurrent(context->display, context->draw_surface,
+	// If the saved context is a null-context, we must use the current
+	// display instead of the saved display because eglMakeCurrent() can't
+	// handle EGL_NO_DISPLAY.
+	EGLDisplay display = context->display == EGL_NO_DISPLAY ?
+		eglGetCurrentDisplay() : context->display;
+
+	// If the current display is also EGL_NO_DISPLAY, we assume that there
+	// is currently no context set and no action needs to be taken to unset
+	// the context.
+	if (display == EGL_NO_DISPLAY) {
+		return true;
+	}
+
+	return eglMakeCurrent(display, context->draw_surface,
 			context->read_surface, context->context);
 }
 


emersion:
I think the current usage is okay:

> To release the current context without assigning a new one, set context to EGL_NO_CONTEXT and set draw and read to EGL_NO_SURFACE. […] This is the only case where an uninitialized display may be passed to eglMakeCurrent.

emersion:
Ah, maybe not. "not a valid EGLDisplay handle" is different from "uninitialized".

render/egl.c
```diff
@@ -459,7 +459,9 @@ void wlr_egl_save_context(struct wlr_egl_context *context) {
 }
 
 bool wlr_egl_restore_context(struct wlr_egl_context *context) {
-	return eglMakeCurrent(context->display, context->draw_surface,
+	EGLDisplay display = context->display ?
+		context->display : eglGetCurrentDisplay();
```
emersion, line 463: `eglGetCurrentDisplay` can return `EGL_NO_DISPLAY` as well

any1: Yes, but this is not an actual use case.

any1: Actually, perhaps it's safe to say that `context->display == eglGetCurrentDisplay() == EGL_NO_DISPLAY` means that there was and is no context, so no action needs to be taken.

emersion: Yeah, LGTM

any1: Updated

any1: Oh, oops. Need to fix one thing!

any1: Fixed now

render/egl.c
```diff
@@ -459,8 +459,18 @@ void wlr_egl_save_context(struct wlr_egl_context *context) {
...
+	// If the saved context is a null-context, we must use the current
+	// display instead of the saved display because eglMakeCurrent() can't
+	// handle EGL_NO_DISPLAY.
+	EGLDisplay display = context->display == EGL_NO_DISPLAY ?
+		eglGetCurrentDisplay() : context->display;
+
+	// If the current display is also EGL_NO_DISPLAY, we assume that there
+	// is currently no context set and no action needs to be taken to unset
+	// the context.
+	return display == EGL_NO_DISPLAY ? true :
```
emersion, line 471: Hmm, this is a little bit hard to read. Could we switch to a more common `if`?

any1: Done

render/egl.c
```diff
@@ -459,7 +459,19 @@ void wlr_egl_save_context(struct wlr_egl_context *context) {
...
+	// If the saved context is a null-context, we must use the current
+	// display instead of the saved display because eglMakeCurrent() can't
+	// handle EGL_NO_DISPLAY.
+	EGLDisplay display = context->display == EGL_NO_DISPLAY ?
+		eglGetCurrentDisplay() : context->display;
+
+	// If the current display is also EGL_NO_DISPLAY, we assume that there
+	// is currently no context set and no action needs to be taken to unset
+	// the context.
+	if (display == EGL_NO_DISPLAY)
```
emersion, line 471: Style: braces are mandatory

any1: oops, done

PR #2329, opened 2020-07-13
WiP: Cursor spy

any1:
This implements a protocol for spying on cursors on a given output. This is needed for implementing client side cursor rendering in VNC. See https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#cursor-pseudo-encoding

This feature significantly improves user experience on high latency VNC.

This implementation currently assumes that the pixel format read from cursor textures is ARGB8888. This will be fixed later, I just want some feedback on the protocol.

wayland-protocols MR is here: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/39

### DIFF
--- a/examples/cursor-spy.c
+++ b/examples/cursor-spy.c
@@ -1,0 +1,266 @@
+/*
+ * Copyright © 2008 Kristian Høgsberg
+ * Copyright © 2020 Andri Yngvason
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+#define _POSIX_C_SOURCE 200112L
+#include <errno.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <png.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/param.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#include <assert.h>
+#include <wayland-client-protocol.h>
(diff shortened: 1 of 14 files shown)


emersion:
I'd prefer the mechanism to get the cursor image to integrate with screencopy. The same kind of thing would be necessary for capturing toplevels. Maybe we'll need an abstract way of talking about buffer sources, like we have for buffers (`wl_buffer`).

any1:
This "buffer source" is something that has read-only access to surfaces, so maybe we could call it something like "zext_surface_view" or "zext_surface_reader"? Or maybe "zext_buffer_source" is just fine?

emersion:
I'd prefer not to use the "surface" word, since it would be nice to allow other things to be captured as well, e.g. workspaces or groups of surfaces.

any1:
Any thoughts on how to discover buffer sources? For an outputs and maybe even seats (for cursors) we can just request to get the buffer source for that specific object. But for toplevels, maybe we can use app_id?

progandy:
app_id seems too unspecific without some other selection mechanism. An application may have multiple toplevels or multiple instances. Maybe you could use a ```wlr-foreign-toplevel-management``` handle for privileged access or a ```xdg-foreign``` handle if clients should negotiate for access permissions through some other protocol?

ammen99:
> app_id seems too unspecific without some other selection mechanism. An application may have multiple toplevels or multiple instances. Maybe you could use a `wlr-foreign-toplevel-management` handle for privileged access or a `xdg-foreign` handle if clients should negotiate for access permissions through some other protocol?
> 
> A random thought: Does it make sense to allow the creation of xdg-foreign handles from wlr-foreign-toplevel-management that in turn could be passed around through some other request protocol? I.e. a client asks for any toplevel, a toplevel manager receives the request, asks the user to choose a toplevel and hands back an xdg-foreign reference?

I am not sure I understand your question. The toplevel manager is the compositor. So why can't it just create an `xdg-foreign` from the actual `wlr_surface`/`wlr_xdg_toplevel`? Why the roundtrip via `foreign-toplevel-management`?

progandy:
> I am not sure I understand your question. The toplevel manager is the compositor. So why can't it just create an `xdg-foreign` from the actual `wlr_surface`/`wlr_xdg_toplevel`? Why the roundtrip via `foreign-toplevel-management`?

I think I was trying to allow interactive toplevel selection in an separate privileged application, instead of having sway itself do that. (similar to the screencast portal SelectSources, but in pure wayland)

emersion:
>I am not sure I understand your question. The toplevel manager is the compositor. So why can't it just create an xdg-foreign from the actual wlr_surface/wlr_xdg_toplevel? Why the roundtrip via foreign-toplevel-management?

Because the client capturing a toplevel doesn't own the captured toplevel. Without foreign-toplevel-management, the client doing the capture would need to obtain the xdg-foreign token via a side-channel.

xdg-foreign tokens are maybe the wrong tool here. They allow setting the parent of the toplevel referenced by the token, we don't want capture tools to be able to do that.

>Any thoughts on how to discover buffer sources?

The plan so far was to split foreign-toplevel-management into two protocols: one protocol with read-only access to foreign toplevel info, another protocol which also allows to manage the foreign toplevels.

any1:
> > Any thoughts on how to discover buffer sources?
> 
> The plan so far was to split foreign-toplevel-management into two protocols: one protocol with read-only access to foreign toplevel info, another protocol which also allows to manage the foreign toplevels. Maybe we need to discuss about this again.

But what about cursor buffer sources? Should that be implemented in foreign-toplevel-management too?

emersion:
No, cursors aren't toplevels.

any1:
Yeah, sorry, I was confused because the topic is cursor capturing but diverged to discuss toplevels. I'm just wondering if there should be a protocol for discovering "cursor buffer sources" or if it should be a part of some other protocol.

any1:
How about something like this? https://gist.github.com/any1/0fd43c2a849366a38c5580f3f613c217

emersion:
Why do we need an abstraction? Can't we just add a `capture_cursor` request or something?

any1:
I was just entertaining a previous comment where you said:

> I'd prefer the mechanism to get the cursor image to integrate with screencopy. The same kind of thing would be necessary for capturing toplevels. Maybe we'll need an abstract way of talking about buffer sources, like we have for buffers (`wl_buffer`).

But, yeah, I could add a  request to screencopy. I do wonder how to implement `copy_with_damage` for a "cursor frame" though...

PR #2333, opened 2020-07-16
Improve DRM plane matching logic

### LINKED ISSUE #1943: failed assertion on rbp4 with multiple monitors
I'm hitting an assertion on a Raspberry Pi 4 with two monitors (archlinuxarm)

* Sway Version: `1.2` (also tried building `sway` and `wlroots` from `git`)

* Debug Log:

```
2019-12-02 21:50:57 - [sway/main.c:152] Linux rpi4 4.19.86-1-ARCH swaywm/sway#1 SMP PREEMPT Sat Nov 30 18:54:34 UTC 2019 armv7l GNU/Linux
2019-12-02 21:50:57 - [sway/main.c:168] Contents of /etc/os-release:
2019-12-02 21:50:57 - [sway/main.c:152] NAME="Arch Linux ARM"
2019-12-02 21:50:57 - [sway/main.c:152] PRETTY_NAME="Arch Linux ARM"
2019-12-02 21:50:57 - [sway/main.c:152] ID=archarm
2019-12-02 21:50:57 - [sway/main.c:152] ID_LIKE=arch
2019-12-02 21:50:57 - [sway/main.c:152] BUILD_ID=rolling
2019-12-02 21:50:57 - [sway/main.c:152] ANSI_COLOR="0;36"
2019-12-02 21:50:57 - [sway/main.c:152] HOME_URL="https://archlinuxarm.org/"
2019-12-02 21:50:57 - [sway/main.c:152] DOCUMENTATION_URL="https://archlinuxarm.org/wiki"
2019-12-02 21:50:57 - [sway/main.c:152] SUPPORT_URL="https://archlinuxarm.org/forum"
2019-12-02 21:50:57 - [sway/main.c:152] BUG_REPORT_URL="https://github.com/archlinuxarm/PKGBUILDs/issues"
2019-12-02 21:50:57 - [sway/main.c:152] LOGO=archlinux
2019-12-02 21:50:57 - [sway/main.c:140] LD_LIBRARY_PATH=(null)
2019-12-02 21:50:57 - [sway/main.c:140] LD_PRELOAD=(null)
2019-12-02 21:50:57 - [sway/main.c:140] PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
2019-12-02 21:50:57 - [sway/main.c:140] SWAYSOCK=(null)
2019-12-02 21:50:57 - [sway/server.c:39] Preparing Wayland server initialization
2019-12-02 21:50:57 - [backend/session/logind.c:675] Successfully loaded logind session
2019-12-02 21:50:57 - [backend/backend.c:192] Found 2 GPUs
2019-12-02 21:50:57 - [backend/drm/backend.c:152] Initializing DRM backend for /dev/dri/card1 (vc4)
2019-12-02 21:50:57 - [backend/drm/drm.c:69] Using atomic DRM interface
2019-12-02 21:50:57 - [backend/drm/drm.c:254] Found 2 DRM CRTCs
2019-12-02 21:50:57 - [backend/drm/drm.c:175] Found 6 DRM planes
sway: backend/drm/drm.c:85: add_plane: Assertion `!(type == DRM_PLANE_TYPE_PRIMARY && crtc->primary)' failed
```

* Configuration File:

Reproducible with no (default) config file.

schauveau:
This is an improved implementation of the plane matching logic when the kernel DRM does not enforce a single CRTC per plane.

This is not a bullet proof implementation but it should work fine for the Raspberry Pi 4 where all CRTCs are possible for all planes. 

Closes: https://github.com/swaywm/wlroots/issues/1943

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -197,16 +197,19 @@ static bool init_planes(struct wlr_drm_backend *drm) {
 			goto error;
 		}
 
+		// We don't really care about overlay planes, as we don't support them
+		// yet.
+		if (type == DRM_PLANE_TYPE_OVERLAY) {
+			drmModeFreePlane(plane);
+			continue;
+		}
+
 		/*
 		 * This is a very naive implementation of the plane matching
 		 * logic. Primary and cursor planes should only work on a
 		 * single CRTC, and this should be perfectly adequate, but
 		 * overlay planes can potentially work with multiple CRTCs,
 		 * meaning this could return inefficient/skewed results.
-		 *
-		 * However, we don't really care about overlay planes, as we
-		 * don't support them yet. We only bother to keep basic
-		 * tracking of them for DRM lease clients.
 		 *
 		 * possible_crtcs is a bitmask of crtcs, where each bit is an
 		 * index into drmModeRes.crtcs. So if bit 0 is set (ffs starts
@@ -218,18 +221,6 @@ static bool init_planes(struct wlr_drm_backend *drm) {
 		assert(crtc_bit >= 0 && (size_t)crtc_bit < drm->num_crtcs);
 
 		struct wlr_drm_crtc *crtc = &drm->crtcs[crtc_bit];
-
-		if (type == DRM_PLANE_TYPE_OVERLAY) {
-			uint32_t *tmp = realloc(crtc->overlays,
-				sizeof(*crtc->overlays) * (crtc->num_overlays + 1));
-			if (tmp) {
-				crtc->overlays = tmp;
-				crtc->overlays[crtc->num_overlays++] = id;
-			}
-			drmModeFreePlane(plane);
-			continue;
-		}
-
 		if (!add_plane(drm, crtc, plane, type, &props)) {
 			drmModeFreePlane(plane);
 			goto error;
@@ -314,7 +305,6 @@ void finish_drm_resources(struct wlr_drm_backend *drm) {
 			wlr_drm_format_set_finish(&crtc->cursor->formats);
 			free(crtc->cursor);
 		}
-		free(crtc->overlays);
 	}
 
 	free(drm->crtcs);
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -204,23 +204,26 @@ static bool init_planes(struct wlr_drm_backend *drm) {
 			continue;
 		}
 
-		/*
-		 * This is a very naive implementation of the plane matching
-		 * logic. Primary and cursor planes should only work on a
-		 * single CRTC, and this should be perfectly adequate, but
-		 * overlay planes can potentially work with multiple CRTCs,
-		 * meaning this could return inefficient/skewed results.
-		 *
-		 * possible_crtcs is a bitmask of crtcs, where each bit is an
-		 * index into drmModeRes.crtcs. So if bit 0 is set (ffs starts
-		 * counting from 1), crtc 0 is possible.
-		 */
-		int crtc_bit = ffs(plane->possible_crtcs) - 1;
+		assert(drm->num_crtcs <= 32);
+		struct wlr_drm_crtc *crtc = NULL;
+		for (size_t j = 0; j < drm->num_crtcs ; j++) {
+			uint32_t crtc_bit = 1 << j;
+			if ((plane->possible_crtcs & crtc_bit) == 0) {
+				continue;
+			}
 
-		// This would be a kernel bug
-		assert(crtc_bit >= 0 && (size_t)crtc_bit < drm->num_crtcs);
+			struct wlr_drm_crtc *candidate = &drm->crtcs[j];
+			if ((type == DRM_PLANE_TYPE_PRIMARY && !candidate->primary) ||
+					(type == DRM_PLANE_TYPE_CURSOR && !candidate->cursor)) {
+				crtc = candidate;
+				break;
+			}
+		}
+		if (!crtc) {
+			drmModeFreePlane(plane);
+			continue;
+		}
 
-		struct wlr_drm_crtc *crtc = &drm->crtcs[crtc_bit];
 		if (!add_plane(drm, crtc, plane, type, &props)) {
 			drmModeFreePlane(plane);
 			goto error;
--- a/include/backend/drm/drm.h
+++ b/include/backend/drm/drm.h
@@ -63,13 +63,6 @@ struct wlr_drm_crtc {
 	struct wlr_drm_plane *primary;
 	struct wlr_drm_plane *cursor;
 
-	/*
-	 * We don't support overlay planes yet, but we keep track of them to
-	 * give to DRM lease clients.
-	 */
-	size_t num_overlays;
-	uint32_t *overlays;
-
 	union wlr_drm_crtc_props props;
 };
 


backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		 * the last possible CRTC.
+                 */
+
+		int crtc_bit ;
+		int num_crtcs = (int)drm->num_crtcs ;
+		assert(num_crtcs<=64);
+		uint64_t bits = plane->possible_crtcs ;
+		// Some drm implementations (e.g. vc4 on raspberry pi) may set more
+		// bits than there are CRTCs so the unused upper bits must be cleared
+		// to insure a proper detection of the last possible CRTC.
```
ascent12, line 226: Can you post the output of [drm_info](https://github.com/ascent12/drm_info) here?
That sounds like a kernel bug if this is true, and I hate adding workarounds for bugs.

schauveau: Here it is.

Recent 5.x kernels for Raspberry Pi are always producing CRTCs: {0, 1, 2, 3, 4, 5, 6, 7} for all planes regardless of the actual number of CRTCs:   
https://drmdb.emersion.fr/devices/01cd892b4221

Older 4.9 kernels (i.e. the kernel officially supported by the Raspberry Pi OS) are not setting unused bits and so produce CRTCs: {0,1} with 2 monitors.  
https://drmdb.emersion.fr/devices/e973c8924b7e

However, I just looked at the kernel driver code and I discovered that this is caused by the hard-coded 0xff here
 https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/gpu/drm/vc4/vc4_firmware_kms.c#L820

This is a modified kernel provided by the Raspberry Pi foundation. As far as I can tell, the vc4 driver in the mainline Linux kernels is not behaving like that.  Even worse, the association of the planes to the CRTCs is also hardcoded meaning that a  `possible_crtcs` with more than one bit is wrong. 

For example, according to the current `possible_crtcs` it should be legal to assign the 1st cursor plane to the 2nd crtc and the 2nd cursor plane to the 1st crtc. I just tried that in sway and the effect was that my left screen was showing the right screen cursor and the other way around. So this is definitely a bug and I will fill a bug report in the rpi kernel. 

Though, it would be nice if my pull request was accepted to allow rpi4 users to use multiple screens without having to wait for a new kernel. A basic plane matching logic is better than nothing. 

Also clearing the unused upper bits is not really needed even for the rpi4 because the driver is allocating exactly one plane of each type for each crtc so the fallback is never needed. However, that could become a problem if the rpi4 driver was modified to allow more than one overlay per crtc.

schauveau: If your main concern is the cleaning of the upper bits, I can rewrite the loop as follow. The behavior should be identical with the last valid bit being used as a fallback.

    int num_crtcs = (int)drm->num_crtcs ;
    assert(num_crtcs<=64);
    int crtc_bit = -1;
    uint64_t bits = plane->possible_crtcs ;
    for (int j=0;j<num_crtcs;j++) {
       if (bits & 1) {
          crtc_bit = j;
          struct wlr_drm_crtc *crtc = &drm->crtcs[crtc_bit];
          if (type == DRM_PLANE_TYPE_PRIMARY && crtc->primary) {
             continue;
          }
          if (type == DRM_PLANE_TYPE_CURSOR && crtc->cursor) {
             continue;
          }
          if (type == DRM_PLANE_TYPE_OVERLAY && crtc->num_overlays>0) {
             continue;
          }
          break;
       }
       bits >>= 1;                  
    }

ascent12: >However, I just looked at the kernel driver code and I discovered that this is caused by the hard-coded 0xff here

That's just straight up broken.

>Though, it would be nice if my pull request was accepted to allow rpi4 users to use multiple screens without having to wait for a new kernel

I absolutely despise driver-specific code, and driver-specific workarounds is even worse. Once you add them in, they become extremely hard to remove, and just leads to a maintenance burden. The case is even worse when we're not talking about an upstream kernel.

>However, that could become a problem if the rpi4 driver was modified to allow more than one overlay per crtc.

I believe it's kernel UAPI[1] to have exactly 1 primary plane per CRTC, meaning the current `ffs()` algorithm works fine.
I'm not quite as sure if it's UAPI for cursor planes, but it doesn't actually make sense to have more than 1 cursor plane anyway; it's kind of a legacy idea for userspace's convenience, and any sane driver writer would make that an overlay plane instead.

Overlay planes currently are not important. We track them (because we kind of have to), but we don't actually use them. The comment in the code explains that the algorithm we use is intentionally crap, and we're willing to accept bad results. Once we're doing planes properly with libliftoff, then a real matching algorithm needs to be used (which can't reliably be done ahead of time), but libliftoff itself handles that.

---
[1] https://01.org/linuxgraphics/gfx-docs/drm/gpu/drm-kms.html#plane-abstraction : "All drivers should provide one primary plane per CRTC to avoid surprising userspace too much"

schauveau: I assume that this means that the pull request will not be accepted. I do not really mind. My patched wlroots is working fine so I will simply attach a patch to #1943. That should be enough to please the few raspberry pi users concerned with that problem.

emersion: @ascent12 I asked on #dri-devel and they said it's not a driver bug (https://github.com/swaywm/wlroots/issues/1943#issuecomment-561261556).

schauveau: @emersion It depends what you asked for. I agree than setting bit for non-existing CRTCs is not really a bug. Also, setting multiple CRTCs for each plane is legal and that could be an argument for my pull request.. 

However in that specific case, I do not believe anymore that the Raspberry Pi kernel implementation is correct because a quick look at the `vc4_fkms_create_screen` function (https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/gpu/drm/vc4/vc4_firmware_kms.c#L1738) shows that they are manually creating and associating 3 planes to each CRTC. As I explained previously, swapping the association order of the cursor planes results in the cursors appearing on the wrong screen. This is definitely not the expected behavior.

backend/drm/drm.c
```diff
@@ -196,8 +196,9 @@ static bool init_planes(struct wlr_drm_backend *drm) {
 			goto error;
 		}
 
-		/*
-		 * This is a very naive implementation of the plane matching
+
+                /*
```
emersion, line 200: Style: we indent with tabs

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
 		 * tracking of them for DRM lease clients.
 		 *
 		 * possible_crtcs is a bitmask of crtcs, where each bit is an
-		 * index into drmModeRes.crtcs. So if bit 0 is set (ffs starts
-		 * counting from 1), crtc 0 is possible.
-		 */
-		int crtc_bit = ffs(plane->possible_crtcs) - 1;
+		 * index into drmModeRes.crtcs. So if bit i is set crtc i is
+		 * possible.
+                 *
```
emersion, line 214: Style: we indent with tabs

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		 * The loop below is selecting the first possible CRTC without a
+		 * associated plane of the considered type or, as a fallback,
+		 * the last possible CRTC.
+                 */
```
emersion, line 218: Style: we indent with tabs

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+
+		int crtc_bit ;
```
emersion, line 220: Style: no space before `;`

(Applies to the rest of the file as well)

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		int num_crtcs = (int)drm->num_crtcs ;
+		assert(num_crtcs<=64);
```
emersion, line 222: Style: add spaces around arithmetic operators

(Applies to the rest of the file as well)

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		// bits than there are CRTCs so the unused upper bits must be cleared
+		// to insure a proper detection of the last possible CRTC.
+		bits &= (((uint64_t)1u)<<num_crtcs) - 1;
+		for (crtc_bit = 0 ; crtc_bit < num_crtcs ; crtc_bit++) {
+			if (bits & 1) {
+				if (bits==1) {
+					break; // The last CRTC
+				}
+				struct wlr_drm_crtc *crtc = &drm->crtcs[crtc_bit];
+				if (type == DRM_PLANE_TYPE_PRIMARY && crtc->primary) {
```
emersion, line 234: A `switch` would make this more readable.

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+		assert(num_crtcs<=64);
+		uint64_t bits = plane->possible_crtcs ;
+		// Some drm implementations (e.g. vc4 on raspberry pi) may set more
+		// bits than there are CRTCs so the unused upper bits must be cleared
+		// to insure a proper detection of the last possible CRTC.
+		bits &= (((uint64_t)1u)<<num_crtcs) - 1;
+		for (crtc_bit = 0 ; crtc_bit < num_crtcs ; crtc_bit++) {
+			if (bits & 1) {
+				if (bits==1) {
+					break; // The last CRTC
```
emersion, line 231: Why is this necessary? `crtc_bit < num_crtcs` should end the loop at the correct time, right?

backend/drm/drm.c
```diff
@@ -208,10 +209,41 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
+                 */
+
+		int crtc_bit ;
+		int num_crtcs = (int)drm->num_crtcs ;
+		assert(num_crtcs<=64);
+		uint64_t bits = plane->possible_crtcs ;
+		// Some drm implementations (e.g. vc4 on raspberry pi) may set more
+		// bits than there are CRTCs so the unused upper bits must be cleared
+		// to insure a proper detection of the last possible CRTC.
+		bits &= (((uint64_t)1u)<<num_crtcs) - 1;
```
emersion, line 227: This seems to only be required because of the `if (bits==1)` check.

backend/drm/drm.c
schauveau: No. If the for loop terminates via crtc_bit < num_crtcs  then this is an error case (no valid CRTC for the plane) and the assert is triggered.  That break is there to catch the last bit set in possible_crtcs. 
For example, consider the following hypothetical case:
 - 4 crtcs in 2 groups of 2 
 - 8 overlay planes. the first half have possible_crtcs = 0b0011 while the second half have 0b1100.    
The break is there to insure that the supplementary overlay planes (those that would not pass the test crtc->num_overlays>0) are assigned to a valid crtc. 

More generally, if you plan to accept the pull request then it is probably better to consider the the alternative implementation I proposed in https://github.com/swaywm/wlroots/pull/2333#discussion_r456779587.

emersion: >alternative implementation

Yeah that one looks better

schauveau:
So here is the updated loop inspired from my previous comment  https://github.com/swaywm/wlroots/pull/2333#discussion_r456779587. I did not use a switch as proposed by @emersion but I reversed the test condition to avoid using continue which imho makes the code more intuitive.

backend/drm/drm.c
emersion: https://lists.freedesktop.org/archives/dri-devel/2020-August/275142.html

tmlind:
Looks like thigs gets things initialized also for droid4 with the linux_openvrsgx and related blobs. I suggest that you do an updated pull request with the several patches merged together as the end changes are quite simple. Also, it seems you can decrease the nestedness a bit with checking for if (!((plane->possible_crtcs >> j) & 1)) in the beginning of the loop and do continue :)

emersion:
All right, I've changed the logic a little to make the code more readable. Should have the same behaviour as before.

Can you test?

schauveau:
I am afraid that testing is currently not very easy for me because I do not build anymore wlroots & sway from sources and also I switched to a more recent 5.x kernel where the CRT planes bits are not 'broken' anymore.

backend/drm/drm.c
```diff
@@ -197,36 +197,28 @@ static bool init_planes(struct wlr_drm_backend *drm) {
...
-		 * counting from 1), crtc 0 is possible.
-		 */
-		int crtc_bit = ffs(plane->possible_crtcs) - 1;
-
-		// This would be a kernel bug
-		assert(crtc_bit >= 0 && (size_t)crtc_bit < drm->num_crtcs);
+		// We don't really care about overlay planes, as we don't support them
+		// yet.
+		if (type == DRM_PLANE_TYPE_OVERLAY) {
+			continue;
```
ammen99, line 203: should drmModeFreePlane be called here? It is called in master.

ammen99 (commented):
I believe I understand the logic and it seems an improvement over the previous one, however, take my opinion with a grain of salt :)

backend/drm/drm.c
emersion: Good catch, fixed!

tmlind:
Works for me on droid4 now thanks!

PR #2336, opened 2020-07-18
xwm: reject requests for WM_STATE_ICONIC

junglerobba:
Fix for https://github.com/swaywm/sway/issues/4324
Fix for this issue in i3: https://github.com/i3/i3/pull/3421

For compatibility reasons, wine sets the iconic state without
waiting for the window manager, so if it is unsupported, there is
no way to recover the window from that state.
This simply sets the state back to normal immediately to work around
that problem.


emersion:
Thanks for sending a PR!

Some wlroots compositors might want to handle iconic state differently than Sway. wlroots cannot be opinionated and reject iconic on behalf of the compositor.

Instead, we should implement something similar to https://github.com/swaywm/wlroots/pull/2264

PR #2337, opened 2020-07-18
xwm: add support for xwayland minimize

junglerobba:
Since https://github.com/swaywm/wlroots/pull/2336 was the wrong approach,
I adopted the work from https://github.com/swaywm/wlroots/pull/2264 and I hope I didn't
miss any of the open issues on it.

I'm also not quite sure if the minimize event is necessary, so I left it in a separate commit for now.

A patch for sway that simply rejects minimize requests can be found here https://github.com/junglerobba/sway/commit/9a5d8422d3aee1e1c1fa05def9d26cd261cb7398 (caution, because it causes heavy flickering on wine games that were
previously acting as described here https://github.com/swaywm/sway/issues/4324 )


ammen99:
> I'm also not quite sure if the minimize event is necessary, so I left it in a separate commit for now.

Looking at #2264 the minimize event should be necessary for any compositor which actually supports minimization (for example Wayfire).

soreau:
> > I'm also not quite sure if the minimize event is necessary, so I left it in a separate commit for now.
> 
> Looking at #2264 the minimize event should be necessary for any compositor which actually supports minimization (for example Wayfire).

I have tested and wayfire can work with or without the second patch, it's just an extra LOC or two in wayfire with it. I tested minimizing from xwayland csd client as well as minimize from the panel window list. If I understand correctly, the event is present regardless but with the second patch, there is an associated event struct, which doesn't seem to be necessary.

ammen99:
> If I understand correctly, the event is present regardless but with the second patch, there is an associated event struct, which doesn't seem to be necessary.

The event struct is necessary to differentiate between client minimizing and restoring itself. While usually you don't interact with a minimized view (so it won't restore itself), maybe this would be necessary for applications which can be activated via dbus or something?

xwayland/xwm.c
```diff
@@ -1165,6 +1177,19 @@ static void xwm_handle_net_wm_state_message(struct wlr_xwm *xwm,
...
 		wlr_signal_emit_safe(&xsurface->events.request_maximize, xsurface);
 	}
+
+	if (minimized != xsurface->minimized) {
+		if (xsurface->minimized) {
+			xsurface->saved_width = xsurface->width;
+			xsurface->saved_height = xsurface->height;
+		}
+
+		struct wlr_xwayland_minimize_event minimize_event;
```
emersion, line 1187: Can we use an initializer here to avoid future fields from being undefined?

```c
struct wlr_xwayland_minimize_event event = {
	.surface = xsurface,
	…
};
```

xwayland/xwm.c
```diff
@@ -1202,6 +1227,29 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+static void xwm_handle_wm_change_state_message(struct wlr_xwm *xwm,
+		xcb_client_message_event_t *ev) {
+	struct wlr_xwayland_surface *xsurface = lookup_surface(xwm, ev->window);
+	uint32_t detail = ev->data.data32[0];
+
+	if (xsurface == NULL) {
+		return;
+	}
+
+	struct wlr_xwayland_minimize_event minimize_event;
```
emersion, line 1239: Ditto

PR #2339, opened 2020-07-19
xwayland: do not allow apps to change focus after wlroots request

### LINKED ISSUE #2324: Xwayland gets stuck in endless loop with multiple-windows-per-process clients
Initially I found this in Wayfire: https://github.com/WayfireWM/wayfire/issues/561

However, I could reproduce in Sway too:

1. Add bindings to move window to another workspace and to focus that workspace and the new window in it (it is important to focus immediately when switching, otherwise it works):
```
bindsym $mod+Shift+1 move container to workspace 1; workspace 1
```
2. Start Sway with at least 2 monitors
3. Open chrome on one workspace.
4. Open a new chrome window (ctrl-N)
5. Use the binding to move the chrome window to the workspace on the other monitor.

The window freezes, Xwayland and wlroots both have a very high CPU usage (~100%). @damianatorrpm managed to bisect it and the problem is https://github.com/swaywm/wlroots/pull/2219

My theory as to why it happens (say the two windows are X and Y)

1. Client tries to focus X
2. Compositor focuses Y
3. Message from 1. arrives in wlroots, we focus X
4. Message from 2. arrives in client, it resets the focus to Y
5. Message from 3. arrives in client, it resets focus to X
6. Message from 4. arrives in wlroots, it focuses Y

Notice 5./6. is same as 1./2. so we get an infinite loop. Not sure if this is really the problem but something like that is happening. Ideas?

ammen99:
This is an attempt to fix #2324 

@Xyene or someone else who uses JetBrains IDEs: can you check that this does not break #2219?

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -258,8 +258,9 @@ static void xwm_send_focus_window(struct wlr_xwm *xwm,
 	} else {
 		xwm_send_wm_message(xsurface, &message_data, XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT);
 
-		xcb_set_input_focus(xwm->xcb_conn, XCB_INPUT_FOCUS_POINTER_ROOT,
-			xsurface->window_id, XCB_CURRENT_TIME);
+		xcb_void_cookie_t cookie = xcb_set_input_focus(xwm->xcb_conn,
+			XCB_INPUT_FOCUS_POINTER_ROOT, xsurface->window_id, XCB_CURRENT_TIME);
+		xwm->last_focus_seq = cookie.sequence;
 	}
 
 	uint32_t values[1];
@@ -1246,6 +1247,11 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
 	struct wlr_xwayland_surface *requested_focus = lookup_surface(xwm, ev->event);
 	if (xwm->focus_surface && requested_focus &&
 			requested_focus->pid == xwm->focus_surface->pid) {
+		// We let the application override focus immediately after our focus change
+		// request, but not later
+		if (ev->sequence != xwm->last_focus_seq) {
+			return;
+		}
 		xwm->focus_surface = requested_focus;
 	}
 
--- a/include/xwayland/xwm.h
+++ b/include/xwayland/xwm.h
@@ -125,6 +125,7 @@ struct wlr_xwm {
 #if WLR_HAS_XCB_ERRORS
 	xcb_errors_context_t *errors_context;
 #endif
+	unsigned int last_focus_seq;
 
 	struct wl_listener compositor_new_surface;
 	struct wl_listener compositor_destroy;


xwayland/xwm.c
```diff
@@ -1246,6 +1248,11 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
 	struct wlr_xwayland_surface *requested_focus = lookup_surface(xwm, ev->event);
 	if (xwm->focus_surface && requested_focus &&
 			requested_focus->pid == xwm->focus_surface->pid) {
+		// We let the application override focus immediately after our focus change
+		// request, but not later
+		if (ev->sequence != xwm->last_focus_seq) {
```
psychon, line 1253: Are there any cases where this is `==`? It feels to me like this is basically an `if (true)`:

- `xwm_send_focus_window()` calls `xcb_set_input_focus()` and records the sequence number
- Immediately afterwards it calls `xcb_configure_window()`

Most likely the two requests are sent to the X11 server in a single call to `write()`, thus it it basically impossible for anything to cause a focus change in between the two requests. (and the sequence number of an event is the sequence of the last request that was handled by the server, so should always be the newer one.)

ammen99: Yes, there are actually quite a lot of cases where this is `ev->sequence == xwm->last_focus_seq`. Just add a `wlr_log` after this `if`, and run steam.

Unfortunately, I can't disprove your logic either :/

psychon: Okay, thanks for checking. In other news, I have no idea what is happening here. :-)

Xyene:
Thanks, I can confirm this doesn't break JetBrains IDEs.

xwayland/xwm.c
```diff
@@ -1236,6 +1237,7 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
 		return;
 	}
 
+
```
Xyene, line 1240: Nit: this newline shouldn't be here.

ammen99: fixed, thanks :)

xwayland/xwm.c
psychon: Well, I do have an idea what is going on. Of course, `xcb_set_input_focus()` causes a focus change! That's its whole job. That focus change causes a `FocusIn` event to be sent, which will satisfy the `sequence` condition.

I guess that means that this patch basically "kills" what this code is supposed to do: Allowing apps to change the focus themselves. From the comment above, that should bring back https://github.com/swaywm/sway/issues/1865.

ammen99: > From the comment above, that should bring back [swaywm/sway#1865](https://github.com/swaywm/sway/issues/1865)

Except that it doesn't. I have verified steam and it works as expected, even with this patch.

Also what you said cannot be true - otherwise, the current code in master would trigger an infinite loop with Xwayland, because we'd be continuously sending focus in response to focus in event.

soreau:
I can confirm this fixes #2324 and does not bring back swaywm/sway#1865, steam menu click still works.

xwayland/xwm.c
psychon: A `FocusIn` event is only generated when the focus actually changes. So, setting the focus to whatever already has the focus is a no-op. That would stop the loop.

Thanks for testing that Steam still works and sorry that I am always coming up with broken theories.

xwayland/xwm.c
```diff
@@ -1246,6 +1247,11 @@ static void xwm_handle_focus_in(struct wlr_xwm *xwm,
 	struct wlr_xwayland_surface *requested_focus = lookup_surface(xwm, ev->event);
 	if (xwm->focus_surface && requested_focus &&
 			requested_focus->pid == xwm->focus_surface->pid) {
+		// We let the application override focus immediately after our focus change
+		// request, but not later
+		if (ev->sequence != xwm->last_focus_seq) {
+			return;
```
emersion, line 1253: This ignores focus change events instead of denying them. This means X11 clients can now overwrite the focus (and wlroots' state will be out-of-sync with xserver's).

ammen99:
Unfortunately @emersion is right, and also steam needs to change focus long after we focus it, so this approach cannot work.

Issue #1378, opened 2018-11-19
Support 10bit color 


maletor:
Discussed previously in the past. 10 bit color should work if configured (or detected?) properly.

https://github.com/swaywm/sway/issues/1348#issuecomment-328394287
https://github.com/swaywm/wlroots/issues/54#issuecomment-321395585

emersion:
Can we composite in 10-bit? (See `GL_EXT_color_buffer_float`)

AntiSC2:
Wayland/Weston has now started work on HDR support with a proposed design here: https://lists.freedesktop.org/archives/wayland-devel/2019-January/039808.html

AndreasBackx:
Any update on this perhaps? Would be up for testing this if needed.

tadeokondrak:
I was able to get wlroots working with 10-bit output:
```diff
diff --git a/backend/drm/drm.c b/backend/drm/drm.c
index 60f8ecaa..fb0602f3 100644
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -109,10 +109,11 @@ static bool add_plane(struct wlr_drm_backend *drm,
 	for (size_t j = 0; j < drm_plane->count_formats; ++j) {
 		uint32_t fmt = drm_plane->formats[j];
 
-		if (fmt == DRM_FORMAT_ARGB8888) {
-			// Prefer formats with alpha channel
+		if (fmt == DRM_FORMAT_XRGB2101010) {
 			rgb_format = fmt;
 			break;
+		} else if (fmt == DRM_FORMAT_ARGB8888) {
+			rgb_format = fmt;
 		} else if (fmt == DRM_FORMAT_XRGB8888) {
 			rgb_format = fmt;
 		}
diff --git a/backend/drm/renderer.c b/backend/drm/renderer.c
index 25101a4b..7ff0f023 100644
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -32,7 +32,7 @@ bool init_drm_renderer(struct wlr_drm_backend *drm,
 		EGL_NONE,
 	};
 
-	renderer->gbm_format = GBM_FORMAT_ARGB8888;
+	renderer->gbm_format = GBM_FORMAT_ARGB2101010;
 	renderer->wlr_rend = create_renderer_func(&renderer->egl,
 		EGL_PLATFORM_GBM_MESA, renderer->gbm,
 		config_attribs, renderer->gbm_format);
```
[...]

PR #2340, opened 2020-07-20
[WIP] add 10bit color support

DerVerruckteFuchs:
This is an initial pull request to address issue #1378 using @tadeokondrak's patch posted in the issue as a starting point. I've added some of the color formats contained within the `wl_shm_format` enum that are in` /usr/include/wayland-server-protocol.h `. The initial 10bit color formats that have been added seem like a reasonable starting place, given that they look like 10bit versions of the 8bit color formats already being used. In future, adding YCbCr color formats seem like a good addition for TV/HDMI output.

I'm not quite sure if `wlr_output_set_format()` is doing enough when setting the color format. It looks fine, but I don't know if I need to be using `output->pending.format = format;` instead of `output->format = format;`. From what I gather from reading `types/wlr_output.h` and `interfaces/wlr_output.h`, the pending state is used for checking output state changes before committing them. I'm thinking checking before committing the change is possibly desirable in the event someone tries to set 10bit output to an 8bit monitor (and/or using an inadequate video cable). I'm not sure how to check for that at the moment.

Also, I'm not sure if `renderer->gbm_format` should get changed. I'm supposing 8bit color by default is safer than trying to use 10bit color on an 8bit monitor.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -109,11 +109,17 @@ static bool add_plane(struct wlr_drm_backend *drm,
 	for (size_t j = 0; j < drm_plane->count_formats; ++j) {
 		uint32_t fmt = drm_plane->formats[j];
 
-		if (fmt == DRM_FORMAT_XRGB2101010) {
+		if (fmt == DRM_FORMAT_ARGB2101010) {
+			// Prefer formats with 10 bit color
+			// and/or alhpa channel.
+			rgb_format = fmt;
+			break;
+		} else if (fmt == DRM_FORMAT_XRGB2101010) {
 			rgb_format = fmt;
 			break;
 		} else if (fmt == DRM_FORMAT_ARGB8888) {
 			rgb_format = fmt;
+			break;
 		} else if (fmt == DRM_FORMAT_XRGB8888) {
 			rgb_format = fmt;
 		}
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -109,10 +109,11 @@ static bool add_plane(struct wlr_drm_backend *drm,
 	for (size_t j = 0; j < drm_plane->count_formats; ++j) {
 		uint32_t fmt = drm_plane->formats[j];
 
-		if (fmt == DRM_FORMAT_ARGB8888) {
-			// Prefer formats with alpha channel
+		if (fmt == DRM_FORMAT_XRGB2101010) {
 			rgb_format = fmt;
 			break;
+		} else if (fmt == DRM_FORMAT_ARGB8888) {
+			rgb_format = fmt;
 		} else if (fmt == DRM_FORMAT_XRGB8888) {
 			rgb_format = fmt;
 		}
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -138,6 +138,7 @@ struct wlr_output {
 	bool enabled;
 	float scale;
 	enum wl_output_subpixel subpixel;
+	enum wl_shm_format format;
 	enum wl_output_transform transform;
 	enum wlr_output_adaptive_sync_status adaptive_sync_status;
 
@@ -286,6 +287,12 @@ void wlr_output_set_scale(struct wlr_output *output, float scale);
 void wlr_output_set_subpixel(struct wlr_output *output,
 	enum wl_output_subpixel subpixel);
 void wlr_output_set_description(struct wlr_output *output, const char *desc);
+/**
+ * Set the color format mode for the output.
+ *
+ */
+void wlr_output_set_format(struct wlr_output *output,
+	enum wl_shm_format format);
 /**
  * Schedule a done event.
  *
--- a/render/gles2/pixel_format.c
+++ b/render/gles2/pixel_format.c
@@ -39,9 +39,40 @@ static const struct wlr_gles2_pixel_format formats[] = {
 		.gl_type = GL_UNSIGNED_BYTE,
 		.has_alpha = true,
 	},
+		// 10 bit formats
+	{	
+		.wl_format = WL_SHM_FORMAT_ARGB2101010,
+		.depth = 32,
+		.bpp = 32,
+		.gl_format = GL_BGRA_EXT,
+		.gl_type = GL_UNSIGNED_BYTE,
+		.has_alpha = true,
+	},
+	{
+		.wl_format = WL_SHM_FORMAT_XRGB2101010,
+		.depth = 24,
+		.bpp = 32,
+		.gl_format = GL_BGRA_EXT,
+		.gl_type = GL_UNSIGNED_BYTE,
+		.has_alpha = false,
+	},
+	{
+		.wl_format = WL_SHM_FORMAT_XBGR2101010,
+		.depth = 24,
+		.bpp = 32,
+		.gl_format = GL_RGBA,
+		.gl_type = GL_UNSIGNED_BYTE,
+		.has_alpha = false,
+	},
+	{
+		.wl_format = WL_SHM_FORMAT_ABGR2101010,
+		.depth = 32,
+		.bpp = 32,
+		.gl_format = GL_RGBA,
+		.gl_type = GL_UNSIGNED_BYTE,
+		.has_alpha = true,
+	},
 };
-
-// TODO: more pixel formats
 
 const struct wlr_gles2_pixel_format *get_gles2_format_from_wl(
 		enum wl_shm_format fmt) {
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -267,6 +267,15 @@ void wlr_output_set_subpixel(struct wlr_output *output,
 	wlr_output_schedule_done(output);
 }
 
+void wlr_output_set_format(struct wlr_output *output,
+		enum wl_shm_format format) {
+	if (output->format == format) {
+		return;
+	}
+
+	output->format = format;
+}
+
 void wlr_output_set_description(struct wlr_output *output, const char *desc) {
 	if (output->description != NULL && desc != NULL &&
 			strcmp(output->description, desc) == 0) {
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -35,7 +35,7 @@ bool init_drm_renderer(struct wlr_drm_backend *drm,
 		EGL_NONE,
 	};
 
-	renderer->gbm_format = GBM_FORMAT_ARGB8888;
+	renderer->gbm_format = GBM_FORMAT_ARGB2101010;
 	renderer->wlr_rend = create_renderer_func(&renderer->egl,
 		EGL_PLATFORM_GBM_MESA, renderer->gbm,
 		config_attribs, renderer->gbm_format);
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -34,7 +34,8 @@ bool init_drm_renderer(struct wlr_drm_backend *drm,
 		EGL_ALPHA_SIZE, 1,
 		EGL_NONE,
 	};
-
+	
+	// TODO: allow setting from sway config/args. Have sane default set.
 	renderer->gbm_format = GBM_FORMAT_ARGB2101010;
 	renderer->wlr_rend = create_renderer_func(&renderer->egl,
 		EGL_PLATFORM_GBM_MESA, renderer->gbm,


types/wlr_output.c
```diff
@@ -267,6 +267,15 @@ void wlr_output_set_subpixel(struct wlr_output *output,
 	wlr_output_schedule_done(output);
 }
 
+void wlr_output_set_format(struct wlr_output *output,
```
emersion, line 270: This seems to be a no-op.

emersion:
There should at least be a way to disable 10-bit, because it uses more bandwidth than 8-bit (so it could prevent additional screens from being turned on because of GPU limitations).

include/wlr/types/wlr_output.h
```diff
@@ -286,6 +287,12 @@ void wlr_output_set_scale(struct wlr_output *output, float scale);
 void wlr_output_set_subpixel(struct wlr_output *output,
 	enum wl_output_subpixel subpixel);
 void wlr_output_set_description(struct wlr_output *output, const char *desc);
+/**
+ * Set the color format mode for the output.
+ *
```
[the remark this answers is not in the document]
DerVerruckteFuchs: Got it.

DerVerruckteFuchs:
@emersion From the sway side of things, I'm wanting to add an output argument for setting 8bit/10bit color. It should let the user pick which color bitness they want on a per output basis. It's been a while since I've looked at the sway code for setting output arguments, but I do have a start with my 10bit branch for sway. I'm going to look closer at how the other output defaults are set in wlroots. I think 8bit color is probably a sane default.

The limitation I've noticed most often with 10bit color/HDR on monitors, is lower FPS for high refresh rate monitors. I do want the user to be able to choose between high(est) FPS or HDR. Switching from high FPS for games to HDR for video content (or vice versa) would be a use case I'd expect. I'm less familiar with color bitness reducing the number of monitors that can be used for output, but that sounds like that would be more likely to come up with Displayport or Thunderbolt daisy-chaining.

types/wlr_output.c
DerVerruckteFuchs: Right, so my understanding of the other output setting functions was to skip all of the `output->pending` and state setting if the output setting had not changed. Right now `wlr_output_set_format()` doesn't have any of the `output->pending` or state change stuff, and I wasn't sure if any of it was necessary to do things safely. In `wlr_output_enable_adaptive_sync()`, both state changing and `output->pending` shows up there, while `wlr_output_set_subpixel()` is doing something different.

backend/drm/drm.c
```diff
@@ -109,8 +109,15 @@ static bool add_plane(struct wlr_drm_backend *drm,
 	for (size_t j = 0; j < drm_plane->count_formats; ++j) {
 		uint32_t fmt = drm_plane->formats[j];
 
-		if (fmt == DRM_FORMAT_ARGB8888) {
-			// Prefer formats with alpha channel
+		if (fmt == DRM_FORMAT_ARGB2101010) {
+			// Prefer formats with 10 bit color
```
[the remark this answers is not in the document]
DerVerruckteFuchs: I've fixed the typo and collapsed the if blocks. I'm not sure what the ordering is. From skimming the code, it might be related to `formats` in `render/gles2/pixel_format.c`. The order added may be from the order added with `wlr_drm_format_set_add()`. For all I know, the ordering is related to the numbered order of monitors. If that's the case, it would be fine since some monitors could be set to 10-bit and others to 8-bit color. I'd need to read quite a bit more to see how everything is fitting together with `drm_plane->formats` though.

types/wlr_output.c
emersion: `wlr_output_set_subpixel` doesn't need to interact with the backend, so doesn't need to use `output->pending`.

`wlr_output_set_format` should make the backend de-allocate and re-allocate a buffer with the correct format right? This doesn't seem to be the case right now, `output->format` is never used.

render/gles2/pixel_format.c
```diff
@@ -39,10 +39,41 @@ static const struct wlr_gles2_pixel_format formats[] = {
...
+		.wl_format = WL_SHM_FORMAT_ARGB2101010,
+		.depth = 32,
+		.bpp = 32,
+		.gl_format = GL_BGRA_EXT,
+		.gl_type = GL_UNSIGNED_BYTE,
+		.has_alpha = true,
+	},
+	{
+		.wl_format = WL_SHM_FORMAT_XRGB2101010,
+		.depth = 24,
```
emersion, line 53: Depth is 30

render/gles2/pixel_format.c
```diff
@@ -39,10 +39,41 @@ static const struct wlr_gles2_pixel_format formats[] = {
...
+		.bpp = 32,
+		.gl_format = GL_BGRA_EXT,
+		.gl_type = GL_UNSIGNED_BYTE,
+		.has_alpha = false,
+	},
+	{
+		.wl_format = WL_SHM_FORMAT_XBGR2101010,
+		.depth = 24,
```
emersion, line 61: Ditto

render/gles2/pixel_format.c
```diff
@@ -39,10 +39,41 @@ static const struct wlr_gles2_pixel_format formats[] = {
 		.gl_type = GL_UNSIGNED_BYTE,
 		.has_alpha = true,
 	},
+		// 10 bit formats
```
emersion, line 42: I don't think this is enough to support these buffer formats. I'd assume we need shaders as well?

Have you tested a 10-bit client?

emersion:
Support for FP16 client buffers would be nice to have as well. As far as I know games will in general use this format.

>From the sway side of things, I'm wanting to add an output argument for setting 8bit/10bit color

Looks like a good idea.

>I'm less familiar with color bitness reducing the number of monitors that can be used for output, but that sounds like that would be more likely to come up with Displayport or Thunderbolt daisy-chaining.

Also more likely to come up with Intel, which has a rather small bandwidth limitation shared by all CRTCs.

render/gles2/pixel_format.c
emersion: Also, this list is only used for `wl_shm` clients. GPU-accelerated clients don't use it.

render/gles2/pixel_format.c
DerVerruckteFuchs: Got it.

render/gles2/pixel_format.c
DerVerruckteFuchs: Got this one as well.

render/gles2/pixel_format.c
DerVerruckteFuchs: I have a 10-bit branch of sway I tested with my 10-bit wlroots branch some time ago. I used the method of running my compiled version of sway/wlroots like in the wiki, and start it with a line in my `~/.zprofile` so that my compiled version of sway gets used on login instead of the package managed version. It worked mostly as expected, aside from the mouse cursor having a corrupted sprite/texture. However, I'm not sure if it was from my changes, or if it was from some other commits/changes. At the time it was mostly just applying @tadeokondrak's patch, and compiling that with sway. I started some very initial work with 10-bit color in sway that includes the aforementioned output argument. I'm not sure if the output argument was working properly, since I think the patch I started with may have been hard coding the `gbm_format` stuff. I was able to log in.

It's been awhile since I've tested it that way. I wasn't confident with the HDMI cable I was using then, so since then I have gotten a proper certified 2.0b cable. I've been using a HDR capable smart tv. I can toggle between YCbCr and RGB in the TV settings. Assuming the TV's menu is reporting things accurately, it should also tell me if it is receiving 10-bit HDR input. I'll need to check if just having it be 10-bit color is enough for that, or if 10-bit HDR is a more specific specification than that. I'm assuming that it may need to be YCbCr for it to work, but the RGB option on the TV makes me think it might be more flexible. If not, I'll need to check my collection of monitors to see if any of them are 10-bit or not.

> Also, this list is only used for `wl_shm` clients. GPU-accelerated clients don't use it.

Duly noted. I'll be looking into that along with the shaders you've mentioned.

DerVerruckteFuchs: I've tested my more recent commits. The sway log is spitting out some errors.

```
00:01:16.638 [ERROR] [EGL] command: eglCreatePlatformWindowSurfaceEXT, error: EGL_BAD_MATCH (0x3009), message: "EGL config not compatible with GBM format"
00:01:16.638 [ERROR] [render/egl.c:406] Failed to create EGL surface
00:01:16.638 [ERROR] [backend/drm/renderer.c:103] Failed to create EGL surface
00:01:16.638 [ERROR] [backend/drm/drm.c:895] Cannot allocate cursor resources
00:01:16.638 [DEBUG] [types/wlr_output.c:1050] Falling back to software cursor on output 'HDMI-A-4'
```

I can log in, but the TV is telling me it is not getting HDR input with either the RGB or YCbCr input setting on the TV. AFAIK, the EGL stuff has to do with shaders, so I'll need to look closer at what is going on with the GBM format. Given there was a fallback to software cursors and the older issue I had was a garbled cursor, I'm wondering if I had a similar EGL/GBM issue with the older issue as well.


I'll test again after adding something for GPU-accelerated clients. It looks like my login session was GPU-accelerated. It seems I didn't add color formats for GPU-accelerated clients, so mine couldn't find what it needed for 10-bit color.

DerVerruckteFuchs:
>Support for FP16 client buffers would be nice to have as well. As far as I know games will in general use this format.

Good to know. I do want my games to work properly.

>Also more likely to come up with Intel, which has a rather small bandwidth limitation shared by all CRTCs.

Intel is getting into dedicated GPUs, so I suspect this may be limited to their integrated GPUs unless Intel beefs them up in the future. I'm guessing this is at least part of the reason Intel enables frame buffer compression by default for Skylake+ in Linux.

p4block:
I built your branch and tested on my system in an attempt to mitigate the massive color banding issues I was having. I'm pretty sure there has been a massive improvement, but _placebo is a thing_. As a user, can I do something to help test more?

DerVerruckteFuchs:
> I built your branch and tested on my system in an attempt to mitigate the massive color banding issues I was having. I'm pretty sure there has been a massive improvement, but _placebo is a thing_. As a user, can I do something to help test more?

In my 10-bit sway branch, I'm working on getting color bitness to be displayed in the `swaymsg -t get_outputs` output. Right now it just crashes since I need to set things up properly so that the color bitness status value actually exists. It's been awhile since I've played with the color status color bitness, so that can't be tested until it's fixed. Also making sure that the color bitness status value properly correlates to the running client will need to be done.

As far as right now, if you're using sway, check sway's log output. I don't have GPU-accelerated clients working yet since I need to add them. If you're using anything else with wlroots, if there is log output, make sure you're actually getting 10-bit color.

Other than that, just keep on eye on any new commits coming into this PR if you want to test them.

render/gles2/pixel_format.c
DerVerruckteFuchs: >This is surprising to me. Isn't the GPU responsible for switching between YCbCr and RGB?

As far as I'm aware. The colors do visibly change when I switch between YCbCr and RGB. It's a Vizio if you're curious. It looks like some Samsung TVs may also have this option. I guess it just asks the source for YCbCr/RGB output? As I look into it more, it seems to be for PC input. I guess some GPUs might not handle YCbCr for some reason or another? Or maybe it's a weird manufacturer design decision?

types/wlr_output.c
```diff
@@ -267,6 +267,15 @@ void wlr_output_set_subpixel(struct wlr_output *output,
...
+		enum wl_shm_format format) {
+	if (output->format == format) {
```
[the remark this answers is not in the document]
DerVerruckteFuchs: I was looking at the `wlr_output_set_subpixel()` function above it, and noticed it had a similar `if` statement. I assumed it was there to avoid setting `output->subpixel` to the same subpixel layout it was already set to, and maybe it would be bad to set it twice when not needed. I figured if there was a reason the `if` statement exists in `wlr_output_set_subpixel()`, then maybe it would be a good idea to also not set `output->format` twice when not needed. If there is no benefit to the `if` statement in `wlr_output_set_format()`, I can remove it.

render/gles2/pixel_format.c
```diff
@@ -39,10 +39,41 @@ static const struct wlr_gles2_pixel_format formats[] = {
...
+	{	
+		.wl_format = WL_SHM_FORMAT_ARGB2101010,
+		.depth = 32,
+		.bpp = 32,
+		.gl_format = GL_BGRA_EXT,
+		.gl_type = GL_UNSIGNED_BYTE,
```
emersion, line 48: This can't work. I think it needs `GL_UNSIGNED_INT_2_10_10_10_REV`, which is only available in GLES3.

We need a way to test it to make sure it works anyways.

swick:
Just FYI most scanout engines which support scanout from fp16 or 10/12 bit formats use dithering if the connector format has less precision which can help if the user perceives banding artifacts.

emersion:
Note, 10-bit is not enough to composite HDR content. We'll really want fp16. Also see: https://docs.google.com/document/d/1A__vvTDKXt4qcuCcSN-vLzcQKuchnYmu8TFMDseIZJE/edit

emersion:
Ref https://gitlab.freedesktop.org/wayland/weston/-/issues/467

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2340

PR #2341, opened 2020-07-20
xwm: Set _NET_WM_STATE_FOCUSED property for the focused surface

soreau:
Certain clients require this property to be set for expected behavior.
Most notably, steam client CSD maximize button no longer worked
after unmaximizing once, unless the state was changed by another
method.

Tested with sway and wayfire. In wayfire the problem is easy to identify
because the window is able to be resized on maximize toggle. On sway,
without the patch, notice clicking the maximize button in steam will make
it 'stick' in the maximized state. With the patch, the button state continuously
cycles on subsequent clicks as expected.


ammen99:
I have a feeling that the focused state might not be reset after focus leaves the window.

soreau:
> I have a feeling that the focused state might not be reset after focus leaves the window.

I added a commit that unsets this property for the focus_surface if it was set, before setting it to a new surface and verified with xprop that it's being set/unset. Note that it only is unset when another xwayland window gains focus.

soreau:
I have merged both patches into one and rebased.

kode54:
I'm having an issue when I tried to combine this with #2339 and resulted in Discord losing keyboard focus semi-permanently. It only regains focus if another Xwayland app or window receives focus first. Reproduced in both sway and wayfire.

kode54:
Verified, this patch alone applied to current master, restarting Discord crashes Wayfire. Something needs debugging.

Log, for posterity:

[wayfire.5.log](https://github.com/swaywm/wlroots/files/4977043/wayfire.5.log)

kode54:
Okay, confirmed, this latest version fixes the issues that caused Discord to crash.

#2339 fixes the issues moving Discord's popout to a second output (without it, and even with this pull, it locks up Wayfire when moving between outputs) and with this pull, probably fixes other unseen issues, and no longer has the keyboard input focus issue it had before. I give it my clean bill of health.

xwayland/xwm.c
```diff
@@ -238,13 +239,20 @@ static void xwm_set_net_client_list(struct wlr_xwm *xwm) {
 			XCB_ATOM_WINDOW, 32, mapped_surfaces, windows);
 }
 
+static void xsurface_set_net_wm_state(struct wlr_xwayland_surface *xsurface);
+
 static void xwm_send_focus_window(struct wlr_xwm *xwm,
```
emersion, line 244: This function now not only sends focus events but also sets focus. Maybe rename to `xwm_set_focus_window`?

xwayland/xwm.c
```diff
@@ -270,6 +278,13 @@ static void xwm_send_focus_window(struct wlr_xwm *xwm,
 	values[0] = XCB_STACK_MODE_ABOVE;
 	xcb_configure_window(xwm->xcb_conn, xsurface->window_id,
 		XCB_CONFIG_WINDOW_STACK_MODE, values);
+
+	xsurface_set_net_wm_state(xsurface);
+unfocus:
```
emersion, line 283: Maybe move the `if` block at the top of the function, so that we don't need to have `goto`s?

xwayland/xwm.c
```diff
@@ -238,13 +239,20 @@ static void xwm_set_net_client_list(struct wlr_xwm *xwm) {
 			XCB_ATOM_WINDOW, 32, mapped_surfaces, windows);
 }
 
+static void xsurface_set_net_wm_state(struct wlr_xwayland_surface *xsurface);
+
 static void xwm_send_focus_window(struct wlr_xwm *xwm,
 		struct wlr_xwayland_surface *xsurface) {
+	struct wlr_xwayland_surface *unfocus_surface = xwm->focus_surface;
+
+	xwm->focus_surface = xsurface;
```
emersion, line 248: This function is a little special because it isn't a no-op when the surface is already focused (`xwm->focus_surface == xsurface`). This is necessary because we want to be able to deny `FocusIn` events. Maybe add a comment about this?

xwayland/xwm.c
```diff
@@ -238,12 +239,25 @@ static void xwm_set_net_client_list(struct wlr_xwm *xwm) {
...
+
+	if (unfocus_surface) {
+		xsurface_set_net_wm_state(unfocus_surface);
+	}
+
 	if (!xsurface) {
 		xcb_set_input_focus_checked(xwm->xcb_conn,
 			XCB_INPUT_FOCUS_POINTER_ROOT,
 			XCB_NONE, XCB_CURRENT_TIME);
+
```
emersion, line 260: Nit: unnecessary newline here

soreau: woops.. removed.

PR #2342, opened 2020-07-20
xdg-shell: split last-acked and current state

### LINKED ISSUE #762: xdg shell ack_configure copies pending state to current immediately
Should happen on the next commit instead

ifreund:
These states are distinct in the time period between the ack_configure
and the next commit on the surface. Splitting these states avoids the
following race for example:

- client starts at 1000x1000
- wlr_xdg_toplevel_set_size 500x500
- size is different -> configure sent
- client acks the configure
- wlr_xdg_toplevel_set_size 1000x1000
- compare_xdg_toplevel_state returns true since there is no pending
  configure and the currently committed size is still 1000x1000
- no new configure is sent
- client commits at the size it last acked, 500x500

For reference, this is the cause of https://github.com/ifreund/river/issues/31

Closes #762

### DIFF
--- a/types/xdg_shell/wlr_xdg_toplevel.c
+++ b/types/xdg_shell/wlr_xdg_toplevel.c
@@ -13,60 +13,44 @@ void handle_xdg_toplevel_ack_configure(
 	assert(surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL);
 	assert(configure->toplevel_state != NULL);
 
-	surface->toplevel->current.maximized =
-		configure->toplevel_state->maximized;
-	surface->toplevel->current.fullscreen =
-		configure->toplevel_state->fullscreen;
-	surface->toplevel->current.resizing =
-		configure->toplevel_state->resizing;
-	surface->toplevel->current.activated =
-		configure->toplevel_state->activated;
-	surface->toplevel->current.tiled =
-		configure->toplevel_state->tiled;
+	surface->toplevel->last_acked = *configure->toplevel_state;
 }
 
 bool compare_xdg_surface_toplevel_state(struct wlr_xdg_toplevel *state) {
-	struct {
-		struct wlr_xdg_toplevel_state state;
-		uint32_t width, height;
-	} configured;
-
 	// is pending state different from current state?
 	if (!state->base->configured) {
 		return false;
 	}
 
+	struct wlr_xdg_toplevel_state *configured = NULL;
 	if (wl_list_empty(&state->base->configure_list)) {
-		// last configure is actually the current state, just use it
-		configured.state = state->current;
-		configured.width = state->base->surface->current.width;
-		configured.height = state->base->surface->current.height;
+		// There are currently no pending configures, so check against the last
+		// state acked by the client.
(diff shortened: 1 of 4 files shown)


ifreund:
Found #762 which is fixed by this patch.

types/xdg_shell/wlr_xdg_toplevel.c
```diff
@@ -13,55 +13,39 @@ void handle_xdg_toplevel_ack_configure(
...
-		configure->toplevel_state->maximized;
-	surface->toplevel->current.fullscreen =
-		configure->toplevel_state->fullscreen;
-	surface->toplevel->current.resizing =
-		configure->toplevel_state->resizing;
-	surface->toplevel->current.activated =
-		configure->toplevel_state->activated;
-	surface->toplevel->current.tiled =
-		configure->toplevel_state->tiled;
+	surface->toplevel->last_acked = *configure->toplevel_state;
```
emersion, line 16: I don't know if we can just copy states like this, given that there is a `struct wl_listener fullscreen_output_destroy` field?

types/xdg_shell/wlr_xdg_toplevel.c
```diff
@@ -13,55 +13,39 @@ void handle_xdg_toplevel_ack_configure(
...
-		struct wlr_xdg_toplevel_state state;
-		uint32_t width, height;
-	} configured;
-
 	// is pending state different from current state?
 	if (!state->base->configured) {
 		return false;
 	}
 
+	struct wlr_xdg_toplevel_state configured;
```
emersion, line 25: Always zero-init structs

types/xdg_shell_v6/wlr_xdg_toplevel_v6.c
```diff
@@ -295,50 +295,36 @@ void handle_xdg_toplevel_v6_ack_configure(struct wlr_xdg_surface_v6 *surface,
...
-		struct wlr_xdg_toplevel_v6_state state;
-		uint32_t width, height;
-	} configured;
-
 	// is pending state different from current state?
 	if (!state->base->configured) {
 		return false;
 	}
 
+	struct wlr_xdg_toplevel_v6_state configured;
```
emersion, line 307: Ditto

emersion:
The motivation makes sense to me!

types/xdg_shell/wlr_xdg_toplevel.c
ifreund: As far as I could tell, these fields are only used in the `client_pending` state. Arguably they should simply be moved to `wlr_xdg_toplevel`.

types/xdg_shell/wlr_xdg_toplevel.c
ifreund: Done.

types/xdg_shell/wlr_xdg_toplevel.c
ifreund: #2366

emersion (approved):
LGTM, thanks!

Issue #1721, opened 2019-06-12
Crash when `xwayland_surface_create` returns `NULL`


FreeFull:
https://github.com/swaywm/sway/issues/4203#issuecomment-500917642

This can be reproduced in both sway and rootston, by launching three instances of pysolfc.

FreeFull:
I'm not sure if this bug is still reproducible. At least, I can't reproduce it using pysolfc any more (although that might be due to a change in pysolfc itself).

emersion:
Here's a fix: https://github.com/swaywm/wlroots/pull/2343

>I'm not sure if this bug is still reproducible. At least, I can't reproduce it using pysolfc any more (although that might be due to a change in pysolfc itself).

Probably because libwayland doesn't create one timerfd per event loop timer anymore.

PR #2343, opened 2020-07-21
xwayland/xwm: don't insert surface in list on error

emersion:
In case wl_event_loop_add_timer errors out, don't insert the free'd
wlr_xwayland_surface in the list.

Closes: https://github.com/swaywm/wlroots/issues/1721


ddevault:
Thanks!

PR #2348, opened 2020-07-25
Document minimum version of meson required

markstos:


emersion:
I'm not a fan of documenting minimum version in the README, because it's too easy for these to get out-of-sync with `meson.build`.

markstos:
I'm a software maintainer too -- fair enough. 

Thanks for the reply.

PR #2351, opened 2020-07-28
render/gles2: remove gles2_procs global

emersion:
_See individual commits._

* * *

Breaking change: the `wlr_gles2_texture_from_*` family of functions are no longer public API.


ddevault:
Thanks!

PR #2353, opened 2020-07-29
color management

akvadrako:
This patch adds some basic color management by using the lcms2 library. I first tried to use the higher-level libplacebo but integration was a mess. The changes to the shaders is pretty simple, so it was easier to just do it manually.

Some notes:
- The compositor can assign ICC profiles to wlr_surface and wlr_renderer. They are used when rendering textures and rectangles.
- The default is a built-in sRGB profile.
- If the input and output profiles match, the conversion is skipped.
- Currently it depends on some OpenGL extensions.
  - GL_EXT_texture_norm16 - this is required for 16-bit integer color lookup table. We could fallback to 8-bit ints or even floats without much trouble.
  - GL_OES_texture_3D - this is used to create the 3D CLUT. It could be done with a 1D or 2D lookup also - I don't know if there are performance considerations. If this doesn't exist, the color conversion is just skipped. However I do use glTexImage3D from the gl32.h header.
  - I wasn't sure if we need to stick to ES 2.0. ES 3.0 is 8 years old now.
- It doesn't seem to affect full-screen apps.

I did some benchmarks - with color conversion turned on during a almost fullscreen movie, my iGPU uses about 0.1W more and runs a few Mhz higher. glmark2 scores go down by about 5%. When color conversion is turned off, the difference was negligible from current master.

See also #1378.

### DIFF
--- a/include/wlr/render/color.h
+++ b/include/wlr/render/color.h
@@ -1,0 +1,19 @@
+#ifndef _WLR_COLOR_H
+#define _WLR_COLOR_H
+
+struct wlr_color_config {
+    char *icc_profile_path; // not null
+};
+
+/**
+ * Create a color config.
+ *
+ * icc_profile_path should not be NULL.
+ * It will be copied into the config.
+ */
+struct wlr_color_config *wlr_color_config_load(const char *icc_profile_path);
+
+struct wlr_color_config *wlr_color_config_copy(struct wlr_color_config *value);
+void wlr_color_config_free(struct wlr_color_config *config);
+
+#endif
--- a/render/gles2/color.c
+++ b/render/gles2/color.c
@@ -1,0 +1,241 @@
+// OpenGL Color conversion
+
+#define _XOPEN_SOURCE 500 // for strdup
+
+#include <GLES3/gl32.h>
+
+#include <assert.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <time.h>
+
+#include <wlr/types/wlr_list.h>
+#include <wlr/render/color.h>
+
+#include <lcms2.h>
+
+#include "render/gles2.h"
+
+// GL3+ or GL_EXT_texture_norm16 on ES 3.2
+#define GL_RGBA16 0x805B
+
+struct lut {
+	GLuint tex_id;
+	struct wlr_color_config *input, *output;
+	cmsHTRANSFORM transform;
+};
+
+#define LUTS_MAX 100
+
+static struct wlr_list luts;
+
+static int color_compare(void *av, void *bv) {
+	struct wlr_color_config *a = av, *b = bv;
+	if(a == b)
+		return 0;
+	if(!a)
+		return 1;
+	if(!b)
+		return -1;
+
+	return strcmp(a->icc_profile_path, b->icc_profile_path);
+}
+
+struct wlr_color_config *wlr_color_config_copy(struct wlr_color_config *value) {
+	if(! value)
+		return NULL;
+
+	struct wlr_color_config *copy = calloc(1, sizeof(*copy));
+	copy->icc_profile_path = strdup(value->icc_profile_path);
+	return copy;
+}
+
+static cmsHPROFILE color_load(struct wlr_color_config *color) {
+	if(color) {
+		cmsHPROFILE pro = cmsOpenProfileFromFile (color->icc_profile_path, "r");
+		if (pro == NULL) {
+			wlr_log(WLR_ERROR, "error loading icc profile %s", color->icc_profile_path);
+		} else {
+			return pro;
+		}
+	}
+	return cmsCreate_sRGBProfile();
+
+}
+
+static const char *color_describe(struct wlr_color_config *color) {
+	if(! color)
+		return "sRGB";
+
+	char *base = strrchr(color->icc_profile_path, '/');
+	return base ? base + 1 : color->icc_profile_path;
+}
+
+struct wlr_color_config *wlr_color_config_load(const char *icc_profile_path) {
+	assert(icc_profile_path);
+
+	if(0 == strlen(icc_profile_path)) {
+		return NULL;
+	}
+
+	bool can_access = access(icc_profile_path, F_OK) != -1;
+	if (!can_access) {
+		wlr_log(WLR_ERROR, "Unable to access color profile '%s'", icc_profile_path);
+		return NULL;
+	}
+
+	struct wlr_color_config *color = calloc(1, sizeof(*color));
+	color->icc_profile_path = strdup(icc_profile_path);
+	return color;
+}
+
+void wlr_color_config_free(struct wlr_color_config *color) {
+	if(! color)
+		return;
+	free(color->icc_profile_path);
+	free(color);
+}
+
+static struct lut *load_lut(struct wlr_color_config *input, struct wlr_color_config *output) {
+	if(0 == color_compare(input, output))
+		return NULL;
+
+	struct lut *lut;
+
+	// find an existing cached lut
+	for (size_t i = 0; i < luts.length; i++) {
+		lut = luts.items[i];
+		if(color_compare(lut->input, input) || color_compare(lut->output, output))
+			continue;
+
+		// move to head to list
+		if(i > 5) {
+			wlr_list_del(&luts, i);
+			wlr_list_insert(&luts, 0, lut);
+		}
+		return lut;
+	}
+
+	// clean an old LUT to avoid memory leak
+	if(luts.length > LUTS_MAX) {
+		struct lut *lut = wlr_list_pop(&luts);
+		wlr_log(WLR_DEBUG, "destroy CLUT %s -> %s", color_describe(lut->input), color_describe(lut->output));
+		wlr_color_config_free(lut->input);
+		wlr_color_config_free(lut->output);
+		cmsDeleteTransform(lut->transform);
+		glDeleteTextures(1, &lut->tex_id);
+		free(lut);
+	}
+
+	wlr_log(WLR_INFO, "create CLUT %s -> %s", color_describe(input), color_describe(output));
+	cmsHPROFILE inp = color_load(input);
+	cmsHPROFILE outp = color_load(output);
+
+	// According to lcms docs, black point compensation does not apply to perceptual intent, so we omit the flag.
+	cmsHTRANSFORM xform = cmsCreateTransform(
+			inp, TYPE_RGB_16,
+			outp, TYPE_RGB_16,
+			INTENT_PERCEPTUAL,
+			cmsFLAGS_HIGHRESPRECALC | cmsFLAGS_NOCACHE);
+
+	cmsCloseProfile(inp);
+	cmsCloseProfile(outp);
+
+	if (xform == NULL) {
+		wlr_log(WLR_ERROR, "failed to create color transform");
+		return NULL;
+	}
+
+	// save the new LUT
+	lut = malloc(sizeof(*lut));
+	*lut = (struct lut){
+		.input = wlr_color_config_copy(input),
+			.output = wlr_color_config_copy(output),
+			.transform = xform,
+	};
+	wlr_list_insert(&luts, 0, lut);
+	return lut;
+}
+
+void color_convert(struct wlr_color_config *ic, struct wlr_color_config *oc, const float input[static 4], float output[static 4]) {
+	struct lut *lut = load_lut(ic, oc);
+	if(! lut) {
+		memcpy(output, input, 4 * sizeof(float));
+		return;
+	}
+
+	uint16_t out[3], in[3] = {
+		input[0] * UINT16_MAX,
+		input[1] * UINT16_MAX,
+		input[2] * UINT16_MAX,
+	};
+	cmsDoTransform(lut->transform, in, out, 1);
+	output[0] = (float)out[0] / UINT16_MAX;
+	output[1] = (float)out[1] / UINT16_MAX;
+	output[2] = (float)out[2] / UINT16_MAX;
+	output[3] = input[3];
+}
+
+GLuint color_build_lut(struct wlr_color_config *input, struct wlr_color_config *output) {
+	struct lut *lut = load_lut(input, output);
+	if(! lut)
+		return 0;
+
+	if(lut->tex_id)
+		return lut->tex_id;
+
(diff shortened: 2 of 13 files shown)


render/gles2/renderer.c
```diff
@@ -182,11 +191,14 @@ static void gles2_render_quad_with_matrix(struct wlr_renderer *wlr_renderer,
 	float transposition[9];
 	wlr_matrix_transpose(transposition, matrix);
 
+	float converted[4];
+	color_convert(NULL, wlr_renderer->color, color, converted);
```
cyanreg, line 195: Of course this won't even do anything without a protocol to provide a client profile or colorspace. Which is why this _really_ needs a protocol before it can be merged. Otherwise people who currently care about color and let clients apply the profile themselves will be broken as they'll have a profile applied twice before display.

render/gles2/color.c
```diff
@@ -0,0 +1,241 @@
...
+	struct wlr_color_config *input, *output;
+	cmsHTRANSFORM transform;
+};
+
+#define LUTS_MAX 100
+
+static struct wlr_list luts;
+
+static int color_compare(void *av, void *bv) {
+	struct wlr_color_config *a = av, *b = bv;
```
cyanreg, line 36: You're comparing a struct containing a pointer to another struct containing a pointer. This won't ever work.
You should compare either the ICC profile bytes, the ICC profile parameters (better) or the colorspace (if provided).

include/wlr/render/color.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _WLR_COLOR_H
+#define _WLR_COLOR_H
+
+struct wlr_color_config {
+    char *icc_profile_path; // not null
```
cyanreg, line 5: You need a colorspace here as well. The colorspace is as good of a substitute of a profile as an actual profile.
That's why the eternally stuck in WIP color management protocol specifies both.
You need color primaries, transfer curve, and colorspace. And HDR parameters (which I'm not sure the protocol describes, which would be bad).

render/gles2/renderer.c
akvadrako: This case is covered - for color managed clients you can specify a color profile that matches the monitor.

render/gles2/color.c
akvadrako: The strcmp is there - I'm not quite sure what you're saying.

include/wlr/render/color.h
akvadrako: I originally had a colorspace too when I was using libplacebo. However libplacebo just ignores the colorspace when an ICC profile is set. My understanding, and that of the libplacebo developer, is that an ICC profile makes a colorspace redundant.

What would it be used for if we provided it?

cyanreg: Because colorspaces are an alternative to ICC profiles. If a colorspace is specified and a client specifies the same colorspace then no conversion should happen. I expect this to be common in the future, especially with HDR. Because you'll only know of a monitor's colorspace rather than get an ICC profile, and if you hope to have HDR on it you'll definitely need something different than bt709.

render/gles2/color.c
cyanreg: Didn't see it.

render/gles2/renderer.c
cyanreg: And how exactly is that sent such that the compositor doesn't do it again?

include/wlr/render/color.h
akvadrako: I agree colorspaces are an alternative, but you can do "color profile /usr/share/color/argyll/ref/Rec2020.icm" or some well-known path instead of making an additional config option.

The easier way to tell that the conversion can be skipped would be to check if the CLUT is a 1-to-1 mapping.

I don't know why you wouldn't have an ICC profile for a monitor - usable defaults are provided by colord from the EDID data.

render/gles2/renderer.c
akvadrako: Take a look at the sway PR https://github.com/swaywm/sway/pull/5586 - the ICC profile is set for each window.

myfreeweb: hm neither this nor the sway PR seem to [register with colord](https://gitlab.freedesktop.org/wayland/weston/-/blob/master/compositor/cms-colord.c) which is what would *inform* the clients of the profile so that they could do the conversion themselves

akvadrako: Correct - this is just basic manual color management - you specify the profiles for windows and outputs in your sway config.

emersion: colord is out-of-scope for wlroots.

include/wlr/render/color.h
cyanreg: EDID data is highly unreliable and inaccurate. I'd trust sRGB over an ICC generated from an EDID. To get a good ICC profile you really need either equipment, or the guarantee of a manufacturer.
And not everyone wants to install argyll at all.

akvadrako: You don't need to use the EDID data. You can use a standard profile. I'm not really against hard-coding colorspaces, but it seems to complicate things with no benefit. If you really want to be user friendly, wlroots could include some standard color profiles.

render/gles2/renderer.c
akvadrako: I don't understand you. If you want, you can completely ignore it - things will work as they do now. Otherwise, it's the backend solution which can be built on with a protocol, in which case the config will just provide defaults.

akvadrako:
This solves two problems for me:
- I want to view HDR content on an HDR monitor, while keeping correct colors in most apps.
- My monitor needs calibration to look correct.

This isn't meant to be a total solution to everybody's color needs, but it's designed so it can be built on - since it's just about assigning profiles to surfaces and outputs.

render/gles2/renderer.c
LaserEyess: What cyanreg is talking about is this https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14. There is a WIP upstream protocol for color management. Right now it doesn't have HDR but you could help with that. I think the benefit of standardizing this with a protocol is that you could start getting patches for client support. Wayland protocols also need an implementation before they're merged, so if you modified your patches to work with this protocol that would count.

LaserEyess:
@akvadrako Maybe I'm missing something but how does this work with 10bit? A lot of HDR content is 10bit and it doesn't look like you're modifying the renderer to work with 10bit color. There's a PR for that https://github.com/swaywm/wlroots/pull/2340 so I'm wondering if/how these two would work together.

render/gles2/renderer.c
akvadrako: Support for that protocol would better fit in a different PR, since the concerns are mostly orthogonal.
- This PR is assumes a `wlr_color_config` (being a profile or something else) is assigned to each `wlr_surface` or output (`wlr_renderer`) and applies them during rendering.
- The protocol is about reading and writing the color config assign to surfaces and outputs via clients, so it's the other side of the equation.

akvadrako:
> @akvadrako Maybe I'm missing something but how does this work with 10bit? A lot of HDR content is 10bit and it doesn't look like you're modifying the renderer to work with 10bit color. There's a PR for that #2340 so I'm wondering if/how these two would work together.

For good HDR support you need both this and the other PR. This doesn't affect anything with bit depth of clients or outputs, though it does the color correction in 16 bits. This one allows you to set your monitor to HDR while still allowing non-HDR clients to show non-HDR content correctly.

LaserEyess:
Ok that makes sense, but I believe some displays' HDR modes require 10bit depth; smart TVs, for example.

emersion:
I think it makes sense to have a per-monitor color profile, however I'm not sure about per-window color profiles. If this is just a temporary solution until a proper color management protocol is sorted out, I'd prefer to leave it out. 

>I want to view HDR content on an HDR monitor, while keeping correct colors in most apps.

But HDR metadata isn't sent to the monitor with this PR, right?

>I wasn't sure if we need to stick to ES 2.0. ES 3.0 is 8 years old now.

We can use ES 3.0 if available, but still need to support ES 2.0.

I wonder if this would be simpler to implement with e.g. a Vulkan renderer.

akvadrako:
> I think it makes sense to have a per-monitor color profile, however I'm not sure about per-window color profiles. If this is just a temporary solution until a proper color management protocol is sorted out, I'd prefer to leave it out.

I wouldn't say it's temporary; what do you think the alternative design would be? You need a profile to convert from; surfaces that do their own color management would have a `wlr_color_config` matching a monitor, apps by default would be sRGB (I assume) and HDR apps would be Rec2020 or something.

The color for each surface could be set with a protocol, but that doesn't change this PR. As for the sway PR, you could omit the "color profile" option and always assume sRGB, but it's less flexible. 

> But HDR metadata isn't sent to the monitor with this PR, right?

Right. I don't know anything about HDR metadata; I just set my monitor to HDR manually. It seems like with 10-bit output that happens automatically, but haven't tested that.

> > I wasn't sure if we need to stick to ES 2.0. ES 3.0 is 8 years old now.
> 
> We can use ES 3.0 if available, but still need to support ES 2.0.

Okay. Do we need to support color management with ES 2.0? Because 3D textures aren't available it will be a little less accurate.

progandy:
Rec2020 isn't good enough I think. I tried to understand the following discussion (which is ongoing for over a year), but only understood that color management is hard: https://discuss.pixls.us/t/wayland-color-management/10804 (in response to a proposal on the wayland-devel list https://lists.freedesktop.org/archives/wayland-devel/2019-January/039808.html )

The only reals consensus is that color management should be implemented correctly from the start so nobody starts relying on a broken implementation.

By the way, according to that discussion and the mailing list thread blending surfaces in the display colorspace is not a good idea at all (only blitting is OK). Blending should be done in a linear colorspace that won't lose color information.

akvadrako:
> Rec2020 isn't good enough

Rec2020 is just an example; surfaces can be in any colorspace. I agree the *protocol* should be designed carefully, because it's hard to change. I tried to stay away from relying on any details of that and just focus on the renderer.

> By the way, according to that discussion and the mailing list thread blending surfaces in the display colorspace is not a good idea at all (only blitting is OK). Blending should be done in a linear colorspace that won't lose color information.

That sounds reasonable and could be done, though it'll require two step rendering with an intermediate 16-bit texture. It may not always make sense if you care about performance over correctness.

emersion:
>I wouldn't say it's temporary; what do you think the alternative design would be? You need a profile to convert from; surfaces that do their own color management would have a wlr_color_config matching a monitor, apps by default would be sRGB (I assume) and HDR apps would be Rec2020 or something.

HDR apps would probably be scRGB FP16 or something (see https://www.x.org/wiki/Events/XDC2016/Program/xdc-2016-hdr.pdf). On the long run users won't have to configure per-app color profiles, everything will happen over the Wayland protocol.

>Okay. Do we need to support color management with ES 2.0? Because 3D textures aren't available it will be a little less accurate.

We don't need to. We could even require Vulkan for color management IMHO, if it allows us to have a less messy implementation.

>it'll require two step rendering with an intermediate 16-bit texture

This sounds like the direction Weston is going in.

akvadrako:
> HDR apps would probably be scRGB FP16 or something (see e.g. https://www.x.org/wiki/Events/XDC2016/Program/xdc-2016-hdr.pdf). On the long run users won't have to configure per-app color profiles, everything will happen over the Wayland protocol.

Nice presentation. And what about legacy applications that only provide sRGB? In scRGB, `#FFFFFF` means 80 nits, so beside a 1000 nit app on a 1000 nit monitor, it will appear very dim. By allowing an input profile you could say the app is in an sRGB-500nit space.

It could also be useful for non-sRGB legacy apps (like GIMP in X11) and special effects, like inverting colors.

> We don't need to. We could even require Vulkan for color management IMHO, if it allows us to have a less messy implementation.

The messiness is getting a couple functions from EGL and skipping them if they aren't available.

I personally would use the Vulkan renderer if it was ready - but I don't know how far off that is.

emersion:
>And what about legacy applications that only provide sRGB?

Isn't tone-mapping from sRGB to scRGB  supposed to fix this?

>special effects, like inverting colors

I'd prefer not to design the color management infrastructure for this kind of use-case.

>I personally would use the Vulkan renderer if it was ready - but I don't know how far off that is.

https://github.com/swaywm/wlroots/pull/2240 should help a lot. We also need the extension to import DMA-BUFs in Vulkan, which isn't implemented by any driver yet.

akvadrako:
> > And what about legacy applications that only provide sRGB?
> 
> Isn't tone-mapping from sRGB to scRGB supposed to fix this?

I'm not sure what you mean - are you suggesting an additional tone-mapping in sway/wlroots for non-color-managed SDR?

Probably you want the brightness (at least) of SDR surfaces to be configurable, like Windows does. One way of doing that in a flexible manner is allowing source ICC profiles.

aufkrawall:
Would this allow gamut mapping for every visible element? Vast majority of web and games content likely is BT.709 (more noteworthy exceptions should be HDR games and UHD BDs) and it's very unfortunate that in most graphical environments you end up with a patchwork rug of color-managed applications and such that are not. Thus imho it would be nice to have a global "fits most" setting (ideally it should also apply to fullscreen games) to allow proper sRGB calibration everywhere.

emersion:
Note, color management protocols are being discussed at https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/14

The MR contains docs.

emersion:
This is what a compositor needs to do to implement color management: https://gitlab.freedesktop.org/wayland/weston/-/issues/467

akvadrako:
Some comments on what isn't done from that list:

- [ ] Gamma-correct blending. (!596)
- [ ] Special cases:
  - [ ]  Parametric transformation?
- [ ]  Stop KMS planes from being used with client buffers when color management is on.
- [ ]  Compositor fallback rendering intent configuration.
   - Currently rendering intent is hardcoded.
- [ ]  Optionally use EDID colorimetry information to manufacture a monitor color profile for an output (not HDR).
- [ ]  Extract monitor HDR properties from EDID etc. (!583)
   - Are these just to build a monitor color profile?
- [ ]  Experimental option to set a monitor into HDR mode in DRM-backend.
   - I guess this is only for DP 1.4 and maybe HDMI 2.0?
- [ ]  HDR brightness and black level controls.
    - this can be done manually by adjusting the client color profiles.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2353

PR #2357, opened 2020-07-30
[DON'T MERGE] wlr-workspaces implementation

ammen99:
This is my work so far on implementing https://github.com/swaywm/wlr-protocols/pull/35 

I decided to first implement the protocol in its current form (with slight modifications as I saw fit), and I'd like feedback on them.
If it looks okay I will rename `wlr` to `ext` and open a PR against wayland-protocols as well.

The only notable change from the PR is that all synchronization is done at the workspace-manager level (both commit and done are moved there), although we could make them per workspace-group.

About the implementation, it should be very similar to foreign-toplevel-management, in almost all aspects. I am not 100% sure on the commit part (with pending, current & server_state).

If someone wants to test, I have added a wayfire plugin to support this protocol: https://github.com/WayfireWM/wayfire/pull/647

### DIFF
--- a/examples/wlr-workspace.c
+++ b/examples/wlr-workspace.c
@@ -31,15 +31,15 @@ struct workspace_state {
 
 static void copy_state(struct workspace_state *current,
 		struct workspace_state *pending) {
-	if (current->name && pending->name) {
-		free(current->name);
-	}
 
 	current->state = pending->state;
 	wl_array_copy(&current->coordinates, &pending->coordinates);
 
-	current->name = pending->name;
-	pending->name = NULL;
+	if (pending->name) {
+		free(current->name);
+		current->name = pending->name;
+		pending->name = NULL;
+	}
 }
 
 struct workspace_v1 {
@@ -309,8 +309,6 @@ int main(int argc, char **argv) {
 	}
 	wl_display_roundtrip(display); // load workspace groups
 	wl_display_roundtrip(display); // load details
-	wl_display_roundtrip(display); // load details
-	wl_display_roundtrip(display); // load details
 
 	if (focus_name != NULL) {
 		struct workspace_v1 *focus = workspace_by_name_or_bail(focus_name);
@@ -325,6 +323,7 @@ int main(int argc, char **argv) {
 			}
 		}
 		zwlr_workspace_handle_v1_activate(focus->handle);
+		zwlr_workspace_manager_v1_commit(workspace_manager);
 	}
 
 	wl_display_flush(display);
--- a/examples/wlr-workspace.c
+++ b/examples/wlr-workspace.c
@@ -1,0 +1,339 @@
+#define _POSIX_C_SOURCE 200809L
+#include <getopt.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wayland-client.h>
+#include "wlr-workspace-unstable-v1-client-protocol.h"
+
+#define WLR_WORKSPACE_VERSION 1
+
+/**
+ * Usage:
+ * 1. wlr-workspace
+ *    List all workspace groups and their workspaces
+ * 2. wlr-workspace -w X
+ *    Focus workspace with name X
+ * 3. wlr-workspace -m
+ *    Continuously monitor for changes and print new state.
+ */
+
+enum workspace_state_field {
+	WORKSPACE_FOCUSED = (1 << 0),
+};
+
+struct workspace_state {
+	char *name;
+	struct wl_array coordinates;
+	uint32_t state;
+};
+
+static void copy_state(struct workspace_state *current,
+		struct workspace_state *pending) {
+	if (current->name && pending->name) {
+		free(current->name);
+	}
+
+	current->state = pending->state;
+	wl_array_copy(&current->coordinates, &pending->coordinates);
+
+	current->name = pending->name;
+	pending->name = NULL;
+}
+
+struct workspace_v1 {
+	struct wl_list link;
+	struct zwlr_workspace_handle_v1 *handle;
+	struct workspace_state current, pending;
+};
+
+static void print_workspace(struct workspace_v1 *workspace) {
+	printf("--> workspace name=%s, focused=%d, coordinates=(",
+			workspace->current.name ?: "(n/a)",
+			!!(workspace->current.state & WORKSPACE_FOCUSED));
+
+	bool is_first = true;
+	int32_t *pos;
+	wl_array_for_each(pos, &workspace->current.coordinates) {
+		if (!is_first) {
+			printf(",");
+		}
+		printf("%d", *pos);
+		is_first = false;
+	}
+
+	printf(")\n");
+}
+
+static uint32_t array_to_state(struct wl_array *array) {
+	uint32_t state = 0;
+	uint32_t *entry;
+	wl_array_for_each(entry, array) {
+		if (*entry == ZWLR_WORKSPACE_HANDLE_V1_STATE_ACTIVE)
+			state |= WORKSPACE_FOCUSED;
+	}
+
+	return state;
+}
+
+
+static void workspace_handle_name(void *data,
+		struct zwlr_workspace_handle_v1 *workspace_handle_v1,
+		const char *name) {
+	struct workspace_v1 *workspace = (struct workspace_v1*)
+		zwlr_workspace_handle_v1_get_user_data(workspace_handle_v1);
+
+	free(workspace->pending.name);
+	workspace->pending.name = strdup(name);
+}
+
+static void workspace_handle_coordinates(void *data,
+		struct zwlr_workspace_handle_v1 *workspace_handle,
+		struct wl_array *coordinates) {
+	struct workspace_v1 *workspace = (struct workspace_v1*)
+		zwlr_workspace_handle_v1_get_user_data(workspace_handle);
+	wl_array_copy(&workspace->pending.coordinates, coordinates);
+}
+
+static void workspace_handle_state(void *data,
+		struct zwlr_workspace_handle_v1 *workspace_handle,
+		struct wl_array *state) {
+	struct workspace_v1 *workspace = (struct workspace_v1*)
+		zwlr_workspace_handle_v1_get_user_data(workspace_handle);
+	workspace->pending.state = array_to_state(state);
+}
+
+static void workspace_handle_remove(void *data,
+		struct zwlr_workspace_handle_v1 *workspace_handle) {
+	struct workspace_v1 *workspace = (struct workspace_v1*)
+		zwlr_workspace_handle_v1_get_user_data(workspace_handle);
+	zwlr_workspace_handle_v1_destroy(workspace_handle);
+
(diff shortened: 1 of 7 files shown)


examples/wlr-workspace.c
```diff
@@ -0,0 +1,338 @@
...
+	wl_registry_add_listener(registry, &registry_listener, NULL);
+	wl_display_dispatch(display);
+	wl_display_roundtrip(display);
+
+	if (workspace_manager == NULL) {
+		fprintf(stderr, "wlr-workspace not available\n");
+		return EXIT_FAILURE;
+	}
+	wl_display_roundtrip(display); // load workspace groups
+	wl_display_roundtrip(display); // load details
```
Anakael, line 311: Why it called twice after 304th line?
As I see here https://wayland.freedesktop.org/docs/html/apb.html
it blocks for all requests

include/wlr/types/wlr_workspace_v1.h
```diff
@@ -0,0 +1,122 @@
...
+};
+
+struct wlr_workspace_handle_v1 {
+	struct wl_list link; // wlr_workspace_group_handle_v1::workspaces
+	struct wl_list resources;
+
+	struct wlr_workspace_group_handle_v1 *group;
+
+	// request from the client
+	uint32_t pending, current;
```
Anakael, line 69: `pending_state`, `current_state` better, I think

types/wlr_workspace_v1.c
```diff
@@ -0,0 +1,544 @@
...
+	}
+
+	workspace->pending &= ~WLR_WORKSPACE_HANDLE_V1_STATE_ACTIVE;
+}
+
+static void workspace_handle_resource_destroy(struct wl_resource *resource) {
+	wl_list_remove(wl_resource_get_link(resource));
+}
+
+static bool fill_array_from_workspace_state(struct wl_array *array,
```
Anakael, line 73: What does this function do?

types/wlr_workspace_v1.c
```diff
@@ -0,0 +1,544 @@
...
+		if (index == NULL) {
+			return false;
+		}
+		*index = ZWLR_WORKSPACE_HANDLE_V1_STATE_ACTIVE;
+	}
+
+	return true;
+}
+
+static void workspace_handle_send_state_to_resource(
```
Anakael, line 86: I think state not clear here. Something like `data` or `details` link in foreign would be better

include/wlr/types/wlr_workspace_v1.h
```diff
@@ -0,0 +1,122 @@
...
+
+#include <wayland-server-core.h>
+#include <wlr/types/wlr_output.h>
+
+struct wlr_workspace_manager_v1 {
+	struct wl_event_loop *event_loop;
+	struct wl_event_source *idle_source;
+
+	struct wl_global *global;
+	struct wl_list resources; // wl_resource_get_link
```
Anakael, line 20: May you please explain what is purpose of resources?

types/wlr_workspace_v1.c
```diff
@@ -0,0 +1,544 @@
...
+
+static void group_send_output(struct wlr_workspace_group_handle_v1 *group,
+		struct wlr_output *output, bool enter) {
+	struct wl_resource *resource, *tmp;
+	wl_resource_for_each_safe(resource, tmp, &group->resources) {
+		send_output_to_group_resource(resource, output, enter);
+	}
+}
+
+static void toplevel_handle_output_destroy(struct wl_listener *listener,
```
Anakael, line 307: Fix name?

types/wlr_workspace_v1.c
```diff
@@ -0,0 +1,544 @@
...
+		group_output_destroy(output);
+	}
+
+	struct wl_resource *tmp3, *resource;
+	wl_resource_for_each_safe(resource, tmp3, &group->resources) {
+		zwlr_workspace_group_handle_v1_send_remove(resource);
+
+		wl_resource_set_user_data(resource, NULL);
+		wl_list_remove(&resource->link);
+		wl_list_init(&resource->link);
```
Anakael, line 426: Why init here?

types/wlr_workspace_v1.c
```diff
@@ -0,0 +1,544 @@
...
+		zwlr_workspace_handle_v1_send_coordinates(resource, coordinates);
+	}
+
+	workspace_manager_update_idle_source(workspace->group->manager);
+}
+
+void wlr_workspace_handle_v1_set_active(
+		struct wlr_workspace_handle_v1 *workspace, bool activate) {
+	if (activate) {
+		workspace->server_state |= WLR_WORKSPACE_HANDLE_V1_STATE_ACTIVE;
```
Anakael, line 142: I think we should change pending here, not server_state

ammen99: This is the function that the compositor calls, not the client. The client requests are `workspace_handle_activate/deactivate`

examples/wlr-workspace.c
ammen99: The comments say it. On the firsst roundtrip, we get all events for workspaces. We bind all workspaces during this first roundtrip. We also connect our listeners. Then we need a second roundtrip during which we get all details for the workspaces.

include/wlr/types/wlr_workspace_v1.h
ammen99: A resource is typically a connected client, we just need to keep a list of connected clients so we can send them events.

types/wlr_workspace_v1.c
ammen99: Because the resource will be destroyed later here: https://github.com/swaywm/wlroots/pull/2357/files#diff-4adb740cfee4446f54fc395585e39a8fR247

We want that call there to be a valid list, otherwise, we may get a crash.

examples/wlr-workspace.c
Anakael: I see.

include/wlr/types/wlr_workspace_v1.h
Anakael: Ok, ty.

types/wlr_workspace_v1.c
Anakael: But `server_state` not used outside here.

types/wlr_workspace_v1.c
Anakael: I see.

types/wlr_workspace_v1.c
ammen99: The client may request a workspace to be focused. But the compositor may decide to ignore this request. Thus, there is one state which the client requested, and another state, that the compositor has set.

types/wlr_workspace_v1.c
ammen99: Exactly what it says. It takes a state which is a bitmask of the workspace state, and fills an array which contains the set bits (array is needed because that's what the protocol has).

types/wlr_workspace_v1.c
ammen99: done

types/wlr_workspace_v1.c
ammen99: done

types/wlr_workspace_v1.c
Anakael: Hmm. 
Commit method just change pending state to current? And clients change pending state directly?
I think we should prevent changing in this case via pending and current.
Because sync 3 state it's not obvious and easy.

types/wlr_workspace_v1.c
Anakael: I see. thnaks

types/wlr_workspace_v1.c
Anakael: Client call request to change state
Compositor recive request and decide to change pending state or not.
Don't see any reasons for server state

ammen99: In the future we may have more states than just active/inactive so the `server_state` field will have more information that just active/inactive. In such cases, we will need to store them separately from pending/current (but yes, right now we could remove them. I decided to try to make the implementation a little bit future-proof because I expect a few more states like urgent).

Anakael: I expect too, but can't imagine the case when we want to separate server_state from current and pending state.
If compositor refuse to update state after request - it can just don't update pending.
If required some logic for complex change state by one request - it also can be performed via pending_state.

ammen99: I think you are misunderstanding what current and pending is. Pending is what the client has requested, but not yet committed. Once the client commits, that state becomes current - currently requested. So you can think of the `pending` field as a private implementation detail, compositors should not use it. They should use current only, to see what the client has requested.

And in server_state we save what the compositor has last requested too.

Anakael: -> And in server_state we save what the compositor has last requested too.
compositor requested from who?

Anakael: Why compositor not should use pending?
I see it like:
Client, Server:
- C: I want A
- S: -
- C: I want B
- S; -
- C: commit
- S: No, i don't give you A, but I'll give you B and C. *transorm pending to current and sends such events*

server_state seems like exceed duplicate of pending.

ammen99: I think we clarified why this is needed on IRC.

Anakael: Yes

ammen99:
wayland-protocols MR: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40

protocol/wlr-workspace-unstable-v1.xml
```diff
@@ -0,0 +1,270 @@
...
+      </description>
+      <arg name="state" type="array"/>
+    </event>
+
+    <enum name="state">
+      <description summary="types of states on the workspace">
+        The different states that a workspace can have.
+      </description>
+
+      <entry name="active" value="0" summary="the workspace is active"/>
```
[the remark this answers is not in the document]
ammen99: In the wayland-protocols PR here: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40
I have added urgent and hidden. I'll also update this PR once it is approved upstream.

examples/wlr-workspace.c
```diff
@@ -0,0 +1,338 @@
...
+				zwlr_workspace_handle_v1_deactivate(workspace->handle);
+			}
+		}
+		zwlr_workspace_handle_v1_activate(focus->handle);
+		zwlr_workspace_manager_v1_commit(workspace_manager);
+	}
+
+	wl_display_flush(display);
+
+	if (monitor != false) {
```
MisterDA, line 331: `if (monitor)` is better

ammen99:
I have now updated the implementation to match the latest changes in the PR (so this is ready for review)

Anakael:
Any updates here?

ammen99:
> Any updates here?

PRs are waiting for a review.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2357

Issue #2330, opened 2020-07-14
Compositors may need to send no-op configure events


emersion:
When the client sends a `set_maximized` or `set_fullscreen` request (or one of the `unset` variants), the protocol states that the compositor must send a `configure` event, even if it's a no-op because the compositor refuses.

We should at least document this properly, because this can be a little surprising and cause frozen clients.

Ideally wlroots should also help with this. However wlroots can't schedule a configure event right away, because this breaks asynchronous configure events (in case the compositor defers the decision).

See https://github.com/swaywm/sway/pull/5544

RPigott:
What if this was handled in `wlr_xdg_toplevel_set_maximized`, for example? That is a function the compositor would already call to indicate it has made a decision, but the issue is wlroots does _extra_ work in this case to compare the states when that's really unnecessary, since it has to schedule a configure anyway. So instead of

https://github.com/swaywm/wlroots/blob/aaf490d794bd077330c306face305f4a87de5809/types/xdg_shell/wlr_xdg_toplevel.c#L584

We would use `wlr_xdg_surface_schedule_configure` to skip the comparison there for those requests in which scheduling a configure is required regardless.

The difference is that for compositors that don't care about maximizing like sway, instead of not implementing set_maximized they'd be required to implement a stub handler that just returns false, e.g. `wlr_xdg_toplevel_set_maximized(surface, false)` to let wlroots know it's done considering the client's request. Compositors that don't do that would get the same old buggy behavior, so we can document that requirement somewhere. As long as compositors implement handlers like they normal, they will get the correct Wayland behavior because wlroots will know when a configure is required or not.

Does that sounds good? I can draft a pull for wlroots/sway if does.

emersion:
So far we've allowed compositors to repeatedly call e.g. `wlr_xdg_toplevel_set_maximized` without caring about the current state. This has allowed compositors to blindly call it even if the toplevel had the right state. Your change would break that (not saying it's necessarily a bad thing).

Note that the guarantee in the protocol to always send a `configure` event after a `set_maximized` request is racy anyways:

- Compositor sends `configure` on its own
- Client sends `set_maximized`
- Compositor sends `configure` in response to the request
- Client receives the first `configure` event, assumes it's a response to its `set_maximized` request

The client has no way to map `configure` events to `set_*` requests (if any).

RPigott:
[...]

PR #2361, opened 2020-08-01
Always schedule configure on maximize and fullscreen requests

RPigott:
This is my approach for fixing #2330. The idea is that compositors already implement handlers that call these functions (probably), so wlroots should do the "right thing" and always schedule a configure event even if the compositor decided not to change the window state.

If a compositor like sway doesn't care about maximize requests, it still needs to implement a stub handler that calls set_maximized to indicate it is done considering the request.

### DIFF
--- a/types/xdg_shell/wlr_xdg_toplevel.c
+++ b/types/xdg_shell/wlr_xdg_toplevel.c
@@ -581,7 +581,7 @@ uint32_t wlr_xdg_toplevel_set_maximized(struct wlr_xdg_surface *surface,
 	assert(surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL);
 	surface->toplevel->server_pending.maximized = maximized;
 
-	return schedule_xdg_surface_configure(surface);
+	return wlr_xdg_surface_schedule_configure(surface);
 }
 
 uint32_t wlr_xdg_toplevel_set_fullscreen(struct wlr_xdg_surface *surface,
@@ -589,7 +589,7 @@ uint32_t wlr_xdg_toplevel_set_fullscreen(struct wlr_xdg_surface *surface,
 	assert(surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL);
 	surface->toplevel->server_pending.fullscreen = fullscreen;
 
-	return schedule_xdg_surface_configure(surface);
+	return wlr_xdg_surface_schedule_configure(surface);
 }
 
 uint32_t wlr_xdg_toplevel_set_resizing(struct wlr_xdg_surface *surface,


types/xdg_shell/wlr_xdg_toplevel.c
```diff
@@ -581,15 +581,15 @@ uint32_t wlr_xdg_toplevel_set_maximized(struct wlr_xdg_surface *surface,
 	assert(surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL);
 	surface->toplevel->server_pending.maximized = maximized;
 
-	return schedule_xdg_surface_configure(surface);
+	return wlr_xdg_surface_schedule_configure(surface);
 }
 
 uint32_t wlr_xdg_toplevel_set_fullscreen(struct wlr_xdg_surface *surface,
```
ammen99, line 587: This function is what **the compositor** calls when it wants to fullscreen a client, not what the client calls.

RPigott: That's the point, isn't it?

I'm not trying to make this work with _no_ intervention from the compositor, I want wlroots to send a configure event when the compositor _does_ call this function to set the client fullscreen state, whether or not the client has actually changed state.

The issue before wasn't just that sway ignored the maximize request, it was that there was no way to implement the handler using the usual api (this function) that gave the correct Wayland behavior (send a configure regardless of state change). Now compositors using the usual api can get the correct behavior, and the compositor ignoring the request is never correct.

ammen99: Oh, I see, my bad.

RPigott:
Actually, I didn't consider that these functions may be called outside of responding to a client request as well. Like if the compositor decides on it's own it wants to fullscreen a client. In those cases the compositor will call these functions and the state comparison will be skipped even though it could have been applied without breaking anything.

@ammen99 Do you think that might be an issue?

ammen99:
> Actually, I didn't consider that these functions may be called outside of responding to a client request as well. Like if the compositor decides on it's own it wants to fullscreen a client. In those cases the compositor will call these functions and the state comparison will be skipped even though it could have been applied without breaking anything.
> 
> @ammen99 Do you think that might be an issue?

I don't think it will be an issue, but I am no expert in the xdg-shell protocol.

ammen99:
I think this is not a good way to solve the issue. There are two possible solutions:

1. We can schedule configures unconditionally after this line: https://github.com/swaywm/wlroots/blob/master/types/xdg_shell/wlr_xdg_toplevel.c#L381 (and in similar locations for unset_maximize and fullscreen). The compositor should have already updated the state in the event handler. This will also not require anything special in the compositor. However, it makes it impossible for the compositor to postpone the configure event.

2. Make the compositor itself call `wlr_xdg_surface_schedule_configure()`, maybe we could document something. This would require changes in all compositors to properly fix the issue.

RPigott:
Of those choices I definitely prefer (2). I don't want to restrict the compositor if we don't have to, and ultimately the required change is not large. So for that reason I don't like (1).

The thing is, it would be totally pointless to schedule a configure in wlroots when the state changes _and_ require the compositor to schedule a configure on it's own, so if we do (2) the state comparison should be removed in wlroots and the wlr_xdg_toplevel_set_maximized and friends will never schedule a configure on their own.

But then, if the compositor is calling wlr_xdg_toplevel_set_maximized and friends in their handler, they have already completed their decision on how to handle the request and have nothing left to do but schedule a configure. So... we could fold the schedule_configure call back into wlr_xdg_toplevel_set_maximized function and friends and their semantics once again falls back in line with the remaining wlr_xdg_toplevel_set_* functions.

That is how I arrived at this proposal, (3). Here's why I think think (3) > (2) > (1).

The drawback of (1) is the worst imo and should be avoided. It removes the ability of a wlroots compositor to ever handle the set_maximized requests in an asynchronous way.

The drawback of (2) is that the wlroots toplevel_set_maximized and friends functions have different semantics compared to their siblings that must be documented and then accounted for in compositors. A bother maybe, but not terrible.

The drawback of (3) is that now that these special toplevel_set_* functions always schedule a configure, It's possible the compositor could accidentally schedule an extra configure if and only if it calls them from _outside_ a handler in a way that doesn't update the client state, e.g. spuriously calling toplevel_set_maximized on an already maximized client.

Given that I think that behavior is both rare (this currently happens _nowhere_ in sway's code, at least. `container_set_fullscreen` already contains an "if fullscreen" early return) and harmless when it does happen, it seems worth it for the simplicity in both wlroots and compositors. The fact that compositors which implement any handler whatsoever for these xdg_toplevel requests will also be immediately fixed with no additional change is a bonus.

ammen99:
> The drawback of (2) is that the wlroots toplevel_set_maximized and friends functions have different semantics compared to their siblings that must be documented and then accounted for in compositors. A bother maybe, but not terrible.

This must be considered anyway, your proposed change also requires the compositor to always call `set_maximized` even if it does not wish to change the maximized state.

ammen99:
> The fact that compositors which implement any handler whatsoever for these xdg_toplevel requests will also be immediately fixed with no additional change is a bonus.

This is a very specific assumption about compositors that have such handlers :) For one thing, this PR will not fix the problem for Wayfire, it calls the `wlr_xdg_toplevel_set_maximized` function only when the state should actually change.

RPigott:
> your proposed change also requires the compositor to always call set_maximized even if it does not wish to change the maximized state.

That's true.

> For one thing, this PR will not fix the problem for Wayfire, it calls the wlr_xdg_toplevel_set_maximized function only when the state should actually change.

There's never any reason to do that though. In the status quo, all the wlr_xdg_toplevel_set_* functions already only send a configure when the state changes. The trouble here is that in these few cases it's wrong to do that according to the Wayland protocol. Doing the comparison twice doesn't make it right – I would consider that a bug in Wayfire. Regardless of the decision here, that code should be removed.

If we choose (3) or (2) you will either need to add `wlr_xdg_toplevel_set_maximized` to your handler or `wlr_xdg_surface_schedule_configure` respectively to correct the behavior and wlroots will have documentation to reflect that. If we move the responsibility of scheduling a configure to the compositor in this case, option (2), compositors will be broken until they get an update because they will no longer schedule a configure for even the permitted state changes. With (3) compositor behavior is no worse than before the wlroots change.

ammen99:
I don't see why compositors are broken with (2). I think you may have misunderstood me. (2) is basically saying "let's leave things as they are in master, and fix the bug in compositors, not in wlroots"

RPigott:
So we leave the original schedule_configure in? If `wlr_xdg_toplevel_set_maximized` will schedule a configure on a change, and then compositors will schedule a configure again, won't that potentially cause compositors to send two configure events if they handle the request asynchronously?

Only if they leave the handler before calling schedule_configure on their own I guess, which doesn't seem likely, but it still feels kinda weird. Even if I'm not swayed, I don't feel so strongly about it to keep up the discussion. Either (2) or (3) is probably fine in my book.

I'll defer and let you or emersion decide how to handle this. I don't have a compositor of my own yet.

ammen99:
> So we leave the original schedule_configure in? If `wlr_xdg_toplevel_set_maximized` will schedule a configure on a change, and then compositors will schedule a configure again, won't that potentially cause compositors to send two configure events if they handle the request asynchronously?

Yes, I would leave the original configure. But why would it send two configure events? If the compositor decides to postpone configuring the client, nothing will call schedule_configure so you get only one event when the compositor decides to configure.

RPigott:
> If the compositor decides to postpone configuring the client, nothing will call schedule_configure so you get only one event when the compositor decides to configure.

I think it could only happen if:

1. compositor calls `wlr_xdg_toplevel_set_*`, the state has changed and wlroots schedules a configure
2. compositor does some more async stuff and wlroots sends the configure
3. compositor calls `wlr_xdg_surface_schedule_configure` as required.

If wlroots doesn't send a configure in between, the check in `schedule_configure` will prevent sending two configures. That would be pretty unusual behavior from a compositor though, I think. They should just call `wlr_xdg_toplevel_set_*` and `wlr_xdg_surface_schedule_configure` together.

RPigott:
This is no longer necessary.

PR #2363, opened 2020-08-05
gamma-control-v1: fix use-after-free in gamma_control_handle_set_gamma

emersion:
gamma_control_send_failed destroys gamma_control.


ammen99:
Let's wait for @minus7 to test this

minus7:
Doesn't crash anymore (though I'm not sure what exactly caused it before), but gamma isn't re-applied after the monitors wake up from DPMS (have to restart Redshift)

emersion:
Hmm, makes sense. This will require more work. Can you open an issue about it, @minus7?

PR #2366, opened 2020-08-10
xdg_shell: simplify wlr_xdg_toplevel_state

ifreund:
the fullscreen_output and fullscreen_output_destroy fields were only
used if the state was wlr_xdg_toplevel::client_pending. Moving these
fields to wlr_xdg_toplevel reduces confusion and simplifies the code.


ifreund:
Guess this is a breaking change and sway does access one of these fields directly here: https://github.com/swaywm/sway/blob/4dd46f06acc520449b980a5ea52be544cc5bfb6d/sway/desktop/xdg_shell.c#L449-L452

ifreund:
Super stale and superceeded by #3199

PR #2374, opened 2020-08-22
Renderer v6: headless backend

emersion:
Same as https://github.com/swaywm/wlroots/pull/2240 but for the headless backend. Supersedes https://github.com/swaywm/wlroots/pull/2268.

### DIFF
--- a/backend/headless/output.c
+++ b/backend/headless/output.c
@@ -1,65 +1,18 @@
 #include <assert.h>
-#include <GLES2/gl2.h>
-#include <GLES2/gl2ext.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <wlr/interfaces/wlr_output.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/util/log.h>
 #include "backend/headless.h"
+#include "render/swapchain.h"
+#include "render/wlr_renderer.h"
 #include "util/signal.h"
 
 static struct wlr_headless_output *headless_output_from_output(
 		struct wlr_output *wlr_output) {
 	assert(wlr_output_is_headless(wlr_output));
 	return (struct wlr_headless_output *)wlr_output;
-}
-
-static bool create_fbo(struct wlr_headless_output *output,
-		unsigned int width, unsigned int height) {
-	if (!wlr_egl_make_current(output->backend->egl, EGL_NO_SURFACE, NULL)) {
-		return false;
-	}
-
-	GLuint rbo;
-	glGenRenderbuffers(1, &rbo);
-	glBindRenderbuffer(GL_RENDERBUFFER, rbo);
-	glRenderbufferStorage(GL_RENDERBUFFER, output->backend->internal_format,
-		width, height);
-	glBindRenderbuffer(GL_RENDERBUFFER, 0);
-
-	GLuint fbo;
-	glGenFramebuffers(1, &fbo);
-	glBindFramebuffer(GL_FRAMEBUFFER, fbo);
-	glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
-		GL_RENDERBUFFER, rbo);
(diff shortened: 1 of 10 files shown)


any1:
Works fine with wayvnc. Any plans on implementing export-dmabuf?

backend/headless/output.c
```diff
@@ -120,6 +120,23 @@ static void output_rollback_render(struct wlr_output *wlr_output) {
...
+		struct wlr_dmabuf_attributes *attribs) {
+	struct wlr_headless_output *output =
+		headless_output_from_output(wlr_output);
+
+	if (!output->front_buffer) {
+		return false;
+	}
+
+	struct wlr_dmabuf_attributes tmp;
+	if (!wlr_buffer_get_dmabuf(output->front_buffer, &tmp)) {
```
any1, line 133: Should this not be `back_buffer`? This works well with wayvnc if the back-buffer is used.

emersion: The DRM backend [exports the front-buffer too](https://github.com/swaywm/wlroots/blob/297354f8477285231826f6441b943188ec7d7399/backend/drm/drm.c#L642), so I just went with that. But the [wlr_output docs](https://github.com/swaywm/wlroots/blob/297354f8477285231826f6441b943188ec7d7399/include/wlr/types/wlr_output.h#L389) seem to not agree.

any1: wlr-screencopy expects the back-buffer and wlr-export-dmabuf seems to also expect it. What should be the source of truth here?

emersion: That's a good question.

I think the original behaviour was to hand over the back buffer. Then this commit changed it to use the front-buffer: https://github.com/swaywm/wlroots/commit/8da9d9679e4cad640b0f0a518ca1529c8cdecbe0#diff-8b05a774317ee8e87d51422170f82d4bL663

However, I think moving forward it'd make more sense to hand over the front buffer. Acquiring the back-buffer has a number of limitations: there's no guarantee we even have a back buffer, except right before a commit. I'd like to move away from implicit state like this.

This would involve updating the docs + screencopy/export-dmabuf to export on `commit` instead of `precommit`.

any1: Can you make `glReadPixels` read the front buffer without implementing this PR for all backends?

emersion: Unfortunately, it's not possible for GLES2 to select the front buffer. `glReadBuffer(GL_FRONT)` is only available on desktop GL.

any1: There are two options, right?
 1. Have drm read from the back-buffer and keep this PR as is, and slowly work towards the goal of implementing this for all the backends, one backend at a time. Then, eventually, switch to reading from and exporting the front buffer.
 2. Implement this PR for all backends and change screencopy and export-dmabuf to use the front-buffer in one big PR.

emersion: Submitted https://github.com/swaywm/wlroots/pull/2448 to do (1).

emersion: >Submitted #2448 to do (1).

This PR is toast: we don't have access to the back-buffer prior to the output commit: https://github.com/swaywm/wlroots/blob/79c7591dc196b14dfa857caf2c183276f937ee02/backend/drm/renderer.c#L282

>2. Implement this PR for all backends and change screencopy and export-dmabuf to use the front-buffer in one big PR.

Actually, we don't need to read from the front-buffer when using the `glReadPixels` code-path. We can continue reading from the back-buffer for shm, *and* read from the front-buffer for DMA-BUF. This will probably create friction for screencopy which supports both, but I'll see if it's manageable.

any1: Whoops. :)

I suspect that the damage tracking is going to be the hardest part.

emersion: >We can continue reading from the back-buffer for shm, and read from the front-buffer for DMA-BUF.

I've done that here: https://github.com/swaywm/wlroots/pull/2463

Issue #2376, opened 2020-08-25
Direct session crashes on FreeBSD


jbeich:
Regressed by #2364. direct.c vs. direct-freebsd.c grew out of sync?

```
$ sway -dc /dev/null
00:00:00.000 [sway/main.c:327] Sway version 1.5-fd216b3a (Aug 25 2020, branch 'master')
00:00:00.000 [sway/main.c:154] FreeBSD 13.0-CURRENT #0 r364769M: Tue Aug 25 14:22:19 UTC 2020 foo@bar:/usr/obj/usr/src/amd64.amd64/sys/MYKERNEL amd64
00:00:00.000 [sway/main.c:170] Contents of /etc/os-release:
00:00:00.000 [sway/main.c:154] NAME=FreeBSD
00:00:00.000 [sway/main.c:154] VERSION=13.0-CURRENT
00:00:00.000 [sway/main.c:154] VERSION_ID=13.0
00:00:00.000 [sway/main.c:154] ID=freebsd
00:00:00.000 [sway/main.c:154] ANSI_COLOR="0;31"
00:00:00.000 [sway/main.c:154] PRETTY_NAME="FreeBSD 13.0-CURRENT"
00:00:00.000 [sway/main.c:154] CPE_NAME=cpe:/o:freebsd:freebsd:13.0
00:00:00.000 [sway/main.c:154] HOME_URL=https://FreeBSD.org/
00:00:00.000 [sway/main.c:154] BUG_REPORT_URL=https://bugs.FreeBSD.org/
00:00:00.000 [sway/main.c:142] LD_LIBRARY_PATH=
00:00:00.000 [sway/main.c:142] LD_PRELOAD=
00:00:00.000 [sway/main.c:142] PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin
00:00:00.000 [sway/main.c:142] SWAYSOCK=
00:00:00.000 [sway/server.c:43] Preparing Wayland server initialization
00:00:00.001 [INFO] [backend/session/direct-freebsd.c:207] Using tty /dev/ttyv3
00:00:00.010 [INFO] [backend/session/direct-freebsd.c:296] Successfully loaded direct session
Segmentation fault

(lldb) bt all
* thread #1, name = 'sway', stop reason = signal SIGSEGV
  * frame #0: 0x0000000800904b07 libwayland-server.so.0`wl_list_insert + 7
[...]

PR #2379, opened 2020-08-26
session: Add missing init to direct-freebsd

kennylevinsen:
bad1e9afa8ea ("session: Add libseat backend") introduced a change to to
how session backends initialize, but failed to update the FreeBSD
specific version of the direct backend accordingly.

Closes: https://github.com/swaywm/wlroots/issues/2376


emersion (approved):
Thanks!

PR #2382, opened 2020-08-27
render: Don't crash on 0 dimensions

guidocella:
Don't force compositors to check when an empty shape is being renderered.
References #2282. This was motivated by dwl crashing when setting window
borders to 0 (djpohly/dwl#51).


emersion:
Can we early-return if either width or height is zero?

guidocella:
I changed the code to return early (that's actually how I did it at first).

emersion:
Thanks, this LGTM. Can you squash these commits into a single one?

guidocella:
Done.

emersion (approved):
Thanks!

PR #2384, opened 2020-08-30
Maintain correct cursor size on displays with scale factors

quantum5:
When using a scale factor on a display, the cursor size was incorrectly handled. With multiple displays and differing scale factors on each display, this caused the cursor size to differ visibly between displays.

For example, I have two 27" monitors side by side. Monitor 1 runs at 2560×1440 with scale 1 and monitor 2 runs at 3840×2160 with scale 1.5. Everything should have the same apparent size on this setup. On monitor 1, everything looks good. On monitor 2:
* in any UI rendered by sway, the cursor appears to be half size.
* in native wayland applications, the cursor appears to be be quarter size.
* on xwayland applications, the cursor appears to be the correct size.

It appears that `cursor->output->impl->set_cursor` [considers the output scale but also divides the dimensions of the cursor by `scale`](https://github.com/swaywm/wlroots/blob/b0144c7ded2b655085eb8e7e9b64908dc9420d60/backend/drm/drm.c#L922-L923). Indeed, everywhere else this function is called in `types/wlr_output.c`, `1` is always passed as `scale`.

Passing `1` in `output_cursor_attempt_hardware` causes the cursor size to be correct in all cases.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -1061,8 +1061,8 @@ static void output_cursor_commit(struct wlr_output_cursor *cursor,
 
 	// Some clients commit a cursor surface with a NULL buffer to hide it.
 	cursor->enabled = wlr_surface_has_buffer(surface);
-	cursor->width = surface->current.width * cursor->output->scale;
-	cursor->height = surface->current.height * cursor->output->scale;
+	cursor->width = surface->current.width * surface->current.scale * cursor->output->scale;
+	cursor->height = surface->current.height * surface->current.scale * cursor->output->scale;
 	output_cursor_update_visible(cursor);
 	if (update_hotspot) {
 		cursor->hotspot_x -= surface->current.dx * cursor->output->scale;
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -981,12 +981,10 @@ static void output_cursor_update_visible(struct wlr_output_cursor *cursor) {
 }
 
 static bool output_cursor_attempt_hardware(struct wlr_output_cursor *cursor) {
-	float scale = cursor->output->scale;
 	enum wl_output_transform transform = WL_OUTPUT_TRANSFORM_NORMAL;
 	struct wlr_texture *texture = cursor->texture;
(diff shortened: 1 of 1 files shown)


quantum5:
Turns out software cursors were also broken, but in a different way. There are two cases to consider:

* In `wlr_output_cursor_set_image`, which appears to be used for cursors rendered by sway's own UI. Multiplying the logical cursor size by the output scale factor fixes the software cursor size issues on sway's UI.
* In `output_cursor_commit`, which appears to be used for cursors submitted by wayland clients. GTK, for example, [divides the cursor size by the current scale, then keeps track of the scale](https://github.com/GNOME/gtk/blob/ec9fd76c8700b52011c53e31987691a07f406b19/gdk/wayland/gdkcursor-wayland.c#L195-L197). This information was tracked in `wlr_surface_state::scale`, but was not used. Multiplying by `wlr_surface_state::scale` fixes the software cursor size issues in these wayland clients.

quantum5:
I have been running this locally for a while now and it seems to work perfectly. This should be ready for review.

mozartilize:
Hi, I only have one display and scaled with 1.12. I tested your branch and my cursor on gtk apps goes big :)

quantum5:
@mozartilize I believe sway controls the titles and background, while wayland clients like gtk apps set their own cursors.

The problem is that on the current master, the cursors over gtk apps are a factor of two smaller than they are supposed to be when using a scale factor > 1 (due to the missing multiply by cursor surface scale factor). If you scaled up cursors in gtk apps so that they appear normal on current master, they will appear too big on my branch.

If this is the case, then changing the scale factor to 1 while on wlroots master should also result in cursors appearing too large over gtk applications.

quantum5:
It appears that this patch only works in cases where the minimum cursor size in the active cursor theme is larger than the cursor size setting. In all other cases it would create cause the cursor to appear overly large Instead, increasing the cursor setting to the minimum cursor size in the active cursor theme would fix the sizing issues.

PR #2385, opened 2020-08-30
Fix undefined behavior in xcursor

e00E:
Copied from the two commit messages:

Use fixed size integer type
This type is meant to be 4 bytes large as seen in _XcursorReadUInt which
always reads 4 bytes. An unsigned int is often 4 bytes large but this
isnt' guaranteed so it is cleaner to use the exact type we want.

Fix undefined behavior
Without the casts the bytes accesses get converted to int. but int is
not guaranteed to be 4 bytes large. Even when it is 4 bytes large
`bytes[3] << 24` does not fit because int is signed.

### DIFF
--- a/include/xcursor/xcursor.h
+++ b/include/xcursor/xcursor.h
@@ -26,8 +26,10 @@
 #ifndef XCURSOR_H
 #define XCURSOR_H
 
-typedef int		XcursorBool;
-typedef unsigned int	XcursorUInt;
+#include <stdint.h>
+
+typedef int XcursorBool;
+typedef uint32_t XcursorUInt;
 
 typedef XcursorUInt	XcursorDim;
 typedef XcursorUInt	XcursorPixel;
--- a/xcursor/xcursor.c
+++ b/xcursor/xcursor.c
@@ -285,11 +285,12 @@ _XcursorReadUInt (XcursorFile *file, XcursorUInt *u)
         return XcursorFalse;
 
     if ((*file->read) (file, bytes, 4) != 4)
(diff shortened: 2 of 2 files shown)


emersion (commented):
Can you change the commit author to match the original commit's? You can add yourself in a Co-authored-by line.

e00E:
Which commit do you mean with "the original"? I wrote both of the commits in this PR.

emersion:
Oh, my bad, I thought this was backported from xorg-xserver.

emersion:
It would be nice to submit this upstream: https://gitlab.freedesktop.org/xorg/lib/libxcursor

e00E:
I didn't realize there was an upstream. Maybe we should put that in a readme somewhere (or maybe it is and I missed it). I am going to try to submit the patch upstream, too.

emersion:
Thanks. Upstream merge request: https://gitlab.freedesktop.org/xorg/lib/libxcursor/-/merge_requests/2

emersion (approved):
LGTM, thanks!

zzag:
> Thanks. Upstream merge request: https://gitlab.freedesktop.org/xorg/lib/libxcursor/-/merge_requests/2

https://gitlab.freedesktop.org/wayland/wayland has a copy of xcursor.{h,c} as well.

emersion:
@e00E: if you have time, it would be great if you could fix libwayland-cursor as well. There are more reviewers for Wayland repos than for X11.

e00E:
Will do.

Issue #2386, opened 2020-08-30
dmabuf-capture - wl_output can't find display modes 


meMuszr:
Having an issue trying to get the dmabuf-capture example working. 
Looks like it's not being able to pass the right screen geometry to libav and not quite sure how to debug this one. 
Here's a few command outputs showcasing the problem. 

```sh

❯ WAYLAND_DEBUG=1 ./dmabuf-capture -o 36 test.mkv
[1307836.005]  -> wl_display@1.get_registry(new id wl_registry@2)
[1307836.034]  -> wl_display@1.sync(new id wl_callback@3)
[1307836.146] wl_display@1.delete_id(3)
[1307836.162] wl_registry@2.global(1, "wl_shm", 1)
[1307836.174] wl_registry@2.global(2, "wl_drm", 2)
[1307836.191] wl_registry@2.global(3, "zwp_linux_dmabuf_v1", 3)
[1307836.200] wl_registry@2.global(4, "wl_compositor", 4)
[1307836.208] wl_registry@2.global(5, "wl_subcompositor", 1)
[1307836.222] wl_registry@2.global(6, "wl_data_device_manager", 3)
[1307836.232] wl_registry@2.global(7, "zwlr_gamma_control_manager_v1", 1)
[1307836.240] wl_registry@2.global(8, "gtk_primary_selection_device_manager", 1)
[1307836.251] wl_registry@2.global(9, "zxdg_output_manager_v1", 3)
[1307836.263] wl_registry@2.global(10, "org_kde_kwin_idle", 1)
[1307836.274] wl_registry@2.global(11, "zwp_idle_inhibit_manager_v1", 1)
[1307836.284] wl_registry@2.global(12, "zwlr_layer_shell_v1", 2)
[1307836.299] wl_registry@2.global(13, "xdg_wm_base", 2)
[1307836.316] wl_registry@2.global(14, "zwp_tablet_manager_v2", 1)
[1307836.335] wl_registry@2.global(15, "org_kde_kwin_server_decoration_manager", 1)
[1307836.349] wl_registry@2.global(16, "zxdg_decoration_manager_v1", 1)
[1307836.361] wl_registry@2.global(17, "zwp_relative_pointer_manager_v1", 1)
[1307836.372] wl_registry@2.global(18, "zwp_pointer_constraints_v1", 1)
[1307836.384] wl_registry@2.global(19, "wp_presentation", 1)
[...]

PR #2387, opened 2020-08-31
examples/dmabuf-capture: add extra roundtrip for wl_output listener

emersion:
This example was relying on wl_display_dispatch being enough to fetch
output information. This worked by chance.

Add an explicit wl_display_roundtrip.

Other examples don't setup wl_output listeners, so they should be fine.

Fixes: 297354f84772 ("Remove unnecessary wl_display_dispatch calls")
Closes: https://github.com/swaywm/wlroots/issues/2386


ddevault:
To be pedantic, it could take several round-trips for output info to show up. The compositor can send it along whenever it likes. Xwayland for example just keeps round-tripping until it has everything.

PR #2389, opened 2020-09-02
Accommodate for CONFIG_VT=0, all TTYs are in seat0, but not all seat0s have TTYs

n3rdopolis:
This has not been tested on a kernel that has been compiled with CONFIG_VT=0, but it has been tested on one with CONFIG_VT=1


emersion (commented):
This looks fine to me.

Implementation in systemd:

https://github.com/systemd/systemd/blob/1cabd2d0c56b7de73e4a4fb645f3bbed4a528d2c/src/login/logind-seat.c#L551

cc @ascent12

emersion (approved):
Thanks!

PR #2390, opened 2020-09-02
Quieten failure to set login session type

tokyovigilante:
(almost certainly due to systemd version <246)

This will move a log regarding a failure to set the logind session type to `wayland` down to debug level. It will always fail on systemd <246, and (hopefully never) on systemd >246. 

I've left the `XDG_SESSION_TYPE` as an error, as failure to set an environment variable seems like a error to me, and apparently not setting this breaks Zoom at least. I've also made the whole function return void as there seems to be no reason to check this error as we can do nothing about it and it's not fatal. 

Fixes https://github.com/swaywm/sway/issues/5655

### DIFF
--- a/backend/session/logind.c
+++ b/backend/session/logind.c
@@ -242,7 +242,7 @@ static bool take_control(struct logind_session *session) {
 	return ret >= 0;
 }
 
-static bool set_type(struct logind_session *session) {
+static void set_type(struct logind_session *session) {
 	int ret;
 	sd_bus_message *msg = NULL;
 	sd_bus_error error = SD_BUS_ERROR_NULL;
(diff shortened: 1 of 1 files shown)


backend/session/logind.c
```diff
@@ -251,23 +251,17 @@ static bool set_type(struct logind_session *session) {
...
 
 	sd_bus_error_free(&error);
 	sd_bus_message_unref(msg);
 
-	if (ret < 0) {
-		return false;
-	}
-
-	ret = setenv("XDG_SESSION_TYPE", "wayland", 1);
+	setenv("XDG_SESSION_TYPE", "wayland", 1);
```
emersion, line 261: We shouldn't set XDG_SESSION_TYPE if SetType failed, because the two will become out-of-sync in this case.

tokyovigilante: Even though it can't be set on old systemd? They will be out of sync but setting 1/2 seems more correct than 0/2. I'm relaxed though if you want to return early in that case though.

emersion: Yes, Lennard said keeping both in sync is better.

tokyovigilante: Ok, will update.

tokyovigilante:
Whoops didn't mean to remove resetting `ret` for the `setenv()` call.

emersion (approved):
Thanks!

PR #2392, opened 2020-09-03
output-management-v1: add head identifying information

subdiff:
Add events to zwlr_output_head_v1 for make, model and serial number of a display. This allows clients to identify displays from previous compositor sessions and reload accordingly configurations that were saved previously.

I have an example implementation with:
* Generic Wayland client library: https://gitlab.com/kwinft/wrapland/-/merge_requests/75
* Output management client library: https://gitlab.com/kwinft/disman/-/merge_requests/9
* KDisplay client: https://gitlab.com/kwinft/kdisplay/-/merge_requests/6

screenshot

The form in which the client combines this information to recognize the display again is unspecified and depends on how permissive it is.

For example in case of Disman I use a string `<make>:<model>:<serial_number>:<name>` which I then MD5-hash and hex convert for the name of the output configuration files.

I also use the name in there because I only want to load a configuration when the display is connected to the same connector (displays can have different capabilities on different connectors).

### DIFF
--- a/types/wlr_output_management_v1.c
+++ b/types/wlr_output_management_v1.c
@@ -6,7 +6,7 @@
 #include "util/signal.h"
 #include "wlr-output-management-unstable-v1-protocol.h"
 
-#define OUTPUT_MANAGER_VERSION 1
+#define OUTPUT_MANAGER_VERSION 2
 
 enum {
 	HEAD_STATE_ENABLED = 1 << 0,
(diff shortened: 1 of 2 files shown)


emersion:
Thanks for working on this! Can you submit a PR against wlr-protocol too?

subdiff:
> Thanks for working on this! Can you submit a PR against wlr-protocol too?

Sure, wasn't sure how this is supposed to work so I first put it here. I will create one with only the first commit there.

types/wlr_output_management_v1.c
```diff
@@ -761,6 +761,16 @@ static void manager_send_head(struct wlr_output_manager_v1 *manager,
 			output->phys_width, output->phys_height);
 	}
 
+	if (version >= ZWLR_OUTPUT_HEAD_V1_MAKE_SINCE_VERSION) {
```
emersion, line 764: Can we not send anything if these are empty strings?

subdiff: I tried with strnlen and strnlen_s but they are not declared. What's the right way to check if the string is empty in C (could be not null-terminated)?

subdiff: Checking the first element of the strings is possible though because initialized to 0 and then `make[0] == '\0' == 0`. Patch updated.

emersion (approved):
Code LGTM, let's wait for the wlr-protocols MR to get merged before merging this

emersion:
Thanks!

PR #2394, opened 2020-09-03
wayland: emit relative pointer events only for current pointer

ammen99:
As discussed on IRC, without this check, the wayland backend receives the relative motion event once per output, and emits these events on all pointer devices. This can cause confusion in the compositor.


emersion (approved):
Thanks!

PR #2396, opened 2020-09-04
xwayland: disconnect display destroy listener even if xwayland didn't initialize

ammen99:
I got this shutdown crash where Xwayland isn't deinitialized properly. I think the problem might be that Xwayland itself crashes after shutdown has started but fails to restart (since we're already tearing things down).

Anyway, it can be seen that this is a problem: https://github.com/swaywm/wlroots/blob/master/xwayland/server.c#L250

The listener is added, then display sockets are opened, if that fails, `server_finish_display` is called but it is a no-op because of the check in the beginning.


emersion (approved):
Good catch, thanks!

PR #2402, opened 2020-09-16
backend: remove check for _WAYLAND_DISPLAY

emersion:
I'm not sure what this was used for, but it's not used by libwayland.
Setting _WAYLAND_DISPLAY would result in the Wayland backend being
picked but would ignore the actual value of the env variable.


ddevault (approved):

ddevault:
Thanks!

PR #2405, opened 2020-09-19
xwm: add loop detection for read_surface_parent

Emantor:
Implement a simple loop detection while trying to retrieve the parent
for a TRANSIENT_FOR window.

Fixes swaywm/sway#4624

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -467,19 +467,40 @@ static void read_surface_title(struct wlr_xwm *xwm,
 	wlr_signal_emit_safe(&xsurface->events.set_title, xsurface);
 }
 
+static bool has_parent(struct wlr_xwayland_surface *parent,
+		struct wlr_xwayland_surface *child) {
+	while (parent) {
+		if (child == parent) {
+			return true;
+		}
+
+		parent = parent->parent;
+	}
+
+	return false;
+}
+
 static void read_surface_parent(struct wlr_xwm *xwm,
 		struct wlr_xwayland_surface *xsurface,
 		xcb_get_property_reply_t *reply) {
+	struct wlr_xwayland_surface *found_parent = NULL;
 	if (reply->type != XCB_ATOM_WINDOW) {
 		return;
 	}
 
 	xcb_window_t *xid = xcb_get_property_value(reply);
 	if (xid != NULL) {
-		xsurface->parent = lookup_surface(xwm, *xid);
+		found_parent = lookup_surface(xwm, *xid);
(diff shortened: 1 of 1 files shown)


xwayland/xwm.c
```diff
@@ -467,20 +467,39 @@ static void read_surface_title(struct wlr_xwm *xwm,
 	wlr_signal_emit_safe(&xsurface->events.set_title, xsurface);
 }
 
+static bool check_transient_loop(struct wlr_xwayland_surface *candidate,
```
emersion, line 470: This could be renamed `has_parent`

xwayland/xwm.c
```diff
@@ -467,20 +467,39 @@ static void read_surface_title(struct wlr_xwm *xwm,
...
+		struct wlr_xwayland_surface *parent) {
+	while (parent) {
+		if (candidate == parent)
```
emersion, line 473: Style: braces are mandatory

xwayland/xwm.c
```diff
@@ -467,20 +467,39 @@ static void read_surface_title(struct wlr_xwm *xwm,
...
 
 	xcb_window_t *xid = xcb_get_property_value(reply);
 	if (xid != NULL) {
-		xsurface->parent = lookup_surface(xwm, *xid);
+		candidate = lookup_surface(xwm, *xid);
 	} else {
 		xsurface->parent = NULL;
 	}
 
+	if (!check_transient_loop(xsurface, candidate)) {
```
emersion, line 497: `check_transient_loop` is called even if `xsurface->parent` is set to NULL above, which is surprising. Can we move this in the `if (xid != NULL)` block?

xwayland/xwm.c
Emantor: I don't think `has_parent` fits here, maybe `is_parent`?

xwayland/xwm.c
Emantor: Yep, definitely.

xwayland/xwm.c
emersion: `has_parent` would be fine if the args were reversed. BTW maybe we can find better names for these args.

Emantor: I'll go with `parent` and `child` and `has_parent`. Also renamed the `candidate` variable to `found_parent`.

crocket:
Check https://github.com/swaywm/sway/issues/4624#issuecomment-695192840

Emantor:
Gentle Ping.

emersion (approved):
Thanks!

PR #2408, opened 2020-09-21
data-control: Don't send offers to the clients that sent them

### LINKED ISSUE #2406: wlr-data-control makes offers to clients that sent them
When using wlr-data-control protocol and listening for offers, any offer sent is also broadcast back to the client that sent it. This means that when the client tries to receive the data, it read()s from a fd that the same client has to write(). Of course, this causes the client to block forever. It doesn't really make sense to send something back to the client that it just sent. This makes me think that the protocol should state that sent sources are not rebroadcast back to the same client as offers. Thoughts?

soreau:
If a client is listening for offers and sends an offer, it will block forever
when it attempts to receive the offer and read from the file descriptor. This
is because the offer is sent back to the client that sent it. Since the writing
and reading ends are in the same client, it never has a chance to write
(barring methods to work around this problem). Even if the client could read
its own selection, it is pointless to do so.

Here we fix the problem by not making offers back to the client that sent them.
The protocol has been updated to reflect this change.
Closes #2406.

### DIFF
--- a/protocol/wlr-data-control-unstable-v1.xml
+++ b/protocol/wlr-data-control-unstable-v1.xml
@@ -108,7 +108,8 @@
         the primary clipboard selections). Immediately following the
         wlr_data_control_device.data_offer event, the new data_offer object
         will send out wlr_data_control_offer.offer events to describe the MIME
-        types it offers.
+        types it offers. Offers will be propagated to other listening clients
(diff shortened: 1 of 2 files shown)


emersion:
This breaks backwards compatibility and is inconsistent with the core Wayland protocol.

PR #2411, opened 2020-10-01
output-management-v1: send complete head state on enable change

subdiff:
Send all head data when an output gets enabled since data can change while the head is disabled and the changed data in this case not sent.

I noticed that issue with sway and Disman on the repeated hot-plug of an external display with a position different to (0,0). The position event was only received the first time. Afterwards the position event was not received anymore. This also happened when Disman was restarted but the sway session continued.


subdiff:
The freebsd build failed, but I don't think it's because of this change here. Can I restart the build somehow?

agx:
I just hit the same this weekend and this fixes it.

agx (approved):

types/wlr_output_management_v1.c
```diff
@@ -680,6 +680,7 @@ static void head_send_state(struct wlr_output_head_v1 *head,
 
 	if (state & HEAD_STATE_ENABLED) {
 		zwlr_output_head_v1_send_enabled(head_resource, head->state.enabled);
+		state = ~(state & 0);
```
agx, line 683: just wonder if setting `state = HEAD_STATE_ALL` improves readbility?

subdiff: Definitely. Didn't notice this variable. I'll change it. Thanks!

emersion (approved):
Good catch. LGTM, thanks!

PR #2413, opened 2020-10-03
Factor out wlr_wl_seat in Wayland backend

ony:
~~*Note  that this PR is based on `0.11.0`, because `master` at the moment is crashing my sway for some reason.*~~

#### Change in output to use `current_pointer`

This is just to bring in sync `seat.c` (we propagate events to `current_pointer`) and `output.c`. Though we have at most one pointer at the moment, this may simplify some further work by dropping implicit assumption `assert(!wl->current_pointer || wl->current_pointer->wl_pointer == wl->pointer)` that we already seems to have in `seat.c`.

### Testing
```sh
WLR_WL_OUTPUTS=2 WLR_NO_HARDWARE_CURSORS=1 WLR_BACKENDS=wayland cage -s -D -- weston-terminal
```
Started under Sway (running system `wlroots-0.11.0`) with single seat.

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -103,9 +103,9 @@ static void registry_global(void *data, struct wl_registry *registry,
 		wl->compositor = wl_registry_bind(registry, name,
 			&wl_compositor_interface, 4);
 	} else if (strcmp(iface, wl_seat_interface.name) == 0) {
-		wl->seat = wl_registry_bind(registry, name,
+		struct wl_seat *wl_seat = wl_registry_bind(registry, name,
 			&wl_seat_interface, 5);
-		wl_seat_add_listener(wl->seat, &seat_listener, wl);
+		create_wl_seat(wl_seat, wl);
 	} else if (strcmp(iface, xdg_wm_base_interface.name) == 0) {
 		wl->xdg_wm_base = wl_registry_bind(registry, name,
 			&xdg_wm_base_interface, 1);
@@ -155,13 +155,16 @@ static bool backend_start(struct wlr_backend *backend) {
 
 	wl->started = true;
 
-	if (wl->keyboard) {
-		create_wl_keyboard(wl->keyboard, wl);
-	}
+	struct wlr_wl_seat *seat = wl->seat;
+	if (seat != NULL) {
+		if (seat->keyboard) {
+			create_wl_keyboard(seat->keyboard, wl);
+		}
 
-	if (wl->tablet_manager && wl->seat) {
-		wl_add_tablet_seat(wl->tablet_manager,
-			wl->seat, wl);
+		if (wl->tablet_manager) {
+			wl_add_tablet_seat(wl->tablet_manager,
+				seat->wl_seat, wl);
+		}
 	}
 
 	for (size_t i = 0; i < wl->requested_outputs; ++i) {
@@ -192,8 +195,6 @@ static void backend_destroy(struct wlr_backend *backend) {
 
 	wl_list_remove(&wl->local_display_destroy.link);
 
-	free(wl->seat_name);
-
 	wl_event_source_remove(wl->remote_display_src);
 
 	wlr_renderer_destroy(wl->renderer);
@@ -201,12 +202,7 @@ static void backend_destroy(struct wlr_backend *backend) {
 
 	wlr_drm_format_set_finish(&wl->linux_dmabuf_v1_formats);
 
-	if (wl->pointer) {
-		wl_pointer_destroy(wl->pointer);
-	}
-	if (wl->seat) {
-		wl_seat_destroy(wl->seat);
-	}
+	destroy_wl_seats(wl);
 	if (wl->zxdg_decoration_manager_v1) {
 		zxdg_decoration_manager_v1_destroy(wl->zxdg_decoration_manager_v1);
 	}
--- a/backend/wayland/seat.c
+++ b/backend/wayland/seat.c
@@ -48,6 +48,7 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 	struct wlr_wl_output *output = wl_surface_get_user_data(surface);
 	assert(output);
 	struct wlr_wl_pointer *pointer = output_get_pointer(output);
+	assert(!backend->current_pointer || backend->current_pointer == pointer);
 
 	output->enter_serial = serial;
 	backend->current_pointer = pointer;
--- a/backend/wayland/seat.c
+++ b/backend/wayland/seat.c
@@ -372,12 +372,49 @@ static struct wlr_wl_input_device *get_wl_input_device_from_input_device(
 	return (struct wlr_wl_input_device *)wlr_dev;
 }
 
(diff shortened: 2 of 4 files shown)


include/backend/wayland.h
```diff
@@ -14,6 +14,15 @@
 #include <wlr/types/wlr_box.h>
 #include <wlr/render/drm_format_set.h>
 
+struct wlr_wl_seat {
+	struct wl_seat *wl_seat;
+
+	/* TODO: struct wl_list link; */
```
ammen99, line 20: is there a reason not to use `wl_list` here?

ony: This PR is only for factoring out stuff into structure that will be in the list in next https://github.com/swaywm/wlroots/pull/2417 that actually introduce multi-seat support.

backend/wayland/seat.c
```diff
@@ -282,12 +283,40 @@ static struct wlr_wl_input_device *get_wl_input_device_from_input_device(
 	return (struct wlr_wl_input_device *)wlr_dev;
 }
 
+void create_wl_seat(struct wl_seat *wl_seat, struct wlr_wl_backend *wl) {
```
emersion, line 286: `create_*` functions should allocate a new structure and return it.

OTOH, `init_*` functions fill an already-existing structure.

backend/wayland/seat.c
```diff
@@ -282,12 +283,40 @@ static struct wlr_wl_input_device *get_wl_input_device_from_input_device(
...
+	wl->seat.wl_seat = wl_seat;
+	wl_seat_add_listener(wl_seat, &seat_listener, wl);
+}
+
+void destroy_wl_seats(struct wlr_wl_backend *wl) {
```
emersion, line 291: Ditto, but with `destroy_*`/`finish_*`.

emersion, line 291: This should cleanup `keyboard` too.

backend/wayland/seat.c
```diff
@@ -282,12 +283,40 @@ static struct wlr_wl_input_device *get_wl_input_device_from_input_device(
...
+
+static struct wlr_wl_seat *backend_get_seat(struct wlr_wl_backend *backend, struct wl_seat *wl_seat) {
+	assert(backend->seat.wl_seat == wl_seat);
+	return &backend->seat;
+}
+
+static struct wlr_wl_seat *input_device_get_seat(struct wlr_input_device *wlr_dev) {
+	struct wlr_wl_input_device *dev =
+		get_wl_input_device_from_input_device(wlr_dev);
+	// TODO: assert on pointers match (e.g. keyboard/pointer)?
```
emersion, line 309: We should do this.

emersion (changes requested):
Thanks for your PR!

include/backend/wayland.h
```diff
@@ -114,11 +120,18 @@ void create_wl_pointer(struct wl_pointer *wl_pointer, struct wlr_wl_output *outp
...
 extern const struct wl_seat_listener seat_listener;
 
 struct wlr_wl_tablet_seat *wl_add_tablet_seat(
 		struct zwp_tablet_manager_v2 *manager,
 		struct wl_seat *seat, struct wlr_wl_backend *backend);
 
+/* Similar to wl_list_for_each(). No guarantee that after loop var will be set to last value
+ * For now we have only one or none seats
+ */
+#define wlr_wl_backend_for_each_seat(var, wl) for (var = &wl->seat; var && var->wl_seat; var = NULL)
```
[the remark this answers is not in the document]
ony: I think you misinterpreted this one. In this PR there is no list of seats. This one is just to abstract away if we have 0..1 or 0..n seats in the code. This macro replaces `if (bakcend->wl_seat)` and similar code so later we don't need to touch it.  
In https://github.com/swaywm/wlroots/pull/2417 this becomes just a wrapper around `wl_list_for_each`.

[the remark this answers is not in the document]
emersion: All right. We don't do macros in wlroots, so we need to remove it.

backend/wayland/seat.c
ony: This matches `create_wl_pointer` and `create_wl_keyboard` that do not return created objects. Though they link them in `backend->devices`.  
Similar to those this one takes care of "linking" in backend. It just happens that in this PR it is not behind `wl_list` in `backend`.

Note that in https://github.com/swaywm/wlroots/pull/2417 this method actually going to create objects and link it in `backend`. But I'm not fully sure it is correct to write `init_*` because we are not initializing single field.

I think `add_wl_seat` will be a better fit here.

emersion: Name it `init` in this PR, rename to `create` in your other PR.

backend/wayland/seat.c
ony: This is just re-factoring. I merely moved out exiting code from `backend_destroy`:

```c
	free(wl->seat_name);
	// ...
	if (wl->pointer) {
		wl_pointer_destroy(wl->pointer);
	}
	if (wl->seat) {
		wl_seat_destroy(wl->seat);
	}
```

Note that there is a difference in how pointer and keyboard are handled (see `seat_handle_capabilities`).  
In particular, `wlr_keyboard` created  and destroyed right away as soon as backend is started and exist in single instance. Hence `wl_keyboard_destroy` done in `input_device_destroy` since sole ownership allows it.  
OTOH, `wlr_pointer` is created per output. Thus ownership for host `wl_pointer` will have to be somehow shared and as a simpliest way is to achieve that is lift ownership to object with matching life-cyle e.g. `wlr_wl_backend` originally and `wlr_wl_seat` in this PR (effectively aligned with `wlr_wl_backend`).

backend/wayland/seat.c
ony: Unfortunately at this moment we don't store any association between `wlr_input_device` and say `wl_keyboard` other than subscribing to it. I left this to-do at moment where I didn't grasped Wayland backend fully enough.

backend/wayland/seat.c
ony: As per suggestion in IRC, now `wlr_wl_seat` gets allocated independently from `wlr_wl_backend` and thus name `create_*` is applicable.

backend/wayland/backend.c
```diff
@@ -155,13 +155,16 @@ static bool backend_start(struct wlr_backend *backend) {
 
 	wl->started = true;
 
-	if (wl->keyboard) {
-		create_wl_keyboard(wl->keyboard, wl);
-	}
+	struct wlr_wl_seat *seat;
+	if ((seat = wl->seat) != NULL) {
```
emersion, line 159: Style: can we move out the assignment above?

```c
struct wlr_wl_seat *seat = wl->seat;
if (seat != NULL) {
```

backend/wayland/seat.c
emersion: Hm, OK, I see.

emersion (changes requested):
Looks good to me otherwise! Thanks for bearing with me.

emersion (approved):
LGTM, thanks!

PR #2414, opened 2020-10-03
xwayland: introduce WLR_XWAYLAND for specifying which Xwayland to use

Xyene:
When debugging Xwayland-related issues, a common first step in debugging
has been to ask the reporter to move their real Xwayland to
/usr/bin/Xwayland.bin, and create a shell script starting Xwayland with
extra arguments under the original /usr/bin/Xwayland location.

Introducing a `WLR_XWAYLAND` environment variable makes this less
invasive, by allowing the user to swap out Xwayland without resorting to
global system changes (or source patches).

### DIFF
--- a/xwayland/server.c
+++ b/xwayland/server.c
@@ -107,8 +107,16 @@ noreturn static void exec_xwayland(struct wlr_xwayland_server *server) {
 		dup2(devnull, STDERR_FILENO);
 	}
 
+	const char *xwayland_path = getenv("WLR_XWAYLAND");
+	if (xwayland_path) {
+		wlr_log(WLR_INFO, "Using Xwayland binary to '%s' due to WLR_XWAYLAND",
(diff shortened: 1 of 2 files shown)


Emantor (approved):

xwayland/server.c
```diff
@@ -107,8 +107,13 @@ noreturn static void exec_xwayland(struct wlr_xwayland_server *server) {
 		dup2(devnull, STDERR_FILENO);
 	}
 
+	const char *xwayland_path = getenv("WLR_XWAYLAND");
```
emersion, line 110: Can we print a debug message when this is used? Since it's a debugging variable, I wouldn't want users to have issues because of a left-over env.

Xyene: Good idea, done.

emersion (approved):
Thanks!

Issue #2416, opened 2020-10-03
Intermittent SEGV in wl_list_insert on shotdown of two-level Sway with wl-roots from master (87836dcb)


ony:
```plain
  Command Line: sway -d -c /home/nikolay/contrib/wlroots/swaysway-config
    Executable: /usr/x86_64-pc-linux-gnu/bin/sway
 Control Group: /user.slice/user-1000.slice/session-223.scope
          Unit: session-223.scope
         Slice: user-1000.slice
       Session: 223
     Owner UID: 1000 (nikolay)
       Boot ID: 7aee15a50c7446fa9994c2ab0b7dc118
    Machine ID: 5fb39b2aa86d4cd2a3446853c714b43c
      Hostname: ony
       Storage: /var/lib/systemd/coredump/core.sway.1000.7aee15a50c7446fa9994c2ab0b7dc118.1049167.1601761366000000.zst
       Message: Process 1049167 (sway) of user 1000 dumped core.

                Stack trace of thread 1049167:
                #0  0x00007f00691c1540 wl_list_insert (libwayland-server.so.0 + 0xd540)
                #1  0x000055bd462f8582 n/a (sway + 0x25582)
                #2  0x00007f006916a9f8 n/a (/tmp/ws/build-wlroots-gcc9-debug/libwlroots.so.6 + 0x869f8)
                #3  0x00007f006913ff25 n/a (/tmp/ws/build-wlroots-gcc9-debug/libwlroots.so.6 + 0x5bf25)
                #4  0x00007f0069161f8e n/a (/tmp/ws/build-wlroots-gcc9-debug/libwlroots.so.6 + 0x7df8e)
                #5  0x00007f006916220c n/a (/tmp/ws/build-wlroots-gcc9-debug/libwlroots.so.6 + 0x7e20c)
                #6  0x00007f00686cd96d n/a (libffi.so.7 + 0x696d)
                #7  0x00007f00686ccf7b n/a (libffi.so.7 + 0x5f7b)
                #8  0x00007f00691c0cdd n/a (libwayland-server.so.0 + 0xccdd)
                #9  0x00007f00691bd607 n/a (libwayland-server.so.0 + 0x9607)
                #10 0x00007f00691bf0ea wl_event_loop_dispatch (libwayland-server.so.0 + 0xb0ea)
                #11 0x00007f00691bd7e5 wl_display_run (libwayland-server.so.0 + 0x97e5)
[...]

Issue #2415, opened 2020-10-03
Software cursor in wayland backend doesn't render for tablet and no movements reported for pointing tool


ony:
Reproduce with running Sway (started from agetty) and tablet per-attached:
```sh
WLR_NO_HARDWARE_CURSORS=1 cage -s -- wev
```

When starting `wev` from root compositor both relative movements with touch-pad mode and absolute with pointing tool hovering over tablet results in:
```
[13:      wl_pointer] motion: time: -1341421357; x, y: 45.039062, 75.171875
```

From inner compositor only touch-pad shows cursors and reports events. Pointing device only reported during  "clicks" when pressing with a tool.

P.S.
Tablet:
```
interface: 'zwp_tablet_manager_v2',                      version:  1, name: 14
	tablet_seat: seat0
		tablet: Wacom Bamboo 16FG 4x5 Pen
			vendor: 1386
			product: 222
			path: /sys/devices/pci0000:00/0000:00:14.0/usb1/1-14/1-14.4/1-14.4:1.0/0003:056A:00DE.013E/input/input342/event31
		pad:
			buttons: 4
			path: /sys/devices/pci0000:00/0000:00:14.0/usb1/1-14/1-14.4/1-14.4:1.1/0003:056A:00DE.013F/input/input347/event257
			group:
				modes: 1
				strips: 0
				rings: 0
				buttons: 0 1 2 3
```

Issue #1499, opened 2019-01-25
Add multi-seat support to Wayland backend


YaLTeR:
209210d30780ec64995594b77fde3d718b655542
Crash on launch.
```
2019-01-25 20:33:53 - [rootston/config.c:537] No config file found. Using sensible defaults.
2019-01-25 20:33:53 - [backend/wayland/backend.c:186] Creating wayland backend
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: wl_shm v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: zwp_linux_dmabuf_v1 v3
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: wl_drm v2
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: wl_compositor v4
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: wl_subcompositor v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: wl_data_device_manager v3
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: gamma_control_manager v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: zwlr_gamma_control_manager_v1 v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: gtk_primary_selection_device_manager v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: zxdg_output_manager_v1 v2
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: org_kde_kwin_idle v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: zwp_idle_inhibit_manager_v1 v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: zwlr_layer_shell_v1 v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: zxdg_shell_v6 v1
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: xdg_wm_base v2
2019-01-25 20:33:53 - [backend/wayland/backend.c:65] Remote wayland global: org_kde_kwin_server_decoration_manager v1
[...]

PR #2417, opened 2020-10-04
Mutli-seat support in Wayland backend

ony:
*~~Note  that this PR is based on https://github.com/swaywm/wlroots/pull/2413 `0.11.0`, because latest changes at the moment are crashing my sway test setup https://github.com/swaywm/wlroots/issues/2416.~~*

Intention is to get close to resolving https://github.com/swaywm/wlroots/issues/1499 by getting in one of these:
- ~~exclusive seat per backend (or output)~~.
- ~~muxing pointers/keyboards into virtual ones (same seat)~~
- ~~muxing multiple seats into single virtual one with single/multiple pointer and keyboard~~
- proxying seats as inputs 1:1

Any of those should work for my needs.

Worth to note that as of now we multiple pointers by amount of outputs ~~. Not sure purpose of this.~~ to map absolute coordinates on particular output.

### Progress
* [x] Track multiple seats from outer compositor.
* ~~Switching (`backend->current_pointer`) between seats based on entered/left pointer (for exclusive behavior).~~
* [x] Listen to events from multiple pointers.
* [x] Expose multiple inputs downstream.
* [x] Distinguishable input names to allow seat management in downstream Sway.
* [x] Track down bug triggered by switching VTs.
* [x] Make ~~`current_pointer`~~ *`active_pointer`* be local to `wrl_wl_seat` and point which `wlr_pointer` needs to receive events based on output we entered with host `wl_pointer`.
* [x] Track down bug with clicks/focus propagation in sway-(cage-sway-alacrity)x2 to get two pointers working simulatenously.  
*We were always reporting to `backend->current_pointer`*
* [x] Move `backend->current_pointer` to `wlr_wl_output` since right now it only required there for sending host pointer cursor surface.  
*For me it is still not clear how cursor is set ~~for individual seats since we receive only output in `wlr_output_impl`~~ and how compositor knows if back-end support hardware cursor. For now moved to `output->cursor.pointer` in d6f7dffb547d640041cb2caa892691a8e2d52696 assuming that potentially different outputs might be a different GPUs with own hardware cursors.*

### Testing
Root Sway compositor with two non-empty seats

Checking how seats/inputs are advertised:

*
  ```sh
  cage -s -- wayland-info
  ```
* 
  ```sh
  # dummy-config
  # run: sway -c dummy-config
  exec "swaymsg -t get_inputs; swaymsg -t get_seats; swaymsg exit"
  ```

Checking cursor and events propagation:
```sh
cage -s -D -- wev
cage -s -D -- weston-terminal
```
* Slowly moving through boundaries triggering cursor change to re-size icon before entering output window.
* Check motion events from pointer.

Multi-output: `WLR_WL_OUTPUTS=2` 
Software cursor: `WLR_NO_HARDWARE_CURSORS=1`

#### Goal
Share screen for two people:
```
# inputs
seat seat1 attach {
  # ...
}

# minimal visual split
default_border pixel
hide_edge_borders both

# bootstrap cells
exec cage -s -- alacritty
exec cage -s -- alacritty
```
(from `alacritty` we can spawn `login` or `exec` compositor)

* :heavy_check_mark:  Works for keyboard with input focus following pointer from same seat.  
* :heavy_check_mark: Mouse hover/click and movements do propagate to both windows independently.
* ~~:grey_question: Didn't checked touch/pen events. I have only one digitizing tablet, so can only employ Sway ability to attach same device to both seats in order to check it.~~ Doesn't work regardless of this PR. See https://github.com/swaywm/wlroots/issues/2415 .

### DIFF
--- a/backend/wayland/seat.c
+++ b/backend/wayland/seat.c
@@ -448,7 +448,11 @@ struct wlr_wl_input_device *create_wl_input_device(
 	struct wlr_input_device *wlr_dev = &dev->wlr_input_device;
 
 	unsigned int vendor = 0, product = 0;
-	const char *name = "wayland";
+
+	size_t name_size = 8 + strlen(seat->name) + 1;
+	char name[name_size];
+	(void) snprintf(name, name_size, "wayland-%s", seat->name);
+
 	wlr_input_device_init(wlr_dev, type, &input_device_impl, name, vendor,
 		product);
 	wl_list_insert(&seat->backend->devices, &wlr_dev->link);
--- a/backend/wayland/seat.c
+++ b/backend/wayland/seat.c
@@ -412,8 +412,8 @@ void destroy_wl_seats(struct wlr_wl_backend *wl) {
 static struct wlr_wl_seat *input_device_get_seat(struct wlr_input_device *wlr_dev) {
 	struct wlr_wl_input_device *dev =
 		get_wl_input_device_from_input_device(wlr_dev);
-	assert(dev->backend->seat);
-	return dev->backend->seat;
+	assert(dev->seat);
+	return dev->seat;
 }
 
 static void input_device_destroy(struct wlr_input_device *wlr_dev) {
@@ -437,14 +437,15 @@ bool wlr_input_device_is_wl(struct wlr_input_device *dev) {
 }
 
 struct wlr_wl_input_device *create_wl_input_device(
-		struct wlr_wl_backend *backend, enum wlr_input_device_type type) {
+		struct wlr_wl_seat *seat, enum wlr_input_device_type type) {
 	struct wlr_wl_input_device *dev =
 		calloc(1, sizeof(struct wlr_wl_input_device));
 	if (dev == NULL) {
 		wlr_log_errno(WLR_ERROR, "Allocation failed");
 		return NULL;
 	}
-	dev->backend = backend;
+	dev->backend = seat->backend;
+	dev->seat = seat;
 
 	struct wlr_input_device *wlr_dev = &dev->wlr_input_device;
 
@@ -452,7 +453,7 @@ struct wlr_wl_input_device *create_wl_input_device(
 	const char *name = "wayland";
 	wlr_input_device_init(wlr_dev, type, &input_device_impl, name, vendor,
 		product);
-	wl_list_insert(&backend->devices, &wlr_dev->link);
+	wl_list_insert(&seat->backend->devices, &wlr_dev->link);
 	return dev;
 }
 
@@ -617,7 +618,9 @@ static void pointer_handle_output_destroy(struct wl_listener *listener,
 	wlr_input_device_destroy(&pointer->input_device->wlr_input_device);
 }
 
-void create_wl_pointer(struct wl_pointer *wl_pointer, struct wlr_wl_output *output) {
+void create_wl_pointer(struct wlr_wl_seat *seat, struct wlr_wl_output *output) {
+	assert(seat->pointer);
+	struct wl_pointer *wl_pointer = seat->pointer;
 	struct wlr_wl_backend *backend = output->backend;
 
 	struct wlr_input_device *wlr_dev;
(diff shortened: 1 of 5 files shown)


backend/wayland/seat.c
```diff
@@ -559,21 +605,21 @@ void create_wl_keyboard(struct wl_keyboard *wl_keyboard, struct wlr_wl_backend *
...
 		wl_list_for_each(output, &backend->outputs, link) {
-			create_wl_pointer(wl_pointer, output);
+			create_wl_pointer(seat, output);
 		}
 	}
-	if (!(caps & WL_SEAT_CAPABILITY_POINTER) && backend->pointer != NULL) {
+	if (!(caps & WL_SEAT_CAPABILITY_POINTER) && seat->pointer != NULL) {
 		wlr_log(WLR_DEBUG, "seat %p dropped pointer", (void *)wl_seat);
 
 		struct wlr_input_device *device, *tmp;
```
ony, line 625: We should destroy only inputs for particular seat. Same for keyboard.

Right now unplugging all inputs forming capability of one seat destroy corresponding pointer/keyboard for another seat.

backend/wayland/tablet_v2.c
```diff
@@ -818,7 +819,7 @@ static const struct zwp_tablet_tool_v2_listener tablet_tool_listener = {
 	.frame = handle_tablet_tool_frame,
 };
 
-static void handle_tool_added(void *data,
+static void handle_tool_added(void *unused_data,
```
emersion, line 821: Let's keep `data` here, no need for the `unused_` prefix.

include/backend/wayland.h
```diff
@@ -38,7 +38,7 @@ struct wlr_wl_backend {
 	struct wp_presentation *presentation;
 	struct zwp_linux_dmabuf_v1 *zwp_linux_dmabuf_v1;
 	struct zwp_relative_pointer_manager_v1 *zwp_relative_pointer_manager_v1;
-	struct wlr_wl_seat *seat;
+	struct wl_list seats;
```
emersion, line 41: We generally add a comment describing the list items:

```c
struct wl_list seats; // wlr_wl_seat.link
```

include/backend/wayland.h
```diff
@@ -108,6 +108,7 @@ struct wlr_wl_pointer {
 struct wlr_wl_seat {
 	struct wl_seat *wl_seat;
 
+	struct wl_list link;
```
emersion, line 111: Ditto:

```c
struct wl_list link; // wlr_wl_backend.seats
```

backend/wayland/seat.c
```diff
@@ -448,7 +448,11 @@ struct wlr_wl_input_device *create_wl_input_device(
 	struct wlr_input_device *wlr_dev = &dev->wlr_input_device;
 
 	unsigned int vendor = 0, product = 0;
-	const char *name = "wayland";
+
+	size_t name_len = 8 + strlen(seat->name) + 1;
```
emersion, line 466: I generally try to use "len" for the result of `strlen` (number of items) and "size" for the number of bytes. Since this include the final `\0` I'd be more comfortable with `name_size` here.

emersion (changes requested):
Overall looks very good! I just have a few nits, the rest of the patches LGTM. The review was easy thanks to the clear split into logical commits. :)

include/backend/wayland.h
ony: I see benefit of previous one because `wl_list` doesn't suggest type of list elements. But I'm not fully sure about practice of referencing all the places where we reference this list type.

I don't insist on changing practice in this project. Just suggest to re-consider what we get out of it and how it works in case if we have zero (e.g. callbacks binding) or more than one fields where we have list of list-capable structs.

emersion: The comments have proven useful when new lists are added to a struct and when an item can be added to multiple lists.

emersion (approved):
Thanks!

PR #2419, opened 2020-10-06
foreign-toplevel-management: Report parent toplevel

dkondor:
This is an implementation for https://github.com/swaywm/wlr-protocols/pull/52

Works together with: https://github.com/dkondor/wayfire/tree/foreign-toplevel-report-parent

Example client: https://github.com/dkondor/cairo-dock-core/tree/wayland_egl

Tested: not displaying child views in the taskbar

Untested: actually grouping together views with the same parent

I'm happy to work on additional test cases if needed.

Potential issues:

- I've added a field storing the parent to the wlr_foreign_toplevel_handle_v1 struct. This is necessary so that foreign_toplevel_manager_bind() can send information about parents to the new client. This breaks ABI though.
- In wlr_foreign_toplevel_handle_v1_destroy(), I've added code to check if the destroyed handle is not stored in the parent field for any other toplevel. I don't really like this solution, since this makes destroying a toplevel handle scale with the number of toplevels managed here (and destroying N toplevels will have a complexity of N^2). Since the number of toplevels is expected to be low, I don't think it is a big problem in practice though. Potential alternatives:
  -- Store the children of a toplevel in a list. I don't like this, since it essentially duplicates functionality that is likely already present in the compositor elsewhere.
  -- Do not do check this and require the caller to ensure that either (1) a parent handle is only destroyed after all child handles; (2) wlr_foreign_toplevel_handle_v1_set_parent() is called with NULL or another parent before destroying the parent handle
- Strangely, I could not find any documentation for wl_resource_find_for_client(), so I used it based on the assumption that it does what I think. I'd be grateful if someone more familiar with these interfaces could confirm that I'm using it correctly.

### DIFF
--- a/include/wlr/types/wlr_foreign_toplevel_management_v1.h
+++ b/include/wlr/types/wlr_foreign_toplevel_management_v1.h
@@ -50,6 +50,7 @@ struct wlr_foreign_toplevel_handle_v1 {
 
 	char *title;
 	char *app_id;
+	struct wlr_foreign_toplevel_handle_v1 *parent;
 	struct wl_list outputs; // wlr_foreign_toplevel_v1_output
 	uint32_t state; // wlr_foreign_toplevel_v1_state
 
@@ -104,6 +105,12 @@ struct wlr_foreign_toplevel_manager_v1 *wlr_foreign_toplevel_manager_v1_create(
 
 struct wlr_foreign_toplevel_handle_v1 *wlr_foreign_toplevel_handle_v1_create(
 	struct wlr_foreign_toplevel_manager_v1 *manager);
+/* Destroy the given toplevel handle, sending the closed event to any
+ * client. Also, if the destroyed toplevel is set as a parent of any
+ * other valid toplevel, clients still holding a handle to both are
+ * sent a parent signal with NULL parent. If this is not desired, the
+ * caller should ensure that any child toplevels are destroyed before
+ * the parent. */
 void wlr_foreign_toplevel_handle_v1_destroy(
 	struct wlr_foreign_toplevel_handle_v1 *toplevel);
 
@@ -125,5 +132,15 @@ void wlr_foreign_toplevel_handle_v1_set_activated(
 	struct wlr_foreign_toplevel_handle_v1 *toplevel, bool activated);
 void wlr_foreign_toplevel_handle_v1_set_fullscreen(
 	struct wlr_foreign_toplevel_handle_v1* toplevel, bool fullscreen);
+
+/* Set the parent of a toplevel. If the parent changed from its previous
+ * value, also sends a parent event to all clients that hold handles to
+ * both toplevel and parent (no message is sent to clients that have
+ * previously destroyed their parent handle). NULL is allowed as the
+ * parent, meaning no parent exists. */
+void wlr_foreign_toplevel_handle_v1_set_parent(
+	struct wlr_foreign_toplevel_handle_v1 *toplevel,
+	struct wlr_foreign_toplevel_handle_v1 *parent);
+
 
 #endif
--- a/types/wlr_foreign_toplevel_management_v1.c
+++ b/types/wlr_foreign_toplevel_management_v1.c
@@ -9,7 +9,7 @@
 #include "util/signal.h"
 #include "wlr-foreign-toplevel-management-unstable-v1-protocol.h"
 
-#define FOREIGN_TOPLEVEL_MANAGEMENT_V1_VERSION 2
+#define FOREIGN_TOPLEVEL_MANAGEMENT_V1_VERSION 3
 
 static const struct zwlr_foreign_toplevel_handle_v1_interface toplevel_handle_impl;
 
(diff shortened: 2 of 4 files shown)


dkondor:
Related PR for wayfire: https://github.com/WayfireWM/wayfire/pull/769

include/wlr/types/wlr_foreign_toplevel_management_v1.h
```diff
@@ -126,4 +127,14 @@ void wlr_foreign_toplevel_handle_v1_set_activated(
 void wlr_foreign_toplevel_handle_v1_set_fullscreen(
 	struct wlr_foreign_toplevel_handle_v1* toplevel, bool fullscreen);
 
+/* Send a parent change event to all clients that hold a handle to
```
ammen99, line 130: I wouldn't say `send a parent change` because we are setting the toplevel parent. The change is also sent to any clients which come afterwards (so it should be `set` not `send`)

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -405,6 +405,28 @@ void wlr_foreign_toplevel_handle_v1_set_fullscreen(
 	toplevel_send_state(toplevel);
 }
 
+static void send_parent(struct wlr_foreign_toplevel_handle_v1 *parent,
+		struct wl_resource *toplevel_resource) {
+	struct wl_client *client = wl_resource_get_client(toplevel_resource);
+	struct wl_resource *parent_resource = NULL;
+	if(parent) parent_resource = wl_resource_find_for_client(&parent->resources, client);
```
ammen99, line 412: Style issue: use `{}`

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -405,6 +405,28 @@ void wlr_foreign_toplevel_handle_v1_set_fullscreen(
...
+	/* note: parent_resource will be NULL if this client does not hold
+	 * a handle for the parent or if a NULL parent was supplied */
+	zwlr_foreign_toplevel_handle_v1_send_parent(toplevel_resource,
```
ammen99, line 415: Do we really want null? https://github.com/swaywm/wlr-protocols/pull/52#issuecomment-705080313

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -560,9 +599,12 @@ static void foreign_toplevel_manager_bind(struct wl_client *client, void *data,
 	wl_list_insert(&manager->resources, wl_resource_get_link(resource));
 
 	struct wlr_foreign_toplevel_handle_v1 *toplevel, *tmp;
+	wl_list_for_each_safe(toplevel, tmp, &manager->toplevels, link) {
```
ammen99, line 602: I think it would be nice having a comment to explain why we do it in two steps.

ammen99 (changes requested):
See comments.

types/wlr_foreign_toplevel_management_v1.c
dkondor: Not necessarily; at least I don't have a preference for either, I'll update this to only send NULL if the caller explicitly supplies a NULL parent.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -405,6 +405,37 @@ void wlr_foreign_toplevel_handle_v1_set_fullscreen(
 	toplevel_send_state(toplevel);
 }
 
+static void send_parent(struct wlr_foreign_toplevel_handle_v1 *parent,
+		struct wl_resource *toplevel_resource) {
+	struct wl_client *client = wl_resource_get_client(toplevel_resource);
+	struct wl_resource *parent_resource = NULL;
+	if(parent) {
```
ammen99, line 412: nitpick: space between `if` and `(` ...

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -405,6 +405,37 @@ void wlr_foreign_toplevel_handle_v1_set_fullscreen(
...
+		parent_resource = wl_resource_find_for_client(&parent->resources, client);
+		if(!parent_resource) {
```
ammen99, line 414: same

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -405,6 +405,37 @@ void wlr_foreign_toplevel_handle_v1_set_fullscreen(
...
+		}
+	}
+	zwlr_foreign_toplevel_handle_v1_send_parent(toplevel_resource,
+		parent_resource);
+}
+
+void wlr_foreign_toplevel_handle_v1_set_parent(
+		struct wlr_foreign_toplevel_handle_v1 *toplevel,
+		struct wlr_foreign_toplevel_handle_v1 *parent) {
+	if(parent == toplevel->parent)
```
ammen99, line 426: same, and brace must be on the same line.

ammen99 (approved):
LGTM once the style issues are fixed.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -432,6 +462,19 @@ void wlr_foreign_toplevel_handle_v1_destroy(
...
 
+	/* need to ensure no other toplevels hold a pointer to this one as
+	 * a parent, so that a later call to foreign_toplevel_manager_bind()
+	 * will not result in a segfault */
+	struct wlr_foreign_toplevel_manager_v1 *manager = toplevel->manager;
+	struct wlr_foreign_toplevel_handle_v1 *tl, *tmp3;
+	wl_list_for_each_safe(tl, tmp3, &manager->toplevels, link) {
+		if (tl->parent == toplevel) {
+			/* Note: we do not send a parent signal to the client, that
+			 * should be done by the caller if it wishes */
```
emersion, line 473: Hmm. This won't work because the `parent == toplevel->parent` check in `wlr_foreign_toplevel_handle_v1_set_parent` will bail out early.

I think it'd make sense to just send the event here.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -432,6 +462,19 @@ void wlr_foreign_toplevel_handle_v1_destroy(
 
 	wl_list_remove(&toplevel->link);
 
+	/* need to ensure no other toplevels hold a pointer to this one as
+	 * a parent, so that a later call to foreign_toplevel_manager_bind()
+	 * will not result in a segfault */
+	struct wlr_foreign_toplevel_manager_v1 *manager = toplevel->manager;
```
emersion, line 468: This assignment can be omitted.

types/wlr_foreign_toplevel_management_v1.c
```diff
@@ -405,6 +405,36 @@ void wlr_foreign_toplevel_handle_v1_set_fullscreen(
 	toplevel_send_state(toplevel);
 }
 
+static void send_parent(struct wlr_foreign_toplevel_handle_v1 *parent,
+		struct wl_resource *toplevel_resource) {
```
emersion, line 409: The order of these arguments is a little bit misleading, I'd prefer having the `wl_resource` first and then the parent. And maybe we can replace the `wl_resource` argument with a proper `wlr_foreign_toplevel_handle_v1`, to avoid mixing resource types up.

emersion (changes requested):
Apart from this, looks fine to me.

emersion:
If possible, please update our test client in `examples/`.

dkondor:
HI,
thanks, I've addressed the review comment and also updated the example client. Let me know if there's something more (and also, if I should squash everything into one commit).

emersion (approved):
LGTM, thanks!

PR #2422, opened 2020-10-09
layer-shell: error on 0 dimension without anchors

ifreund:
The protocol requires clients to set opposing anchors when requesting
a width or height of 0.

The goal of this patch is not to break clients that rely on this
behavior but to improve the consistency of the layer shell ecosystem
through adherence to the protocol.


emersion (approved):
LGTM, thanks!

Issue #1278, opened 2018-10-03
Support GPU hotplug


emersion:
This means keep scanning for GPU nodes with `udev` and create/destroy subbackends when they show up or go away.

Oliph:
I would love to see egpu hotplug support. I will keep sending donation for that in the following months. (sorry if it is not the place for such comment)

ascent12:
I acquired a USB displaylink connector not too long ago, which for all intents and purposes acts as a separate GPU, just lacking its own rendering capabilities.
Once I get around to getting that to work, it would solve this, but it's not something I'm actively working on right now.

Ongy:
We did chat wit someone at XDC who was working on getting it into gnome (collabora employee iirc), we can probably use some of that work.

Iirc a lot of the weirdness is about picking the "best" main gpu

Also something I saw at work: some of the Wacom tablets ( the one I saw was a businessy signing pad with display ) use display link
So this has more of a usecase than immediately obvious

OTOH, the kernel side of displaylink isn't exactly the nicest IMO.

Oliph:
I have seen that a couple of month ago on Phoronix where they talk about that work in mutter https://www.phoronix.com/scan.php?page=news_item&px=GNOME-Mutter-GPU-Hotplug.

They also link to the merge: https://gitlab.gnome.org/GNOME/mutter/commit/ad7d6e4a37a6258a5de876a85859f7f57415dffa

AndreasBackx:
@emersion when I hotplug my MST monitors, they show up with different names and so are just empty screens that my configuration does not work for (though they do work and sway uses them). I was gonna file an issue, though I'm unsure whether it belongs to this issue. I have the same issue with i3 so it might be an issue above wlroots and i3, no? Happy to provide logs or other information.

emersion:
MST is unrelated to GPU hotplug, so this would be a different issue. If it also happens on i3, it's probably a graphics driver issue.

AndreasBackx:
@emersion thank you, I'll file an issue with for the amdgpu stuff.

ascent12:
Recently #1696 was merged as a workaround, but we eventually want to add proper support to hotplugging DRM devices. It's particularly interesting for things like docking stations or other things which are putting monitors over USB 3 or thunderbolt, but also applies to external GPUs.

### Just as a little bit of background

There are really 2 types of DRM devices: **display controllers** and **rendering devices**.

[...]

PR #2423, opened 2020-10-10
GPU hotplugging - loads/unloads based upon udev events

neon64:
(this description is mostly copied from the main commit message)

This has been implemented specifically for the use case of Nvidia
Optimus laptops with display ports hardwired to the dGPU. I'm hoping it
may prove useful for other use-cases as well, such as hotplugging
external GPUs.

The implementation is not pretty, and likely is full of bugs and
special-cases. I am not familiar with the internal
architecture of wlroots, so would love any constructive criticism of the
implementation / C style.

The status quo:
- with both i915 and nouveau drivers loaded, sway can use both the laptop
  monitor AND external monitors, enabling and disabling outputs as
  needed
- BUT, the `nouveau` driver always stays loaded, and so my laptop idles
  at 20W instead of 3-5W :(
- unloading `nouveau` requires restarting sway. This is annoying.

After this patch:
- run `sudo modprobe nouveau` while sway is running - wlroots will
  detect the udev event and add a new drm backend and then add it to the multi
  backend.
- unplug all outputs from the dGPU, and sway will destroy the dGPU drm
  backend.
  You can verify that the nouveau driver is no longer in use,
  by running `lsof /dev/dri/card*`
- The user must still manually run `sudo rmmod nouveau` and
  then also use `bbswitch`, to fully power off the dGPU and achieve
  power-saving gains. Though still somewhat manual, this whole process can
  be done WITHOUT shutting down sway, which is awesome for productivity (no longer have
  to restart the swaywm several times throughout the day).

The future (perhaps a separate script):
- Automatically loads `nouveau` when a cable is connected (unsure if
  this is possible due to hardwiring of HDMI port, see below)
- Automatically unloads `nouveau` and runs bbswitch after all monitors
  have been disconnected for a while.
- This would be BLISS - basically as automatic as Windows when it comes to
  using Nvidia Optimus configuration - except of course that sway won't
  actually *render* from the dGPU. For my workloads, this isn't a problem,
  but perhaps someone will take this further and reimplement something
  like bumblebee but for Wayland. Is that even a good idea? I don't know.

Note: at the moment, this driver unloading only occurs when you unplug all monitors from a given card. This results in some undesirable behaviour (https://github.com/swaywm/wlroots/issues/1278#issuecomment-705874280)
Perhaps there should be an explicit command, e.g.: "unload
all unused drm backends" that can be called by some other script, which
will trigger unloading. Or perhaps, if a drm backend is unused for a
given amount of time (e.g.: a few minutes) it will get unloaded.

Also note, at least on my laptop, there seemingly no way to tell when a
cable is plugged into the HDMI port (probably because it is hardwired to
the Nvidia GPU). If this were possible, then you could write a script
which will automatically power on the nvidia GPU when an external
monitor is connected. Currently, you need to do this manually, with `sudo
modprobe nouveau`.

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -20,6 +20,7 @@
 
 #if WLR_HAS_X11_BACKEND
 #include <wlr/backend/x11.h>
+#include <include/backend/drm/drm.h>
 #endif
 
 void wlr_backend_init(struct wlr_backend *backend,
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -307,6 +307,7 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display,
 		wlr_backend_destroy(backend);
 		return NULL;
 	}
+	wlr_multi_backend_init_gpu_hotplug(multi, primary_drm);
 
 	return backend;
 
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -3,18 +3,15 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 #include <wayland-server-core.h>
 #include <wlr/backend/interface.h>
 #include <wlr/backend/session.h>
 #include <wlr/interfaces/wlr_output.h>
-#include <wlr/render/egl.h>
 #include <wlr/types/wlr_list.h>
 #include <wlr/util/log.h>
 #include <xf86drm.h>
 #include "backend/drm/drm.h"
 #include "util/signal.h"
-#include <backtrace.h>
 #include <include/wlr/backend/multi.h>
 
 struct wlr_drm_backend *get_drm_backend_from_backend(
@@ -27,28 +24,6 @@ static bool backend_start(struct wlr_backend *backend) {
 	struct wlr_drm_backend *drm = get_drm_backend_from_backend(backend);
 	scan_drm_connectors(drm);
 	return true;
-}
-
-static int bt_callback(void *data, uintptr_t pc,
-		const char *filename, int lineno,
-		const char *function) {
-	wlr_log(WLR_DEBUG, "Backtrace: %s in %s:%d", function, filename, lineno);
-	return 0;
-}
-
-static void bt_error(void *data, const char *msg,
-		int errnum) {
-	wlr_log(WLR_ERROR, "Backtrace error%s", msg);
-}
-
-static void print_trace(void) {
-	wlr_log(WLR_DEBUG, "Trying to obtain a backtrace");
-
-	struct backtrace_state *s = backtrace_create_state(NULL, 0,bt_error, NULL);
-
-	wlr_log(WLR_DEBUG, "Tried to obtain a backtrace");
-
-	backtrace_full(s, 0, bt_callback, bt_error, NULL);
 }
 
 static void backend_destroy(struct wlr_backend *backend) {
@@ -115,10 +90,10 @@ static void handle_add_gpu(struct wl_listener* listener, void *data) {
 	struct wlr_event_add_gpu *event = data;
 
 	wlr_log(WLR_INFO, "parent drm fd is %d", drm->fd);
-
 	wlr_log(WLR_INFO, "got handle_gpu signal with fd = %d", event->gpu_fd);
 
-	// TODO:
+	// TODO: create_renderer_func should not be forced NULL
+	//       even though this currently works in sway, it may not work in general.
 	struct wlr_backend *child_drm = wlr_drm_backend_create(drm->display, drm->session,
 						event->gpu_fd, &drm->backend, NULL);
 
@@ -131,10 +106,10 @@ static void handle_add_gpu(struct wl_listener* listener, void *data) {
 
 	fprintf(stderr, "is multi? %d\n\n\n", wlr_backend_is_multi(&drm->backend));
 
-	if(!wlr_multi_backend_add(&drm->multi->backend, child_drm)) {
-		wlr_log(WLR_INFO, "Failed to add to multi backend");
+	if (!wlr_multi_backend_add(&drm->multi->backend, child_drm)) {
+		wlr_log(WLR_ERROR, "Failed to add new drm backend to multi backend");
 	} else {
-		wlr_log(WLR_INFO, "Added to multi backend");
+		wlr_log(WLR_ERROR, "NOTICE: successfully new drm backend to multi backend");
 	}
 }
 
@@ -197,23 +172,12 @@ static void drm_invalidated(struct wl_listener *listener, void *data) {
 	seen_len = wl_list_length(&drm->outputs);
 	wlr_log(WLR_DEBUG, "%ld outputs after scan", seen_len);
 
-	// TODO: only destroy backend if we find that all drm_connectors have been lost ?
-	// i.e.: have unplugged all monitors
-
 	bool should_destroy = !was_all_disconnected && is_all_disconnected;
 	wlr_log(WLR_INFO, "Should we destroy the DRM backend? %s - %d", name, should_destroy);
 	free(name);
-	
-	if(drm->parent != NULL) {
-		name = drmGetDeviceNameFromFd2(drm->parent->fd);
-		wlr_log(WLR_INFO, "Parent drm backend: %p %s", drm->parent, name);
-		free(name);
-	}
-
-	print_trace();
 
 	if(should_destroy) {
-		wlr_log(WLR_INFO, "Destroying DRM backend anyway... hehe");
+		wlr_log(WLR_ERROR, "NOTICE: All outputs disconnected, destroying drm backend");
 		// this is what we added - to destroy the backend when drm is invalidated...
 		wlr_backend_destroy(&drm->backend);
 	}
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -31,7 +31,7 @@ static void backend_destroy(struct wlr_backend *backend) {
 		return;
 	}
 
-	wlr_log(WLR_INFO, "Destroying DRM BACKEND!!!");
+	wlr_log(WLR_INFO, "destroying DRM backend");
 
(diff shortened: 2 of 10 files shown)


J0nnyMak0:
One other point, I think `DRI_PRIME=1` should automatically load nouveau.

neon64:
@J0nnyMak0 thank you once again for testing this, I have a few thoughts below, but no concrete solutions as of yet

> First off, if I first plugin external monitors and then sudo modprobe nouveau, then it works great. The scan is completed and my external monitors come to life. Now I can unplug the monitors and sudo rmmod nouveau and everything works as expected. I then proceed to switch off the card with bbswitch. All good.
> 
> However, if I load nouveau module without external monitors connected, then there is a problem. Nouveau is loaded as expected, but then when I go to rmmod nouveau, it fails with: rmmod: ERROR: Module nouveau is in use

I think this is expected given the current implementation. Currently, the only time a backend will be destroyed is if the backend transitions from having some output enabled, to no outputs enabled. That certainly isn't ideal if you have no external monitors to plug in-and-out. I wasn't sure that my original implementation (unload when 0 outputs are enabled) was ideal either, because you could load nouveau without any monitors connected, and sway would immediately destroy the backend again. Then if you connected a monitor, you'd have to `rmmod nouveau` and `modprobe nouveau` again to get it to recognise.

I'm not sure what the most robust way to approach this is. Sometimes, you may not care about unloading/loading drivers, e.g.: if your particular drivers already do fantastic power management. Perhaps it should be become a config option that could be e.g.: exposed in sway's config?

> One other point, I think `DRI_PRIME=1` should automatically load nouveau.

From what I understand, `DRI_PRIME=1` is an env-var which can be given to individual applications, that forces them to use the dGPU? I'm not entirely sure how to get wlroots involved in that process. I might try it out for myself on e.g.: `glxgears`

emersion:
Thanks for working on this!

IMHO, it's not great to destroy the DRM sub-backend when all outputs are disconnected. Instead, I'd much prefer to listen to udev events. This will be what we'll need for dGPU that can be hot-unplugged by the user as well (e.g. via USB-C).

You can tell udev to trigger a unplug event for a GPU like so:

```
# MINOR=0 triggers removal for card0
udevadm trigger --verbose --type=devices --action=remove --subsystem-match=drm --property-match="MINOR=0"
```

Then wlroots can listen to these events, and destroy the sub-backend appropriately.

J0nnyMak0:
Ok, thanks! Also how are you monitoring the power state of dGPU with nouveau? For some reason powertop doesn't seem to be telling me the truth about dGPU state...

neon64:
Thanks @emersion for the advice, I've changed it now to listen to udev "remove" events as well. Now most of the code for handling GPU hotplugging is inside of the `multi/backend.c`. I'm not sure if that's the best place to put it though, seeing as the multi backend previously didn't handle specifics like GPUs etc...

> Ok, thanks! Also how are you monitoring the power state of dGPU with nouveau? For some reason powertop doesn't seem to be telling me the truth about dGPU state...

Have you tried `cat /proc/acpi/bbswitch`? It seems to work insofar as when bbswitch says the dGPU is off, power consumption drops significantly.

J0nnyMak0:
@neon64 thanks for the pointer. I'm trying to get some useful info about what I'm observing. It's a shame powertop is unreliable. The problem is when I use this branch to unload nouveau and turn off dGPU, my laptop seems to run hotter with fans and run out of battery faster than if I keep nouveau loaded and dGPU powered up. Powertop reports 100% usage of dGPU but no power reading, so I don't trust it. And bbswitch reports dGPU is off...

neon64:
Hmm that is fishy. May I ask which generation GPU it is? I only have a GTX 1050-Ti to test on

J0nnyMak0:
> Hmm that is fishy. May I ask which generation GPU it is? I only have a GTX 1050-Ti to test on

Its a Quadro K3100M

emersion:
>Now most of the code for handling GPU hotplugging is inside of the multi/backend.c. I'm not sure if that's the best place to put it though, seeing as the multi backend previously didn't handle specifics like GPUs etc...

Right, the multi backend isn't supposed to handle this. Instead, we should probably move the udev stuff to `backend/drm/backend.c` (we can grab the `struct udev` from the `wlr_session` there). The libinput backend handles its own udev event dispatching as well (via `libinput_udev_create_context`).

neon64:
So I tried to have a read of how libinput does it (https://cgit.freedesktop.org/wayland/libinput/tree/src/udev-seat.c). Are you suggesting completely getting rid of all the udev stuff from `session/session.c` (including the existing handler that seems to invoke the `drm_invalidated` function), and then creating a fresh `struct udev_monitor` for each drm backend with its own separate event handler which (a) calls `drm_invalidated` on a relevant "change" event, and (b) destroys the backend on an appropriate "remove" event?

Also from what I understand of the wlroots architecture, in order to add a new drm backend, you need access to the multi backend from somewhere to be able to call `wlr_multi_backend_add`, and also need access to the primary drm backend to make it the "parent backend". Do you have a recommendation on where to store that information? Originally I had a pointer to the `wlr_multi_backend` inside `struct wlr_drm_backend`, but that seemed far from ideal.

I've also come across some errors when destroying the backend after the "remove" udev event (also got some crashes / kernel panics on exit which I assume are related https://github.com/neon64/wlroots/issues/1). This didn't occur before when unloading after monitor disconnect.

```
00:00:05.296 [ERROR] [backend/drm/legacy.c:45] HDMI-A-1: failed to set DPMS property: Permission denied
00:00:05.339 [ERROR] [backend/session/logind.c:122] Failed to release device '9': Device not taken
```

emersion:
>completely getting rid of all the udev stuff from session/session.c (including the existing handler that seems to invoke the drm_invalidated function)

Ah, right, I forgot that `session.c` had some DRM-specific stuff:

- `wlr_session_find_gpus` being there
- The monitoring of the "change" events is also somewhat DRM-specific

>Also from what I understand of the wlroots architecture, in order to add a new drm backend, you need access to the multi backend from somewhere to be able to call wlr_multi_backend_add, and also need access to the primary drm backend to make it the "parent backend".

Yes, this is correct. We're somewhat hitting some wlroots API limitations here.

Some notes:

- `wlr_session` doesn't have access to the multi backend, and I'd prefer to keep it that way
- `wlr_multi_backend` shouldn't do DRM-specific stuff
- It would be natural to add the new DRM udev monitoring to `wlr_session` and expose add/remove as events

I can't find the issue, but there were discussions about either:

- Having the DRM backend being able to manage multiple devices without having to resort to multiple sub-backends (like [glider](https://github.com/emersion/glider/blob/619cf43a630f08998a9f7a9e328996342a29dfd1/include/backend/backend.h#L119) does)
  - Maybe doesn't give enough control to the compositor
  - Maybe would be too monolithic
- Making the compositor responsible for creating and destroying DRM backends
  - Would allow compositors to decide how DRM devices are used, which is a good thing
  - Need to think about how buffer sharing between backends should work (the trend is to remove any renderer bits from the DRM backends, but we can probably defer this work for now)
  - Maybe provide a helper that handles most of that complexity (e.g. takes a `wlr_multi_backend` plus a `wlr_session`, and creates the DRM sub-backends as necessary)

cc @ascent12

emersion:
Some of the work I described in my previous comment is done in https://github.com/swaywm/wlroots/pull/2465.

neon64:
Thanks for the PR refractoring all that stuff @emersion! I've just come out of an exam period so have been away from this for a while, I'll have a go rebasing off your branch and see what happens.

u6bkep:
I would also like add a comment for consideration by the devs here. This kind of dynamic unloading of GPU drivers is perfect for passing a GPU through to a VM for gaming without having a second discreet gaming capable GPU. The ideal case would be to run the desktop on an integrated or other cheap GPU, and have a second GPU that can be easily detached and passed to a VM, then reattached without rebooting for accelerated workloads on the host.
I know that supporting this fringe workload is probably not a priority, but the dozen or so of us with this use case would greatly appreciate it if a few moments thought where put into it. for example, allowing the user to trigger the attaching and detaching easily without having to do something stupid like unplug the monitor.

emersion:
[This comment](https://github.com/swaywm/wlroots/pull/2423#issuecomment-706545379) describes a way to tell wlroots to stop using a device from the CLI.

emersion:
https://github.com/swaywm/wlroots/pull/2575 adds support for GPU unplug.

emersion:
> Maybe provide a helper that handles most of that complexity (e.g. takes a wlr_multi_backend plus a wlr_session, and creates the DRM sub-backends as necessary)

As noted above, this is the other half of the work. This can be implemented as a private helper for now that `attempt_drm_backend` would create. Maybe something like this:

```c
struct wlr_drm_backend_monitor {
    struct wlr_backend *multi;
    struct wlr_backend *primary;
    struct wlr_session *session;

    struct wl_listener multi_destroy;
    struct wl_listener primary_destroy;
    struct wl_listener session_destroy;
    struct wl_listener session_add_drm_card;
};

struct wlr_drm_backend_monitor *wlr_drm_backend_monitor_create(struct wlr_backend *multi, struct wlr_session *session);
```

J0nnyMak0:
I hope to test this a little more soon, but I think this branch needs to be rebased first with all the changes to wlroots.

neon64:
You're right, I've let this branch get very stale - I've tried building wlroots master a couple of times in recent weeks but have been hit by a number of bugs/crashes/performance issues with multi GPU including https://github.com/swaywm/wlroots/issues/2526, so I haven't rewritten GPU hotplugging yet because I'd just be hotplugging a blank screen! Hopefully in the new year I will come back and finish the missing part emersion has outlined.

neon64:
So I've created a new PR https://github.com/swaywm/wlroots/pull/2657, based upon emersion's https://github.com/swaywm/wlroots/pull/2575 and rebased onto the latest master.

Rather large caveat: it doesn't really work at all (at least for me) - see the PR description and https://github.com/neon64/wlroots/issues/2

PR #2426, opened 2020-10-11
Using %m in `wlr_log` is broken, use `wlr_log_errno` / `perror` instead

Xyene:
This one was awful to track down, but calls to `wlr_log` with `%m` have
the errno masked by the `isatty` call in `log_stderr`. Switch them to
`wlr_log_errno` instead.

Cue quality "how can `read(2)` POSSIBLY be returning `ENOTTY`??" moments :)

While at it, I replaced the remaining usages of `%m` (in examples/) with
`perror`, since `%m` is a GNU extension.


emersion (approved):
Good catch, thanks!

Issue #2192, opened 2020-05-11
Pasting in Emacs causes wlroots to run out of pipes


mssdvd:
* Step to reproduce
1) Open emacs
2) Paste some text from the primary clipboard
4) Emacs and any other GUI application crash
After that I can't open any applications but sway still works.

When I use my usual config the first thing to crash is redshift and after a few seconds every window crash. 

* Sway Version:
sway version 1.4-a8a6ed66 (May 11 2020, branch 'master')

* Debug Log:
https://gist.github.com/mssdvd/d00ef14c7df092dfed9c582abd8aedc0

* Configuration File:
````
bindsym Mod4+c exec clipman pick -t rofi
exec wl-paste -t text --watch clipman store
include /etc/sway/config
include /etc/sway/config.d/*
````
* Stack Trace:
None because Sway didn't crash

emersion:
Can you try wlroots master and see if this is fixed by https://github.com/swaywm/wlroots/pull/2428?

mssdvd:
I tried it and the system freezed (reboot required). I'll make a more detailed
report later.

mssdvd:
I'm having difficulty producing a decent size log. The only time that the system has behaved badly the log was about 140MB. That time the system behaved like I described before. All other times the system did not respond to inputs.

emersion:
Consider using SSH to investigate the frozen machine.

Xyene:
Could you try wlroots master again? A lot of Xwayland clipboard fixes went in a few days ago, and I think this may have been fixed.

mssdvd:
Unfortunately it keeps crashing. 

Log: https://gist.github.com/mssdvd/7ebf06bb55885d61f45fc13c5b5f6faf

Xyene:
Thanks for testing. It's now failing in a different way than before. In your first log, the requestor window was the same in each attempt; here, Emacs is creating a new window for each attempt. This sounds like an Emacs bug... https://github.com/swaywm/wlroots/issues/2326 would prevent the fd exhaustion, but not fix Emac's behavior here.

Some questions to try to get to the bottom of this:

* What version of Emacs are you running?
* Is running `wl-paste` still a requirement in reproducing?
* Does the issue reproduce with `wl-paste`/`clipman` master?
* Does the issue reproduce if the primary selection owner is another X11 window?
* Does the issue reproduce if the primary selection owner is a Wayland window?
[...]

PR #2428, opened 2020-10-12
xwayland: remove stale transfers from the same requestor

Xyene:
It seems that if we ever try to reply to a selection request after
another has been sent by the same requestor (we reply in FIFO order),
the requestor never reads from it, and we end up stalling forever on a
transfer that will never complete.

It appears that `XCB_SELECTION_REQUEST` has some sort of singleton
semantics, and new requests for the same selection are meant to replace
outstanding older ones. I couldn't find a reference for this, but
empirically this does seem to be the case.

Real (contrived) case where we don't currently do this, and things break:

* run fcitx
* run Slack
* wl-copy < <(base64 /opt/firefox/libxul.so)  # or some other large file
* focus Slack (no need to paste)

fcitx will send in an `XCB_SELECTION_REQUEST`, and we'll start
processing it. Immediately after, Slack sends its own. fcitx hangs for a
long, long time. In the meantime, Slack retries and sends another
selection request (with a different mimetype). We now have two
pending requests from Slack.

Eventually fcitx gives up (or it can be `pkill`'d), and we start
processing the first request Slack gave us (FIFO). Slack (Electron?)
isn't listening on the other end anymore, and this transfer never
completes. The X11 clipboard becomes unusable until Slack is killed.

After this patch, the clipboard is immediately usable again after fcitx
bails. Also added a bunch of debug-level logging that makes diagnosing
this sort of issue easier.

Refs swaywm/sway#4007.

### DIFF
--- a/xwayland/selection/outgoing.c
+++ b/xwayland/selection/outgoing.c
@@ -71,9 +71,11 @@ static void xwm_selection_transfer_destroy_outgoing(
 	struct wlr_xwm_selection *selection = transfer->selection;
 	bool was_first = transfer == xwm_selection_transfer_get_first(selection);
 	wl_list_remove(&transfer->outgoing_link);
+	wlr_log(WLR_DEBUG, "Destroying transfer %p", transfer);
 
 	// Start next queued transfer if we just removed the active one.
 	if (was_first && !wl_list_empty(&selection->outgoing)) {
+		wlr_log(WLR_DEBUG, "Destroyed transfer was active, starting next");
 		xwm_selection_transfer_start_outgoing(
 			xwm_selection_transfer_get_first(selection));
 	}
@@ -230,6 +232,8 @@ static void xwm_selection_transfer_start_outgoing(
 	struct wlr_xwm *xwm = transfer->selection->xwm;
 	struct wl_event_loop *loop =
 		wl_display_get_event_loop(xwm->xwayland->wl_display);
+	wlr_log(WLR_DEBUG, "Starting transfer %p", transfer);
+	assert(transfer == xwm_selection_transfer_get_first(transfer->selection));
 	transfer->source = wl_event_loop_add_fd(loop, transfer->source_fd,
 		WL_EVENT_READABLE, xwm_data_source_read, transfer);
 }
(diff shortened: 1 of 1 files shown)


Xyene:
Tentatively I want to say this also would close https://github.com/swaywm/wlroots/issues/2192. The [log](https://gist.github.com/mssdvd/d00ef14c7df092dfed9c582abd8aedc0) there is full of

```
00:00:23.525 [DEBUG] [xwayland/selection/outgoing.c:302] Sending Wayland selection 254 to Xwayland window with MIME type text/plain;charset=utf-8, target 254
00:00:23.526 [DEBUG] [xwayland/selection/outgoing.c:366] XCB_SELECTION_REQUEST (time=0 owner=2097153, requestor=2097153 selection=272, target=254, property=273)
00:00:23.526 [DEBUG] [xwayland/selection/outgoing.c:302] Sending Wayland selection 254 to Xwayland window with MIME type text/plain;charset=utf-8, target 254
00:00:23.526 [DEBUG] [xwayland/selection/outgoing.c:366] XCB_SELECTION_REQUEST (time=0 owner=2097153, requestor=2097153 selection=272, target=254, property=273)
```

Since we'd be replacing the requests (`requestor` is the same in all of these), we wouldn't run out of pipes anymore. Not sure what'd happen to Emacs, but in the worst case, a frozen Emacs feels better than a frozen Sway :)

xwayland/selection/outgoing.c
```diff
@@ -312,14 +316,34 @@ static void xwm_selection_send_data(struct wlr_xwm_selection *selection,
...
+		req->target, transfer);
 	xwm_selection_source_send(selection, mime_type, p[1]);
 
+	// It seems that if we ever try to reply to a selection request after
+	// another has been sent by the same requestor, the requestor never reads
+	// from it. It appears to only ever read from the latest, so purge stale
+	// transfers to prevent clipboard hangs.
+	struct wlr_xwm_selection_transfer *outgoing, *tmp;
+	wl_list_for_each_safe(outgoing, tmp, &selection->outgoing, outgoing_link) {
+		if (transfer->request.requestor == outgoing->request.requestor) {
```
emersion, line 329: Side note: maybe it would be a good idea to filter by other `xcb_selection_request_event_t` fields if this turns out not being narrow enough.

emersion (approved):
Thanks!

PR #2430, opened 2020-10-13
xwayland: notify requestor when we fail to respond to their request

Xyene:
We already mostly did this, but there were a couple of branches
(`calloc` failures) where we'd bail without letting the other side know.

Refs swaywm/sway#4007. Likely not going to be a real improvement there
(if `calloc` fails you're already pretty screwed), but it does address a
theoretical possibility.


emersion (approved):
Thanks!

PR #2431, opened 2020-10-14
xwayland: add set_geometry event

ammen99:
This is necessary to react to changes in position of override-redirect views.
Example of where this is necessary is dragging the "Search everywhere" dialog in Android Studio. It configures itself in order to move around the screen, but compositors up to now can only react when the surface is committed. In the meantime, the compositor and the Xserver state is out-of-sync causing jumps of the view because the mouse coordinates are wrong.

See also https://github.com/WayfireWM/wayfire/pull/784

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -198,6 +198,7 @@ struct wlr_xwayland_surface {
 		struct wl_signal set_hints;
 		struct wl_signal set_decorations;
 		struct wl_signal set_override_redirect;
+		struct wl_signal set_geometry;
 		struct wl_signal ping_timeout;
 	} events;
 
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -164,6 +164,7 @@ static struct wlr_xwayland_surface *xwayland_surface_create(
 	wl_signal_init(&surface->events.set_decorations);
 	wl_signal_init(&surface->events.set_override_redirect);
 	wl_signal_init(&surface->events.ping_timeout);
+	wl_signal_init(&surface->events.set_geometry);
 
 	xcb_get_geometry_reply_t *geometry_reply =
 		xcb_get_geometry_reply(xwm->xcb_conn, geometry_cookie, NULL);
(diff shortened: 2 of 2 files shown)


include/wlr/xwayland.h
```diff
@@ -199,6 +199,7 @@ struct wlr_xwayland_surface {
 		struct wl_signal set_decorations;
 		struct wl_signal set_override_redirect;
 		struct wl_signal ping_timeout;
+        struct wl_signal configure_notify;
```
emersion, line 202: Style: tabs

xwayland/xwm.c
```diff
@@ -940,6 +941,8 @@ static void xwm_handle_configure_notify(struct wlr_xwm *xwm,
 	xsurface->width = ev->width;
 	xsurface->height = ev->height;
 
+    wlr_signal_emit_safe(&xsurface->events.configure_notify, xsurface);
```
emersion, line 944: Ditto

emersion, line 944: I'd prefer to trigger this event after the event has been completely handled by wlroots, ie. at the end of the function after the `override_redirect` handling. Otherwise the compositor will see half-updated state.

emersion:
BTW, maybe this new event supersedes the `set_override_redirect` event?

xwayland/xwm.c
emersion, line 944: One more nit: so far we've been cargo-culting the fact that we always provide a data pointer for events (here, with `xsurface`). This has the unfortunate consequence of always creating a breaking change each time we want to replace it with something useful.

Can we change it to `NULL`, so that we can easily change it to something useful later (e.g. a bitflag of properties that changed)?

include/wlr/xwayland.h
ammen99: Fixed.

xwayland/xwm.c
```diff
@@ -935,15 +936,25 @@ static void xwm_handle_configure_notify(struct wlr_xwm *xwm,
 		return;
 	}
 
-	xsurface->x = ev->x;
-	xsurface->y = ev->y;
-	xsurface->width = ev->width;
-	xsurface->height = ev->height;
+	bool geometry_changed =
+		(xsurface->x != ev->x || xsurface->y != ev->y ||
+		 xsurface->width != ev->width || xsurface->height || ev->height);
```
emersion, line 941: Typo: `xsurface->height || ev->height` should read `xsurface->height != ev->height`

xwayland/xwm.c
```diff
@@ -935,15 +936,25 @@ static void xwm_handle_configure_notify(struct wlr_xwm *xwm,
...
+		xsurface->height = ev->height;
+	}
 
 	if (xsurface->override_redirect != ev->override_redirect) {
 		xsurface->override_redirect = ev->override_redirect;
 		wlr_signal_emit_safe(&xsurface->events.set_override_redirect, xsurface);
 	}
+
+	if (geometry_changed) {
+		wlr_signal_emit_safe(&xsurface->events.set_geometry, NULL);
```
emersion, line 956: Since this is only about geometry, I'm fine with emitting this before `set_override_redirect` if you want to change that back.

emersion:
Can you also update the commit message? Otherwise this looks good!

xwayland/xwm.c
ammen99: I don't think the order matters much.

emersion: Agreed.

emersion (approved):
Looks good to me, thanks!

xwayland/xwm.c
ammen99: very nice catch!

PR #2433, opened 2020-10-15
Use uint32_t in wlr_renderer_begin signature

ifreund:
This matches the signature of wlr_renderer_impl.begin


ifreund:
This is technically a breaking change but I don't think it will cause any issues as the implicit cast from `int` to `uint32_t` is currently happening, presumably without issue, inside wlroots.

ammen99:
> This is technically a breaking change but I don't think it will cause any issues as the implicit cast from `int` to `uint32_t` is currently happening, presumably without issue, inside wlroots.

The change LGTM, however, it is still a breaking change. For example, if some language uses auto-generated bindings and in that language the implicit conversion is not available.

emersion (approved):
Thanks!

PR #2434, opened 2020-10-15
Swap arg order of wlr_surface_accepts_touch()

ifreund:
This puts the wlr_surface argument first, as should be the case for
functions namespaced with wlr_surface_*.


ifreund:
This is a breaking change of course, I'd be happy to track down and send patches to compositors using this if it's deemed to be worth merging and not just bikeshedding.

emersion:
Note, `wlr_surface_accepts_tablet_v2` is also affected.

ifreund:
> Note, wlr_surface_accepts_tablet_v2 is also affected.

The commit now corrects `wlr_surface_accepts_tablet_v2()` as well.

ammen99:
> Still wondering whether this breaking change is worth it.

IMO the question is: if wlroots 1.0 were to be released, would you find the current syntax acceptable?

ifreund:
Added another commit fixing another style error in the same vein

ifreund:
Added another commit fixing the prefix of two `wlr_xwayland_surface` functions.

I still don't have a strong opinion on whether it makes sense to merge this or not, I'm just using this to keep track of these minor inconsistencies as I notice them.

include/wlr/xwayland.h
```diff
@@ -322,11 +322,11 @@ void wlr_xwayland_surface_ping(struct wlr_xwayland_surface *surface);
  * Returns: true if the window should receive focus
  *          false if it should be ignored
  */
-bool wlr_xwayland_or_surface_wants_focus(
+bool wlr_xwayland_surface_or_wants_focus(
```
emersion, line 325: While at it we might want to rename to `wlr_xwayland_surface_o_r_wants_focus` or even `wlr_xwayland_surface_override_redirect_wants_focus` to avoid the confusion with the English word "or".

ifreund: Yeah we should really just be explicit here and spell out "override redirect." I actually already did so in zig-wlroots...

emersion: Also the whole function is just an ugly hack, ideally we should get rid of it and rely on the input model only, but I'm too scared to touch X11 input stuff again.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2434

PR #2435, opened 2020-10-15
wayland: add touch support to the wayland backend

ammen99:
Goal currently is to get support working for a single output, thus there is only one touch device created.

Multi-output support is left for later.

### DIFF
--- a/backend/wayland/seat.c
+++ b/backend/wayland/seat.c
@@ -276,6 +276,95 @@ static struct wl_keyboard_listener keyboard_listener = {
 	.repeat_info = keyboard_handle_repeat_info
 };
 
+static void touch_coordinates_to_absolute(struct wlr_wl_input_device *device,
+		wl_fixed_t x, wl_fixed_t y, double *sx, double *sy) {
+	// TODO: each output needs its own touch
+	struct wlr_wl_output *output, *tmp;
+	wl_list_for_each_safe(output, tmp, &device->backend->outputs, link) {
+		*sx = wl_fixed_to_double(x) / output->wlr_output.width;
+		*sy = wl_fixed_to_double(y) / output->wlr_output.height;
+		return; // Choose the first output in the list
+	}
+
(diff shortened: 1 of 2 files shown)


backend/wayland/seat.c
```diff
@@ -276,6 +276,95 @@ static struct wl_keyboard_listener keyboard_listener = {
 	.repeat_info = keyboard_handle_repeat_info
 };
 
+static void touch_coordinates_to_absolute(struct wlr_wl_input_device *device,
+		wl_fixed_t x, wl_fixed_t y, double *sx, double *sy) {
+	// TODO: each output needs its own touch
+	struct wlr_wl_output *output, *tmp;
+	wl_list_for_each_safe(output, tmp, &device->backend->outputs, link) {
```
ony, line 283: I think `wl_list_for_each` is sufficient here since we don't modify list:
```suggestion
	struct wlr_wl_output *output;
	wl_list_for_each(output, &device->backend->outputs, link) {
```

emersion (approved):
LGTM, thanks!

backend/wayland/seat.c
```diff
@@ -563,6 +652,28 @@ void create_wl_keyboard(struct wl_keyboard *wl_keyboard, struct wlr_wl_backend *
...
+	if (!dev) {
+		return;
+	}
+
+	struct wlr_input_device *wlr_dev = &dev->wlr_input_device;
+
+	wlr_dev->touch = calloc(1, sizeof(*wlr_dev->touch));
+	if (!wlr_dev->touch) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		free(dev);
```
ony, line 667: `create_wl_input_device` does a bit more than allocation. In particular it inserts device in `backend->devices` and freeing here without removing there will lead to crash.
```suggestion
		wlr_input_device_destroy(dev);
```

emersion:
Good catches, can you submit these as a PR?

PR #2436, opened 2020-10-16
xdg_positioner: remove unused field

ifreund:
The resource field of wlr_xdg_positioner is never initialized or
accessed within wlroots. The wl_resource for this interface is stored
in the wlr_xdg_positioner_resource struct.


emersion (approved):
Thanks!

Issue #2437, opened 2020-10-17
wlr_texture_from_pixels: missing piece of documentation


werererer:
### Problem
When I use wlr_texture_from_pixels inside wlr_render_begin and wlr_render_end my program crashes after calling the next function that has something to do with rendering with the following message:
```
test: render/gles2/renderer.c:37: gles2_get_renderer_in_context: Assertion `wlr_egl_is_current(renderer->egl)' failed.
```
### Solution
- Maybe wlr_texture_from_pixels should assert when being called when it is inside wlr_render_begin/end block.
- adding Documentation about it (I will do that)

emersion:
>Maybe wlr_texture_from_pixels should assert when being called when it is inside wlr_render_begin/end block.

Yeah. Adding `assert(!r->rendering)` in `render/wlr_renderer.c` should do the trick.

werererer:
I think it is more compilcated than I initialy thought. It also crashes when I call `wlr_output_attach_render` this assert won't help in that case I think.

emersion:
`wlr_output_attach_render` needs to be called right before `begin`, so can add the assert there too I guess.

PR #2438, opened 2020-10-17
append info about the behaviour of wlr_texture_from_pixels

werererer:
See #2437

### DIFF
--- a/include/wlr/render/wlr_texture.h
+++ b/include/wlr/render/wlr_texture.h
@@ -24,6 +24,8 @@ struct wlr_texture {
 /**
  * Create a new texture from raw pixel data. `stride` is in bytes. The returned
  * texture is mutable.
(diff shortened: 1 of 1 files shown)


emersion:
I think this generally applies to all functions creating or updating a texture.

werererer:
Didnt test that but may be true. I'll test it later

werererer:
btw It seems that this issue didn't exist in prior versions couldn't reproduce it there

emersion:
Yes, this behaviour has been introduced after properly cleaning up EGL contexts behind wlroots calls.

werererer:
Is it cleanly fixable? I mean sometimes you have to introduce new surfaces inbetween wlr_render_begin() and end()

emersion:
No, you shouldn't need to do this.

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2486

PR #2441, opened 2020-10-18
Minor fixes in Wayland backend seat and input device handling alloc errors

ony:
This is partially follow on comments for https://github.com/swaywm/wlroots/pull/2413 and https://github.com/swaywm/wlroots/pull/2435 .

Note that releasing `wl_keyboard` is managed by `wlr_wl_keyboard` created in association 1:1 most of the time. With one exception when back-end is not yet started.


emersion (approved):
Thanks!

PR #2445, opened 2020-10-20
foreign toplevel: Fix whitespace error

soreau:


emersion:
Does this need to be upstreamed at wlr-protocols?

emersion (approved):
Right, this error isn't in wlr-protocols.

Thanks

PR #2446, opened 2020-10-21
Replace wlr_key_state with wl_keyboard_key_state

ifreund:
There's no reason to have duplicate enums.

* * *

Breaking change: `enum wlr_key_state` has been replaced with `enum wl_keyboard_key_state`.

### DIFF
--- a/backend/libinput/keyboard.c
+++ b/backend/libinput/keyboard.c
@@ -72,10 +72,10 @@ void handle_keyboard_key(struct libinput_event *event,
 		libinput_event_keyboard_get_key_state(kbevent);
 	switch (state) {
 	case LIBINPUT_KEY_STATE_RELEASED:
-		wlr_event.state = WLR_KEY_RELEASED;
+		wlr_event.state = WL_KEYBOARD_KEY_STATE_RELEASED;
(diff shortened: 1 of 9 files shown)


ifreund:
I've just realized that we have the same issue with wlr_button state. If this is deemed to be the direction we want to take I would be happy to make the parallel change there.

emersion:
I think the original intention was to keep apart the protocol bits and the backend interface bits. But duplicating the enums doesn't really make sense to me, since wlroots is a *Wayland* compositor library, not a generic GUI library.

ifreund:
Found another: `wlr_edge` duplicates `xdg_toplevel.resize_edge`.

Xyene:
Note that Sway, at least uses `wlr_edges` to represent resize edges of non-`xdg_toplevel` things, e.g. Xwayland containers. (It seems to be used to similar effect in `xwayland/xwm.c` etc.)

emersion (approved):
LGTM

The `enum wl_keyboard_key_state` values have the same values as `enum wlr_key_state`.

emersion:
Thanks!

PR #2448, opened 2020-10-22
export-dmabuf, screencopy: various fixes

emersion:
_See each commit._

cc @any1: does this fix the off-by-one-buffer bug on DRM?


emersion:
Closing per https://github.com/swaywm/wlroots/pull/2374#discussion_r510398382

Issue #2432, opened 2020-10-14
race condition when closing popups


dkondor:
Hi,

copying my original bug report for Wayfire (https://github.com/WayfireWM/wayfire/issues/772), since this is probably an issue with wlroots.

Steps to reproduce:

1. Build and run cairo-dock from this branch: https://github.com/dkondor/cairo-dock-core/tree/wayland_egl
2. Start wayfire and cairo-dock (cairo-dock -o -L)
3. Start any app (gnome-calculator in my example)
4. Minimize it
5. Right click on the icon
6. From the right click menu, click on the magnifying glass icon (see screenshot)

Expected result: app is restored and activated
Actual result: wayfire crashes (see below for stacktrace)

In this case, cairo-dock is using the activate request of wlr-foreign-toplevel-management.
![cd_crash](https://user-images.githubusercontent.com/6224679/95998447-727ac180-0e67-11eb-95a0-68e42f608fcc.png)

dkondor:
I've run both wayfire and cairo-dock with WAYLAND_DEBUG=1, and this is the part of the output I think is relevant:

For wayfire:
```
[4140950.033] zwlr_foreign_toplevel_handle_v1@4278190080.activate(wl_seat@11)
...
[4140953.836]  -> xdg_popup@40.popup_done()
[4140962.837] wl_surface@45.commit()
```

For cairo-dock
```
[4140949.599]  -> zwlr_foreign_toplevel_handle_v1@4278190080.activate(wl_seat@11)
...
[4140955.739]  -> wl_surface@45.commit()
[4140956.080] xdg_popup@40.popup_done()
[4140956.100]  -> xdg_popup@40.destroy()
[4140956.114]  -> xdg_surface@41.destroy()
[4140956.140]  -> wl_surface@45.destroy()
```

Here, as far as I can tell, `wl_surface@45`, `xdg_surface@41` and `xdg_popup@40` belong to the menu. From wayfire's perspective, it sends the popup_done signal in response to the activate request, but receives a commit request for the popup's wl_surface after it. From cairo-dock's perspective, the popup_done event does not arrive until after the commit request is sent.

emersion:
The crash seems like a duplicate of https://github.com/swaywm/wlroots/issues/2056

dkondor:
I've created a bit contrived example that reliably crashes Wayfire for me:
https://github.com/dkondor/wlroots/tree/issue-2432/issue-2432

[...]

PR #2453, opened 2020-10-27
xdg_shell: fix a typo

ammen99:
This could help with #2432, it seems like this overwrites undefined memory locations...


emersion (approved):
Thanks!

PR #2454, opened 2020-10-27
xdg-surface: handle inert toplevels and popups in when committing

### LINKED ISSUE #2432: race condition when closing popups

ammen99:
~~This is an attempt to~~ Fixes #2432 
@dkondor please test

### DIFF
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -336,7 +336,9 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
 		return;
 	}
 
-	if (surface->role == WLR_XDG_SURFACE_ROLE_NONE) {
+	// surface->role might be NONE for inert popups
+	// So we check surface->surface->role
+	if (surface->surface->role == NULL) {
 		wl_resource_post_error(surface->resource,
 			XDG_SURFACE_ERROR_NOT_CONSTRUCTED,
 			"xdg_surface must have a role");
@@ -361,7 +363,8 @@ void handle_xdg_surface_commit(struct wlr_surface *wlr_surface) {
 
 	switch (surface->role) {
 	case WLR_XDG_SURFACE_ROLE_NONE:
-		assert(false);
+		// inert toplevel or popup
+		return;
 	case WLR_XDG_SURFACE_ROLE_TOPLEVEL:
 		handle_xdg_surface_toplevel_committed(surface);
 		break;


dkondor:
Just tested this (this branch applied on top of latest wlroots master + wayfire master).

The crash after dismissing the dialog is gone, but there is a new crash when closing a toplevel that used to have a dialog.

This happens in my simplified example: https://github.com/dkondor/wlroots/blob/issue-2432/issue-2432/wlr-crasher-2.c

To reproduce, close the blue dialog either by clicking on it or by activating another app. After that, when the main toplevel is closed (clicking on it or even Ctrl-C from terminal), wayfire crashes.

I cannot test my original issues with cairo-dock, since trying to start it already crashes wayfire with this patch.

dkondor:
Stackrace:
```
EE 28-10-20 11:56:08.144 - [src/main.cpp:233] Fatal error: Segmentation fault
EE 28-10-20 11:56:08.183 - #1  signal_handler(int) ../src/main.cpp:234
EE 28-10-20 11:56:08.361 - #2  killpg ??:?
EE 28-10-20 11:56:08.370 - #3  wl_list_remove /home/dkondor/program/wayland/src/wayland-util.c:55
EE 28-10-20 11:56:08.383 - #4  reset_xdg_surface ../types/xdg_shell/wlr_xdg_surface.c:513
EE 28-10-20 11:56:08.397 - #5  destroy_xdg_surface ../types/xdg_shell/wlr_xdg_surface.c:517
EE 28-10-20 11:56:08.410 - #6  xdg_surface_handle_resource_destroy ../types/xdg_shell/wlr_xdg_surface.c:325
EE 28-10-20 11:56:08.418 - #7  destroy_resource /home/dkondor/program/wayland/src/wayland-server.c:726
EE 28-10-20 11:56:08.429 - #8  wl_resource_destroy /home/dkondor/program/wayland/src/wayland-server.c:743
EE 28-10-20 11:56:08.441 - #9  xdg_surface_handle_destroy ../types/xdg_shell/wlr_xdg_surface.c:309
EE 28-10-20 11:56:08.449 - #10 ffi_call_unix64 ??:?
EE 28-10-20 11:56:08.459 - #11 ffi_call ??:?
EE 28-10-20 11:56:08.467 - #12 wl_closure_invoke /home/dkondor/program/wayland/src/connection.c:1020
EE 28-10-20 11:56:08.476 - #13 wl_client_connection_data /home/dkondor/program/wayland/src/wayland-server.c:432
EE 28-10-20 11:56:08.486 - #14 wl_event_loop_dispatch /home/dkondor/program/wayland/src/event-loop.c:1027
EE 28-10-20 11:56:08.496 - #15 wl_display_run /home/dkondor/program/wayland/src/wayland-server.c:1349
EE 28-10-20 11:56:08.527 - #16 main ../src/main.cpp:394
EE 28-10-20 11:56:08.662 - #17 __libc_start_main ../csu/libc-start.c:344
EE 28-10-20 11:56:08.703 - #18 ?? ??:0
```

ammen99:
@dkondor I think I saw the problem, `reset_xdg_surface` might be called multiple times for the same toplevel/popup. I have fixed this issue, please try again.

dkondor:
Yes, it's working now; also, my original issue with cairo-dock is resolved as well. Thanks!

travankor:
Should it send a protocol error, too? See: https://github.com/swaywm/wlroots/pull/2057

ammen99:
> Should it send a protocol error, too? See: #2057

No, it shouldn't. This is simply a race condition which neither wlroots nor the client can prevent, and necessitates an inert state of objects on the wlroots side.

@travankor does this also fix your issue with chromium on Wayland?

travankor:
>does this also fix your issue with chromium on Wayland?

I can't test that anymore since the bug only worked on v80 and previous releases.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -507,6 +508,11 @@ void reset_xdg_surface(struct wlr_xdg_surface *xdg_surface) {
 	}
 
 	xdg_surface->role = WLR_XDG_SURFACE_ROLE_NONE;
+
+	wl_list_remove(&xdg_surface->surface_destroy.link);
+	wl_list_remove(&xdg_surface->surface_commit.link);
```
emersion, line 513: Doesn't this break re-mapping a surface after it has been unmapped?

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -507,6 +508,11 @@ void reset_xdg_surface(struct wlr_xdg_surface *xdg_surface) {
 	}
 
 	xdg_surface->role = WLR_XDG_SURFACE_ROLE_NONE;
+
+	wl_list_remove(&xdg_surface->surface_destroy.link);
```
emersion, line 512: Doesn't this cause a use-after-free when destroying a `wl_surface` without destroying the `xdg_surface`?

types/xdg_shell/wlr_xdg_surface.c
ammen99: I also thought this but after grepping in the source code, you can see `reset_xdg_surface` is called in three cases:

- destruction of xdg_toplevel
- destruction of xdg_popup
- destruction of xdg_surface

So this is not a problem

types/xdg_shell/wlr_xdg_surface.c
ammen99: When the `wl_surface` is destroyed, the function `xdg_surface_handle_surface_destroy` is called which then calls `reset_xdg_surface()`. This means that this code will be run when the surface is being destroyed, and will not cause a crash.

types/xdg_shell/wlr_xdg_surface.c
emersion: The client is allowed to destroy the xdg_toplevel object without destroying the xdg_surface, and then re-create the xdg_toplevel.

The `surface_commit` listener is set up in `create_xdg_surface`, I'd rather not remove it in `reset_xdg_surface`.

types/xdg_shell/wlr_xdg_surface.c
emersion: This situation can result in a use-after-free:

- `reset_xdg_surface` is called (e.g. because the `xdg_toplevel` is destroyed by the client). The `surface_destroy` listener is removed.
- The client destroys the `wl_surface`.

In this case the `xdg_surface` object is still alive, and `wlr_xdg_surface.surface` points to free'd memory.

ammen99:
I did not know that you can destroy and re-create the xdg_toplevel without doing the same for xdg_surface. Unfortunately, this means we just need to remove one of the errors for clients, let me know if you find this acceptable.

emersion:
Can we add an explanation of why this is necessary to the commit message? Otherwise I won't remember. Something among those lines:

```
xdg_popups can be destroyed by the compositor when closed. When this happens,
wlroots makes the xdg_popup surface inert and resets the xdg_surface role to
NONE.

Currently, wlroots sends a protocol error and asserts that an xdg_surface has
a role when committed. This is racy if at the same time the client commits an
xdg_popup and the compositor closes it. This patch removes the assertion and
ignores commits on xdg_surfaces without a role set.
```

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -337,9 +337,7 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
 	}
 
 	if (surface->role == WLR_XDG_SURFACE_ROLE_NONE) {
```
emersion, line 339: Maybe we can keep the protocol error, and replace the check `surface->surface->role == NULL` instead? This doesn't get reset by `reset_xdg_surface`.

ammen99: Fixed.

ammen99:
I updated the patch as you suggested. Thanks for the in-depth review, I know so little about the xdg-shell implementation.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -336,7 +336,8 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
 		return;
 	}
 
-	if (surface->role == WLR_XDG_SURFACE_ROLE_NONE) {
+	// surface->role might be NONE for inert popups
+	if (surface->surface->role == WLR_XDG_SURFACE_ROLE_NONE) {
```
emersion, line 340: `surface->surface` is the `wlr_surface`, so it has nothing to do with xdg-shell. `surface->surface->role` is a `struct wlr_surface_role *` and not an enum, so we need to check whether it's NULL.

emersion (changes requested):
>Thanks for the in-depth review, I know so little about the xdg-shell implementation.

No problem, the xdg-shell implementation isn't the most straightforward part of wlroots sadly.

types/xdg_shell/wlr_xdg_surface.c
ammen99: I am sorry, this was really stupid :/

emersion (changes requested):
Forgot to push?

emersion (approved):
Thanks for fixing this!

PR #2456, opened 2020-10-29
backend/drm: Check if output is enabled before sending frame event

maringuu:
This is my first contribution to wlroots so I'm not quite sure if this is the right place to fix the bug.
But since it is only a one line change it should be easy to review.

When an output is disabled one last pageflip will happen to disable it.
Since the output is disabled we don't want to send frame events anymore.
Currently one last event is send after the output is disabled (and the disable is committed ofc).


emersion (approved):
LGTM, thanks!

PR #2458, opened 2020-10-31
Replace raw enum flags with their equivalent bitmasks

Xyene:
This is a straightforward cleanup PR, covering all relevant enums that I could find.


emersion (approved):
Thanks!

PR #2463, opened 2020-11-04
Fix wlr_output_export_dmabuf off-by-one-frame error

emersion:
_See individual commits._

cc @any1


any1:
I tested this with both screencopy-dmabuf and regular screencopy in wayvnc, and after the change everything works as expected.

I also looked over the diffs, trying to find cases where damage handling and/or resource management might go wrong and did not find any such cases.

Issue #2466, opened 2020-11-06
Send wl_surface_enter when wl_output is bound late


kennylevinsen:
If a client has not bound a wl_output at the time where the compositor calls wlr_surface_send_enter, then no wl_surface_enter event is sent.

As the compositor is unaware of this, and no logic is present to retry when a wl_output object is finally bound by the client, this leaves the client unaware of its current output association.

A solution to this would be to have a wlr_surface_(enter|leave)_output API, where instead of blindly sending an event, we'd instead add the output to a list on the surface, allowing us to send late enter events, and possibly also to send leave events when an output disappears.

Application issue stemming from this: https://codeberg.org/dnkl/foot/issues/182

emersion:
Lazy IRC log dump of a discussion about this

```
22:50 <kennylevinsen> Hmm, wlroots is discarding the wl_surface::enter event sway is trying to send
22:51 <kennylevinsen> wlr_surface_send_enter looks for a wl_output resource for the client, finds none, and does nothing.
22:52 <emersion> eh, that's normal then
22:52 <kennylevinsen> sway moves on with its life thinking all went well of course, and thus the client is never notified of having entered an output despite being visible on one.
22:52 <emersion> the client needs to bind to the output, otherwise wlroots can't send any event
22:52 <emersion> (having no wl_output object to tie the event to)
22:53 <emersion> what we could do though is send wl_output.enter when the client binds to a wl_output
22:53 <emersion> i've been thining about a better way to handle "entered" outputs in wlroots
22:54 <emersion> iirc weston just has a per-surface list of entered outputs
22:57 <emersion> right, weston_view.output_mask
22:59 <kennylevinsen> it's normal that there's no object, but for newly appearing outputs we seem to be racing with the client, and never trying again if it fails
23:00 <kennylevinsen> so yeah - we'd need to send output enter once the client binds it then
23:07 <emersion> i wonder how (if?) we can nicely implement this in wlroots 
23:07 <emersion> wlr_surface and wlr_output are completely separate
23:09 <emersion> we need to somehow store a ref to the outputs entered in the surfaces (be it a list of a mask), but have to be careful about correctly handling output destroy
23:09 <emersion> *be it a list or a mask
23:11 <emersion> the naive approach would be to have a linked list of (wlr_output, output destroy listener) in wlr_surface, but that's not too great
[...]

PR #2468, opened 2020-11-07
Implement new surface enter/leave API

kennylevinsen:
surface: Make send_enter store entered outputs
    
wlr_surface_send_enter now stores outputs that have been entered.
Combined with a new 'bind' event on wlr_output, this allows us to delay
enter events as necessary until the respective wl_output global has been
bound.

Closes: https://github.com/swaywm/wlroots/issues/2466

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -93,6 +93,13 @@ static void output_bind(struct wl_client *wl_client, void *data,
 	send_current_mode(resource);
 	send_scale(resource);
 	send_done(resource);
+
+	struct wlr_output_event_bind evt = {
+		.output = output,
+		.resource = resource,
+	};
+
+	wlr_signal_emit_safe(&output->events.bind, &evt);
 }
 
 void wlr_output_create_global(struct wlr_output *output) {
@@ -333,6 +340,7 @@ void wlr_output_init(struct wlr_output *output, struct wlr_backend *backend,
 	wl_signal_init(&output->events.precommit);
 	wl_signal_init(&output->events.commit);
 	wl_signal_init(&output->events.present);
+	wl_signal_init(&output->events.bind);
 	wl_signal_init(&output->events.enable);
 	wl_signal_init(&output->events.mode);
 	wl_signal_init(&output->events.scale);
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -166,6 +166,8 @@ struct wlr_output {
 		struct wl_signal commit; // wlr_output_event_commit
 		// Emitted right after the buffer has been presented to the user
 		struct wl_signal present; // wlr_output_event_present
+		// Emitted after a client bound the wl_output global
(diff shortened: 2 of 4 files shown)


kennylevinsen:
Now that I think of it, it might not make sense at all to listen for client unbind - there's no events we could send at that point anyway.

However, the second use for unbind - when we tear down the resources during destroy - could be served by a global_destroy so that we can send the leave just before we kill the resource. Alternatively, we could stick to using just global_remove, in which case the only event we're interested in is bind.

kennylevinsen:
My main outstanding question is whether the API should be renamed as done here to indicate the new functionality, or if the old name should be maintained for compatibility. I don't feel strongly either way, just defaulted to updating the name accordingly.

ammen99:
> My main outstanding question is whether the API should be renamed as done here to indicate the new functionality, or if the old name should be maintained for compatibility. I don't feel strongly either way, just defaulted to updating the name accordingly.

I see the new situation as an upgrade from the previous one, so I'm for keeping the name.

types/wlr_output.c
```diff
@@ -93,6 +93,8 @@ static void output_bind(struct wl_client *wl_client, void *data,
 	send_current_mode(resource);
 	send_scale(resource);
 	send_done(resource);
+
+	wl_signal_emit(&output->events.bind, resource);
```
ammen99, line 97: I thought the intention for new code is to either use a dedicated struct or NULL for signals?

kennylevinsen: Changed.

kennylevinsen:
> I see the new situation as an upgrade from the previous one, so I'm for keeping the name.

Changed. I don't think the names are entirely accurate anymore, but the change is minor, so eh.

kennylevinsen:
Note: alpine build failure is due to a temporary package CDN problem, but I don't think I can trigger a re-run without force-pushing a new commit.

Xyene:
Doing `git commit --amend --allow-empty && git push --force` should be fine here, and the easiest way to force the CI to rerun.

kennylevinsen:
> Doing `git commit --amend --allow-empty --no-edit && git push --force` should be fine here, and the easiest way to force the CI to rerun.

That's what I meant by a new commit - it's just slightly annoying when you have to pull tricks for CI reasons. :/

Oh well, pushed and passed.

types/wlr_output.c
emersion: We can use `wl_resource`, but only if we're 100% sure that we'll never need to extend the API again. It's always safer to just introduce a new struct.

emersion:
>it's just slightly annoying when you have to pull tricks for CI reasons

Right. FWIW, I can re-spin the CI manually as well.

>I don't think the names are entirely accurate anymore, but the change is minor, so eh.

Yeah, that's a little bit unfortunate. I don't feel strongly about it.

emersion (approved):
LGTM

emersion:
Thanks!

PR #2469, opened 2020-11-11
Replace wlr_button_state with protocol enums

ifreund:
This is a breaking change, see #2446 for justification.

### DIFF
--- a/backend/headless/output.c
+++ b/backend/headless/output.c
@@ -2,6 +2,7 @@
 #include <GLES2/gl2.h>
 #include <GLES2/gl2ext.h>
 #include <stdlib.h>
+#include <stdio.h>
 #include <wlr/interfaces/wlr_output.h>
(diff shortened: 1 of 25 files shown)


Xyene:
Hmm, I'm not sure this is a net win... it feels like this shifts this enum from wlroots into every compositor wanting to support tablet input.

After this change, a compositor supporting tablet input would have to perform a mapping from {WL_POINTER_BUTTON_STATE, ZWP_TABLET_PAD_V2_BUTTON_STATE} &rarr; some custom type (or more confusingly, both to WL_POINTER_BUTTON_STATE) and operate on that type in order to implement pointer emulation.

(Strictly speaking zwp_tablet_pad_v2_button_state is structurally the same as wl_pointer_button_state so one could get away with casting between them, but I don't think a compositor should be relying on that detail.)

PR #2470, opened 2020-11-11
Remove unneeded includes from wlr_input_device.h

ifreund:
This uncovered many places where we were using things without directly including them.

This allows us to avoid including tablet-unstable-v2 in many of the example compositors.

~~depends on #2469~~

* * *

Breaking change: `wlr/types/wlr_input_device.h` no longer pulls in all device-specific headers.

### DIFF
--- a/include/wlr/types/wlr_input_device.h
+++ b/include/wlr/types/wlr_input_device.h
@@ -9,6 +9,8 @@
 #ifndef WLR_TYPES_WLR_INPUT_DEVICE_H
 #define WLR_TYPES_WLR_INPUT_DEVICE_H
 
+#include <wayland-server-core.h>
+
 enum wlr_button_state {
 	WLR_BUTTON_RELEASED,
 	WLR_BUTTON_PRESSED,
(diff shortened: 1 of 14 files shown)


include/wlr/types/wlr_input_device.h
```diff
@@ -9,6 +9,8 @@
 #ifndef WLR_TYPES_WLR_INPUT_DEVICE_H
 #define WLR_TYPES_WLR_INPUT_DEVICE_H
 
+#include "wayland-server-core.h"
```
emersion, line 12: This should use angle brackets since it's an external header.

emersion:
Hmm. I know it'll be annoying for you, but could you re-order commits so that this PR can be merged before #2469? Otherwise we'll end up in an intermediate state where all compositors including `wlr_input_device.h` will need to generate the tablet protocol header.

include/wlr/types/wlr_input_device.h
ifreund: Thanks for the catch

ifreund:
> Hmm. I know it'll be annoying for you, but could you re-order commits so that this PR can be merged before #2469? Otherwise we'll end up in an intermediate state where all compositors including `wlr_input_device.h` will need to generate the tablet protocol header.

yeah that makes sense, no worries!

emersion (approved):
LGTM, thanks!

PR #2482, opened 2020-11-18
backend/drm: stop force-probing connectors

emersion:
After discussing with Pekka and Daniel on #dri-devel, we concluded [1]
that user-space shouldn't need to force-probe connectors. Force-probing
can take some time, so using drmModeGetConnectorCurrent can result in
faster start-up.

Users can manually trigger a force-probe if necessary:

    echo detect | sudo tee /sys/class/drm/card0-DP-1/status

Or just by running a tool like drm_info.

A similar change has been submitted to Weston [2].

[1]: https://lists.freedesktop.org/archives/dri-devel/2020-November/287728.html
[2]: https://gitlab.freedesktop.org/wayland/weston/-/issues/437

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1282,8 +1282,8 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 	struct wlr_drm_connector *new_outputs[res->count_connectors + 1];
 
 	for (int i = 0; i < res->count_connectors; ++i) {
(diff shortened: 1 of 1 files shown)


ammen99:
This results in a noticeable improvement when starting Wayfire with an external monitor connected.

ammen99:
Another improvement - displayport monitors no longer disconnect when waking up from DPMS: https://github.com/WayfireWM/wayfire/issues/898#issuecomment-734425107

Issue #2480, opened 2020-11-17
Sway fails to start with EGL_BAD_DISPLAY in eglQueryDeviceString


RPigott:
Sway fails to start, on both the wayland and drm backends. This behavior begins for me in c8d95acc.
```
$ sway --version
sway version 1.5-f5ca4c26 (Nov 16 2020, branch 'master')
$ pacman -Q mesa libglvnd
mesa 20.2.2-2
libglvnd 1.3.2-1
$ lspci -kd::300
00:02.0 VGA compatible controller: Intel Corporation HD Graphics 630 (rev 04)
	DeviceName:  Onboard IGD
	Subsystem: Dell Device 07be
	Kernel driver in use: i915
	Kernel modules: i915
$ uname -a
Linux rxps 5.9.8-arch1-1 #1 SMP PREEMPT Tue, 10 Nov 2020 22:44:11 +0000 x86_64 GNU/Linux
```

The following debug log was captured when attempting to start sway:

<details>
<summary>
sway log
</summary>

```
00:00:00.000 [sway/main.c:327] Sway version 1.5-f5ca4c26 (Nov 16 2020, branch 'master')
00:00:00.002 [sway/main.c:154] Linux rxps 5.9.8-arch1-1 #1 SMP PREEMPT Tue, 10 Nov 2020 22:44:11 +0000 x86_64 GNU/Linux
00:00:00.002 [sway/main.c:170] Contents of /etc/lsb-release:
00:00:00.002 [sway/main.c:154] LSB_VERSION=1.4
00:00:00.002 [sway/main.c:154] DISTRIB_ID=Arch
00:00:00.002 [sway/main.c:154] DISTRIB_RELEASE=rolling
00:00:00.002 [sway/main.c:154] DISTRIB_DESCRIPTION="Arch Linux"
00:00:00.002 [sway/main.c:170] Contents of /etc/os-release:
00:00:00.002 [sway/main.c:154] NAME="Arch Linux"
00:00:00.002 [sway/main.c:154] PRETTY_NAME="Arch Linux"
00:00:00.002 [sway/main.c:154] ID=arch
00:00:00.002 [sway/main.c:154] BUILD_ID=rolling
00:00:00.002 [sway/main.c:154] ANSI_COLOR="38;2;23;147;209"
00:00:00.002 [sway/main.c:154] HOME_URL="https://www.archlinux.org/"
[...]

PR #2484, opened 2020-11-18
render/egl: recognize EGL_BAD_DEVICE_EXT error

RPigott:
This error is generated by various egl functions that handle egl devices, and also potentially by glvnd when dispatching those functions. Printing a human-readable name for it is helpful in debugging issues like #2480.


emersion (approved):
Thanks

PR #2486, opened 2020-11-19
texture: document that functions should not be called while rendering

ammen99:
Supersedes: https://github.com/swaywm/wlroots/pull/2438


emersion (approved):
LGTM, thanks!

PR #2487, opened 2020-11-19
implementation of xdg-foreign protocol

### LINKED ISSUE #1702: Add xdg-foreign support

ammen99:
Supersedes #1841 (for now only rebased) and thus fixes #1702 

TODO:

- [x] Make xdg-foreign v1 and v2 share resources (as discussed in https://github.com/swaywm/wlroots/pull/1841#issuecomment-721125163)

### DIFF
--- a/include/util/uuid.h
+++ b/include/util/uuid.h
@@ -1,0 +1,8 @@
+#ifndef UTIL_UUID_H
+#define UTIL_UUID_H
+
+#include <stdbool.h>
+
+bool generate_uuid(char out[static 37]);
+
+#endif
--- a/include/wlr/types/wlr_xdg_foreign_registry.h
+++ b/include/wlr/types/wlr_xdg_foreign_registry.h
@@ -1,0 +1,75 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_XDG_FOREIGN_REGISTRY_H
+#define WLR_TYPES_WLR_XDG_FOREIGN_REGISTRY_H
+
+#include <wayland-server-core.h>
+
+#define WLR_XDG_FOREIGN_HANDLE_SIZE 37
+
+/**
+ * wlr_xdg_foreign_registry is used for storing a list of exported surfaces with
+ * the xdg-foreign family of protocols.
+ *
+ * It can be used to allow interoperability between clients using different
+ * versions of the protocol (if all versions use the same registry).
+ */
+struct wlr_xdg_foreign_registry {
+	struct wl_list exported_surfaces; // struct wlr_xdg_foreign_exported_surface
+
+	struct wl_listener display_destroy;
+	struct {
+		struct wl_signal destroy;
+	} events;
+};
+
+struct wlr_xdg_foreign_exported {
+	struct wl_list link; // wlr_xdg_foreign_registry::exported_surfaces
+	struct wlr_xdg_foreign_registry *registry;
+
+	struct wlr_surface *surface;
+
+	char handle[WLR_XDG_FOREIGN_HANDLE_SIZE];
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+};
+
+/**
+ * Create an empty wlr_xdg_foreign_registry.
+ *
+ * It will be destroyed when the associated display is destroyed.
+ */
+struct wlr_xdg_foreign_registry *wlr_xdg_foreign_registry_create(
+	struct wl_display *display);
+
+/**
+ * Add the given exported surface to the registry and assign it a unique handle.
+ * The caller is responsible for removing the exported surface from the repository
+ * if it is destroyed.
+ *
+ * Returns true if the initialization was successful.
+ */
+bool wlr_xdg_foreign_exported_init(struct wlr_xdg_foreign_exported *surface,
+	struct wlr_xdg_foreign_registry *registry);
+
+/**
+ * Find an exported surface with the given handle, or NULL if such a surface
+ * does not exist.
+ */
+struct wlr_xdg_foreign_exported *wlr_xdg_foreign_registry_find_by_handle(
+	struct wlr_xdg_foreign_registry *registry, const char *handle);
+
+/**
+ * Remove the given surface from the registry it was previously added in.
+ */
+void wlr_xdg_foreign_exported_finish(struct wlr_xdg_foreign_exported *surface);
+
+#endif
--- a/include/wlr/types/wlr_xdg_foreign_v1.h
+++ b/include/wlr/types/wlr_xdg_foreign_v1.h
@@ -1,0 +1,64 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_XDG_FOREIGN_V1_H
+#define WLR_TYPES_WLR_XDG_FOREIGN_V1_H
+
+#include <wayland-server-core.h>
+#include <wlr/types/wlr_xdg_foreign_registry.h>
+
+struct wlr_xdg_foreign_v1 {
+	struct {
+		struct wl_global *global;
+		struct wl_list objects; // wlr_xdg_exported_v1::link or wlr_xdg_imported_v1::link
+	} exporter, importer;
+
+	struct wl_listener foreign_registry_destroy;
+	struct wl_listener display_destroy;
+
+	struct wlr_xdg_foreign_registry *registry;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_xdg_exported_v1 {
+	struct wlr_xdg_foreign_exported base;
+
+	struct wl_resource *resource;
+	struct wl_listener xdg_surface_destroy;
+
+	struct wl_list link; // wlr_xdg_foreign_v1::exporter::objects
+};
+
+struct wlr_xdg_imported_v1 {
+	struct wlr_xdg_foreign_exported *exported;
+	struct wl_listener exported_destroyed;
+
+	struct wl_resource *resource;
+	struct wl_list link; // wlr_xdg_foreign_v1::importer::objects
+	struct wl_list children;
+};
+
+struct wlr_xdg_imported_child_v1 {
+	struct wlr_xdg_imported_v1 *imported;
+	struct wlr_surface *surface;
+
+	struct wl_list link; // wlr_xdg_imported_v1::children
+
+	struct wl_listener xdg_surface_unmap;
+	struct wl_listener xdg_toplevel_set_parent;
+};
+
+struct wlr_xdg_foreign_v1 *wlr_xdg_foreign_v1_create(
+		struct wl_display *display, struct wlr_xdg_foreign_registry *registry);
+
+#endif
--- a/include/wlr/types/wlr_xdg_foreign_v2.h
+++ b/include/wlr/types/wlr_xdg_foreign_v2.h
@@ -1,0 +1,64 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_XDG_FOREIGN_V2_H
+#define WLR_TYPES_WLR_XDG_FOREIGN_V2_H
+
+#include <wayland-server-core.h>
+#include <wlr/types/wlr_xdg_foreign_registry.h>
+
+struct wlr_xdg_foreign_v2 {
+	struct {
+		struct wl_global *global;
+		struct wl_list objects; // wlr_xdg_exported_v2::link or wlr_xdg_imported_v2::link
+	} exporter, importer;
+
+	struct wl_listener foreign_registry_destroy;
+	struct wl_listener display_destroy;
+
+	struct wlr_xdg_foreign_registry *registry;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_xdg_exported_v2 {
+	struct wlr_xdg_foreign_exported base;
+
+	struct wl_resource *resource;
+	struct wl_listener xdg_surface_destroy;
+
+	struct wl_list link; // wlr_xdg_foreign_v2::exporter::objects
+};
+
+struct wlr_xdg_imported_v2 {
+	struct wlr_xdg_foreign_exported *exported;
+	struct wl_listener exported_destroyed;
+
+	struct wl_resource *resource;
+	struct wl_list link; // wlr_xdg_foreign_v2::importer::objects
+	struct wl_list children;
+};
+
+struct wlr_xdg_imported_child_v2 {
+	struct wlr_xdg_imported_v2 *imported;
+	struct wlr_surface *surface;
+
+	struct wl_list link; // wlr_xdg_imported_v2::children
+
+	struct wl_listener xdg_surface_unmap;
+	struct wl_listener xdg_toplevel_set_parent;
+};
+
+struct wlr_xdg_foreign_v2 *wlr_xdg_foreign_v2_create(
+		struct wl_display *display, struct wlr_xdg_foreign_registry *registry);
+
+#endif
--- a/meson.build
+++ b/meson.build
@@ -87,6 +87,7 @@ conf_data.set10('WLR_HAS_X11_BACKEND', false)
 conf_data.set10('WLR_HAS_XWAYLAND', false)
 conf_data.set10('WLR_HAS_XCB_ERRORS', false)
 conf_data.set10('WLR_HAS_XCB_ICCCM', false)
+conf_data.set10('WLR_HAS_XDG_FOREIGN', false)
 
 # Clang complains about some zeroed initializer lists (= {0}), even though they
 # are valid
@@ -109,8 +110,16 @@ pixman = dependency('pixman-1')
 math = cc.find_library('m')
 rt = cc.find_library('rt')
 
-uuid = dependency('uuid', required: false)
-uuid_create = cc.has_function('uuid_create')
+if not get_option('xdg-foreign').disabled()
+	uuid = dependency('uuid', required: false)
+	uuid_create = cc.has_function('uuid_create')
+	if uuid.found() or uuid_create
+		conf_data.set10('WLR_HAS_XDG_FOREIGN', true)
+	elif get_option('xdg-foreign').enabled()
+		error('Missing dependency uuid and uuid_create function not available ' +
+				'cannot build with xdg-foreign support')
+	endif
+endif
 
 wlr_files = []
 wlr_deps = [
@@ -171,6 +180,7 @@ summary({
 	'x11_backend': conf_data.get('WLR_HAS_X11_BACKEND', 0) == 1,
 	'xcb-icccm': conf_data.get('WLR_HAS_XCB_ICCCM', 0) == 1,
 	'xcb-errors': conf_data.get('WLR_HAS_XCB_ERRORS', 0) == 1,
+	'xdg-foreign': conf_data.get('WLR_HAS_XDG_FOREIGN', 0) == 1,
 }, bool_yn: true)
 
 if get_option('examples')
--- a/meson.build
+++ b/meson.build
@@ -109,6 +109,9 @@ pixman = dependency('pixman-1')
 math = cc.find_library('m')
 rt = cc.find_library('rt')
 
+uuid = dependency('uuid', required: false)
+uuid_create = cc.has_function('uuid_create')
+
 wlr_files = []
 wlr_deps = [
 	wayland_server,
--- a/types/meson.build
+++ b/types/meson.build
@@ -66,3 +66,9 @@ wlr_files += files(
 	'wlr_xdg_decoration_v1.c',
 	'wlr_xdg_output_v1.c',
 )
+
+if conf_data.get('WLR_HAS_XDG_FOREIGN', 0) == 1
+	wlr_files += files(
+		'wlr_xdg_foreign_registry.c',
+	)
+endif
--- a/types/meson.build
+++ b/types/meson.build
@@ -70,6 +70,7 @@ wlr_files += files(
 if conf_data.get('WLR_HAS_XDG_FOREIGN', 0) == 1
 	wlr_files += files(
 		'wlr_xdg_foreign_v1.c',
+		'wlr_xdg_foreign_v2.c',
 		'wlr_xdg_foreign_registry.c',
 	)
 endif
--- a/types/meson.build
+++ b/types/meson.build
@@ -69,6 +69,7 @@ wlr_files += files(
 
 if conf_data.get('WLR_HAS_XDG_FOREIGN', 0) == 1
 	wlr_files += files(
+		'wlr_xdg_foreign_v1.c',
 		'wlr_xdg_foreign_registry.c',
 	)
 endif
--- a/types/wlr_xdg_foreign_registry.c
+++ b/types/wlr_xdg_foreign_registry.c
@@ -1,0 +1,73 @@
+#include <wlr/types/wlr_xdg_foreign_registry.h>
+#include "util/signal.h"
+#include "util/uuid.h"
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+
+bool wlr_xdg_foreign_exported_init(
+		struct wlr_xdg_foreign_exported *exported,
+		struct wlr_xdg_foreign_registry *registry) {
+	do {
+		if (!generate_uuid(exported->handle)) {
+			return false;
+		}
+	} while (wlr_xdg_foreign_registry_find_by_handle(registry, exported->handle) != NULL);
+
+	exported->registry = registry;
+	wl_list_insert(&registry->exported_surfaces, &exported->link);
+
+	wl_signal_init(&exported->events.destroy);
+	return true;
+}
+
+struct wlr_xdg_foreign_exported *wlr_xdg_foreign_registry_find_by_handle(
+		struct wlr_xdg_foreign_registry *registry, const char *handle) {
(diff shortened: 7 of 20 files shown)


ammen99:
I added a new `wlr_xdg_foreign_registry`. The compositors now can choose whether to let v1 and v2 share handles or not.

A basic implementation in a compositor is here: https://github.com/WayfireWM/wayfire/pull/892

include/util/uuid.h
```diff
@@ -0,0 +1,6 @@
+#ifndef UTIL_UUID_H
+#define UTIL_UUID_H
+
+int generate_uuid(char out[static 37]);
```
emersion, line 4: For consistency with the rest of wlroots, we could change this to return a `bool`.

util/uuid.c
```diff
@@ -0,0 +1,32 @@
+#include "util/uuid.h"
```
emersion, line 1: Nit: local includes should come after external includes.

util/meson.build
```diff
@@ -7,3 +7,14 @@ wlr_files += files(
 	'signal.c',
 	'time.c',
 )
+
+if uuid.found()
+	wlr_deps += uuid
+	add_project_arguments('-DHAS_LIBUUID', language: 'c')
```
emersion, line 13: We try to avoid `#ifdef`s and use `#if` instead so that forgetting to define the macro triggers an error.

types/meson.build
```diff
@@ -66,3 +66,11 @@ wlr_files += files(
 	'wlr_xdg_decoration_v1.c',
 	'wlr_xdg_output_v1.c',
 )
+if uuid.found() or uuid_create
+	conf_data.set10('WLR_HAS_XDG_FOREIGN', true)
```
emersion, line 70: Can we move this one line to the main `meson.build`? It's too easy to incorrectly include another Meson file that depends on this being set before executing `subdir('types')`.

meson.build
```diff
@@ -107,6 +108,8 @@ libinput = dependency('libinput', version: '>=1.14.0')
 xkbcommon = dependency('xkbcommon')
 udev = dependency('libudev')
 pixman = dependency('pixman-1')
+uuid = dependency('uuid', required: false)
+uuid_create = cc.has_function('uuid_create')
```
emersion, line 112: I think we should have a Meson feature option for xdg-foreign, just like we have for all other features. Package maintainers don't want auto-detection of features, they want hard failures when a dep is missing.

include/wlr/types/wlr_xdg_foreign_registry.h
```diff
@@ -0,0 +1,76 @@
...
+	struct {
+		struct wl_signal destroy;
+	} events;
+};
+
+struct wlr_xdg_foreign_exported {
+	struct wl_list link; // wlr_xdg_foreign_registry::exported_surfaces
+	struct wlr_xdg_foreign_registry *registry;
+
+	struct wlr_surface *surface;
```
emersion, line 36: We don't have a destroy listener for this surface. Do we rely on the caller to properly destroy the `wlr_xdg_foreign_exported` when the surface is destoyed?

If so, we should properly document that.

include/wlr/types/wlr_xdg_foreign_registry.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+struct wlr_xdg_foreign_registry *wlr_xdg_foreign_registry_create(
+	struct wl_display *display);
+
+/**
+ * Add the given exported surface to the registry and assign it a unique handle.
+ *
+ * @return True if the initialization was successful.
+ */
+bool wlr_xdg_foreign_registry_initialize_exported(
```
emersion, line 58: Maybe rename to `wlr_xdg_foreign_exported_init` for consistency with the rest of wlroots?

include/wlr/types/wlr_xdg_foreign_registry.h
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * It will be destroyed when the associated display is destroyed.
+ */
+struct wlr_xdg_foreign_registry *wlr_xdg_foreign_registry_create(
+	struct wl_display *display);
+
+/**
+ * Add the given exported surface to the registry and assign it a unique handle.
+ *
+ * @return True if the initialization was successful.
```
emersion, line 56: We don't use `@return` comments.

include/wlr/types/wlr_xdg_foreign_registry.h
```diff
@@ -0,0 +1,76 @@
...
+ * Find an exported surface with the given handle, or NULL if such a surface
+ * does not exist.
+ */
+struct wlr_xdg_foreign_exported *wlr_xdg_foreign_registry_find_by_handle(
+	struct wlr_xdg_foreign_registry *registry, const char *handle);
+
+/**
+ * Remote the given surface from the registry.
+ */
+void wlr_xdg_foreign_registry_remove_exported(
```
emersion, line 72: Maybe rename to `wlr_xdg_foreign_exported_fini` for consistency with the rest of wlroots?

include/wlr/types/wlr_xdg_foreign_v2.h
```diff
@@ -0,0 +1,85 @@
...
+#ifndef WLR_TYPES_WLR_XDG_FOREIGN_V2_H
+#define WLR_TYPES_WLR_XDG_FOREIGN_V2_H
+
+#include <wayland-server-core.h>
+#include <wlr/types/wlr_xdg_foreign_registry.h>
+
+struct wlr_xdg_foreign_v2 {
+	struct {
+		struct wl_global *global;
+		struct wl_list resources;
```
emersion, line 18: We don't have a `resources` list for globals anymore.

include/wlr/types/wlr_xdg_foreign_v2.h
```diff
@@ -0,0 +1,85 @@
...
+		struct wl_list clients;
```
emersion, line 19: Do we really need per-client lists? Can't we just have directly a list of `wlr_xdg_exported_v2`/`wlr_xdg_imported_v2`?

include/wlr/types/wlr_xdg_foreign_v2.h
```diff
@@ -0,0 +1,85 @@
...
+
+	struct wl_list link; // wlr_xdg_imported_v2::children
+
+	struct wl_listener xdg_surface_unmap;
+	struct wl_listener xdg_toplevel_set_parent;
+};
+
+struct wlr_xdg_foreign_v2 *wlr_xdg_foreign_v2_create(
+		struct wl_display *display, struct wlr_xdg_foreign_registry *registry);
+void wlr_xdg_foreign_v2_destroy(struct wlr_xdg_foreign_v2 *foreign);
```
emersion, line 83: We don't expose destructors for globals like this one anymore.

types/wlr_xdg_foreign_registry.c
```diff
@@ -0,0 +1,77 @@
+#include <wlr/types/wlr_xdg_foreign_registry.h>
+#include "util/signal.h"
+#include "util/uuid.h"
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+
+/**
```
emersion, line 8: The doc-comment is already in the header file. Would prefer not to duplicate it.

types/wlr_xdg_foreign_registry.c
```diff
@@ -0,0 +1,77 @@
...
+	} while (wlr_xdg_foreign_registry_find_by_handle(registry, exported->handle) != NULL);
+
+	exported->registry = registry;
+	wl_list_insert(&registry->exported_surfaces, &exported->link);
+	return true;
+}
+
+struct wlr_xdg_foreign_exported *wlr_xdg_foreign_registry_find_by_handle(
+		struct wlr_xdg_foreign_registry *registry, const char *handle) {
+
```
emersion, line 27: Nit: unnecessary newline.

types/wlr_xdg_foreign_registry.c
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	return NULL;
+}
+
+void wlr_xdg_foreign_registry_remove_exported(
+		struct wlr_xdg_foreign_registry *registry,
+		struct wlr_xdg_foreign_exported *surface) {
+
```
emersion, line 45: Nit: unnecessary newline.

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,472 @@
...
+		&xdg_imported_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void xdg_imported_handle_destroy(struct wl_client *client,
+		struct wl_resource *resource) {
+	wl_resource_destroy(resource);
+}
+
+static bool verify_is_toplevel(struct wl_resource *client_resource,
```
emersion, line 31: Hm, why do we need to check whether the surface is a toplevel and send a protocol error if not? The protocol doesn't define any errors like this.

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,472 @@
...
+void wlr_xdg_foreign_v2_destroy(
+		struct wlr_xdg_foreign_v2 *foreign) {
+	if (!foreign) {
+		return;
+	}
+
+	struct wl_resource *resource, *tmp_resource;
+	wl_resource_for_each_safe(resource, tmp_resource,
+			&foreign->importer.resources) {
+		wl_resource_destroy(resource);
```
emersion, line 402: The resources should already be destroyed by the time the `wl_display` is.

emersion (changes requested):
Thanks for continuing the effort on this feature! Here's a first round of comments.

Apart from these comments, I'd also like the commits to be re-organized and split. For instance the xdg-shell change can be in a standalone commit, so can be the introduction of `util/uuid`. You can use [`Co-authored-by` tags](https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) in the commit message to indicate a secondary commit author.

include/wlr/types/wlr_xdg_foreign_registry.h
ammen99: I don't like `wlr_xdg_foreign_exported_init` because the registry initializes the given exported surface, it is not like the exported surface initializes itself (if it were C++ I'd put this method inside the registry class). If you however really don't like the name I can rename it, no problem.

emersion: I was thinking of changing the order of arguments too:

```c
bool wlr_xdg_foreign_exported_init(struct wlr_xdg_foreign_exported *surface, struct wlr_xdg_foreign_registry *registry);
```

which would mean "initialize the exported foreign given this registry".

I'm also fine with keeping the current order, but I'd really like to use the `init`/`finish` wording we use elsewhere.

include/wlr/types/wlr_xdg_foreign_registry.h
emersion, line 72: Also, we shouldn't need to provide the registry here, `wlr_xdg_foreign_exported` already has a registry field.

include/wlr/types/wlr_xdg_foreign_v2.h
ammen99: yeah the list is actually of wlr_xdg_exported_v2 not of clients, I think `objects` would be a better name, or what do you think?

ammen99:
Thanks for the review, I'm working on these changes. I will be just pushing new commits until we're ready for merging, and then I'll reorganize the commits as you suggested.

include/wlr/types/wlr_xdg_foreign_registry.h
ammen99: yeah, I explained in in the docstring for init()

types/wlr_xdg_foreign_v2.c
ammen99: Should we instead silently fail?

meson.build
ammen99: I have added this, let me know what you think of the approach.

ammen99:
I think I have addressed all of your comments, except the one about erroring when the handle is not a toplevel. The protocol says `Only xdg_surface surfaces may be exported.` which I think is enough grounds to disconnect a client which exports something else.

include/util/uuid.h
```diff
@@ -0,0 +1,8 @@
+#ifndef UTIL_UUID_H
+#define UTIL_UUID_H
+
+#include <stdbool.h>
+
+bool generate_uuid(char out[static 37]);
```
emersion, line 6: Maybe we should have a constant for this?

```c
#ifndef UUID_STR_LEN
#define UUID_STR_LEN 37
#endif
```

meson.build
```diff
@@ -110,6 +111,17 @@ pixman = dependency('pixman-1')
 math = cc.find_library('m')
 rt = cc.find_library('rt')
 
+if not get_option('xdg-foreign').disabled()
+	uuid = dependency('uuid', required: false)
+	uuid_create = cc.has_function('uuid_create')
+	if uuid.found() or uuid_create
+		conf_data.set10('WLR_HAS_XDG_FOREIGN', 1)
```
emersion, line 118: Nit: `set10` takes a `bool`, so should be `true` instead of 1.

meson.build
```diff
@@ -110,6 +111,17 @@ pixman = dependency('pixman-1')
...
+	elif get_option('xdg-foreign').enabled()
+		error('Missing dependency uuid and uuid_create function not available' +
```
emersion, line 120: Nit: missing space in error message.

include/util/uuid.h
emersion: Hm, I guess it wouldn't help too much, since this is a private header.

util/meson.build
```diff
@@ -7,3 +7,16 @@ wlr_files += files(
...
+
+if conf_data.get('WLR_HAS_XDG_FOREIGN', 0) == 1
+	if uuid.found()
+		wlr_deps += uuid
+		add_project_arguments('-DHAS_LIBUUID=1', language: 'c')
+	else
+		add_project_arguments('-DHAS_LIBUUID=0', language: 'c')
+	endif
+	wlr_files += files(
+			'uuid.c',
```
emersion, line 20: Nit: this can just be single-line

util/uuid.c
```diff
@@ -0,0 +1,31 @@
...
+	uuid_create(&uuid, &status);
+	if (status != uuid_s_ok) {
+		return false;
+	}
+	char *str;
+	uuid_to_string(&uuid, &str, &status);
+	if (status != uuid_s_ok) {
+		return false;
+	}
+	strcpy(out, str);
```
emersion, line 27: `strcpy` is insecure. Use `memcpy` instead. Maybe `assert(strlen(str) + 1 == 37)` would be nice too.

include/wlr/types/wlr_xdg_foreign_v1.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include <wayland-server-core.h>
+#include <wlr/types/wlr_xdg_foreign_registry.h>
+
+struct wlr_xdg_foreign_v1 {
+	struct wl_global *exporter_global;
+	struct wl_global *importer_global;
+
+	struct {
+		struct wl_global *global;
```
emersion, line 20: We have `exporter_global` and `exporter.global`. Do we need both?

types/wlr_xdg_foreign_v1.c
```diff
@@ -0,0 +1,458 @@
...
+
+	wl_list_insert(&foreign->importer.objects, &importer->link);
+}
+
+static void xdg_foreign_destroy(struct wlr_xdg_foreign_v1 *foreign) {
+	if (!foreign) {
+		return;
+	}
+
+	wlr_signal_emit_safe(&foreign->events.destroy, foreign);
```
emersion, line 396: Use NULL as data argument to allow future extensibility.

types/wlr_xdg_foreign_v1.c
```diff
@@ -0,0 +1,458 @@
...
+		&zxdg_exported_v1_interface,
+		wl_resource_get_version(exporter->resource),
+		id);
+
+	wl_resource_set_implementation(exported->resource, &xdg_exported_impl,
+			exported, xdg_exported_handle_resource_destroy);
+
+	wl_list_insert(&exporter->exports, &exported->link);
+	wl_list_init(&exported->imports);
+	wl_signal_init(&exported->base.events.destroy);
```
emersion, line 222: This is already done by `wlr_xdg_foreign_exported_init`

types/wlr_xdg_foreign_v1.c
```diff
@@ -0,0 +1,458 @@
...
+	.destroy = xdg_exporter_handle_destroy,
+	.export = xdg_exporter_handle_export
+};
+
+static void xdg_exporter_handle_resource_destroy(
+		struct wl_resource *resource) {
+	struct wlr_xdg_exporter_v1 *exporter = xdg_exporter_from_resource(resource);
+	struct wlr_xdg_exported_v1 *exported, *exported_tmp;
+	wl_list_for_each_safe(exported, exported_tmp, &exporter->exports, link) {
+		wl_resource_destroy(exported->resource);
```
emersion, line 244: Nothing in the protocol says that destroying the exporter also destroys the child objects.

types/wlr_xdg_foreign_v1.c
```diff
@@ -0,0 +1,458 @@
...
+	.import = xdg_importer_handle_import
+};
+
+static void xdg_importer_handle_resource_destroy(
+		struct wl_resource *resource) {
+	struct wlr_xdg_importer_v1 *importer =
+		xdg_importer_from_resource(resource);
+	struct wlr_xdg_imported_v1 *imported, *imported_tmp;
+	wl_list_for_each_safe(imported, imported_tmp, &importer->imports, link) {
+		disconnect_imported(imported);
```
emersion, line 357: Nothing in the protocol says that destroying the importer also destroys the child objects.

include/wlr/types/wlr_xdg_foreign_v1.h
```diff
@@ -0,0 +1,86 @@
...
+	struct wlr_xdg_foreign_registry *registry;
+
+	struct {
+		struct wl_signal destroy;
+	} events;
+
+	void *data;
+};
+
+struct wlr_xdg_exporter_v1 {
```
emersion, line 36: I still don't think we need this intermediate struct. We can just store the imported/exported objects in the `wlr_xdg_foreign_v1`.

include/wlr/types/wlr_xdg_foreign_v1.h
ammen99: Indeed, this is a mistake, will remove.

types/wlr_xdg_foreign_v1.c
ammen99: It isn't done but probably should be :)

include/wlr/types/wlr_xdg_foreign_v1.h
ammen99: You are right! I misunderstood your previous comment. This change removed a lot of unnecessary code.

ammen99:
Alright, made it so that we don't store anything about the intermediate exporter/importer resources. They directly have the `foreign` as their user data, which is enough.

types/wlr_xdg_foreign_registry.c
```diff
@@ -0,0 +1,72 @@
...
+	wl_list_for_each(exported, &registry->exported_surfaces, link) {
+		if (strcmp(handle, exported->handle) == 0) {
+			return exported;
+		}
+	}
+
+	return NULL;
+}
+
+void wlr_xdg_foreign_exported_finish(struct wlr_xdg_foreign_exported *surface) {
```
emersion, line 40: Maybe we should trigger the `destroy` event here?

types/wlr_xdg_foreign_v2.c
emersion: No, reading the protocol again, sending a protocol error is what we should do, but we're just missing the error values.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/57

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+	struct wlr_xdg_imported_v2 *imported =
+		calloc(1, sizeof(struct wlr_xdg_imported_v2));
+	if (imported == NULL) {
+		wl_client_post_no_memory(wl_client);
+		return;
+	}
+
+	imported->exported = wlr_xdg_foreign_registry_find_by_handle(
+		foreign->registry, handle);
+	imported->resource = wl_resource_create(wl_client,
```
emersion, line 304: Missing a NULL check here

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+		wl_resource_get_version(client_resource),
+		id);
+	wl_resource_set_implementation(imported->resource, &xdg_imported_impl,
+			imported, xdg_imported_handle_resource_destroy);
+
+	wl_list_init(&imported->children);
+	wl_list_insert(&foreign->importer.objects, &imported->link);
+
+	if (imported->exported == NULL) {
+		wl_list_init(&imported->exported_destroyed.link);
```
emersion, line 315: I'd prefer to make imported objects without an exported inert, by making the user data pointer NULL and free'ing the `struct wlr_xdg_imported_v2`. This is the standard thing that wlroots does when objects are still alive but can't be used anymore.

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	if (!wlr_xdg_foreign_exported_init(&exported->base, foreign->registry)) {
+		wl_client_post_no_memory(wl_client);
+		free(exported);
+		return;
+	}
+
+	exported->base.surface = surface;
+	exported->resource = wl_resource_create(wl_client,
```
emersion, line 212: Missing a NULL check here.

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+	wl_resource_set_implementation(exported->resource, &xdg_exported_impl,
+			exported, xdg_exported_handle_resource_destroy);
+
+	wl_list_insert(&foreign->exporter.objects, &exported->link);
+	wl_list_init(&exported->imports);
+
+	zxdg_exported_v2_send_handle(exported->resource, exported->base.handle);
+
+	exported->xdg_surface_unmap.notify = handle_xdg_surface_unmap;
+	if (wlr_surface_is_xdg_surface(surface)) {
```
emersion, line 226: We checked `verify_is_toplevel` above, so this should be unnecessary?

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+	wl_list_remove(&exported->xdg_surface_unmap.link);
+	wl_list_remove(&exported->link);
+	free(exported);
+}
+
+static void handle_xdg_surface_unmap(
+		struct wl_listener *listener, void *data) {
+	struct wlr_xdg_exported_v2 *exported =
+		wl_container_of(listener, exported, xdg_surface_unmap);
+	wl_resource_destroy(exported->resource);
```
emersion, line 183: This destroys the resource behind the client's back. We can't do this, otherwise the client will die when sending an `xdg_exported.destroy` request.

Why do we need to handle `unmap` in the first place? The protocol doesn't seem to mention anything about it.

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+	}
+	struct wlr_surface *wlr_surface = imported->exported->surface;
+	struct wlr_surface *wlr_surface_child =
+		wlr_surface_from_resource(child_resource);
+
+	if (!verify_is_toplevel(resource, wlr_surface_child)) {
+		return;
+	}
+	if (wlr_surface_is_xdg_surface(wlr_surface)
+			!= wlr_surface_is_xdg_surface(wlr_surface_child)) {
```
emersion, line 86: Hm, `verify_is_toplevel` should take care of this too I think?

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+	child = calloc(1, sizeof(struct wlr_xdg_imported_child_v2));
+	if (child == NULL) {
+		wl_client_post_no_memory(client);
+		return;
+	}
+	child->surface = wlr_surface_child;
+	child->xdg_surface_unmap.notify = handle_child_xdg_surface_unmap;
+	child->xdg_toplevel_set_parent.notify = handle_xdg_toplevel_set_parent;
+
+	if (wlr_surface_is_xdg_surface(wlr_surface)) {
```
emersion, line 107: This should always be true as well I think.

types/wlr_xdg_foreign_v2.c
```diff
@@ -0,0 +1,411 @@
...
+			wlr_xdg_surface_from_wlr_surface(wlr_surface);
+		struct wlr_xdg_surface *surface_child =
+			wlr_xdg_surface_from_wlr_surface(wlr_surface_child);
+
+		wlr_xdg_toplevel_set_parent(surface_child, surface);
+		wl_signal_add(&surface_child->events.unmap,
+				&child->xdg_surface_unmap);
+		wl_signal_add(&surface_child->toplevel->events.set_parent,
+				&child->xdg_toplevel_set_parent);
+	}
```
emersion, line 118: `child` is leaked in the else branch, because `wlr_xdg_toplevel_set_parent` won't free it in `xdg_imported_handle_resource_destroy`.

types/wlr_xdg_foreign_v2.c
ammen99: Yes, it was a leftover from the time when there was both xdg6 and xdg-stable

types/wlr_xdg_foreign_v2.c
ammen99: if we remove the if() there is no else, right?

types/wlr_xdg_foreign_v2.c
ammen99: A good question. I will remove that as well.

ammen99: Hmm, I suppose we want to listen not for unmap, but for the case where the toplevel is destroyed.

ammen99:
I think I have addressed your new review comments. Inert state is used in two different places:

- For imported, if the exported could not be found
- For exported, if the xdg-surface is destroyed (not unmapped, as it was previously).

emersion (commented):
This looks good to me, thanks. Can you split everything into logical commits? See https://github.com/swaywm/wlroots/pull/2487#pullrequestreview-546948373

ammen99:
> This looks good to me, thanks. Can you split everything into logical commits? See [#2487 (review)](https://github.com/swaywm/wlroots/pull/2487#pullrequestreview-546948373)

Thanks for the review! I have split everything into five commits.

emersion (approved):
Nice work, thanks for you contribution!

Issue #2429, opened 2020-10-12
Changing gamma doesn't change the display's gamma


msrd0:
I used to use redshift-wlr until it recently started segfaulting on me. Since the project seemed abandoned, I tried gammastep and wlsunset, which both claimed to be updating the colour temperature but nothing changed on my screen. Also, the `gamma-control` example from wlroots does not do anything, it just idles until I Ctrl+C it.

This is the debug output:

```
~/git/wlroots [master|✔]$ WAYLAND_DEBUG=1 ./build/examples/gamma-control -b 0.5
[  1362.623]  -> wl_display@1.get_registry(new id wl_registry@2)
[  1362.682]  -> wl_display@1.sync(new id wl_callback@3)
[  1362.995] wl_display@1.delete_id(3)
[  1363.051] wl_registry@2.global(1, "wl_shm", 1)
[  1363.087] wl_registry@2.global(2, "wl_drm", 2)
[  1363.130] wl_registry@2.global(3, "zwp_linux_dmabuf_v1", 3)
[  1363.168] wl_registry@2.global(4, "wl_compositor", 4)
[  1363.203] wl_registry@2.global(5, "wl_subcompositor", 1)
[  1363.250] wl_registry@2.global(6, "wl_data_device_manager", 3)
[  1363.294] wl_registry@2.global(7, "zwlr_gamma_control_manager_v1", 1)
[  1363.335]  -> wl_registry@2.bind(7, "zwlr_gamma_control_manager_v1", 1, new id [unknown]@4)
[  1363.413] wl_registry@2.global(8, "gtk_primary_selection_device_manager", 1)
[  1363.449] wl_registry@2.global(9, "zxdg_output_manager_v1", 3)
[  1363.492] wl_registry@2.global(10, "org_kde_kwin_idle", 1)
[  1363.536] wl_registry@2.global(11, "zwp_idle_inhibit_manager_v1", 1)
[  1363.573] wl_registry@2.global(12, "zwlr_layer_shell_v1", 2)
[  1363.605] wl_registry@2.global(13, "xdg_wm_base", 2)
[  1363.651] wl_registry@2.global(14, "zwp_tablet_manager_v2", 1)
[  1363.691] wl_registry@2.global(15, "org_kde_kwin_server_decoration_manager", 1)
[  1363.734] wl_registry@2.global(16, "zxdg_decoration_manager_v1", 1)
[...]

PR #2489, opened 2020-11-19
backend/drm: Use legacy gamma size for legacy backend

kennylevinsen:
This is a quick fix for https://github.com/swaywm/wlroots/issues/2429.


emersion (approved):
LGTM, thanks!

PR #2493, opened 2020-11-22
POSIX/GNU compliance macros fixes for uClibc support

pcercuei:
Hi,

These two patches allow me to build wlroots with uClibc.

- `_XOPEN_SOURCE` is required for `strdup` to become available. Since it defines `_POSIX_C_SOURCE` (see `man feature_test_macros(7)`), remove all inline defines of these two macros, and have meson provide only the former.
- `F_DUPFD_CLOEXEC` seems to be a GNU-only thing, and will be provided by <fcntl.h> only when _GNU_SOURCE is defined on uClibc.


emersion:
>meson: define _XOPEN_SOURCE globally

Nack, we record what each POSIX revision each source file needs, just like we record which header files it requires.

>render/dmabuf: define _GNU_SOURCE for F_DUPFD_CLOEXEC

Nack, this is POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html

pcercuei:
Well, it needs _XOPEN_SOURCE defined everywhere strdup() is used, and _GNU_SOURCE for F_DUPFD_CLOEXEC (even though these are POSIX) to compile on uClibc.

emersion:
uClibc is wrong here. strdup is POSIX, and so is F_DUPFD_CLOEXEC, so only _POSIX_C_SOURCE should be required. Please report a uClibc bug.

pcercuei:
uClibc is wrong here, but so is glibc. strdup() conforms to POSIX.1-2001 (according to its man page), so half of your code base could probably use _POSIX_C_SOURCE=200112L, but doesn't just so that it compiles with a broken library (glibc only exports strdup when _POSIX_C_SOURCE=200809L).

I'll try to send a patch to uClibc without too much hope (they probably have _reasons_ why this has been broken for 20 years) but that won't change the fact that wlroots cannot be compiled with the current uClibc-based toolchains.

emersion:
>glibc only exports strdup when _POSIX_C_SOURCE=200809L

Then report the bug to glibc.

pcercuei:
I forced-push to a new commit that adds some feature macros in order for `struct timespec` to be visible on uClibc. The commit is based on 0.12.0, I will need to try with the latest master too.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2493

PR #2494, opened 2020-11-24
backend/libseat: fix change_vt return value

ifreund:
This should return true on success and false on failure not vice-versa.


kennylevinsen:
LGTM, although I see now that the libseat logind backend incorrectly returns 1 on success and -1 on error. That will be fixed in the next release.

emersion (approved):
OK, thanks both of you!

PR #2496, opened 2020-11-25
backend/drm: query render formats

emersion:
~~Depends on https://github.com/swaywm/wlroots/pull/2483~~

On some platforms it's possible that the display engine supports
modifiers not supported by the render engine.

Query render formats and intersect them with plane formats to accommodate
for this.

### DIFF
--- a/render/drm_format_set.c
+++ b/render/drm_format_set.c
@@ -137,3 +137,39 @@ struct wlr_drm_format *wlr_drm_format_dup(const struct wlr_drm_format *format) {
 	memcpy(duped_format, format, format_size);
 	return duped_format;
 }
+
+struct wlr_drm_format *wlr_drm_format_intersect(
+		const struct wlr_drm_format *a, const struct wlr_drm_format *b) {
+	assert(a->format == b->format);
+
+	size_t format_cap = a->len < b->len ? a->len : b->len;
+	size_t format_size = sizeof(struct wlr_drm_format) +
+		format_cap * sizeof(a->modifiers[0]);
+	struct wlr_drm_format *format = calloc(1, format_size);
+	if (format == NULL) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+	format->format = a->format;
+	format->cap = format_cap;
+
+	for (size_t i = 0; i < a->len; i++) {
+		for (size_t j = 0; j < b->len; j++) {
+			if (a->modifiers[i] == b->modifiers[j]) {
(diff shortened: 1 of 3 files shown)


emersion:
Side note, another reason why we should query render formats is that when a driver adds support for modifiers, at some point KMS will return modifiers but Mesa won't understand them. This is what is happening right now on AMD GFX9+.

render/drm_format_set.c
```diff
@@ -137,3 +137,37 @@ struct wlr_drm_format *wlr_drm_format_dup(const struct wlr_drm_format *format) {
...
+				format->modifiers[format->len] = a->modifiers[i];
+				format->len++;
+				break;
+			}
+		}
+	}
+
+	// If both formats support modifiers, but the intersection is empty, then
+	// the formats aren't compatible with each other
+	if (format->len == 0 && (a->len > 0 && b->len == 0)) {
```
ammen99, line 167: Did you mean `a->len > 0 && b->len > 0` in the second part of the condition?

render/drm_format_set.c
```diff
@@ -137,3 +137,37 @@ struct wlr_drm_format *wlr_drm_format_dup(const struct wlr_drm_format *format) {
...
+
+	size_t format_len = a->len < b->len ? a->len : b->len;
+	size_t format_size = sizeof(struct wlr_drm_format) +
+		format_len * sizeof(a->modifiers[0]);
+	struct wlr_drm_format *format = calloc(1, format_size);
+	if (format == NULL) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+	format->format = a->format;
```
ammen99, line 153: What is `format->cap` and does it have to be zero in all the cases? (I'd have thought that if `*a == *b`, then we expect `*a == *format` in the end)

ammen99 (commented):
LGTM aside from the two comments I have left.

render/drm_format_set.c
emersion: Good catch!

render/drm_format_set.c
emersion: That's a good point. `format->cap` is set appropriately now.

emersion:
Also added some docs for `wlr_drm_format_intersect` in the latest version.

ammen99 (approved):
LGTM, should I merge?

emersion:
Please do!

ammen99:
Thanks :+1:

PR #2500, opened 2020-11-27
xwm: add wlr_xwayland_surface_restack()

ammen99:
This PR adds a simple function for restacking Xwayland surfaces above or below each other.

To allow the most flexibility, this function receives a sibling and a mode argument, like the corresponding X11 configure event does.

EDIT: this is needed because in X11 the surface which gets pointer events is determined by the stacking order. Thus, if a compositor has the concept of always-on-top views or similar, they need to tell the Xserver what the real stacking order is.

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -252,6 +252,14 @@ void wlr_xwayland_set_cursor(struct wlr_xwayland *wlr_xwayland,
 void wlr_xwayland_surface_activate(struct wlr_xwayland_surface *surface,
 	bool activated);
 
+/**
+ * Restack surface relative to sibling.
+ * If sibling is NULL, then the surface is moved to the top or the bottom
+ * of the stack (depending on the mode).
+ */
+void wlr_xwayland_surface_restack(struct wlr_xwayland_surface *surface,
+	struct wlr_xwayland_surface *sibling, enum xcb_stack_mode_t mode);
+
 void wlr_xwayland_surface_configure(struct wlr_xwayland_surface *surface,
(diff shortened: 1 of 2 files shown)


include/wlr/xwayland.h
```diff
@@ -252,6 +252,12 @@ void wlr_xwayland_set_cursor(struct wlr_xwayland *wlr_xwayland,
 void wlr_xwayland_surface_activate(struct wlr_xwayland_surface *surface,
 	bool activated);
 
+/**
+ * Restack @surface relative to @sibling (which might be NULL).
```
emersion, line 256: Nit: we don't use @-references in our doc comments.

xwayland/xwm.c
```diff
@@ -1548,6 +1548,23 @@ void wlr_xwayland_surface_activate(struct wlr_xwayland_surface *xsurface,
 	}
 }
 
+void wlr_xwayland_surface_restack(struct wlr_xwayland_surface *surface,
+		struct wlr_xwayland_surface *sibling, enum xcb_stack_mode_t mode) {
+	struct wlr_xwm *xwm = surface->xwm;
+	uint32_t values[2];
+	uint32_t idx = 0;
```
emersion, line 1555: Nit: should be `size_t`

include/wlr/xwayland.h
emersion, line 256: What happens if sibling is NULL? We should probably document that.

emersion (commented):
Looks like a good addition to me, apart from these minor comments

emersion (approved):
LGTM, thanks!

Issue #2499, opened 2020-11-26
After 713c166, output disconnection/reconnection doesn't work reliably


luispabon:
After updating past 713c166, sway seems to have trouble knowing when outputs have been disconnected or reconnected. I'm attaching a debug log of a sway session where:

  - Start sway
  - Open two terminals on different workspaces in different outputs
  - Disconnect my thunderbolt dock (2 displays on it, one on vga and another on hdmi): at this point, sway does not react to the disconnection. The workspace with the second terminal is still in a display that's not connected anymore. `get_outputs` still reports all 3 displays connected
  - Reconnect the dock: displays do not get reactivated

sway log: https://paste.ubuntu.com/p/HXMt3c8mVC/
get_outputs: https://paste.ubuntu.com/p/9yQPkrBPsV/

uname -a:
```
Linux luis-XPS-15-9560 5.4.0-53-generic #59-Ubuntu SMP Wed Oct 21 09:38:44 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```
i7-7700HQ & intel graphics hd630

wlroots: 6485fadc16a45ccbcbdcc76b064bcd2ee6256cda
sway: sway version 1.5-32b93ef6 (Nov 26 2020, branch 'master')

Rolling back my wlroots version to 154fe8696f40151dae999cfc69299aef5132dd14 fixes the issue for me.

emersion:
>00:00:12.133 [ERROR] [backend/drm/atomic.c:36] DP-9: Atomic commit failed (pageflip): Invalid argument

Can you obtain DRM debug logs? See https://github.com/swaywm/wlroots/wiki/DRM-Debugging

luispabon:
Of course:

dmesg: https://paste.ubuntu.com/p/srPqWBh6Jr/
sway.log: https://paste.ubuntu.com/p/WmBxZsnZpm/

luispabon:
Another thing to note. If I use just the HDMI display and I connect it straight into the connector on the laptop the issue does not happen.

synaptiko:
I also started having troubles after https://github.com/swaywm/wlroots/commit/713c1661b742f93a7d2167321837c0d99541ca87

My external monitor is connected over HDMI over USB-C dock. Its highest resolution is 2560x1440. Previously it worked fine but after that commit, I only get 1920x1080 resolution.

[...]

PR #2503, opened 2020-11-30
Revert "backend/drm: stop force-probing connectors"

emersion:
This reverts commit 713c1661b742f93a7d2167321837c0d99541ca87.

It turns out we do need to force-probe on startup and on hotplug [1].
This is unfortunate, but that's just how the uAPI works. Fixing this
would require patching the kernel.

[1]: https://lists.freedesktop.org/archives/dri-devel/2020-November/289506.html

Closes: https://github.com/swaywm/wlroots/issues/2499

cc @ammen99

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1282,8 +1282,8 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 	struct wlr_drm_connector *new_outputs[res->count_connectors + 1];
 
 	for (int i = 0; i < res->count_connectors; ++i) {
-		drmModeConnector *drm_conn =
-			drmModeGetConnectorCurrent(drm->fd, res->connectors[i]);
+		drmModeConnector *drm_conn = drmModeGetConnector(drm->fd,
+			res->connectors[i]);
 		if (!drm_conn) {
 			wlr_log_errno(WLR_ERROR, "Failed to get DRM connector");
 			continue;


ammen99:
Is it not possible to not call this each time?

emersion:
No, this isn't possible without fixing the kernel.

ammen99:
Alright, but this fixed more problems than just slow startup. For example, force-probing forces outputs to disconnect and reconnect when waking up from DPMS. Why do we need to force probe in that case?

emersion:
Because we receive a hotplug event, supposedly?

emersion:
Oh, and #intel-gfx said your issue might be an amdgpu bug

```
2020-11-26 19:25:45     danvet  I think that's an amdgpu bug
2020-11-26 19:25:53     danvet  you need to whack screens pretty hard for probing to work
2020-11-26 19:26:41     danvet  especially dp
```

So you might want to report a kernel bug.

ammen99:
> Because we receive a hotplug event, supposedly?

Ah ok. I read up through the code a bit and I see what you mean. Sad, but I don't see how to avoid that. I'll patch my wlroots locally though :)

JordanL2:
@emersion Am I right in understand that this command:
```
echo detect | sudo tee /sys/class/drm/card0-DP-1/status | sudo tee /sys/class/drm/card0-DP-2/status
```
should cause the same behaviour as calling `drmModeGetConnector` for the same set of connectors?

It's just that I can't recreate the reconnect issue by running that command.

emersion:
Merging because https://patchwork.freedesktop.org/patch/404887/ has collected enough reviews.

Issue #2154, opened 2020-05-04
xwayland: don't use SIGUSR1


emersion:
Right now to signal the compositor that Xwayland has started, we use SIGUSR1: https://github.com/swaywm/wlroots/blob/46c83cbf3daf6c38c70d0dcb5492e164b6c8ab29/xwayland/xwayland.c#L355

We could use a pipe instead. This would fix quite a few compositor issues with global state.

PR #2504, opened 2020-11-30
xwayland: use pipe instead of SIGUSR1 to signal readiness

dominikh:
Closes: https://github.com/swaywm/wlroots/issues/2154

### DIFF
--- a/xwayland/server.c
+++ b/xwayland/server.c
@@ -57,6 +57,7 @@ noreturn static void exec_xwayland(struct wlr_xwayland_server *server) {
 	}
 
 	/* Make Xwayland signal us when it's ready */
+	/* TODO: can we use -displayfd instead? */
 	signal(SIGUSR1, SIG_IGN);
 
 	char *argv[] = {
@@ -139,8 +140,8 @@ static void server_finish_process(struct wlr_xwayland_server *server) {
 		wl_list_remove(&server->client_destroy.link);
 		wl_client_destroy(server->client);
 	}
-	if (server->sigusr1_source) {
-		wl_event_source_remove(server->sigusr1_source);
+	if (server->pipe_source) {
+		wl_event_source_remove(server->pipe_source);
 	}
 
 	safe_close(server->wl_fd[0]);
@@ -184,8 +185,8 @@ static void handle_client_destroy(struct wl_listener *listener, void *data) {
 	struct wlr_xwayland_server *server =
 		wl_container_of(listener, server, client_destroy);
 
-	if (server->sigusr1_source) {
-		// Xwayland failed to start, let the sigusr1 handler deal with it
+	if (server->pipe_source) {
+		// Xwayland failed to start, let the readiness handler deal with it
 		return;
 	}
 
@@ -219,7 +220,21 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	wlr_xwayland_server_destroy(server);
 }
 
-static int xserver_handle_ready(int signal_number, void *data) {
+static int xserver_handle_ready(int fd, uint32_t mask, void *data) {
+	// There are three ways in which we can end up here, from server_start:
+	// 1. the second fork failed
+	// 2. the exec failed
+	// 3. Xwayland sent a SIGUSR1
+	//
+	// All three cases result in a write to the pipe, which triggers us.
+	//
+	// For the first two cases, the first fork will exit with
+	// EXIT_FAILURE, notifying us that startup failed.
+	//
+	// For the third case, the first fork will exit with EXIT_SUCCESS
+	// and we'll know that Xwayland started successfully.
+
+	close(fd);
 	struct wlr_xwayland_server *server = data;
 
 	int stat_val = -1;
@@ -236,8 +251,8 @@ static int xserver_handle_ready(int signal_number, void *data) {
 	}
 	wlr_log(WLR_DEBUG, "Xserver is ready");
 
-	wl_event_source_remove(server->sigusr1_source);
-	server->sigusr1_source = NULL;
+	wl_event_source_remove(server->pipe_source);
+	server->pipe_source = NULL;
 
 	struct wlr_xwayland_server_ready_event event = {
 		.server = server,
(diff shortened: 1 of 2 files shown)


xwayland/server.c
```diff
@@ -332,15 +345,24 @@ static bool server_start(struct wlr_xwayland_server *server) {
...
 			_exit(EXIT_FAILURE);
 		} else if (pid == 0) {
 			exec_xwayland(server);
 		}
 
 		int sig;
 		sigwait(&sigset, &sig);
-		kill(ppid, SIGUSR1);
-		wlr_log(WLR_DEBUG, "sent SIGUSR1 to process %d", ppid);
+		if (write(p[1], "", 0) == -1) {
```
emersion, line 356: Maybe we can just write a single byte like `\n`? I'm worried about a zero-byte write not waking up poll…

xwayland/server.c
```diff
@@ -332,15 +345,24 @@ static bool server_start(struct wlr_xwayland_server *server) {
 		pid_t pid = fork();
 		if (pid < 0) {
 			wlr_log_errno(WLR_ERROR, "second fork failed");
+			(void) !write(p[1], "", 0);
```
emersion, line 348: Can we drop the `!`?

emersion, line 348: Style: no space after cast

```c
(void)write(…);
```

xwayland/server.c
```diff
@@ -332,15 +345,24 @@ static bool server_start(struct wlr_xwayland_server *server) {
...
 			_exit(EXIT_FAILURE);
 		} else if (pid == 0) {
```
emersion, line 350: Should we `close(p[1])` here?

xwayland/server.c
emersion, line 348: Actually, do we even need that write? If the parent receives `WL_EVENT_HANGUP`, it can detect the failure?

xwayland/server.c
```diff
@@ -310,19 +312,30 @@ static bool server_start(struct wlr_xwayland_server *server) {
...
 	wl_client_add_destroy_listener(server->client, &server->client_destroy);
 
+	int p[2];
+	if (pipe(p) == -1) {
+		wlr_log_errno(WLR_ERROR, "pipe failed");
+		server_finish_process(server);
+		return false;
+	}
+	fcntl(p[1], F_SETFD, FD_CLOEXEC);
+	fcntl(p[0], F_SETFD, FD_CLOEXEC);
```
emersion, line 322: We have a `set_cloexec` helper for this.

xwayland/server.c
```diff
@@ -219,7 +220,8 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	wlr_xwayland_server_destroy(server);
 }
 
-static int xserver_handle_ready(int signal_number, void *data) {
+static int xserver_handle_ready(int fd, uint32_t mask, void *data) {
```
emersion, line 223: I think it'd be better to check for the mask here. If it contains `WL_EVENT_READABLE`, then we're good. If it contains `WL_EVENT_HANGUP`, then Xwayland startup failed.

emersion (changes requested):
Thanks for working on this! Here are a few comments.

xwayland/server.c
dominikh: Without the `!`, gcc still complains that the result of `write` is unused.

xwayland/server.c
dominikh: Exiting the process will close the file descriptor.

xwayland/server.c
dominikh: Can you explain to me what triggers WL_EVENT_HANGUP? The write end of the pipe being closed? The general issue with that is that we might leak the write fd to another process, as creating the pipe and setting the fds CLOEXEC isn't atomic, so there's a race. If it weren't for that, none of the code paths would have to write, we could just close the fd. But since another process might keep the write end open, we have to actually do a write.

xwayland/server.c
dominikh: Considering the race I described above, we can't rely on seeing a WL_EVENT_HANGUP. Yeah, we could return early if we see one, but not seeing one won't imply that there was no failure. IOW, handling WL_EVENT_HANGUP won't remove any of the existing code in the function, it's superfluous.

xwayland/server.c
dominikh: Does using the helper gain us anything here? We're not interested in keeping any other flags in tact, and avoiding the call to F_GETFD reduces the size of the race window.

xwayland/server.c
emersion: Right, I guess `exec_xwayland` will either exit or exec, so it doesn't matter.

xwayland/server.c
emersion: We shouldn't assume that `FD_CLOEXEC` is the only flag, and we shouldn't change default flags. Additionally, we should error out if setting `FD_CLOEXEC` fails, instead of leaking a FD (leaking FDs are hard to debug).

xwayland/server.c
emersion: I'd really like to be able to detect Xwayland exec failures. Xwayland might fail to exec if the executable is missing for instance.

>Can you explain to me what triggers WL_EVENT_HANGUP? The write end of the pipe being closed?

Yes.

>The general issue with that is that we might leak the write fd to another process, as creating the pipe and setting the fds CLOEXEC isn't atomic, so there's a race.

This can only happen if the compositor is multi-threaded and forks+execs outside of the wlroots thread. POSIX doesn't give us tools (yet) to atomically create a CLOEXEC pipe, so I'd just ignore the race for now.

dominikh: > I'd really like to be able to detect Xwayland exec failures. Xwayland might fail to exec if the executable is missing for instance.

But we are currently detecting that. 

- If the second fork (line 345) fails, the write to the pipe (line 348) wakes up `xserver_handle_ready`, which will waitpid on the first fork and see that it exited with EXIT_FAILURE
- If the exec fails, the write (line 356) will wake up `xserver_handle_ready`, but not before the `if (sig == SIGCHLD)` branch exited with EXIT_FAILURE

When there was no failure, the same write (line 356) wakes up the handler, which now sees EXIT_SUCCESS – which can only happen if the exec succeeded and Xwayland sent a SIGUSR1.

> This can only happen if the compositor is multi-threaded and forks+execs outside of the wlroots thread.

Sure, but it _can_ happen, and when it does happen, we'll get hangs. The write avoids that risk.

xwayland/server.c
emersion: OK

xwayland/server.c
emersion: >waitpid on the first fork and see that it exited with EXIT_FAILURE

Ah. I forgot we waitpid'ed the first fork. An issue with this behavior is compositors can't set a global SIGCHLD handler without stealing the Xwayland process status. But that's a completely separate issue.

emersion (approved):
OK, this is already a big improvement over the status quo. Can you squash all of these changes into a single logical commit?

emersion:
Rebased and squashed locally. Thanks!

dominikh:
Sorry for the delay; thanks for merging!

PR #2505, opened 2020-12-01
Move primary wlr_swapchain handling out of backends

emersion:
Move the allocator/renderer-related code out of backends into the common backend/output code for everything related to the primary buffer. This process shouldn't break any compositor.

The DRM backend has not been migrated yet, because it requires more work.

- [x] Allow backends to be created without `get_renderer`
- [x] Allow outputs to be created without `{attach,rollback}_render`
- [x] Create allocator and renderer in `wlr_backend`
- [x] Create swapchain in `wlr_output`
- [x] Add support for the Pixman renderer
- [x] Migrate backends
  - [x] Headless
  - [x] Wayland
  - [x] X11
  - [x] ~~DRM~~ (done separately: https://github.com/swaywm/wlroots/pull/2903)

References: https://github.com/swaywm/wlroots/issues/1352
~~Depends on: https://github.com/swaywm/wlroots/pull/2496~~
~~Depends on: https://github.com/swaywm/wlroots/pull/2498~~
~~Depends on: https://github.com/swaywm/wlroots/pull/2495~~
~~Depends on: https://github.com/swaywm/wlroots/pull/2511~~
~~Depends on: https://github.com/swaywm/wlroots/pull/2507~~
~~Depends on: https://github.com/swaywm/wlroots/pull/2561~~

### DIFF
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -70,8 +70,6 @@ static void backend_destroy(struct wlr_backend *wlr_backend) {
 
 	wlr_backend_finish(wlr_backend);
 
-	free(backend->format);
-
 	close(backend->drm_fd);
 	free(backend);
 }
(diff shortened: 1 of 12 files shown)


emersion:
- We need to select a CRTC for the connector prior to `wlr_output_test`/`wlr_output_commit`, so that we can return a meaningful list of formats out of `wlr_output_impl.get_dmabuf_primary_formats`.
- Ideally we should completely wire up test-only commits to `wlr_output_test` to implement the modifier-less fallback path. Although I suppose we can also let the atomic commit fail, just like we currently do internally in the DRM backend.

emersion:
This has been rebased and updated to work with the Pixman renderer.

This is ready for review.

bl4ckb0ne (approved):

PR #2510, opened 2020-12-04
backend/headless: use config-less context

emersion:
This is preliminary work for https://github.com/swaywm/wlroots/pull/2505.

With `wlr_swapchain`, we stop using `EGLSurface` entirely, so we can skip selecting an `EGLConfig`.

### DIFF
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -165,22 +165,12 @@ struct wlr_backend *wlr_headless_backend_create(struct wl_display *display,
 		return NULL;
 	}
 
-	static const EGLint config_attribs[] = {
-		EGL_SURFACE_TYPE, 0,
-		EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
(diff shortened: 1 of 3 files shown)


ammen99 (commented):
LGTM. However, it seems that libweston will fall back to using some EGL config attribs in case that the EGL extensions are not present. Is there someplace we can find how widely supported these are?

emersion:
>However, it seems that libweston will fall back to using some EGL config attribs in case that the EGL extensions are not present. Is there someplace we can find how widely supported these are?

I thought of doing that too, but I figured we could try without the workaround first and see if anyone complains about the missing extension.

Mesa unconditionally enables the extension for all drivers: https://gitlab.freedesktop.org/mesa/mesa/-/blob/661b29283425c568d1fa403be703775e02c50d08/src/egl/drivers/dri2/egl_dri2.c#L928

ammen99 (approved):
Mesa supporting this unconditionally sounds good enough, let us hope no one complains. Thanks :)

PR #2513, opened 2020-12-04
drm: do not unset make/model before emitting destroy event

ammen99:
Otherwise, compositors do not have access to make/model when the output is being removed.

Note: not sure about the name `drm_connector_reset`, if someone has a better suggestion, I'll update the PR.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1617,10 +1617,6 @@ static void drm_connector_cleanup(struct wlr_drm_connector *conn) {
 		conn->output.enabled = false;
 		conn->output.width = conn->output.height = conn->output.refresh = 0;
 
-		memset(&conn->output.make, 0, sizeof(conn->output.make));
-		memset(&conn->output.model, 0, sizeof(conn->output.model));
-		memset(&conn->output.serial, 0, sizeof(conn->output.serial));
-
 		if (conn->output.idle_frame != NULL) {
 			wl_event_source_remove(conn->output.idle_frame);
 			conn->output.idle_frame = NULL;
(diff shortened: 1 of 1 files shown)


emersion:
What about just moving the `memset` calls below?

ammen99:
> What about just moving the `memset` calls below?

 Below where? They are unset in only two of the switch cases

emersion:
Below `wlr_signal_emit_safe` in the original code. Beware, these switches are fallthrough: the `WLR_DRM_CONN_NEEDS_MODESET` case is executed even if the state is `WLR_DRM_CONN_CONNECTED`.

Yes, this is confusing. But if we're going to refactor something, I think I'd prefer to call `wlr_output_destroy` properly instead of manually resetting all of the `wlr_output` fields and triggering the event.

emersion:
For reference, the glider approach is to use `wlr_output_impl.destroy` to clear the backend-specific fields: https://github.com/emersion/glider/blob/619cf43a630f08998a9f7a9e328996342a29dfd1/backend/connector.c#L199

And call `wlr_output_destroy` when the connector gets disconnected: https://github.com/emersion/glider/blob/619cf43a630f08998a9f7a9e328996342a29dfd1/backend/connector.c#L335

ammen99:
@emersion I didn't want to move it below initially since this would be a change of behavior which I am not sure about. However if you say that's fine...

emersion:
Yeah that should be fine, I don't see how it could cause issues, because it would execute the reset _more often_.

ammen99:
Okay, I hope it is good now.

emersion (approved):
Thanks!

PR #2514, opened 2020-12-04
backend/session/libseat: set XDG_SESSION_TYPE

emersion:
libseat will call logind's SetType method if necessary.

cc @kennylevinsen


kennylevinsen:
LGTM

PR #2515, opened 2020-12-04
screencopy: handle compositor not setting damage

ifreund:
Damage the full output if the compositor didn't submit damage but did
submit a buffer.

### DIFF
--- a/types/wlr_screencopy_v1.c
+++ b/types/wlr_screencopy_v1.c
@@ -51,15 +51,14 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 	int width, height;
 	wlr_output_transformed_resolution(output, &width, &height);
 
-	// If the compositor did not submit damage but did submit a buffer,
-	if (!(output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) &&
-			(output->pending.committed & WLR_OUTPUT_STATE_BUFFER)) {
-		// damage everything
-		pixman_region32_union_rect(region, region, 0, 0, width, height);
-	} else {
-		// otherwise copy over the current damage
+	if (output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) {
+		// If the compositor submitted damage, copy it over
 		pixman_region32_union(region, region, &output->pending.damage);
 		pixman_region32_intersect_rect(region, region, 0, 0, width, height);
+	} else if (output->pending.committed & WLR_OUTPUT_STATE_BUFFER) {
+		// If the compositor did not submit damage but did submit a buffer
+		// damage everything
+		pixman_region32_union_rect(region, region, 0, 0, width, height);
 	}
 
 	damage->last_commit_seq = output->commit_seq;
--- a/types/wlr_screencopy_v1.c
+++ b/types/wlr_screencopy_v1.c
@@ -48,9 +48,20 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 		return;
 	}
 
-	pixman_region32_union(region, region, &output->pending.damage);
-	pixman_region32_intersect_rect(region, region, 0, 0, output->width,
-		output->height);
+	// If the compositor did not submit damage but did submit a buffer,
+	if (!(output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) &&
+			(output->pending.committed & WLR_OUTPUT_STATE_BUFFER)) {
+		// damage everything
+		int width, height;
+		wlr_output_transformed_resolution(output, &width, &height);
+		pixman_region32_union_rect(region, region, 0, 0, width, height);
+	} else {
+		// otherwise copy over the current damage
+		pixman_region32_union(region, region, &output->pending.damage);
+		pixman_region32_intersect_rect(region, region, 0, 0, output->width,
+			output->height);
+	}
+
 	damage->last_commit_seq = output->commit_seq;
 }
 
--- a/types/wlr_screencopy_v1.c
+++ b/types/wlr_screencopy_v1.c
@@ -48,18 +48,18 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 		return;
 	}
 
+	int width, height;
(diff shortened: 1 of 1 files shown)


types/wlr_screencopy_v1.c
```diff
@@ -48,9 +48,20 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
...
-	pixman_region32_union(region, region, &output->pending.damage);
-	pixman_region32_intersect_rect(region, region, 0, 0, output->width,
-		output->height);
+	// If the compositor did not submit damage but did submit a buffer,
+	if (!(output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) &&
+		(output->pending.committed & WLR_OUTPUT_STATE_BUFFER)) {
+		// damage everything
+		int width, height;
+		wlr_output_transformed_resolution(output, &width, &height);
+		pixman_region32_union_rect(region, region, 0, 0, width, height);
```
ifreund, line 57: I'm not sure if I chose the right pixman function here, they don't seem to have any docs...

types/wlr_screencopy_v1.c
```diff
@@ -48,9 +48,20 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 		return;
 	}
 
-	pixman_region32_union(region, region, &output->pending.damage);
-	pixman_region32_intersect_rect(region, region, 0, 0, output->width,
-		output->height);
+	// If the compositor did not submit damage but did submit a buffer,
+	if (!(output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) &&
+		(output->pending.committed & WLR_OUTPUT_STATE_BUFFER)) {
```
emersion, line 53: Style nit: use two tabs for `if` condition continuation lines, to tell them apart from the `if` body.

types/wlr_screencopy_v1.c
emersion: Yes, this sounds like the right function to use.

types/wlr_screencopy_v1.c
```diff
@@ -48,9 +48,20 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
...
+		// damage everything
+		int width, height;
+		wlr_output_transformed_resolution(output, &width, &height);
+		pixman_region32_union_rect(region, region, 0, 0, width, height);
+	} else {
+		// otherwise copy over the current damage
+		pixman_region32_union(region, region, &output->pending.damage);
+		pixman_region32_intersect_rect(region, region, 0, 0, output->width,
+			output->height);
```
emersion, line 62: Looking at it now, it sounds like this breaks when using an output transform? We should use `wlr_output_transformed_resolution` like above, I think.

Can you fix that in a different commit (makes bisecting easier if we find bugs about it)?

types/wlr_screencopy_v1.c
```diff
@@ -48,9 +48,20 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
...
-	pixman_region32_intersect_rect(region, region, 0, 0, output->width,
-		output->height);
+	// If the compositor did not submit damage but did submit a buffer,
+	if (!(output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) &&
+		(output->pending.committed & WLR_OUTPUT_STATE_BUFFER)) {
+		// damage everything
+		int width, height;
+		wlr_output_transformed_resolution(output, &width, &height);
+		pixman_region32_union_rect(region, region, 0, 0, width, height);
+	} else {
```
emersion, line 58: I know the previous code wasn't doing it, but we should check for `output->pending.committed & WLR_OUTPUT_STATE_DAMAGE` here. Otherwise we might be reading invalid state from `output->pending.damage`.

Can you fix that in a different commit (makes bisecting easier if we find bugs about it)?

ifreund:
Added your suggestions in separate commits and retested to make sure my use-case still works (which it does). If the commits look good to you I'd say this is good to go.

emersion (approved):
LGTM, thanks!

types/wlr_screencopy_v1.c
```diff
@@ -48,18 +48,18 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 		return;
 	}
 
+	int width, height;
+	wlr_output_transformed_resolution(output, &width, &height);
```
any1, line 52: This looks like it will break stuff.

ammen99: Ah indeed, damage is accumulated in transformed coordinates, but when it is committed, it is in buffer coordinates ...

emersion: Ah, interesting. I suppose we should just revert to `output->{width,height}` then?

I'll also create an issue about removing all of this mess.

any1: I haven't confirmed that it breaks anything. I just thought this looked suspicious because I've never run into an issue with damage reporting on transformed outputs before. I dunno, maybe it fixes bugs that I didn't know about.

emersion: I don't see any damage conversion in `wlr_output_set_damage` nor in `wlr_output_commit`. So I believe `wlr_output_state.damage` is in buffer-local coords. The comment in the header file even says so:

```c
struct wlr_output_state {
	pixman_region32_t damage; // output-buffer-local coordinates
```

In fact, Sway already transforms the damage manually before calling `wlr_output_set_damage`:

```c
	enum wl_output_transform transform =
		wlr_output_transform_invert(wlr_output->transform);
	wlr_region_transform(&frame_damage, &output->damage->current,
		transform, width, height);
	wlr_output_set_damage(wlr_output, &frame_damage);
	pixman_region32_fini(&frame_damage);
```

tl;dr we should use `output->{width,height}` here.

emersion: https://github.com/swaywm/wlroots/issues/2518

ifreund: #2519

PR #2517, opened 2020-12-06
wlr-output-management: Handle modes added after initializing

maringuu:
The DRM backend adds custom modes to wlr_output.modes
Currently modes that are added after the first occurence of
wlr_output_configuration_head_v1 are not added to wlr_output_head.mode_resources.

### DIFF
--- a/types/wlr_output_management_v1.c
+++ b/types/wlr_output_management_v1.c
@@ -819,6 +819,26 @@ static bool manager_update_head(struct wlr_output_manager_v1 *manager,
 		state |= HEAD_STATE_SCALE;
 	}
 
+	// If  a mode was added to wlr_output.modes we need to add the new mode
+	// to the wlr_output_head
+	struct wlr_output_mode *mode;
+	wl_list_for_each(mode, &head->state.output->modes, link) {
+		bool found = false;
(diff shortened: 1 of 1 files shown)


types/wlr_output_management_v1.c
```diff
@@ -819,6 +819,25 @@ static bool manager_update_head(struct wlr_output_manager_v1 *manager,
 		state |= HEAD_STATE_SCALE;
 	}
 
+	// When a mode was added to wlr_output.modes we need to add the new mode to the wlr_output_head
```
ifreund, line 822: style nit: you should wrap this comment to 80 columns, also "if" sounds better than "when" here.

ifreund (commented):
With this patch and https://github.com/ifreund/river/pull/143 I am able to set and use a custom mode of 1816x1080 using wdisplays in river. I'm not sure why, but even with this patch attempting to set the mode to the same dimensions in sway does nothing.

Nice work @maringuu!

emersion (approved):
LGTM, thanks!

Note to self: modes can't be removed from wlr_output right now,  so we don't need to destroy mode resources that are no longer in `wlr_output.modes`.

PR #2519, opened 2020-12-07
screencopy: don't use output transform for damage copy

ifreund:
Only wlr_output_damage works in transformed coordinates,
wlr_output->damage is in output-buffer-local coordinates.

This essentially reverts 1ecc1b5 and fixes 1477401.


emersion (approved):
Thanks for the fix!

Issue #2479, opened 2020-11-16
Assertion failed when creating cursor FBO on multi-GPU setup


J0nnyMak0:
With latest sources, sway fails to start. Log attached.
[sway.log](https://github.com/swaywm/sway/files/5548472/sway.log)

neon64:
I don't know nearly enough about wlroots to understand this bug, but in the meantime adding this hack inside of [gles2_get_renderer_in_context](https://github.com/swaywm/wlroots/blob/master/render/gles2/renderer.c#L36) at least prevents the crash. I'm sure the actual fix is elsewhere though as `gles2_get_renderer_in_context` is supposed to assume the context is already current, right?

```c
if(!wlr_egl_is_current(renderer->egl)) {
    wlr_egl_make_current(renderer->egl, EGL_NO_SURFACE, NULL);
}
```

function calls leading up to the crash:
`wlr_output_cursor_set_surface` -> `drm_connector_set_cursor` -> `drm_fb_lock_surface` -> `drm_surface_unset_current` -> `wlr_renderer_bind_buffer` -> `gles2_bind_buffer` -> `gles2_get_renderer_in_context`

J0nnyMak0:
FWIW I see the assert only when external monitors are connected.

J0nnyMak0:
I can create a pull request if this is a satisfactory solution. Even with this change, I still get freezes from time to time when hotplugging external monitors. Not sure if that is related to this or some other change. (I haven't kept up with more recent changes to wlroots/sawy as I've not been using sway on a daily basis because of this issue.)

emersion:
`drm_surface_unset_current` must only be called after making the surface current. Apparently that's not the case here.

emersion:
Can you try https://github.com/swaywm/wlroots/pull/2520?

PR #2520, opened 2020-12-07
backend/drm: unset current surface before importing

emersion:
drm_fb_import_wlr may need to change the current EGL context. For
instance by calling drm_fb_clear, which calls wlr_buffer_unlock, which
may destroy a buffer if the cursor swapchain size has changed, which
calls gles2's destroy_buffer, which calls glDeleteFramebuffers.

Closes: https://github.com/swaywm/wlroots/issues/2479

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -322,12 +322,15 @@ void drm_fb_clear(struct wlr_drm_fb *fb) {
 bool drm_fb_lock_surface(struct wlr_drm_fb *fb, struct wlr_drm_surface *surf) {
 	assert(surf->back_buffer != NULL);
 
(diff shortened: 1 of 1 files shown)


J0nnyMak0:
This gets rid of the crash, but the external monitors display just a black screen with the cursor. I'm attaching the debug log when booted with external monitors connected. The result is the same even when hotplugging after booting without external monitors connected.

[sway.log](https://github.com/swaywm/wlroots/files/5655725/sway.log)

emersion:
Can you open a separate bug for this? I think the workaround in https://github.com/swaywm/wlroots/blob/c9760569ae380d99df4fb9c7d1f00d0c70f9e580/backend/drm/drm.c#L984 is no longer working.

emersion:
Actually, I already found this bug: https://github.com/swaywm/wlroots/issues/2525

PR #2521, opened 2020-12-07
docs: fix error in wlr_output_set_damage() comment

ifreund:
output-buffer-local coordinates are neither scaled nor transformed


emersion (approved):
Thanks!

PR #2522, opened 2020-12-07
output-management: use same types as wlr_output

ifreund:
This is more correct and also makes things much nicer for languages
that don't allow implicit conversions between these types.

### DIFF
--- a/types/wlr_output_management_v1.c
+++ b/types/wlr_output_management_v1.c
@@ -240,7 +240,7 @@ static void config_head_handle_set_scale(struct wl_client *client,
 		return;
 	}
 
-	double scale =  wl_fixed_to_double(scale_fixed);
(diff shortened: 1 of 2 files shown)


types/wlr_output_management_v1.c
```diff
@@ -240,7 +240,7 @@ static void config_head_handle_set_scale(struct wl_client *client,
 		return;
 	}
 
-	double scale =  wl_fixed_to_double(scale_fixed);
+	float scale =  wl_fixed_to_double(scale_fixed);
```
ifreund, line 243: This makes me feel icky.  We could also go the route of making `wlr_output.scale` a double instead of a float, but that would be a more widely reaching change.

emersion (approved):
Thanks

Issue #2527, opened 2020-12-08
render/gles2: gles2_preferred_read_format relies on EGL config


emersion:
The EGL config is not to be trusted when an FBO is used. Additionally, since https://github.com/swaywm/wlroots/pull/2510 we may not have any EGL config selected.

Instead, we should check whether the current buffer format has an alpha channel.

```
[EGL] command: eglGetConfigAttrib, error: EGL_BAD_CONFIG (0x3005), message: "eglGetConfigAttrib"
```

PR #2528, opened 2020-12-08
render/gles2: query alpha size from render buffer 

emersion:
If we don't have an EGL config, don't try to query anything from it.

If we're using a render buffer, query the alpha size from it.

Closes: #2527

cc @any1


any1 (approved):
Yeah, that takes care if it. Thanks!

PR #2529, opened 2020-12-09
backend/drm: remove EGL config

emersion:
Since we're using wlr_swapchain, we don't need to provide an EGL config.


ammen99 (approved):
LGTM, thanks!

PR #2530, opened 2020-12-09
backend/drm: stop using GBM flags

emersion:
gbm_bo_create_with_modifiers doesn't take GBM flags, so our
wlr_gbm_allocator interface doesn't either. We were still internally
using GBM flags in the DRM backend, leading to awkward back-and-forth
conversions.

The only flag passed to drm_plane_init_surface was GBM_BO_USE_LINEAR, so
turn that into a bool to make sure other flags can't be passed in.

Move the "force linear" logic out of init_drm_surface, because the
supplied wlr_drm_format should already contain that information.

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -66,7 +66,7 @@ void finish_drm_renderer(struct wlr_drm_renderer *renderer) {
 
 static bool init_drm_surface(struct wlr_drm_surface *surf,
 		struct wlr_drm_renderer *renderer, uint32_t width, uint32_t height,
-		const struct wlr_drm_format *drm_format, uint32_t flags) {
+		const struct wlr_drm_format *drm_format) {
 	if (surf->width == width && surf->height == height) {
 		return true;
 	}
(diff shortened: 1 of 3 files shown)


backend/drm/renderer.c
```diff
@@ -277,22 +263,38 @@ bool drm_plane_init_surface(struct wlr_drm_plane *plane,
 		drm_format = wlr_drm_format_dup(&format_no_modifiers);
 	}
 
+	struct wlr_drm_format *drm_format_linear =
+		calloc(1, sizeof(struct wlr_drm_format) + sizeof(uint64_t));
+	if (drm_format_linear == NULL) {
+		return false;
```
ammen99, line 269: Don't we need to free `drm_format`?

ammen99: I also don't like we are doing an allocation we might not need but the alternative seems much more complex code ...

backend/drm/renderer.c
```diff
@@ -277,22 +263,38 @@ bool drm_plane_init_surface(struct wlr_drm_plane *plane,
...
 	if (!drm->parent) {
-		ok = init_drm_surface(&plane->surf, &drm->renderer, width, height,
-			drm_format, flags | GBM_BO_USE_SCANOUT);
+		ok = init_drm_surface(&plane->surf, &drm->renderer,
+			width, height, drm_format);
 	} else {
 		ok = init_drm_surface(&plane->surf, &drm->parent->renderer,
-			width, height, drm_format, flags | GBM_BO_USE_LINEAR);
+			width, height, drm_format_linear);
 		if (ok && !init_drm_surface(&plane->mgpu_surf, &drm->renderer,
```
ammen99, line 290: A question, unrelated to the PR: what is `mgpu_surf` aka what does purpose does this second attempt at init_drm_surface achieve?

ammen99 (changes requested):
LGTM overall, however, I think you have a memory leak :)

backend/drm/renderer.c
emersion: Good catch, fixed.

> I also don't like we are doing an allocation we might not need but the alternative seems much more complex code ...

Yes. It shouldn't matter in practice, it's not a hot path at all, and micro-optimizations detrimental to code readability are a bad idea.

backend/drm/renderer.c
emersion: On multi-GPU setups, we need two surfaces: one on the primary GPU for rendering operations, and one on the secondary GPU to copy over the final buffer for scan-out. The code initializes both surfaces here.

ammen99 (approved):
LGTM. Thanks!

PR #2535, opened 2020-12-10
egl: fix memory leak

ammen99:
I have noticed this with LeakSanitizer, I hope these are all occurrences.


emersion (approved):
Nice catch!

PR #2540, opened 2020-12-12
backend/x11: use DRI3Open to get DRM FD 

emersion:
Instead of relying on EGL to retrieve the DRM FD, query it from the
DRI3 extension.

Use the EGL GBM platform, and drop the EGL config.

~~Depends on https://github.com/swaywm/wlroots/pull/2495~~


ammen99 (approved):
Looks good, I have tested this in both real X11 and Xwayland, no noticeable change :)
Thanks!

ammen99:
And needs a rebase ...

emersion:
Rebased. Thanks for the review!

PR #2545, opened 2020-12-13
backend/x11: implement a real rendering loop

emersion:
This is a respin of https://github.com/swaywm/wlroots/pull/1894

~~Depends on https://github.com/swaywm/wlroots/pull/2544~~

- [x] ~~Fix stuttering when playing a video with mpv (maybe need to fill the msc field in the present request?)~~ This is due to https://github.com/swaywm/wlroots/issues/2566, likely an mpv bug.

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -21,13 +21,6 @@
 #include "util/signal.h"
 #include "util/time.h"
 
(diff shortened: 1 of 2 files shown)


emersion:
Turns out the mpv issue was a red herring.

Issue #2543, opened 2020-12-13
Race condition when closing popups


dkondor:
This is a second incarnation of #2432 -- the compositor crash I reported there was fixed by #2454 but now I'm seeing the client get an error in the same conditions.

### Steps to reproduce:
1. Download the test program from here: https://github.com/dkondor/wlroots/blob/issue-2432/issue-2432/wlr-crasher.c
2. Compile it (a meson.build is in the same directory -- it can be compiled with meson independently of wlroots, by running everything inside in this dir)
3. Run the resulting wlrc program
4. Click on the blue popup (if there is no blue popup, just exit and try again -- it is not shown in some cases)

Expected behavior: the child process is focused and the popup is closed by the compositor

Actual behavior: the parent process exits with the following error: `xdg_surface@16: error 3: xdg_surface has never been configured`

### Alternative, real-world case

My original problem with cairo-dock (from [this](https://github.com/dkondor/cairo-dock-core/tree/wayland_new) branch) as described in #2432 -- instead of the compositor crashing, now only cairo-dock crashes when clicking on the magnifier. This does not always happen, but can be observed fairly regularly.

### Notes

Tested on Wayfire master + wlroots master. The strange thing is that I was testing this originally when #2454 was merged and could not see this problem. Now I've tried again, using wlroots 9595f95452a1d7420bbb190dc29c3806c29ca41e + Wayfire [a36b931](https://github.com/WayfireWM/wayfire/commit/a36b9313e57e7dad1a1fdd86c987225bbe403e27) and I can see the problem in that case. So I'm confused if this is a regression or if I just missed this previously (although I'm quite certain it was working at that time).

[...]

PR #2551, opened 2020-12-17
Fix for #2543

dkondor:
This makes handling of inert popups explicit. Tested with Wayfire; fixes my problems in #2543


types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -362,8 +366,12 @@ void handle_xdg_surface_commit(struct wlr_surface *wlr_surface) {
 	}
 
 	switch (surface->role) {
+	case WLR_XDG_SURFACE_ROLE_INERT_POPUP:
+		// inert popup
+		return;
 	case WLR_XDG_SURFACE_ROLE_NONE:
-		// inert toplevel or popup
+		// should not happen
+		wlr_log(WLR_ERROR, "xdg_surface %p commit with role == NONE", surface);
```
travankor, line 374: should there be an assert here instead of an error?

dkondor: I wasn't sure if it should be an assert or a protocol error. I feel a buggy client could trigger this if committing without creating a role object? So in this case a protocol error might be better. What do you think?

ammen99: If a client manages to trigger that (and it is indeed forbidden by the protocol), then the client should be sent an error and disconnected.

If it is simply because of inert objects, there is no error at all.

dkondor: Changed this to an error, since inert popups are handled separately. I did not actually see any case that triggers this now; I might try to do a test with a client that intentionally tries to trigger this error.

dkondor:
I have another question: is it allowed to re-create an xdg_surface? E.g. assign it to a popup, destroy it, and then assign it to an xdg_toplevel? If yes, this will be a problem [here](https://github.com/dkondor/wlroots/blob/06232961dec1fbcb370138dd0c3e9707e14b8547/types/xdg_shell/wlr_xdg_toplevel.c#L502), so that would need to be adjusted as well.

ammen99:
> I have another question: is it allowed to re-create an xdg_surface? E.g. assign it to a popup, destroy it, and then assign it to an xdg_toplevel? If yes, this will be a problem [here](https://github.com/dkondor/wlroots/blob/06232961dec1fbcb370138dd0c3e9707e14b8547/types/xdg_shell/wlr_xdg_toplevel.c#L502), so that would need to be adjusted as well.

It is possible to destroy the xdg_toplevel and recreate it (but not to switch its role afaik), see also the review on my previous patches: https://github.com/swaywm/wlroots/pull/2454#discussion_r518594072

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -96,6 +96,7 @@ enum wlr_xdg_surface_role {
 	WLR_XDG_SURFACE_ROLE_NONE,
 	WLR_XDG_SURFACE_ROLE_TOPLEVEL,
 	WLR_XDG_SURFACE_ROLE_POPUP,
+	WLR_XDG_SURFACE_ROLE_INERT_POPUP,
```
emersion, line 99: Not a fan of this. We can already check whether a surface is an inert popup: `xdg_surface->role == WLR_XDG_SURFACE_ROLE_NONE && surface->role == &xdg_popup_surface_role`.

dkondor: I understand and was thinking what would be an alternative. Thanks for the suggestion, I'll make an alternate version that way.

dkondor:
Hi,
I just remembered this issue and thought to check in.
I think I was able to address the main concern about adding an enum value. Also, with this patch, everything is working reliably for me, fixing #2543. Is there a chance to get this reviewed again? Let me know if there is anything more needed.
Thanks!

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2551

PR #2555, opened 2020-12-18
Keyboard interactivity options

dkondor:
Implementation for https://github.com/swaywm/wlr-protocols/pull/100

The goal of this is to enable layer-shell clients to have "regular" keyboard focus semantics. This can make it easier to implement layer-shell clients (no need to adjust the keyboard-interactivity setting when needed); also, it allows better support for keyboard interaction on the main layer-shell surface, including being notified of keyboard notifiers along with mouse clicks.

For clients using the old version of the protocol, any value is still interpreted as requesting exclusive focus. For clients with using the new protocol version, the compositor can interpret the given value. according to its policies.

Question: should there be a way for compositors to specify which version to use in wlr_layer_shell_v1_create()?

* * *

Breaking change: `wlr_layer_surface_v1_state.keyboard_interactivity` has now the type `enum zwlr_layer_surface_v1_keyboard_interactivity` to reflect protocol changes.

### DIFF
--- a/examples/layer-shell.c
+++ b/examples/layer-shell.c
@@ -50,7 +50,8 @@ static int32_t margin_top = 0;
 static double alpha = 1.0;
 static bool run_display = true;
 static bool animate = false;
-static bool keyboard_interactive = false;
+static enum zwlr_layer_surface_v1_keyboard_interactivity keyboard_interactive =
+	ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_NONE;
 static double frame = 0;
 static int cur_x = -1, cur_y = -1;
 static int buttons = 0;
@@ -453,8 +454,8 @@ static void handle_global(void *data, struct wl_registry *registry,
 				&wl_seat_interface, 1);
 		wl_seat_add_listener(seat, &seat_listener, NULL);
 	} else if (strcmp(interface, zwlr_layer_shell_v1_interface.name) == 0) {
-		layer_shell = wl_registry_bind(
-				registry, name, &zwlr_layer_shell_v1_interface, 1);
+		layer_shell = wl_registry_bind(registry, name,
+			&zwlr_layer_shell_v1_interface, version < 4 ? version : 4);
 	} else if (strcmp(interface, xdg_wm_base_interface.name) == 0) {
 		xdg_wm_base = wl_registry_bind(
 				registry, name, &xdg_wm_base_interface, 1);
@@ -478,7 +479,7 @@ int main(int argc, char **argv) {
 	int32_t margin_right = 0, margin_bottom = 0, margin_left = 0;
 	bool found;
 	int c;
-	while ((c = getopt(argc, argv, "knw:h:o:l:a:x:m:t:")) != -1) {
+	while ((c = getopt(argc, argv, "k:nw:h:o:l:a:x:m:t:")) != -1) {
 		switch (c) {
 		case 'o':
 			output = atoi(optarg);
@@ -558,9 +559,29 @@ int main(int argc, char **argv) {
 		case 'n':
 			animate = true;
 			break;
-		case 'k':
-			keyboard_interactive = true;
+		case 'k': {
+			const struct {
+				const char *name;
+				enum zwlr_layer_surface_v1_keyboard_interactivity value;
+			} kb_int[] = {
+				{ "none", ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_NONE },
+				{ "exclusive", ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE },
+				{ "on_demand", ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_ON_DEMAND }
+			};
+			found = false;
+			for (size_t i = 0; i < sizeof(kb_int) / sizeof(kb_int[0]); ++i) {
+				if (strcmp(optarg, kb_int[i].name) == 0) {
+					keyboard_interactive = kb_int[i].value;
+					found = true;
+					break;
+				}
+			}
+			if (!found) {
+				fprintf(stderr, "invalid keyboard interactivity setting %s\n", optarg);
+				return 1;
+			}
 			break;
+		}
 		default:
 			break;
 		}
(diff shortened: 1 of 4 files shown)


dkondor:
Matching Wayfire PR: https://github.com/WayfireWM/wayfire/pull/962

types/wlr_layer_shell_v1.c
```diff
@@ -156,7 +156,18 @@ static void layer_surface_handle_set_keyboard_interactivity(
 	if (!surface) {
 		return;
 	}
-	surface->client_pending.keyboard_interactive = !!interactive;
+
+	if (wl_resource_get_version(resource) <= 3) {
```
emersion, line 160: We can use `ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_ON_DEMAND_SINCE_VERSION` instead (yeah, the constant name is a little bit long…).

types/wlr_layer_shell_v1.c
```diff
@@ -156,7 +156,18 @@ static void layer_surface_handle_set_keyboard_interactivity(
...
+		surface->client_pending.keyboard_interactive = !!interactive;
+	} else {
+		if (interactive > 2) {
```
emersion, line 163: Maybe replace with `> ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_ON_DEMAND`?

types/wlr_layer_shell_v1.c
```diff
@@ -156,7 +156,18 @@ static void layer_surface_handle_set_keyboard_interactivity(
...
+			wl_resource_post_error(resource,
+			ZWLR_LAYER_SURFACE_V1_ERROR_INVALID_KEYBOARD_INTERACTIVITY,
```
emersion, line 165: Style: indent continuation lines with an extra tab:

```c
wl_resource_post_error(resource,
	ZWLR_LAYER_SURFACE_V1_ERROR_INVALID_KEYBOARD_INTERACTIVITY, …);
```

include/wlr/types/wlr_layer_shell_v1.h
```diff
@@ -49,7 +49,7 @@ struct wlr_layer_surface_v1_state {
 	struct {
 		uint32_t top, right, bottom, left;
 	} margin;
-	bool keyboard_interactive;
+	uint32_t keyboard_interactive;
```
emersion, line 52: We can now use `enum zwlr_layer_surface_v1_keyboard_interactivity` instead here.

examples/layer-shell.c
```diff
@@ -453,8 +454,11 @@ static void handle_global(void *data, struct wl_registry *registry,
 				&wl_seat_interface, 1);
 		wl_seat_add_listener(seat, &seat_listener, NULL);
 	} else if (strcmp(interface, zwlr_layer_shell_v1_interface.name) == 0) {
+		if (zwlr_layer_shell_v1_interface.version < (int)version) {
```
emersion, line 457: Why the cast to `int`?

We can't auto-magically use `zwlr_layer_shell_v1_interface.version` here. If there is ever a v5, we don't want to automatically use v5 when re-compiling this example. We should explicitly cap the version with 4 instead.

examples/layer-shell.c
```diff
@@ -558,9 +562,29 @@ int main(int argc, char **argv) {
 		case 'n':
 			animate = true;
 			break;
-		case 'k':
-			keyboard_interactive = true;
+		case 'k': {
+			struct {
```
emersion, line 566: This can be `const`.

types/wlr_layer_shell_v1.c
```diff
@@ -498,7 +509,7 @@ struct wlr_layer_shell_v1 *wlr_layer_shell_v1_create(struct wl_display *display)
 	}
 
 	struct wl_global *global = wl_global_create(display,
-		&zwlr_layer_shell_v1_interface, 2, layer_shell, layer_shell_bind);
+		&zwlr_layer_shell_v1_interface, zwlr_layer_shell_v1_interface.version, layer_shell, layer_shell_bind);
```
emersion, line 512: We can't advertise `zwlr_layer_shell_v1_interface.version` here, because re-compiling with an XML file that has v5 won't make the wlroots code handle v5 properly.

types/wlr_layer_shell_v1.c
```diff
@@ -498,7 +509,7 @@ struct wlr_layer_shell_v1 *wlr_layer_shell_v1_create(struct wl_display *display)
 	}
 
 	struct wl_global *global = wl_global_create(display,
-		&zwlr_layer_shell_v1_interface, 2, layer_shell, layer_shell_bind);
```
emersion, line 501: Hm, wierd, we only advertise v2? A destroy request was added in v3: https://github.com/swaywm/wlr-protocols/commit/23490a04798bd0ee5d927f3f717087ded416f391

It seems like we implemented the new request, but forgot to bump the version: https://github.com/swaywm/wlroots/commit/45c0877e3433f2b057f9125a8bc735057e088e6c#diff-8feb2aaa9e09a85d35f888167858b8c51abb54a3dcd2e676279a8c4a5d67936c

cc @ifreund

emersion:
Also need to update the protocol XML to the latest version.

types/wlr_layer_shell_v1.c
ifreund: Good catch: #2608

types/wlr_layer_shell_v1.c
dkondor: Ok, I just rebased on master after #2608 -- I'm still wondering if it would make sense to allow a compositor to request a lower version than available in wlroots (so if it does not support the new version, only the lower version will be advertised to clients). It would still be a breaking change though since that would need to be an extra parameter for this function.

dkondor:
I've addressed the comments, aside from my question about letting the compositor set the version it wants to support; if you think it's worth it, I can try to add a separate interface for that.

emersion (approved):
LGTM, can you squash everything into a single commit?

types/wlr_layer_shell_v1.c
emersion: Our policy so far has been that compositors always need to support the version supported by wlroots.

In this particular case it's not a big deal because compositors are not required to focus layer surfaces.

dkondor:
> LGTM, can you squash everything into a single commit?

Great, I just squashed it, thanks!

emersion:
Thanks!

PR #2556, opened 2020-12-18
backend/drm: remove special linear case for cursor plane

emersion:
_See individual commits._


ammen99 (approved):
Tested it, seems to work (and also LGTM).
Thanks!

PR #2558, opened 2020-12-18
xwayland: avoid crash on repeated server_finish_display() call

ifreund:
This function may end up being called more than once if the Xwayland
binary does not exist on the system.

Fixes https://github.com/ifreund/river/issues/167

### DIFF
--- a/xwayland/server.c
+++ b/xwayland/server.c
@@ -164,6 +164,7 @@ static void server_finish_display(struct wlr_xwayland_server *server) {
 	}
 
 	wl_list_remove(&server->display_destroy.link);
+	wl_list_init(&server->display_destroy.link);
 
 	if (server->display == -1) {
 		return;


emersion:
Hm, I don't see how we can end up in this state, `server_finish_display` is supposed to be called exactly once.

If Xwayland doesn't exist, `xserver_handle_ready` will be called and `waitpid` will return a non-zero `stat_val`. Then `server_finish_process` and `server_finish_display` will be called, and nothing else should happen?

ifreund:
The issue is that river isn't aware that this error has occurred and calls `wlr_xwayland_destroy()` on shutdown, which calls `server_finish_display()` again. Is there a way I should be checking if xwayland failed to init? `wlr_xwayland_create()` returns cleanly in this case which makes me think that calling `wlr_xwayland_destroy()` should be safe no matter what...

emersion (approved):
LGTM, thanks!

PR #2562, opened 2020-12-19
backend/wayland: remove EGL config

emersion:
We don't use EGLSurface anymore, so we don't need to choose an EGL
config anymore.

This is the last EGL config we use in wlroots.


ammen99 (approved):
LGTM, and seems to work.

PR #2569, opened 2020-12-23
drm: add fallback drmModeAddFB for drivers which do not support drmModeAddFB2

kaniini:
This pull request implements a fallback `get_fb_for_bo_legacy` function which uses the legacy `drmAddModeFB` function to set up the framebuffer.

This is based on the logic in xf86-video-modesetting, and allows both Sway and my own compositor to run on older cards which otherwise support the necessary features for Wayland to function.  Tested on a machine with ATI Radeon 7500 AGP graphics.

### DIFF
--- a/backend/drm/util.c
+++ b/backend/drm/util.c
@@ -170,6 +170,35 @@ const char *conn_get_name(uint32_t type_id) {
 	}
 }
 
+static uint32_t get_fb_for_bo_legacy(struct gbm_bo *bo) {
+	struct gbm_device *gbm = gbm_bo_get_device(bo);
+
+	/* We only support this as a fallback of last resort for ARGB8888 visuals,
+	 * like xf86-video-modesetting does. This is necessary on BE machines. */
+	if (gbm_bo_get_format(bo) != GBM_FORMAT_ARGB8888 ||
+			gbm_bo_get_plane_count(bo) != 1) {
+		wlr_log(WLR_DEBUG,
+			"Invalid visual %x (%d planes) requested for legacy DRM framebuffer",
+			gbm_bo_get_format(bo), gbm_bo_get_plane_count(bo));
+		return 0;
+	}
+
+	int fd = gbm_device_get_fd(gbm);
+	uint32_t width = gbm_bo_get_width(bo);
+	uint32_t height = gbm_bo_get_height(bo);
+	uint32_t depth = 32;
+	uint32_t bpp = gbm_bo_get_bpp(bo);
+	uint32_t pitch = gbm_bo_get_stride(bo);
+	uint32_t handle = gbm_bo_get_handle(bo).u32;
+
+	uint32_t id = 0;
+	if (drmModeAddFB(fd, width, height, depth, bpp, pitch, handle, &id)) {
+		wlr_log_errno(WLR_ERROR, "Unable to add DRM framebuffer");
+	}
+
+	return id;
+}
+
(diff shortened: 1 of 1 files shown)


backend/drm/util.c
```diff
@@ -170,6 +170,25 @@ const char *conn_get_name(uint32_t type_id) {
 	}
 }
 
+static uint32_t get_fb_for_bo_legacy(struct gbm_bo *bo) {
+	struct gbm_device *gbm = gbm_bo_get_device(bo);
+
+	int fd = gbm_device_get_fd(gbm);
+	uint32_t width = gbm_bo_get_width(bo);
+	uint32_t height = gbm_bo_get_height(bo);
+	uint32_t depth = 24;
```
emersion, line 179: This depends on the format.

kaniini: Per discussion in IRC, we locked this to `GBM_FORMAT_ARGB8888` and changed depth to 32.  We've also validated this is working as expected on a machine where ADDFD2 is not working (due to the driver not supporting it on big-endian).

backend/drm/util.c
```diff
@@ -170,6 +170,30 @@ const char *conn_get_name(uint32_t type_id) {
 	}
 }
 
+static uint32_t get_fb_for_bo_legacy(struct gbm_bo *bo) {
+	struct gbm_device *gbm = gbm_bo_get_device(bo);
+
+	/* we only support this as a fallback of last resort for ARGB8888 visuals, like
+	 * xf86-video-modesetting does. */
+	if (gbm_bo_get_format(bo) == GBM_FORMAT_ARGB8888)
```
emersion, line 178: Shouldn't this be `!= GBM_FORMAT_ARGB8888`?

kaniini: oh, yes.

kaniini: I pushed a fix for that typo, we're going to test it on a few G4 and G5 systems.  I'll update this PR once we've finished testing this version of the patch.

kaniini: Alright, everything looks good!

backend/drm/util.c
```diff
@@ -170,6 +170,30 @@ const char *conn_get_name(uint32_t type_id) {
 	}
 }
 
+static uint32_t get_fb_for_bo_legacy(struct gbm_bo *bo) {
+	struct gbm_device *gbm = gbm_bo_get_device(bo);
+
+	/* we only support this as a fallback of last resort for ARGB8888 visuals, like
+	 * xf86-video-modesetting does. */
+	if (gbm_bo_get_format(bo) != GBM_FORMAT_ARGB8888)
```
emersion, line 178: Style: braces are mandatory

backend/drm/util.c
```diff
@@ -170,6 +170,30 @@ const char *conn_get_name(uint32_t type_id) {
...
+		return 0;
```
emersion, line 179: In this case, we don't print an error message (the caller prints a debug message only). We probably should.

backend/drm/util.c
emersion, line 178: Can we also add a safety `|| gbm_bo_get_plane_count(bo) != 1` check? Just in case a client submits a bad buffer.

emersion (changes requested):
LGTM apart from these minor comments!

kaniini:
All of these revisions should now be incorporated :)

emersion (approved):
Thanks!

PR #2575, opened 2020-12-27
backend/drm: handle unplug event

emersion:
This is half of the GPU hotplug work, see https://github.com/swaywm/wlroots/pull/2423.

To test:

```
# MINOR=1 triggers removal for card1
udevadm trigger --verbose --type=devices --action=remove --subsystem-match=drm --property-match="MINOR=1"
```

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -50,6 +50,7 @@ static void backend_destroy(struct wlr_backend *backend) {
 	wl_list_remove(&drm->display_destroy.link);
 	wl_list_remove(&drm->session_destroy.link);
 	wl_list_remove(&drm->session_active.link);
+	wl_list_remove(&drm->parent_destroy.link);
 	wl_list_remove(&drm->dev_change.link);
 
 	finish_drm_resources(drm);
@@ -144,6 +145,12 @@ static void handle_display_destroy(struct wl_listener *listener, void *data) {
 	backend_destroy(&drm->backend);
 }
 
+static void handle_parent_destroy(struct wl_listener *listener, void *data) {
+	struct wlr_drm_backend *drm =
+		wl_container_of(listener, drm, parent_destroy);
+	backend_destroy(&drm->backend);
+}
+
 struct wlr_backend *wlr_drm_backend_create(struct wl_display *display,
 		struct wlr_session *session, struct wlr_device *dev,
 		struct wlr_backend *parent) {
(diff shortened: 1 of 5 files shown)


neon64:
Is there a good way of debugging why a kernel module is still in use? Unfortunately with this patch after manually sending a remove udev event, and `ls /dev/dri/card*` showing only `/dev/dri/card0` (my iGPU), I get 

    rmmod: ERROR: Module nouveau is in use
    
which is weird because it doesn't look like this patch is doing much different compared to my last PR which allowed `rmmod` fine (although that PR was admittedly against wlroots from several months ago)

emersion:
`lsmod` will tell you know a kernel module is still in use. For instance, this indicates drm_kms_helper is still used by amdgpu:

```
Module                  Size  Used by
drm_kms_helper        274432  1 amdgpu
```

In my testing, `rmmod -f` was sometimes necessary because the "Used by" column indicated nothing and `lsof /dev/dri/card0` revealed nothing.

emersion:
This sounds unrelated. Please open a separate issue.

neon64:
so I've implemented the other half of this here: https://github.com/neon64/wlroots/tree/drm-hotplug . Should I make a separate PR to wlroots which includes/depends on the changes in this one?

emersion:
>Should I make a separate PR to wlroots which includes/depends on the changes in this one?

Nice! Sure, feel free to open a new PR.

Have you run into any bug with this PR?

neon64:
I still haven't fully diagnosed why nouveau is still in-use unfortunately. `lsof /dev/dri*` shows only Intel stuff, and `lsmod` shows no modules depending on nouveau, but also the refcount is `5`, not `0`, so I think `rmmod -f` isn't  safe to do.

```
nouveau              2359296  5
mxm_wmi                16384  1 nouveau
ttm                   114688  3 drm_vram_helper,drm_ttm_helper,nouveau
wmi                    36864  5 intel_wmi_thunderbolt,thinkpad_wmi,wmi_bmof,mxm_wmi,nouveau
video                  53248  3 thinkpad_acpi,i915,nouveau
i2c_algo_bit           16384  2 i915,nouveau
drm_kms_helper        274432  4 drm_vram_helper,vboxvideo,i915,nouveau
drm                   569344  30 drm_kms_helper,drm_vram_helper,vboxvideo,drm_ttm_helper,i915,ttm,nouveau
agpgart                53248  4 intel_gtt,ttm,nouveau,drm
```

After some more testing, I've worked out that if, before trying to destroy the card1 backend, I pull out the HDMI cable, then the refcount drops from 5 to 1, and then after running `sudo udevadm trigger --verbose --action=remove --subsystem-match=drm --property-match="MINOR=0"`, the refcount drops from 1 to 0, and rmmod works.

If I don't disconnect the physical cable before using udevadm, then the refcount stays at 5 as long as sway is open. So my theory is that there's something missing in the cleanup code for outputs / drm-connectors, if you don't disconnect the output before the backend destroy.

May or may not be relevant from [sway.log](https://github.com/swaywm/wlroots/files/5825896/sway-not-releasing-nouveau.txt):
```
00:00:21.104 [DEBUG] [wlr] [backend/drm/drm.c:1064] connector HDMI-A-1: De-allocating CRTC 3
00:00:21.104 [ERROR] [wlr] [backend/drm/legacy.c:42] connector HDMI-A-1: Failed to set DPMS property: Permission denied
00:00:21.154 [ERROR] [wlr] [backend/session/logind.c:120] Failed to release device '9': Device not taken
```

emersion:
> If I don't disconnect the physical cable before using udevadm, then the refcount stays at 5 as long as sway is open. So my theory is that there's something missing in the cleanup code for outputs / drm-connectors, if you don't disconnect the output before the backend destroy.

Hmm. Yeah, if the commit fails, we don't destroy the FBs in `dealloc_crtc` (it stops at `drm_crtc_commit`). Does it work if you comment the `return` for `drm_crtc_commit` failures?

neon64:
> Does it work if you comment the return for drm_crtc_commit failures?

Yep that appears to work - refcount drops to zero and `rmmod nouveau` works.
Is there a downside to continuing through the rest of the function if the commit fails?

emersion:
The main thing I'm worried about is if this happens prior to shutdown and we try to re-use the CRTC for another connector later on. The kernel would return an error in this case (EINVAL for atomic).

We _could_ only ignore errors when we're shutting down the backend, but not sure it's worth it. In all cases, something will be printed to the logs.

In any case, feel free to submit a pull request.

emersion:
Pushed the fix for EPERM.

Issue #2580, opened 2020-12-28
Examples don't run, could not initialize EGL


fwSmit:
Most examples don't run (in sway).
```
00:00:02.118 [EGL] command: eglQueryDeviceStringEXT, error: EGL_BAD_DISPLAY (0x3008), message: "(null)"
00:00:02.118 [render/egl.c:315] eglQueryDeviceStringEXT(EGL_EXTENSIONS) failed
00:00:02.118 [EGL] command: eglMakeCurrent, error: EGL_BAD_DISPLAY (0x3008), message: "Invalid display (nil)"
00:00:02.120 [render/wlr_renderer.c:281] Could not initialize EGL
00:00:02.120 [backend/wayland/backend.c:325] Could not create renderer
```
I have all required packages installed, except libseat.

emersion:
Please provide full debug logs.

neon64:
This looks like it could be https://github.com/swaywm/wlroots/issues/2480 - the log output looks similar. Do you have any Nvidia  packages installed? If so, what worked for me was installing the latest libglvnd (libglvnd-git on Arch) which has [this patch](https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/235 ) merged

fwSmit:
> This looks like it could be #2480 - the log output looks similar. Do you have any Nvidia packages installed? If so, what worked for me was installing the latest libglvnd (libglvnd-git on Arch) which has [this patch](https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/235) merged

Thanks that works for me. I do have some nvidia packages installed apparently, so that would probably fix this issue as well.

> Please provide full debug logs.

I'll link the debug logs anyways, since I made them already: https://gist.github.com/fwSmit/1d0d9c67d8572b4f15414f934d8d2e94

PR #2581, opened 2020-12-28
Reset pointer focus on surface commit

fwSmit:
This should fix https://github.com/swaywm/sway/issues/5911. I want to test it, but I don't know how to. Do I need to recompile sway before testing? Most of the examples don't run (see #2580), so I can't test those.

### DIFF
--- a/include/wlr/types/wlr_seat.h
+++ b/include/wlr/types/wlr_seat.h
@@ -169,6 +169,7 @@ struct wlr_seat_pointer_state {
 	uint32_t grab_time;
 
 	struct wl_listener surface_destroy;
+	struct wl_listener surface_commit;
 
 	struct {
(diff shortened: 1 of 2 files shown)


ammen99:
I don't like this, wouldn't this be racy with the compositor updating focus on commit?

fwSmit:
I didn't know that was a job for the compositor. I'm not familiar with the codebase. The focus change on surface destroy is handled by wlroots thought (which is where I copied it from).

ammen99:
@fwSmit Yes, the compositor is responsible for setting and clearing the focus. In the case that a surface is unmapped, it is a bit different because the surface cannot continue to be focus anyway and the compositor has to set another focus.

fwSmit:
Alright, thanks for the info. This PR can be closed then.

PR #2582, opened 2020-12-28
types/seat: Clear focus in wlr_seat_destroy()

any1:
This fixes use-after-free in surface destroy signal listeners.

### DIFF
--- a/types/seat/wlr_seat.c
+++ b/types/seat/wlr_seat.c
@@ -159,6 +159,14 @@ void wlr_seat_destroy(struct wlr_seat *seat) {
 		return;
 	}
 
+	wlr_seat_pointer_clear_focus(seat);
+	wlr_seat_keyboard_clear_focus(seat);
+
+	struct wlr_touch_point *point;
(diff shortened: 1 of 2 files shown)


emersion:
I think this makes sense. We do something similar for unmap signals for xdg-shell, by guaranteeing that unmap is always triggered before destroy.

types/seat/wlr_seat.c
```diff
@@ -159,6 +159,14 @@ void wlr_seat_destroy(struct wlr_seat *seat) {
 		return;
 	}
 
+	wlr_seat_pointer_clear_focus(seat);
+	wlr_seat_keyboard_clear_focus(seat);
+
+	struct wlr_touch_point* point;
```
emersion, line 165: Style: `*` on the right

emersion (changes requested):
Can we also document this guarantee in the header?

types/seat/wlr_seat.c
any1: Done.

emersion (approved):
LGTM, thanks!

PR #2583, opened 2020-12-29
noop: listen to display destroy and destroy the backend

ammen99:
Not sure if this is needed, but I guess it is a good idea, since most objects which are related to the display are destroyed when it is destroyed?


ifreund (approved):
I'm in favor of this, but this is a breaking change. I currently manually destroy the noop backend so if this was merged without changing river's code there would be a double free.

emersion (approved):
LGTM, thanks!

PR #2584, opened 2020-12-29
Transient seat

any1:
This implements the transient seat protocol for creating short-lived seats to be used by remote clients such as via VNC or RDP.
A sample implementation exists here: https://github.com/any1/wayvnc/tree/multi-seat

This PR also currently includes some multi-seat related fixes which are also in another PR. They will be removed when the fixes have been merged.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2584

PR #2587, opened 2021-01-01
xwm: don't iterate output parent in has_parent

Emantor:
Commit 5012121d331c ("xwm: add loop detection for read_surface_parent")
introduced a beviour change, previsously we would correctly setup the
child parent relationship:
```
surface -> surface -> surface -> NULL
   |          ^
   +- parent -+
```
but now, since has_parent iterates on the pointer passe by reference,
our setup looks like this:
```
surface -> surface -> surface -> NULL
   |                     ^
   +------- parent ------+
```
Introduce a local pointer variable in has_parent to fix this issue.


Emantor:
Ignore this, I am too tired after new years eve…

PR #2588, opened 2021-01-02
Implement foreign-toplevel-info-v1

ifreund:
wayland-protocols MR: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/59

I'm not sure how best to namespace these within wlroots such that things are consistent and concise but there is no conflict with the current `wlr_foreign_toplevel` APIs. I've gone with a `wlr_ext` prefix for now but am hoping for suggestions.

PR #2589, opened 2021-01-02
Implement ext-foreign-toplevel-management-v1

ifreund:
wayland-protocols MR: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/60

I'm not sure how best to namespace these within wlroots such that things are consistent and concise but there is no conflict with the current wlr_foreign_toplevel APIs. I've gone with a wlr_ext prefix for now but am hoping for suggestions.

Depends on #2588

Implementation of both new protocols in river on [this branch](https://github.com/ifreund/river/tree/foreign-toplevel-management).


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2589

PR #2591, opened 2021-01-02
Export wlr buffer from output backends

any1:
This implements export-dmabuf and screencopy-dmabuf for all remaining output backends.

It seems that the wayland and x11 backends might delete the front buffer while its in use and I'm not sure what's the best way to fix that. Suggestions are welcome. The simplest solution would be to hold on to it until a new one is committed, but a fancier solution would be a reader reference count on the `wlr_buffer` object, which is probably overkill.


any1:
Superseded by #2615

Issue #2563, opened 2020-12-19
Drop EGL things from wlr_renderer_autocreate


emersion:
Currently `wlr_renderer_autocreate` takes parameters tied to EGL:

1. `EGLint *config_attribs, EGLint visual_id`: with https://github.com/swaywm/wlroots/pull/2562, no one uses those anymore.
2. `EGLenum platform, void *remote_display`: almost everyone uses the GBM platform. The only exception is the Wayland backend, which still has no other way to get the DRM FD used by the parent compositor.

The long-term solution for (2) is to get the DRM FD from dmabuf-hints, however the protocol is still pending. A more realistic solution would be to vendor a copy of the `wl_drm` protocol, which Mesa uses exactly for this.

Once that is done, we can think of picking a new signature for `wlr_renderer_autocreate`. The renderer needs to have access to the DRM FD to initialize itself. There may not be a DRM FD, e.g. on GPU-less setups, in which case we should allow `wlr_renderer_autocreate` to create a software renderer.

Possible new signatures:

```c
/**
 * Automatically create a renderer with the provided DRM FD.
 * Maybe this is too DRM-specific?
 */
struct wlr_renderer *wlr_renderer_autocreate(int drm_fd);

/**
 * Automatically create a renderer suitable for the backend.
 * The rendering device will be queried from the backend.
 */
struct wlr_renderer *wlr_renderer_autocreate(struct wlr_backend *backend);
```

emersion:
(1) is done in https://github.com/swaywm/wlroots/pull/2593.

The issue with the Wayland backend in (2) is fixed in https://github.com/swaywm/wlroots/pull/2592.

emersion:
https://github.com/swaywm/wlroots/pull/2602 removes the `struct wlr_egl *` argument from `wlr_renderer_autocreate`

bl4ckb0ne:
What about the `remote_display` parameter? Should the `wlr_gbm_allocator` be a part of the `wlr_backend` now that all of the backends use the GBM platform?

emersion:
> Should the wlr_gbm_allocator be a part of the wlr_backend now that all of the backends use the GBM platform?

I'd prefer not to add an interface to get the `gbm_device` from the backend. Only the DRM backend needs the GBM device to import buffers, other backends only need it for allocations. Eventually we'll want to completely remove the renderer and allocator from backends, so I'd prefer not to rely on the backend having a GBM allocator.

[...]

PR #2593, opened 2021-01-04
render: remove EGL config and visual from wlr_renderer_autocreate

emersion:
This isn't used anymore by any backend.

Some examples still provide an EGL config to wlr_egl_init, so we can't
drop it yet there.

References: https://github.com/swaywm/wlroots/issues/2563

* * *

Breaking change: `wlr_renderer_autocreate` no longer takes the parameters `EGLint *config_attribs` and `EGLint visual_id`.


ammen99 (approved):
LGTM

PR #2595, opened 2021-01-04
backend/wayland: re-use cursor wl_buffers

emersion:
We were importing cursor buffers as wl_buffers over and over again.
Instead, only import these once.


ammen99 (approved):
Thanks for the quick fix!

PR #2597, opened 2021-01-05
docs: wlr_surface_from_resource cannot return NULL

ifreund:


include/wlr/types/wlr_surface.h
```diff
@@ -256,6 +256,10 @@ void wlr_surface_send_frame_done(struct wlr_surface *surface,
  */
 void wlr_surface_get_extends(struct wlr_surface *surface, struct wlr_box *box);
 
+/**
+ * Get the wlr_surface corresponding to a wl_surface resource. This will
+ * never return NULL given a valid wl_surface resource created by wlroots.
```
emersion, line 261: This makes it sounds like this may return NULL if it's not a valid wl_surface resource. However we assert in this case, right?

ifreund: Indeed, this wording could be improved.

emersion (approved):
Thanks!

PR #2598, opened 2021-01-05
backend/wayland: Avoid uninitialized read

kennylevinsen:
keyboard_handle_leave would always process 1 keycode more than was
pending, which meant reading uninitialized memory from the "pressed"
array.

Found by valgrind.

### DIFF
--- a/backend/wayland/seat.c
+++ b/backend/wayland/seat.c
@@ -235,11 +235,12 @@ static void keyboard_handle_leave(void *data, struct wl_keyboard *wl_keyboard,
 
 	uint32_t time = get_current_time_msec();
 
-	uint32_t pressed[dev->keyboard->num_keycodes + 1];
+	size_t num_keycodes = dev->keyboard->num_keycodes;
+	uint32_t pressed[num_keycodes + 1];
 	memcpy(pressed, dev->keyboard->keycodes,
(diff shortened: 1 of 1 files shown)


backend/wayland/seat.c
```diff
@@ -235,11 +235,12 @@ static void keyboard_handle_leave(void *data, struct wl_keyboard *wl_keyboard,
 
 	uint32_t time = get_current_time_msec();
 
-	uint32_t pressed[dev->keyboard->num_keycodes + 1];
+	size_t num_keycodes = dev->keyboard->num_keycodes;
+	uint32_t pressed[num_keycodes];
```
emersion, line 239: VLAs have undefined behaviour when initialized with a zero length. We should either add +1 here, or assert that `num_keycodes > 0`.

emersion (commented):
Nice catch! I'm not even sure `sizeof` is guaranteed to work with VLAs.

backend/wayland/seat.c
kennylevinsen: Ah, I had forgotten that. I imagine num_keycodes can be zero, so let's just keep the +1 on the array length as it's inconsequential.

emersion (approved):
Thanks!

PR #2602, opened 2021-01-06
Remove wlr_egl from backends

bl4ckb0ne:
Removes the `wlr_egl` member from various backend to keep it into the `wlr_renderer`. This also removes various includes of the `render/egl.h` header that were made useless by this patch.

References: https://github.com/swaywm/wlroots/issues/2563

* * *

Breaking change: `wlr_renderer_autocreate` no longer takes a `struct wlr_egl *` as parameter.

### DIFF
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -71,10 +71,11 @@ static void backend_destroy(struct wlr_backend *wlr_backend) {
 	wlr_signal_emit_safe(&wlr_backend->events.destroy, backend);
 
 	free(backend->format);
-	if (backend->egl == &backend->priv_egl) {
+
+	if (!backend->has_parent_renderer) {
 		wlr_renderer_destroy(backend->renderer);
-		wlr_egl_finish(&backend->priv_egl);
 	}
+
 	wlr_allocator_destroy(backend->allocator);
 	free(backend);
 }
@@ -114,7 +115,6 @@ static bool backend_init(struct wlr_headless_backend *backend,
 
 	backend->allocator = allocator;
 	backend->renderer = renderer;
-	backend->egl = wlr_gles2_renderer_get_egl(renderer);
 
 	const struct wlr_drm_format_set *formats =
 		wlr_renderer_get_dmabuf_render_formats(backend->renderer);
@@ -206,7 +206,7 @@ struct wlr_backend *wlr_headless_backend_create(struct wl_display *display) {
 		goto error_backend;
 	}
 
-	struct wlr_renderer *renderer = wlr_renderer_autocreate(&backend->priv_egl,
+	struct wlr_renderer *renderer = wlr_renderer_autocreate(
 		EGL_PLATFORM_GBM_KHR, gbm_alloc->gbm_device);
 	if (!renderer) {
 		wlr_log(WLR_ERROR, "Failed to create renderer");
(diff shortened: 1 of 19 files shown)


render/wlr_renderer.c
```diff
@@ -276,6 +275,9 @@ struct wlr_renderer *wlr_renderer_autocreate(struct wlr_egl *egl,
 		config_attribs = all_config_attribs;
 	}
 
+	struct wlr_egl *egl = calloc(1, sizeof(*egl));
```
emersion, line 278: This memory is leaked when the renderer is destroyed.

I think long-term we'll want to either:

* Allocate the `wlr_egl` as part of the `wlr_gles2_renderer`. This means `wlr_gles2_renderer_create` would need to take all arguments necessary for the call to `wlr_egl_init`.
* Make `wlr_gles2_renderer_create` take over the `wlr_egl` (kind of like you're doing in this PR), and replace `wlr_egl_init`/`wlr_egl_finish` with `wlr_egl_create`/`wlr_egl_destroy` so that the GLES2 renderer can cleanly free the `wlr_egl` (this would ensuring it's not part of a larger structure).

For this PR we can probably just `free(renderer->egl)` when destroying the GLES2 renderer, and properly document that `wlr_gles2_renderer_create` takes ownership of the `wlr_egl`?

backend/headless/backend.c
```diff
@@ -71,10 +70,7 @@ static void backend_destroy(struct wlr_backend *wlr_backend) {
 	wlr_signal_emit_safe(&wlr_backend->events.destroy, backend);
 
 	free(backend->format);
-	if (backend->egl == &backend->priv_egl) {
-		wlr_renderer_destroy(backend->renderer);
-		wlr_egl_finish(&backend->priv_egl);
-	}
+	wlr_renderer_destroy(backend->renderer);
```
emersion, line 73: The headless backend is a little special and can be created with a parent renderer. If that's the case, we shouldn't destroy the renderer here, as it's owned by the parent backend.

We could probably add a `bool has_parent_renderer` field to the backend, and skip destroying the renderer if it's true.

emersion (changes requested):
Thanks for working on this! Here are a few comments.

This needs to be rebased on top of https://github.com/swaywm/wlroots/pull/2593.

include/wlr/render/interface.h
```diff
@@ -19,8 +19,6 @@
 #define EGL_NO_PLATFORM_SPECIFIC_TYPES
 #endif
 
-#include <EGL/egl.h>
-#include <EGL/eglext.h>
```
emersion, line 23: We can get rid of the various `MESA_*` and `EGL_*` defines above now.

emersion:
Also, is it possible to get rid of the unnecessary includes in a separate commit?

render/wlr_renderer.c
bl4ckb0ne: Huh, I was sure I added the `free` in the `wlr_renderer_destroy` function. Good catch.
I like the idea of `wlr_gles2_renderer` becoming the owner of `wlr_egl`, it makes more sense when we will add the pixman backend. 

I'll slap a nice comment and a TODO.

backend/headless/backend.c
bl4ckb0ne: Missed that, I just blindly removed both of the `wlr_egl`. Thanks

bl4ckb0ne:
There you we go, I can reword the commit if you want, I put in something generic since it affects two both the backend and the render folder.

render/wlr_renderer.c
emersion: OK. I slightly edited the PR to also add a note in the header defining `wlr_gles2_renderer_create`.

emersion (approved):
LGTM, thanks!

PR #2603, opened 2021-01-06
CI builds tinywl

bl4ckb0ne:
The ci now builds tinywl, which requires wlroots to be installed first.

### DIFF
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -18,11 +18,14 @@ sources:
 tasks:
   - setup: |
       cd wlroots
-      CC=gcc meson build-gcc -Dauto_features=enabled -Dlogind-provider=systemd
+      CC=gcc meson build-gcc -Dauto_features=enabled -Dlogind-provider=systemd --prefix /usr
       CC=clang meson build-clang -Dauto_features=enabled -Dlogind-provider=systemd
   - gcc: |
       cd wlroots/build-gcc
       ninja
+      sudo ninja install
+      cd ../tinywl
+      make
   - clang: |
       cd wlroots/build-clang
       ninja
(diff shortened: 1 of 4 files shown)


tinywl/tinywl.c
```diff
@@ -846,7 +846,7 @@ int main(int argc, char *argv[]) {
 	 * backend uses the renderer, for example, to fall back to software cursors
 	 * if the backend does not support hardware cursors (some older GPUs
 	 * don't). */
-	server.backend = wlr_backend_autocreate(server.wl_display, NULL);
+	server.backend = wlr_backend_autocreate(server.wl_display);
```
emersion, line 849: Can we remove the comment "The NULL argument here optionally allows you […]"?

.builds/archlinux.yml
```diff
@@ -18,11 +18,21 @@ sources:
...
+      cd ../tinywl
+      make
+      cd ../build-gcc
+      sudo ninja uninstall
   - clang: |
       cd wlroots/build-clang
       ninja
+      sudo ninja install
+      cd ../tinywl
+      make
```
emersion, line 36: Here tinywl is built twice with GCC, because `CC` is only set for the `meson` commands. Also I think the wlroots installation won't change a lot between GCC and Clang.

We could only build tinywl with GCC I think, because Clang is already used on FreeBSD.

emersion (changes requested):
Thanks for working on this!

tinywl/tinywl.c
bl4ckb0ne: Oh sure, I squashed the fix into the first tinywl commit.

.builds/archlinux.yml
bl4ckb0ne: Oh right, tinywl is not built with meson. I removed the tinywl build from it and squashed the fix into the ci commit.

emersion (approved):
LGTM, thanks!

Issue #2604, opened 2021-01-06
Correctly handle ICCCM globally active input model


BrassyPanache:
This issue stems from swaywm/sway#3007.

I've found that from 2020.3 onwards IntelliJ sets the input field of WM_HINTS to **false** for certain dialogs.  This is allowable under the ICCCM specifications (see https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.7).  If the input model is globally active the window receives passive input and then issues a SetInputFocus request to receive focus.  However under Wayland there is no mechanism to access input passively.  This leads to the situation under Sway where these dialogs cannot receive user input.

I've created a rather broad fix (perhaps overly so) which I've tested with:

- Chrome
- IntelliJ
- Alacritty
- VSCode

I have not seen any major issues.  The fix is to always enable the input flag.  My rationale is as follows:

1. If wlroots is not compiled with ICCCM support this is the default so it must be somewhat reasonable otherwise I'd expect to see some open issues.
2. If a window really does not require input and receives it this is somewhat easier for the end-user to navigate around rather than the current situation whereby the window can never receive input.

If this fix is too broad then perhaps we can at least open the discussion on how this can be better handled.  The current situation leaves JetBrains products rather unusable.

luispabon:
Do you have the fix you mention on a branch somewhere? I'd love to help testing this out.

emersion:
We had a very hard time getting focus to work somewhat correctly with X11 windows. We'll need extensive testing before merging anything.

Some windows _really_ don't want input focus, like tooltips. How does your patch handle them?

In the spec, with "Globally Active Input" the input field is set to False but WM_TAKE_FOCUS is enabled. Can't we special-case it?

BrassyPanache:
I think supporting special-cases would make sense.  I agree that you don't want tooltips to take focus. I'll have more of a look at this.

Haardt:
Hi,
your patch breaks another behavior. 'Normal' child windows from intelli no longer get the focus. If the mouse not over the new window i can't use shortcuts (for example the quit dialog and 'enter').

BrassyPanache:
Thanks for testing, @Haardt.

Just following up after some further investigation.  It appears as though the dialog boxes in question do not set WM_TAKE_FOCUS.  The relevant logging can be seen below (there are no protocols configured).

```
[...]

PR #2605, opened 2021-01-06
Expose ICCCM input status

BrassyPanache:
In certain situations windows can have their input field set to false.
In these cases windows should only receive focus if WM_TAKE_FOCUS
is enabled. Otherwise the window should never receive focus.

Relates to #2604

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -120,6 +120,18 @@ struct wlr_xwayland_surface_size_hints {
 };
 
 /**
+ * This represents the input focus described as follows:
+ * 
+ * https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#input_focus
+ */
+enum wlr_xwayland_icccm_input_model {
+	WLR_ICCCM_INPUT_MODEL_NONE = 0,
+	WLR_ICCCM_INPUT_MODEL_PASSIVE = 1,
+	WLR_ICCCM_INPUT_MODEL_LOCAL = 2,
+	WLR_ICCCM_INPUT_MODEL_GLOBAL = 3,
+};
+
+/**
  * An Xwayland user interface component. It has an absolute position in
  * layout-local coordinates.
  *
@@ -303,7 +315,10 @@ void wlr_xwayland_surface_ping(struct wlr_xwayland_surface *surface);
  *          false if it should be ignored
  */
 bool wlr_xwayland_or_surface_wants_focus(
-	const struct wlr_xwayland_surface *surface);
+	const struct wlr_xwayland_surface *xsurface);
 
+
+enum wlr_xwayland_icccm_input_model wlr_xwayland_icccm_input_model(
+	const struct wlr_xwayland_surface *xsurface);
 
 #endif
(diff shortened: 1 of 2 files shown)


tsdh:
I've test-driven this PR and the corresponding sway PR.  See https://github.com/swaywm/sway/pull/5948#issuecomment-760813301.

include/wlr/xwayland.h
```diff
@@ -119,6 +119,13 @@ struct wlr_xwayland_surface_size_hints {
 	uint32_t win_gravity;
 };
 
+enum wlr_xwayland_icccm_input_state {
```
emersion, line 122: Can we add a link to https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#input_focus here so that people can figure out what each model means?

include/wlr/xwayland.h
```diff
@@ -119,6 +119,13 @@ struct wlr_xwayland_surface_size_hints {
...
+	ICCCM_INPUT_STATE_NONE = 0,
```
emersion, line 123: Please prefix each symbol with `WLR_`, to avoid potential conflicts with another library.

include/wlr/xwayland.h
```diff
@@ -303,7 +310,10 @@ void wlr_xwayland_surface_ping(struct wlr_xwayland_surface *surface);
  *          false if it should be ignored
  */
 bool wlr_xwayland_or_surface_wants_focus(
-	const struct wlr_xwayland_surface *surface);
+	const struct wlr_xwayland_surface *xsurface);
+
 
+enum wlr_xwayland_icccm_input_state wlr_xwayland_icccm_input_state(
```
emersion, line 316: Can we rename `input_state` to `input_model`? That would match the wording used by the spec and makes it easier to search for its meaning.

include/wlr/xwayland.h
BrassyPanache: No problem.

xwayland/xwm.c
```diff
@@ -1980,12 +1980,32 @@ bool wlr_xwayland_or_surface_wants_focus(
...
+				xsurface->window_type_len, needles[i])) {
+			return false;
 		}
 	}
 
-	return ret;
+	return true;
+}
+
+enum wlr_xwayland_icccm_input_state wlr_xwayland_icccm_input_model(
```
emersion, line 1994: Can we rename the enum as well?

include/wlr/xwayland.h
```diff
@@ -119,6 +119,18 @@ struct wlr_xwayland_surface_size_hints {
 	uint32_t win_gravity;
 };
 
+/**
+ * This represents the input focus described as follows:
+ * 
+ * https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#input_focus
+ */
+enum wlr_xwayland_icccm_input_model {
+	WLR_ICCCM_INPUT_STATE_NONE = 0,
```
emersion, line 128: One last rename for these please :P

emersion (approved):
LGTM, thanks!

emersion:
Now that I think about it, maybe `wlr_xwayland_or_surface_wants_focus` is a bad workaround and we should use `wlr_xwayland_icccm_input_model` for override-redirect windows as well?

BrassyPanache:
I'd be happy to undertake the attempt.

PR #2607, opened 2021-01-07
xwayland/selection: ignore incr transfer requests during incr transfer

### LINKED ISSUE #2548: Sway crashes when an XWayland-client copies large contents to the clipboard and the copyq clipboard manager is running
**Sway Version:**
```
❯ swaymsg -t get_version
sway version 1.5-a1591c23 (Dec  9 2020, branch 'master')
```

To reproduce this issue (and to get debug symbols), I installed https://aur.archlinux.org/packages/sway-git/ and https://aur.archlinux.org/packages/wlroots-git/, and both should be using the current master branch.

**Debug Log:**
I obtained the following logs using `sway -d 2> ~/sway.log`:
https://gist.github.com/zsolt-donca/62f7348ad71de1b566198109612eedef

**Configuration File:**
I reproduced this issue with the default configuration, with the exception of some keybinding coming from `/etc/sway/config.d/swayshot`, and of `/etc/sway/config.d/50-systemd-user.conf`.

**Stack Trace:**
```
#0  0x00007f26711255c0 in xcb_get_property_value_length () at /usr/lib/libxcb.so.1
swaywm/sway#1  0x00007f26711f0b8b in xwm_data_source_write (fd=90, mask=<optimized out>, data=0x564108ffa5f0) at ../xwayland/selection/incoming.c:22
        transfer = 0x564108ffa5f0
        xwm = 0x564108ffa490
        property = 0x20 <error: Cannot access memory at address 0x20>
        remainder = <optimized out>
        len = <optimized out>
swaywm/sway#2  0x00007f267123efaa in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
swaywm/sway#3  0x00007f267123d4e7 in wl_display_run () at /usr/lib/libwayland-server.so.0
swaywm/sway#4  0x00005641074c0422 in server_run (server=<optimized out>) at ../sway/sway/server.c:247
swaywm/sway#5  0x00005641074b4614 in main (argc=2, argv=0x7ffcacd2fc58) at ../sway/sway/main.c:412
        verbose = 0
        debug = 1
        validate = 0
        allow_unsupported_gpu = 0
        long_options = 
            {{name = 0x56410750647b "help", has_arg = 0, flag = 0x0, val = 104}, {name = 0x564107509da9 "config", has_arg = 1, flag = 0x0, val = 99}, {name = 0x564107506480 "validate", has_arg = 0, flag = 0x0, val = 67}, {name = 0x564107506489 "debug", has_arg = 0, flag = 0x0, val = 100}, {name = 0x5641075063df "version", has_arg = 0, flag = 0x0, val = 118}, {name = 0x564107505557 "verbose", has_arg = 0, flag = 0x0, val = 86}, {name = 0x56410750648f "get-socketpath", has_arg = 0, flag = 0x0, val = 112}, {name = 0x56410750649e "unsupported-gpu", has_arg = 0, flag = 0x0, val = 117}, {name = 0x5641075064ae "my-next-gpu-wont-be-nvidia", has_arg = 0, flag = 0x0, val = 117}, {name = 0x0, has_arg = 0, flag = 0x0, val = 0}}
        config_path = 0x0
        usage = 0x564107506820 "Usage: sway [options] [command]\n\n  -h, --help", ' ' <repeats 13 times>, "Show help message and quit.\n  -c, --config <config>  Specify a config file.\n  -C, --validate         Check the validity of the config file, th"...
        c = <optimized out>
```

**Description:**
Sway crashes when an XWayland-client copies the contents of a large file to the clipboard and the copyq clipboard manager is running. The issue does not appear I am copying the same file contents from a Wayland-native app, such as `gedit`. Also, the issue is not present if the `copyq` clipboard manager is not running. When sway crashes, I lose my entire session and get back to the TTY login.

I am not very convinced that this issue is indeed related to the size of the clipboard contents, as so far it reproduced to me with two files only, both being large-ish. It most definitely does not reproduce with the contents of just any large file. So far I couldn't find a pattern.

I am using a wayland-native build of copyq, using the `wayland` branch (https://github.com/hluk/CopyQ/tree/wayland).

The issue happens to me very random, only with some files (few), but it always reproduces with the same file. In the latest case the file is just 330KiB, and is a .csv (not sure if relevant). I cannot share the file itself, as it contains sensitive data.

I followed exactly the following steps to reproduce the issue and to obtain the above logs and stack trace:
1. Started sway;
2. Opened `copyq`;
3. Opened jetbrains-toolbox;
4. Started IntelliJ IDEA from jetbrains-toolbox;
5. Opened a specific file that I know to reproduce the issue;
6. Selected all contents of the file and hit CTRL+C.

Even though the issue is very specific, I feel I need to raise this issue as the impact of losing all my running apps and my work is very inconvenient. I feel like sway should never, ever crash, even if this turns out to be a bug in some other component (copyq, xwayland).

Xyene:
Currently, if:

1. an X11 application requests the contents of the clipboard
2. the contents are too large to be sent in one go, so incr transfer
   starts
3. wlroots registers an fd to write the chunks to in the event loop

If

4. the X11 application requests the contents of the clipboard again
   (before the transfer completed)

Then

5. wlroots registers the same fd with the same void *data into the event
   loop

Finally, when

6. the "first" transfer completes
7. wlroots frees the void *data, removes one instance of the fd from the
   event loop
8. one is still there; if it fires due to closing, wlroots accesses the
    freed void *data and segfaults

This commit makes the request in 4. a no-op, preventing the segfault.

I do not know if this makes any sense from a protocol perspective, or if
this should be disallowed in a different way elsewhere (or if it's
somehow allowed!)

Closes #2548.

### DIFF
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -56,20 +56,25 @@ static int xwm_data_source_write(int fd, uint32_t mask, void *data) {
 
 static void xwm_write_property(struct wlr_xwm_selection_transfer *transfer,
 		xcb_get_property_reply_t *reply) {
-	struct wlr_xwm *xwm = transfer->selection->xwm;
+	if (transfer->source != NULL) {
+		// Incremental transfer requested while one is already in progress; ignore
+		// the request.
+		return;
+	}
 
-	transfer->property_start = 0;
-	transfer->property_reply = reply;
-
-	xwm_data_source_write(transfer->source_fd, WL_EVENT_WRITABLE, transfer);
-
-	if (transfer->property_reply != NULL) {
+	if (reply != NULL) {
+		struct wlr_xwm *xwm = transfer->selection->xwm;
 		struct wl_event_loop *loop =
 			wl_display_get_event_loop(xwm->xwayland->wl_display);
 		transfer->source = wl_event_loop_add_fd(loop,
 			transfer->source_fd, WL_EVENT_WRITABLE, xwm_data_source_write,
 			transfer);
 	}
+
+	transfer->property_start = 0;
+	transfer->property_reply = reply;
+
+	xwm_data_source_write(transfer->source_fd, WL_EVENT_WRITABLE, transfer);
 }
 
 void xwm_get_incr_chunk(struct wlr_xwm_selection_transfer *transfer) {


emersion:
>the X11 application requests the contents of the clipboard again

Shouldn't that queue a new transfer?

Xyene:
> Shouldn't that queue a new transfer?

I'm not sure. At least, I think I'm not using precise enough terminology.

Outgoing transfers are queued up in `xwm_selection_send_data`, whose only caller is `xwm_handle_selection_request`, which is the handler for `XCB_SELECTION_REQUEST`. This outgoing path is not what triggers this bug.

The issue is in the incoming path, in `xwm_write_property`. The incoming path can only have a single ongoing transfer, and performs no sort of queuing like the outgoing path. However, `xwm_write_property` has no guard against starting a second transfer while one is already running. It can double-add the same `transfer->source_fd` into the epoll loop &mdash; this is the eventual cause of the crash.

From [the logs](https://gist.github.com/zsolt-donca/72e67feb89ee67a25b3ae69df2dec159) provided by the OP in https://github.com/swaywm/wlroots/issues/2548, what's happening is `xwm_get_incr_chunk` is called during an incoming transfer, for the same transfer that's already running. This causes [`xwm_write_property` to be called again](https://github.com/swaywm/wlroots/blob/5012121d331c97bc6af856dd58d68b5be336b573/xwayland/selection/incoming.c#L99) for the same transfer, eventually leading to the crash.

`xwm_get_incr_chunk` is just [called](https://github.com/swaywm/wlroots/blob/5012121d331c97bc6af856dd58d68b5be336b573/xwayland/selection/selection.c#L93) in response to `XCB_PROPERTY_NOTIFY`, and there's no logic on that path preventing this issue.

Xyene:
As per the IRC discussion today, I'll close this PR in favor of reworking xwm to use a per-transfer window, which would prevent (this entire class of) bugs in a less hacky way. $DAY_JOB permitting, I'll hopefully have some patches up in the next little while.

PR #2608, opened 2021-01-07
layer shell: advertise version 3

ifreund:
This should have been done in 45c0877 but was overlooked unfortunately.


emersion (approved):
LGTM, thanks!

PR #2609, opened 2021-01-07
xdg shell: add wlr_xdg_surface_for_each_popup_surface() 

ifreund:
~~Omitting these is a bug. Firefox for example only renders stuff to
the subsurfaces of its popups for whatever reason.~~

When rendering, it is necessary to iterate the subsurfaces as well,
so add a function that makes this easy.

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -393,6 +393,14 @@ void wlr_xdg_surface_for_each_surface(struct wlr_xdg_surface *surface,
 		wlr_surface_iterator_func_t iterator, void *user_data);
 
 /**
+ * Call `iterator` on each popup's surface and popup's subsurface in the
+ * xdg-surface tree, with the surfaces's position relative to the root
+ * xdg-surface. The function is called from root to leaves (in rendering order).
+ */
+void wlr_xdg_surface_for_each_popup_surface(struct wlr_xdg_surface *surface,
+		wlr_surface_iterator_func_t iterator, void *user_data);
+
+/**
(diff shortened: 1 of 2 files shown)


emersion:
Ah, but this is about `wlr_xdg_surface_for_each_popup`. This one only iterates over xdg-popup surfaces, nothing else.

Use `wlr_xdg_surface_for_each_surface` to iterate over the whole surface tree.

ifreund:
I've changed this commit to introduce a new function suitable for rendering instead of modifying the existing one. 

The API of the existing function is cleaned up in #2610.

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -392,6 +392,14 @@ void wlr_xdg_surface_get_geometry(struct wlr_xdg_surface *surface,
 void wlr_xdg_surface_for_each_surface(struct wlr_xdg_surface *surface,
 		wlr_surface_iterator_func_t iterator, void *user_data);
 
+/**
+ * Call `iterator` on each popup surface and subsurface in the xdg-surface
```
emersion, line 396: Can we make it clearer that this won't iterate over subsurfaces of the root xdg-surface? Maybe s/subsurface/popup's subsurface/?

ifreund: Done, it sounds a little funny now but it's certainly more clear that way.

emersion (approved):
LGTM, thanks!

Can you send a PR to rename `wlr_layer_surface_v1_for_each_popup` as well?

ifreund:
> Can you send a PR to rename `wlr_layer_surface_v1_for_each_popup` as well?

#2617

PR #2610, opened 2021-01-07
xdg shell: remove wlr_xdg_surface_for_each_popup() 

ifreund:
~~This function is not well suited to rendering as it does not iterate
over subsurfaces. Therefore, it makes little sense to use the standard
surface iterator function. Introduce an new xdg-popup iterator function
and use that instead.~~

This function is inferior to wlr_xdg_surface_for_each_popup_surface()
for rendering as it does not iterate over subsurfaces. Furthermore,
no compositor known to use this to iterate popups for any purpose other
than rendering. Therefore remove the function which may of course be
reintroduced at a later date if a use-case is found.

* * *

Breaking change: `wlr_xdg_surface_for_each_popup` has been removed, use `wlr_xdg_surface_for_each_popup_surface` instead.

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -415,12 +415,4 @@ void wlr_xdg_surface_for_each_popup_surface(struct wlr_xdg_surface *surface,
  */
 uint32_t wlr_xdg_surface_schedule_configure(struct wlr_xdg_surface *surface);
 
-/**
- * Call `iterator` on each popup in the xdg-surface tree, with the popup's
- * position relative to the root xdg-surface. The function is called from root
- * to leaves (in rendering order).
- */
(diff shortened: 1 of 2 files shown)


ifreund:
We should consider simply removing this function instead if #2609 is accepted. I don't know of any use case for iterating popups in a compositor aside from rendering.

emersion:
I agree. We can always re-introduce the function if a compositor finds a use-case.

ifreund:
Sounds good to me, I changed the commit to remove this function instead.

emersion (approved):
LGTM, will wait a few days in case someone wants to object.

emersion:
(Needs a rebase.)

Sway pull request: https://github.com/swaywm/sway/pull/5937

emersion:
Thanks!

PR #2507, opened 2020-12-02
output: take a wlr_buffer in set_cursor

emersion:
We want to remove the rendering bits from the backends. To do so, we need to stop feeding them `wlr_texture` for cursors.

Switch to `wlr_buffer`. The common `wlr_output` code is responsible for allocating and rendering cursor textures.

This will also help with [explicit synchronization](https://github.com/swaywm/wlroots/pull/2070).

TODO:

- [x] Fix disappearing cursor when near to output edge on DRM
- [x] Fix output scaling and transform
- [x] Multi-GPU

~~Depends on https://github.com/swaywm/wlroots/pull/2498~~
~~Depends on https://github.com/swaywm/wlroots/pull/2556~~
~~Depends on https://github.com/swaywm/wlroots/pull/2613~~
~~Depends on https://github.com/swaywm/wlroots/pull/2901~~
~~Depends on https://github.com/swaywm/wlroots/pull/2947~~

* * *

Breaking change for backends: `wlr_output_impl.set_cursor` now takes a `wlr_buffer`. Backends should no longer allocate a swapchain for the cursor. Backends can optionally specify cursor buffer format and size restrictions with `get_cursor_formats` and `get_cursor_size`.

PR #2613, opened 2021-01-07
backend: add get_drm_fd to interface

emersion:
This is part of https://github.com/swaywm/wlroots/pull/2507 and https://github.com/swaywm/wlroots/pull/2505, split into a separate pull request because it might be useful before these are merged.

~~Depends on https://github.com/swaywm/wlroots/pull/2592~~


emersion:
Merged in https://github.com/swaywm/wlroots/pull/2639.

PR #2615, opened 2021-01-07
Simpler screencopy

any1:
This simplifies screencopy by copying pixels to shm in the commit callback rather than the pre-commit callback, so both dmabuf and shm buffers are handled in commit.

Because commit always happens after pre-commit, damage accumulation is also simplified because we no longer need to ascertain whether the damage callback or the frame-copy callback happened first.

Supersedes #2591

### DIFF
--- a/types/wlr_screencopy_v1.c
+++ b/types/wlr_screencopy_v1.c
@@ -21,7 +21,6 @@ struct screencopy_damage {
 	struct pixman_region32 damage;
 	struct wl_listener output_precommit;
 	struct wl_listener output_destroy;
-	uint32_t last_commit_seq;
 };
 
 static const struct zwlr_screencopy_frame_v1_interface frame_impl;
@@ -44,13 +43,6 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 	struct pixman_region32 *region = &damage->damage;
 	struct wlr_output *output = damage->output;
 
-	/* This check is done so damage that has been added and cleared in the
-	 * frame precommit handler is not added again after it has been handled.
-	 */
-	if (damage->last_commit_seq == output->commit_seq) {
-		return;
-	}
-
 	if (output->pending.committed & WLR_OUTPUT_STATE_DAMAGE) {
 		// If the compositor submitted damage, copy it over
 		pixman_region32_union(region, region, &output->pending.damage);
@@ -62,8 +54,6 @@ static void screencopy_damage_accumulate(struct screencopy_damage *damage) {
 		pixman_region32_union_rect(region, region, 0, 0,
 			output->width, output->height);
 	}
-
-	damage->last_commit_seq = output->commit_seq;
 }
 
 static void screencopy_damage_handle_output_precommit(
@@ -98,7 +88,6 @@ static struct screencopy_damage *screencopy_damage_create(
 	}
 
 	damage->output = output;
-	damage->last_commit_seq = output->commit_seq - 1;
 	pixman_region32_init_rect(&damage->damage, 0, 0, output->width,
 		output->height);
 	wl_list_insert(&client->damages, &damage->link);
@@ -154,7 +143,6 @@ static void frame_destroy(struct wlr_screencopy_frame_v1 *frame) {
 		}
 	}
 	wl_list_remove(&frame->link);
-	wl_list_remove(&frame->output_precommit.link);
 	wl_list_remove(&frame->output_commit.link);
 	wl_list_remove(&frame->output_destroy.link);
 	wl_list_remove(&frame->output_enable.link);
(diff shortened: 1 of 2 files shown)


emersion:
I haven't looked in detail yet, but

>renderer: Add buffer argument to read_pixels

I think we can just `bind_buffer`, `read_pixels`, then `bind_buffer(NULL)` to avoid this change?

any1:
> I haven't looked in detail yet, but
> 
> > renderer: Add buffer argument to read_pixels
> 
> I think we can just `bind_buffer`, `read_pixels`, then `bind_buffer(NULL)` to avoid this change?

Hmm, yeah, you're most likely right. I wasn't thinking in terms of the "global picture"; just what would be needed to call `read_pixels` from pretty much anywhere at any time. This change should make it possible to skip scheduling a frame and waiting for the commit event when the user doesn't want to wait for damage or when there is already damage. But maybe that's not really a useful "optimisation".

any1:
I tried using `bind_buffer`, but it failed on `assert(wlr_egl_is_current(renderer->egl))`. If we want to use these functions, we need to (at least) change them so that they set an egl context and they should probably also save/restore the old egl context.

emersion:
>we need to (at least) change them so that they set an egl context

I think this would be a good change, and it's also needed to remove the EGL dependency from the backends. Would you mind opening a PR for this?

> save/restore the old egl context

I think saving/restoring makes sense when we need to temporarily set the EGL context in a single function (like `blit_dmabuf`). However I'm a little worried about saving in one function call and restoring in another one. This could cause issues if the saved EGL objects become invalid, for instance.

any1:
> > we need to (at least) change them so that they set an egl context
> 
> I think this would be a good change, and it's also needed to remove the EGL dependency from the backends. Would you mind opening a PR for this?

Yeah, I'll give it a go

any1:
We are putting this on hold until #2505 is merged. Specifically https://github.com/swaywm/wlroots/pull/2505/commits/47f59dd3042891903f2682b9f3e3cc93afe07ef5.

emersion:
#2505 has been merged, I think now would be a good time to revisit this. https://github.com/swaywm/wlroots/pull/2955 adds `wlr_output.front_buffer`.

emersion:
Ah, still one missing piece is the DRM backend, which still uses the old infrastructure. This will be fixed in https://github.com/swaywm/wlroots/pull/2903.

emersion:
@any1, would you be willing to continue work on this?

any1:
> @any1, would you be willing to continue work on this?

Yeah, sure.

emersion:
We should now be able to listen to output commits exclusively, check for `event->committed & WLR_OUTPUT_STATE_BUFFER`, then grab `output->front_buffer`.

any1:
I believe I have adapted my previous changes to master now. It does seem to work with wayvnc on both the wayland and headless backends at least.

types/wlr_screencopy_v1.c
```diff
@@ -21,7 +21,6 @@ struct screencopy_damage {
 	struct pixman_region32 damage;
 	struct wl_listener output_precommit;
```
emersion, line 22: This can probably be removed?

types/wlr_screencopy_v1.c
```diff
@@ -244,31 +208,24 @@ static void frame_handle_output_precommit(struct wl_listener *listener,
 	wl_shm_buffer_begin_access(shm_buffer);
 	void *data = wl_shm_buffer_get_data(shm_buffer);
 	uint32_t renderer_flags = 0;
+	wlr_renderer_begin_with_buffer(renderer, output->front_buffer);
```
emersion, line 211: We should check for errors here

types/wlr_screencopy_v1.c
```diff
@@ -297,21 +254,45 @@ static bool blit_dmabuf(struct wlr_renderer *renderer,
...
+	// TODO: add support for copying regions with DMA-BUFs
+	if (frame->box.x != 0 || frame->box.y != 0 ||
+			output->width != frame->box.width ||
+			output->height != frame->box.height) {
+		zwlr_screencopy_frame_v1_send_failed(frame->resource);
+		frame_destroy(frame);
+		return false;
+	}
+
+	bool ok = output->front_buffer && blit_dmabuf(renderer, dma_buffer,
```
emersion, line 273: I think we can assume that if an output is enabled, it'll always have a buffer.

types/wlr_screencopy_v1.c
```diff
@@ -297,21 +254,45 @@ static bool blit_dmabuf(struct wlr_renderer *renderer,
...
+	struct wlr_dmabuf_v1_buffer *dma_buffer = frame->dma_buffer;
+	struct wlr_output *output = frame->output;
+	struct wlr_renderer *renderer = wlr_backend_get_renderer(output->backend);
+	assert(renderer);
+
+	// TODO: add support for copying regions with DMA-BUFs
+	if (frame->box.x != 0 || frame->box.y != 0 ||
+			output->width != frame->box.width ||
+			output->height != frame->box.height) {
+		zwlr_screencopy_frame_v1_send_failed(frame->resource);
```
emersion, line 268: I think this is already done by the caller.

types/wlr_screencopy_v1.c
any1: The precommit handler? It's what accumulates damage per output on the client object.

emersion: Ah, sorry, not this one. But `wlr_screencopy_frame_v1.output_precommit` can be removed, now that the `wl_signal_add` call has been removed from `frame_handle_copy`.

emersion:
Ready to merge with that one last cleanup comment fixed

types/wlr_screencopy_v1.c
any1: `wlr_screencopy_frame_v1.output_precommit` is removed now

emersion (approved):
Thanks!

PR #2617, opened 2021-01-08
layer shell: rename popup iterator for consistency

ifreund:
This iterates over the subsurfaces of popups as well, so rename it to
match wlr_xdg_surface_for_each_popup_surface().

* * *

Breaking change: `wlr_layer_surface_v1_for_each_popup` has been renamed to `wlr_layer_surface_v1_for_each_popup_surface`.


emersion (approved):
LGTM, thanks for the follow-up patch!

PR #2626, opened 2021-01-09
Add some X11 error tracing facilities

psychon:
This PR adds some infrastructure to better pinpoint where an X11 error was caused.

As an example, causing an X11 error on purpose like this:
```diff
diff --git a/xwayland/xwm.c b/xwayland/xwm.c
index bb7a4759..0f604f0a 100644
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -168,6 +168,7 @@ static struct wlr_xwayland_surface *xwayland_surface_create(
 	wl_signal_init(&surface->events.ping_timeout);
 	wl_signal_init(&surface->events.set_geometry);
 
+	xcb_change_window_attributes(xwm->xcb_conn, 1234, 0, NULL);
 	xcb_get_geometry_reply_t *geometry_reply =
 		xcb_get_geometry_reply(xwm->xcb_conn, geometry_cookie, NULL);
 	if (geometry_reply != NULL) {
```
produces the following output:
```
00:00:02.140 [wlr] [xwayland/xwm.c:1432] xcb error: op 2:0, code 3, sequence 111, value 1234
00:00:02.140 [wlr] [xwayland/trace.c:64] X11 error happened somewhere after xwayland/xwm.c:128
00:00:02.140 [wlr] [xwayland/trace.c:67] X11 error happened somewhere before xwayland/xwm.c:193
```
Future work on this:
- [ ] Make this a compile time option so that it can be turned off when not needed. Dunno if this is necessary and I didn't want to start the bikeshedding with an own name for the option, so this is up to you.
- [ ] Do something similar for `backend/x11`. Since this is best done by sharing code between `xwayland` and the backend, which is not done so far, I did not want to do this just yet. I'd be happy if someone who knows what they are doing does this as a follow-up PR.

### DIFF
--- a/include/xwayland/trace.h
+++ b/include/xwayland/trace.h
@@ -1,0 +1,23 @@
+#ifndef XWAYLAND_TRACE_H
+#define XWAYLAND_TRACE_H
+
+#include <xcb/xcb.h>
+#include <wayland-util.h>
+
+struct wlr_x11_trace {
+	struct wl_list trace_events; // trace_event::link
+};
+
+void wlr_x11_trace_init(struct wlr_x11_trace *trace);
+void wlr_x11_trace_deinit(struct wlr_x11_trace *trace);
+
+void wlr_x11_trace_received_event(struct wlr_x11_trace *trace, xcb_generic_event_t *event);
+void wlr_x11_trace_log_error_trace(struct wlr_x11_trace *trace, uint32_t sequence);
+
+void wlr_x11_trace_trace(struct wlr_x11_trace *trace, unsigned int sequence,
+	const char *file, unsigned int line);
+
+#define WLR_X11_TRACE(trace, conn) \
+	wlr_x11_trace_trace((trace), xcb_no_operation(conn).sequence, __FILE__, __LINE__)
+
+#endif
--- a/include/xwayland/xwm.h
+++ b/include/xwayland/xwm.h
@@ -12,6 +12,7 @@
 #include <xcb/xcb_errors.h>
 #endif
 #include "xwayland/selection.h"
+#include "xwayland/trace.h"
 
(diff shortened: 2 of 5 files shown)


emersion:
We have [something similar](https://github.com/swaywm/wlroots/blob/master/include/render/gles2.h#L132-L135) for OpenGL, except the OpenGL driver takes care of keeping track of the trace points. In this regard, this implementation is more complicated.

I wonder if it's worth it. In my experience, the X11 error messages contain the request name, and that's in general enough to understand where the error is coming from. Did you encounter a case where this PR was helpful?

> Make this a compile time option so that it can be turned off when not needed.

Could maybe just turn it off at runtime if the debug level is lower than debug.

psychon:
My motivation is "just because". And to see how complicated it is to get this working. If you say that this does not solve a problem that you have, then that is totally fine with me.

> Did you encounter a case where this PR was helpful?

Nope. My hope was rather the other way: Someone shows up and says "I wish we already had this a year ago". Otherwise, this is not worth the trouble.

(Random side note: It is possible for the `uint32_t` sequence numbers to wrap around. Internally, libxcb keeps `uint64_t` numbers which one can assume never wrap. I don't think the PR handles this case correctly yet (but it's also hard to test since mapping sequence numbers wrap takes some time). My first idea on how to handle this would be to just throw away the whole queue when the sequence numbers wrap around and hope no X11 errors show up that would have needed this information.)

emersion:
Thanks. I guess we can keep this PR open in case someone needs it, and revive this discussion when this happens.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2626

Issue #2188, opened 2020-05-12
Add support for hotplug_mode_update


emersion:
This is set by VM software when user-space is supposed to modeset with the preferred mode on hotplug.

Related: the "suggested X" and "suggested Y" props.

emersion:
We may also want to decide to NACK the whole KMS VM support. The whole thing is pretty hacky.

emersion:
With https://github.com/swaywm/wlroots/pull/2645, regardless of whatever wlroots chooses to support, compositors should be able to handle these properties if they want to.

hedgepigdaniel:
@emersion I'm curious how you see this stuff working out (especially in terms of what Sway can support).

My goal: to make Sway (and perhaps other wlroots compositors) work in the way you would expect when running in a virtual machine. IMO that means that it should:
 - use the preferred mode for each output, including if it changes (as it does if the host window is resized)
 - layout multiple outputs in the same way that they are on the host system, in the case that this information is provided, and including if it changes for a particular output (as it sometimes does if other outputs are added/removed from the VM)

Currently, i.e. without these abilities, using sway on a VM is a poor experience, because:
 - if a host window is resized, Sway does not mode set to match the new preferred mode, so the output dimensions become incorrect (aspect ratio, resolution, and scale)
 - if the host displays the VM in full screen on multiple outputs, their relative position is incorrect by default.
 - a configuration that solves either of the above problems is specific to a particular window/output arrangement on the host. A correct configuration becomes wrong if a window is resized, full screen mode is toggled, or a screen is added/removed.

As best as I understand the suggested X/Y props, they are useful if and only if the user wants the outputs to be configured as they are on the host system. In this case, the host system is responsible for choosing output modes, rotation, position, and relative scale. The chosen configuration is made visible in the VM with the preferred mode and suggested X/Y props (in pixels). 

If the user wants to choose their own rotation, relative scale, or position for any of the outputs, then they are choosing not to use the host configuration, and therefore the suggested X/Y positions are irrelevant.

The positions are in pixels, so there is an assumption that all pixels are the same size. This assumption is consistently made based on all the testing I've done (VMware Fusion on OSX and Virtualbox on a Sway host):
 - the output physical size props are not set in the VM (they are always  0x0)
 - the pixels presented to the VM are scaled according to the host configuration. For example on VMware Fusion with a high and low resolution display, each VM pixel corresponds to 1 pixel on the high resolution display and 2x2 pixels on the low resolution display.

[...]

Issue #2300, opened 2020-06-22
output: remove non-atomic events


emersion:
Some events in `wlr_output` are remnants of the pre-atomic logic:

- `enable`
- `mode`
- `scale`
- `transform`

All of these can be replaced with a `commit` listener.

Related: we can de-duplicate these event listeners, e.g. https://github.com/swaywm/wlroots/blob/d5530b26d7dfc142ee1f37020436d11696864fd3/types/wlr_output_damage.c#L16

emersion:
> All of these can be replaced with a commit listener.

Not quite. Here are two exceptions:

- The backend turns an output on on its own. This can happen when the compositor requested to enable an output, the backend failed the commit because of a missing CRTC, and then a CRTC becomes available again. See https://github.com/swaywm/wlroots/issues/2094
- With the Wayland and X11 backends, the user can resize the window. This makes the backend perform a mode change on its own.

emersion:
https://github.com/swaywm/wlroots/pull/2631 has removed `scale` and `transform`.

For `enable` and `mode`, as noted above, things get more complicated. I see two potential solutions:

1. Let backends perform output commits under-the-hood. Compositors would receive `commit` events for commits they haven't performed.
2. Add a new `wlr_output` event to request an output configuration change. The Wayland/X11 backends would request a custom mode, the DRM backend would request to enable the output and set a fixed mode.

We discussed this on IRC with @ammen99. It sounds like while (1) doesn't require any action from the compositor, (1) would force the compositor to accept whatever the backend chooses. This could be an issue if the user has explicitly configured a mode for an output, for instance. Also, we'd like to move away from automatically enabling outputs in the DRM backend: https://github.com/swaywm/wlroots/issues/2094.

(2) would let the compositor have the final word, and would probably just be a few extra lines of code.

PR #2628, opened 2021-01-10
backend/drm: handle changes in the set of fixed modes for an output

hedgepigdaniel:
Related to https://github.com/swaywm/wlroots/issues/2188

`vmwgfx`, `qxl`,  and possibly other drivers for virtualised displays emit a udev change event for outputs when the corresponding window on the host system is resized. In this case, the set of fixed modes for the virtualised output changes such that the preferred mode matches the new dimensions of the window on the host system.

This change handles this case by updating the modes in `(struct wlr_output).modes`, and adding a new signal `available_modes` which compositors may use to be notified when such a change occurs. The mode is not automatically changed, instead the compositor may change the mode as it sees fit.

Notes:
 - I didn't use `hotplug_mode_update` because I didn't see any need to refer to it. If the modes change, then it is handled, and if they don't, nothing happens.
 - A mode is freed when it has been removed and is no longer the current or pending mode. I assume that a compositor will not try to access a mode after this time. If the compositor does not handle the `available_modes` signal, then it is possible that the old/removed mode will remain active. In this case, it should work the same as before.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1221,6 +1221,94 @@ static uint32_t get_possible_crtcs(int fd, drmModeRes *res,
 
 static void disconnect_drm_connector(struct wlr_drm_connector *conn);
 
+static bool update_modes(drmModeConnector *drm_conn,
+		struct wlr_drm_connector *wlr_conn) {
+	bool changes_made = false;
+
+	struct wlr_output_mode *previous_mode, *tmp_mode;
+	wl_list_for_each_safe(previous_mode, tmp_mode, &wlr_conn->output.modes, link) {
+
+		struct wlr_drm_mode *previous_drm_mode = (struct wlr_drm_mode *) previous_mode;
+		bool matched = false;
+
+		for (int j = 0; j < drm_conn->count_modes; ++j) {
+			if (memcmp(&previous_drm_mode->drm_mode, &drm_conn->modes[j],
+					sizeof(previous_drm_mode->drm_mode)) == 0) {
+				matched = true;
+				continue;
+			}
+		}
+
+		if (matched) {
+			continue;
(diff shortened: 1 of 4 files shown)


emersion:
Marking this as a breaking change, since modes can now be removed from the list. Previously, compositors could stash a `wlr_output_mode` somewhere and apply it later.

This PR is causing quite some issues related to memory management. This PR tries to destroy output modes, except if they are still in use by `wlr_output.mode` or `wlr_output.pending.mode`. I'm not sure if there's a better solution.

hedgepigdaniel:
Yes I struggled with the memory management. The only alternative I can think of: don't free any modes when they are removed, and instead introduce a list of old/invalid modes to be freed when the output is destroyed.

emersion:
I think there's some overlap between this and https://github.com/swaywm/wlroots/issues/2300#issuecomment-761819673. If we go with the solution (2) mentioned there, we could just ask nicely the compositor to set the new mode.

emersion:
Something I'd like to fix is that these properties aren't standard right now. We should really move them in the kernel DRM core and [document them](https://dri.freedesktop.org/docs/drm/gpu/drm-kms.html#standard-connector-properties) so that all virtual drivers can use them.

emersion:
Thinking about this a little bit more: maybe it doesn't make sense to populate `output->modes` at all when a connector has the `hotplug_mode_update` property. The mode list is mostly populated with standard modes so that user-space which expects physical outputs isn't confused, right?

Maybe DRM outputs backed by a virtual driver can behave a bit more like the Wayland and X11 backends.

hedgepigdaniel:
Closed in favour of https://github.com/swaywm/wlroots/pull/2712

PR #2629, opened 2021-01-10
backend/drm: add support for suggested output position prop

hedgepigdaniel:
Depends on and includes commits from https://github.com/swaywm/wlroots/pull/2628 (could also be independently rebased onto master, but wouldn't be as useful on its own)

Relates to https://github.com/swaywm/wlroots/issues/2188

Virtualised outputs use the "suggested X" and "suggested Y" DRM props to indicate their relative position on the host WM.

This change adds members `suggested_x` and `suggested_y` to `struct wlr_output` which contain the value of these props. Also, a `suggested_position` signal is added so that compositors can be notified if there is a changed in the suggested position.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1221,6 +1221,94 @@ static uint32_t get_possible_crtcs(int fd, drmModeRes *res,
 
 static void disconnect_drm_connector(struct wlr_drm_connector *conn);
 
+static bool update_modes(drmModeConnector *drm_conn,
+		struct wlr_drm_connector *wlr_conn) {
+	bool changes_made = false;
+
+	struct wlr_output_mode *previous_mode, *tmp_mode;
+	wl_list_for_each_safe(previous_mode, tmp_mode, &wlr_conn->output.modes, link) {
+
+		struct wlr_drm_mode *previous_drm_mode = (struct wlr_drm_mode *) previous_mode;
+		bool matched = false;
+
+		for (int j = 0; j < drm_conn->count_modes; ++j) {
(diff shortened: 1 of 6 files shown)


emersion:
The KMS suggested position property is fragile, it doesn't handle well things like output scaling and rotation. Sway auto-detects the scale of outputs. So I'm not a big fan of this kernel API.

It's not clear what this is used for. This seems to be a workaround to get pointer events properly routed to the right position.

hedgepigdaniel:
Nothing to do with pointer events - the purpose is so that in a virtualised system with multiple virtualised outputs, compositors do not need to be separately configured with the same output layout that the host system is configured with, and instead the correct layout can be read from the suggested X/Y properties.

That's what https://github.com/swaywm/sway/pull/5944 does - positions the outputs the same way that they are positioned in the host system. Scale is handled there, including the default values (the suggested X/Y is multiplied by the effective scale to give the scaled position coordinates).

Rotation is not handled in https://github.com/swaywm/sway/pull/5944 - but I think its very unlikely that a user would want their virtualised output rotated differently to the physical output on the host system. If that is desired then it can be overridden by configuring a position, which overrides the default suggested properties.

hedgepigdaniel:
Now that I think about it, it is relevant to pointer events, because when the output positions in sway don't match the positions in the host WM the pointer behaves very strangely. In my testing it works normally with this change and https://github.com/swaywm/sway/pull/5944.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2629

PR #2635, opened 2021-01-11
Make libinput and DRM backends optional

emersion:
Some compositors may not need the libinput or DRM backend, because they provide
their own backends. Allow to build wlroots without these.


hizukiayaka:
My proposal is at #2809

emersion:
Closing in favor of https://github.com/swaywm/wlroots/pull/3058

PR #2639, opened 2021-01-14
Drop EGL from wlr_renderer_autocreate

### LINKED ISSUE #2563: Drop EGL things from wlr_renderer_autocreate

bl4ckb0ne:
Fixes #2563, requires https://github.com/swaywm/wlroots/pull/2613

* * *
Breaking change: `wlr_renderer_autocreate` no longer takes the arguments `EGLenum platform, void *remote_display`. Instead it takes a `struct wlr_backend *`.

### DIFF
--- a/include/wlr/render/egl.h
+++ b/include/wlr/render/egl.h
@@ -41,6 +41,7 @@ struct wlr_egl {
 	EGLConfig config; // may be EGL_NO_CONFIG
 	EGLContext context;
 	EGLDeviceEXT device; // may be EGL_NO_DEVICE_EXT
+	struct gbm_device *gbm_device;
 
 	struct {
 		// Display extensions
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -1,6 +1,8 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <gbm.h>
+#include <wlr/render/egl.h>
 #include <wlr/render/gles2.h>
 #include <wlr/render/interface.h>
 #include <wlr/render/wlr_renderer.h>
@@ -8,6 +10,7 @@
 #include <wlr/util/log.h>
 #include "util/signal.h"
 #include "render/wlr_renderer.h"
+#include "backend/backend.h"
 
 void wlr_renderer_init(struct wlr_renderer *renderer,
 		const struct wlr_renderer_impl *impl) {
@@ -247,13 +250,28 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 	return true;
 }
 
-struct wlr_renderer *wlr_renderer_autocreate(EGLenum platform,
-		void *remote_display) {
-	struct wlr_egl *egl = wlr_egl_create(platform, remote_display, NULL);
-	if (egl == NULL) {
-		wlr_log(WLR_ERROR, "Could not initialize EGL");
+struct wlr_renderer *wlr_renderer_autocreate(struct wlr_backend *backend) {
+	int fd = backend_get_drm_fd(backend);
+	if (fd < 0) {
+		wlr_log(WLR_ERROR, "Failed to get DRM FD from backend");
 		return NULL;
 	}
+
+	struct gbm_device *gbm_device = gbm_create_device(fd);
+	if (!gbm_device) {
+		wlr_log(WLR_ERROR, "Failed to create GBM device");
+		return NULL;
+	}
+
+	struct wlr_egl *egl = wlr_egl_create(EGL_PLATFORM_GBM_KHR, gbm_device,
+			NULL);
+	if (egl == NULL) {
+		wlr_log(WLR_ERROR, "Could not initialize EGL");
+		gbm_device_destroy(gbm_device);
+		return NULL;
+	}
+
+	egl->gbm_device = gbm_device;
 
 	struct wlr_renderer *renderer = wlr_gles2_renderer_create(egl);
 	if (!renderer) {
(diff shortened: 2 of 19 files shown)


render/wlr_renderer.c
```diff
@@ -247,9 +249,10 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 	return true;
 }
 
-struct wlr_renderer *wlr_renderer_autocreate(EGLenum platform,
-		void *remote_display) {
-	struct wlr_egl *egl = wlr_egl_create(platform, remote_display, NULL);
+struct wlr_renderer *wlr_renderer_autocreate(struct wlr_backend *backend) {
+	int fd = backend_get_drm_fd(backend);
```
emersion, line 253: We need to fail if `fd < 0`.

render/wlr_renderer.c
```diff
@@ -247,9 +249,10 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
...
+	struct wlr_egl *egl = wlr_egl_create(EGL_PLATFORM_GBM_KHR,
+			gbm_create_device(fd), NULL);
```
emersion, line 255: We need to handle `gbm_create_device` failures by checking whether it returns NULL here.

emersion, line 255: When the `wlr_egl` is destroyed, we need to destroy the GBM device.

A simple fix would be to add a `struct gbm_device *` field to `wlr_egl`.

emersion (changes requested):
The meat of the patch looks good.

Can you squash the last two commits together? Each commit should compile correctly, otherwise that makes bisecting difficult.

emersion:
Also, I think we can drop the `<wlr/render/egl.h>` include from `include/wlr/render/wlr_renderer.h`.

render/wlr_renderer.c
bl4ckb0ne: Could it be 0 or is it already checked when initializing it in the backends?

emersion: 0 is a valid FD. For instance, if a compositor closes stdout and then opens a DRM FD, that FD would be assigned the number 0. Yes, it's misleading.

emersion: The check should be `< 0`, not `== 0`.

include/wlr/render/egl.h
```diff
@@ -24,6 +24,7 @@
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
 #include <pixman.h>
+#include <gbm.h>
```
emersion, line 27: We don't actually need to include this header, I think. It should work fine without.

render/wlr_renderer.c
```diff
@@ -247,14 +249,29 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 	return true;
 }
 
-struct wlr_renderer *wlr_renderer_autocreate(EGLenum platform,
-		void *remote_display) {
-	struct wlr_egl *egl = wlr_egl_create(platform, remote_display, NULL);
+struct wlr_renderer *wlr_renderer_autocreate(struct wlr_backend *backend) {
+	int fd = backend_get_drm_fd(backend);
+	if (fd == 0) {
+		wlr_log(WLR_ERROR, "Invalid DRM fd");
```
emersion, line 255: A better error message might be "Failed to get DRM FD from backend".

render/wlr_renderer.c
bl4ckb0ne: Oh derp

include/wlr/render/egl.h
bl4ckb0ne: I moved it to the `egl.c` file where it's needed for the `gbm_destroy_device` call

render/wlr_renderer.c
bl4ckb0ne: Makes more sense yeah

render/wlr_renderer.c
bl4ckb0ne: Done

bl4ckb0ne: Done

bl4ckb0ne:
I dropped the `egl.h` header from the `wlr_renderer` one, but it's required in the backends when getting the `wlr_egl` from the `wlr_gles_renderer`.

include/wlr/render/egl.h
emersion: Cool. It seems this line was still added in this commit (and removed in the next one), fixed it locally.

emersion (approved):
LGTM, thanks!

Issue #2618, opened 2021-01-08
render: better integrate bind_buffer with begin/end


emersion:
Right now, one needs to do the following to render:

```c
wlr_egl_make_current(egl, EGL_NO_SURFACE); // done in the backend
wlr_renderer_bind_buffer(renderer, buffer); // done in the backend
wlr_renderer_begin(renderer, width, height); // done in the compositor

// Perform the rendering operations

wlr_renderer_end(renderer); // done in the compositor
wlr_renderer_bind_buffer(renderer, NULL); // done in the backend
wlr_egl_unset_current(egl); // done in the backend
```

Once https://github.com/swaywm/wlroots/pull/2615#issuecomment-756687006 is done, this could be simplified to:

```c
wlr_renderer_bind_buffer(renderer, buffer);
wlr_renderer_begin(renderer, width, height);

// Perform the rendering operations

wlr_renderer_end(renderer);
wlr_renderer_bind_buffer(renderer, NULL);
```

I'm still not very happy with the API, something like this would be better:

```c
wlr_renderer_begin(renderer, buffer);

// Perform the rendering operations

wlr_renderer_end(renderer);
```

However this would be a breaking change, because compositors call `wlr_renderer_begin`. Also need to figure out if/how compositors should acquire the buffer.

PR #2640, opened 2021-01-14
render/gles2: make EGL context current in bind_buffer

emersion:
Instead of requiring callers to manually make the EGL context current
before binding a buffer and unsetting it after unbinding a buffer, do
it inside wlr_renderer_bind_buffer.

This hides renderer-specific implementation details inside the
wlr_renderer interface. Non-GLES2 renderers may not use EGL.
This removes almost all EGL dependencies from the backends, the only
remaining one is wlr_renderer_autocreate args.

References: https://github.com/swaywm/wlroots/issues/2618
References: https://github.com/swaywm/wlroots/pull/2615#issuecomment-756687006

cc @bl4ckb0ne @any1

* * *

This is _not_ a breaking change because `wlr_renderer_bind_buffer` is a private wlroots API.

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -8,8 +8,6 @@
 #include <string.h>
 #include <unistd.h>
 #include <wayland-util.h>
-#include <wlr/render/egl.h>
-#include <wlr/render/gles2.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_matrix.h>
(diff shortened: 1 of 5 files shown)


any1:
With this I should be able to do what you suggested in https://github.com/swaywm/wlroots/pull/2615#issuecomment-756398842, right?

emersion:
Yes, I believe this should work.

PR #2641, opened 2021-01-14
render/gles2: restore EGL context after texture operations

emersion:
It can be surprising and unexpected that texture operations (such as
texture upload and import) change the current EGL context, especially
when it's done under-the-hood by wlroots in response to wl_surface
requests.

To prevent surprises, save and restore the previous EGL context.

cc @any1


any1:
Makes sense

Issue #2601, opened 2021-01-05
Crash with nouveau in gbm_bo_destroy


J0nnyMak0:
With latest sources, `swaymsg exit` consistently crashes:
```
#0  0x00007f739fdc93fc in  () at /usr/lib/libgbm.so.1
#1  0x00007f73a06b9475 in buffer_destroy (wlr_buffer=0x564f90f46a70) at ../subprojects/wlroots/render/gbm_allocator.c:58
        buffer = 0x564f90f46a70
#2  0x00007f73a06f3dfb in buffer_consider_destroy (buffer=0x564f90f46a70) at ../subprojects/wlroots/types/wlr_buffer.c:26
#3  0x00007f73a06f3e55 in wlr_buffer_drop (buffer=0x564f90f46a70) at ../subprojects/wlroots/types/wlr_buffer.c:36
        __PRETTY_FUNCTION__ = "wlr_buffer_drop"
#4  0x00007f73a06bd350 in slot_reset (slot=0x564f90dfeba8) at ../subprojects/wlroots/render/swapchain.c:43
#5  0x00007f73a06bd3af in wlr_swapchain_destroy (swapchain=0x564f90dfeb40) at ../subprojects/wlroots/render/swapchain.c:52
        i = 2 
#6  0x00007f73a06c6ea3 in finish_drm_surface (surf=0x564f90009db8) at ../subprojects/wlroots/backend/drm/renderer.c:106
#7  0x00007f73a06c7352 in drm_plane_finish_surface (plane=0x564f90009db0) at ../subprojects/wlroots/backend/drm/renderer.c:194
#8  0x00007f73a06c3f28 in dealloc_crtc (conn=0x564f90edf5c0) at ../subprojects/wlroots/backend/drm/drm.c:1072
        drm = 0x564f90006ce0
#9  0x00007f73a06c3d5e in drm_connector_destroy_output (output=0x564f90edf5c0) at ../subprojects/wlroots/backend/drm/drm.c:1015
        conn = 0x564f90edf5c0
        mode = 0x564f8f7df2e0
        mode_tmp = 0x0 
#10 0x00007f73a070a456 in wlr_output_destroy (output=0x564f90edf5c0) at ../subprojects/wlroots/types/wlr_output.c:393
        cursor = 0x564f90edf800
        tmp_cursor = 0x564f90edf800
[...]

PR #2642, opened 2021-01-15
render/gbm_allocator: fix gbm_device use-after-free

emersion:
We need to destroy any gbm_bo we've created before gbm_device_destroy.

Closes: https://github.com/swaywm/wlroots/issues/2601


J0nnyMak0:
This has fixed the crash for me. Thanks!

PR #2648, opened 2021-01-16
WIP: Vulkan allocator

emersion:
This is a first step towards a Vulkan renderer.

This makes use of [VK_EXT_image_drm_format_modifier](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_image_drm_format_modifier.html) and [VK_EXT_physical_device_drm](https://github.com/KhronosGroup/Vulkan-Docs/pull/1356) to allocate buffers and export DMA-BUFs.


322997am:
Question, does this mean that one may hypothetically be able to avoid GBM through this, or is it still necessary?

emersion:
The other place where GBM is used is in the DRM backend. This is necessary because of [GEM handles ref'counting issues](https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/110).

In general, for setups that lack a GPU, it's desirable to get rid of GBM. There are ways around the ref'counting issues as outlined in the libdrm MR.

Note: proprietary drivers are not and won't be officially supported.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2648

PR #2650, opened 2021-01-16
handle some common client behaviour

damianatorrpm:
Some programs are resetting _NET_WM_STATE properties to their current value,
to indicate that values may have changed.

Example program: "StepMania"
1) configure in windowed mode
2) change resolution (maybe a few times is required)

Current behaviour: 
Results in unhandled _NET_WM_STATE messages of type _NET_WM_STATE_FULLSCREEN / max_h / max_v
No other unhandled messages are logged.
Nothing happens.

Behaviour with PR:
Emits a new signal.
Compositors can decide what to do.

Other:
Works fine on gnome.

### DIFF
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -199,7 +199,8 @@ struct wlr_xwayland_surface {
 		struct wl_signal set_override_redirect;
 		struct wl_signal set_geometry;
 		struct wl_signal ping_timeout;
-	} events;
+		struct wl_signal request_reload_configure;
+        } events;
 
 	struct wl_listener surface_destroy;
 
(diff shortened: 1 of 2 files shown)


include/wlr/xwayland.h
```diff
@@ -199,7 +199,8 @@ struct wlr_xwayland_surface {
 		struct wl_signal set_override_redirect;
 		struct wl_signal set_geometry;
 		struct wl_signal ping_timeout;
-	} events;
+		struct wl_signal request_reload_configure;
```
ammen99, line 202: I think it would be a lot clearer if we emit a single `set_wm_state` event. This way, compositors can recalculate allowed actions like mutter does.

emersion:
What does the spec say about _NET_WM_STATE client messages? Does it state that XWMs _must_ replace the _NET_WM_STATE prop in response? If so, wlroots should handle it.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2650

PR #2653, opened 2021-01-16
backend/drm: use local DRM FD for wlr_rend

emersion:
The new wlr_renderer_autocreate API is great for compositors, however
it causes some issues with DRM multi-GPU support.
    
A DRM child backend wants the compositor to use the parent GPU, so it
exposes the parent's DRM FD in get_drm_fd. However, in order to be able
to perform multi-GPU buffer copies, the child DRM backend still needs to
create a local renderer.
    
Use the new private wlr_renderer_autocreate_with_drm_fd function to
avoid creating a renderer for the parent GPU.
    
Fixes: e128e6c08dc0 ("render: drop egl parameters from wlr_renderer_autocreate")

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -28,7 +28,7 @@ bool init_drm_renderer(struct wlr_drm_backend *drm,
 		return false;
 	}
 
-	renderer->wlr_rend = wlr_renderer_autocreate(&drm->backend);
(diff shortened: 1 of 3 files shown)


J0nnyMak0:
This brings back external monitors with `WLR_DRM_NO_MODIFIERS=1` Thanks!

PR #2654, opened 2021-01-17
render: add wlr_egl_create_from_drm_fd function

### LINKED ISSUE #2649: render/egl: use EGLDevice instead of the GBM platform
We use the GBM platform to initialize EGL on a particular GPU. Instead, we could use the EGL device platform.

See https://github.com/swaywm/wlroots/issues/2563#issuecomment-757869784 for details.

cc @bl4ckb0ne

bl4ckb0ne:
Fixes #2649

### DIFF
--- a/include/wlr/render/egl.h
+++ b/include/wlr/render/egl.h
@@ -53,6 +53,9 @@ struct wlr_egl {
 
 		// Device extensions
 		bool device_drm_ext;
+
+		// Client extensions
+		bool device_query_ext;
 	} exts;
 
 	struct {
@@ -70,6 +73,7 @@ struct wlr_egl {
 		PFNEGLDEBUGMESSAGECONTROLKHRPROC eglDebugMessageControlKHR;
 		PFNEGLQUERYDISPLAYATTRIBEXTPROC eglQueryDisplayAttribEXT;
 		PFNEGLQUERYDEVICESTRINGEXTPROC eglQueryDeviceStringEXT;
+		PFNEGLQUERYDEVICESEXTPROC eglQueryDevicesEXT;
 	} procs;
 
 	struct wl_display *wl_display;
@@ -86,6 +90,15 @@ struct wlr_egl {
  */
 struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 	const EGLint *config_attribs);
+
+/**
+ * Creates and EGL context from a given drm fd. This function uses the
+ * following extensions:
+ * - EXT_device_enumeration to get the list of available EGLDeviceEXT
+ * - EXT_device_drm to get the name of the EGLDeviceEXT
+ * - EXT_platform_device to create an EGLDisplay from an EGLDeviceEXT
+ */
+struct wlr_egl *wlr_egl_create_from_drm_fd(int drm_fd);
 
 /**
  * Frees all related EGL resources, makes the context not-current and
--- a/include/wlr/render/egl.h
+++ b/include/wlr/render/egl.h
@@ -41,7 +41,6 @@ struct wlr_egl {
 	EGLConfig config; // may be EGL_NO_CONFIG
 	EGLContext context;
 	EGLDeviceEXT device; // may be EGL_NO_DEVICE_EXT
-	struct gbm_device *gbm_device;
 
 	struct {
 		// Display extensions
--- a/render/egl.c
+++ b/render/egl.c
@@ -150,8 +150,7 @@ out:
 	free(formats);
 }
 
-struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
-		const EGLint *config_attribs) {
+static struct wlr_egl *egl_create(void) {
 	struct wlr_egl *egl = calloc(1, sizeof(struct wlr_egl));
 	if (egl == NULL) {
 		wlr_log_errno(WLR_ERROR, "Allocation failed");
@@ -167,6 +166,7 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 		}
 		return NULL;
 	}
+	wlr_log(WLR_INFO, "Supported EGL client extensions: %s", client_exts_str);
 
 	if (!check_egl_ext(client_exts_str, "EGL_EXT_platform_base")) {
 		wlr_log(WLR_ERROR, "EGL_EXT_platform_base not supported");
@@ -176,6 +176,18 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 		"eglGetPlatformDisplayEXT");
 	load_egl_proc(&egl->procs.eglCreatePlatformWindowSurfaceEXT,
 		"eglCreatePlatformWindowSurfaceEXT");
+
+	if (check_egl_ext(client_exts_str, "EGL_EXT_device_enumeration")) {
+		load_egl_proc(&egl->procs.eglQueryDevicesEXT, "eglQueryDevicesEXT");
+	}
+
+	if (check_egl_ext(client_exts_str, "EGL_EXT_device_query")) {
+		egl->exts.device_query_ext = true;
+		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
+			"eglQueryDeviceStringEXT");
+		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
+			"eglQueryDisplayAttribEXT");
+	}
 
 	if (check_egl_ext(client_exts_str, "EGL_KHR_debug")) {
 		load_egl_proc(&egl->procs.eglDebugMessageControlKHR,
@@ -193,26 +205,31 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 
 	if (eglBindAPI(EGL_OPENGL_ES_API) == EGL_FALSE) {
 		wlr_log(WLR_ERROR, "Failed to bind to the OpenGL ES API");
-		goto error;
+		return NULL;
 	}
 
+	return egl;
+}
+
+static bool egl_init(struct wlr_egl *egl, EGLenum platform,
+		void *remote_display, const EGLint *config_attribs) {
 	egl->display = egl->procs.eglGetPlatformDisplayEXT(platform,
 		remote_display, NULL);
 	if (egl->display == EGL_NO_DISPLAY) {
 		wlr_log(WLR_ERROR, "Failed to create EGL display");
-		goto error;
+		return false;
 	}
 
 	EGLint major, minor;
 	if (eglInitialize(egl->display, &major, &minor) == EGL_FALSE) {
 		wlr_log(WLR_ERROR, "Failed to initialize EGL");
-		goto error;
+		return false;
 	}
 
 	const char *display_exts_str = eglQueryString(egl->display, EGL_EXTENSIONS);
 	if (display_exts_str == NULL) {
 		wlr_log(WLR_ERROR, "Failed to query EGL display extensions");
-		return NULL;
+		return false;
 	}
 
 	if (check_egl_ext(display_exts_str, "EGL_KHR_image_base")) {
@@ -251,17 +268,12 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 	}
 
 	const char *device_exts_str = NULL;
-	if (check_egl_ext(client_exts_str, "EGL_EXT_device_query")) {
-		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
-			"eglQueryDisplayAttribEXT");
-		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
-			"eglQueryDeviceStringEXT");
-
+	if (egl->exts.device_query_ext) {
 		EGLAttrib device_attrib;
 		if (!egl->procs.eglQueryDisplayAttribEXT(egl->display,
 				EGL_DEVICE_EXT, &device_attrib)) {
 			wlr_log(WLR_ERROR, "eglQueryDisplayAttribEXT(EGL_DEVICE_EXT) failed");
-			goto error;
+			return false;
 		}
 		egl->device = (EGLDeviceEXT)device_attrib;
 
@@ -269,7 +281,7 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			egl->procs.eglQueryDeviceStringEXT(egl->device, EGL_EXTENSIONS);
 		if (device_exts_str == NULL) {
 			wlr_log(WLR_ERROR, "eglQueryDeviceStringEXT(EGL_EXTENSIONS) failed");
-			goto error;
+			return false;
 		}
 
 		egl->exts.device_drm_ext =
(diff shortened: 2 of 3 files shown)


render/egl.c
```diff
@@ -358,6 +358,58 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+			wlr_log(WLR_INFO, "nodes[%d] %s", j, device->nodes[j]);
+
+	EGLDeviceEXT *egl_device = NULL;
+	for(int i = 0; i < nb_devices; i++)
+	{
+		const char *dev_name = eglQueryDeviceStringEXT(devices[i],
+				EGL_DRM_DEVICE_FILE_EXT);
+
+		// TODO is nodes[0] always the card name?
+		if (strcmp(dev_name, device->nodes[0]) == 0) {
```
neon64, line 398: I think there's a function called `device_has_name` (also in egl.c) which performs this check for you

(incidentally I also had a go at making this change before I saw your PR - maybe there's something of use here: https://github.com/swaywm/wlroots/compare/master...neon64:egl-from-drm-fd

render/egl.c
```diff
@@ -358,6 +358,58 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 	return NULL;
 }
 
+struct wlr_egl *wlr_egl_create_from_drm_fd(int drm_fd) {
+	PFNEGLQUERYDEVICESTRINGEXTPROC eglQueryDeviceStringEXT;
+	PFNEGLQUERYDEVICESEXTPROC eglQueryDevicesEXT;
+
+	load_egl_proc(&eglQueryDevicesEXT, "eglQueryDevicesEXT");
+	load_egl_proc(&eglQueryDeviceStringEXT,"eglQueryDeviceStringEXT");
```
emersion, line 366: We need to check extension strings before doing that. EGL might return garbage function pointers otherwise.

render/egl.c
```diff
@@ -358,6 +358,58 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	drmGetDevice(drm_fd, &device);
+	wlr_log(WLR_INFO, "drmdeviceptr size %d", device->available_nodes);
+
+	for (int j = 0; j < DRM_NODE_MAX; j++)
+		if (device->available_nodes & 1 << j)
+			wlr_log(WLR_INFO, "nodes[%d] %s", j, device->nodes[j]);
+
+	EGLDeviceEXT *egl_device = NULL;
+	for(int i = 0; i < nb_devices; i++)
+	{
```
emersion, line 393: Style: `{` should be on the previous line.

render/egl.c
```diff
@@ -358,6 +358,58 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+		return NULL;
+	}
+
+	eglQueryDevicesEXT(MAX_DEVICES, devices, &nb_devices);
+
+	drmDevicePtr device = NULL;
+	drmGetDevice(drm_fd, &device);
+	wlr_log(WLR_INFO, "drmdeviceptr size %d", device->available_nodes);
+
+	for (int j = 0; j < DRM_NODE_MAX; j++)
```
emersion, line 387: Style: braces are mandatory

render/egl.c
```diff
@@ -358,6 +358,58 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	PFNEGLQUERYDEVICESTRINGEXTPROC eglQueryDeviceStringEXT;
+	PFNEGLQUERYDEVICESEXTPROC eglQueryDevicesEXT;
+
+	load_egl_proc(&eglQueryDevicesEXT, "eglQueryDevicesEXT");
+	load_egl_proc(&eglQueryDeviceStringEXT,"eglQueryDeviceStringEXT");
+
+	static const EGLint MAX_DEVICES = 64;
+	EGLint nb_devices = 0;
+
+	eglQueryDevicesEXT(MAX_DEVICES, NULL, &nb_devices);
```
emersion, line 371: I'd prefer to enable EGL debugging prior to any EGL calls (see `eglDebugMessageControlKHR` in `wlr_egl_create`).

emersion, line 371: We need to check the return value of `eglQueryDevicesEXT` (there's are more unchecked calls below).

include/wlr/render/egl.h
```diff
@@ -87,6 +87,15 @@ struct wlr_egl {
 struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 	const EGLint *config_attribs);
 
+/**
+ * Creates and EGL contect from a given drm fd. This function uses the
```
myfreeweb, line 91: typo: `s/contect/context/`

render/egl.c
bl4ckb0ne: I thought it was a "load or die" function https://github.com/swaywm/wlroots/blob/master/render/egl.c#L91
I'll pull the exts list before loading then.

emersion: Yeah, EGL doesn't guarantee that sadly. The assertions are just in case the EGL impl is nice enough.

include/wlr/render/egl.h
bl4ckb0ne: good catch!

render/egl.c
bl4ckb0ne: thanks for the tip!

render/egl.c
bl4ckb0ne: Fixed in v2

bl4ckb0ne: I ended up splitting the `wlr_egl_create` function in two to get the basic EGL functions first, and then initialize the rest.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+
 	return egl;
 
 error:
 	eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
 	if (egl->display) {
 		eglTerminate(egl->display);
 	}
+	if (egl) {
+		free(egl);
```
emersion, line 396: Since `free(NULL)` is a no-op, we don't need the `if` here.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+
+	return egl;
+
+error:
+	eglMakeCurrent(EGL_NO_DISPLAY, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
+	if (egl->display) {
+		eglTerminate(egl->display);
+	}
+	if (egl) {
+		free(egl);
```
emersion, line 477: Since `free(NULL)` is a no-op, we don't need the `if` here.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	if (!egl->procs.eglQueryDevicesEXT(MAX_DEVICES, devices, &nb_devices)) {
+		wlr_log(WLR_ERROR, "Failed to query EGL devices");
+		goto error;
+	}
+
+	drmDevicePtr device = NULL;
+	drmGetDevice(drm_fd, &device);
+
+	EGLDeviceEXT *egl_device = NULL;
+	for(int i = 0; i < nb_devices; i++) {
```
emersion, line 448: Style: missing space after `for`.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+		goto error;
+	}
+
+	if (!egl->procs.eglQueryDevicesEXT(MAX_DEVICES, devices, &nb_devices)) {
+		wlr_log(WLR_ERROR, "Failed to query EGL devices");
+		goto error;
+	}
+
+	drmDevicePtr device = NULL;
+	drmGetDevice(drm_fd, &device);
```
emersion, line 445: `drmGetDevice` can fail, we should handle the error appropriately.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	}
+
+	if (egl->procs.eglQueryDevicesEXT == NULL) {
+		wlr_log(WLR_ERROR, "EGL_EXT_device_enumeration not supported");
+		goto error;
+	}
+
+	static const EGLint MAX_DEVICES = 64;
+	EGLint nb_devices = 0;
+	if (!egl->procs.eglQueryDevicesEXT(MAX_DEVICES, NULL, &nb_devices)) {
```
emersion, line 428: The spec says:

> If devices is NULL, then max_devices will be ignored, no devices will be returned in devices, and num_devices will be set to the number of supported devices in the system.

So I think we can just drop `MAX_DEVICES` here, and call `eglQueryDevicesEXT(0, NULL, &nb_devices)`.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+		goto error;
+	}
+
+	EGLDeviceEXT *devices = calloc(nb_devices, sizeof(EGLDeviceEXT));
+	if (devices == NULL) {
+		wlr_log_errno(WLR_ERROR, "Failed to allocate device list");
+		goto error;
+	}
+
+	if (!egl->procs.eglQueryDevicesEXT(MAX_DEVICES, devices, &nb_devices)) {
```
emersion, line 439: And here we can just pass `nb_devices` instead of `MAX_DEVICES`.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+		wlr_log(WLR_ERROR, "Failed to query EGL devices");
+		goto error;
+	}
+
+	drmDevicePtr device = NULL;
+	drmGetDevice(drm_fd, &device);
+
+	EGLDeviceEXT *egl_device = NULL;
```
emersion, line 447: I don't think this should be a pointer. This will be set to an element from the `EGLDeviceEXT *devices` variable, so `devices[i]` should be an `EGLDeviceEXT` and not an `EGLDeviceEXT *`.

Mind, EGL/GL is pretty unhelpful wrt. type safety.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+		wlr_log_errno(WLR_ERROR, "Failed to allocate device list");
+		goto error;
+	}
+
+	if (!egl->procs.eglQueryDevicesEXT(MAX_DEVICES, devices, &nb_devices)) {
+		wlr_log(WLR_ERROR, "Failed to query EGL devices");
+		goto error;
+	}
+
+	drmDevicePtr device = NULL;
```
emersion, line 444: Please don't use the `XXXPtr` types from libdrm, prefer types like `drmDevice *`.

render/egl.c
```diff
@@ -347,13 +369,113 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+				break;
+		}
+	}
+
+	if (egl_device == NULL) {
+		wlr_log(WLR_ERROR, "Failed to find EGLDeviceEXT");
+		goto error;
+	}
+
+	if (!egl_init(egl, EGL_PLATFORM_DEVICE_EXT, egl_device, NULL)) {
```
emersion, line 464: Note: we should still pass an `EGLDeviceEXT *` here. Mesa will de-reference the pointer and expect an `EGLDeviceEXT`.

render/egl.c
```diff
@@ -250,26 +263,34 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+			wlr_log(WLR_ERROR, "EGL_EXT_client_extensions not supported");
+		} else {
+			wlr_log(WLR_ERROR, "Failed to query EGL client extensions");
+		}
+		return false;
+	}
+
 	const char *device_exts_str = NULL;
 	if (check_egl_ext(client_exts_str, "EGL_EXT_device_query")) {
 		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
```
emersion, line 278: Maybe we can move this `load_egl_proc` call into `egl_create` as well, so that we load procs for client extensions all in the same spot?

emersion (changes requested):
Looks pretty good, here are a few comments.

render/egl.c
bl4ckb0ne: It seems I was calling `goto error` if `egl_create` fails, which was a bit dumb. I changed the error handling logic in `wlr_egl_create`.

render/egl.c
bl4ckb0ne: Same as my previous comment, but this time I simply return `NULL`if `egl_create` fails.

render/egl.c
bl4ckb0ne:
```
typedef void *EGLDeviceEXT;
```
I see your point, EGL is indeed pretty unhelpful.

render/egl.c
bl4ckb0ne: 100% agreed

render/egl.c
bl4ckb0ne: Sure!

render/egl.c
```diff
@@ -250,26 +265,31 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			"eglQueryWaylandBufferWL");
 	}
 
+	const char *client_exts_str = eglQueryString(EGL_NO_DISPLAY, EGL_EXTENSIONS);
```
emersion, line 268: Instead of querying again the client exts, I'd prefer to add a new `bool device_query_ext` field to `wlr_egl.exts`.

render/egl.c
```diff
@@ -250,26 +265,31 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
-			"eglQueryDisplayAttribEXT");
-		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
-			"eglQueryDeviceStringEXT");
-
 		EGLAttrib device_attrib;
 		if (!egl->procs.eglQueryDisplayAttribEXT(egl->display,
 				EGL_DEVICE_EXT, &device_attrib)) {
 			wlr_log(WLR_ERROR, "eglQueryDisplayAttribEXT(EGL_DEVICE_EXT) failed");
-			goto error;
+			return false;
```
emersion, line 284: This function returns a pointer, so should s/false/NULL/

emersion: Nevermind, I got confused.

render/egl.c
```diff
@@ -347,13 +368,110 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+		goto error;
+	}
+
+	EGLDeviceEXT egl_device;
+	for (int i = 0; i < nb_devices; i++) {
+		const char *egl_device_name = egl->procs.eglQueryDeviceStringEXT(
+				devices[i], EGL_DRM_DEVICE_FILE_EXT);
+
+		if (device_has_name(device, egl_device_name)) {
+				wlr_log(WLR_DEBUG, "Using EGL device %s", egl_device_name);
```
emersion, line 451: Style: extra indent here

render/egl.c
```diff
@@ -347,13 +368,110 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	int ret = drmGetDevice(drm_fd, &device);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to get DRM device: %s", strerror(-ret));
+		goto error;
+	}
+
+	EGLDeviceEXT egl_device;
+	for (int i = 0; i < nb_devices; i++) {
+		const char *egl_device_name = egl->procs.eglQueryDeviceStringEXT(
+				devices[i], EGL_DRM_DEVICE_FILE_EXT);
```
emersion, line 448: I think this function can return NULL.

render/egl.c
```diff
@@ -347,13 +368,110 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	}
+
+	drmDevice *device = NULL;
+	int ret = drmGetDevice(drm_fd, &device);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to get DRM device: %s", strerror(-ret));
+		goto error;
+	}
+
+	EGLDeviceEXT egl_device;
```
emersion, line 445: This should be initialized to NULL, so that we don't read an undefined value if `nb_devices == 0`.

render/egl.c
bl4ckb0ne: I also moved the output if the `client_exts_str` from `egl_init` to `egl_create`, let me know if you want something else.

emersion: Yes, perfectly fine by me.

render/wlr_renderer.c
```diff
@@ -251,22 +251,12 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
...
-	struct wlr_egl *egl = wlr_egl_create(EGL_PLATFORM_GBM_KHR, gbm_device,
-			NULL);
+	struct wlr_egl *egl = wlr_egl_create_from_drm_fd(drm_fd);
 	if (egl == NULL) {
 		wlr_log(WLR_ERROR, "Could not initialize EGL");
-		gbm_device_destroy(gbm_device);
 		return NULL;
 	}
 
-	egl->gbm_device = gbm_device;
```
emersion, line 268: This field is now unused, I think we can get rid of it (and also get rid of the `gbm.h` include).

emersion (changes requested):
LGTM apart from this last comment!

render/wlr_renderer.c
bl4ckb0ne: Done :D

emersion (approved):
Alright, this LGTM, thanks!

Issue #2586, opened 2020-12-31
backend/x11: X11 does not support required DRI3 version


CaptainBloodz:
```
sway -d
00:00:00.000 [sway/main.c:346] Sway version 1.5-53f5197c (Dec 31 2020, branch 'HEAD')
00:00:00.021 [sway/main.c:154] Linux amd64 5.10.4-gentoo-classic #2 SMP Thu Dec 31 13:26:15 EAT 2020 x86_64 Intel(R) Core(TM) i3-6100 CPU @ 3.70GHz GenuineIntel GNU/Linux
00:00:00.021 [sway/main.c:170] Contents of /etc/lsb-release:
00:00:00.021 [sway/main.c:154] DISTRIB_ID="Gentoo"
00:00:00.021 [sway/main.c:170] Contents of /etc/os-release:
00:00:00.021 [sway/main.c:154] NAME=Gentoo
00:00:00.021 [sway/main.c:154] ID=gentoo
00:00:00.021 [sway/main.c:154] PRETTY_NAME="Gentoo/Linux"
00:00:00.021 [sway/main.c:154] ANSI_COLOR="1;32"
00:00:00.021 [sway/main.c:154] HOME_URL="https://www.gentoo.org/"
00:00:00.021 [sway/main.c:154] SUPPORT_URL="https://www.gentoo.org/support/"
00:00:00.021 [sway/main.c:154] BUG_REPORT_URL="https://bugs.gentoo.org/"
00:00:00.021 [sway/main.c:170] Contents of /etc/gentoo-release:
00:00:00.021 [sway/main.c:154] Gentoo Base System release 2.7
00:00:00.022 [sway/main.c:142] LD_LIBRARY_PATH=
00:00:00.022 [sway/main.c:142] LD_PRELOAD=
00:00:00.022 [sway/main.c:142] PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin:/usr/lib64/opencascade-7.3.0/ros/bin:/usr/lib/llvm/11/bin:/usr/lib/llvm/10/bin
00:00:00.022 [sway/main.c:142] SWAYSOCK=
00:00:00.022 [sway/server.c:43] Preparing Wayland server initialization
00:00:00.022 [wlr] [backend/x11/backend.c:412] X11 does not support required DRI3 version
00:00:00.022 [sway/server.c:49] Unable to create backend
```

Used to work a couple of weeks ago.

[...]

PR #2655, opened 2021-01-17
backend/x11: add support for DRI3 1.0

emersion:
Closes: #2586


emersion:
I've tested on xf86-video-amdgpu, and jbeich has tested on xf86-video-intel.

PR #2656, opened 2021-01-17
backend/headless: add WLR_HEADLESS_DEVICE env variable

emersion:
This lets users select the DRM node to use for rendering.

* * *

It would be nice to have a generalized env variable to select the rendering device regardless of the backend. However this would break the DRM backend, which expects buffers coming from the primary device.

### DIFF
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -153,6 +153,16 @@ static bool backend_init(struct wlr_headless_backend *backend,
 }
 
 static int open_drm_render_node(void) {
+	const char *name = getenv("WLR_HEADLESS_DEVICE");
+	if (name != NULL) {
+		wlr_log(WLR_DEBUG, "Opening '%s' due to WLR_HEADLESS_DEVICE", name);
+		int fd = open(name, O_RDWR | O_CLOEXEC);
+		if (fd < 0) {
+			wlr_log_errno(WLR_ERROR, "Failed to open '%s'", name);
+		}
+		return fd;
+	}
+
 	uint32_t flags = 0;
 	int devices_len = drmGetDevices2(flags, NULL, 0);
 	if (devices_len < 0) {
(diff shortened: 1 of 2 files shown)


ammen99:
> It would be nice to have a generalized env variable to select the rendering device regardless of the backend. However this would break the DRM backend, which expects buffers coming from the primary device.

Hmm, how would it break the DRM backend? The env var would be selecting the primary GPU device, both for DRM, headless, etc.

emersion:
First, we'll always want the allocator and the renderer to operate on the same device. I don't think anything supports rendering on a non-local DMA-BUF.

Then, the DRM backend has two modes of operation: parent or child. In parent mode, the buffers submitted are directly scanned out. In child mode, the buffers are first copied locally, and then scanned out.

Right now the DRM backend mode is picked at creation time, and the renderer/allocator is created after the backend (because by default the renderer/allocator should default to the backend's device). If we were to add a generic "pick this device for rendering" env var, we would need to make sure the DRM parent matches the rendering device. So far `WLR_DRM_*` env vars always were to configure the DRM backend, and having two env vars would leave room for a mismatch.

We could also change how the DRM backend operates: instead of deciding of the parent/child mode at creation time, maybe we could do something smarter like try to import the buffer to KMS and perform a copy if that fails (assuming it's a remote DMA-BUF). That's a little fragile though, the import could fail because of another reason and a copy may not help. Unfortunately there's no "is this DMA-BUF local?" syscall.

ammen99:
> Maybe we could make it so the ordering in `WLR_DRM_DEVICES` doesn't matter and the primary device is selected via the generic env var.

This is what I had in mind too.

emersion:
Note, we should probably error out if `WLR_DRM_DEVICES` doesn't contain the primary device env var.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2656

PR #2657, opened 2021-01-17
backend/drm: handle add_drm_card event

neon64:
Depends on / includes changes from https://github.com/swaywm/wlroots/pull/2575, supersedes https://github.com/swaywm/wlroots/pull/2423.

The main API is pretty much identical to what emersion suggested in https://github.com/swaywm/wlroots/pull/2423#issuecomment-751475833. At the moment this backend monitor struct is created inside of `wlr_backend_autocreate`, and it is destroyed whenever one of the multi backend, the primary drm backend, or the wlr_session is destroyed
```c
struct wlr_drm_backend_monitor {
    struct wlr_backend *multi;
    struct wlr_backend *primary_drm;
    struct wlr_session *session;

    struct wl_listener multi_destroy;
    struct wl_listener primary_drm_destroy;
    struct wl_listener session_destroy;
    struct wl_listener session_add_drm_card;
};

struct wlr_drm_backend_monitor *wlr_drm_backend_monitor_create(struct wlr_backend *multi, struct wlr_backend *primary_drm, struct wlr_session *session);
```

Alas there is a pretty major issue I've encountered so far: https://github.com/neon64/wlroots/issues/2. Hotplugging doesn't work at all because the EGLDevice returned for the nouveau card isn't actually valid. In order to make it work, you can load both drivers before starting sway, but then force sway to only use one card initially (`WLR_DRM_DEVICES=/dev/dri/card0`), but then manually load the other one with `sudo udevadm ...`. However this basically defeats the point of hotplugging. I think this is a bug in libglvnd - anyway the issue is described in much greater detail there.


emersion:
I've rebased this against master.

Do you think you'll have time to continue working on this?

neon64:
> Do you think you'll have time to continue working on this?

I'd like to - earlier in the year I was stalled because I didn't have access to an external monitor to test with, and also was blocked by https://gitlab.freedesktop.org/glvnd/libglvnd/-/issues/215. I note that a [patch](https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/239) has been submitted - I need to give it a go and see if it works though

emersion:
Note that the `EGLDevice` stuff has been removed for now, and can't be re-introduced for now without a fallback because it breaks on split render/display SoCs. So we should be able to add hotplug support without waiting for glvnd.

neon64:
I've finally had a chance to test this out again. Unfortunately without https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/239 , creating a renderer for a previously-unseen GPU still fails

https://github.com/swaywm/wlroots/blob/430018465385aba2ca4e71ab45ccb1f8a60bf99e/render/egl.c#L286

I believe this is consistent with my understanding of the current libglvnd behaviour - the list of valid devices is lazy-initialized once on wlroots startup, and thus the EGLDevice returned by the call to `eglQueryDisplayAttribEXT(EGL_DEVICE_EXT)` is deemed invalid the next time we try to use it, because this device wasn't returned by the initial call to `eglQueryDevicesEXT`

Good news is, after compiling the libglvnd PR, and using that instead, I can get hotplugging / unplugging working.

That said, repeatedly enabling/disabling nouveau seems to be touching some fairly brittle code-paths in kernel/driver code, as I have encountered a few kernel panics when turning things on/off in quick succession. Unfortunately I haven't reproduced those panics reliably in order to get logs / stack traces yet.

backend/drm/drm.c
```diff
@@ -1540,16 +1540,16 @@ static void page_flip_handler(int fd, unsigned seq,
...
 		.version = 3,
 		.page_flip_handler2 = page_flip_handler,
 	};
 
 	if (drmHandleEvent(fd, &event) != 0) {
 		wlr_log(WLR_ERROR, "drmHandleEvent failed");
-		wl_display_terminate(drm->display);
+		wlr_log(WLR_ERROR, "ignoring failed drmHandleEvent - since could be triggered by device unplug");
+		// TODO: how to handle this gracefully. Ignoring the error doesn't seem good long-term.
+		// wl_display_terminate(drm->display);
```
emersion, line 1552: Can we check `mask` here?

```c
if (mask & (WL_EVENT_HANGUP | WL_EVENT_ERROR)) {
    // Destroy backend
}
```

backend/backend.c
```diff
@@ -367,6 +367,11 @@ struct wlr_backend *wlr_backend_autocreate(struct wl_display *display) {
 
 	struct wlr_backend *primary_drm =
 		attempt_drm_backend(display, backend, multi->session);
+
+	// discard the returned pointer - the monitor should destroy itself
+	// when the multi-backend is destroyed
+	wlr_drm_backend_monitor_create(backend, primary_drm, multi->session);
```
emersion, line 373: We should create the monitor after the `if (!primary_drm)` check.

backend/drm/monitor.c
```diff
@@ -0,0 +1,103 @@
...
+	wlr_log(WLR_DEBUG, "Received session_add_drm_card signal: %s", event->path);
+
+	struct wlr_device *wlr_dev =
+		session_open_if_kms(backend_monitor->session, event->path);
+	if (!wlr_dev) {
+		wlr_log(WLR_ERROR, "Unable to open %s as DRM device", event->path);
+		return;
+	}
+
+	wlr_log(WLR_DEBUG, "Opened %s as DRM device, fd = %d", event->path, wlr_dev->fd);
```
emersion, line 33: Logging the FD number is probably not very useful here

backend/drm/monitor.c
```diff
@@ -0,0 +1,103 @@
+#include <wlr/backend/drm.h>
+#include <wlr/util/log.h>
+#include <backend/drm/drm.h>
+#include <backend/multi.h>
+#include <stdlib.h>
+#include <backend/session/session.h>
+
+static void destroy_drm_backend_monitor(struct wlr_drm_backend_monitor* monitor) {
```
emersion, line 8: Style: `*` on the right

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2657

PR #2658, opened 2021-01-17
xwayland/cursor: free cursor in wlr_xwayland_destroy

neon64:
Running the asan build has detected what seems like a large number of memory leaks on shutdown. This was the first one I saw and seems like a relatively simple fix (the old cursor memory is freed in [wlr_xwayland_set_cursor.c:119](https://github.com/neon64/wlroots/blob/28584167dd50b388e3e1755be2946e45d9d0c2bd/xwayland/xwayland.c#L119) but not on shutdown)

### DIFF
--- a/xwayland/xwayland.c
+++ b/xwayland/xwayland.c
@@ -64,6 +64,7 @@ void wlr_xwayland_destroy(struct wlr_xwayland *xwayland) {
 
 	wl_list_remove(&xwayland->server_destroy.link);
 	wl_list_remove(&xwayland->server_ready.link);
+	free(xwayland->cursor);
 
 	wlr_xwayland_set_seat(xwayland, NULL);
 	wlr_xwayland_server_destroy(xwayland->server);


xwayland/xwayland.c
```diff
@@ -64,6 +64,9 @@ void wlr_xwayland_destroy(struct wlr_xwayland *xwayland) {
 
 	wl_list_remove(&xwayland->server_destroy.link);
 	wl_list_remove(&xwayland->server_ready.link);
+	if (xwayland->cursor) {
+		free(xwayland->cursor);
```
emersion, line 68: Since `free(NULL)` is a no-op, no need for an `if` here

emersion:
> a large number of memory leaks on shutdown

Yeah, most of these are due to Sway not cleaning up its structures on shutdown.

emersion (approved):
LGTM, can you squash these two commits into a single one?

neon64:
> Yeah, most of these are due to Sway not cleaning up its structures on shutdown.

huh, is that by design or is it just not worth fixing all at once?

emersion:
Partly because everything is cleaned up by the OS when the process exits, so it's not really a big deal. Partly because nobody has bothered yet.

However it does make _real_ memory leaks harder to spot, because ASan/Valgrind don't produce very useful info. So I'd merge patches to clean that up.

emersion:
Thanks for the fix!

Issue #2659, opened 2021-01-17
backend/x11: cursor hidden causes issues when debugging crash with gdb


imirkin:
I was debugging a crash in the nvc0 driver which sway triggered. I did this with the X11 backend (because X11 is what I use, and also because it's nice to be able to have your regular dev environment + gdb when debugging crashes).

The one downside of this is that while sway had abort'd, the window wasn't gone as far as Xorg was concerned. So the cursor remained hidden, making it difficult to navigate my regular environment.

Either wlroots shouldn't hide the cursor in the first place and do an "accelerated" cursor (i.e. set X's cursor), or there should be a way to allow the native cursor to not be hidden.

imirkin:
@emersion mentioned that as a first pass we could XDefineCursor with a fully transparent cursor instead of hiding the cursor outright. That way the cursor is still there, and when not over the window should return to the "regular" cursor.

A second pass could be to try to hook up the full pipeline into XRenderCreateCursor. It's unclear whether this can work directly with DRI3 Pixmaps or even with a Window + Present, or if we just need to read the data out "by hand".

Worth noting that in practice GPUs support only a very limited quantity of cursor sizes, not sure what form the cursor textures come in.

PR #2660, opened 2021-01-17
backend/x11: set a blank cursor rather than hiding globally

imirkin:
This actually simplifies the logic since we no longer have to wait for
enter/leave events, and also improves the UX when e.g. handling a crash
with gdb attached.

See #2659

### DIFF
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -613,6 +613,23 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
 	x11->display_destroy.notify = handle_display_destroy;
 	wl_display_add_destroy_listener(display, &x11->display_destroy);
 
+	// Create an empty pixmap to be used as the cursor. The
+	// default GC foreground is 0, and that is what it will be
+	// filled with.
+	xcb_pixmap_t blank = xcb_generate_id(x11->xcb);
+	xcb_create_pixmap(x11->xcb, 1, blank, x11->screen->root, 1, 1);
+	xcb_gcontext_t gc = xcb_generate_id(x11->xcb);
+	xcb_create_gc(x11->xcb, gc, blank, 0, NULL);
+	xcb_rectangle_t rect = { .x = 0, .y = 0, .width = 1, .height = 1 };
+	xcb_poly_fill_rectangle(x11->xcb, blank, gc, 1, &rect);
+
+	x11->cursor = xcb_generate_id(x11->xcb);
+	xcb_create_cursor(x11->xcb, x11->cursor, blank, blank,
+		0, 0, 0, 0, 0, 0, 0, 0);
+
+	xcb_free_gc(x11->xcb, gc);
+	xcb_free_pixmap(x11->xcb, blank);
+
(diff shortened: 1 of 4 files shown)


imirkin:
Actually I should probably create the cursor once globally and then use it, rather than leaking it at per-output creation (not that I'd expect that to happen too frequently...)

imirkin:
> Actually I should probably create the cursor once globally and then use it, rather than leaking it at per-output creation (not that I'd expect that to happen too frequently...)

Actually can't do that. The Pixmap needs to be attached to a Window. I could come up with a way to not leak the cursor, but since outputs with the X11 backend aren't really dynamic, probably not necessary.

imirkin:
Duh, just use the screen root (which the window is rendered on). And now we can set the cursor as part of the window creation, so simpler.

imirkin:
Let me know if the dep on xcb-image is undesirable, the cursor could also be created using XFillRectangle. This involves creating a GC, so more X gunk to do. But definitely doable.

imirkin:
FWIW the fill rectangle variant looks like

```
        xcb_pixmap_t blank = xcb_generate_id(x11->xcb);
        xcb_create_pixmap(x11->xcb, 1, blank, x11->screen->root, 1, 1);
        xcb_gcontext_t gc = xcb_generate_id(x11->xcb);
        xcb_create_gc(x11->xcb, gc, blank, 0, NULL);
        xcb_rectangle_t rect = {
          .x = 0,
          .y = 0,
          .width = 1,
          .height = 1,
        };
        xcb_poly_fill_rectangle(x11->xcb, blank, gc, 1, &rect);
```

Let me know which is preferred.

backend/x11/backend.c
```diff
@@ -613,6 +614,13 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
 	x11->display_destroy.notify = handle_display_destroy;
 	wl_display_add_destroy_listener(display, &x11->display_destroy);
 
+	uint8_t zero = 0;
+	xcb_pixmap_t blank = xcb_create_pixmap_from_bitmap_data(
+		x11->xcb, x11->screen->root, &zero, 1, 1, 1, 0, 0, NULL);
+	x11->cursor = xcb_generate_id(x11->xcb);
```
emersion, line 620: Can we cleanup the cursor and pixmap in `backend_destroy`? wlroots is a library, and compositors are free to create/destroy as many backends as they want, so I'd prefer not to leak anything if possible.

emersion (changes requested):
Thanks for working on this! Just a small comment below.

> dep on xcb-image

I don't mind the xcb-image dependency.

>  I could come up with a way to not leak the cursor

The backend and output both have functions called when destroyed: `backend_destroy` and `output_destroy`.

backend/x11/backend.c
imirkin: backend_destroy destroys the connection, which should also free any server-side resources. However the pixmap should be freed earlier, will do.

emersion (approved):
LGTM, thanks a lot!

Issue #2399, opened 2020-09-08
Pixman software renderer


emersion:
After renderer v6, we'll be able to start working on a pixman software renderer. It'll probably work this way:

- Allocator: add a dumb buffer allocator. The returned `wlr_buffer` can be exported to a DMA-BUF via `drmPrimeHandleToFD`.
- Renderer: import the DMA-BUF via `drmPrimeFDToHandle` and mmap the dumb buffer.
- DRM backend: import the DMA-BUF via `drmPrimeFDToHandle` (or GBM?) and scan it out
- Other backends: ???

Need to be careful wrt. GEM handle ref-counting.

bl4ckb0ne:
Where should `wlr_egl` go? Sould it belong to the implementation of the `wlr_renderer`?

Also how would the renderer choosing strategy work? egl then pixman, always egl unless an env var is specified?

emersion:
>Where should wlr_egl go? Sould it belong to the implementation of the wlr_renderer?

Probably in `wlr_gles2_renderer` yes.

>Also how would the renderer choosing strategy work? egl then pixman, always egl unless an env var is specified?

By default, if there's a render node, try GLES2. Otherwise try pixman. We probably want an env var to force a specific renderer.

emersion:
More thoughts:

- As discussed on IRC, the DRM dumb buffer allocator will get GEM handles and somehow needs to share them with the DRM backend. We need to be careful wrt. ref'counting issues with GEM handles.
  - One solution is to open the primary DRM node multiple times (via the session) and then the allocator can do some `drmPrimeHandleToFD` calls to implement `wlr_buffer_impl.get_dmabuf`.
  - Another solution would be to have a `wlr_buffer_impl.get_gem_handle` function, but this would be fragile as it would completely break if both parties don't have the same file description.
- The DRM dumb buffer allocator should probably check for `DRM_CAP_DUMB_BUFFER`.
- For the Wayland/X11/headless backends, DRM dumb buffers are not appropriate. We'll probably need a shm allocator (which uses `shm_open`), use `wl_shm` on the Wayland backend, use MIT-SHM's `AttachFd` on X11, and a new `wlr_buffer_impl.get_shm_fd` function to share the FDs between the allocator and the backends.
- We need a way to access the buffers from the Pixman renderer.
  - We could special-case the DRM dumb buffer allocator in the Pixman renderer, but it would prevent the shm allocator and third-party allocators to work.
[...]

PR #2661, opened 2021-01-18
Pixman renderer

bl4ckb0ne:
Closes: https://github.com/swaywm/wlroots/issues/2399

### DIFF
--- a/include/render/pixman.h
+++ b/include/render/pixman.h
@@ -1,0 +1,49 @@
+#ifndef RENDER_PIXMAN_H
+#define RENDER_PIXMAN_H
+
+#include <wlr/render/pixman.h>
+#include <wlr/render/wlr_renderer.h>
+#include <wlr/render/drm_format_set.h>
+#include "render/pixel_format.h"
+
+struct wlr_pixman_pixel_format {
+	uint32_t drm_format;
+	pixman_format_code_t pixman_format;
+};
+
+struct wlr_pixman_buffer;
+
+struct wlr_pixman_renderer {
+	struct wlr_renderer wlr_renderer;
+
+	struct wl_list buffers; // wlr_pixman_buffer.link
+
+	struct wlr_pixman_buffer *current_buffer;
+	int32_t width, height;
+
+	struct wlr_drm_format_set drm_formats;
+};
+
+struct wlr_pixman_buffer {
+	struct wlr_buffer *buffer;
+	struct wlr_pixman_renderer *renderer;
+
+	pixman_image_t *image;
+
+	struct wl_listener buffer_destroy;
+	struct wl_list link; // wlr_pixman_renderer.buffers
+};
+
+struct wlr_pixman_texture {
+	struct wlr_texture wlr_texture;
+	struct wlr_pixman_renderer *renderer;
+
+	void *data;
+	pixman_image_t *image;
+	const struct wlr_pixel_format_info *format;
+};
+
+pixman_format_code_t get_pixman_format_from_drm(uint32_t fmt);
+const uint32_t *get_pixman_drm_formats(size_t *len);
+
+#endif
--- a/meson.build
+++ b/meson.build
@@ -17,8 +17,13 @@ project(
 # between minor releases.
 soversion = 9
 
+little_endian = host_machine.endian() == 'little'
+big_endian = host_machine.endian() == 'big'
+
 add_project_arguments([
 	'-DWLR_USE_UNSTABLE',
+	'-DWLR_LITTLE_ENDIAN=@0@'.format(little_endian.to_int()),
+	'-DWLR_BIG_ENDIAN=@0@'.format(big_endian.to_int()),
 ], language: 'c')
 
 cc = meson.get_compiler('c')
--- a/render/pixman/pixel_format.c
+++ b/render/pixman/pixel_format.c
@@ -1,0 +1,59 @@
+#include <drm_fourcc.h>
+#include <wlr/util/log.h>
+
+#include "render/pixman.h"
+
+static const struct wlr_pixman_pixel_format formats[] = {
+	{
+		.drm_format = DRM_FORMAT_ARGB8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_b8g8r8a8,
+#else
+		.pixman_format = PIXMAN_a8r8g8b8,
+#endif
+	},
+	{
+		.drm_format = DRM_FORMAT_XBGR8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_r8g8b8x8,
+#else
+		.pixman_format = PIXMAN_x8b8g8r8,
+#endif
+	},
+	{
+		.drm_format = DRM_FORMAT_XRGB8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_b8g8r8x8,
+#else
+		.pixman_format = PIXMAN_x8r8g8b8,
+#endif
+	},
+	{
+		.drm_format = DRM_FORMAT_ABGR8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_r8g8b8a8,
+#else
+		.pixman_format = PIXMAN_a8b8g8r8,
+#endif
+	}
+};
+
+pixman_format_code_t get_pixman_format_from_drm(uint32_t fmt) {
+	for (size_t i = 0; i < sizeof(formats) / sizeof(*formats); ++i) {
+		if (formats[i].drm_format == fmt) {
+			return formats[i].pixman_format;
+		}
+	}
+
+	wlr_log(WLR_ERROR, "DRM format 0x%"PRIX32" has no pixman equivalent", fmt);
+	return 0;
+}
+
+const uint32_t *get_pixman_drm_formats(size_t *len) {
+	static uint32_t drm_formats[sizeof(formats) / sizeof(formats[0])];
+	*len = sizeof(formats) / sizeof(formats[0]);
+	for (size_t i = 0; i < sizeof(formats) / sizeof(formats[0]); i++) {
+		drm_formats[i] = formats[i].drm_format;
+	}
+	return drm_formats;
+}
--- a/render/pixman/renderer.c
+++ b/render/pixman/renderer.c
@@ -1,0 +1,383 @@
+#include <assert.h>
+#include <drm_fourcc.h>
+#include <pixman.h>
+#include <stdlib.h>
+#include <wayland-server.h>
+#include <wlr/render/interface.h>
+#include <wlr/types/wlr_matrix.h>
+#include <wlr/util/log.h>
+
+#include "render/pixman.h"
+#include "types/wlr_buffer.h"
+
+static const struct wlr_renderer_impl renderer_impl;
+
+static struct wlr_pixman_renderer *get_renderer(
+		struct wlr_renderer *wlr_renderer) {
+	assert(wlr_renderer->impl == &renderer_impl);
+	return (struct wlr_pixman_renderer *)wlr_renderer;
+}
+
+static struct wlr_pixman_buffer *get_buffer(
+		struct wlr_pixman_renderer *renderer, struct wlr_buffer *wlr_buffer) {
+	struct wlr_pixman_buffer *buffer;
+	wl_list_for_each(buffer, &renderer->buffers, link) {
+		if (buffer->buffer == wlr_buffer) {
+			return buffer;
+		}
+	}
+	return NULL;
+}
+
+static const struct wlr_texture_impl texture_impl;
+
+static struct wlr_pixman_texture *get_texture(
+		struct wlr_texture *wlr_texture) {
+	assert(wlr_texture->impl == &texture_impl);
+	return (struct wlr_pixman_texture *)wlr_texture;
+}
+
+static bool texture_is_opaque(struct wlr_texture *wlr_texture) {
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+	return texture->format->has_alpha;
+}
+
+static void texture_destroy(struct wlr_texture *wlr_texture) {
+	if (wlr_texture == NULL) {
+		return;
+	}
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+
+	pixman_image_unref(texture->image);
+	free(texture->data);
+	free(texture);
+}
+
+static const struct wlr_texture_impl texture_impl = {
+	.is_opaque = texture_is_opaque,
+	.destroy = texture_destroy,
+};
+
+struct wlr_pixman_texture *pixman_create_texture(
+		struct wlr_texture *wlr_texture, struct wlr_pixman_renderer *renderer);
+
+static void handle_destroy_buffer(struct wl_listener *listener, void *data) {
+	struct wlr_pixman_buffer *buffer =
+		wl_container_of(listener, buffer, buffer_destroy);
+	wl_list_remove(&buffer->link);
+	wl_list_remove(&buffer->buffer_destroy.link);
+
+	pixman_image_unref(buffer->image);
+
+	free(buffer);
+}
+
+static struct wlr_pixman_buffer *create_buffer(
+		struct wlr_pixman_renderer *renderer, struct wlr_buffer *wlr_buffer) {
+	struct wlr_pixman_buffer *buffer = calloc(1, sizeof(*buffer));
+	if (buffer == NULL) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+	buffer->buffer = wlr_buffer;
+	buffer->renderer = renderer;
+
+	uint32_t drm_format;
+	struct wlr_dmabuf_attributes dmabuf = {0};
+	struct wlr_shm_attributes shm = {0};
+	if (wlr_buffer_get_dmabuf(wlr_buffer, &dmabuf)) {
+		drm_format = dmabuf.format;
+	} else if (wlr_buffer_get_shm(wlr_buffer, &shm)) {
+		drm_format = shm.format;
+	} else {
+		goto error_buffer;
+	}
+
+	uint32_t format = get_pixman_format_from_drm(drm_format);
+
+	void *data = NULL;
+	size_t stride;
+	if (!buffer_get_data_ptr(wlr_buffer, &data, &stride)) {
+		wlr_log(WLR_ERROR, "Failed to get buffer data");
+		goto error_buffer;
+	}
+
+	buffer->image = pixman_image_create_bits(format, wlr_buffer->width,
+			wlr_buffer->height, data, stride);
+	if (!buffer->image) {
+		wlr_log(WLR_ERROR, "Failed to allocate pixman image");
+		goto error_buffer;
+	}
+
+	buffer->buffer_destroy.notify = handle_destroy_buffer;
+	wl_signal_add(&wlr_buffer->events.destroy, &buffer->buffer_destroy);
+
+	wl_list_insert(&renderer->buffers, &buffer->link);
+
+	wlr_log(WLR_DEBUG, "Created pixman buffer %dx%d",
+		wlr_buffer->width, wlr_buffer->height);
+
+	return buffer;
+
+error_buffer:
+	free(buffer);
+	return NULL;
+}
+
+static void pixman_begin(struct wlr_renderer *wlr_renderer, uint32_t width,
+		uint32_t height) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	renderer->width = width;
+	renderer->height = height;
+}
+
+static void pixman_clear(struct wlr_renderer *wlr_renderer,
+		const float color[static 4]) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	const struct pixman_color colour = {
+		.red = color[0] * 0xFFFF,
+		.green = color[1] * 0xFFFF,
+		.blue = color[2] * 0xFFFF,
+		.alpha = color[3] * 0xFFFF,
+	};
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+
+	pixman_image_composite32(PIXMAN_OP_SRC, fill, NULL, buffer->image, 0, 0, 0,
+			0, 0, 0, renderer->width, renderer->height);
+
+	pixman_image_unref(fill);
+}
+
+static void pixman_scissor(struct wlr_renderer *wlr_renderer,
+		struct wlr_box *box) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	if (box != NULL) {
+		struct pixman_region32 region = {0};
+		pixman_region32_init_rect(&region, box->x, box->y, box->width,
+				box->height);
+		pixman_image_set_clip_region32(buffer->image, &region);
+		pixman_region32_fini(&region);
+	} else {
+		pixman_image_set_clip_region32(buffer->image, NULL);
+	}
+}
+
+static void matrix_to_pixman_transform(struct pixman_transform *transform,
+		const float mat[static 9]) {
+	struct pixman_f_transform ftr;
+	ftr.m[0][0] = mat[0];
+	ftr.m[0][1] = mat[1];
+	ftr.m[0][2] = mat[2];
+	ftr.m[1][0] = mat[3];
+	ftr.m[1][1] = mat[4];
+	ftr.m[1][2] = mat[5];
+	ftr.m[2][0] = mat[6];
+	ftr.m[2][1] = mat[7];
+	ftr.m[2][2] = mat[8];
+
+	pixman_transform_from_pixman_f_transform(transform, &ftr);
+}
+
+static bool pixman_render_subtexture_with_matrix(
+		struct wlr_renderer *wlr_renderer, struct wlr_texture *wlr_texture,
+		const struct wlr_fbox *fbox, const float matrix[static 9],
+		float alpha) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	// TODO: don't create a mask if alpha == 1.0
+	struct pixman_color mask_colour = {0};
+	mask_colour.alpha = 0xFFFF * alpha;
+	pixman_image_t *mask = pixman_image_create_solid_fill(&mask_colour);
+
+	float m[9];
+	memcpy(m, matrix, sizeof(m));
+	wlr_matrix_scale(m, 1.0 / fbox->width, 1.0 / fbox->height);
+
+	struct pixman_transform transform = {0};
+	matrix_to_pixman_transform(&transform, m);
+	pixman_transform_invert(&transform, &transform);
+
+	pixman_image_set_transform(texture->image, &transform);
+
+	// TODO clip properly with src_x and src_y
+	pixman_image_composite32(PIXMAN_OP_OVER, texture->image, mask,
+			buffer->image, 0, 0, 0, 0, 0, 0, renderer->width,
+			renderer->height);
+
+	pixman_image_unref(mask);
+
+	return true;
+}
+
+static void pixman_render_quad_with_matrix(struct wlr_renderer *wlr_renderer,
+		const float color[static 4], const float matrix[static 9]) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	struct pixman_color colour = {
+		.red = color[0] * 0xFFFF,
+		.green = color[1] * 0xFFFF,
+		.blue = color[2] * 0xFFFF,
+		.alpha = color[3] * 0xFFFF,
+	};
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+
+	float m[9];
+	memcpy(m, matrix, sizeof(m));
+
+	// TODO get the width/height from the caller instead of extracting them
+	// TODO detect non rotation with matrix[1] == 0 and matrix[4] == 0 to avoid
+	// doing the calculation
+	float width = sqrt(matrix[0] * matrix[0] + matrix[1] * matrix[1]);
+	float height = sqrt(matrix[3] * matrix[3] + matrix[4] * matrix[4]);
+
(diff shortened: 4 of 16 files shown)


render/pixman/renderer.c
```diff
@@ -0,0 +1,44 @@
...
+	.get_dmabuf_texture_formats = gles2_get_dmabuf_texture_formats,
+	.get_dmabuf_render_formats = gles2_get_dmabuf_render_formats,
+	.preferred_read_format = gles2_preferred_read_format,
+	.read_pixels = gles2_read_pixels,
+	.texture_from_pixels = gles2_texture_from_pixels,
+	.texture_from_wl_drm = gles2_texture_from_wl_drm,
+	.texture_from_dmabuf = gles2_texture_from_dmabuf,
+	.init_wl_display = gles2_init_wl_display,
+	.blit_dmabuf = gles2_blit_dmabuf,
+	.get_drm_fd = gles2_get_drm_fd,
```
emersion, line 30: We probably want to leave that one out.

bl4ckb0ne: Should I make it return -1?

emersion: Most wlr_renderer_impl functions are optional, you should be able to leave them NULL.

List of mandatory functions: https://github.com/swaywm/wlroots/blob/04d89a8bc5c5b0db0ae24ca3e2b35f5632ed6d23/render/wlr_renderer.c#L17

bl4ckb0ne: alrighty then

render/pixman/renderer.c
```diff
@@ -0,0 +1,234 @@
...
+
+	struct wlr_dmabuf_attributes dmabuf = {0};
+	if (!wlr_buffer_get_dmabuf(wlr_buffer, &dmabuf)) {
+		goto error_buffer;
+	}
+
+	uint32_t format;
+	switch (dmabuf.format) {
+	case DRM_FORMAT_XRGB8888:
+		format = PIXMAN_x8r8g8b8;
```
emersion, line 61: On a little-endian system, this should actually be `PIXMAN_b8g8r8x8`. Pixman formats use the endianness of the system, DRM formats are all little-endian.

Meson can be used to detect system endianness.

render/pixman/renderer.c
```diff
@@ -0,0 +1,234 @@
...
+static void init_dmabuf_formats(struct wlr_pixman_renderer *renderer) {
+	static const int fmts[] = {
+		DRM_FORMAT_ARGB8888,
+		DRM_FORMAT_XRGB8888,
+	};
+
+	static const size_t len = sizeof(fmts) / sizeof(fmts[0]);
+	for (size_t i = 0; i < len; ++i) {
+		wlr_drm_format_set_add(&renderer->dmabuf_render_formats, fmts[i],
+				0);
```
emersion, line 215: Nit: `DRM_FORMAT_MOD_LINEAR`

render/pixman/renderer.c
```diff
@@ -0,0 +1,234 @@
...
+static const struct wlr_renderer_impl renderer_impl = {
+	.begin = pixman_begin,
+	.clear = pixman_clear,
+	.scissor = pixman_scissor,
+	.render_subtexture_with_matrix = pixman_render_subtexture_with_matrix,
+	.render_quad_with_matrix = pixman_render_quad_with_matrix,
+	.render_ellipse_with_matrix = pixman_render_ellipse_with_matrix,
+	.get_shm_texture_formats = pixman_get_shm_texture_formats,
+	.texture_from_pixels = pixman_texture_from_pixels,
+	.get_dmabuf_render_formats = pixman_get_dmabuf_render_formats,
```
emersion, line 202: Hmm. I see the backends use `get_dmabuf_render_formats` to figure out which formats are supported by the renderer. However the Pixman renderer doesn't deal with DMA-BUFs, so it's a little bit weird to use this function.

Not sure what's the best way to address this yet.

render/pixman/renderer.c
```diff
@@ -0,0 +1,234 @@
...
+	};
+
+	static const size_t len = sizeof(fmts) / sizeof(fmts[0]);
+	for (size_t i = 0; i < len; ++i) {
+		wlr_drm_format_set_add(&renderer->dmabuf_render_formats, fmts[i],
+				0);
+	}
+}
+
+struct wlr_renderer *wlr_pixman_renderer_create(int drm_fd) {
```
emersion, line 219: I don't think we need the DRM FD for anything in the Pixman renderer.

render/pixman/renderer.c
```diff
@@ -0,0 +1,234 @@
...
+	case DRM_FORMAT_ARGB8888:
+		format = PIXMAN_a8r8g8b8;
+		break;
+	default:
+		wlr_log(WLR_ERROR, "Unsupported DRM format 0x%x\n", dmabuf.format);
+		goto error_buffer;
+	}
+
+	// TODO: ensure wlr_buffer is a drm dumb buffer
+	struct wlr_drm_dumb_buffer *drm_dumb_buffer =
```
emersion, line 72: Brain dump about accessing the dumb buffer data from the pixman renderer: https://github.com/swaywm/wlroots/issues/2399#issuecomment-779742867

render/pixman/renderer.c
bl4ckb0ne: Yeah there's a check somewhere in the DRM backend that fails if there's no dmabuf formats.

render/pixman/renderer.c
bl4ckb0ne: Fixed in #2700

render/pixman/renderer.c
```diff
@@ -0,0 +1,339 @@
...
+	#if WLR_BIG_ENDIAN
+		fmt = PIXMAN_x8r8g8b8;
+	#elif WLR_LITTLE_ENDIAN
+		fmt = PIXMAN_b8g8r8x8;
+	#else
+	#error pixman endian error
+	#endif
+		break;
+	default:
+		// TODO: wl_shm_format to string?
```
emersion, line 99: It's probably enough to just print the hex value, e.g.`"Unsupported wl_shm_format 0x%"PRIX32`. We don't really want to have a big table of `wl_shm_format` to strings, and could be missing entries in it anyways.

render/pixman/renderer.c
```diff
@@ -0,0 +1,339 @@
...
+static const struct wlr_texture_impl texture_impl = {
+	.is_opaque = texture_is_opaque,
+	.write_pixels = texture_write_pixels,
+	.to_dmabuf = texture_to_dmabuf,
+	.destroy = texture_destroy,
+};
+
+static uint32_t get_pixman_format_from_wl(enum wl_shm_format wl_fmt) {
+	uint32_t fmt = 0;
+	switch(wl_fmt) {
```
emersion, line 79: Style nit: missing space after `switch`

render/pixman/renderer.c
```diff
@@ -0,0 +1,339 @@
...
+	case WL_SHM_FORMAT_ARGB8888:
+	#if WLR_BIG_ENDIAN
```
emersion, line 81: Style: preprocessor directives should not be indented.

render/pixman/renderer.c
```diff
@@ -0,0 +1,339 @@
...
+	struct wlr_pixman_buffer *buffer = calloc(1, sizeof(*buffer));
+	if (buffer == NULL) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+	buffer->buffer = wlr_buffer;
+	buffer->renderer = renderer;
+
+	struct wlr_dmabuf_attributes dmabuf = {0};
+	if (!wlr_buffer_get_dmabuf(wlr_buffer, &dmabuf)) {
```
emersion, line 132: Other allocators may not support DMA-BUFs. It would be nice if the information we need was returned from `wlr_buffer_get_data_ptr`, or was part of the `wlr_buffer` struct.

render/pixman/renderer.c
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	void *data = NULL;
+	size_t size = 0;
+	if (!wlr_buffer_get_data_ptr(wlr_buffer, &data, &size)) {
+		wlr_log(WLR_ERROR, "Failed to get buffer data");
+		goto error_buffer;
+	}
+
+	uint32_t stride = size / wlr_buffer->height;
```
emersion, line 156: Hm. This sounds a little fragile, but is probably good enough for now. Maybe add an assert to check that the size is divisible by the height.

bl4ckb0ne: Maybe if we have a way to get it directly from the buffer itself, like adding a func to the impl `get_stride` ?

render/pixman/renderer.c
bl4ckb0ne: `wlr_buffer_get_format` ?

render/pixman/renderer.c
emersion: We can't have a generic `get_stride` function, because some buffers don't have a single stride. Multi-planar buffers may have up to 4 different strides.

Maybe we can just add the stride as an out-param of `get_data_ptr`.

render/pixman/renderer.c
emersion: We already have `wlr_buffer.width` and `wlr_buffer.height`. I think we could have `wlr_buffer.format` with a DRM format in it.

Another solution would be to add the DRM format as an out-param of `get_data_ptr`.

bl4ckb0ne: I prefer having `wlr_buffer.format` and keep `get_data_ptr` for only getting the data.

render/pixman/renderer.c
bl4ckb0ne: `bool wlr_buffer_get_stride` and return `false` if its not supported?

emersion: Hm, I'm not sure. I think I'd prefer the assumption "there is only a single stride" to be tied to `get_shm` and `get_data_ptr`.

emersion: BTW, we could just replace the size out-param with stride. There's non need to know the exact size when you know you can access the buffer up to `stride * height`.

bl4ckb0ne: Works for me, the only place where I use the size is to calculate the stride. Shall I do the edit in this branch or in `shm-allocator`?

emersion: As you prefer. I'll backport to my branch if you change it here.

meson.build
```diff
@@ -21,6 +21,18 @@ add_project_arguments([
...
+	add_project_arguments([
+		'-DWLR_LITTLE_ENDIAN=0',
+		'-DWLR_BIG_ENDIAN=1'
+	],language: 'c')
+else
+	add_project_arguments([
+		'-DWLR_LITTLE_ENDIAN=1',
+		'-DWLR_BIG_ENDIAN=0'
+	],language: 'c')
+endif
```
emersion, line 34: Can simplify to something like this:

```meson
le = host_machine.endian() == 'little'
be = host_machine.endian() == 'big'
add_project_arguments([
	'-DWLR_LITTLE_ENDIAN=@0@'.format(le.to_int()),
	'-DWLR_BIG_ENDIAN=@0@'.format(be.to_int()),
], language: 'c')
```

render/pixman/pixel_format.c
```diff
@@ -0,0 +1,57 @@
...
+		.drm_format = DRM_FORMAT_ABGR8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_r8g8b8a8,
+#else
+		.pixman_format = PIXMAN_a8b8g8r8,
+#endif
+	}
+};
+
+const struct wlr_pixman_pixel_format *get_pixman_format_from_drm(
```
emersion, line 40: Maybe this can return directly a `pixman_format_code_t`?

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+}
+
+static const struct wlr_texture_impl texture_impl = {
+	.is_opaque = texture_is_opaque,
+	.write_pixels = texture_write_pixels,
+	.to_dmabuf = texture_to_dmabuf,
+	.destroy = texture_destroy,
+};
+
+static uint32_t get_pixman_format_from_wl(enum wl_shm_format wl_fmt) {
```
emersion, line 104: This needs to be updated to use `get_pixman_format_from_drm`, I think.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+}
+
+static bool texture_is_opaque(struct wlr_texture *wlr_texture) {
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+	wlr_log(WLR_INFO, "texture_is_opaque");
+
+#if WLR_BIG_ENDIAN
+	return texture->format == PIXMAN_b8g8r8x8;
+#else
+	return texture->format == PIXMAN_x8r8g8b8;
```
emersion, line 48: We can probably use our shared pixel format table for this.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+	} else if (wlr_buffer_get_shm(wlr_buffer, &shm)) {
+		drm_format = shm.format;
+	} else {
+		goto error_buffer;
+	}
+
+	uint32_t format;
+	switch (drm_format) {
+	case DRM_FORMAT_XRGB8888:
+		format = PIXMAN_x8r8g8b8;
```
emersion, line 174: This can be replaced with `get_pixman_format_from_drm` as well.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+	struct wlr_shm_attributes shm = {0};
+	if (wlr_buffer_get_dmabuf(wlr_buffer, &dmabuf)) {
+		drm_format = dmabuf.format;
+	} else if (wlr_buffer_get_shm(wlr_buffer, &shm)) {
+		drm_format = shm.format;
+	} else {
+		goto error_buffer;
+	}
+
+	uint32_t format;
```
emersion, line 171: The `pixman_format_code_t` type would be more descriptive.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+
+	pixman_image_composite32(PIXMAN_OP_SRC, fill, NULL, buffer->image, 0, 0, 0,
+			0, 0, 0, renderer->width, renderer->height);
+
+	pixman_image_unref(fill);
+}
+
+static void pixman_scissor(struct wlr_renderer *wlr_renderer,
+		struct wlr_box *box) {
+	wlr_log(WLR_ERROR, "todo pixman_scissor");
```
emersion, line 246: I wonder if `pixman_image_set_clip_region32` does what we want here.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+}
+
+static void pixman_render_quad_with_matrix(struct wlr_renderer *wlr_renderer,
+		const float color[static 4], const float matrix[static 9]) {
+	wlr_log(WLR_ERROR, "todo pixman_render_quad_with_matrix");
+}
+
+static void pixman_render_ellipse_with_matrix(struct wlr_renderer *wlr_renderer,
+		const float color[static 4], const float matrix[static 9]) {
+	wlr_log(WLR_ERROR, "todo pixman_render_ellipse_with_matrix");
```
emersion, line 298: This will be removed soon, no need to bother with implementing this.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+	wlr_log(WLR_INFO, "src_x %zu src_y %zu", (size_t)src_x, (size_t)src_y);
+
+	uint32_t *start = (uint32_t*)data + (src_y * stride) + (src_x * 4);
+	assert(start);
+
+	pixman_image_t *pixels = pixman_image_create_bits_no_clear(texture->format,
+			width, height, start, stride);
+
+	pixman_image_composite32(PIXMAN_OP_OVER, texture->image, NULL, pixels,
+			dst_x, dst_y, 0, 0, 0, 0, width, height);
```
emersion, line 74: I don't think we should do this: `texture->image` points to client memory, we shouldn't write to client buffers.

We can probably just `return false` and the wlr_buffer code will re-create the texture (which isn't particularly expensive since no buffer copying is involved).

emersion: Even better: we can just leave `write_pixels` unimplemented.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+
+static void pixman_render_ellipse_with_matrix(struct wlr_renderer *wlr_renderer,
+		const float color[static 4], const float matrix[static 9]) {
+	wlr_log(WLR_ERROR, "todo pixman_render_ellipse_with_matrix");
+}
+
+static const uint32_t *pixman_get_shm_texture_formats(
+		struct wlr_renderer *wlr_renderer, size_t *len) {
+	wlr_log(WLR_ERROR, "todo pixman_get_shm_texture_formats");
+	return 0;
```
emersion, line 304: Here we should probably list all formats from the table which return true for `pixman_format_supported_source()`.

render/pixman/renderer.c
emersion, line 171: We could sanity-check the format with `pixman_format_supported_destination()` and print a good error message if it returns false.

render/pixman/renderer.c
```diff
@@ -0,0 +1,410 @@
...
+
+	for (size_t i = 0; i < len; ++i) {
+		wlr_drm_format_set_add(&renderer->drm_formats, formats[i],
+				DRM_FORMAT_MOD_LINEAR);
+	}
+
+
+	return &renderer->wlr_renderer;
+}
+
```
emersion, line 410: Style nit: extra newline here

include/render/pixman.h
```diff
@@ -38,10 +39,11 @@ struct wlr_pixman_texture {
 	struct wlr_pixman_renderer *renderer;
 
 	pixman_image_t *image;
-	uint32_t format;
+	const struct wlr_pixel_format_info *format;
+	pixman_format_code_t pixman_format;
```
emersion, line 43: Nit: no need to store the format here, it can be directly accessed with `pixman_image_get_format`

emersion:
The last commit on this branch fixes crashes on resize for me: https://github.com/emersion/wlroots/commits/pixman_render

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+	buffer->buffer = wlr_buffer;
+	buffer->renderer = renderer;
+
+	uint32_t drm_format;
+	struct wlr_dmabuf_attributes dmabuf = {0};
+	struct wlr_shm_attributes shm = {0};
+	if (wlr_buffer_get_dmabuf(wlr_buffer, &dmabuf)) {
```
emersion, line 88: Note to self: calling `get_dmabuf`/`get_shm` just to get the pixel format isn't too great. But it's fine for now, let's fix this later.

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+	void *data = NULL;
+	size_t stride;
+	if (!buffer_get_data_ptr(wlr_buffer, &data, &stride)) {
+		wlr_log(WLR_ERROR, "Failed to get buffer data");
+		goto error_buffer;
+	}
+
+	wlr_log(WLR_INFO, "width %zu, height %zu", (size_t)wlr_buffer->width,
+			(size_t)wlr_buffer->height);
+	wlr_log(WLR_INFO, "data %p, stride %zu", data, stride);
```
emersion, line 107: Can we clarify these messages a little?

- Maybe downgrade to DEBUG, since it's not really important for end-users.
- Maybe explain what's happening. For instance the GLES2 renderer logs something like "Imported 42x42 buffer as FBO"
- The value of the data pointer doesn't seem very useful.

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+
+	buffer->image = pixman_image_create_bits(format, wlr_buffer->width,
+			wlr_buffer->height, data, stride);
+	if (!buffer->image) {
+		wlr_log(WLR_ERROR, "Failed to allocate pixman image\n");
```
emersion, line 112: `wlr_log` already appends a newline, no need to do it manually.

render/pixman/renderer.c
emersion: Ah, there's already a debug message below that reads "Created pixman buffer %dx%d". Maybe it's fine to just remove these two INFOs then?

bl4ckb0ne: Oh woops that's dev output, I think we can remove those.

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+}
+
+static void pixman_scissor(struct wlr_renderer *wlr_renderer,
+		struct wlr_box *box) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	if (box != NULL) {
+		struct pixman_region32 region = {0};
+		pixman_region32_init_rect(&region, box->x, box->y, box->width,
```
emersion, line 165: We need to `pixman_region32_fini` after the `pixman_image_set_clip_region32` call, to avoid leaking memory.

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	struct pixman_color mask_colour = {0};
+	mask_colour.alpha = 0xFFFF * alpha;
+	pixman_image_t *mask = pixman_image_create_solid_fill(&mask_colour);
+
+	struct pixman_transform transform = {0};
+	wlr_matrix_scale((float*)matrix, 1.0 / fbox->width, 1.0 / fbox->height);
```
emersion, line 202: Style nit: missing space between `float` and `*`

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+	matrix_to_pixman_transform(&transform, matrix);
+	pixman_transform_invert(&transform, &transform);
+
+	pixman_image_set_transform(texture->image, &transform);
+
+	pixman_image_composite32(PIXMAN_OP_OVER, texture->image, mask,
```
emersion, line 208: Can't we just leave the `mask` NULL here, since it's not really used?

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+			buffer->image, 0, 0, 0, 0, 0, 0, renderer->width,
```
emersion, line 209: It seems like we're not clipping the source texture properly, because `src_x` and `src_y` are left as zero. We can implement this later, but please add a TODO about clipping.

render/pixman/renderer.c
bl4ckb0ne: I think we're good because the `pixman_region32` is allocated on the stack https://cgit.freedesktop.org/pixman/tree/pixman/pixman-region.c#n413

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+	if (texture == NULL) {
+		wlr_log(WLR_ERROR, "Failed to allocate pixman texture");
+		return NULL;
+	}
+
+	wlr_texture_init(&texture->wlr_texture, &texture_impl, width, height);
+	texture->renderer = renderer;
+
+	texture->format = drm_get_pixel_format_info(drm_format);
+	assert(texture->format);
```
emersion, line 276: Can we turn this assert into a graceful failure? I'd prefer not to die when a format which isn't in our DRM pixel table is passed.

render/pixman/renderer.c
bl4ckb0ne: it's created from the alpha, is it unused by the compositors?

render/pixman/renderer.c
emersion: `FREE_DATA` hides a `free` call: https://cgit.freedesktop.org/pixman/tree/pixman/pixman-region.c#n216

render/pixman/renderer.c
emersion: Oh. My bad, that looks good then! Maybe add a TODO: don't create a mask if alpha == 1.0

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+
+	struct pixman_color colour = {
+		.red = color[0] * 0xFFFF,
+		.green = color[1] * 0xFFFF,
+		.blue = color[2] * 0xFFFF,
+		.alpha = color[3] * 0xFFFF,
+	};
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+	pixman_image_t *image = pixman_image_create_bits (PIXMAN_a8r8g8b8,
```
emersion, line 230: Style: extra space before `(`

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+						matrix[0], matrix[4], NULL, 0);
```
emersion, line 231: Style: don't align, indent continuation lines with 1 or 2 tabs

render/pixman/renderer.c
```diff
@@ -0,0 +1,365 @@
...
+	pixman_image_composite32 (PIXMAN_OP_SRC, fill, NULL, image,
+				  0, 0, 0, 0, 0, 0, matrix[0], matrix[4]);
```
emersion, line 233: Ditto

render/pixman/renderer.c
```diff
@@ -0,0 +1,367 @@
...
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+	struct wlr_pixman_buffer *buffer = renderer->current_buffer;
+
+	// TODO: don't create a mask if alpha == 1.0
+	struct pixman_color mask_colour = {0};
+	mask_colour.alpha = 0xFFFF * alpha;
+	pixman_image_t *mask = pixman_image_create_solid_fill(&mask_colour);
+
+	struct pixman_transform transform = {0};
+	wlr_matrix_scale((float *)matrix, 1.0 / fbox->width, 1.0 / fbox->height);
```
emersion, line 200: This mutates the `matrix` input argument, which is marked as `const`. We can't do this.

We need to copy `matrix` to a local variable before mutating it.

render/pixman/renderer.c
```diff
@@ -0,0 +1,367 @@
...
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+	pixman_image_t *image = pixman_image_create_bits(PIXMAN_a8r8g8b8,
+			matrix[0], matrix[4], NULL, 0);
+	pixman_image_composite32 (PIXMAN_OP_SRC, fill, NULL, image,
+			0, 0, 0, 0, 0, 0, matrix[0], matrix[4]);
+	pixman_image_unref(fill);
+
+	struct pixman_transform transform = {0};
+	wlr_matrix_scale((float*)matrix, 1.0 / matrix[0], 1.0 / matrix[4]);
```
emersion, line 236: Same here.

render/pixman/renderer.c
```diff
@@ -0,0 +1,367 @@
...
+	struct pixman_color colour = {
+		.red = color[0] * 0xFFFF,
+		.green = color[1] * 0xFFFF,
+		.blue = color[2] * 0xFFFF,
+		.alpha = color[3] * 0xFFFF,
+	};
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+	pixman_image_t *image = pixman_image_create_bits(PIXMAN_a8r8g8b8,
+			matrix[0], matrix[4], NULL, 0);
```
emersion, line 230: This code assumes `width = matrix[0]` and `height = matrix[4]`. However this only works if there's no rotation in the matrix, because in fact `matrix[0] = width * cos(rot)` and `matrix[4] = height * cos(rot)`.

render/pixman/renderer.c
```diff
@@ -0,0 +1,367 @@
...
+
+	struct pixman_color colour = {
+		.red = color[0] * 0xFFFF,
+		.green = color[1] * 0xFFFF,
+		.blue = color[2] * 0xFFFF,
+		.alpha = color[3] * 0xFFFF,
+	};
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+	pixman_image_t *image = pixman_image_create_bits(PIXMAN_a8r8g8b8,
```
emersion, line 229: I was able to apply this patch without observing a change in behaviour: https://l.sr.ht/psia.txt

The patch removes `image`, and only deals with `fill`.

render/pixman/renderer.c
emersion: Fixing this sounds complicated, because we don't have an explicit width/height param. We'd need to assume the matrix scale is the quad size, and compute `width = sqrt(matrix[0] * matrix[0] + matrix[1] * matrix[1])` (and something similar for `height`). In the meantime, maybe slap a TODO here too.

emersion: Note: we can detect the no-rotation special case with `matrix[1] == 0`.

render/pixman/renderer.c
```diff
@@ -0,0 +1,379 @@
...
+		.alpha = color[3] * 0xFFFF,
+	};
+
+	pixman_image_t *fill = pixman_image_create_solid_fill(&colour);
+
+	float m[9];
+	memcpy(m, matrix, sizeof(m));
+
+	float width = sqrt(matrix[0] * matrix[0] + matrix[1] * matrix[1]);
+	float height = sqrt(matrix[4] * matrix[4] + matrix[1] * matrix[1]);
```
emersion, line 237: I think this should be `sqrt(matrix[3] * matrix[3] + matrix[4] * matrix[4])`.

bl4ckb0ne: I was thinking the same thing, but wasn't able to make it crash with `matrix[1]`. I'll go `matrix[3]` just in case.

emersion: A mix-up won't make it crash, but will give the wrong result if the matrix contains a transformation.

render/pixman/renderer.c
```diff
@@ -0,0 +1,379 @@
...
+
+	wlr_matrix_scale(m, 1.0 / width, 1.0 / height);
+
+	pixman_image_t *image = pixman_image_create_bits(PIXMAN_a8r8g8b8, width,
```
emersion, line 241: Can we just slap a TODO here to avoid this image allocation?

emersion:
I'm no longer able to reproduce the `render_quad` glitches in Sway. I'm guessing memcpy'ing the matrices fixed it.

With that out of the way, I think we have a fairly complete renderer, even if there's some room for future improvements. :)

render/pixman/renderer.c
bl4ckb0ne: Heh I know, all of our examples have square textures, that's why I didn't get any wrong results.

bl4ckb0ne:
I second the fix on the `render_quad` fix, wasn't able to reproduce.

emersion (approved):
LGTM, thanks a lot!

PR #2669, opened 2021-01-21
backend/x11: Use native cursors

imirkin:
Note that I played around with returning 'false' for non-NULL textures. However this did not have the desired effect. But I don't think it's a shortcoming of this change. The native cursor was correctly removed in that case though.

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -7,12 +7,14 @@
 #include <drm_fourcc.h>
 #include <xcb/dri3.h>
 #include <xcb/present.h>
+#include <xcb/render.h>
 #include <xcb/xcb.h>
 #include <xcb/xinput.h>
 
 #include <wlr/interfaces/wlr_output.h>
 #include <wlr/interfaces/wlr_pointer.h>
 #include <wlr/interfaces/wlr_touch.h>
+#include <wlr/types/wlr_matrix.h>
 #include <wlr/util/log.h>
 
 #include "backend/x11.h"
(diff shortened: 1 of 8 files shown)


backend/x11/output.c
```diff
@@ -331,12 +338,179 @@ static void output_rollback_render(struct wlr_output *wlr_output) {
...
+		width = texture->width * wlr_output->scale / scale;
+		height = texture->height * wlr_output->scale / scale;
+	}
+
+	struct wlr_box hotspot = { .x = hotspot_x, .y = hotspot_y };
+	wlr_box_transform(&hotspot, &hotspot,
+		wlr_output_transform_invert(wlr_output->transform),
+		width, height);
+
+	// TODO: use output->wlr_output.transform to transform pixels and hotpot
```
emersion, line 480: This seems implemented.

backend/x11/output.c
```diff
@@ -331,12 +338,179 @@ static void output_rollback_render(struct wlr_output *wlr_output) {
 	wlr_renderer_bind_buffer(x11->renderer, NULL);
 }
 
+static void update_x11_output_cursor(struct wlr_x11_output *output,
+		int32_t hotspot_x, int32_t hotspot_y) {
+	struct wlr_x11_backend *x11 = output->x11;
+
+	xcb_cursor_t cursor = x11->cursor; /* transparent cursor */
```
emersion, line 345: Maybe we can rename `x11->cursor` to something like `x11->transparent_cursor` to avoid confusion?

backend/x11/output.c
imirkin: OK. That was just some copy/paste from the wayland one.

backend/x11/output.c
imirkin: Sure thing!

emersion (approved):
LGTM, thanks!

PR #2671, opened 2021-01-21
render/egl: initialize wlr_egl with EGL_PLATFORM_DEVICE_EXT

### LINKED ISSUE #2649: render/egl: use EGLDevice instead of the GBM platform

bl4ckb0ne:
This function relies on the extension EGL_WL_bind_wayland_display being present. If not, the creation of `wlr_egl` will fallback on using GBM.

I see in the logs that the client extensions are being printed twice, would you prefer the fallback to be in `wlr_egl_create_from_drm_fd` rather than in `wlr_renderer_autocreate_with_drm_fd`?

I also I fixed an issue regarding `eglMakeCurrent` when processing the error, `EGL_NO_DISPLAY` is not a valid display handle and will cause an `EGL_BAD_DISPLAY` error.

> As with other commands taking EGLDisplay parameters, if display is not a valid EGLDisplay handle, an EGL_BAD_DISPLAY error is generated. (Some implementations have chosen to allow EGL_NO_DISPLAY as a valid display parameter for eglMakeCurrent. This behavior is not portable to all EGL implementations, and should be considered as an undocumented vendor extension).

https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglMakeCurrent.xhtml

Fixes #2649

### DIFF
--- a/include/wlr/render/egl.h
+++ b/include/wlr/render/egl.h
@@ -44,6 +44,11 @@ struct wlr_egl {
 		// Device extensions
 		bool EXT_device_drm;
 		bool EXT_device_drm_render_node;
+
+		// Client extensions
+		bool EXT_device_query;
+		bool KHR_platform_gbm;
+		bool EXT_platform_device;
 	} exts;
 
 	struct {
@@ -56,6 +61,7 @@ struct wlr_egl {
 		PFNEGLDEBUGMESSAGECONTROLKHRPROC eglDebugMessageControlKHR;
 		PFNEGLQUERYDISPLAYATTRIBEXTPROC eglQueryDisplayAttribEXT;
 		PFNEGLQUERYDEVICESTRINGEXTPROC eglQueryDeviceStringEXT;
+		PFNEGLQUERYDEVICESEXTPROC eglQueryDevicesEXT;
 	} procs;
 
 	struct wlr_drm_format_set dmabuf_texture_formats;
--- a/render/egl.c
+++ b/render/egl.c
@@ -156,7 +156,7 @@ out:
 	free(formats);
 }
 
-struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
+static struct wlr_egl *egl_create(void) {
 	struct wlr_egl *egl = calloc(1, sizeof(struct wlr_egl));
 	if (egl == NULL) {
 		wlr_log_errno(WLR_ERROR, "Allocation failed");
@@ -172,11 +172,7 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 		}
 		return NULL;
 	}
-
-	if (!check_egl_ext(client_exts_str, "EGL_KHR_platform_gbm")) {
-		wlr_log(WLR_ERROR, "EGL_KHR_platform_gbm not supported");
-		return NULL;
-	}
+	wlr_log(WLR_INFO, "Supported EGL client extensions: %s", client_exts_str);
 
 	if (!check_egl_ext(client_exts_str, "EGL_EXT_platform_base")) {
 		wlr_log(WLR_ERROR, "EGL_EXT_platform_base not supported");
@@ -184,6 +180,24 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 	}
 	load_egl_proc(&egl->procs.eglGetPlatformDisplayEXT,
 		"eglGetPlatformDisplayEXT");
+
+	egl->exts.KHR_platform_gbm = check_egl_ext(client_exts_str,
+			"EGL_KHR_platform_gbm");
+
+	egl->exts.EXT_platform_device = check_egl_ext(client_exts_str,
+			"EGL_EXT_platform_device");
+
+	if (check_egl_ext(client_exts_str, "EGL_EXT_device_enumeration")) {
+		load_egl_proc(&egl->procs.eglQueryDevicesEXT, "eglQueryDevicesEXT");
+	}
+
+	if (check_egl_ext(client_exts_str, "EGL_EXT_device_query")) {
+		egl->exts.EXT_device_query = true;
+		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
+			"eglQueryDeviceStringEXT");
+		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
+			"eglQueryDisplayAttribEXT");
+	}
 
 	if (check_egl_ext(client_exts_str, "EGL_KHR_debug")) {
 		load_egl_proc(&egl->procs.eglDebugMessageControlKHR,
@@ -201,32 +215,31 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 
 	if (eglBindAPI(EGL_OPENGL_ES_API) == EGL_FALSE) {
 		wlr_log(WLR_ERROR, "Failed to bind to the OpenGL ES API");
-		goto error;
+		return NULL;
 	}
 
-	egl->gbm_device = gbm_create_device(drm_fd);
-	if (!egl->gbm_device) {
-		wlr_log(WLR_ERROR, "Failed to create GBM device");
-		goto error;
-	}
+	return egl;
+}
 
-	egl->display = egl->procs.eglGetPlatformDisplayEXT(EGL_PLATFORM_GBM_KHR,
-		egl->gbm_device, NULL);
+static bool egl_init(struct wlr_egl *egl, EGLenum platform,
+		void *remote_display) {
+	egl->display = egl->procs.eglGetPlatformDisplayEXT(platform,
+		remote_display, NULL);
 	if (egl->display == EGL_NO_DISPLAY) {
 		wlr_log(WLR_ERROR, "Failed to create EGL display");
-		goto error;
+		return false;
 	}
 
 	EGLint major, minor;
 	if (eglInitialize(egl->display, &major, &minor) == EGL_FALSE) {
 		wlr_log(WLR_ERROR, "Failed to initialize EGL");
-		goto error;
+		return false;
 	}
 
 	const char *display_exts_str = eglQueryString(egl->display, EGL_EXTENSIONS);
 	if (display_exts_str == NULL) {
 		wlr_log(WLR_ERROR, "Failed to query EGL display extensions");
-		return NULL;
+		return false;
 	}
 
 	if (check_egl_ext(display_exts_str, "EGL_KHR_image_base")) {
@@ -247,17 +260,12 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 	}
 
 	const char *device_exts_str = NULL, *driver_name = NULL;
-	if (check_egl_ext(client_exts_str, "EGL_EXT_device_query")) {
-		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
-			"eglQueryDisplayAttribEXT");
-		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
-			"eglQueryDeviceStringEXT");
-
+	if (egl->exts.EXT_device_query) {
 		EGLAttrib device_attrib;
 		if (!egl->procs.eglQueryDisplayAttribEXT(egl->display,
 				EGL_DEVICE_EXT, &device_attrib)) {
 			wlr_log(WLR_ERROR, "eglQueryDisplayAttribEXT(EGL_DEVICE_EXT) failed");
-			goto error;
+			return false;
 		}
 		egl->device = (EGLDeviceEXT)device_attrib;
 
@@ -265,7 +273,7 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 			egl->procs.eglQueryDeviceStringEXT(egl->device, EGL_EXTENSIONS);
 		if (device_exts_str == NULL) {
 			wlr_log(WLR_ERROR, "eglQueryDeviceStringEXT(EGL_EXTENSIONS) failed");
-			goto error;
+			return false;
 		}
 
 		if (check_egl_ext(device_exts_str, "EGL_MESA_device_software")) {
@@ -276,7 +284,7 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 				wlr_log(WLR_ERROR, "Software rendering detected, please use "
 					"the WLR_RENDERER_ALLOW_SOFTWARE environment variable "
 					"to proceed");
-				goto error;
+				return false;
 			}
 		}
 
@@ -297,21 +305,19 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 			!check_egl_ext(display_exts_str, "EGL_MESA_configless_context")) {
 		wlr_log(WLR_ERROR, "EGL_KHR_no_config_context or "
 			"EGL_MESA_configless_context not supported");
-		goto error;
+		return false;
 	}
 
 	if (!check_egl_ext(display_exts_str, "EGL_KHR_surfaceless_context")) {
-		wlr_log(WLR_ERROR,
-			"EGL_KHR_surfaceless_context not supported");
-		goto error;
+		wlr_log(WLR_ERROR, "EGL_KHR_surfaceless_context not supported");
+		return false;
 	}
 
-	wlr_log(WLR_INFO, "Using EGL %d.%d", (int)major, (int)minor);
-	wlr_log(WLR_INFO, "Supported EGL client extensions: %s", client_exts_str);
 	wlr_log(WLR_INFO, "Supported EGL display extensions: %s", display_exts_str);
 	if (device_exts_str != NULL) {
 		wlr_log(WLR_INFO, "Supported EGL device extensions: %s", device_exts_str);
 	}
+	wlr_log(WLR_INFO, "Using EGL %d.%d", (int)major, (int)minor);
 	wlr_log(WLR_INFO, "EGL vendor: %s", eglQueryString(egl->display, EGL_VENDOR));
 	if (driver_name != NULL) {
 		wlr_log(WLR_INFO, "EGL driver name: %s", driver_name);
@@ -345,7 +351,7 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 		EGL_NO_CONTEXT, attribs);
 	if (egl->context == EGL_NO_CONTEXT) {
 		wlr_log(WLR_ERROR, "Failed to create EGL context");
-		goto error;
+		return false;
 	}
 
 	if (request_high_priority) {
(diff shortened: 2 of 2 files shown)


render/wlr_renderer.c
```diff
@@ -251,23 +251,28 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 }
 
 struct wlr_renderer *wlr_renderer_autocreate_with_drm_fd(int drm_fd) {
+	struct wlr_renderer *renderer = NULL;
+	struct wlr_egl *egl = wlr_egl_create_from_drm_fd(drm_fd);
+	if (egl != NULL) {
+		goto create_renderer;
+	}
+
+	// Fallback to GBM
```
emersion, line 260: I'd prefer the fallback to be in `wlr_egl_create_from_drm_fd` if possible. This would allow to re-use the `wlr_egl` returned by `egl_create` and avoid the double initialization and double log messages.

render/egl.c
```diff
@@ -248,28 +265,28 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			"eglUnbindWaylandDisplayWL");
 		load_egl_proc(&egl->procs.eglQueryWaylandBufferWL,
 			"eglQueryWaylandBufferWL");
+	} else {
+		if (platform == EGL_PLATFORM_DEVICE_EXT) {
+			wlr_log(WLR_ERROR, "EGL_WL_bind_wayland_display not supported");
```
emersion, line 270: It would probably be more appropriate to not log an error when this happens. Maybe an INFO which says something like "EGL_WL_bind_wayland_display not supported on device platform, falling back to GBM platform"?

emersion (changes requested):

emersion:
Note, the Mesa MR that makes the device platform work is: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/8638

render/egl.c
bl4ckb0ne: Do you prefer INFO or DEBUG?

render/wlr_renderer.c
bl4ckb0ne: Good idea, fixed in v2

render/egl.c
```diff
@@ -347,13 +364,124 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
...
+	}
+
+	// Fallback to GBM
+	struct gbm_device *gbm_device = gbm_create_device(drm_fd);
+	if (!gbm_device) {
+		wlr_log(WLR_ERROR, "Failed to create GBM device");
+		goto error;
+	}
+
+	if (egl_init(egl, EGL_PLATFORM_GBM_KHR, gbm_device, NULL)) {
```
emersion, line 472: We need to set `egl->gbm_device` here so that we don't leak it.

render/egl.c
emersion: As you prefer, I don't feel strongly either way. FWIW, INFO are user-visible messages, DEBUG are only visible to developers.

bl4ckb0ne: Can go for INFO then

render/egl.c
```diff
@@ -248,28 +265,28 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			"eglUnbindWaylandDisplayWL");
 		load_egl_proc(&egl->procs.eglQueryWaylandBufferWL,
 			"eglQueryWaylandBufferWL");
+	} else {
+		if (platform == EGL_PLATFORM_DEVICE_EXT) {
+			wlr_log(WLR_INFO, "EGL_WL_bind_wayland_display not supported");
```
emersion, line 270: Please indicate that this is about the device platform and that we'll fallback to the GBM platform.

bl4ckb0ne: Debug or info level

bl4ckb0ne: I'll add the message in `wlr_egl_create_from_drm_fd`

emersion:
Unfortunately this PR breaks split display/render devices. This happens because Mesa doesn't create an `EGLDevice` for the display-only device, even if kmsro would make it work. See https://gitlab.freedesktop.org/mesa/mesa/-/issues/4178.

bl4ckb0ne:
Want me to take a look at that issue?

emersion:
It's a tricky issue, I'm not sure what's the best way to fix it. I'd like to hear ideas from Mesa devs.

bl4ckb0ne:
I'll keep an eye on the issue then, let me know if I can do anything to help

render/egl.c
```diff
@@ -317,7 +333,8 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 
 	// On DRM, request a high priority context if possible
 	bool request_high_priority = ext_context_priority &&
-		platform == EGL_PLATFORM_GBM_MESA;
+		(platform == EGL_PLATFORM_GBM_MESA
+		|| platform == EGL_PLATFORM_DEVICE_EXT);
```
emersion, line 337: https://github.com/swaywm/wlroots/pull/2817

render/egl.c
```diff
@@ -248,28 +265,28 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			"eglUnbindWaylandDisplayWL");
 		load_egl_proc(&egl->procs.eglQueryWaylandBufferWL,
 			"eglQueryWaylandBufferWL");
+	} else {
+		if (platform == EGL_PLATFORM_DEVICE_EXT) {
```
emersion, line 269: This can be an `else if` block.

Would be nice to explain why we `return false` here: we expect the caller to fallback to the GBM platform.

include/wlr/render/egl.h
```diff
@@ -87,6 +91,15 @@ struct wlr_egl {
...
 	const EGLint *config_attribs);
 
+/**
+ * Creates and EGL context from a given drm fd. This function uses the
+ * following extensions:
+ * - EXT_device_enumeration to get the list of available EGLDeviceEXT
+ * - EXT_device_drm to get the name of the EGLDeviceEXT
+ * - EXT_platform_device to create an EGLDisplay from an EGLDeviceEXT
+ */
+struct wlr_egl *wlr_egl_create_from_drm_fd(int drm_fd);
```
emersion, line 101: Can we keep this private for now?

emersion:
> Unfortunately this PR breaks split display/render devices.

If we keep the GBM fallback, it should work fine. Unfortunately this means we can't get rid of GBM in the renderer yet.

include/wlr/render/egl.h
bl4ckb0ne: I moved it to a private header

bl4ckb0ne:
Alpine failure because of some python trouble

render/egl.c
```diff
@@ -355,13 +370,123 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display) {
...
+	drmDevice *device = NULL;
+	int ret = drmGetDevice(drm_fd, &device);
+	if (ret < 0) {
+		wlr_log(WLR_ERROR, "Failed to get DRM device: %s", strerror(-ret));
+		goto error;
+	}
 
+	EGLDeviceEXT egl_device = NULL;
+	for (int i = 0; i < nb_devices; i++) {
+		const char *egl_device_name = egl->procs.eglQueryDeviceStringEXT(
```
emersion, line 448: We should check for `EGL_EXT_device_drm` in the device extension string before doing this.

bl4ckb0ne:
The fix is finally in Mesa [21.1.0][1], I'll rebase the patch.

[1]: https://docs.mesa3d.org/relnotes/21.1.0.html

bl4ckb0ne:
@emersion rebased and ready to be reviewed :D

render/egl.c
```diff
@@ -185,6 +186,23 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+		egl->exts.device_query_ext = true;
+		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
+			"eglQueryDeviceStringEXT");
+		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
+			"eglQueryDisplayAttribEXT");
+	}
+
+	if (check_egl_ext(client_exts_str, "EGL_EXT_device_base")) {
+		egl->exts.device_base_ext = true;
+
```
emersion, line 203: Style nit: stray newline

render/egl.c
```diff
@@ -246,26 +263,26 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 			"eglQueryDmaBufModifiersEXT");
 	}
 
-	const char *device_exts_str = NULL, *driver_name = NULL;
-	if (check_egl_ext(client_exts_str, "EGL_EXT_device_query")) {
-		load_egl_proc(&egl->procs.eglQueryDisplayAttribEXT,
-			"eglQueryDisplayAttribEXT");
-		load_egl_proc(&egl->procs.eglQueryDeviceStringEXT,
-			"eglQueryDeviceStringEXT");
+	if (platform == EGL_PLATFORM_DEVICE_EXT && !egl->exts.device_base_ext) {
```
emersion, line 266: If `EGL_PLATFORM_DEVICE_EXT` is not supported, `eglGetPlatformDisplayEXT` should have failed already. We should move this check up to before the `eglGetPlatformDisplayEXT` call.

render/egl.c
```diff
@@ -172,6 +172,7 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 		}
 		return NULL;
 	}
+	wlr_log(WLR_INFO, "Supported EGL client extensions: %s", client_exts_str);
 
 	if (!check_egl_ext(client_exts_str, "EGL_KHR_platform_gbm")) {
```
emersion, line 177: If we use `EGL_PLATFORM_DEVICE_EXT`, then we don't need the GBM platform. We should move this check to `egl_init` and guard it behind a `platform` check.

render/wlr_renderer.c
```diff
@@ -10,11 +10,11 @@
 #include <wlr/config.h>
 
 #if WLR_HAS_GLES2_RENDERER
-#include <wlr/render/egl.h>
```
emersion, line 13: Changes to this file seem unnecessary.

render/egl.c
emersion: This is the wrong ext to check for `EGL_PLATFORM_DEVICE_EXT` support. The ext that provides the device platform is `EGL_EXT_platform_device`.

render/egl.c
bl4ckb0ne: I added a bool in `wlr_egl` since `egl_init` doesn't have access to the client exts string. I also move the platform check on top of the function to not initialize too many EGL stuff for nothing.

render/wlr_renderer.c
bl4ckb0ne: Seems like this came from a rebase, I dropped the changes.

render/egl.c
bl4ckb0ne: Made the check against the right ext and moved the platform check at the beginning of `egl_init`

render/egl.c
```diff
@@ -357,13 +362,93 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+		wlr_log(WLR_ERROR, "Failed to find EGLDeviceEXT");
+		goto error;
+	}
+
+	if (egl_init(egl, EGL_PLATFORM_DEVICE_EXT, egl_device)) {
+		wlr_log(WLR_DEBUG, "Using EGL_PLATFORM_DEVICE_EXT");
+		return egl;
+	}
+
+	// Fallback to GBM
```
emersion, line 431: The fallback no longer sets `egl->gbm_device`, which means we leak it when destroying the `wlr_egl`.

render/wlr_renderer.c
emersion: The changes seem to still be there

render/egl.c
```diff
@@ -359,13 +364,93 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
+	struct wlr_egl *egl = egl_create();
+	if (egl == NULL) {
+		wlr_log(WLR_ERROR, "Failed to create EGL context");
+		return NULL;
+	}
+
+	if (egl->procs.eglQueryDevicesEXT == NULL) {
+		wlr_log(WLR_ERROR, "EGL_EXT_device_enumeration not supported");
+		goto error;
```
emersion, line 381: Here we should fallback to GBM. And downgrade that log message.

render/egl.c
```diff
@@ -359,13 +364,93 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+		}
+
+		if (device_has_name(device, egl_device_name)) {
+			wlr_log(WLR_DEBUG, "Using EGL device %s", egl_device_name);
+			egl_device = devices[i];
+			break;
+		}
+	}
+
+	if (egl_device == NULL) {
```
emersion, line 423: Should use `EGL_NO_DEVICE` here

render/wlr_renderer.c
bl4ckb0ne: Should be gone for real now :) I think I got lost in rebase.

render/egl.c
```diff
@@ -359,13 +364,93 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+		wlr_log(WLR_ERROR, "Failed to create EGL context");
+		return NULL;
+	}
+
+	if (egl->procs.eglQueryDevicesEXT == NULL) {
+		wlr_log(WLR_ERROR, "EGL_EXT_device_enumeration not supported");
+		goto error;
+	}
+
+	EGLint nb_devices = 0;
```
emersion, line 384: Maybe we can move all of this `EGLDevice` logic into a function among these lines?

```c
static EGLDevice get_egl_device_from_drm_fd(struct wlr_egl *egl, int drm_fd)
```

render/egl.c
```diff
@@ -359,13 +364,93 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+		if (device_has_name(device, egl_device_name)) {
+			wlr_log(WLR_DEBUG, "Using EGL device %s", egl_device_name);
+			egl_device = devices[i];
+			break;
+		}
+	}
+
+	if (egl_device == NULL) {
+		wlr_log(WLR_ERROR, "Failed to find EGLDeviceEXT");
+		goto error;
```
emersion, line 425: If we didn't find a matching `EGLDevice`, maybe we're on a split render/display SoC. We should try falling back to GBM.

render/egl.c
emersion: This comment still needs to be addressed. If an EGL impl doesn't have `EGL_KHR_platform_gbm` but does have `EGL_EXT_platform_device`, we should be able to initialize EGL with an `EGLDevice`.

render/egl.c
```diff
@@ -359,13 +364,93 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+	}
+
+	if (egl_init(egl, EGL_PLATFORM_DEVICE_EXT, egl_device)) {
```
emersion, line 428: It would be nice to only try GBM if (1) one of the required extensions (`EGL_EXT_platform_device` or `EGL_EXT_device_enumeration`) isn't supported or (2) we couldn't find an `EGLDevice`.

This line uses the fallback even if the failure was unrelated.

render/egl.c
```diff
@@ -359,13 +364,96 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+		return EGL_NO_DEVICE;
+	}
+
+	EGLint nb_devices = 0;
+	if (!egl->procs.eglQueryDevicesEXT(0, NULL, &nb_devices)) {
+		wlr_log(WLR_ERROR, "Failed to query EGL devices");
+		return EGL_NO_DEVICE;
+	}
+
+	EGLDeviceEXT *devices = calloc(nb_devices, sizeof(EGLDeviceEXT));
```
emersion, line 385: We need to free this before returning.

render/egl.c
```diff
@@ -359,13 +369,97 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+	struct wlr_egl *egl = egl_create();
+	if (egl == NULL) {
+		wlr_log(WLR_ERROR, "Failed to create EGL context");
+		return NULL;
+	}
+
+	// Using eglQueryDevicesEXT
+	EGLDeviceEXT egl_device = get_egl_device_from_drm_fd(egl, drm_fd);
+	if (egl_device != EGL_NO_DEVICE_EXT
+			&& egl_init(egl, EGL_PLATFORM_DEVICE_EXT, egl_device)) {
```
emersion, line 438: Can we hard fail here if `egl_init` fails when `EXT_platform_device` is supported? That probably means something else is wrong, and GBM won't help. This would avoid some confusing double-errors in the logs.

render/egl.c
```diff
@@ -201,32 +213,37 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
-	if (!egl->gbm_device) {
-		wlr_log(WLR_ERROR, "Failed to create GBM device");
-		goto error;
+	return egl;
+}
+
+static bool egl_init(struct wlr_egl *egl, EGLenum platform,
+		void *remote_display) {
+
+	if (platform == EGL_PLATFORM_GBM_KHR && !egl->exts.KHR_platform_gbm) {
```
emersion, line 225: We should do a similar check for `EGL_PLATFORM_DEVICE_EXT`.

render/egl.c
```diff
@@ -201,32 +213,37 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
-	egl->gbm_device = gbm_create_device(drm_fd);
-	if (!egl->gbm_device) {
-		wlr_log(WLR_ERROR, "Failed to create GBM device");
-		goto error;
+	return egl;
+}
+
+static bool egl_init(struct wlr_egl *egl, EGLenum platform,
+		void *remote_display) {
+
```
emersion, line 224: Style nit: stray newline

render/egl.c
```diff
@@ -185,6 +181,22 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 	load_egl_proc(&egl->procs.eglGetPlatformDisplayEXT,
 		"eglGetPlatformDisplayEXT");
 
+	if (check_egl_ext(client_exts_str, "EGL_KHR_platform_gbm")) {
```
emersion, line 184: Can be simplified to `egl->exts.KHR_platform_gbm = check_egl_ext(…)`

render/egl.c
```diff
@@ -359,13 +369,97 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
...
+	if (egl_device != EGL_NO_DEVICE_EXT
+			&& egl_init(egl, EGL_PLATFORM_DEVICE_EXT, egl_device)) {
+		wlr_log(WLR_DEBUG, "Using EGL_PLATFORM_DEVICE_EXT");
+		return egl;
+	}
+
+	egl->gbm_device = gbm_create_device(drm_fd);
+	if (!egl->gbm_device) {
+		wlr_log(WLR_ERROR, "Failed to create GBM device");
+		goto error;
```
emersion, line 446: This will segfault in `gbm_device_destroy`

render/egl.c
bl4ckb0ne: Added check for `EGL_EXT_platform_device` as well

render/egl.c
bl4ckb0ne: I changed the logic to check for `EGL_PLATFORM_DEVICE_EXT` before getting the device from eglQueryDevicesEXT, and a check to `EGL_PLATFORM_GBM_KHR` before creating the GBM device.

emersion (approved):
Thanks for your patience. This version LGTM!

PR #2673, opened 2021-01-21
backend/x11: add support for scanout mode

imirkin:
This makes full-screen with weston-terminal work.

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -218,12 +218,24 @@ static struct wlr_x11_buffer *get_or_create_x11_buffer(
 static bool output_commit_buffer(struct wlr_x11_output *output) {
 	struct wlr_x11_backend *x11 = output->x11;
 
-	assert(output->back_buffer != NULL);
+	struct wlr_buffer *buffer = NULL;
+	switch (output->wlr_output.pending.buffer_type) {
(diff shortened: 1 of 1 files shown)


backend/x11/output.c
```diff
@@ -218,12 +218,24 @@ static struct wlr_x11_buffer *get_or_create_x11_buffer(
...
+	case WLR_OUTPUT_STATE_BUFFER_SCANOUT:
+		buffer = wlr_buffer_lock(output->wlr_output.pending.buffer);
+		break;
+	}
+	assert(buffer != NULL);
 
 	struct wlr_x11_buffer *x11_buffer =
-		get_or_create_x11_buffer(output, output->back_buffer);
+		get_or_create_x11_buffer(output, buffer);
 	if (!x11_buffer) {
```
emersion, line 239: I think we'll want to unlock the buffer here, so that we don't keep a ref on a buffer we aren't using. This can happen if the buffer depth doesn't match the window's for instance.

emersion: I mean, unlock the buffer if `get_or_create_x11_buffer` fails.

imirkin: This was already handled in the goto error case below ... except I forgot to update it. Oops! Done now.

emersion (approved):
Thanks!

PR #2674, opened 2021-01-21
backend/x11: keep track of exposed rects, add them to damage regions 

### LINKED ISSUE #2670: backend/x11: unmap/remap (?) does not cause redraw
When switching workspaces in WindowMaker back to a workspace which has a wlroots-made x11 output running, the output contains random data. Same thing happens when "rolling up" the window, as well as miniaturizing. However regular damage (e.g. dragging another window over it) does not cause problems.

I'm guessing that we need to listen to Map events and make sure to redisplay things in that case?

imirkin:
Fixes #2670

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -75,6 +75,8 @@ static void output_destroy(struct wlr_output *wlr_output) {
 	struct wlr_x11_output *output = get_x11_output_from_output(wlr_output);
 	struct wlr_x11_backend *x11 = output->x11;
 
+	pixman_region32_fini(&output->exposed);
+
 	wlr_input_device_destroy(&output->pointer_dev);
 	wlr_input_device_destroy(&output->touch_dev);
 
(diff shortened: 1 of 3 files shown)


backend/x11/output.c
```diff
@@ -229,11 +231,17 @@ static bool output_commit_buffer(struct wlr_x11_output *output) {
...
+	pixman_region32_t damage;
+	pixman_region32_init(&damage);
+	pixman_region32_copy(&damage, &output->exposed);
+	pixman_region32_clear(&output->exposed);
 	if (output->wlr_output.pending.committed & WLR_OUTPUT_STATE_DAMAGE) {
-		pixman_region32_t *damage = &output->wlr_output.pending.damage;
+		pixman_region32_union(&damage, &damage, &output->wlr_output.pending.damage);
+	}
 
+	if (pixman_region32_not_empty(&damage)) {
```
emersion, line 242: The compositor may not support damage tracking. In that case, the `WLR_OUTPUT_STATE_DAMAGE` flag won't be set and we want to damage the whole window (`region = XCB_NONE`).

If we got an Expose event (so the `expose` region isn't empty) but the compositor doesn't support damage tracking, we still want to damage the whole window. So we should probably merge this block into the previous `if`.

imirkin: Ah OK. I wasn't quite aware of the various permutations. So then we should only pay attention to the damage from events if "committed & WLR_OUTPUT_STATE_DAMAGE" is set, and otherwise not worry about it, since it'll be a full redraw anyways?

emersion: Yeah, that sounds correct.

emersion (approved):
LGTM, thanks!

Issue #1865, opened 2019-10-21
wl_subsurface_place_above ignores parent


christianrauch:
Using `wl_subsurface_place_above` fails with `wl_subsurface@92: error 0: place_above: wl_surface@9 is not a parent or sibling` on a parent surface that has previously been used in `wl_subcompositor_get_subsurface`.
This works on GNOME shell and `weston`. Therefore, I assume for the moment that this is a sway compositor issue. This might as well be a client issue, but sway should at least behave like the reference compositor weston and show the same window surfaces.

sway 1.2 debug log: [sway.log](https://github.com/swaywm/sway/files/3748601/sway.log)

steps:
1. decompress the example [libdecoration_subsurf_above_bug.tar.xz.txt](https://github.com/swaywm/sway/files/3748582/libdecoration_subsurf_above_bug.tar.xz.txt)
2. build via `meson build && ninja -C build`
3. run demo via `demo.sh`

wayland log on sway 1.2: [wl_sway.txt](https://github.com/swaywm/sway/files/3748602/wl_sway.txt)

wayland log on GNOME shell 3.28: [wl_gnome.log](https://github.com/swaywm/sway/files/3748607/wl_gnome.log)

ascent12:
Moved to wlroots, because it would be an issue with wlroots's subsurface code rather than sway's.

Yeah, I have a suspicion our implementation of subsurfaces is not 100% conformant.
Subsurfaces are still a feature not used by a lot of clients, but I have my own test client which fails with a protocol error on wlroots and not weston.

emersion:
Weston achieves this with a dummy subsurface for the parent in the subsurface list (see `weston_subsurface_create_for_parent`). IIRC we had this at some point but then removed it for some reason.

berylline:
Hi there @christianrauch,

Do you mind trying out my PR(#2676) to see if the issue has been fixed for you, when you have the time?

Thanks!

christianrauch:
I am not using sway on a regular basis. I only test libdecoration against released versions from time-to-time.

If the example above works in your setup, I am happy to close this issue after the PR has been merged.

rmader:
I'm seeing a similar issue with a new experimental Firefox backend (https://bugzilla.mozilla.org/show_bug.cgi?id=1695500). Will try the MR in the coming days, in any case it would be great to see this fixed :)

emersion:
[...]

PR #2676, opened 2021-01-22
[DON'T MERGE YET] subsurface: fix incorrect behaviour

berylline:
Correct `subsurface_handle_place_above`, `subsurface_handle_place_below` and some other subsurface-related functions to have it match Weston's equivalent code for correctness, sanity and implementation parity.

See the commits for more info.

Code refs:

For `subsurface_handle_place_above`:

https://github.com/swaywm/wlroots/blob/master/types/wlr_surface.c#L787
https://github.com/wayland-project/weston/blob/master/libweston/compositor.c#L4423

For `subsurface_handle_place_below`:

https://github.com/swaywm/wlroots/blob/master/types/wlr_surface.c#L814
https://github.com/wayland-project/weston/blob/master/libweston/compositor.c#L4447

And for the missing `assert`:

https://github.com/wayland-project/weston/blob/master/libweston/compositor.c#L4417

Issue refs:

#1865, varmd/wine-wayland#23

@ascent12, @emersion:

Do the subsurface test clients work with this PR?

EDIT:

Just in case it wasn't clear, `subsurface_handle_place_above` is currently doing what `subsurface_handle_place_below` should be doing and `subsurface_handle_place_below` was doing vice versa and both of the handler functions lacked an `assert` for making sure that the parent of the subsurface matched the parent of the sibling that gets found, if I understood the code correctly.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -804,8 +804,10 @@ static void subsurface_handle_place_above(struct wl_client *client,
 		return;
 	}
 
+	assert(sibling->parent == subsurface->parent);
+
 	wl_list_remove(&subsurface->parent_pending_link);
-	wl_list_insert(&sibling->parent_pending_link,
+	wl_list_insert(sibling->parent_pending_link.prev,
 		&subsurface->parent_pending_link);
 
 	subsurface->reordered = true;
@@ -831,8 +833,10 @@ static void subsurface_handle_place_below(struct wl_client *client,
 		return;
 	}
 
+	assert(sibling->parent == subsurface->parent);
+
 	wl_list_remove(&subsurface->parent_pending_link);
-	wl_list_insert(sibling->parent_pending_link.prev,
+	wl_list_insert(&sibling->parent_pending_link,
 		&subsurface->parent_pending_link);
 
 	subsurface->reordered = true;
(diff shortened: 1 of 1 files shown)


emersion:
Test case: `subsurfaces` in https://github.com/emersion/wleird, probably.

types/wlr_surface.c
```diff
@@ -804,8 +804,10 @@ static void subsurface_handle_place_above(struct wl_client *client,
 		return;
 	}
 
+	assert(sibling->parent == subsurface->parent);
```
emersion, line 807: I don't think these asserts are necessary, because `subsurface_find_sibling` already searches in the parent. Or am I missing something?

berylline: The equivalent sibling searching code that's in wlroots isn't in its own function and where I had put the `assert` line was the closest place that I could put it in to make both of the handler functions(`subsurface_handle_place_above` and `subsurface_handle_place_below`) be on par with what Weston is doing. Sorry if I didn't make that clear at first.

emersion:
I've tested this PR with wleird's example and it gave the wrong result. wlroots master had the correct result.

types/wlr_surface.c
rmader: Note: the sibling can also be the surfaces parent. I.e. `sibling->parent == subsurface->parent || sibling == subsurface->parent`

types/wlr_surface.c
```diff
@@ -831,8 +833,10 @@ static void subsurface_handle_place_below(struct wl_client *client,
...
```
rmader, line 836: Same here, sibling can be the parent.

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2948

PR #2666, opened 2021-01-20
render/egl: remove *config_attribs, wlr_egl->config and wlr_egl_(create, destroy)_surface

berylline:
Just another part of #2624 and #1352 that I felt needed to be done.

Breaking changes:

`EGLint *config_attribs`, `wlr_egl->config`, `wlr_egl_create_surface` and `wlr_egl_destroy_surface` have been removed and no longer exist.

CC(if needed):

@emersion @bl4ckb0ne @ascent12 

EDIT: Fixed a typo.

EDIT 2: Removed `wlr_egl_create_surface` and `wlr_egl_destroy_surface` and removed a dependency on `wlr_egl` from some example clients.

Issue #2624, opened 2021-01-09
Cleanup wlr_egl


emersion:
There are some EGL features that are no longer used by backends:

- [x] `wlr_egl_swap_buffers` (https://github.com/swaywm/wlroots/pull/2623)
- [x] `wlr_egl_create_surface` and `wlr_egl_destroy_surface` (still used by client examples)
- [x] Buffer age support in `wlr_egl_make_current` (https://github.com/swaywm/wlroots/pull/2623)

emersion:
I think we should remove any `wlr_egl` usage from the examples. I'm not yet clear what's the best way to do this, since EGL initialization is quite verbose. Maybe a we need simple `bool setup_wl_egl(struct wl_display *display, EGLDisplay *display, EGLContext *context)` helper for examples, but I'm not a fan of having examples that aren't self-contained. OTOH these aren't self-contained right now since they depend on `wlr_egl`.

ammen99:
> OTOH these aren't self-contained right now since they depend on `wlr_egl`.

What is the definition of self-contained? They surely depend on wlroots?

emersion:
The clients using `wlr_egl` directly are Wayland clients, for instance [toplevel-decoration](https://github.com/swaywm/wlroots/blob/master/examples/toplevel-decoration.c). The compositor examples don't use `wlr_egl` directly, instead they rely on `wlr_renderer` and `wlr_backend`.

emersion:
Fixed in https://github.com/swaywm/wlroots/pull/2666.

PR #2677, opened 2021-01-22
render/egl: remove wlr_egl_context->(draw, read)_surface

berylline:
Just another bit of EGL housekeeping done for #1352 and #2624.

Breaking changes:

`wlr_egl_context->draw_surface` and `wlr_egl_context->read_surface` have been deleted and no longer exist.

Cc:

@emersion, @ascent12, @ammen99

EDIT: Just to clarify and to clear up confusion, this PR is intended to accompany #2666(which got merged not too long ago).

### DIFF
--- a/include/wlr/render/egl.h
+++ b/include/wlr/render/egl.h
@@ -32,8 +32,6 @@
 struct wlr_egl_context {
 	EGLDisplay display;
 	EGLContext context;
-	EGLSurface draw_surface;
-	EGLSurface read_surface;
 };
 
 struct wlr_egl {
--- a/render/egl.c
+++ b/render/egl.c
@@ -426,8 +426,6 @@ bool wlr_egl_is_current(struct wlr_egl *egl) {
 void wlr_egl_save_context(struct wlr_egl_context *context) {
 	context->display = eglGetCurrentDisplay();
 	context->context = eglGetCurrentContext();
-	context->draw_surface = eglGetCurrentSurface(EGL_DRAW);
-	context->read_surface = eglGetCurrentSurface(EGL_READ);
 }
 
 bool wlr_egl_restore_context(struct wlr_egl_context *context) {
@@ -444,8 +442,8 @@ bool wlr_egl_restore_context(struct wlr_egl_context *context) {
 		return true;
 	}
 
-	return eglMakeCurrent(display, context->draw_surface,
-			context->read_surface, context->context);
+	return eglMakeCurrent(display, EGL_NO_SURFACE,
+			EGL_NO_SURFACE, context->context);
 }
 
 EGLImageKHR wlr_egl_create_image_from_wl_drm(struct wlr_egl *egl,
--- a/render/egl.c
+++ b/render/egl.c
@@ -292,6 +292,12 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display) {
 		goto error;
 	}
 
+	if (!check_egl_ext(display_exts_str, "EGL_KHR_surfaceless_context")) {
+		wlr_log(WLR_ERROR, 
+			"EGL_KHR_surfaceless_context not supported");
+		goto error;
+	}
+
 	wlr_log(WLR_INFO, "Using EGL %d.%d", (int)major, (int)minor);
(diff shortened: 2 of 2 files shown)


emersion:
`wlr_egl_restore_context` is supposed to restore the EGL state to where it was at `wlr_egl_save_context` time. With this patch `wlr_egl_restore_context` resets the current surfaces to EGL_NO_SURFACE even if a surface was previously current.

We don't use surfaces in wlroots, but compositors might.

berylline:
Now that you've brought that possibility to my attention and now that I've thought about it, I think that's a good possibility to take into consideration, but are there any compositors that actually do something like this when it comes to rendering? I'm not sure if something like Wayfire or other compositors that are similar to or different from it would save and restore the `draw` and `read` surfaces, if it uses them. @ammen19, do you know if it does?

Also, don't forget: I've removed the `EGLSurface`s from `wlr_egl_context` itself, so even if a compositor were to use `EGLSurface`s, it'll probably have to use them on its own without the help of wlroots since it doesn't use surfaces anymore, if I know what I'm talking about, that is.

Actually, now that I'm thinking about it, I think that's actually a good idea, although I don't know what you think of it.

Compositors can store their own `EGLSurface`s and can use `eglMakeCurrent` (as well as the other EGL functions that require surfaces) to use them if they want to.

Do you think this is a good idea and a good obligation to have for compositors that want to do the above? Let me know.

emersion:
My point is: with this PR wlroots will unexpectedly reset the draw/read EGL surfaces at arbitrary points in time. This can happen e.g. when a buffer or texture is destroyed. I'd rather not do that.

berylline:
Sorry, ignore everything I said in the previous comment.

I think what I should've clarified is that this PR is meant to accompany #2666.

Anyway, moving on:

I think I understand what you're saying, but if some wlroots compositors were to use `EGLSurface`s, then technically, they wouldn't be able to do that since `wlr_egl_make_current` sets the `draw` and `read` surfaces to `EGL_NO_SURFACE`(and has been doing that since #2623 got merged not long ago) and since `wlr_egl_swap_buffers` has been removed(also as a result of #2623 being merged), so I'm not sure how they would be able to do so.

Since then(and because my other PR that's related to this one, #2666, was merged not long ago), there's virtually no room left for `EGLSurface`s to be used in wlroots now, AFAIK. And almost certainly because of this, my guess would be that both `wlr_egl_context->draw_surface` and `wlr_egl_context->read_surface` are going to return `EGL_NO_SURFACE` since no `EGLSurface` is actually being made current with a context, so IMO, there's virtually no reason to keep those `EGLSurface` struct members in `wlr_egl_context`.

Also, AFAIK, this is consistent with what wlroots can do when the (EGL_KHR_surfaceless_context)[https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_surfaceless_context.txt] extension is available.

And according to #1352(the `wlr_egl` section, specifically) isn't surface-less rendering supposed to be mandated for `wlr_egl` in wlroots anyway or am I missing or misunderstanding something?

Let me know if I am.

emersion:
> they wouldn't be able to do that

Nothing stops compositors from using EGL directly. Quite a few just do this.

berylline:
Uh.... Okay, this is embarrassing. I think my previous comment about `wlr_egl` should be ignored now. 

I forgot and/or didn't realise that `wlr_egl_context` is not linked or related to `wlr_egl` in any way, shape, sense or fashion. Sorry about that.

@emersion, actually taking what you're saying (in your current comment and your previous comments) into consideration now, I think you're right and you have a point. Sorry for misunderstanding.

Do you want to close this PR or do you want to leave it open? Personally, I think it should be closed because I think I wasted my own time basically trying to convince you to have me remove something that didn't need removing in the first place. 

Basically and personally, I don't mind what you do with it now, and that includes closing it, if you want to, that is.

berylline:
You know what, never mind, I'll close it.

emersion:
I think your second commit ("render/egl: add check for EGL_KHR_surfaceless_context") would be nice to have though. Feel free to submit it in a new PR.

berylline:
That's what I was thinking about doing as well! I'll submit that as a PR when I get the time and when I'm ready.

berylline:
I've split off the second commit into a separate PR now: #2702

PR #2679, opened 2021-01-22
output: split wlr_output_enable and wlr_output_enable_adaptive_sync

berylline:
Split `wlr_output_enable(output, enable)` into `wlr_output_enable(output)` and `wlr_output_disable(output)` and split `wlr_output_enable_adaptive_sync(output, enabled)` into `wlr_output_enable_adaptive_sync(output)` and `wlr_output_disable_adaptive_sync(output)`.

Breaking changes:

`wlr_output_enable` no longer takes a `bool enable` parameter and `wlr_output_enable_adaptive_sync` no longer takes a `bool enabled` parameter.

Solution:

Replace `wlr_output_enable(output, true)` with `wlr_output_enable(output)` and replace `wlr_output_enable(output, false)` with `wlr_output_disable(output)`.

Matching Sway PR: swaywm/sway#5965

Mentions:

@emersion

### DIFF
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -286,7 +286,7 @@ void wlr_output_set_custom_mode(struct wlr_output *output, int32_t width,
 void wlr_output_set_transform(struct wlr_output *output,
 	enum wl_output_transform transform);
 /**
- * Enables or disables adaptive sync (ie. variable refresh rate) on this
+ * Enables adaptive sync (ie. variable refresh rate) on this
  * output. This is just a hint, the backend is free to ignore this setting.
  *
  * When enabled, compositors can submit frames a little bit later than the
@@ -294,7 +294,14 @@ void wlr_output_set_transform(struct wlr_output *output,
  *
  * Adaptive sync is double-buffered state, see `wlr_output_commit`.
  */
-void wlr_output_enable_adaptive_sync(struct wlr_output *output, bool enabled);
+void wlr_output_enable_adaptive_sync(struct wlr_output *output);
+/**
+ * Disables adaptive sync (ie. variable refresh rate) on this
+ * output. This is just a hint, the backend is free to ignore this setting.
+ *
+ * Adaptive sync is double-buffered state, see `wlr_output_commit`.
+ */
+void wlr_output_disable_adaptive_sync(struct wlr_output *output);
 /**
  * Sets a scale for the output.
  *
--- a/include/wlr/types/wlr_output.h
+++ b/include/wlr/types/wlr_output.h
@@ -241,13 +241,20 @@ struct wlr_output_event_bind {
 struct wlr_surface;
 
 /**
- * Enables or disables the output. A disabled output is turned off and doesn't
- * emit `frame` events.
+ * Enables the output.
  *
- * Whether an output is enabled is double-buffered state, see
+ * Whether an output is enabled or disabled is double-buffered state, see
  * `wlr_output_commit`.
  */
-void wlr_output_enable(struct wlr_output *output, bool enable);
+void wlr_output_enable(struct wlr_output *output);
+/**
+ * Disables the output. A disabled output is turned off and doesn't
+ * emit `frame` events.
+ *
+ * Whether an output is enabled or disabled is double-buffered state, see
+ * `wlr_output_commit`.
+ */
+void wlr_output_disable(struct wlr_output *output);
 void wlr_output_create_global(struct wlr_output *output);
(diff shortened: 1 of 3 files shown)


emersion:
What is the motivation for this change?

berylline:
(Sorry again for not quote-replying. I'm using my phone again)

I figured that I needed an explanation for this. I have a feeling that I'm not going to explain this very well for it to make my reasons convincing, so brace yourself and please bear with me:

As I've come to educate myself more and more about design patterns and about how some people think (some) API functions should look, I've come to see why boolean parameters for on/off(or, for this case, enable/disable) functions can be off-putting for certain programmers and why they can be bad for certain types of situations.

For example, `wlr_output_enable(output, true)` enables the output and `wlr_output_enable(output, false)` disables the output, but it's not obvious to the compositor author that it does either of these two actions unless you read the description that sits above `wlr_output_enable` and look at the part of the `wlr_output_enable` function declaration that says `bool enable`, which tells you what it does. 

Note that my reason for splitting these functions is *not* to split `wlr_output_enable` *and* remove the description that sits above it. It is simply to make `wlr_output_enable` more readable and obvious as a function(or *two* functions, as is proposed here in this PR) so compositor authors can see `wlr_output_enable(output)` and `wlr_output_disable(output)` for what they obviously are as functions for enabling and disabling outputs instead of seeing `wlr_output_enable(output, enable)` and figuring that they have to use `enable` to enable and disable outputs, which makes them keep an extra thing in mind instead of just using the function as is and not worry about making a potential mistake, which may or may not sound like a big deal, but which I think is a life improvement and one less mistake made, anyway and regardless, take it or leave it.

Another thing which may or may not be a big deal(I'm guessing you'd probably think the latter, but I don't know what you think yet, so I'm all ears and I'm going to remain impartial about this), but which I also think is an improvement is having one less thing to type. Whether that matters or not depends on what you think.

Note also that I'm trying to see this from other people's perspectives. On one hand, I personally think that me(or someone else, I guess, for that matter) typing something like `wlr_output_enable(output, true)` isn't going to be the end of the world, but on the other hand, I also think it's going to be irritating and aggravating to look at and I can understand why some people would get irritated and aggravated by looking at something like `wlr_output_enable(output, true)`.

Of course, my(and possibly some other people's) feelings themselves don't explain why I think `wlr_output_enable(output, enable)` should be changed to `wlr_output_enable(output)` and `wlr_output_disable(output)`, hence why I've written this comment.

I don't think my comment itself is going to explain my reasons entirely, so I'm going to leave a few links to articles written by some people so you can understand why I made this PR, hopefully(I also think they make for good reading in general, by the way), if you want to read them, that is:

https://martinfowler.com/bliki/FlagArgument.html (a slightly more impartial article about boolean arguments, but I think it still makes for good reading, regardless)

https://www.clearslide.com/blog/thought-leadership/psa-avoid-flag-arguments/ (harsh-sounding article, but the author is displaying their frustration, which I think is understanding in this case)

https://joelforjava.com/blog/2018/09/21/refactoring-replace-parameter-with-explicit-methods.html (it's for Java and it's also essentially a guide, but the first part of the article explains why its author refactored their code for a project of theirs)

https://ardalis.com/are-boolean-flags-on-methods-a-code-smell/ (what I think is a detailed and reasonable article about why boolean arguments are not a good idea, in their opinion)

https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3 (ever so *slightly* about Objective-C, but I still think this is generally a good article to look at, and it's from an iOS developer's perspective too, which isn't relevant to this discussion, but I still think it's good to take different perspectives into consideration here)

emersion:
Sorry, but I don't find this compelling. While I agree functions which do too many things are a bad design, I don't think this is the case here.

`wlr_output_enable` would perhaps have been better named `wlr_output_set_enabled`, but it's not enough to justify a breaking API change IMO.

Next time, please open a proposal issue or discuss on IRC before working on a PR with breaking changes. This will avoid you wasting time on rejected PRs.

berylline:
Sorry for the somewhat late message.

Yeah, I think you're right about proposing such changes on IRC and I agree I could've done that instead, although to be fair, at the time, I didn't really think about it. I'll try doing that next time, though. I also now think that how I went about with my now closed PRs wasn't really formal or agreed upon and I can see how that can be perceived as being rude, so I apologise for doing that and coming off that way.

I'm not sure if this is the end of the road here or if we should discuss my PRs more on IRC.

What do you think?

emersion:
I don't find it rude or anything. It's just that I don't want to waste your time.

PR #2680, opened 2021-01-23
output: split wlr_output_update_enabled

berylline:
This PR splits `wlr_output_update_enabled(output, enabled)` into `wlr_output_update_enabled(output)` and `wlr_output_update_disabled(output)`. 

I'm not going to repeat the reasons why I would like `wlr_output_update_enabled` to be split in this PR, so please see #2679 for the reasons why as the reasons stated there apply to this PR as well.

And please see this comment too, for clarification and more info:

https://github.com/swaywm/wlroots/pull/2679#issuecomment-765792893

(Although the above comment applies to a different function, `wlr_output_enable`, it also applies to why I'm doing it here with `wlr_output_update_enabled`, just FYI to clarify and not cause confusion, I hope.)

(And although I think because `wlr_output_update_enabled` seems to be a private function that also seems to only be used by the backends and not compositors that utilize wlroots (also), I think, it would seem that changing/refactoring it doesn't really matter as much and/or as such (and I don't really think so, in my opinion), unless I'm wrong and it *does* matter, that is.)

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -785,7 +785,7 @@ bool drm_connector_set_mode(struct wlr_drm_connector *conn,
 			realloc_crtcs(drm);
 			attempt_enable_needs_modeset(drm);
 		}
-		wlr_output_update_enabled(&conn->output, false);
+		wlr_output_update_disabled(&conn->output);
 		return true;
 	}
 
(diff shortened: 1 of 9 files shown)


emersion:
See https://github.com/swaywm/wlroots/pull/2679#issuecomment-765890235.

PR #2681, opened 2021-01-23
Add documenation and output selection to foreign toplevel example

fwSmit:
Made this for testing https://github.com/swaywm/sway/pull/5955.
While I was at it, I added some documentation as well.

### DIFF
--- a/examples/foreign-toplevel.c
+++ b/examples/foreign-toplevel.c
@@ -9,26 +9,27 @@
 
 #define WLR_FOREIGN_TOPLEVEL_MANAGEMENT_VERSION 3
 
-/**
- * Usage:
- * 1. foreign-toplevel
- * 	  Prints a list of opened toplevels
- * 2. foreign-toplevel -f <id>
- *    Focus the toplevel with the given id
- * 3. foreign-toplevel -a <id>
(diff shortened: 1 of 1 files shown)


emersion:
I don't think we should have an option to print all outputs. Users can already use `wayland-info` to print them.

Please fix the indentation of these changes. We use tabs.

fwSmit:
Sure, I fixed.

examples/foreign-toplevel.c
```diff
@@ -48,6 +49,62 @@ struct toplevel_state {
...
 };
 
+struct wayland_output {
+	struct wl_list link;
+	uint32_t id;
+	struct wl_output *output;
+	char *make;
+	char *model;
+	int width;
+	int height;
```
ifreund, line 59: These are never used.

ifreund (commented):
it might be simpler to handle this in the same way the layer-shell example client does instead of storing a list.

fwSmit:
I guess I copied from the wrong example. The layer-shell example is much easier. I've implemented it that way now.

emersion (approved):
LGTM, thanks!

PR #2684, opened 2021-01-24
xwayland: use strerror(...) instead of %m

Xyene:
Previously, any error would be masked by an internal isatty call:

```
24:31:48.174 [DEBUG] [wlr] [xwayland/selection/incoming.c:386] XCB_SELECTION_NOTIFY (selection=277, property=278, target=256)
24:31:48.174 [ERROR] [wlr] [xwayland/selection/incoming.c:30] write error to target fd: Inappropriate ioctl for device
```


emersion (changes requested):
Good catch! Can we use `wlr_log_errno` instead?

Xyene:
Done, wasn't aware it existed :)

emersion (approved):
Thanks!

PR #2685, opened 2021-01-25
xwayland/selection: prevent fd leak on unsupported MIME type

Xyene:
Since we never end up calling xcb_convert_selection, the file descriptor
ends up getting leaked (i.e., not cleaned up within
xwm_data_source_write).


emersion (approved):
Thanks!

PR #2687, opened 2021-01-26
xwayland/selection: simplify incremental transfer control flow

Xyene:
This PR makes incremental transfer error-handling a bit simpler to follow, and easier to ascertain the correctness of. There should be no behavioral changes; I've tested transfers (incremental and non-incremental) still work. See individual commits for details and rationale.

### DIFF
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -61,9 +61,12 @@ static void xwm_write_property(struct wlr_xwm_selection_transfer *transfer,
 	transfer->property_start = 0;
 	transfer->property_reply = reply;
 
-	xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer);
-
-	if (transfer->property_reply != NULL) {
+	bool wl_client_finished_consuming =
+		!xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer);
+	if (!wl_client_finished_consuming) {
+		// Wrote out part of the property to the Wayland client, but the client was
+		// unable to accept all of it. Schedule an event to asynchronously complete
+		// the transfer.
 		struct wl_event_loop *loop =
 			wl_display_get_event_loop(xwm->xwayland->wl_display);
 		transfer->event_source = wl_event_loop_add_fd(loop,
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -100,8 +100,8 @@ void xwm_get_incr_chunk(struct wlr_xwm_selection_transfer *transfer) {
 	if (xcb_get_property_value_length(transfer->property_reply) > 0) {
 		xwm_write_selection_property_to_wl_client(transfer);
 	} else {
-		xwm_selection_transfer_finish(transfer);
 		wlr_log(WLR_DEBUG, "incremental transfer complete");
+		xwm_selection_transfer_finish(transfer);
 	}
 }
 
@@ -117,7 +117,6 @@ static void xwm_selection_get_data(struct wlr_xwm_selection *selection) {
 		transfer->incr = true;
 		xwm_selection_transfer_destroy_property_reply(transfer);
 	} else {
-		transfer->incr = false;
 		// Reply's ownership is transferred to wm, which is responsible for freeing
 		// it.
 		xwm_write_selection_property_to_wl_client(transfer);
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -61,12 +61,8 @@ static int xwm_data_source_write(int fd, uint32_t mask, void *data) {
 	return 1;
 }
 
-static void xwm_write_property(struct wlr_xwm_selection_transfer *transfer,
-		xcb_get_property_reply_t *reply) {
+static void xwm_write_property(struct wlr_xwm_selection_transfer *transfer) {
 	struct wlr_xwm *xwm = transfer->selection->xwm;
-
-	transfer->property_start = 0;
-	transfer->property_reply = reply;
 
 	bool wl_client_finished_consuming =
 		!xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer);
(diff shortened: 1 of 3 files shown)


xwayland/selection/incoming.c
```diff
@@ -61,7 +61,9 @@ static void xwm_write_property(struct wlr_xwm_selection_transfer *transfer,
 	transfer->property_start = 0;
 	transfer->property_reply = reply;
 
-	xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer);
+	if (!xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer)) {
```
emersion, line 71: This is maybe a little confusing, because `xwm_data_source_write` doesn't return a `bool` indicating success/error. Instead, it returns an int which indicates whether we need to re-check the FD. Maybe we could split that up in a variable to make it more explicit? e.g.

```c
bool needs_recheck = xwm_data_source_write(…);
```

xwayland/selection/incoming.c
```diff
@@ -28,7 +28,7 @@ static int xwm_data_source_write(int fd, uint32_t mask, void *data) {
 		xwm_selection_transfer_destroy_property_reply(transfer);
 		xwm_selection_transfer_remove_event_source(transfer);
 		xwm_selection_transfer_close_wl_client_fd(transfer);
-		return 1;
+		return 0;
```
emersion, line 41: The docs say:

> If the event source is registered for re-check with wl_event_source_check(): 0 for all done, 1 for needing a re-check. If not registered, the return value is ignored and should be zero. 

The event source has not be registered with `wl_event_source_check`, so the return value is ignored.

I can see how this can be useful in the next commit though.

Xyene: Good point, I've fixed the commit message.

xwayland/selection/incoming.c
Xyene: Good idea, done.

xwayland/selection/incoming.c
```diff
@@ -25,113 +40,115 @@ static int xwm_data_source_write(int fd, uint32_t mask, void *data) {
...
-		/* Reply's ownership is transferred to xwm, which is responsible
-		 * for freeing it */
-		xwm_write_property(transfer, reply);
+	if (xcb_get_property_value_length(transfer->property_reply) > 0) {
+		xwm_write_selection_property_to_wl_client(transfer);
 	} else {
-		wlr_log(WLR_DEBUG, "transfer complete");
-		xwm_selection_transfer_close_wl_client_fd(transfer);
-		free(reply);
+		xwm_selection_transfer_finished(transfer);
```
Xyene, line 131: Note for reviewer: if we hit this branch, in principle the event source has already been destroyed. However, we use `xwm_selection_transfer_finished`, which has an internal "useless" call to `xwm_selection_transfer_remove_event_source` so future readers don't have to figure this out :)

xwayland/selection/incoming.c
```diff
@@ -25,113 +40,115 @@ static int xwm_data_source_write(int fd, uint32_t mask, void *data) {
...
+		return false;
+	}
+
+	return true;
+}
+
+void xwm_get_incr_chunk(struct wlr_xwm_selection_transfer *transfer) {
+	wlr_log(WLR_DEBUG, "xwm_get_incr_chunk");
+
+	if (transfer->property_reply) {
```
Xyene, line 119: Note for reviewer: I ran into this being a source of compositor segfaults while introducing bugs in early drafts of this code. Since it's possible for an X11 client to do this, I think it's worth guarding against.

Current behavior doesn't try to do anything fancy in this obscure case. An error message will be printed, and the clipboard will become "stuck" because we're no longer servicing the transfer.

xwayland/selection/incoming.c
```diff
@@ -61,9 +61,12 @@ static void xwm_write_property(struct wlr_xwm_selection_transfer *transfer,
 	transfer->property_start = 0;
 	transfer->property_reply = reply;
 
-	xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer);
-
-	if (transfer->property_reply != NULL) {
+	bool wl_client_finished_consuming =
+		!xwm_data_source_write(transfer->wl_client_fd, WL_EVENT_WRITABLE, transfer);
+	if (!wl_client_finished_consuming) {
```
emersion, line 76: The double negation is a little complicated here. At least it happens in two steps :P

xwayland/selection/selection.c
```diff
@@ -33,6 +33,13 @@ void xwm_selection_transfer_destroy_property_reply(
 	transfer->property_reply = NULL;
 }
 
+void xwm_selection_transfer_finished(
```
emersion, line 36: Maybe rename to `xwm_selection_transfer_finish` for consistency with the rest of wlroots?

Xyene: Changed to `_finish`.

xwayland/selection/incoming.c
Xyene: Agreed, I tried to compensate by being **very** explicit with the naming :)

Xyene:
I've snuck in one more fix: https://github.com/swaywm/wlroots/pull/2687/commits/025a754e1ac13cfce3898a1f7a7b92e3857204d0

xwayland/selection/incoming.c
```diff
@@ -25,113 +40,87 @@ static int xwm_data_source_write(int fd, uint32_t mask, void *data) {
...
-		 * for freeing it */
-		xwm_write_property(transfer, reply);
+	if (xcb_get_property_value_length(transfer->property_reply) > 0) {
+		xwm_write_selection_property_to_wl_client(transfer);
 	} else {
-		wlr_log(WLR_DEBUG, "transfer complete");
-		xwm_selection_transfer_close_wl_client_fd(transfer);
-		free(reply);
+		wlr_log(WLR_DEBUG, "incremental transfer complete");
+		transfer->incr = false;
```
Xyene, line 104: Maybe we want this to be part of `xwm_selection_transfer_finish`. I haven't thought too much about it.

Xyene: Yeah, I think I like that more.

emersion (approved):
LGTM, thanks for working on this!

PR #2688, opened 2021-01-26
xwayland: fix extraneous NET_WM_STATE log messages

emersion:
wlroots would log "Unhandled NET_WM_STATE property change" log
messages for atoms we know about. Simplify the code structure
and remove these extra messages.


ammen99 (approved):
LGTM.

PR #2689, opened 2021-01-27
render/egl: fail to create wlr_egl on software renderer

bl4ckb0ne:
The creation of `wlr_egl` will fail is the device extension
EGL_MESA_device_software is defined. The creation process is allowed to
continue only if the environment variable `WLR_RENDERER_ALLOW_SOFTWARE`
is defiled.

### DIFF
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -39,6 +39,11 @@ wlroots reads these environment variables
 
 * *WLR_X11_OUTPUTS*: when using the X11 backend specifies the number of outputs
 
+## gles2 renderer
+
+* *WLR_RENDERER_ALLOW_SOFTWARE*: allows the gles2 renderer to use software
+  rendering
+
 # Generic
 
 * *DISPLAY*: if set probe X11 backend in `wlr_backend_autocreate`
--- a/render/egl.c
+++ b/render/egl.c
@@ -272,6 +272,18 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			goto error;
 		}
 
+		if (check_egl_ext(device_exts_str, "EGL_MESA_device_software")) {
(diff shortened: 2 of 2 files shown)


render/egl.c
```diff
@@ -272,6 +272,14 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display,
 			goto error;
 		}
 
+		if (check_egl_ext(device_exts_str, "EGL_MESA_device_software") &&
+				!getenv("WLR_RENDERER_ALLOW_SOFTWARE")) {
```
emersion, line 276: Can we only lift the check if the env var is exactly "1"? Other `WLR_` env vars ignore other values (like "0").

emersion (changes requested):
Can we log an INFO message when we're using a software renderer and `WLR_RENDERER_ALLOW_SOFTWARE=1` is set? Juts to make it clear software rendering is used.

Can we document this new env var in `doc/env_vars.md`?

render/egl.c
bl4ckb0ne: Sure!

bl4ckb0ne:
I added the doc, tell me if you see any spelling mistake.

docs/env_vars.md
```diff
@@ -39,6 +39,11 @@ wlroots reads these environment variables
 
 * *WLR_X11_OUTPUTS*: when using the X11 backend specifies the number of outputs
 
+## gles2 renderer
+
+* *WLR_RENDERER_ALLOW_SOFTWARE*: allows the gles2 renderer to use software
```
emersion, line 44: Note to self: this is a GLES2-specific variable, but it doesn't contain "GLES2" in the name. I'd like to implement a similar check in other non-GLES2 renderers such as Vulkan.

emersion (approved):
LGTM, thanks!

PR #2690, opened 2021-01-27
Build tinywl from meson

bl4ckb0ne:
I am using this in my dev branches, so I figured it would benefit the community. Option is off by default.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -17,15 +17,11 @@ sources:
 tasks:
   - setup: |
       cd wlroots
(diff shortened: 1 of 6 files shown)


emersion:
I'm not a fan of having to maintain two separate build systems. I don't want to drop the Makefile because it's self-contained and most developers know this file format.

bl4ckb0ne:
How much maintainance has been done to tinywl's makefile in the past? Both build files are relatively small, I don't think we will have too much work to maintain those.

bl4ckb0ne:
Closed for being not relevant.

PR #2692, opened 2021-01-28
WIP: pixman allocator

bl4ckb0ne:


emersion:
We probably want a DRM dumb buffer allocator. Pixman is not meant for buffer allocation.

PR #2693, opened 2021-01-28
output: introduce request_state event

emersion:
This allows backends to request the compositor to change the state of an output.

References: https://github.com/swaywm/wlroots/issues/2300#issuecomment-761819673

* * *

Breaking change: Backends no longer change the output mode behind the compositor's back. To retain the previous behavior, compositors can listen to the `request_state` event and call `wlr_output_queue_state` followed by `wlr_output_commit`.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2693

PR #2694, opened 2021-01-29
backend/wayland: flush remote display

emersion:
It turns out wl_event_source_check is not enough to guarantee that the
remote wl_display will be flushed after we queue requests. We need to
explicitly flush, just like we do in our X11 code.

References: https://gitlab.freedesktop.org/wayland/wayland/-/issues/187


emersion:
It turns out this is not a workaround, this is what needs to be done for correct `wl_event_loop` usage. See upstream issue for details.

PR #2697, opened 2021-01-29
virtual_keyboard: Release keys on destroy

### LINKED ISSUE #2034: virtual-keyboard: release unreleased keys
Follow https://github.com/swaywm/sway/pull/4932#issuecomment-586108940.

It's not restricted to IME. Any application aborted may leave some unreleased virtual_keyboard key events.  I think there should be similar thing to https://github.com/swaywm/wlroots/blob/master/types/wlr_keyboard_group.c#L203-L219.

As it is using wlr_keyboard, so pressed keys are recorded. So keyboard_group can handle unreleased keys, then why can not virtual_keyboard?

tadeokondrak:
Fixes #2034.

### DIFF
--- a/include/wlr/interfaces/wlr_keyboard.h
+++ b/include/wlr/interfaces/wlr_keyboard.h
@@ -19,6 +19,7 @@ struct wlr_keyboard_impl {
 
 void wlr_keyboard_init(struct wlr_keyboard *keyboard,
 		const struct wlr_keyboard_impl *impl);
+void wlr_keyboard_release_keys(struct wlr_keyboard *keyboard);
 void wlr_keyboard_destroy(struct wlr_keyboard *keyboard);
 void wlr_keyboard_notify_key(struct wlr_keyboard *keyboard,
 		struct wlr_event_keyboard_key *event);
--- a/types/wlr_keyboard.c
+++ b/types/wlr_keyboard.c
@@ -1,7 +1,9 @@
+#define _POSIX_C_SOURCE 200809L
 #include "util/array.h"
 #include <assert.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <wayland-server-core.h>
 #include <wlr/interfaces/wlr_keyboard.h>
 #include <wlr/types/wlr_keyboard.h>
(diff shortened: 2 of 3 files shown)


emersion:
Hm, but maybe this applies to physical keyboards as well? What happens if you hold a key and disconnect your keyboard? Same applies to the keyboards for the Wayland/X11 backends.

Maybe this logic should be part of [`wlr_keyboard_destroy`](https://github.com/swaywm/wlroots/blob/7b50f5d3249173d93736d6e2a3c6f61e86b367bd/types/wlr_keyboard.c#L127)?

tadeokondrak:
> Hm, but maybe this applies to physical keyboards as well? What happens if you hold a key and disconnect your keyboard?

It doesn't keep repeating, I'm not sure what handles this.

> Wayland backend

Is the parent Wayland compositor leaving a key pressed when removing the keyboard capability from a seat something that clients should handle?

> X11 backend

It only has one seat, so keyboards can't be removed during runtime.

include/wlr/interfaces/wlr_keyboard.h
```diff
@@ -19,6 +19,7 @@ struct wlr_keyboard_impl {
 
 void wlr_keyboard_init(struct wlr_keyboard *keyboard,
 		const struct wlr_keyboard_impl *impl);
+void wlr_keyboard_release_keys(struct wlr_keyboard *keyboard);
```
tadeokondrak, line 22: Maybe this shouldn't be a public function. Not sure what to do instead, though.

types/wlr_keyboard.c
```diff
@@ -124,6 +126,22 @@ void wlr_keyboard_init(struct wlr_keyboard *kb,
...
+			.keycode = kb->keycodes[0],
+			.update_state = false,
+			.state = WL_KEYBOARD_KEY_STATE_RELEASED,
+		};
+		wlr_keyboard_notify_key(kb, &event);
+	}
+}
+
 void wlr_keyboard_destroy(struct wlr_keyboard *kb) {
 	if (kb == NULL) {
```
jaens, line 146: Could `wlr_keyboard_release_keys` be called inside `wlr_keyboard_destroy` instead?
That way it could be private.

emersion: That would be best indeed. I'm not a fan of having keyboard-specific stuff inside the common code.

tadeokondrak: `wlr_keyboard_destroy` is called after the `destroy` signal on the `wlr_input_device` is emitted.
If the keys are released after that signal, the original issue isn't fixed anymore.

emersion:
I wonder if instead we should keep the status quo and make it so destroying a keyboard implicitly releases all keys.

emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2918

PR #2698, opened 2021-01-29
examples: remove "major" and "minor" from egl_common.c

berylline:
Remove "major" and "minor" from egl_common.c as they are not used by the examples that use `egl_common.c`.

In other words, just polishing `egl_common.c`

Also, whoops! Sorry about that. I should've removed those two variables as those were leftovers when I was copying from `render/egl.c` before #2666 got merged. :grimacing:

### DIFF
--- a/examples/egl_common.c
+++ b/examples/egl_common.c
@@ -81,8 +81,7 @@ bool egl_init(struct wl_display *display) {
 		goto error;
 	}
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

berylline:
You're welcome! :grinning:

PR #2699, opened 2021-01-29
Clipboard fixes part 2

Xyene:
I was originally going to open a separate PR for each change I made, then realized having them all in one place made testing easier &mdash; so here it is. Commits are independent, and their messages have more details.

### DIFF
--- a/include/xwayland/selection.h
+++ b/include/xwayland/selection.h
@@ -77,6 +77,6 @@ bool primary_selection_source_is_xwayland(
 void xwm_seat_handle_start_drag(struct wlr_xwm *xwm, struct wlr_drag *drag);
 
 void xwm_selection_init(struct wlr_xwm *xwm);
-void xwm_selection_finish(struct wlr_xwm *xwm);
+void xwm_selection_finish(struct wlr_xwm_selection *selection);
 
 #endif
--- a/include/xwayland/selection.h
+++ b/include/xwayland/selection.h
@@ -77,7 +77,8 @@ bool primary_selection_source_is_xwayland(
 
 void xwm_seat_handle_start_drag(struct wlr_xwm *xwm, struct wlr_drag *drag);
 
-void xwm_selection_init(struct wlr_xwm *xwm);
+void xwm_selection_init(struct wlr_xwm_selection *selection,
+	struct wlr_xwm *xwm, xcb_atom_t atom);
 void xwm_selection_finish(struct wlr_xwm_selection *selection);
 
 #endif
--- a/include/xwayland/selection.h
+++ b/include/xwayland/selection.h
@@ -47,6 +47,7 @@ void xwm_selection_transfer_close_wl_client_fd(
 	struct wlr_xwm_selection_transfer *transfer);
 void xwm_selection_transfer_destroy_property_reply(
 	struct wlr_xwm_selection_transfer *transfer);
+void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer);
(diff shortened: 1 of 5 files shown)


xwayland/selection/incoming.c
```diff
@@ -382,11 +397,14 @@ void xwm_handle_selection_notify(struct wlr_xwm *xwm,
...
 	if (event->property == XCB_ATOM_NONE) {
 		wlr_log(WLR_ERROR, "convert selection failed");
+		xwm_selection_transfer_finish(&selection->incoming);
 	} else if (event->target == xwm->atoms[TARGETS]) {
 		// No xwayland surface focused, deny access to clipboard
 		if (xwm->focus_surface == NULL) {
 			wlr_log(WLR_DEBUG, "denying write access to clipboard: "
 				"no xwayland surface focused");
+			// Would leak this transfer otherwise. Should never happen.
+			assert(selection->incoming.wl_client_fd <= 0);
```
Xyene, line 407: If we were being defensive about this, this ought to be a `xwm_selection_transfer_finish`. However, getting here with valid `wl_client_fd` is indicative of much larger problems, so it'd be good to be alerted (by way of a bug report) if this happens.

xwayland/selection/incoming.c
```diff
@@ -150,6 +150,21 @@ static void source_send(struct wlr_xwm_selection *selection,
 		return;
 	}
 
+	// FIXME: we currently can't handle two X11-to-Wayland transfers at once due
+	// to reusing the same X11 window. Proceeding further here would lead us to
+	// lose track of the current `transfer->wl_client_fd` and use-after-free
+	// during cleanup. This doesn't happen often, but bail now to avoid a
+	// compositor crash later.
+	if (transfer->wl_client_fd > 0) {
```
emersion, line 158: Zero is a valid value for an FD. If the compositor closes stdin, FD 0 could be re-used for a selection.

Xyene: You're right, but 0 is also what `wl_client_fd` is "set" to post-`calloc`, so including 0 in this check as-is would totally break copy/paste (the very first transfer would perpetually be rejected). I've introduced some more logic to initialize `wl_client_fd` to -1. In the future, we can use `xwm_selection_transfer_init` to do the X11 window allocation stuff, too.

emersion: Ah, good catch!

include/xwayland/selection.h
```diff
@@ -77,6 +77,6 @@ bool primary_selection_source_is_xwayland(
 void xwm_seat_handle_start_drag(struct wlr_xwm *xwm, struct wlr_drag *drag);
 
 void xwm_selection_init(struct wlr_xwm *xwm);
-void xwm_selection_finish(struct wlr_xwm *xwm);
+void xwm_selection_finish(struct wlr_xwm_selection *selection);
```
emersion, line 81: Hm, this creates a discrepancy with `xwm_selection_init`. Either:

- Maybe just rename to `xwm_finish_selection`?
- Maybe take a `struct wlr_xwm_selection *` in `xwm_selection_init` too?

emersion:
I've pushed the first 3 commits in master because they were reviewed. Feel free to rebase.

include/xwayland/selection.h
Xyene: I've made `xwm_selection_finish` take a `struct wlr_xwm_selection *` too now.

emersion (approved):
Thanks!

PR #2701, opened 2021-01-31
xwayland/selection: use one X11 window per incoming transfer

Xyene:
Depends on https://github.com/swaywm/wlroots/pull/2699.

### DIFF
--- a/include/xwayland/selection.h
+++ b/include/xwayland/selection.h
@@ -20,26 +20,32 @@ struct wlr_xwm_selection_transfer {
 	struct wl_array source_data;
 	int wl_client_fd;
 	struct wl_event_source *event_source;
+	struct wl_list link;
 
 	// when sending to x11
 	xcb_selection_request_event_t request;
-	struct wl_list outgoing_link;
 
 	// when receiving from x11
 	int property_start;
 	xcb_get_property_reply_t *property_reply;
+	xcb_window_t incoming_window;
 };
 
 struct wlr_xwm_selection {
 	struct wlr_xwm *xwm;
+
 	xcb_atom_t atom;
 	xcb_window_t window;
 	xcb_window_t owner;
 	xcb_timestamp_t timestamp;
 
-	struct wlr_xwm_selection_transfer incoming;
+	struct wl_list incoming;
 	struct wl_list outgoing;
 };
+
+struct wlr_xwm_selection_transfer *
+xwm_selection_find_incoming_transfer_by_window(
+	struct wlr_xwm_selection *selection, xcb_window_t window);
 
 void xwm_selection_transfer_remove_event_source(
 	struct wlr_xwm_selection_transfer *transfer);
@@ -47,10 +53,10 @@ void xwm_selection_transfer_close_wl_client_fd(
 	struct wlr_xwm_selection_transfer *transfer);
 void xwm_selection_transfer_destroy_property_reply(
 	struct wlr_xwm_selection_transfer *transfer);
-void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer);
-void xwm_selection_transfer_finish(struct wlr_xwm_selection_transfer *transfer);
-bool xwm_selection_transfer_get_selection_property(
-	struct wlr_xwm_selection_transfer *transfer, bool delete);
+void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer,
+	struct wlr_xwm_selection *selection);
+void xwm_selection_transfer_destroy(
+	struct wlr_xwm_selection_transfer *transfer);
 
 void xwm_selection_transfer_destroy_outgoing(
 	struct wlr_xwm_selection_transfer *transfer);
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -11,13 +11,84 @@
 #include "xwayland/selection.h"
 #include "xwayland/xwm.h"
 
+static struct wlr_xwm_selection_transfer *
+xwm_selection_transfer_create_incoming(struct wlr_xwm_selection *selection) {
+	struct wlr_xwm_selection_transfer *transfer = calloc(1, sizeof(*transfer));
+	if (!transfer) {
+		return NULL;
+	}
+
+	xwm_selection_transfer_init(transfer, selection);
+
+	wl_list_insert(&selection->incoming, &transfer->link);
+
+	struct wlr_xwm *xwm = selection->xwm;
+	transfer->incoming_window = xcb_generate_id(xwm->xcb_conn);
+	xcb_create_window(
+		xwm->xcb_conn,
+		XCB_COPY_FROM_PARENT,
+		transfer->incoming_window,
+		xwm->screen->root,
+		0, 0,
+		10, 10,
+		0,
+		XCB_WINDOW_CLASS_INPUT_OUTPUT,
+		xwm->screen->root_visual,
+		XCB_CW_EVENT_MASK, (uint32_t[]){
+			XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_PROPERTY_CHANGE
+		}
+	);
+	xcb_flush(xwm->xcb_conn);
+
+	return transfer;
+}
+
+struct wlr_xwm_selection_transfer *
+xwm_selection_find_incoming_transfer_by_window(
+		struct wlr_xwm_selection *selection, xcb_window_t window) {
+	struct wlr_xwm_selection_transfer *transfer;
+	wl_list_for_each(transfer, &selection->incoming, link) {
+		if (transfer->incoming_window == window) {
+			return transfer;
+		}
+	}
+
+	return NULL;
+}
+
+static bool xwm_selection_transfer_get_incoming_selection_property(
+		struct wlr_xwm_selection_transfer *transfer, bool delete) {
+	struct wlr_xwm *xwm = transfer->selection->xwm;
+
+	xcb_get_property_cookie_t cookie = xcb_get_property(
+		xwm->xcb_conn,
+		delete,
+		transfer->incoming_window,
+		xwm->atoms[WL_SELECTION],
+		XCB_GET_PROPERTY_TYPE_ANY,
+		0, // offset
+		0x1fffffff // length
+	);
+
+	transfer->property_start = 0;
+	transfer->property_reply =
+		xcb_get_property_reply(xwm->xcb_conn, cookie, NULL);
+
+	if (!transfer->property_reply) {
+		wlr_log(WLR_ERROR, "cannot get selection property");
+		return false;
+	}
+
+	return true;
+}
+
 static void xwm_notify_ready_for_next_incr_chunk(
 		struct wlr_xwm_selection_transfer *transfer) {
 	struct wlr_xwm *xwm = transfer->selection->xwm;
 	assert(transfer->incr);
 
 	wlr_log(WLR_DEBUG, "deleting property");
-	xcb_delete_property(xwm->xcb_conn, transfer->selection->window,
+	xcb_delete_property(xwm->xcb_conn, transfer->incoming_window,
 		xwm->atoms[WL_SELECTION]);
 	xcb_flush(xwm->xcb_conn);
 
@@ -40,7 +111,7 @@ static int write_selection_property_to_wl_client(int fd, uint32_t mask,
 	ssize_t len = write(fd, property + transfer->property_start, remainder);
 	if (len == -1) {
 		wlr_log_errno(WLR_ERROR, "write error to target fd %d", fd);
-		xwm_selection_transfer_finish(transfer);
+		xwm_selection_transfer_destroy(transfer);
 		return 0;
 	}
 
@@ -56,7 +127,7 @@ static int write_selection_property_to_wl_client(int fd, uint32_t mask,
 		xwm_notify_ready_for_next_incr_chunk(transfer);
 	} else {
 		wlr_log(WLR_DEBUG, "transfer complete");
-		xwm_selection_transfer_finish(transfer);
+		xwm_selection_transfer_destroy(transfer);
 	}
 
 	return 0;
@@ -93,7 +164,7 @@ void xwm_get_incr_chunk(struct wlr_xwm_selection_transfer *transfer) {
 		return;
 	}
 
-	if (!xwm_selection_transfer_get_selection_property(transfer, false)) {
+	if (!xwm_selection_transfer_get_incoming_selection_property(transfer, false)) {
 		return;
 	}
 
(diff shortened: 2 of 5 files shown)


Xyene:
This is ready for review now -- all commits except the last are from #2699, and I can rebase them away once #2699 is merged.

xwayland/selection/outgoing.c
```diff
@@ -297,7 +297,6 @@ static bool xwm_selection_send_data(struct wlr_xwm_selection *selection,
 		return false;
 	}
 
-	xwm_selection_transfer_init(transfer);
```
Xyene: Temporarily removed because this is the outgoing path, which has no logic to release the X11 window yet (so this would just leak).

emersion: Maybe replace with `transfer->wl_client_fd = -1` temporarily, and add a TODO comment about it?

include/xwayland/selection.h
```diff
@@ -28,19 +28,34 @@ struct wlr_xwm_selection_transfer {
...
 	xcb_get_property_reply_t *property_reply;
+
+	// FIXME: these are only used for X11 -> Wayland for now, but are initialized
+	// for both directions.
+	//
+	// Wayland -> X11 continues to use the `wlr_xwm_selection` versions.
+	xcb_window_t window;
+	xcb_timestamp_t timestamp;
+
+	struct wl_list incoming_link;
```
emersion, line 39: We could use a single `struct wl_list link` field which is either outgoing, either incoming. That way `xwm_selection_transfer_finish` can just `wl_list_remove(&transfer->link)`.

xwayland/selection/selection.c
```diff
@@ -33,15 +46,48 @@ void xwm_selection_transfer_destroy_property_reply(
...
+			XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_PROPERTY_CHANGE
+		}
+	);
+
+	xcb_set_selection_owner(xwm->xcb_conn, transfer->window,
+			xwm->atoms[CLIPBOARD_MANAGER], XCB_TIME_CURRENT_TIME);
+	xcb_flush(xwm->xcb_conn);
 }
 
 void xwm_selection_transfer_finish(
```
emersion, line 73: This function now frees the transfer, so should be renamed to `xwm_selection_transfer_destroy`.

xwayland/selection/selection.c
```diff
@@ -33,15 +46,48 @@ void xwm_selection_transfer_destroy_property_reply(
...
+	}
+
 	transfer->incr = false;
 	xwm_selection_transfer_destroy_property_reply(transfer);
 	xwm_selection_transfer_remove_event_source(transfer);
 	xwm_selection_transfer_close_wl_client_fd(transfer);
+
+	if (transfer->window) {
+		xcb_destroy_window(transfer->selection->xwm->xcb_conn, transfer->window);
+		transfer->window = XCB_WINDOW_NONE;
```
emersion, line 86: Nit: no need to reset the window, we'll free the whole struct right after

xwayland/selection/incoming.c
```diff
@@ -150,37 +149,23 @@ static void source_send(struct wlr_xwm_selection *selection,
...
+		transfer->window,
+		transfer->selection->atom,
 		mime_type_atom,
 		xwm->atoms[WL_SELECTION],
 		XCB_TIME_CURRENT_TIME);
 
 	xcb_flush(xwm->xcb_conn);
 
 	fcntl(fd, F_SETFL, O_WRONLY | O_NONBLOCK);
+	assert(transfer->wl_client_fd < 0);
```
emersion, line 162: Nice assert!

xwayland/selection/incoming.c
```diff
@@ -395,23 +382,29 @@ void xwm_handle_selection_notify(struct wlr_xwm *xwm,
...
-		xwm_selection_transfer_finish(&selection->incoming);
+		xwm_selection_transfer_finish(transfer);
 	} else if (event->target == xwm->atoms[TARGETS]) {
 		// No xwayland surface focused, deny access to clipboard
 		if (xwm->focus_surface == NULL) {
 			wlr_log(WLR_DEBUG, "denying write access to clipboard: "
 				"no xwayland surface focused");
 			// Would leak this transfer otherwise. Should never happen.
-			assert(selection->incoming.wl_client_fd <= 0);
+			assert(transfer->wl_client_fd <= 0);
```
emersion, line 400: Hm, this should be `assert(wl_client_fd < 0)` after "xwayland/selection: introduce `xwm_selection_transfer_init`", right?

xwayland/selection/incoming.c
```diff
@@ -450,22 +444,30 @@ int xwm_handle_xfixes_selection_notify(struct wlr_xwm *xwm,
...
+		// We have to use XCB_TIME_CURRENT_TIME when we claim the selection, so
+		// grab the actual timestamp here so we can answer TIMESTAMP conversion
+		// requests correctly.
+		transfer->timestamp = event->timestamp;
 		return 1;
 	}
 
-	struct wlr_xwm_selection_transfer *transfer = &selection->incoming;
-	transfer->incr = false;
+	transfer = calloc(1, sizeof(*transfer));
```
emersion, line 457: Can we have `wlr_xwm_selection_transfer_create_incoming` maybe?

xwayland/selection/selection.c
```diff
@@ -33,15 +46,48 @@ void xwm_selection_transfer_destroy_property_reply(
 }
 
 void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer) {
+	struct wlr_xwm *xwm = transfer->selection->xwm;
+
 	transfer->wl_client_fd = -1;
+
+	transfer->window = xcb_generate_id(xwm->xcb_conn);
+	xcb_create_window(xwm->xcb_conn,
```
emersion, line 54: OK, so this is where the leak is coming from. Maybe move this out of `xwm_selection_transfer_init` for now, because it's only used by incoming transfers? That way we don't leak it for outgoing.

xwayland/selection/incoming.c
Xyene: Good catch, fixed in the other PR.

include/xwayland/selection.h
Xyene: Good idea, done.

xwayland/selection/selection.c
Xyene: Right, done.

xwayland/selection/selection.c
Xyene: Done.

xwayland/selection/outgoing.c
Xyene: Sure, I've done that. I didn't originally because I thought the outgoing path didn't do comparisons with 0, but I was wrong and it does.

xwayland/selection/selection.c
Xyene: I think just not calling `xwm_selection_transfer_init` is nicer here rather than shuffling the code back and forth once the outgoing path stops using a queue (soon™).

xwayland/selection/incoming.c
Xyene: Sure, done.

Xyene:
Argh, [the new assert](https://github.com/swaywm/wlroots/pull/2701/files#diff-3dbc47e7547e30f281b113b9b4e77f2c90ed2a47ca8de6dfcd3d09920071c7f0R162) that I thought would never fire does, in fact, fire. Will have to look into this.

Xyene:
OK, the issue was we weren't guarding against self-copies. New check fixes the issue: 

https://github.com/swaywm/wlroots/pull/2701/commits/3093149be74544c25e741e72b90f2c8338617942#diff-3dbc47e7547e30f281b113b9b4e77f2c90ed2a47ca8de6dfcd3d09920071c7f0R422-R428

Xyene:
There were some fundamental issues with the first round of this patch: `wlr_data_source.send` may be called _multiple_ times for the same source, just like `wl_data_offer.receive` can be called multiple times. The transfer window was being destroyed after the first send completed, so a second send would fail.

Now:

* Selection offer mimetypes are queried upon xfixes selection notify, on the **selection** window. There is no race here, but there might be an issue with the primary and clipboard selections sharing the same window, but using the same target. TBD.
* A new transfer window is created on each `source_send` call.

xwayland/selection/selection.c
```diff
@@ -32,16 +45,63 @@ void xwm_selection_transfer_destroy_property_reply(
...
+		0,
+		XCB_WINDOW_CLASS_INPUT_OUTPUT,
+		xwm->screen->root_visual,
+		XCB_CW_EVENT_MASK, (uint32_t[]){
+			XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_PROPERTY_CHANGE
+		}
+	);
+
+	xcb_set_selection_owner(xwm->xcb_conn, transfer->window,
+		xwm->atoms[CLIPBOARD_MANAGER], XCB_TIME_CURRENT_TIME);
```
emersion, line 72: Hm, this doesn't seem right. We should only do this for the outgoing transfer window?

xwayland/selection/selection.c
```diff
@@ -32,16 +45,63 @@ void xwm_selection_transfer_destroy_property_reply(
...
+	if (!transfer) {
+		return;
+	}
+
 	xwm_selection_transfer_destroy_property_reply(transfer);
 	xwm_selection_transfer_remove_event_source(transfer);
 	xwm_selection_transfer_close_wl_client_fd(transfer);
+
+	if (transfer->window) {
+		xcb_destroy_window(transfer->selection->xwm->xcb_conn, transfer->window);
```
emersion, line 100: In theory we'd need to flush here. Maybe it's worth it to let the other side know we're not interested in the clipboard content anymore in case of a write error?

xwayland/selection/incoming.c
```diff
@@ -450,22 +444,30 @@ int xwm_handle_xfixes_selection_notify(struct wlr_xwm *xwm,
 
 	selection->owner = event->owner;
 
-	// We have to use XCB_TIME_CURRENT_TIME when we claim the
-	// selection, so grab the actual timestamp here so we can
-	// answer TIMESTAMP conversion requests correctly.
+	if (selection->window == event->owner) {
```
emersion, line 447: This new `if` seems to conflict with the following `if`.

xwayland/selection/selection.c
emersion: It seems like the plan has changed, maybe we should reconsider this?

xwayland/selection/selection.c
Xyene: Right, this particular copy-paste operation predates my knowing of what `CLIPBOARD_MANAGER` is :)

xwayland/selection/incoming.c
Xyene: It does, and I thought I pushed a fix to remove it, but it looks like I didn't.

xwayland/selection/selection.c
```diff
@@ -51,12 +107,12 @@ bool xwm_selection_transfer_get_selection_property(
 	xcb_get_property_cookie_t cookie = xcb_get_property(
 		xwm->xcb_conn,
 		delete,
-		transfer->selection->window,
+		transfer->window,
```
emersion, line 110: Doesn't this break outgoing?

xwayland/selection/selection.c
```diff
@@ -32,16 +45,59 @@ void xwm_selection_transfer_destroy_property_reply(
...
 
-void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer) {
+void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer,
+		struct wlr_xwm_selection *selection) {
+	transfer->selection = selection;
 	transfer->wl_client_fd = -1;
 }
 
-void xwm_selection_transfer_finish(
+struct wlr_xwm_selection_transfer *xwm_selection_transfer_create_incoming(
```
emersion, line 54: Maybe we should consider moving this to `incoming.c`.

xwayland/selection/selection.c
```diff
@@ -10,6 +10,19 @@
 #include "xwayland/selection.h"
 #include "xwayland/xwm.h"
 
+struct wlr_xwm_selection_transfer *
+xwm_selection_find_incoming_transfer_by_window(
```
emersion, line 14: Ditto

include/xwayland/selection.h
```diff
@@ -20,35 +20,45 @@ struct wlr_xwm_selection_transfer {
...
+	struct wl_list link;
 
 	// when sending to x11
 	xcb_selection_request_event_t request;
-	struct wl_list outgoing_link;
 
 	// when receiving from x11
 	int property_start;
 	xcb_get_property_reply_t *property_reply;
+	xcb_window_t window;
```
emersion, line 31: Maybe we should consider renaming this to `incoming_window`, to make it clear it's unused for outgoing.

xwayland/selection/selection.c
```diff
@@ -32,16 +45,59 @@ void xwm_selection_transfer_destroy_property_reply(
 	transfer->property_reply = NULL;
 }
 
-void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer) {
+void xwm_selection_transfer_init(struct wlr_xwm_selection_transfer *transfer,
```
emersion, line 48: Idea for a future patch: make this perform the allocation and rename to `xwm_selection_transfer_create`.

xwayland/selection/selection.c
Xyene: Done.

xwayland/selection/selection.c
Xyene: Done.

xwayland/selection/selection.c
Xyene: No, because this function is only used for incoming transfers. That's confusing though, so I've moved it to `incoming.c` and renamed it appropriately.

include/xwayland/selection.h
Xyene: Sure, done.

xwayland/selection/incoming.c
```diff
@@ -11,13 +11,82 @@
 #include "xwayland/selection.h"
 #include "xwayland/xwm.h"
 
+static struct wlr_xwm_selection_transfer *
+xwm_selection_transfer_create_incoming(struct wlr_xwm_selection *selection) {
```
emersion, line 15: Hm, don't we need to call `xwm_selection_transfer_init` here?

emersion:
It seems like this breaks rich text from xwayland to wayland.

1. Open Chromium
2. Copy some rich text with bold and links (e.g. wikipedia article)
3. Try to paste into lowriter, content is plaintext

`xclip -selection clipboard -o -t TARGETS` doesn't include `text/html` with this patch, but does include it on master.

emersion (approved):
LGTM and seems to work fine in my testing, thanks!

PR #2702, opened 2021-01-31
render/egl: add check for EGL_KHR_surfaceless_context

berylline:
As surfaces are no longer going to be used for `wlr_egl`, I may as well just go and add this check as it is needed for safety whenever surface-less rendering is being used.

EDIT: I think this is also relevant: #1352


emersion (approved):
Thanks!

berylline:
You're welcome! :)

PR #2704, opened 2021-02-01
backend/drm: take detached output state in commit functions

emersion:
Don't rely on wlr_output.pending, instead take a separate
wlr_output_state argument. This is desirable because the
DRM backend sometimes needs to perform commits under-the-hood,
for instance after switching VTs. We'd like to unify and
re-use our wlr_output_commit logic.


emersion:
Superseded by https://github.com/swaywm/wlroots/pull/2828.

PR #2707, opened 2021-02-02
xwayland/selection: allow simultaneous Wayland-to-X11 transfers

Xyene:
There seems to be no reason why we can't service multiple Wayland-to-X11
transfers concurrently, so long as they are to different windows (or
possibly, same windows but different target properties?)

This commit removes the queuing logic, but retains the request
de-duplication from #2428.

### DIFF
--- a/xwayland/selection/outgoing.c
+++ b/xwayland/selection/outgoing.c
@@ -55,29 +55,10 @@ static int xwm_selection_flush_source_data(
 static void xwm_selection_transfer_start_outgoing(
 		struct wlr_xwm_selection_transfer *transfer);
 
-static struct wlr_xwm_selection_transfer *xwm_selection_transfer_get_first(
-		struct wlr_xwm_selection *selection) {
-	struct wlr_xwm_selection_transfer *first = NULL;
-	if (!wl_list_empty(&selection->outgoing)) {
-		first = wl_container_of(selection->outgoing.prev, first, link);
-	}
-
-	return first;
-}
-
 void xwm_selection_transfer_destroy_outgoing(
 		struct wlr_xwm_selection_transfer *transfer) {
-	struct wlr_xwm_selection *selection = transfer->selection;
-	bool was_first = transfer == xwm_selection_transfer_get_first(selection);
 	wl_list_remove(&transfer->link);
 	wlr_log(WLR_DEBUG, "Destroying transfer %p", transfer);
-
-	// Start next queued transfer if we just removed the active one.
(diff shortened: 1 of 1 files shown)


xwayland/selection/outgoing.c
```diff
@@ -331,19 +311,17 @@ static bool xwm_selection_send_data(struct wlr_xwm_selection *selection,
...
 	}
 
-	wl_list_insert(&selection->outgoing, &transfer->link);
-
-	// We can only handle one transfer at a time
-	if (wl_list_length(&selection->outgoing) == 1) {
-		wlr_log(WLR_DEBUG, "No transfer active, starting %p now", transfer);
-		xwm_selection_transfer_start_outgoing(transfer);
-	} else {
+	if (wl_list_length(&selection->outgoing) > 0) {
```
emersion, line 314: Nit: this `if` has no purpose.

xwayland/selection/outgoing.c
```diff
@@ -331,19 +311,17 @@ static bool xwm_selection_send_data(struct wlr_xwm_selection *selection,
...
 		struct wlr_xwm_selection_transfer *outgoing;
 		wl_list_for_each(outgoing, &selection->outgoing, link) {
-			wlr_log(WLR_DEBUG, "Transfer %p still queued", outgoing);
+			wlr_log(WLR_DEBUG, "Transfer %p still running", outgoing);
```
emersion, line 317: Can we check that the window and prop doesn't match the new transfer's? If so, at least log something? (Not sure what else we can do, maybe destroy the old transfer?)

Xyene: We do this here: https://github.com/swaywm/wlroots/pull/2707/files#diff-d54fef3e474a3877c9746a13c616758f56d9d15e7c40176aa95fc1f8c2f1cf98R305 and destroy the old transfer.

We only check the window and not the prop. I think multiple props on the same window should be OK, but as "prior art", KDE only filters by window and not by prop.

emersion: Hm, sorry, this link is broken. Can you provide one which points to the master branch?

emersion: For posterity: https://github.com/swaywm/wlroots/blob/2fa257313ac888f8b61fba8e7c457eac9e1973c3/xwayland/selection/outgoing.c#L325-L332

emersion (approved):
LGTM, thanks!

PR #2708, opened 2021-02-02
Add a stub wl_drm implementation

emersion:
This will allow us to remove all of our EGL wl_drm support code
and remove some weird stuff we need just for wl_drm support. ~~For
instance the only reason why our wl_compositor implementation needs
a wlr_renderer is to query the size of wl_drm buffers.~~


emersion:
This PR will break on split render/display devices, because of https://gitlab.freedesktop.org/mesa/mesa/-/issues/4178. I don't really want to advertise a primary node via `wl_drm` because we'd need to handle legacy `authenticate` requests.

> For instance the only reason why our wl_compositor implementation needs
a wlr_renderer is to query the size of wl_drm buffers.

Not quite, we also use it to create a texture on surface commit. So can't get rid of this part now, although I'd like to eventually get rid of it.

emersion:
> I don't really want to advertise a primary node via wl_drm because we'd need to handle legacy authenticate requests.

Well, the primary node can't be used for rendering on these platforms anyways, so DRM authentication shouldn't matter. Added a fallback to the DRM primary node.

emersion:
Another blocker for this is Mesa's Wayland WSI, which currently prefers to use `wl_drm` over linux-dmabuf.

emersion:
Pushed a new set of commits to add support for `wl_drm` buffers.

emersion:
This is now ready for review.

types/wlr_drm.c
```diff
@@ -36,9 +102,39 @@ static void drm_handle_create_prime_buffer(struct wl_client *client,
...
+
+	struct wlr_drm_buffer *buffer = calloc(1, sizeof(*buffer));
+	if (buffer == NULL) {
+		close(fd);
+		wl_resource_post_no_memory(resource);
+		return;
+	}
+	wlr_buffer_init(&buffer->base, &buffer_impl, width, height);
+
+	buffer->resource = wl_resource_create(client, &wl_buffer_interface, 1, id);
```
bl4ckb0ne, line 124: Should'nt `WLR_DRM_VERSION` be used instead of `1` to create the resource?

emersion: No, `wl_buffer` is an exception here: since it's created by many different parent interfaces (`wl_shm`, linux-dmabuf, etc), it's stuck at version 1 and cannot be extended.

https://gitlab.freedesktop.org/wayland/wayland/-/blob/92038fa39497f0a481b38e623bd604feb6a01882/protocol/wayland.xml#L431

But that's a good reflex :)

bl4ckb0ne:
LGTM, good riddance :D

bl4ckb0ne (approved):

PR #2709, opened 2021-02-02
backend/wayland/output: silence warning in NDEBUG build.

ericonr:
output isn't used anywhere but the assert. Leave below "// nothing else?"
comment, to remind anyone who eventually adds more stuff to remove the
line, if it becomes no longer necessary.


emersion:
I'm not a fan of this patch.

ericonr:
I have switched to building with `-Dwerror=false`, so for me it's ok to drop it. At least, I don't see another way of silencing the warning.

emersion:
OK, I'd prefer not to merge this workaround then.

PR #2711, opened 2021-02-03
xwayland/selection: use one target window per selection

Xyene:
As before, this PR is based on https://github.com/swaywm/wlroots/pull/2701 and only the last commit contains changes.

---

Previously, the clipboard and primary selections shared the same window.
This was racey, and could have led to pasting failures.

On xfixes selection owner change notification, the logic for requesting
the supported mimetypes of the new owner's selection looks like:

```c
xcb_convert_selection(
  xwm->xcb_conn,
  selection->window,
  selection->atom,
  xwm->atoms[TARGETS],
  xwm->atoms[WL_SELECTION],
  selection->timestamp
);
```

This means: ask the selection owner to write its TARGETS for the
`selection->atom` selection (one of PRIMARY, CLIPBOARD, DND_SELECTION)
to `selection->window`'s WL_SELECTION atom.

However, `selection->window` is shared for both PRIMARY and CLIPBOARD
selections, and WL_SELECTION is used as the target atom in both cases.
So, there's a race when both selections change at the same time.

The CLIPBOARD selection might support mimetypes {A, B, C}, and the
PRIMARY only {A, B}. If the ConvertSelection requests/responses "cross
on the wire", so to speak, wlroots can end up believing that the PRIMARY
selection also supports C.

A Wayland client may then ask for the PRIMARY selection in C format,
which will fail with "convert selection failed". The Wayland client would
end up with a failed paste operation.

This commit fixes this by using a separate window for PRIMARY and
CLIPBOARD target requests, so that WL_SELECTION can be used as the
target atom in both cases.

### DIFF
--- a/xwayland/selection/selection.c
+++ b/xwayland/selection/selection.c
@@ -174,11 +174,64 @@ int xwm_handle_selection_event(struct wlr_xwm *xwm,
 
 void xwm_selection_init(struct wlr_xwm_selection *selection,
 		struct wlr_xwm *xwm, xcb_atom_t atom) {
-	selection->xwm = xwm;
-	selection->atom = atom;
-	selection->window = xwm->selection_window;
 	wl_list_init(&selection->incoming);
 	wl_list_init(&selection->outgoing);
+
+	selection->xwm = xwm;
+	selection->atom = atom;
+	selection->window = xcb_generate_id(xwm->xcb_conn);
+
+	if (atom == xwm->atoms[DND_SELECTION]) {
+		xcb_create_window(
+			xwm->xcb_conn,
+			XCB_COPY_FROM_PARENT,
+			selection->window,
+			xwm->screen->root,
+			0, 0,
+			8192, 8192,
+			0,
+			XCB_WINDOW_CLASS_INPUT_ONLY,
+			xwm->screen->root_visual,
+			XCB_CW_EVENT_MASK,
+			(uint32_t[]){
+				XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_PROPERTY_CHANGE
+			}
+		);
+
+		xcb_change_property(
+			xwm->xcb_conn,
+			XCB_PROP_MODE_REPLACE,
+			selection->window,
+			xwm->atoms[DND_AWARE],
+			XCB_ATOM_ATOM,
+			32, // format
+			1,
+			&(uint32_t){XDND_VERSION}
+		);
+	} else {
+		xcb_create_window(
+			xwm->xcb_conn,
+			XCB_COPY_FROM_PARENT,
+			selection->window,
+			xwm->screen->root,
+			0, 0,
+			10, 10,
+			0,
+			XCB_WINDOW_CLASS_INPUT_OUTPUT,
+			xwm->screen->root_visual,
+			XCB_CW_EVENT_MASK,
+			(uint32_t[]){
+				XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_PROPERTY_CHANGE
+			}
+		);
+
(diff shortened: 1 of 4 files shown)


xwayland/selection/selection.c
```diff
@@ -246,12 +246,65 @@ int xwm_handle_selection_event(struct wlr_xwm *xwm,
...
+			XCB_WINDOW_CLASS_INPUT_OUTPUT,
+			xwm->screen->root_visual,
+			XCB_CW_EVENT_MASK,
+			(uint32_t[]){
+				XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_PROPERTY_CHANGE
+			}
+		);
+
+		if (atom == xwm->atoms[CLIPBOARD]) {
+			xcb_set_selection_owner(xwm->xcb_conn, selection->window,
```
Xyene, line 301: This is copied from the pre-existing code, but as an aside: I'm not sure if this actually makes sense? 

This marks wlroots' proxy window... as a _clipboard manager for X11_?

https://www.freedesktop.org/wiki/ClipboardManager/

This seems to come with a bunch of responsibilities, like responding to `SAVE_TARGETS`  and `TARGET_SIZES` requests, none of which wlroots appears to currently do anyway.

emersion: Yeah, I'm not sure. This is coming from Weston: https://gitlab.freedesktop.org/wayland/weston/-/commit/cba022af3d548

However the commit message doesn't explain why this has been added. Maybe it's to avoid having an X11 clipboard manager running, because it wouldn't interact properly with Wayland clients?

> This seems to come with a bunch of responsibilities, like responding to SAVE_TARGETS and TARGET_SIZES requests, none of which wlroots appears to currently do anyway.

That's no big deal, these are optional per the linked spec.

Xyene: > However the commit message doesn't explain why this has been added.

Weston also has a bit more code, which I can't find in wlroots:

https://gitlab.freedesktop.org/wayland/weston/-/commit/cba022af3d548#853f96e105430271638bdfb9a6cd981030f660a3_524_524

> Maybe it's to avoid having an X11 clipboard manager running, because it wouldn't interact properly with Wayland clients?

Hmm, maybe... I don't see why an X11 clipboard manager would break things hard enough to be explicitly disallowed (or that this even disallows one &mdash; I imagine another app can claim the `CLIPBOARD_SELECTION`?)... but I guess we can only speculate at the original decision at this point.

> That's no big deal, these are optional per the linked spec.

Right, so it's not an (obvious) problem that we grab `CLIPBOARD_MANAGER` and do nothing with it, just odd.

emersion: FWIW, I'd be fine with dropping it and see if something breaks.

emersion:
The second commit LGTM.

Xyene:
OK, so I though this was just a theoretical issue that didn't necessarily come up in practice, but I ended up being able to easily reproduce the bug in while testing https://github.com/swaywm/wlroots/pull/2701. So that's cool, I guess.

Test setup:

- one X11 app with the following rich text

    **test** _test_ test

    For posterity, I used Slack, but Chromium selecting the above text is probably sufficient.

- open up epiphany-browser to https://quilljs.com/playground/
- select the test text in the X11 app, and paste into the Quill playground

Half the time it pastes as "test test test", and the other times as "**test** _test_ test". (Note the source selection region needs to be adjusted after each paste attempt in order to get new selection notification events, and thus trigger the race.) 

After this patch, it reliably gets pasted as rich text.

emersion:
Ah, this might very well be the same bug as https://github.com/swaywm/wlroots/issues/2716!

Xyene:
Sounds like it could be :)

I've rebased this branch off master now.

emersion (approved):
LGTM, thanks!

PR #2712, opened 2021-02-03
backend/drm: add support for hotplug_mode_update

hedgepigdaniel:
Fixes https://github.com/swaywm/wlroots/issues/2188
Depends on https://github.com/swaywm/wlroots/pull/2693
Replaces https://github.com/swaywm/wlroots/pull/2628

The first commit adds basic support for the `hotplug_mode_update` DRM prop. It ignores the fixed modes, and treats the preferred mode as a custom mode. This allows the compositor to automatically modeset when the parent window is resized, but as the window is resized, custom modes are added but not removed, even though they are no longer in the list of DRM modes.

So in the second commit I did what @emersion suggested in https://github.com/swaywm/wlroots/pull/2628#issuecomment-767431813 - not populate the `output->modes` list, and behave more like the X11/Wayland backends in this case. Specifically:
 - when `hotplug_mode_update` is set, the fixed modes are ignored and no modes are ever added to `output->modes`
 - there is a new output mode type `WLR_OUTPUT_STATE_MODE_HOT_PLUGGABLE` alongside `WLR_OUTPUT_STATE_MODE_FIXED` and `WLR_OUTPUT_STATE_MODE_CUSTOM`
 - hot pluggable modes are not assigned to `struct wlr_output_state`. Instead compositors may use `wlr_output_set_hot_pluggable_mode()` to use the current preferred mode on a hot pluggable output, and the backend may request a state with `mode_type` `WLR_OUTPUT_STATE_MODE_HOT_PLUGGABLE`.
 - when an output state is committed with `mode_type` `WLR_OUTPUT_STATE_MODE_HOT_PLUGGABLE`, the current preferred mode is read from DRM and applied


emersion:
Do we really need that new `WLR_OUTPUT_STATE_MODE_HOT_PLUGGABLE` mode type? Why can't we just use `WLR_OUTPUT_STATE_MODE_CUSTOM` like the Wayland/X11 backends?

hedgepigdaniel:
Ok, I think I've managed to simplify it including not introducing a new mode type.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2712

PR #2713, opened 2021-02-04
examples/egl_common: make attribs const

berylline:
Make `(config, context)_attribs` `const` just to be on the safe side.

I wanted to do this in the last PR that I made for `egl_common.c`, but then it got merged, so I thought "forget about it", but now I've made this PR, so... Yeah.


emersion (approved):
Thanks

PR #2714, opened 2021-02-04
render/egl: use strstr instead of check_egl_ext

berylline:
Use `strstr` instead of `check_egl_ext` since the former already exists to do the job of the latter for searching for EGL extension strings.

<s>Not sure why `check_egl_ext` exists when `strstr` does the same thing.</s>

EDIT:

`strstr` is similar in functionality to `check_egl_ext` which is why I replaced `check_egl_ext` with `strstr` in this PR. Sorry for not clarifying this at first.

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -4,6 +4,7 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <unistd.h>
 #include <gbm.h>
 #include <wlr/render/egl.h>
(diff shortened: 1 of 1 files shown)


berylline:
> > Not sure why `check_egl_ext` exists when `strstr` does the same thing.
> 
> It doesn't do the same thing. If the reported extensions are "EGL_foo_bar EGL_baz" and you search for extension "EGL_foo" with `strstr`, you'll mistakenly think you have it available when you don't.

I don't think what you're saying is correct because my impression about how EGL extension strings work is that if the extension isn't available, it wouldn't be in the string and I'm pretty sure that's how the extension strings work.

And if a driver is claiming to support an EGL extension that it doesn't actually support, then I think that would be a flaw with the driver.

Unless I misunderstood your point and you're saying that `strstr` only searches for partial matches, which would be incorrect also because `strstr` searches for the whole string that has been specified in the second parameter, not just one part of it.

Also, I think what I should've clarified is that `strstr` returns a pointer to the extension string(as in, it doesn't return `NULL`) when the string is found and `strstr` returns `NULL` when it doesn't exist. Also, my reasoning for this PR is that AFAICT, they're both equivalent in functionality, which is why I replaced `check_egl_ext` with `strstr`.

ascent12:
There are extensions that are literally prefixes of other extensions (`EGL_EXT_image_dma_buf_import`->`EGL_EXT_image_dma_buf_import_modifiers`, `EGL_KHR_image`->`EGL_KHR_image_base`) so anything strstr-based would not guarantee that you found the exact string you're looking for, and doing something like adding the separator (a space) to the search string doesn't work either, because it may not appear on the last extension advertised.

This function actually was strstr-based when I first wrote it, and was changed to this after.

emersion:
Yeah, `strstr` doesn't work here.

berylline:
Then I stand corrected. I'll delete the branch of this PR later.

@emersion Although I was going to do close this PR, thanks for closing this PR for me. It is well appreciated.

PR #2717, opened 2021-02-04
xwayland: check executable exists on init, take 2

emersion:
### xwayland: add dependency on xwayland

Check that the pkg-config file is available. This will be required
in the future to check whether xwayland supports features such as
-listenfd, -initfd or -verbose.

This effectively makes our relationship with xwayland closer to what
a dynamic library is: checked at build-time, but can be overridden
at run-time.

### xwayland: check executable exists on init

Instead of walking PATH like a previous proposal (#2314), this one
checks that the Xwayland path specified in the pkg-config file
exists.

I think this is a reasonable compromise:

- Users that don't have Xwayland installed system-wide won't get
  a bogus DISPLAY env variable set up.
- Users that have WLR_XWAYLAND set won't be affected by this check.
- Users that have Xwayland installed system-wide and a different
  Xwayland in their PATH still get their custom Xwayland.
- Users that don't have Xwayland installed system-wide but have it
  somewhere else in PATH are left out. But this is pretty niche,
  and they can just set WLR_XWAYLAND.

### DIFF
--- a/xwayland/meson.build
+++ b/xwayland/meson.build
@@ -18,6 +18,19 @@ if not get_option('xwayland').disabled()
 	msg += 'Required for Xwayland support.'
 endif
 
+xwayland = dependency('xwayland', required: false)
+if not xwayland.found()
+	# There's no Xwayland release with the pkg-config file shipped yet.
+	xwayland_prog = find_program('Xwayland', required: false)
+	if not xwayland_prog.found()
+		if get_option('xwayland').enabled()
+			error('\n'.join(msg).format('xwayland'))
+		else
+			subdir_done()
+		endif
+	endif
+endif
+
 foreach lib : xwayland_required
 	dep = dependency(lib,
 		required: get_option('xwayland'),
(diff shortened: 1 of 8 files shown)


xwayland/meson.build
```diff
@@ -1,5 +1,6 @@
 xwayland_libs = []
 xwayland_required = [
+	'xwayland',
```
Xyene, line 3: Small nit: the rest of the deps are lexicographically sorted.

Xyene (commented):
The overall approach LGTM, but CI is currently failing. Is this expected?

emersion:
Ah, CI was still missing the new xwayland dep. Fixed that, and your comment!

emersion:
Hm, it seems like the xwayland pkg-config file hasn't been released yet. Will need to wait for the next xwayland release.

emersion:
Added a fallback to look for the Xwayland executable if we don't find a pkg-config file. This should accommodate for distributions that don't ship Xwayland HEAD.

PR #2718, opened 2021-02-04
xdg-activation-v1: new protocol implementation

emersion:
See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/50

### DIFF
--- a/types/wlr_xdg_activation_v1.c
+++ b/types/wlr_xdg_activation_v1.c
@@ -1,0 +1,327 @@
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wlr/types/wlr_seat.h>
+#include <wlr/types/wlr_surface.h>
+#include <wlr/types/wlr_xdg_activation_v1.h>
+#include <wlr/util/log.h>
+#include "util/signal.h"
+#include "util/token.h"
+#include "xdg-activation-v1-protocol.h"
+
+#define XDG_ACTIVATION_V1_VERSION 1
+
+static const struct xdg_activation_token_v1_interface token_impl;
+
+static struct wlr_xdg_activation_token_v1 *token_from_resource(
+		struct wl_resource *resource) {
+	assert(wl_resource_instance_of(resource,
+		&xdg_activation_token_v1_interface, &token_impl));
+	return wl_resource_get_user_data(resource);
+}
+
+static void token_destroy(struct wlr_xdg_activation_token_v1 *token) {
+	if (token == NULL) {
+		return;
+	}
+	if (token->resource != NULL) {
+		wl_resource_set_user_data(token->resource, NULL); // make inert
+	}
+	wl_list_remove(&token->link);
+	wl_list_remove(&token->seat_destroy.link);
+	wl_list_remove(&token->surface_destroy.link);
+	free(token->app_id);
+	free(token->token);
+	free(token);
+}
+
+static void token_handle_resource_destroy(struct wl_resource *resource) {
+	struct wlr_xdg_activation_token_v1 *token = token_from_resource(resource);
+	token_destroy(token);
+}
+
+static void token_handle_destroy(struct wl_client *client,
+		struct wl_resource *token_resource) {
+	wl_resource_destroy(token_resource);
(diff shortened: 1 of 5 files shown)


dnkl:
Getting a crash when I set a surface on the token, but nothing else:
```c
struct xdg_activation_token_v1 *token = xdg_activation_v1_get_activation_token(term->wl->xdg_activation);
xassert(token != NULL);
xdg_activation_token_v1_add_listener(token, &activation_token_listener, term);
xdg_activation_token_v1_set_surface(token, term->window->surface);
xdg_activation_token_v1_commit(token);
```
Backtrace:
```
(gdb) bt full
#0  token_handle_commit (client=0x55555636c620, token_resource=0x555556397870)
    at ../../subprojects/wlroots/types/wlr_xdg_activation_v1.c:75
        token = 0x555556392770
        token_str = "4a0e176c7b01ce178bbaa2da310b0323\000"
        __PRETTY_FUNCTION__ = "token_handle_commit"
#1  0x00007ffff6f90acd in  () at /usr/lib/libffi.so.7
```
```c
	if (token->surface != token->seat->keyboard_state.focused_surface) {
		wlr_log(WLR_DEBUG, "Rejecting token commit request: "
			"surface doesn't have keyboard focus");
		goto error;
	}
```

The reason I did this was because without setting a surface, I got:
```
00:00:11.373 [DEBUG] [wlr] [types/wlr_xdg_activation_v1.c:71] Rejecting token commit request: surface has been destroyed
```
I was under the impression setting the surface was optional? What am I missing?
```
/**
 * @ingroup iface_xdg_activation_token_v1
 *
 * The requesting client can specify a surface to associate the token
 * being created with it.
 *
 * Must be triggered before commit. This information is optional.
 */
```

dnkl:
Thanks, works for me as well now. I've updated foot's xdg-activation branch to set the token surface to the top-level surface, and everything appears to (still) be working as expected.

emersion:
Compositors should probably:

1. Invalidate all tokens on (keyboard?) focus change, assuming the user has moved on to something else and focus shouldn't be transferred. Except it may still be desirable to retain tokens to mark toplevels as urgent.
2. Immediately destroy new tokens for serials older than the current focus. With the same caveats as above.

djpohly:
I'm trying to implement urgency in dwl using foot as a testcase, but I'm getting a null dereference in `wlr_xdg_activation_v1.c` line 75:
```c
if (token->surface != token->seat->keyboard_state.focused_surface) {
```
with `token->seat` being NULL.  A few lines later, the code does check to see `if (token->seat != NULL)` - maybe a few lines need to be reordered?

dnkl:
It looks kind of like it (the xdg-activation) branch has been reverted to the original version, where Sway also crashed.

https://github.com/swaywm/wlroots/commit/677d6a28e1660c240940ea9253dc99a0d1c5408d is the last version I tested (and found working) with foot. It looks vastly different.

emersion:
Hm, sorry, I probably messed up a rebase. I restored the branch.

emersion:
Added a token timeout to avoid these dangling token issues.

types/wlr_xdg_activation_v1.c
```diff
@@ -0,0 +1,352 @@
...
+			"The activation token has already been used");
+		return;
+	}
+
+	token->surface = surface;
+
+	token->surface_destroy.notify = token_handle_surface_destroy;
+	wl_list_remove(&token->surface_destroy.link);
+	wl_signal_add(&surface->events.destroy, &token->surface_destroy);
+
```
kennylevinsen, line 210: nit: newline

types/wlr_xdg_activation_v1.c
```diff
@@ -0,0 +1,352 @@
...
+static const struct xdg_activation_token_v1_interface token_impl = {
+	.destroy = token_handle_destroy,
+	.commit = token_handle_commit,
+	.set_app_id = token_handle_set_app_id,
+	.set_serial = token_handle_set_serial,
+	.set_surface = token_handle_set_surface,
+};
+
+
```
kennylevinsen, line 221: nit: newline

kennylevinsen (approved):
LGTM.

PR #2720, opened 2021-02-05
Add static+const qualifiers

mstoeckl:
The three commits in this PR add const (and in one case, file-static) qualifiers to global variables that were missing them.
* The list of atoms in xwayland/xwm.c
* The cursor data used by the default theme
* A number of "implementation function lists", which include structs of function pointers for wlroots backends, and for handling Wayland events and requests

Since `xwayland/xwm.h` is an internal header, and `wlr/interfaces/wlr_tablet_tool.h` and `wlr/types/wlr_tablet_tool.h`  are unstable API, I presume changing types slightly is fine.

I found these variables by checking ``objdump -t libwlroots.so.7 | grep "\.data" ``for initialized global variables not linked as read only. I did not look at tinywl or any of the examples.

* * *

Breaking change: `wlr_tablet.impl` is now `const`.

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -87,7 +87,7 @@ static int backend_get_drm_fd(struct wlr_backend *backend) {
 	}
 }
 
-static struct wlr_backend_impl backend_impl = {
+static const struct wlr_backend_impl backend_impl = {
 	.start = backend_start,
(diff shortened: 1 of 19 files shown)


emersion (approved):
Thanks!

PR #2723, opened 2021-02-05
region: remove leftover description of resource_list

berylline:
`resource_list` no longer exists for regions. Remove the last remaining description of what it was responsible for.


emersion (approved):
Good catch, thanks

berylline:
> Good catch, thanks

You're welcome! :)

PR #2725, opened 2021-02-06
xdg-popup: remove code repetition

### LINKED ISSUE #2724: Code repetition leading to inconsistent popup handling logic
wlroots handles xdg_popups differently for [layer surfaces](https://github.com/swaywm/wlroots/blob/a02ac01be3e2029f6dcd5479c39ea30e6ea17594/types/wlr_layer_shell_v1.c#L560) and [xdg surfaces](https://github.com/swaywm/wlroots/blob/a02ac01be3e2029f6dcd5479c39ea30e6ea17594/types/xdg_shell/wlr_xdg_surface.c#L639).
While layer surface popups have their coordinates calculated by subtraction of the popup position and the underlying xdg surface position, xdg surface popups use [xdg_popup_get_position()](https://github.com/swaywm/wlroots/blob/a02ac01be3e2029f6dcd5479c39ea30e6ea17594/types/xdg_shell/wlr_xdg_surface.c#L571) with a slightly different logic; it also adds parent xdg surface coordinates. This difference leads to problems when working with layer surfaces' nested popups ([gtk-layer-demo](https://github.com/wmww/gtk-layer-shell), for example). For example, [this part in Sway](https://github.com/swaywm/sway/blob/master/sway/desktop/output.c#L205) is there for some reason instead of something like the following:
```c
struct sway_layer_surface *layer_surface;
wl_list_for_each(layer_surface, layer_surfaces, link) {
	struct wlr_layer_surface_v1 *wlr_layer_surface_v1 =
		layer_surface->layer_surface;
	struct wlr_surface *surface = wlr_layer_surface_v1->surface;
	struct surface_iterator_data data = {
		.user_iterator = iterator,
		.user_data = user_data,
		.output = output,
		.view = NULL,
		.ox = layer_surface->geo.x,
		.oy = layer_surface->geo.y,
		.width = surface->current.width,
		.height = surface->current.height,
		.rotation = 0,
	};
	wlr_layer_surface_v1_for_each_popup_surface(wlr_layer_surface_v1,
		output_for_each_surface_iterator, &data);
}
```

Possible solutions:
- Copy xdg_popup_get_position() to wlr_layer_shell_v1.c, with a check whether the parent surface is actually an xdg surface or not. Easy to do, but bad due to even more code duplication.
- Expose a function like xdg_popup_for_each_surface() and use that.

If the current approach is actually justified, I'd like to know why.

vyivel:
Fixes #2724.

### DIFF
--- a/types/xdg_shell/wlr_xdg_surface.c
+++ b/types/xdg_shell/wlr_xdg_surface.c
@@ -570,10 +570,12 @@ void wlr_xdg_popup_destroy(struct wlr_xdg_surface *surface) {
 
 static void xdg_popup_get_position(struct wlr_xdg_popup *popup,
 		double *popup_sx, double *popup_sy) {
-	struct wlr_xdg_surface *parent =
-		wlr_xdg_surface_from_wlr_surface(popup->parent);
-	struct wlr_box parent_geo;
-	wlr_xdg_surface_get_geometry(parent, &parent_geo);
+	struct wlr_box parent_geo = {0};
+	if (wlr_surface_is_xdg_surface(popup->parent)) {
+		struct wlr_xdg_surface *parent =
+			wlr_xdg_surface_from_wlr_surface(popup->parent);
+		wlr_xdg_surface_get_geometry(parent, &parent_geo);
+	}
 	*popup_sx = parent_geo.x + popup->geometry.x -
 		popup->base->geometry.x;
 	*popup_sy = parent_geo.y + popup->geometry.y -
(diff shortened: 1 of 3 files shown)


types/xdg_shell/wlr_xdg_popup.c
```diff
@@ -191,6 +191,11 @@ struct wlr_xdg_popup_grab *get_xdg_shell_popup_grab_from_seat(
 	return xdg_grab;
 }
 
+struct xdg_surface_iterator_data {
+	wlr_surface_iterator_func_t user_iterator;
+	void *user_data;
+	int x, y;
+};
```
ifreund, line 198: This is unused.
```suggestion
```

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -570,10 +570,12 @@ void wlr_xdg_popup_destroy(struct wlr_xdg_surface *surface) {
...
-	struct wlr_xdg_surface *parent =
-		wlr_xdg_surface_from_wlr_surface(popup->parent);
-	struct wlr_box parent_geo;
-	wlr_xdg_surface_get_geometry(parent, &parent_geo);
+	struct wlr_box parent_geo = {0};
+	if (wlr_surface_is_xdg_surface(popup->parent)) {
+		struct wlr_xdg_surface *parent =
+			wlr_xdg_surface_from_wlr_surface(popup->parent);
+		wlr_xdg_surface_get_geometry(parent, &parent_geo);
+	}
```
ifreund, line 578: Perhaps I've misunderstood the issue, but don't we need to use the parent coordinates if the parent is a layer surface as well? Currently I don't see how this PR actually changes the result.

ifreund (changes requested):
Thanks for working on this! Getting rid of this code duplication is definitely a step in the right direction.

types/xdg_shell/wlr_xdg_surface.c
vyivel: No, layer surfaces don't have "geometry" in the same sense as xdg surfaces do. For example, let's say we have this:
- A
  - Popup 1
    - Popup 2
 If A is an xdg surface, then popup 2 would have its coordinates + popup 1 coordinates + A coordinates.
 If A is an layer surface, then popup 2 would have only its coordinates. Coordinates of A don't matter in this case, as layer surfaces don't have a position. Howerer, popup 1 *has* a position, and it must be added to the position of popup 2, which is what this PR does.

types/xdg_shell/wlr_xdg_popup.c
vyivel: Removed.

vyivel:
Uh, nevermind all that. I was testing with wlroots 0.12.0. Turns out https://github.com/swaywm/wlroots/commit/b482c90e1ac3f04236eaf3b00eb28a58823494aa already fixes it. Sorry for bothering.

types/xdg_shell/wlr_xdg_surface.c
vyivel: Actually I think this explanation is incorrect. Doesn't matter now anyway.

PR #2727, opened 2021-02-06
backend/x11: clamp hotspot to texture bounds

imirkin:
When a new texture is set, the hotspot may actually belong to the
previous texture and be out of bounds. Rather than incur X errors for
these, clamp the hotspot to be inside of the texture.

This fixes weston examples updating their cursors (e.g.
weston-eventdemo).

### DIFF
--- a/backend/x11/output.c
+++ b/backend/x11/output.c
@@ -475,6 +475,19 @@ static bool output_set_cursor(struct wlr_output *wlr_output,
 	if (texture != NULL) {
 		width = texture->width * wlr_output->scale / scale;
 		height = texture->height * wlr_output->scale / scale;
+
+		if (hotspot_x < 0) {
+			hotspot_x = 0;
+		}
+		if ((uint32_t)hotspot_x > texture->width) {
+			hotspot_x = texture->width;
+		}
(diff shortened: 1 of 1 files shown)


backend/x11/output.c
```diff
@@ -475,6 +475,19 @@ static bool output_set_cursor(struct wlr_output *wlr_output,
 	if (texture != NULL) {
 		width = texture->width * wlr_output->scale / scale;
 		height = texture->height * wlr_output->scale / scale;
+
+		if (hotspot_x < 0) {
```
emersion, line 479: Shouldn't we do this after transforming the hotspot?

imirkin: You tell me :)

The logic I was going under was that hotspot_x/y are in the texture coordinate space, so it's fine to clamp them to texture->width/height. It wasn't completely obvious to me what the limits of the post-transform hotspot would be -- I guess width/height?

emersion: Looking at the code again, I think you're right and the hotspot is in texture coords.

emersion (approved):
Thanks!

PR #2732, opened 2021-02-08
render: make gles2 renderer optional

emersion:
Some compositors may not need the GLES2 renderer, because they provide
their own. Allow to build wlroots without it.

~~Depends on https://github.com/swaywm/wlroots/pull/2733~~

### DIFF
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -2,8 +2,8 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <gbm.h>
+#include <wlr/config.h>
 #include <wlr/render/egl.h>
-#include <wlr/render/gles2.h>
 #include <wlr/render/interface.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_matrix.h>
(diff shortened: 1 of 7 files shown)


render/wlr_renderer.c
```diff
@@ -251,6 +255,7 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 }
 
 struct wlr_renderer *wlr_renderer_autocreate_with_drm_fd(int drm_fd) {
+#if WLR_HAS_GLES2_RENDERER
```
Emantor, line 258: Maybe we should import `IS_ENABLED` from the linux kernel.

emersion: Not a fan of magic kernel macros. In particular, `IS_ENABLED` seems to be using a lot of black magic.

Emantor: I found the explanation in [Zephyr](https://github.com/zephyrproject-rtos/zephyr/blob/89a90963343437378a462aa64431b2283b7c397a/include/sys/util_internal.h#L18) good. But it totally looks like black magic. This would also require changes to the conditional include of the header. The advantage is that all code is compiled by the compiler and bit rot of conditional code paths can be avoided.

hizukiayaka:
I think such patch is necessary but I would suggest that can choose the backend you need yourself,
Another one is the GBM, the gbm API from ARM mali or the other embedded platform that not use mesa, it won't implement all the gbm api.
I would submit a PR as soon as possbile

emersion:
@hizukiayaka's patch has been merged.

PR #2733, opened 2021-02-08
examples: remove wlroots dep from client examples

emersion:
_See individual commits._

cc @berylline


berylline (approved):
Sorry for the really late response. Everything looks good to me! LGTM!

berylline:
Another thing is that I was originally going to replace `wlr_log` with `fprintf` for #2666, but I didn't know if you wanted me to do that or not, so I didn't do that in the end, but thanks for doing that anyway!

PR #2737, opened 2021-02-09
render: introduce shared memory allocator

emersion:
It allocates in local main memory via shm_open, and provides a FD
to allow sharing with other processes.

This is suitable for software rendering under the Wayland and X11
backends.

cc @bl4ckb0ne

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -20,6 +20,7 @@
 #include "backend/wayland.h"
 #include "render/drm_format_set.h"
 #include "render/gbm_allocator.h"
+#include "render/pixel_format.h"
 #include "render/wlr_renderer.h"
 #include "util/signal.h"
 
@@ -190,6 +191,17 @@ static const struct wl_drm_listener legacy_drm_listener = {
 	.capabilities = legacy_drm_handle_capabilities,
 };
 
+static void shm_handle_format(void *data, struct wl_shm *shm,
+		uint32_t shm_format) {
+	struct wlr_wl_backend *wl = data;
+	uint32_t drm_format = convert_wl_shm_format_to_drm(shm_format);
+	wlr_drm_format_set_add(&wl->shm_formats, drm_format, DRM_FORMAT_MOD_INVALID);
+}
+
+static const struct wl_shm_listener shm_listener = {
+	.format = shm_handle_format,
+};
+
 static void registry_global(void *data, struct wl_registry *registry,
 		uint32_t name, const char *iface, uint32_t version) {
 	struct wlr_wl_backend *wl = data;
@@ -233,6 +245,9 @@ static void registry_global(void *data, struct wl_registry *registry,
 	} else if (strcmp(iface, wl_drm_interface.name) == 0) {
 		wl->legacy_drm = wl_registry_bind(registry, name, &wl_drm_interface, 1);
 		wl_drm_add_listener(wl->legacy_drm, &legacy_drm_listener, wl);
+	} else if (strcmp(iface, wl_shm_interface.name) == 0) {
+		wl->shm = wl_registry_bind(registry, name, &wl_shm_interface, 1);
+		wl_shm_add_listener(wl->shm, &shm_listener, wl);
 	}
 }
 
@@ -302,6 +317,7 @@ static void backend_destroy(struct wlr_backend *backend) {
 	wlr_allocator_destroy(wl->allocator);
 	close(wl->drm_fd);
 
+	wlr_drm_format_set_finish(&wl->shm_formats);
 	wlr_drm_format_set_finish(&wl->linux_dmabuf_v1_formats);
 
 	struct wlr_wl_buffer *buffer, *tmp_buffer;
@@ -321,6 +337,9 @@ static void backend_destroy(struct wlr_backend *backend) {
 	}
 	if (wl->zwp_linux_dmabuf_v1) {
 		zwp_linux_dmabuf_v1_destroy(wl->zwp_linux_dmabuf_v1);
+	}
+	if (wl->shm) {
+		wl_shm_destroy(wl->shm);
 	}
 	if (wl->zwp_relative_pointer_manager_v1) {
 		zwp_relative_pointer_manager_v1_destroy(wl->zwp_relative_pointer_manager_v1);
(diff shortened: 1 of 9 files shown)


render/shm_allocator.c
```diff
@@ -0,0 +1,98 @@
...
+		struct wlr_allocator *wlr_allocator, int width, int height,
+		const struct wlr_drm_format *format) {
+	struct wlr_shm_buffer *buffer = calloc(1, sizeof(*buffer));
+	if (buffer == NULL) {
+		return NULL;
+	}
+	wlr_buffer_init(&buffer->base, &buffer_impl, width, height);
+
+	// TODO: consider using a single file for multiple buffers
+	int stride = width; // TODO: align
```
bl4ckb0ne, line 55: stride should be `width * 4` as long as the bpp is 32 bits

emersion: Good catch! Fixed.

backend/wayland/backend.c
```diff
@@ -185,6 +186,28 @@ static const struct wl_drm_listener legacy_drm_listener = {
 	.capabilities = legacy_drm_handle_capabilities,
 };
 
+static uint32_t convert_wl_shm_format_to_drm(enum wl_shm_format format) {
```
emersion, line 189: Note: when https://github.com/swaywm/wlroots/pull/2744 lands, this can be replaced with the centralized helper.

render/shm_allocator.c
```diff
@@ -0,0 +1,110 @@
...
+static struct wlr_buffer *allocator_create_buffer(
+		struct wlr_allocator *wlr_allocator, int width, int height,
+		const struct wlr_drm_format *format) {
+	struct wlr_shm_buffer *buffer = calloc(1, sizeof(*buffer));
+	if (buffer == NULL) {
+		return NULL;
+	}
+	wlr_buffer_init(&buffer->base, &buffer_impl, width, height);
+
+	int bytes_per_pixel;
```
bl4ckb0ne, line 55: would it be worth to add that in #2744?

emersion: What do you mean?

bl4ckb0ne: Having a centralized place to get bits/bytes per pixel from a drm format

emersion: Ah, yeah. That might not be a bad idea, e.g. Weston has one. Should probably be done in a new PR anyways, doesn't seem related to #2744.

emersion: https://github.com/swaywm/wlroots/issues/2748

Xyene:
> It allocates in local main memory via shm_open, and provides a FD to allow sharing with other processes.

Hmm, perhaps this might be a case for [`memfd_create`](https://man7.org/linux/man-pages/man2/memfd_create.2.html) instead? I don't have all the context here, but if an fd to a shm region is being passed to untrusted processes, those processes can change the size of the memory mapping under wlroots' feet, e.g. to something smaller than wlroots expects.

`memfd_create` with `F_SEAL_GROW | F_SEAL_SHRINK` would prevent this.

emersion:
The issue is with Wayland clients shrinking the file under the compositor's feet. With this shm allocator wlroots is the client.

`memfd_create` is not POSIX though, so I won't use it. I also think something like OpenBSD's [__MAP_NOFAULT](https://github.com/openbsd/src/blob/d714ed3db694a238d1e0a5018a7fb156af9527f3/sys/sys/mman.h#L61) would make more sense, because it doesn't require each and every client to opt-in and add support for a Linux-specific API.

emersion:
Rebased against master. The PR has been updated with the latest `get_data_ptr` interface, and now uses the shared format table.

backend/wayland/backend.c
```diff
@@ -421,49 +436,69 @@ struct wlr_backend *wlr_wl_backend_create(struct wl_display *display,
...
 	}
 
+	const struct wlr_drm_format_set *remote_formats;
+	const struct wlr_drm_format_set *render_formats;
+	if (wl->drm_fd >= 0) {
+		remote_formats = &wl->linux_dmabuf_v1_formats;
+		render_formats = wlr_renderer_get_dmabuf_render_formats(wl->renderer);
+	} else {
+		remote_formats = &wl->shm_formats;
+		render_formats = NULL; // TODO
```
emersion, line 485: Still not yet sure what the best way forward is here:

1. The renderer will use `get_data_ptr` to write to the buffers. It doesn't care whether the buffers are DMA-BUFs (DRM dumb buffers) or shm.
2. The backend doesn't care how the renderer writes to the buffers. It only needs to make sure that the renderer can write to its shm or DRM dumb buffers.

Maybe we can assume renderer will never have different format constraints for DRM dumb buffers and shm. I wonder if we'll ever have a case of a renderer which is able to write to both GPU buffers and CPU buffers, with a different set of formats for each case?

Potential solutions:

- Assume renderers will only ever support one buffer type for rendering. Rename `wlr_renderer_get_dmabuf_render_formats` to `wlr_renderer_get_render_formats`. Note, this interface is a little bit too permissive, since software renderers don't have modifiers, but the allocator interface has the same issue too.
- Add `wlr_renderer_get_software_render_formats` (or `wlr_renderer_get_shm_render_formats` for consistency with existing functions?), which returns a flat list of DRM formats, just like `wlr_renderer_get_shm_texture_formats`.

bl4ckb0ne: I like the `wlr_renderer_get_shm_render_formats`, its more consistent with the existing functions. Maybe we could hide that behind an impl function and just call `wlr_renderer_get_render_formats`. Is the renderer aware if it's using dmabufs or not?

emersion: > I like the wlr_renderer_get_shm_render_formats, its more consistent with the existing functions.

If we keep it private for now, it's also the safest option, since it doesn't change the existing API. We can always revisit this when we decide to make it public.

> Is the renderer aware if it's using dmabufs or not?

The renderer can't know in advance if the allocator will produce DMA-BUFs or something else.

bl4ckb0ne: I implemented it in the pixman render PR in https://github.com/swaywm/wlroots/pull/2661/commits/090c8f6d6a7070fb31cc9aa3edf0a73ecdb68eeb, and use it in https://github.com/swaywm/wlroots/pull/2661/commits/d7571f25b09e8a3a6be932d7789996f0cfcb7d51. I chose to use the `wlr_drm_format_set` in case we merge the two functions to get the render formats.

emersion:
These patches have been merged as part of https://github.com/swaywm/wlroots/pull/2661.

PR #2739, opened 2021-02-11
xwayland/selection: fix self-copy loop 

### LINKED ISSUE #2192: Pasting in Emacs causes wlroots to run out of pipes

Xyene:
See commits for details.

Fixes #2192.

### DIFF
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -513,9 +513,14 @@ int xwm_handle_xfixes_selection_notify(struct wlr_xwm *xwm,
 		return 1;
 	}
 
+	if (selection->owner == selection->window &&
+			event->owner != selection->owner) {
+		wlr_log(WLR_DEBUG, "proxy window lost selection ownership");
+	}
+
 	selection->owner = event->owner;
 
-	if (event->owner == selection->window) {
+	if (selection->owner == selection->window) {
 		// We have to use XCB_TIME_CURRENT_TIME when we claim the selection, so
 		// grab the actual timestamp here so we can answer TIMESTAMP conversion
 		// requests correctly.
--- a/xwayland/selection/outgoing.c
+++ b/xwayland/selection/outgoing.c
@@ -392,9 +392,22 @@ void xwm_handle_selection_request(struct wlr_xwm *xwm,
 		goto fail_notify_requestor;
 	}
 
-	if (selection->window != req->owner) {
-		wlr_log(WLR_DEBUG, "received selection request with invalid owner");
+	if (req->requestor == selection->window) {
+		wlr_log(WLR_ERROR, "selection request should have been caught before");
 		goto fail_notify_requestor;
+	}
+
+	if (selection->window != req->owner) {
+		if (req->time != XCB_CURRENT_TIME && req->time < selection->timestamp) {
+			wlr_log(WLR_DEBUG, "ignored old request from timestamp %d; expected > %d",
+					req->time, selection->timestamp);
+			goto fail_notify_requestor;
+		}
+
+		wlr_log(WLR_DEBUG, "received selection request with invalid owner");
+		// Don't fail (`goto fail_notify_requestor`) the selection request if we're
+		// no longer the selection owner.
+		return;
 	}
 
 	// No xwayland surface focused, deny access to clipboard
--- a/xwayland/selection/selection.c
+++ b/xwayland/selection/selection.c
@@ -267,12 +267,14 @@ static void xwm_selection_set_owner(struct wlr_xwm_selection *selection,
 			selection->window,
 			selection->atom,
 			XCB_TIME_CURRENT_TIME);
+		xcb_flush(selection->xwm->xcb_conn);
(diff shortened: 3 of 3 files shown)


xwayland/selection/outgoing.c
```diff
@@ -392,6 +392,12 @@ void xwm_handle_selection_request(struct wlr_xwm *xwm,
 		goto fail_notify_requestor;
 	}
 
+	if (req->time <= selection->timestamp) {
+		wlr_log(WLR_DEBUG, "ignoring old request from timestamp %d; expecting > %d",
+				req->time, selection->timestamp);
+		goto fail_notify_requestor;
+	}
+
```
Xyene, line 400: From ICCCM:

> The specified owner and selection will be the values that were specified in the SetSelectionOwner request. The owner should compare the timestamp with the period it has owned the selection and, if the time is outside, refuse the SelectionRequest by sending the requestor window a SelectionNotify event with the property set to None (by means of a SendEvent request with an empty event mask).

xwayland/selection/incoming.c
```diff
@@ -492,8 +492,16 @@ int xwm_handle_xfixes_selection_notify(struct wlr_xwm *xwm,
 		return 0;
 	}
 
+	if (event->timestamp <= selection->timestamp) {
```
Xyene, line 495: I'm not sure this block is strictly necessary (never fired in my testing), but I don't think it can hurt.

xwayland/selection/outgoing.c
Xyene: (Note we're not exactly implementing ICCCM's prescribed behavior.)

Xyene: Prior art: https://invent.kde.org/plasma/kwin/-/blob/master/src/xwl/selection.cpp#L237

Though I think check has to be `<` instead of `<=` for safety, due to limited precision in `xcb_timestamp_t`.

xwayland/selection/incoming.c
Xyene: No prior art here; KWin doesn't do this so their selections can go "back in time": https://invent.kde.org/plasma/kwin/-/blob/master/src/xwl/selection.cpp#L106

Xyene:
Oops, this breaks pasting from Firefox into Slack, with:

```
00:18:09.285 [DEBUG] [wlr] [xwayland/selection/outgoing.c:378] XCB_SELECTION_REQUEST (time=0 owner=2097153, requestor=16777225 selection=277, target=404, property=363)
00:18:09.285 [ERROR] [wlr] [xwayland/selection/outgoing.c:397] ignoring old request from timestamp 0; expecting > 1124841
00:18:09.285 [DEBUG] [wlr] [xwayland/selection/outgoing.c:29] SendEvent destination=16777225 SelectionNotify(31) time=0 re
```

I'll have to look for why this works in KWin as it does...

xwayland/selection/outgoing.c
Xyene: KWin doesn't have the `XCB_CURRENT_TIME` check and works because https://invent.kde.org/plasma/kwin/-/blob/master/src/xwl/selection.cpp#L107 moves `m_timestamp` back to 0.

Qt has a similar `XCB_CURRENT_TIME` check to what's introduced in this PR, though: https://github.com/qt/qtbase/blob/658b93386eea34398de3ddacab436601e1ab5bbb/src/plugins/platforms/xcb/qxcbclipboard.cpp#L569-L570

xwayland/selection/selection.c
```diff
@@ -267,12 +267,14 @@ static void xwm_selection_set_owner(struct wlr_xwm_selection *selection,
...
 			selection->atom,
 			XCB_TIME_CURRENT_TIME);
+		xcb_flush(selection->xwm->xcb_conn);
 	} else {
 		if (selection->owner == selection->window) {
 			xcb_set_selection_owner(selection->xwm->xcb_conn,
 				XCB_WINDOW_NONE,
 				selection->atom,
 				selection->timestamp);
+			xcb_flush(selection->xwm->xcb_conn);
```
Xyene, line 277: Came to this conclusion stepping through the code, but KWin also does this: https://invent.kde.org/plasma/kwin/-/blob/master/src/xwl/selection.cpp#L209

emersion (approved):
Thanks!

PR #2742, opened 2021-02-15
surface: don't send protocol error on invalid buffer size

emersion:
_See individual commits._


ammen99 (approved):
Thanks :) wasn't even aware that this issue exists but the workaround here seems good.

@emersion you have tested this with a crashing cilent I hope, and it no longer crashes?

emersion:
Yup!

PR #2745, opened 2021-02-19
render/gles2: check buffer stride when uploading texture

emersion:
If the stride is too small, the driver could end up segfaulting
(e.g. radeonsi segfaults in __memmove_sse2_unaligned_erms).

cc @bl4ckb0ne


bl4ckb0ne:
Looks good to me!

PR #2747, opened 2021-02-19
Allow static linking

jbuchermn:
This PR includes a declared dependency for the Meson build system to allow wlroots to be linked statically when used as a subproject. This enables more fine-grained access beyond the public API (My use case is accessing gles2_get_renderer and similar methods to write a custom renderer)

Static linking is not enabled by default and has to be set using `meson build -Dwlroots:static=true` on the superproject.


emersion:
There's no need for a new Meson option, Meson already has a built-in option for this: `default_library=static`.

Example usage: https://github.com/Plagman/gamescope/blob/7eb7e8ec6348cf530c548c311c7475938a5f5ea9/meson.build#L48

jbuchermn:
Thanks very much for your quick support - works like a charm.

Issue #2750, opened 2021-02-20
How to create a wl_buffer by public interfaces


zccrs:
I'm trying to make a Wayland compositor useing Qt and wlroots. The QWindow/QSurface must need a native surface, I want use a wl_buffer with QPlatformWindow, and wl_buffer be qualified for this work. But I can only use wlr_gbm_allocator to create wl_buffer through backend/drm/renderer.h, or use wlr_wl_backend in the wayland backend, I think this is not a good idea.

emersion:
You'll need to create a new `wlr_backend` implementation which uses Qt.

zccrs:
> You'll need to create a new `wlr_backend` implementation which uses Qt.

I don't want to use Qt as the backend of wlrroots, but I want to use wlroots as the platform plugin of Qt .

emersion:
Ah. Can you explain your use-case in detail? I'm not sure I fully understand it.

Is it this way: you have a `wlr_backend`, and you want to use Qt to draw buffers for it?

zccrs:
The content rendered to QWindow will be saved in wlr_buffer, and I then the wlr_buffer will be composition into wlr_output, so that QWindow does not depend on the native window (eg: xwindow or wayland surface)

zccrs:
Like the: https://github.com/KDE/kwin/blob/master/src/plugins/qpa/window.cpp
kwin use the egl pbuffer for the QWindow, I want to use wlr_buffer.

zccrs:
QWindow is a drawable surface, and its backend may be an EGLSurface or GLXDrawable (depending on the implementation of the underlying plug-in), we can use opengl/gles to draw anything in QWindow, such as wlr_surface's wlr_texture using opengl to draw To QWindow, if the backend of QWindow is an X11 window, this will cause wlr_surface to be synthesized into an X11 window, just like the x11 backend of wlroots. In my compositor, I use wlr_output as the backend of QWindow, Use the Qt interface to draw the wlr_texture of wlr_surface into this QWindow, so that the windows created by wayland clients can be synthesized to the screen, and my compositor itself needs to use QWindow (these QWindow are ordinary windows, and wlr_output cannot be used. For example, create a dialog box in the synthesizer program to display some information), but the wayland client interface can no longer be used to create the wayland window in the synthesizer process. If the backend of these QWindow can use wlr_buffer, then I can use these wlr_buffer Synthesize with wlr_suface to wlr_output.

emersion:
[...]

PR #2751, opened 2021-02-20
backend: add 'wlr_backend_create_buffer' interface

zccrs:
Support create a new wlr_buffer by wlr_backend.

Closes: https://github.com/swaywm/wlroots/issues/2750

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -322,3 +322,11 @@ error:
 	wlr_backend_destroy(backend);
 	return NULL;
 }
+
+struct wlr_allocator *wlr_backend_get_allocator(struct wlr_backend *backend) {
+	if (!backend->impl->get_allocator) {
+		return NULL;
+	}
+
+	return backend->impl->get_allocator(backend);
+}
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -13,6 +13,7 @@
 #include <xf86drm.h>
 #include "backend/drm/drm.h"
 #include "util/signal.h"
+#include "render/gbm_allocator.h"
 
 struct wlr_drm_backend *get_drm_backend_from_backend(
 		struct wlr_backend *wlr_backend) {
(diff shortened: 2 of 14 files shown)


emersion:
I don't think this is desirable, the backend is not a buffer factory. In the future, the backend will just be a means to display a `wlr_buffer`.

emersion:
Relevant reading:

- https://github.com/swaywm/wlroots/issues/1352
- https://github.com/swaywm/wlroots/pull/2505

zccrs:
> I don't think this is desirable, the backend is not a buffer factory. In the future, the backend will just be a means to display a `wlr_buffer`.

I agree, but I did not find a more suitable place to add it.

zccrs:
I modified the commits of PR, it looks more reasonable, @emersion  please help me confirm whether it conforms to the design philosophy of wlroots

emersion:
I'm not comfortable adding this as a public API, because the backend will not have an allocator in the long run.

zccrs:
> I'm not comfortable adding this as a public API, because the backend will not have an allocator in the long run.

Is there any plan to provide a public API for the allocator?

emersion:
You mean make `wlr_allocator` public? Yes. It's not public yet because I wanted the dust to settle (e.g. making sure software rendering would work fine with this API).

zccrs:
> You mean make `wlr_allocator` public? Yes. It's not public yet because I wanted the dust to settle (e.g. making sure software rendering would work fine with this API).

Thank you, Is this work in progress ? In which version will you achieve this plans?

zccrs:
@emersion  Where is the wlroots Roadmap for the next few years ? I want  focus for the wlroots, how do get some plans of the wlroots release.

emersion:
I guess most of it is summed up in #1352. #1826 may also be relevant.

emersion:
> Thank you, Is this work in progress ? In which version will you achieve this plans?

I hope these new APIs to make it in wlroots 0.14.0.

PR #2752, opened 2021-02-20
Change the WLR_DRM_DEVICES separator to a comma

### LINKED ISSUE #1386: Change separator for WLR_DRM_DEVICES
At some point recently, udev started adding `/dev/dri/by-path` symlinks (similar to the disk ones) which provide a stable way to refer to DRM devices over reboots.

Unfortunately, these paths contain a colon character, referring to PCI bus ids:
```
/dev/dri/by-path/pci-0000:07:00.0-card
```
These colons mess with the way we separate the paths for multiple DRM devices.

It's much more preferable to use these symlinks instead of pointing to device nodes directly, so we need to do something to fix this.
Ideas:
- Change the separator to something else like a comma.
Would probably be fine, but not guaranteed to not cause issues again in the future.
- Introduce a way to escape colons.
More complex processing and generally uglier strings required to be entered, as well as possibly interacting badly with a shells own escaping.
- Change the separator to actually be ":/".
Seems too hacky, but the strings would look the same as they do now.

I'd probably just go with changing to a comma or semi-colon, unless someone else wants to input anything.

mkopec:
Change the `WLR_DRM_DEVICES` device list separator to a comma.
This allows the use of stable `/dev/dri/by-path` symlinks, which contain colons in their paths.

Fixes #1386.

### DIFF
--- a/backend/session/session.c
+++ b/backend/session/session.c
@@ -288,7 +288,7 @@ static ssize_t explicit_find_gpus(struct wlr_session *session,
 
 	size_t i = 0;
 	char *save;
-	char *ptr = strtok_r(gpus, ":", &save);
+	char *ptr = strtok_r(gpus, ",", &save);
 	do {
 		if (i >= ret_len) {
 			break;
@@ -300,7 +300,7 @@ static ssize_t explicit_find_gpus(struct wlr_session *session,
 		} else {
 			++i;
 		}
-	} while ((ptr = strtok_r(NULL, ":", &save)));
+	} while ((ptr = strtok_r(NULL, ",", &save)));
 
 	free(gpus);
 	return i;
--- a/docs/env_vars.md
+++ b/docs/env_vars.md
@@ -14,7 +14,7 @@ wlroots reads these environment variables
 
 ## DRM backend
 
-* *WLR_DRM_DEVICES*: specifies the DRM devices (as a colon separated list)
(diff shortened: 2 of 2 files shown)


ascent12 (approved):

emersion:
Other potential choices for bikeshedding purposes: semicolon, space, comma with a fallback to colon if there's no comma.

ascent12:
Eh, I think the chances of comma ever being an issue is pretty small, but I seriously hope people aren't using that shit in any `platform` devices. I'm pretty sure that's basically a free-form string, meaning any separator we choose could be an issue.

In the long term, this environment variable should be phased out, but until that's ready, I think just changing it to a comma would be fine.

emersion:
Well, comma is used in almost all platform devices: https://drmdb.emersion.fr/devices

mkopec:
I see. Semicolon seems like a safer option, should I change it to that then? Space looks like it would work, too, I don't have a preference either way.
I don't think comma with a fallback to colon would work if one was to enter a single device by it's `by-path` name.

progandy:
> I don't think comma with a fallback to colon would work if one was to enter a single PCI device by it's `by-path` name.

It could work if you add an additional test and check if the whole string is a valid filename before using the fallback.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2752

PR #2756, opened 2021-02-23
render/egl, backend/wayland: add workaround for split render/display setups

emersion:
Split render/display setups have two separate devices: one display-only
with a primary node, and one render-only with a render node. However
in these cases the EGL implementation and the Wayland compositor will
advertise the display device instead of the render device [1]. The EGL
implementation will magically open the render device when the display
device is passed in.

So just pass the display device as if it were a render device. Maybe in
the future Mesa will advertise the render device instead and we'll be
able to remove this workaround.

[1]: https://gitlab.freedesktop.org/mesa/mesa/-/issues/4178

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -143,7 +143,12 @@ static char *get_render_name(const char *name) {
 	if (match == NULL) {
 		wlr_log(WLR_ERROR, "Cannot find DRM device %s", name);
 	} else if (!(match->available_nodes & (1 << DRM_NODE_RENDER))) {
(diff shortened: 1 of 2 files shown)


emersion:
The `render/egl` workaround doesn't work properly because Mesa doesn't expose the right `EGLDevice` for display-only devices.

emersion:
Confirmed to work with Mesa master, now that this patch is merged: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/9697

PR #2760, opened 2021-02-24
Wire up libliftoff

emersion:
This doesn't yet bring any real benefit, since we only create layers
for the primary and cursor surfaces.


emersion:
Superseded by https://github.com/swaywm/wlroots/pull/3095, which uses a better approach.

PR #2761, opened 2021-02-26
Remove gl matrix from public rendering interface part1

bl4ckb0ne:
This patch is the first part of an ongoing progress to remove the GL-assumptions from the public interface of `wlr_renderer`. 

Here any mention of "transform matrix" is removed from the public API, and all of the matrix calculations for rendering a texture is done inside the gles2 renderer. This is done because the future pixman renderer can't use matrices to calculate position and requires an explicit x/y position. 

@emersion I was thinking about leaving the `wlr_render_*_with_matrix` functions for the next release to ease the transition and label them with a deprecation warning, and then remove them later on.

### DIFF
--- a/examples/fullscreen-shell.c
+++ b/examples/fullscreen-shell.c
@@ -16,6 +16,9 @@
 #include <wlr/types/wlr_surface.h>
 #include <wlr/util/log.h>
 
+/* Temp workaround */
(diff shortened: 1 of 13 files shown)


bl4ckb0ne:
Closed for not being relevant anymore since we fixed the matrix issue in #2770

PR #2762, opened 2021-02-27
backend/session/libseat: pass wlroots log level to libseat

alebastr:
Trivial change to get more logs from libseat.
CC @kennylevinsen

### DIFF
--- a/backend/session/libseat.c
+++ b/backend/session/libseat.c
@@ -74,6 +74,18 @@ static enum wlr_log_importance libseat_log_level_to_wlr(
 	}
 }
 
+static enum libseat_log_level wlr_log_importance_to_libseat(
+		enum wlr_log_importance importance) {
+	switch (importance) {
+	case WLR_ERROR:
+		return LIBSEAT_LOG_LEVEL_ERROR;
+	case WLR_INFO:
+		return LIBSEAT_LOG_LEVEL_INFO;
+	default:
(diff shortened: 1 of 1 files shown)


emersion:
I wonder if we really want to enable debug logs for all of our dependencies when wlroots is running in debug mode.

For instance, EGL never requests debug logs: https://github.com/swaywm/wlroots/blob/c5dad8b626539c0492f14075fec63bd0fb881e76/render/egl.c#L186

alebastr:
Debug logs are disabled in release build of libseat. I'd assume that if someone is running debug libseat they expect to see these logs.

emersion:
Oh, why are debug logs disabled in release builds? It makes it quite difficult for libseat users to ask their own users to get debug logs to investigate bugs.

kennylevinsen:
> Oh, why are debug logs disabled in release builds?

Just an occupational injury. With the current amount of debug logging, I wouldn't be opposed to enabling it in release builds.

Most debug logging happens in the server though - I think the only instance of debug logging in libseat is around drm pause/unpause events in logind. Rest is info or error level.

> I wonder if we really want to enable debug logs for all of our dependencies when wlroots is running in debug mode.

We need the ability to set per-module loglevels. :/

emersion:
Thanks for sending a PR, but I'd rather not configure our dependencies to enable debug logs.

PR #2763, opened 2021-02-28
Remove gl matrix from public rendering interface part2

bl4ckb0ne:
This patch is the second part of an ongoing progress to remove the GL-assumptions from the public interface of `wlr_renderer`. 

All of the matrix calculations for rendering a rectangle is done inside the gles2 renderer. This is done because the future pixman renderer can't use matrices to calculate position and requires an explicit x/y position.

Depends on #2761


bl4ckb0ne:
Closed for not being relevant anymore since we fixed the matrix issue in #2770

PR #2768, opened 2021-03-03
Fix #2018, false positive -Wstringop-truncation warning

### LINKED ISSUE #2018: xcursor: strncpy truncation warning with GCC 10 on s390x
When building `0.10.0` for Fedora (Rawhide, currently with GCC 10), there is a `strncpy` warning in `xcursor/xcursor.c`:

```
In file included from /usr/include/string.h:495,
                 from ../xcursor/xcursor.c:30:
In function 'strncpy',
    inlined from '_XcursorAddPathElt' at ../xcursor/xcursor.c:657:5,
    inlined from '_XcursorBuildFullname' at ../xcursor/xcursor.c:731:5,
    inlined from 'XcursorScanTheme' at ../xcursor/xcursor.c:820:13:
/usr/include/bits/string_fortified.h:106:10: error: '__builtin_strncpy' output truncated before terminating nul copying 7 bytes from a string of the same length [-Werror=stringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function 'strncpy',
    inlined from '_XcursorAddPathElt' at ../xcursor/xcursor.c:657:5,
    inlined from '_XcursorBuildFullname' at ../xcursor/xcursor.c:731:5,
    inlined from 'xcursor_load_theme' at ../xcursor/xcursor.c:959:10:
/usr/include/bits/string_fortified.h:106:10: error: '__builtin_strncpy' output truncated before terminating nul copying 7 bytes from a string of the same length [-Werror=stringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Thanks to `-Werror`, this aborts the build.

---

I checked the code around the specified lines, and the `\0` is currently added explicitly:
https://github.com/swaywm/wlroots/blob/f2943bdf61afe0a3ad2227d72fcbcac8b3088b1a/xcursor/xcursor.c#L657-L658
Because this only manifests on one architecture, and the warning is mitigated on the very next line, I'm inclined to blame GCC and consider this a false positive. Nevertheless, if we want to keep `-Werror` on, the offending code probably have to be changed.

For the time being, I will work around this by disabling the `-Werror` flag.

slyon:
Fix false positive stringop-truncation warning/error with GCC 10 on s390x by indicating GCC to explicitly ignore this case, as it is clearly a false positive (NUL is set in the following line).

This allow the compilation to succeed with -Werror on.

Fixes: https://github.com/swaywm/wlroots/issues/2018

### DIFF
--- a/xcursor/xcursor.c
+++ b/xcursor/xcursor.c
@@ -655,8 +655,11 @@ _XcursorAddPathElt (char *path, const char *elt, int len)
 	elt++;
 	len--;
     }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
     strncpy (path + pathlen, elt, len);
     path[pathlen + len] = '\0';
+#pragma GCC diagnostic pop
 }
 
 static char *
--- a/xcursor/xcursor.c
+++ b/xcursor/xcursor.c
@@ -655,11 +655,7 @@ _XcursorAddPathElt (char *path, const char *elt, int len)
 	elt++;
 	len--;
     }
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstringop-truncation"
-    strncpy (path + pathlen, elt, len);
-    path[pathlen + len] = '\0';
-#pragma GCC diagnostic pop
+    strncat (path + pathlen, elt, len);
 }
 
 static char *


emersion:
I don't want to clutter the codebase with compiler-specific directives to disable warnings.

slyon:
Many distributions build with treating warnings as errors, which makes the build fail in this case.
Do you see a better way to improve those two lines of code, to avoid that warning?

kennylevinsen:
Please see: https://github.com/swaywm/wlroots/wiki/Packaging-recommendations

For upstream, if the code is convoluted, improvement that happens to silence warnings is fine, but we shouldn't bend ourselves out of shape to silence a false positive. It should probably just be disabled in the meson file instead in that case.

emersion:
Maybe try to pass `len - 1`, and check if `len > 0`?

ascent12:
>Many distributions build with treating warnings as errors

Not really related to the issue, but this is a terrible idea for distributors in general. -Werror is a developer thing.
Upgrading your compiler could lead to a massive portion of builds breaking, especially for false-positive heavy warnings like this one.

Instead of carrying workarounds upstream, it's a far better idea to just try and fix whatever is actually causing the warning.

slyon:
I agree it is better to fix the actual problem, than cluttering the codebase with compiler specific directives.

What about just using `strncat` instead of `strncpy`? This will append the trailing `\0` automatically, thus no warning.

emersion (approved):
Thanks!

PR #2770, opened 2021-03-04
Update output transform matrix calculation

bl4ckb0ne:
Improve transform matrix calculation in `wlr_output` to help the pixman rendering process.

* * *

Breaking change: the projection has been moved from `wlr_output.transform_matrix` to the GLES2 renderer. Compositors using `wlr_output.transform_matrix` only with the GLES2 renderer don't need an update.

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -141,8 +141,15 @@ void wlr_output_update_enabled(struct wlr_output *output, bool enabled) {
 }
 
 static void output_update_matrix(struct wlr_output *output) {
-	wlr_matrix_projection(output->transform_matrix, output->width,
-		output->height, output->transform);
+	float width = output->width / 2.f;
+	float height = output->height / 2.f;
+
+	wlr_matrix_identity(output->transform_matrix);
+	if (output->transform != WL_OUTPUT_TRANSFORM_NORMAL) {
+		wlr_matrix_translate(output->transform_matrix, width, height);
+		wlr_matrix_transform(output->transform_matrix, output->transform);
+		wlr_matrix_translate(output->transform_matrix, -width, -height);
+	}
 }
 
 void wlr_output_enable(struct wlr_output *output, bool enable) {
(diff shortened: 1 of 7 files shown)


emersion:
Got correct results with this: https://l.sr.ht/zxBA.txt

bl4ckb0ne:
Runs well with `rotation`, `output-layout` and `fullscreen`

types/wlr_output.c
```diff
@@ -141,8 +141,13 @@ void wlr_output_update_enabled(struct wlr_output *output, bool enabled) {
 }
 
 static void output_update_matrix(struct wlr_output *output) {
-	wlr_matrix_projection(output->transform_matrix, output->width,
-		output->height, output->transform);
+	float width = output->width / 2.f;
+	float height = output->height / 2.f;
+
+	wlr_matrix_identity(output->transform_matrix);
+	wlr_matrix_translate(output->transform_matrix, width, height);
```
emersion, line 148: We can skip these 3 lines if `output->transform == WL_OUTPUT_TRANSFORM_NORMAL`.

emersion:
Can you squash these two commits, and add the motivation for this change in the commit message?

bl4ckb0ne:
Done, let me know if you want more info in the commit message.

emersion:
It seems like this breaks cursors. Applied this patch to the Wayland backend and works again:

https://l.sr.ht/zCxh.txt

Can you do something similar for DRM and X11?

bl4ckb0ne:
Good catch, I'll make a fix soon.

bl4ckb0ne:
I squashed the fix into the first commit. Do you think [`drm_surface_blit`][1] could break too?

[1]: https://github.com/swaywm/wlroots/blob/master/backend/drm/renderer.c#L159

emersion:
> Do you think drm_surface_blit could break too?

Yes, I think so. That function passes the matrix to `wlr_renderer_begin`, which already does the `wlr_matrix_projection` call under the hood. We should replace the `wlr_matrix_projection` call with an identity matrix.

To test, use `examples/screencopy-dmabuf`.

bl4ckb0ne:
I pushed a fix, I also found another call to `wlr_matrix_projection` in `gles2_blit_dmabuf`.

emersion (approved):
LGTM, thanks!

PR #2772, opened 2021-03-07
wlr_seat_pointer: Remove log on notify_button

kennylevinsen:
This seems like unnecessary noise.


emersion (approved):
Agreed, it's easy enough to add back the log if someone wants to debug this part of the code.

Thanks!

Issue #2774, opened 2021-03-10
9601a2 breaks rotations


LaserEyess:
After commit 9601a2abf024fb8ac400c0df178bb41ec9e5d215 my rotated monitor no longer works, reverting the commit fixes this issue.

Relevant config section:
```
output DP-1 {
    pos 0 0
    res 2560x1440@144Hz
    adaptive_sync on
}
output DP-2 {
    pos 2560 -500
    transform 270
    res 2560x1440@144Hz
    adaptive_sync on
}
output DP-3 {
    pos 4000 -200
    res 3840x2160@60Hz
    adaptive_sync on
    scale 2
}
```

DP-2 is affected, DP-1 and DP-3 are not. Will try to reproduce with a clean config.

emersion:
Can repro as well. Some notes:

- Only repros with 90 and 270. 180 is fine.
- Everything looks fine when the resolution is a square (width == height), so probably width and height need to be swapped somewhere.

LaserEyess:
On a clean config with nothing except those transforms, I can still reproduce it and as emersion says only 90/270 are affected.

emersion:
https://github.com/swaywm/wlroots/pull/2775

LaserEyess:
#2775 fixes this for me

PR #2775, opened 2021-03-11
output: fix transform matrix for 90/270 rotations

emersion:
We need to adjust the second translation depending on the transform
we applied.

Fixes: 9601a2abf024 ("output: improve transform matrix calculation"
Closes: https://github.com/swaywm/wlroots/issues/2774

cc @bl4ckb0ne


bl4ckb0ne:
I found the steps a bit repetitive, do you think it would be worth adding a function to `wlr_matrix` to wrap that with a single call?

emersion:
I was wondering about this, but then struggled to find a meaningful name for the new function, and realized https://github.com/swaywm/wlroots/pull/2507 would de-duplicate most of these repetitions.

bl4ckb0ne:
#2507 looks promising, LGTM then.

PR #2779, opened 2021-03-12
pixel_format: add format table

### LINKED ISSUE #2748: Centralized format table
Different parts of the code may need access to DRM format properties:

- Bits per pixel: renderers, shm allocator
- Opaque substitute: DRM backend
- Anything else?

It may make sense to have a central table to store this data.

Weston has one: https://gitlab.freedesktop.org/wayland/weston/-/blob/master/libweston/pixel-formats.c

bl4ckb0ne:
Closes #2748 

/cc @emersion


include/render/pixel_format.h
```diff
@@ -0,0 +1,32 @@
...
+	uint32_t drm_format;
+
+	/* Equivalent of the format if it has an alpha channel, 0 if NA */
+	uint32_t opaque_substitute;
+
+	/* Bytes per pixels */
+	uint32_t bpp;
+
+	/* Size of each channels, 0 if NA */
+	struct {
```
bl4ckb0ne, line 16: I don't know if that's useful, but I thought it would be good to have it. I'm open to remove it if that's really useless.

include/render/pixel_format.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <wayland-server-protocol.h>
+
+struct wlr_pixel_format_info {
+	uint32_t drm_format;
+
+	/* Equivalent of the format if it has an alpha channel, 0 if NA */
+	uint32_t opaque_substitute;
+
+	/* Bytes per pixels */
```
panaman67, line 12:
```suggestion
	/* Bits per pixel */
```

include/render/pixel_format.h
```diff
@@ -0,0 +1,24 @@
+#ifndef RENDER_PIXEL_FORMAT_H
+#define RENDER_PIXEL_FORMAT_H
+
+#include <wayland-server-protocol.h>
+
+struct wlr_pixel_format_info {
+	uint32_t drm_format;
+
+	/* Equivalent of the format if it has an alpha channel, 0 if NA */
```
emersion, line 9: I'd prefer to document that it's `DRM_FORMAT_INVALID` if N/A (which happens to be zero).

include/render/pixel_format.h
bl4ckb0ne: Thanks!

include/render/pixel_format.h
bl4ckb0ne: That's a good idea!

emersion:
Would be nice to have additional commits that make use of this table in the DRM backend and the renderer.

bl4ckb0ne:
I crammed everything in the same commit to have something that builds, gonna split it.

emersion:
I'd like to see two more commits to make use of the new data:

- One for the DRM backend, to replace `strip_alpha_channel` with this table: https://github.com/swaywm/wlroots/blob/6230f7be4f92b83df7ea8e6487410b3b81dd4b1d/backend/drm/renderer.c#L191
- One for the GLES2 renderer, to remove the fields we now have in the common table from the GLES-specific table: https://github.com/swaywm/wlroots/blob/6230f7be4f92b83df7ea8e6487410b3b81dd4b1d/render/gles2/pixel_format.c#L10
  The depth is not used and can be removed from the GLES2 table. We should only need the GLES2 table for GLES2-specific fields: `gl_type` and `gl_format`.

Does this sound reasonable?

backend/drm/renderer.c
```diff
@@ -212,8 +204,15 @@ static struct wlr_drm_format *create_linear_format(uint32_t format) {
 bool drm_plane_init_surface(struct wlr_drm_plane *plane,
 		struct wlr_drm_backend *drm, int32_t width, uint32_t height,
 		uint32_t format, bool with_modifiers) {
+	const struct wlr_pixel_format_info *info =
+		drm_get_pixel_format_info(format);
+	if (!info) {
```
emersion, line 209: If we don't have the format in our table, it doesn't mean we can't use it in DRM. It just means we can't fallback to the same format without the alpha channel, because we don't know what it is.

A better solution would be to only fallback to the opaque substitute if `info != NULL`.

backend/drm/renderer.c
```diff
@@ -213,8 +205,16 @@ bool drm_plane_init_surface(struct wlr_drm_plane *plane,
 		struct wlr_drm_backend *drm, int32_t width, uint32_t height,
 		uint32_t format, bool with_modifiers) {
 	if (!wlr_drm_format_set_has(&plane->formats, format, DRM_FORMAT_MOD_INVALID)) {
-		format = strip_alpha_channel(format);
+		const struct wlr_pixel_format_info *info =
+			drm_get_pixel_format_info(format);
+		if (!info) {
+			wlr_log(WLR_ERROR,
+				"Failed to fallback on DRM opaque substitute format");
```
emersion, line 212: Nit: can we log at least the format in hex? (See similar error message below)

backend/drm/renderer.c
```diff
@@ -380,9 +380,15 @@ static struct wlr_drm_fb *drm_fb_create(struct wlr_drm_backend *drm,
 
 	if (formats && !wlr_drm_format_set_has(formats, attribs.format,
 			attribs.modifier)) {
+		const struct wlr_pixel_format_info *info =
+			drm_get_pixel_format_info(attribs.format);
+		if (!info) {
+			wlr_log(WLR_ERROR, "drm format info not found");
```
emersion, line 386: Nit: can we maybe use a more descriptive error message (like "Failed to fallback on DRM opaque substitute format" that you've used above)? And also log the format code in hex?

include/render/gles2.h
```diff
@@ -17,7 +17,6 @@
 struct wlr_gles2_pixel_format {
 	uint32_t drm_format;
 	GLint gl_format, gl_type;
-	int depth, bpp;
 	bool has_alpha;
```
emersion, line 20: Can't we remove this one as well? It seems like it's in the shared table now.

emersion (changes requested):

include/render/gles2.h
bl4ckb0ne: It's required to identify if the format is opaque or not. Otherwise we are not able to tell the difference between ARGB and XRGB because they have the same `gl_format` and `gl_type` value.

emersion: Yes, but we have `has_alpha` in `wlr_pixel_format_info` as well. Why do we need to duplicate it in `wlr_gles2_pixel_format` as well?

bl4ckb0ne: It's a search parameter in `get_gles2_format_from_gl`,  we might not get the good drm format without it.

emersion: Oh. That's unfortunate indeed. We could query the shared table inside the `get_gles2_format_from_gl` loop, but that's not great either.

Let's just leave it at that for now then.

emersion (approved):
LGTM, thanks!

PR #2783, opened 2021-03-15
render/egl: don't fail on EGL_MESA_device_software

emersion:
Mesa may advertise EGL_MESA_device_software even when hardware
drivers are in use [1]. Demote the error to a warning until the Mesa
bug is fixed.

[1]: https://gitlab.freedesktop.org/mesa/mesa/-/issues/4178

References: https://github.com/swaywm/wlroots/pull/2689

cc @bl4ckb0ne


bl4ckb0ne:
Related issue https://gitlab.freedesktop.org/mesa/mesa/-/issues/4178

emersion:
Ah, I actually added the reference in the commit message, but never used the reference. Should be fixed now.

Thanks for having a look!

PR #2785, opened 2021-03-15
backend/session/libseat: Set loglevel to INFO

kennylevinsen:
Without INFO, the user won't know what seat backend got selected, only which ones failed.


emersion (approved):
Thanks!

emersion:
For reference, previous discussion: https://github.com/swaywm/wlroots/pull/2762

PR #2787, opened 2021-03-15
docs: Minor libseat corrections

kennylevinsen:
The optional dependency wasn't mentioned in the readme, and the env docs incorrectly listed "seatd" as being the WLR_SESSION selector for it.


emersion (approved):
Thanks!

Issue #2780, opened 2021-03-13
Multi-GPU outputs blank starting with 9601a2a


J0nnyMak0:
https://github.com/swaywm/wlroots/commit/9601a2abf024fb8ac400c0df178bb41ec9e5d215 has brought back the issue with blank external monitors using intel+nouveau configuration:

https://github.com/swaywm/wlroots/issues/2526

`WLR_DRM_NO_MODIFIERS=1` workaround is broken as well.

J0nnyMak0:
I'm also getting a few meson warnings. (I can open a separate issue for this if required.)
```
|WARNING: Project targeting '>=0.56.0' but tried to use feature deprecated since '0.56.0': meson.build_root. use meson.current_build_dir instead.
```
I'm using meson 0.57.1

emersion:
The warnings are a Meson bug: https://github.com/mesonbuild/meson/pull/7772#issuecomment-775201986

emersion:
Can you try https://github.com/swaywm/wlroots/pull/2788?

Dirbaio:
not OP but I also had the same symptoms: intel+nouveau, black screen since 9601a2a. 

Can confirm with #2788 everything works again!

PR #2788, opened 2021-03-16
Fix crash in screencopy dmabuf and fix black buffer blits

emersion:
_See each commit._

cc @bl4ckb0ne @any1

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -157,6 +157,7 @@ static struct wlr_buffer *drm_surface_blit(struct wlr_drm_surface *surf,
 
 	float mat[9];
 	wlr_matrix_identity(mat);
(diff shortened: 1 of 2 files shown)


any1:
It could be moved into `render/wlr_renderer.c` instead. Right?

emersion:
Could, but nobody else needs it right now.

any1:
But shouldn't then `wlr_renderer_blit_dmabuf` be removed?

emersion:
Done in https://github.com/swaywm/wlroots/pull/2789, submitted as a separate PR because that's a breaking change.

J0nnyMak0:
Yes, this fixes https://github.com/swaywm/wlroots/issues/2780 for me. Thanks!

PR #2791, opened 2021-03-17
Add kernel tracing instrumentation

emersion:
Useful to figure out what happens and when. The trace points can be visualized via gpuvis.

Not sure we want that upstream, I'm just submitting this as a draft so that one can easily add new tracepoints during an investigation.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -26,6 +26,7 @@
 #include "backend/drm/util.h"
 #include "render/swapchain.h"
 #include "util/signal.h"
+#include "util/trace.h"
 
 bool check_drm_features(struct wlr_drm_backend *drm) {
 	uint64_t cap;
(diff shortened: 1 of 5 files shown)


backend/drm/drm.c
```diff
@@ -330,6 +331,9 @@ static void drm_plane_set_committed(struct wlr_drm_plane *plane) {
 static bool drm_crtc_commit(struct wlr_drm_connector *conn, uint32_t flags) {
 	struct wlr_drm_backend *drm = conn->backend;
 	struct wlr_drm_crtc *crtc = conn->crtc;
+
+	wlr_trace("drm_crtc_commit");
```
agx, line 335: would be using `__FILE__` be better to be more robust on refactors?

agx:
I'd be nice to have something like this since correlating events between userspace and kernel is pretty useful.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2791

PR #2794, opened 2021-03-19
layer-shell-v1: introduce wlr_layer_surface_v1_state.committed

emersion:
This allows compositors to figure out which state fields have changed
since the last commit, and avoid needlessly re-computing their
configuration on each commit.

* * *

Not usable as-is, because `client_pending.committed` is reset before commit, and `current.committed` is not well-defined. Ref https://github.com/swaywm/wlroots/issues/2098

### DIFF
--- a/include/wlr/types/wlr_layer_shell_v1.h
+++ b/include/wlr/types/wlr_layer_shell_v1.h
@@ -43,7 +43,17 @@ struct wlr_layer_shell_v1 {
 	void *data;
 };
 
+enum wlr_layer_surface_v1_state_field {
+	WLR_LAYER_SURFACE_V1_STATE_ANCHOR = 1 << 0,
+	WLR_LAYER_SURFACE_V1_STATE_EXCLUSIVE_ZONE = 1 << 1,
+	WLR_LAYER_SURFACE_V1_STATE_MARGIN = 1 << 2,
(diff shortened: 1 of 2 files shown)


ifreund:
It is currently possible to avoid this needless re-configuration on each surface commit fwiw. River simply makes and stores a copy of the layer surface's current state struct to compare with the new one on the next commit.

Having a committed bitfield would certainly be better though.

include/wlr/types/wlr_layer_shell_v1.h
```diff
@@ -43,16 +43,27 @@ struct wlr_layer_shell_v1 {
 	void *data;
 };
 
+enum wlr_layer_surface_v1_state_field {
+	WLR_LAYER_SURFACE_V1_STATE_ANCHOR = 1 << 0,
+	WLR_LAYER_SURFACE_V1_STATE_EXCLUSIVE_ZONE = 1 << 1,
+	WLR_LAYER_SURFACE_V1_STATE_MARGIN = 1 << 2,
+	WLR_LAYER_SURFACE_V1_STATE_KEYBOARD_INTERACTIVITY = 1 << 2,
```
vyivel, line 50: A typo: `WLR_LAYER_SURFACE_V1_STATE_MARGIN` and `WLR_LAYER_SURFACE_V1_STATE_KEYBOARD_INTERACTIVITY` shouldn't be equal.

vyivel:
Superseded by https://github.com/swaywm/wlroots/pull/3214.

Issue #2798, opened 2021-03-22
examples: fails to build with ffmpeg >= 4.4 due to -Werror


jbeich:
Regressed by https://github.com/ffmpeg/ffmpeg/commit/f7db77bd8785d1715d3e7ed7e69bd1cc991f2d07

```c
examples/dmabuf-capture.c:494:4: error: 'av_init_packet' is deprecated [-Werror,-Wdeprecated-declarations]
                        av_init_packet(&pkt);
                        ^
/usr/include/libavcodec/packet.h:487:1: note: 'av_init_packet' has been explicitly marked deprecated here
attribute_deprecated
^
/usr/include/libavutil/attributes.h:100:49: note: expanded from macro 'attribute_deprecated'
#    define attribute_deprecated __attribute__((deprecated))
                                                ^
```

emersion:
https://github.com/swaywm/wlroots/pull/2799

PR #2799, opened 2021-03-22
examples/dmabuf-capture: stop using av_init_packet

emersion:
It's deprecated in ffmpeg >= 4.4.

Closes: https://github.com/swaywm/wlroots/issues/2798

### DIFF
--- a/examples/dmabuf-capture.c
+++ b/examples/dmabuf-capture.c
@@ -490,26 +490,27 @@ static void *vid_encode_thread(void *arg) {
 		}
 
 		while (1) {
-			AVPacket pkt;
-			av_init_packet(&pkt);
-
-			int ret = avcodec_receive_packet(ctx->avctx, &pkt);
+			AVPacket *pkt = av_packet_alloc();
+			int ret = avcodec_receive_packet(ctx->avctx, pkt);
 			if (ret == AVERROR(EAGAIN)) {
+				av_packet_free(&pkt);
 				break;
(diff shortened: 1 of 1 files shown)


examples/dmabuf-capture.c
```diff
@@ -506,10 +505,10 @@ static void *vid_encode_thread(void *arg) {
 				goto end;
 			}
```
kennylevinsen, line 506: If any of the above branches are hit, the AVPacket will be leaked.

emersion: Fixed

examples/dmabuf-capture.c
```diff
@@ -490,10 +491,8 @@ static void *vid_encode_thread(void *arg) {
...
 
 		while (1) {
-			AVPacket pkt;
-			av_init_packet(&pkt);
-
-			int ret = avcodec_receive_packet(ctx->avctx, &pkt);
+			pkt = av_packet_alloc();
+			int ret = avcodec_receive_packet(ctx->avctx, pkt);
 			if (ret == AVERROR(EAGAIN)) {
 				break;
```
kennylevinsen, line 497: We still leak from here as ctx->err is not set (assuming that the write below here doesn't fail for zero-sized packets).

emersion: Eh, indeed. I've just reverted to free'ing the packet inside each branch.

kennylevinsen (approved):
LGTM. Thanks!

PR #2802, opened 2021-03-22
linux-dmabuf-v1: always advertise support for implicit modifiers

emersion:
Some clients (like Xwayland) will fallback to wl_drm if the compositor
doesn't explicitly advertise support for implicit modifiers, even when
the compositor supports explicit modifiers. This behavior sounds correct
from a protocol point of view.

### DIFF
--- a/types/wlr_linux_dmabuf_v1.c
+++ b/types/wlr_linux_dmabuf_v1.c
@@ -385,9 +385,26 @@ struct wlr_linux_dmabuf_v1 *wlr_linux_dmabuf_v1_from_resource(
 	return dmabuf;
 }
 
+static void linux_dmabuf_send_modifiers(struct wl_resource *resource,
+		const struct wlr_drm_format *fmt) {
+	if (wl_resource_get_version(resource) < ZWP_LINUX_DMABUF_V1_MODIFIER_SINCE_VERSION) {
+		zwp_linux_dmabuf_v1_send_format(resource, fmt->format);
+		return;
+	}
+
+	for (size_t i = 0; i < fmt->len; i++) {
+		uint64_t mod = fmt->modifiers[i];
+		zwp_linux_dmabuf_v1_send_modifier(resource, fmt->format,
+			mod >> 32, mod & 0xFFFFFFFF);
+	}
+
+	// We always support buffers with an implicit modifier
+	zwp_linux_dmabuf_v1_send_modifier(resource, fmt->format,
+		DRM_FORMAT_MOD_INVALID >> 32, DRM_FORMAT_MOD_INVALID & 0xFFFFFFFF);
+}
+
 static void linux_dmabuf_send_formats(struct wlr_linux_dmabuf_v1 *linux_dmabuf,
-		struct wl_resource *resource, uint32_t version) {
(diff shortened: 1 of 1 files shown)


types/wlr_linux_dmabuf_v1.c
```diff
@@ -385,40 +385,35 @@ struct wlr_linux_dmabuf_v1 *wlr_linux_dmabuf_v1_from_resource(
 	return dmabuf;
 }
 
+static void linux_dmabuf_send_modifiers(struct wl_resource *resource,
+		const struct wlr_drm_format *fmt) {
+	if (wl_resource_get_version(resource) < ZWP_LINUX_DMABUF_V1_MODIFIER_SINCE_VERSION) {
+		zwp_linux_dmabuf_v1_send_format(resource, fmt->format);
```
kennylevinsen, line 391: We previously called `send_format` once for every modifier in the format equal to `DRM_FORMAT_MOD_LINEAR`, or once if this format had no modifiers, whereas we now just `send_format` once for every format.

The old logic was oddly elaborate and [appears copied from Weston](https://gitlab.freedesktop.org/wayland/weston/-/blob/master/libweston/linux-dmabuf.c#L524), so a confirmation that this behavior change is intended would be good.

kennylevinsen (commented):
Indeed, the spec doesn't seem to have any problem with always advertising implicit modifier support. Do any other compositors do this already?

types/wlr_linux_dmabuf_v1.c
emersion: `DRM_FORMAT_MOD_LINEAR` is a modifier, not a format.

The previous logic was pretty confusing.

- If modifiers were supported by neither the client nor the compositor: `send_format` once 
- If modifiers were not supported by the client but supported by the compositor: the list hopefully had `LINEAR` once, so we were calling `send_format` once
- If modifiers were not supported by the compositor but supported by the client: `send_modifier(INVALID)` once

emersion:
> Do any other compositors do this already?

Doesn't seem like it, but Pekka loosely confirmed this is the expected behavior. Probably worth a clarification in the protocol.

types/wlr_linux_dmabuf_v1.c
kennylevinsen: >> for every modifier in the format equal to `DRM_FORMAT_MOD_LINEAR`
>
> `DRM_FORMAT_MOD_LINEAR` is a modifier, not a format.

That's also what I tried to write. :P

> the list hopefully had LINEAR once, so we were calling send_format once

I imagined that this was an implied assumption, making for a rather odd way to do something once for every format. Appreciate the confirmation.

kennylevinsen (approved):
LGTM.

emersion:
Submitted a spec clarification here: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/75

types/wlr_linux_dmabuf_v1.c
emersion: I suspect the LINEAR thing is a confusion between implicit modifiers and linear layout by the way...

PR #2806, opened 2021-03-25
render: remove wlr_texture_to_dmabuf

emersion:
This is unused in wlroots, and the use-cases for compositors are
pretty niche since they can access the original DMA-BUF via the
wlr_buffer.


emersion:
One potential use-case for `wlr_texture_to_dmabuf` is to avoid keeping DMA-BUF FDs around, to avoid hitting the max FD limit. Though, it's not yet clear a roundtrip through EGL is the right way to address this.

emersion:
A roundtrip through GBM sounds like a better solution, if a roundtrip is deemed desirable.

PR #2811, opened 2021-03-26
xdg_shell: Fix invert_y of top right anchor

kennylevinsen:
Closes: https://github.com/swaywm/sway/issues/5985


emersion (approved):
Wow, good catch, thanks! Talk about finding a needle in a haystack.

PR #2815, opened 2021-03-31
Require INVALID for implicit format modifiers

emersion:
See [1] for the motivation.

[1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/75


emersion:
This breaks DRM cursors when modifiers aren't supported: the plane has only { LINEAR } supported and receives a buffer with INVALID.

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM! Thanks

PR #2816, opened 2021-03-31
backend/drm: use format table in test_buffer

emersion:
Instead of an ad-hoc strip_alpha_channel function, use the
centralized format table to get an opaque substitute.

cc @bl4ckb0ne


bl4ckb0ne:
LGTM

PR #2820, opened 2021-04-02
build: add subproject fallback for libseat

emersion:
This allows libseat to be compiled as a Meson subproject when it's
not installed system-wide. This can ease development and compilation
on distributions where libseat isn't packaged.

cc @kennylevinsen


kennylevinsen (approved):

kennylevinsen:
Thanks!

PR #2823, opened 2021-04-06
Remove unnecessary directories in the include path

zccrs:
Because the included file here is in the same directory as itself, it is
more appropriate to write only the file name, and doing so can help
unify the code style of the entire project.

### DIFF
--- a/include/backend/drm/drm.h
+++ b/include/backend/drm/drm.h
@@ -12,9 +12,9 @@
 #include <wlr/backend/session.h>
 #include <wlr/render/drm_format_set.h>
 #include <xf86drmMode.h>
-#include "iface.h"
-#include "properties.h"
-#include "renderer.h"
+#include "backend/drm/iface.h"
+#include "backend/drm/properties.h"
+#include "backend/drm/renderer.h"
 
 struct wlr_drm_plane {
 	uint32_t type;
--- a/include/types/wlr_keyboard.h
+++ b/include/types/wlr_keyboard.h
@@ -1,4 +1,4 @@
-#include "wlr/types/wlr_keyboard.h"
+#include <wlr/types/wlr_keyboard.h>
 
 void keyboard_key_update(struct wlr_keyboard *keyboard,
 		struct wlr_event_keyboard_key *event);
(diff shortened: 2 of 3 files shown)


zccrs:
Hi.

I don't know the falied cause of "alpine.yml", I think it has nothing to do with this commit. If it is my problem, please let me know, thank you.

emersion:
I prefer to always use absolute include paths.

zccrs:
> I prefer to always use absolute include paths.

But a lot of codes is used the relative path now, I think it should be consistent with them.

emersion:
Do you have examples where a relative path is used for includes?

zccrs:
> > I prefer to always use absolute include paths.
> 
> But a lot of codes is used the relative path now, I think it should be consistent with them.



> Do you have examples where a relative path is used for includes?

image

At present, the header files in most of the subdirectories of include are almost always included in the same directory as themselves when using'include "foo.h"'. For instance, there is another advantage to using relative paths. When refactoring the code, directly rename a directory will not affect the include place, such as #include "render/allocator.h", if you need to rename "render" to "private_render", you need to synchronize #include "render/allocator.h " change to #include "private_render/allocator.h", but using #include "allocator.h" does not need to do so. Unless the header files in the subdirectories of the include directory (except the wlr directory) in most cases need to be referenced by other header files in different directories, I think we should keep using relative directories.

emersion:
- For protocol files, I don't think we have the choice here?
- For `backend/drm.h`, we should switch to absolute includes.
- For `wlr_keyboard_group.h`, we should switch to angle-brackets since these are public headers.
- The rest already seems to be using absolute includes.

emersion:
> For instance, there is another advantage to using relative paths. When refactoring the code, directly rename a directory will not affect the include place, such as #include "render/allocator.h"

When moving existing code from one place to another, absolute paths remove the need to update the include paths.

zccrs:
> * For protocol files, I don't think we have the choice here?
> * For `backend/drm.h`, we should switch to absolute includes.
> * For `wlr_keyboard_group.h`, we should switch to angle-brackets since these are public headers.
> * The rest already seems to be using absolute includes.

Ok, I has been update the commit, plase review it.

emersion (approved):
Thanks!

emersion:
The Alpine build is failing, but that's unrelated to this change.

PR #2824, opened 2021-04-06
build: simplify HAS_LIBUUID definition

emersion:
We can just use to_int() instead of having two if branches.


kennylevinsen (approved):
Ah, neat. LGTM.

PR #2825, opened 2021-04-06
Remove WLR_HAS_XCB_ERRORS

emersion:
wlroots' dependency on this library doesn't change the features
exposed to compositors. It's purely a wlroots implementation detail.
Thus downstream compositors shouldn't really care about it.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -86,8 +86,10 @@ features = {
 	'libseat': false,
 	'x11-backend': false,
 	'xwayland': false,
-	'xcb-errors': false,
 	'xdg-foreign': false,
+}
+internal_features = {
+	'xcb-errors': false,
 }
 
 wayland_server = dependency('wayland-server', version: '>=1.19')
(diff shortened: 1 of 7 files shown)


meson.build
```diff
@@ -86,7 +86,6 @@ features = {
 	'libseat': false,
 	'x11-backend': false,
 	'xwayland': false,
-	'xcb-errors': false,
```
kennylevinsen, line 89: Hmm, while not a huge problem, this does give slightly reduced feature visibility.

emersion: Yeah, I agree. `features` is used for public-facing features since it's exposed through pkgconfig and project variables. Maybe we can have something like `private_features` (TODO: find a better name) for dependencies like xcb-errors?

emersion: Done

kennylevinsen (approved):

kennylevinsen:
Thanks!

Issue #2250, opened 2020-06-04
backend/drm: wire up wlr_output_test to KMS TEST_ONLY commits


emersion:
Perform a KMS TEST_ONLY atomic commit when the compositor calls `wlr_output_test`.

The main blocker is how we handle `WLR_OUTPUT_STATE_BUFFER_RENDER`. We can't swap buffers and call `gbm_surface_lock_front_buffer` during a test commit.

PR #2827, opened 2021-04-06
backend/drm: use atomic test-only commits for direct scan-out

emersion:
This allows callers to use wlr_output_test to figure out whether a
buffer can be scanned out prior to committing the output.

References: https://github.com/swaywm/wlroots/issues/2250


kennylevinsen (approved):
LGTM

emersion:
Pushed a new commit to remove an extraneous `test_buffer` call.

kennylevinsen (approved):
Ah, nice catch. We're still doing a test for every new buffer during direct scanout, but 1 is certainly better than 2. :)

PR #2832, opened 2021-04-07
render/egl: check "EGL_KHR_platform_gbm" for EGL_PLATFORM_GBM_KHR platform

zccrs:
See the https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_platform_gbm.txt
example code. In use EGL_PLATFORM_GBM_KHR before, should be check the
"EGL_KHR_platform_gbm" extension.

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -172,6 +172,13 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display) {
 		return NULL;
 	}
 
+	if (platform == EGL_PLATFORM_GBM_KHR) {
+		if (!check_egl_ext(client_exts_str, "EGL_KHR_platform_gbm")) {
+			wlr_log(WLR_ERROR, "EGL_KHR_platform_gbm not supported");
+			return NULL;
+		}
+	}
+
 	if (!check_egl_ext(client_exts_str, "EGL_EXT_platform_base")) {
 		wlr_log(WLR_ERROR, "EGL_EXT_platform_base not supported");
 		return NULL;


render/egl.c
```diff
@@ -172,6 +172,15 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display) {
 		return NULL;
 	}
 
+#ifdef EGL_KHR_platform_gbm
+	if (platform == EGL_PLATFORM_GBM_KHR) {
```
emersion, line 176: No need for that `#ifdef`, we require the symbols for that ext.

render/egl.c
```diff
@@ -172,6 +172,15 @@ struct wlr_egl *wlr_egl_create(EGLenum platform, void *remote_display) {
...
+		if (!check_egl_ext(client_exts_str, "EGL_KHR_platform_gbm")) {
```
emersion, line 177: Can we check for `EGL_MESA_platform_gbm` as well?

https://www.khronos.org/registry/EGL/extensions/MESA/EGL_MESA_platform_gbm.txt

emersion: Well, I guess this isn't a big deal, since only ancient Mesa version would have `EGL_MESA_platform_gbm` but not `EGL_KHR_platform_gbm`.

zccrs: On some devices that if it is not use MESA driver. In the case, Do we should be check the EGL_MESA_platform_gbm?

emersion: I don't think non-Mesa drivers have the MESA ext

emersion (approved):
LGTM, thanks!

Xyene:
> Well, I guess this isn't a big deal, since only ancient Mesa version would have `EGL_MESA_platform_gbm` but not `EGL_KHR_platform_gbm`.

FWIW, this does break wlroots on Debian stable, which ships Mesa 18.3.6 (released almost exactly 2 years ago).

As a meta-point it does make me a _little_ sad when we break compatibility with Debian stable for things that can't easily be solved on the user end (like upgrading Mesa here, or upgrading the whole of GTK for the removal of `gtk-primary-unstable`).

I think there's a balance to be had (e.g., having to support old `wayland-protocols` or `meson` sounds like a hassle, when it's easily upgradable from backports on the user's end), but it's unfortunate when the options for Debian stable users are {use an older Sway, upgrade to unstable Debian, maintain custom patches reverting Sway/wlroots commits}.

emersion:
I'd just recommend to use older Sway releases.

PR #2833, opened 2021-04-07
util/uuid: replace with util/token, remove libuuid

### LINKED ISSUE #2830: Replace libuuid with a random util func
We only use libuuid to generate good randomness for a string token. libuuid doesn't do [anything interesting](https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/tree/lib/uuid/gen_uuid.c#n650), so we could just internalize the few lines we need and save a dependency.

r-c-f:
Uses getentropy() when available, otherwise /dev/urandom. Closes #2830


meson.build
```diff
@@ -105,14 +105,7 @@ math = cc.find_library('m')
 rt = cc.find_library('rt')
 
 if not get_option('xdg-foreign').disabled()
```
emersion, line 107: We can probably remove this option, since `xdg-foreign` doesn't have any external dep anymore.

util/meson.build
```diff
@@ -10,7 +10,8 @@ wlr_files += files(
 
 
 if features.get('xdg-foreign')
-	add_project_arguments('-DHAS_LIBUUID=@0@'.format(uuid.found().to_int()), language: 'c')
-	wlr_deps += uuid
+	if cc.has_function('getentropy')
```
emersion, line 13: Has `getentropy` some advantages over `/dev/urandom`?

util/uuid.c
```diff
@@ -1,34 +1,45 @@
...
+#include <inttypes.h>
+#include <stdint.h>
+#include <stdio.h>
 #include <string.h>
-#include <stdlib.h>
+#include <unistd.h>
+
+#ifndef HAS_GETENTROPY
+static int getentropy(void *buffer, size_t len)
+{
```
emersion, line 12: Nit: `{` should be on the previous line

util/uuid.c
```diff
@@ -1,34 +1,45 @@
...
+	FILE *urandom;
+	int ret = -1;
+
+	if ((urandom = fopen("/dev/urandom", "r"))) {
+		if (fread(buffer, 1, len, urandom) != 1) {
```
emersion, line 17: This should check `!= len`, I think.

util/uuid.c
```diff
@@ -1,34 +1,45 @@
...
+#include <stdint.h>
+#include <stdio.h>
 #include <string.h>
-#include <stdlib.h>
+#include <unistd.h>
+
+#ifndef HAS_GETENTROPY
+static int getentropy(void *buffer, size_t len)
+{
+	FILE *urandom;
```
emersion, line 13: Maybe we can make this variable `static`, so that it's opened only once?

That means we'll leak a file on exit, don't know it it's a big deal.

util/uuid.c
```diff
@@ -1,34 +1,45 @@
...
+bool generate_uuid(char out[static 37])
+{
+	uint16_t data[8];
 
-bool generate_uuid(char out[static 37]) {
-	uuid_t uuid;
-	uint32_t status;
-	uuid_create(&uuid, &status);
-	if (status != uuid_s_ok) {
+	if (getentropy(data, 16)) {
```
emersion, line 36: `sizeof(data)` instead of hardcoding 16, so that it's easy to know where this number comes from.

util/uuid.c
```diff
@@ -1,34 +1,45 @@
...
 		return false;
 	}
-
-	assert(strlen(str) + 1 == 37);
-	memcpy(out, str, 37);
-	free(str);
 	return true;
 }
-#endif
+
```
emersion, line 45: Style nit: unnecessary newline

jbeich:
Note, DragonFly and NetBSD lack `getentropy` but have `getrandom`. OpenBSD has `getentropy` but lacks `getrandom`. FreeBSD has both. All BSDs (and Android) have `arc4random_buf`.

r-c-f:
The main benefit to `getentropy` is the avoidance of file IO, with the possible future benefit of [being considered for acceptance into POSIX](https://www.austingroupbugs.net/view.php?id=1134#c5025); if that doesn't matter, static `/dev/urandom` for everything would be the simpler (if slightly leakier) approach, given that a Wayland compositor is already opening up too many file descriptors to worry about constraints.

kennylevinsen:
From the proposed POSIX spec:

```
NAME

    getentropy -- fill a buffer with random bytes

<< SNIP >>

APPLICATION USAGE

    The intended use of this function is to create a seed for other pseudo-random number generators.


RATIONALE

    The getentropy() function is not a cancellation point. (See [xref to 2.9.5.2 Cancellation Points].)
 ```
 
Even though they use the same source, `/dev/urandom` appears more portable for now and is in line with what libuuid was doing.

File: util/uuid.c [hunk 8]
```diff
@@ -1,34 +1,44 @@
...
+	return ret;
+}
+#endif
+#define UUID_FORMAT "%04" PRIx16 "%04" PRIx16 \
+	"-%04" PRIx16 \
+	"-%04" PRIx16 \
+	"-%04" PRIx16 \
+	"-%04" PRIx16 "%04" PRIx16 "%04" PRIx16
+bool generate_uuid(char out[static UUID_STRLEN])
```
kennylevinsen, line 30: @emersion should we just make this a 128-bit hex string instead of a UUID?

util/uuid.c
```diff
@@ -1,34 +1,33 @@
...
+{
+	static FILE *urandom = NULL;
+	uint16_t data[8];
+
+	if (!urandom) {
+		if (!(urandom = fopen("/dev/urandom", "r"))) {
+			return false;
```
emersion, line 20: It would be nice to call `wlr_log_errno` here to make it clear something wrong happened.

util/uuid.c
```diff
@@ -1,34 +1,33 @@
...
+		}
+	}
+	if (fread(data, sizeof(data), 1, urandom) != 1) {
 		return false;
```
emersion, line 24: Same here

util/uuid.c
emersion: I'm fine with either. Would it make the string formatting less awkward?

emersion:
Marking as a breaking change since this removes `WLR_HAS_XDG_FOREIGN`.

emersion:
Hm, actually, it removes `xdg-foreign` for `features` but leaves `WLR_HAS_XDG_FOREIGN`. We should either keep both or get rid of both.

emersion:
We can also get rid of libuuid in CI.

util/uuid.c
r-c-f: Without any pretense of UUID formatting one can just use two 64-bit numbers `"%016" PRIx64 "%016" PRIx64` and keep everything on one line. Though that leads to the question of whether it should be renamed to reflect what it's actually doing.

emersion: Yeah. If we go for an hex string, we could s/uuid/token/

util/token.c
```diff
@@ -0,0 +1,30 @@
+#include "util/token.h"
+#include "wlr/util/log.h"
+
+#include <inttypes.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+bool generate_token(char out[static TOKEN_STRLEN])
+{
```
emersion, line 10: Style nit: `{` should be on the previous line.

emersion (commented):
Code LGTM! Can you squash all commits into a single one, and remove the CI dependencies introduced in bf4e2e0eac4b336747979dbda30d4f328ab0bf3d (in `.builds/*.yml`)?

emersion:
Ah, also please remove libuuid from the README.

emersion (approved):
LGTM, thanks!

Issue #2819, opened 2021-04-02
gma500 regression (renderer v6)


stev47:
`examples/simple` used to work with gma500 (drm backend and egl software rendering) and fails since ef846a883950e59f95507daa6d3f96f0e84da3af.
gma500 supports only dumb buffers afaik, so #2399 with a dumb buffer allocator might change the situation, but since `drmPrimeHandleToFD` fails there might be more to it.

<details>
<summary>log bad 7709a965e596f66efabaf6c22967adbac0bbddb2</summary>

```
00:00:00.000 [INFO] [backend/session/logind.c:572] Selecting session from XDG_SESSION_ID: 3
00:00:00.064 [INFO] [backend/session/logind.c:706] Successfully loaded logind session
00:00:00.090 [INFO] [backend/backend.c:167] Found 1 GPUs
00:00:00.091 [INFO] [backend/drm/backend.c:155] Initializing DRM backend for /dev/dri/card0 (gma500)
00:00:00.091 [DEBUG] [backend/drm/drm.c:65] Atomic modesetting unsupported, using legacy DRM interface
00:00:00.091 [DEBUG] [backend/drm/drm.c:82] ADDFB2 modifiers unsupported
00:00:00.091 [INFO] [backend/drm/drm.c:244] Found 2 DRM CRTCs
00:00:00.092 [INFO] [backend/drm/drm.c:171] Found 2 DRM planes
pci id for fd 9: 8086:0be1, driver (null)
MESA-LOADER: failed to open gma500: /usr/lib/dri/gma500_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/dri)
failed to load driver: gma500
pci id for fd 9: 8086:0be1, driver (null)
MESA-LOADER: failed to open gma500: /usr/lib/dri/gma500_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/dri)
failed to load driver: gma500
00:00:00.510 [INFO] [render/egl.c:286] Warning: software rendering may be in use
00:00:00.510 [INFO] [render/egl.c:287] If you experience slow rendering, please check the OpenGL drivers are correctly installed
00:00:00.510 [INFO] [render/egl.c:309] Using EGL 1.4
[...]

PR #2834, opened 2021-04-08
drm: check for PRIME support

stev47:
PRIME support for buffer sharing has become mandatory since the renderer
rewrite. Make sure we check for the appropriate capabilities in backend,
allocator and renderer.

See also #2819.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -30,14 +30,13 @@
 
 bool check_drm_features(struct wlr_drm_backend *drm) {
 	uint64_t cap;
-	if (drm->parent) {
-		if (drmGetCap(drm->fd, DRM_CAP_PRIME, &cap) ||
-				!(cap & DRM_PRIME_CAP_IMPORT)) {
-			wlr_log(WLR_ERROR,
-				"PRIME import not supported on secondary GPU");
-			return false;
-		}
+	if (drmGetCap(drm->fd, DRM_CAP_PRIME, &cap) ||
+			!(cap & DRM_PRIME_CAP_IMPORT)) {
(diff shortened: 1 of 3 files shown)


hizukiayaka:
I wonder whether we need to check that in gbm because it would fail in gbm_bo_import()

stev47:
The gbm allocator unconditionally calls `gbm_bo_get_fd` during `create_buffer`, so it seems to me to be a good idea to check for prime export in the allocator independently of the current global code path.

By the way: for gma500 it's the following failing path:
1. backend check (this patch)
2. gles renderer check (this patch)
3. gbm allocator check (this patch)
4. no render format support (misleading error message due to empty format list)
5. failed `gbm_bo_get_fd` in gbm allocator

backend/drm/drm.c
```diff
@@ -30,6 +30,12 @@
...
 	uint64_t cap;
+	if (drmGetCap(drm->fd, DRM_CAP_PRIME, &cap) ||
+			!(cap & DRM_PRIME_CAP_IMPORT)) {
+		wlr_log(WLR_ERROR, "PRIME import not supported");
+		return false;
+	}
+
 	if (drm->parent) {
 		if (drmGetCap(drm->fd, DRM_CAP_PRIME, &cap) ||
 				!(cap & DRM_PRIME_CAP_IMPORT)) {
```
emersion, line 41: Since we check for this right above, I think this check can be removed.

emersion (commented):
Otherwise looks good!

stev47:
thanks, I've removed the redundant check

emersion (approved):
Thanks!

PR #2835, opened 2021-04-09
Add generic drm framebuufer interface

hizukiayaka:
Although it sounds that only the drm backend would use the drm framebuffer, but I am sure it would be necessary that we could bring a dmabuf which is allocated by somewhere else to be used in gbm or just drm itself.

We could skip those gbm back-compatible code but weston has them but it would detect the requirement of keeping compatible based on gbm version, I would prefer to do a symbol checking, that is all.

@emersion I modified some patches from you for SHM, to work with the other PR you have known.

I didn't full adapt the drm dumb buffer to use drm_dumb_buffer, it could as its only private data is a struct storing the gem handle.


emersion:
GBM is no longer used by the DRM backend, so I think this PR is now obsolete. Please re-open if desirable.

PR #2836, opened 2021-04-09
build: disable libseat subproject server and man pages

emersion:
When libseat is built as a subproject, we're not interested in
building the server or the man pages.


kennylevinsen (approved):

kennylevinsen:
Thanks!

PR #2838, opened 2021-04-10
Fix #2777, compiler false positive in xcursor.c

### LINKED ISSUE #2777: Compilation fails for xcursor.c
After commit 7dffe9339bf8a92a556098d86712c4c38ac95226 I get the following warning (which is treated as an error due to default -werror=true):
```
ninja -C build
ninja: Entering directory `build'
[279/351] Compiling C object libwlroots.so.7.p/xcursor_xcursor.c.o
FAILED: libwlroots.so.7.p/xcursor_xcursor.c.o
cc -Ilibwlroots.so.7.p -I. -I.. -Iinclude -I../include -Iprotocol -I../protocol -I/usr/include/libdrm -I/usr/include/pixman-1 -I/usr/include/uuid -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c11 -O3 -DWLR_USE_UNSTABLE -Wundef -Wlogical-op -Wmissing-include-dirs -Wold-style-definition -Wpointer-arith -Winit-self -Wstrict-prototypes -Wimplicit-fallthrough=2 -Wendif-labels -Wstrict-aliasing=2 -Woverflow -Wmissing-prototypes -Walloca -Wno-missing-braces -Wno-missing-field-initializers -Wno-unused-parameter -fmacro-prefix-map=../= -DHAS_LIBUUID=1 '-DICONDIR="/usr/share/icons"' -fPIC -MD -MQ libwlroots.so.7.p/xcursor_xcursor.c.o -MF libwlroots.so.7.p/xcursor_xcursor.c.o.d -o libwlroots.so.7.p/xcursor_xcursor.c.o -c ../xcursor/xcursor.c
In function ‘_XcursorAddPathElt’,
    inlined from ‘_XcursorBuildFullname’ at ../xcursor/xcursor.c:731:5,
    inlined from ‘XcursorScanTheme’ at ../xcursor/xcursor.c:719:1:
../xcursor/xcursor.c:658:5: error: ‘strncat’ specified bound 7 equals source length [-Werror=stringop-overflow=]
  658 |     strncat (path + pathlen, elt, len);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘_XcursorAddPathElt’,
    inlined from ‘_XcursorBuildFullname’ at ../xcursor/xcursor.c:731:5,
    inlined from ‘xcursor_load_theme’ at ../xcursor/xcursor.c:719:1:
../xcursor/xcursor.c:658:5: error: ‘strncat’ specified bound 7 equals source length [-Werror=stringop-overflow=]
  658 |     strncat (path + pathlen, elt, len);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[292/351] Compiling C object libwlroots.so.7.p/xwayland_xwm.c.o
ninja: build stopped: subcommand failed.
```
Temporary solution is to set werror=false

yuja:
This reverts commit 7dffe9339bf8a92a556098d86712c4c38ac95226, which introduced
another linter error with -O3:
    
      error: ‘strncat’ specified bound 7 equals source length [-Werror=stringop-overflow=]

and replace strncpy() with memcpy() to suppress the original warning.

### DIFF
--- a/xcursor/xcursor.c
+++ b/xcursor/xcursor.c
@@ -655,11 +655,8 @@ _XcursorAddPathElt (char *path, const char *elt, int len)
 	elt++;
 	len--;
     }
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstringop-truncation"
-    strncpy (path + pathlen, elt, len);
+    memcpy (path + pathlen, elt, len);
     path[pathlen + len] = '\0';
-#pragma GCC diagnostic pop
 }
 
 static char *
--- a/xcursor/xcursor.c
+++ b/xcursor/xcursor.c
@@ -655,7 +655,11 @@ _XcursorAddPathElt (char *path, const char *elt, int len)
 	elt++;
 	len--;
     }
-    strncat (path + pathlen, elt, len);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+    strncpy (path + pathlen, elt, len);
+    path[pathlen + len] = '\0';
+#pragma GCC diagnostic pop
 }
 
 static char *


kennylevinsen:
@yuja can you reproduce the original warning to verify that this fixes it? I don't recall experiencing it myself, in which case I'm unable to verify it as fixed.

yuja:
Sorry, I can't reproduce the original truncation warning even with s390x-linux-gnu-gcc-10. Maybe need an s390x VM?
I needed to comment out the `path[pathlen + len] = '\0'` line to get that, which clearly makes the warning correct.

The use of memcpy() is documented way to suppress the false positive, so I think it would 
avoid the problem reported in #2018.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstringop-truncation

emersion:
> The use of memcpy() is documented way to suppress the false positive

The manual says:

>  To avoid the warning when the result is not expected to be NUL-terminated, call memcpy instead.

However we expect the result to be NUL-terminated here.

yuja:
> we expect the result to be NUL-terminated here.

No, we don't always expect that. Since `elt` may be a part of comma-separated string (e.g. `/usr/local:/usr`),
we do append NUL at the end. `strncat()` would do that automatically, but it introduced another false positive.

emersion:
The result is the output of the function, not the input. `elt` is an input.

yuja:
Yes. I mean `elt` is partial string, so we can't always make `strncpy()` copy the NUL byte from `elt`.
So the result of strncpy() is not expected to be NUL terminated.
That's why we do append NUL.

emersion (approved):
Hm, I see. So the `memcpy` can never copy past the end of `elt`.

Thanks for the explanation. This file is terrible.

PR #2786, opened 2021-03-15
session: Remove logind backend

kennylevinsen:
Rely on libseat instead.

See https://github.com/swaywm/wlroots/issues/2755

Issue #2755, opened 2021-02-22
Delegate logind support to seatd


emersion:
Drop our logind code, rely on seatd instead.

cc @kennylevinsen

Conan-Kudo:
I'm not sure this is a good idea, given that [seatd calls itself "alpha" right now](https://git.sr.ht/~kennylevinsen/seatd/tree/34f55a3e2479daf0dae294dd22af25fcb6b25a01/item/README.md).

Can we hold off on this until it matures more?

kennylevinsen:
> seatd calls itself "alpha" right now.

https://git.sr.ht/~kennylevinsen/seatd/commit/0d855a28f20ab3d7b2058fbe58a172a97affbde5

Conan-Kudo:
@kennylevinsen Do you consider it stable to rely on?

emersion:
I'm not too worried about the libseat code.

Conan-Kudo:
Why not?

emersion:
It's pretty similar to wlroots' code, and has been around for a while.

emersion:
I guess the next step would be to make libseat a required dependency for wlroots, and nuke our direct session (and perhaps our session interface).

prez:
>and nuke our direct session

why?

emersion:
Because it's code which I'd rather share with other compositors than maintain, just like e.g. libinput. It's pretty trivial to drop seatd into `subprojects/` and make Meson build it as a subproject anyways.

emersion:
libseat is now required and replaces all of our session code.

emersion:
If you were using the direct backend and libseat is not yet packaged for your distribution, the replacement is now:

- Drop seatd into `subprojects/seatd`
- Run Meson with `-Dseatd:builtin=enabled`

And things should work just like before.

PR #2839, opened 2021-04-10
session: Remove direct and noop backends

kennylevinsen:
Rely on libseat instead.

See #2755. Based on #2786.


emersion:
At this point, maybe we can drop `session_impl` as well?

kennylevinsen:
> At this point, maybe we can drop `session_impl` as well?

Yeah, the only thing left is the noop backend, which is now in libseat master. If we drop that we could just merge session.c and libseat.c.

emersion:
Isn't the noop backend dropped in "backend/session: Remove noop backend" already?

kennylevinsen:
Hah, briefly forgot I had done that already.

All that's left then is cleaning up the interface.

kennylevinsen:
Marking as breaking change as this removes `WLR_HAS_LIBSEAT` (now mandatory), and changes `struct wlr_session` and `struct wlr_device`.

docs/env_vars.md
```diff
@@ -6,8 +6,6 @@ wlroots reads these environment variables
   libinput, drm, wayland, x11, headless, noop)
 * *WLR_NO_HARDWARE_CURSORS*: set to 1 to use software cursors instead of
   hardware cursors
-* *WLR_SESSION*: specifies the `wlr_session` to be used (available sessions:
```
emersion, line 9: Nit: this belongs to the "Remove session_impl" commit.

backend/session/session.c
```diff
@@ -1,32 +1,131 @@
...
+	session->active = false;
+	wlr_signal_emit_safe(&session->events.active, NULL);
+	libseat_disable_seat(session->seat_handle);
+}
+
+static int libseat_event(int fd, uint32_t mask, void *data) {
+	struct libseat *seat = data;
+	if (libseat_dispatch(seat, 0) == -1) {
+		wlr_log_errno(WLR_ERROR, "Failed to dispatch libseat");
+		exit(-1);
```
emersion, line 43: Since wlroots is a library, I'd prefer to not exit here. Although I'm not sure we can do a lot more than just logging an error.

Also see TODOs in `handle_libinput_readable` for instance.

backend/session/session.c
```diff
@@ -177,13 +265,17 @@ void wlr_session_destroy(struct wlr_session *session) {
...
-	session->impl->destroy(session);
+	libseat_session_finish(session);
+	free(session);
 }
 
 struct wlr_device *wlr_session_open_file(struct wlr_session *session,
 		const char *path) {
-	int fd = session->impl->open(session, path);
-	if (fd < 0) {
+
```
emersion, line 274: Style nit: unnecessary newline

backend/session/session.c
kennylevinsen: Indeed, it's not nice to exit. Didn't know what to do here. Just got carried over from the libweston MR.

I'll try to copy the display termination.

docs/env_vars.md
kennylevinsen: Split it for correctness so the "direct" commit removes the direct option, while the "session_impl" commit removes the env var.

backend/session/session.c
kennylevinsen: We now call `wl_display_terminate`.

backend/session/session.c
kennylevinsen: Fixed.

backend/session/session.c
emersion: Handled the rest in https://github.com/swaywm/wlroots/pull/2857

emersion (approved):
LGTM, thanks!

PR #2846, opened 2021-04-12
matrix: deprecate wlr_matrix_projection

emersion:
This function has baked-in GL assumptions. Compositors shouldn't
need to use it.

### DIFF
--- a/include/wlr/types/wlr_matrix.h
+++ b/include/wlr/types/wlr_matrix.h
@@ -44,7 +44,10 @@ void wlr_matrix_transform(float mat[static 9],
 	enum wl_output_transform transform);
 
 /** Writes a 2D orthographic projection matrix to mat of (width, height) with a
- *  specified wl_output_transform*/
+ *  specified wl_output_transform.
+ *
+ *  Deprecated: this function is deprecated and will be removed in a future
+ *  version of wlroots. */
 void wlr_matrix_projection(float mat[static 9], int width, int height,
 	enum wl_output_transform transform);
 


emersion:
I wonder if we can do anything else to make it more obvious that a function is deprecated. There's a GCC attribute, but it's GNU-specific. @kennylevinsen, thoughts?

bl4ckb0ne:
We could output a log message in `WLR_DEBUG` in the function to indicate the deprecation.

kennylevinsen:
Hmm. Deprecation is [supposed to be part of C23](https://en.cppreference.com/w/c/language/attributes/deprecated), but we'll have to hack something up in the meantime.

Options I can think of:
1. Rename it to `wlr_matrix_projection_deprecated` or similar. Minor up-front breakage with very explicit effect.
2. Hide the implementation behind a `#ifdef WLR_USE_DEPRECATED_MATRIX_PROJECTION`. Note that the define should be explicit, as a generic `WLR_USE_DEPRECATED` would just be set once and then hide all deprecation attempts.
3. Just a comment.

None is as elegant as a deprecation attribute, but I don't think 1 is too bad. The up-front cost doubles as explicit notification, and the fixed name ensures that the intent for deprecation remains clear and doesn't get lost in a `#define` prologue somewhere.

emersion:
We can't just break downstream builds, because this is a stable API. See the top of the header:

```c
/*
 * This is a stable interface of wlroots. Future changes will be limited to:
 *
 * - New functions
 * - New struct members
 * - New enum members
 *
 * Note that wlroots does not make an ABI compatibility promise - in the future,
 * the layout and size of structs used by wlroots may change, requiring code
 * depending on this header to be recompiled (but not edited).
 *
 * Breaking changes are announced by email and follow a 1-year deprecation
 * schedule. Send an email to ~sircmpwn/wlroots-announce+subscribe@lists.sr.ht
 * to receive these announcements.
 */
```

emersion:
Note, the mailing list isn't really used. Maybe asking people to look at the release notes would be better.

bl4ckb0ne:
I like the option to have the prototype hidden under `WLR_USE_DEPRECATED`, that gives one or two releases of warning to the user before removing it from the API.

emersion:
> I like the option to have the prototype hidden under WLR_USE_DEPRECATED, that gives one or two releases of warning to the user before removing it from the API.

The problem with this approach is that marking a stable function as deprecated breaks downstream builds, because `WLR_USE_DEPRECATED` is opt-in. We could make it opt-out (`WLR_HIDE_DEPRECATED`, just like Wayland has `WL_HIDE_DEPRECATED`), but then not sure it has real value.

> We could also use a #warning directive to inform the user at compile time that this function will be gone soon.

We'd want to print the warning only if the compositors use the deprecated function. Not sure how to do this. Additionally, we'd want a way to disable the deprecation warnings (to allow compositors which set `-Werror` to continue to build), but this sounds easier to implement.

* * *

All in all, I think a warning in the header and in the release notes is fine. I'd also be OK with using the GNU-specific attribute hidden behind a `#ifdef __GNUC__`, just like we already have `_WLR_ATTRIB_PRINTF`.

progandy:
> The problem with this approach is that marking a stable function as deprecated breaks downstream builds, because `WLR_USE_DEPRECATED` is opt-in. We could make it opt-out (`WLR_HIDE_DEPRECATED`, just like Wayland has `WL_HIDE_DEPRECATED`), but then not sure it has real value.

"```1-year deprecation schedule```" sounds like it could be a gradual process. First only warnings, maybe after half a year you have to opt in to use deprecated functions and after a year the function is gone.

emersion:
Alright, let's just merge this as-is. We can always improve the deprecation process as we go.

PR #2847, opened 2021-04-12
backend: move get_drm_fd to public interface

subdiff:
As discussed on IRC: The get_drm_fd was made available in an internal header with a53ab146f. Move it now to the public header so consumers opting in to the unstable interfaces can make use of it.

### DIFF
--- a/include/wlr/backend.h
+++ b/include/wlr/backend.h
@@ -57,5 +57,12 @@ struct wlr_session *wlr_backend_get_session(struct wlr_backend *backend);
  * Returns the clock used by the backend for presentation feedback.
  */
 clockid_t wlr_backend_get_presentation_clock(struct wlr_backend *backend);
+/**
+ * Returns the DRM node file descriptor used by the backend's underlying
+ * platform. Can be used by consumers for additional rendering operations.
+ * The consumer must not close the file descriptor since the backend continues
+ * to have ownership of it.
+ */
+int wlr_backend_get_drm_fd(struct wlr_backend *backend);
 
 #endif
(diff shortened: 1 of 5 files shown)


include/wlr/backend.h
```diff
@@ -57,5 +57,10 @@ struct wlr_session *wlr_backend_get_session(struct wlr_backend *backend);
  * Returns the clock used by the backend for presentation feedback.
  */
 clockid_t wlr_backend_get_presentation_clock(struct wlr_backend *backend);
+/**
+ * Returns the file descriptor that is internally opened for rendering.
```
emersion, line 61: "rendering" seems a little bit off. In the future the backend will be completely decoupled from the renderer and any rendering operation.

The DRM backend uses this FD for KMS operations. Other backends don't really use the FD, it's just relayed as a hint for the compositor's renderer, so that wlroots and the parent X11 server or Wayland compositor can use the same device.

Maybe "used by the backend's underlying platform" would be more accurate?

emersion, line 61: I think it would be good to explicitly spell out that this is a DRM node file descriptor. And also that the FD is owned by the backend, the caller must not close it.

emersion (changes requested):
LGTM apart from the doc nits.

include/wlr/backend.h
subdiff: What do you think of the following text:

"Returns the DRM node file descriptor used by the backend's underlying platform. Can be used by consumers for additional rendering operations. The consumer must not close the file descriptor since the backend continues to have ownership of it."

emersion: LGTM!

emersion (approved):
Thanks!

Issue #2664, opened 2021-01-19
Stop importing DMA-BUFs on wl_surface.commit


emersion:
Right now, we create `wlr_client_buffer` objects on surface commit. This works fine for `wl_shm` buffers (especially when we can just update the previous buffer), however this imports DMA-BUFs via EGL too often. We'd only need to import them once when the `wl_buffer` is created, and then we could re-use the same `EGLImage` for all subsequent commits.

Some drivers are known to have slow DMA-BUF import paths. This also makes us import too often the DMA-BUFs to KMS when using direct scan-out.

emersion:
Well, apparently re-importing the DMA-BUFs is necessary on some drivers to see the changes to the underlying buffer.

emersion:
However, when we use external EGL textures, the underlying changes are guaranteed to be visible without re-importing.

Also, it seems like to see the external changes for a regular texture (not external), we don't need to re-import the DMA-BUF as an EGLImage -- we can just re-create the texture from the same EGLImage.

J0nnyMak0:
This proposal sounds like it can potentially fix https://github.com/swaywm/wlroots/issues/2636?

emersion:
I don't think so. Before the regression we were already importing DMA-BUFs on each commit.

emersion:
We could add a `wlr_texture_invalidate` function. However, it's important to note that shm buffers work pretty differently from DMA-BUFs.

With the GLES2 renderer and a DMA-BUF texture:

- On `wl_buffer` creation: import as an `EGLImage`
- On `wl_surface.commit`: re-create the texture if it's external

With the GLES2 renderer and a shm texture:

- On `wl_buffer` creation: do nothing
- On `wl_surface.commit`: upload the damaged region to the texture

With the Pixman renderer and a shm texture:

- On `wl_buffer` creation: we can't hook into that, it's all handled inside libwayland
- On `wl_surface.commit`: create the pixman image
- When accessing the shm buffer, make sure to call `wl_shm_buffer_{begin,end}_access`

emersion:
https://github.com/swaywm/wlroots/pull/2851 tries to address this issue by passing a `wlr_buffer` reference to the renderer for upload/import. The renderer can then make the decision to leave the buffer unlocked (upload: shm+GLES2) or lock the buffer (import: DMA-BUF+GLES2 or shm+Pixman).

[...]

PR #2851, opened 2021-04-12
Cache and re-use DMA-BUF textures

emersion:
If a client creates a `wl_buffer` and attaches it to a `wl_surface`, there's a good chance they're going to re-use it.

References: https://github.com/swaywm/wlroots/issues/2664

### DIFF
--- a/render/gles2/texture.c
+++ b/render/gles2/texture.c
@@ -125,6 +125,20 @@ static const struct wlr_texture_impl texture_impl = {
 	.destroy = gles2_texture_destroy,
 };
 
+static struct wlr_gles2_texture *gles2_texture_create(
+		struct wlr_gles2_renderer *renderer, uint32_t width, uint32_t height) {
+	struct wlr_gles2_texture *texture =
+		calloc(1, sizeof(struct wlr_gles2_texture));
+	if (texture == NULL) {
+		wlr_log_errno(WLR_ERROR, "Allocation failed");
+		return NULL;
+	}
+	wlr_texture_init(&texture->wlr_texture, &texture_impl, width, height);
+	texture->renderer = renderer;
+	wl_list_insert(&renderer->textures, &texture->link);
+	return texture;
+}
+
 struct wlr_texture *gles2_texture_from_pixels(struct wlr_renderer *wlr_renderer,
 		uint32_t drm_format, uint32_t stride, uint32_t width,
 		uint32_t height, const void *data) {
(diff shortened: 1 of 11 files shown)


emersion:
Been testing this for a while, seems to work fine so far. Apart from the minor performance gains for most users, the main motivation for this is:

- This PR results significant improvements on multi-GPU setups. It seems like cross-vendor EGL imports benefit a lot from this. We still have other bugs with multi-GPU though.
- `wlr_texture_from_buffer` allows us to remove GL-centric assumptions from `wlr_client_buffer_import`. #2892 takes advantage of it.

I'll self-review this one more time before marking the PR as ready.

One gotcha is that with this change the GLES2 backend ref'counts `wlr_texture`s. So two calls that create a `wlr_texture` may return the same pointer, and `wlr_texture_destroy` may not actually destroy the texture. The Pixman renderer doesn't have this ref'counting implemented even in #2892, since `pixman_image_create` is pretty cheap.

render/wlr_texture.c
```diff
@@ -46,6 +47,14 @@ struct wlr_texture *wlr_texture_from_dmabuf(struct wlr_renderer *renderer,
 	return renderer->impl->texture_from_dmabuf(renderer, attribs);
 }
 
+struct wlr_texture *wlr_texture_from_buffer(struct wlr_renderer *renderer,
+		struct wlr_buffer *buffer) {
```
Emantor, line 51: Maybe add `assert(!renderer->rendering)`?

Emantor (approved):

render/wlr_texture.c
emersion: Good idea

render/gles2/texture.c
```diff
@@ -318,7 +360,48 @@ struct wlr_texture *gles2_texture_from_dmabuf(struct wlr_renderer *wlr_renderer,
...
+		struct wlr_buffer *buffer) {
+	struct wlr_gles2_renderer *renderer = gles2_get_renderer(wlr_renderer);
+
+	struct wlr_dmabuf_attributes dmabuf;
+	if (!wlr_buffer_get_dmabuf(buffer, &dmabuf)) {
+		return false;
+	}
+
+	struct wlr_gles2_texture *texture;
+	wl_list_for_each(texture, &renderer->textures, link) {
```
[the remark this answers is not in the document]
emersion: That's a recurring theme in wlroots, and I've been thinking for a while that something like a hashmap would allow us to optimize more cases. I've written down a proposal here: https://github.com/swaywm/wlroots/issues/2927

kennylevinsen (approved):

emersion:
Thanks for the reviews!

PR #2853, opened 2021-04-14
xwayland/sockets: check directory permissions

### LINKED ISSUE #2793: xwayland: consider not mkdir /tmp/.X11-unix
This directory should be owned by root and have the sticky bit.

Ref:

- https://invent.kde.org/plasma/kwin/-/merge_requests/690
- https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1787

https://github.com/swaywm/wlroots/blob/5e19e0053a5800252a27ce127fd015d641ee2e9e/xwayland/sockets.c#L91

r-c-f:
Rather than mkdir(), ensure that the xwayland socket directory is owned
by root and has the sticky bit set. Closes #2793.


emersion:
The code LGTM, but I have one question: the Mutter MR will still try to create the directory as the current user if it doesn't exist. I guess this can be useful for simple single-user setups without systemd. Creating the directory ourselves will prevent other users from creating an X11 listening socket. Should we do this too?

r-c-f:
For the simplest single-user systems this is indeed useful.

xwayland/sockets.c
```diff
@@ -84,11 +84,44 @@ static int open_socket(struct sockaddr_un *addr, size_t path_size) {
...
+	}
+	return true;
+}
+
 static bool open_sockets(int socks[2], int display) {
 	struct sockaddr_un addr = { .sun_family = AF_UNIX };
 	size_t path_size;
 
-	mkdir(socket_dir, 0777);
+	if (mkdir(socket_dir, 0755) == 0) {
```
emersion, line 115: Is it dangerous to not set the sticky bit here?

Mutter seems to create the socket with mode 01777, and with a non-root owner. This seems dangerous, since this will potentially allow the current user to mess up other users' X11 sockets.

Setting the mode 0755 makes it clear that the directory is to be used exclusively by the current user, since other users won't be able to create files in it.

So I think our code is safer. Am I missing something?

xwayland/sockets.c
```diff
@@ -84,11 +84,44 @@ static int open_socket(struct sockaddr_un *addr, size_t path_size) {
...
+	if (!(buf.st_mode & S_IFDIR)) {
+		wlr_log(WLR_ERROR, "%s is not a directory", socket_dir);
+		return false;
+	}
+	if (!((buf.st_uid == 0) || (buf.st_uid == getuid()))) {
+		wlr_log(WLR_ERROR, "%s not owned by root or us", socket_dir);
+		return false;
+	}
+	if (!(buf.st_mode & S_ISVTX)) {
+		if ((buf.st_mode & (S_IWGRP | S_IWOTH))) {
```
emersion, line 103: Note to self: "Bail out if the sticky bit is not set and the directory is writable by other users"

(Maybe would be nice to turn this into a comment.)

emersion, line 103: Should we maybe bail out if the sticky bit is set, but the directory is not owned by root? This would detect if someone has already set up the X11 socket dir in a way that would be dangerous to other users.

The only motivation for this would be to be nice to other users, I guess. So not that of a big deal.

emersion (commented):
Sorry for the long review, since this is a security-sensitive patch I just keep wondering whether the code can be exploited.

This approach looks good to me, and the comments are minor. I'm fine with merging this as-is. Can you squash all of the commits into a single one?

xwayland/sockets.c
r-c-f: If the current user is the owner, they are never going to be hindered by any mode as they can change it at will. The advantage here is rather in avoiding the trouble of setting the sticky bit portably without ignoring a more restrictive umask (which I always tend to find rather distasteful) if we're content catering to the simplest single user setups (and will ourselves refuse to use another non-root socket directory regardless).

emersion (approved):
LGTM, thanks for the fix!

Vogtinator:
LGTM!

emersion:
Thanks for double-checking, @Vogtinator

PR #2856, opened 2021-04-14
WIP: buffer: drop wlr_client_buffer

emersion:
`wlr_client_buffer` is weird because it's a `wlr_buffer` implementation used for all `wl_buffer` types, created on `wl_surface.commit` and destroyed on the next one. It wraps a `wlr_texture` and forces some assumptions on the renderer.

Split out of https://github.com/swaywm/wlroots/pull/2851 because that's a pretty intrusive change.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2856

PR #2859, opened 2021-04-15
render/gbm_allocator: make wlr_gbm_allocator_create return a wlr_allocator

bl4ckb0ne:
This patch will prepare the field for dealing with multiple allocators (`drm_dumb` and `shm`) in the coming pixman renderer #2661 

/cc @emersion

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -62,7 +62,7 @@ void finish_drm_renderer(struct wlr_drm_renderer *renderer) {
 		return;
 	}
 
-	wlr_allocator_destroy(&renderer->allocator->base);
+	wlr_allocator_destroy(renderer->allocator);
 	wlr_renderer_destroy(renderer->wlr_rend);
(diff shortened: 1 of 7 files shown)


emersion:
The reason why `wlr_gbm_allocator_create` returns ` wlr_gbm_allocator` is to provide type-safety if the caller wants to do something GBM-specific with the backend. For instance we could have `wlr_gbm_allocator_get_gbm_device` in the future, and it would be nice if this takes a `wlr_gbm_allocator` instead of a generic `wlr_allocator`. But this is all theoretical at this point, and it's still a private interface, so I'm fine with returning a `wlr_allocator` too.

emersion:
Note, if we introduce `wlr_allocator_autocreate`, we wouldn't really need this change anymore.

bl4ckb0ne:
Superseded by https://github.com/swaywm/wlroots/pull/2884

PR #2863, opened 2021-04-17
examples: introduce quads example

bl4ckb0ne:
This examples uses `wlr_render_quad_with_matrix` to render coloured
squares on the screen, and uses the rotation to make them spin on their
middle.

The code is copy pasted from `examples/rotation`, with the arguments removed and a new `output_frame_notify` so maintenance should not be too hard.

The idea behind this is to have an example that uses `render_quad` without having a tablet around.

### DIFF
--- a/examples/quads.c
+++ b/examples/quads.c
@@ -1,0 +1,212 @@
+#define _XOPEN_SOURCE 600
+#include <getopt.h>
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <time.h>
+#include <unistd.h>
+#include <wayland-server-core.h>
+#include <wlr/backend.h>
+#include <wlr/render/wlr_renderer.h>
+#include <wlr/types/wlr_keyboard.h>
+#include <wlr/types/wlr_output.h>
+#include <wlr/types/wlr_input_device.h>
+#include <wlr/types/wlr_matrix.h>
+#include <wlr/util/log.h>
+#include <xkbcommon/xkbcommon.h>
+
+struct sample_state {
+	struct wl_display *display;
+	struct wl_listener new_output;
+	struct wl_listener new_input;
+	struct timespec last_frame;
+	struct wlr_renderer *renderer;
+	struct wl_list outputs;
+};
+
+struct sample_output {
+	struct sample_state *sample;
+	struct wlr_output *output;
+	struct wl_listener frame;
+	struct wl_listener destroy;
+	struct wl_list link;
+};
+
+struct sample_keyboard {
+	struct sample_state *sample;
+	struct wlr_input_device *device;
+	struct wl_listener key;
+	struct wl_listener destroy;
+};
+
+static void output_frame_notify(struct wl_listener *listener, void *data) {
+	struct sample_output *sample_output = wl_container_of(listener, sample_output, frame);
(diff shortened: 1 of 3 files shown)


fluix-dev:
https://github.com/swaywm/wlroots/pull/2860 should probably be applied to this as well.

bl4ckb0ne:
I integrated the patch, thanks for the tip.

examples/quads.c
```diff
@@ -0,0 +1,216 @@
+#define _POSIX_C_SOURCE 200112L
+#define _XOPEN_SOURCE 600
+#include <drm_fourcc.h>
+#include <GLES2/gl2.h>
```
emersion, line 4: This is unused.

examples/quads.c
```diff
@@ -0,0 +1,216 @@
+#define _POSIX_C_SOURCE 200112L
+#define _XOPEN_SOURCE 600
```
emersion, line 2: Specify one of these two, but not both. `_XOPEN_SOURCE` automatically enables a specific version of `_POSIX_C_SOURCE`, specifying both with a mismatch confuses some libc implementations.

examples/quads.c
```diff
@@ -0,0 +1,216 @@
...
+	wl_list_remove(&sample_output->frame.link);
+	wl_list_remove(&sample_output->destroy.link);
+	free(sample_output);
+}
+
+static void new_output_notify(struct wl_listener *listener, void *data) {
+	struct wlr_output *output = data;
+	struct sample_state *sample = wl_container_of(listener, sample, new_output);
+	struct sample_output *sample_output = calloc(1, sizeof(struct sample_output));
+	if (!wl_list_empty(&output->modes)) {
```
emersion, line 111: Using `wlr_output_preferred_mode` is cleaner here. It sounds like only `simple` was using it, fixed in https://github.com/swaywm/wlroots/commit/fd7e565ce3ea463b725fcd5b4291411f030a4e2c.

examples/quads.c
```diff
@@ -0,0 +1,216 @@
+#define _POSIX_C_SOURCE 200112L
+#define _XOPEN_SOURCE 600
+#include <drm_fourcc.h>
```
emersion, line 3: I don't think we needs this here.

examples/quads.c
```diff
@@ -0,0 +1,216 @@
...
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <time.h>
+#include <unistd.h>
+#include <wayland-server-core.h>
+#include <wlr/backend.h>
+#include <wlr/backend/session.h>
```
emersion, line 15: This is unused as well.

examples/quads.c
```diff
@@ -0,0 +1,216 @@
...
+#include <wlr/render/wlr_renderer.h>
+#include <wlr/types/wlr_keyboard.h>
+#include <wlr/types/wlr_output.h>
+#include <wlr/types/wlr_output_layout.h>
```
emersion, line 19: Thsi is unused.

emersion (changes requested):
Otherwise looks good!

examples/quads.c
bl4ckb0ne: I removed the unused includes in `examples/rotation` as well.

emersion (approved):
Thanks!

PR #2868, opened 2021-04-17
xwayland: query window PIDs via XResQueryClientIds

### LINKED ISSUE #2189: Use the XRes protocol instead of _NET_WM_PID
_NET_WM_PID is set by the client. This has issues:

- Clients can lie
- Doesn't work when the client is sandboxed in a PID namespace
- Doesn't work with remote clients

Using XRes would allow us to get the info directly form the X11 server: https://gitlab.freedesktop.org/xorg/proto/xorgproto/blob/master/resproto.txt

alebastr:
Fixes #2189


emersion (approved):
Thanks!

PR #2872, opened 2021-04-18
Session cleanup

kennylevinsen:
See the individual commits.


backend/session/session.c
```diff
@@ -310,6 +315,7 @@ void wlr_session_close_file(struct wlr_session *session,
 	if (libseat_close_device(session->seat_handle, dev->device_id) == -1) {
 		wlr_log_errno(WLR_ERROR, "Failed to close device %d", dev->device_id);
 	}
+	close(dev->fd);
```
emersion, line 318: Ah, good catch. This is something we were doing for other sessions back then we had multiple of them: https://github.com/swaywm/wlroots/blob/d5105c42e34439cc6e1084bd8b04712b74e467c0/backend/session/logind.c#L112

emersion (approved):
LGTM, thanks!

PR #2875, opened 2021-04-19
backend/x11: add support for shm buffers

### LINKED ISSUE #2866: backend/x11: add support for shm buffers
With https://github.com/swaywm/wlroots/issues/2399 we have a new Pixman renderer, but it's not yet usable with the X11 backend.

We need to handle shm `wlr_buffer`s. We probably want to use MIT-SHM's `AttachFd`.

tadeokondrak:
Closes #2866.

This doesn't work in Xwayland for me, since `shared_pixmaps` is 0 when running there.
I'm not sure if that's the case for Xwayland in general.

### DIFF
--- a/backend/x11/backend.c
+++ b/backend/x11/backend.c
@@ -17,6 +17,7 @@
 #include <xcb/dri3.h>
 #include <xcb/present.h>
 #include <xcb/render.h>
+#include <xcb/shm.h>
 #include <xcb/xcb_renderutil.h>
 #include <xcb/xfixes.h>
 #include <xcb/xinput.h>
@@ -32,6 +33,7 @@
 #include "backend/x11.h"
 #include "render/drm_format_set.h"
 #include "render/gbm_allocator.h"
+#include "render/shm_allocator.h"
 #include "render/wlr_renderer.h"
 #include "util/signal.h"
 
@@ -198,6 +200,7 @@ static void backend_destroy(struct wlr_backend *backend) {
 	wlr_renderer_destroy(x11->renderer);
 	wlr_allocator_destroy(x11->allocator);
 	wlr_drm_format_set_finish(&x11->dri3_formats);
+	wlr_drm_format_set_finish(&x11->shm_formats);
 	free(x11->drm_format);
 
 #if HAS_XCB_ERRORS
@@ -341,18 +344,24 @@ static bool query_dri3_modifiers(struct wlr_x11_backend *x11,
 	return true;
 }
 
-static bool query_dri3_formats(struct wlr_x11_backend *x11) {
+static bool query_formats(struct wlr_x11_backend *x11) {
 	xcb_depth_iterator_t iter = xcb_screen_allowed_depths_iterator(x11->screen);
 	while (iter.rem > 0) {
 		uint8_t depth = iter.data->depth;
 
 		const struct wlr_x11_format *format = x11_format_from_depth(depth);
 		if (format != NULL) {
-			wlr_drm_format_set_add(&x11->dri3_formats, format->drm,
-				DRM_FORMAT_MOD_INVALID);
+			if (x11->have_shm) {
+				wlr_drm_format_set_add(&x11->shm_formats, format->drm,
+					DRM_FORMAT_MOD_INVALID);
+			}
 
-			if (!query_dri3_modifiers(x11, format)) {
-				return false;
+			if (x11->have_dri3) {
+				wlr_drm_format_set_add(&x11->dri3_formats, format->drm,
+					DRM_FORMAT_MOD_INVALID);
+				if (!query_dri3_modifiers(x11, format)) {
+					return false;
+				}
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


emersion:
I tried a quick hack to enable MIT-SHM shared pixmaps in Xwayland, it seems to work well so far.

<details>

<summary>Hack</summary>

```diff
diff --git a/hw/xwayland/xwayland-present.c b/hw/xwayland/xwayland-present.c
index 666ea15e7aaf..2cc2728dd569 100644
--- a/hw/xwayland/xwayland-present.c
+++ b/hw/xwayland/xwayland-present.c
@@ -400,10 +400,14 @@ xwl_present_check_flip2(RRCrtcPtr crtc,
 {
     struct xwl_window *xwl_window = xwl_window_from_window(present_window);
     ScreenPtr screen = pixmap->drawable.pScreen;
+    struct wl_buffer *buffer = xwl_glamor_pixmap_get_wl_buffer(pixmap);
 
     if (!xwl_window)
         return FALSE;
 
+    if (!buffer)
+        return FALSE;
+
     /* Can't flip if the window pixmap doesn't match the xwl_window parent
      * window's, e.g. because a client redirected this window or one of its
      * parents.
diff --git a/hw/xwayland/xwayland-screen.c b/hw/xwayland/xwayland-screen.c
index c18e2fbb5a6f..bb18e5c94fbc 100644
--- a/hw/xwayland/xwayland-screen.c
+++ b/hw/xwayland/xwayland-screen.c
@@ -52,6 +52,10 @@
 #include "xwayland-present.h"
 #include "xwayland-shm.h"
 
+#ifdef MITSHM
+#include "shmint.h"
+#endif
+
 #include "xdg-output-unstable-v1-client-protocol.h"
 #include "viewporter-client-protocol.h"
 #include "xdg-shell-client-protocol.h"
@@ -662,6 +666,10 @@ xwl_screen_init(ScreenPtr pScreen, int argc, char **argv)
 
     fbPictureInit(pScreen, 0, 0);
 
+#ifdef MITSHM
+    ShmRegisterFbFuncs(pScreen);
+#endif
+
 #ifdef HAVE_XSHMFENCE
     if (!miSyncShmScreenInit(pScreen))
         return FALSE;
```

</summary>

backend/x11/backend.c
```diff
@@ -560,31 +595,41 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
-		close(drm_fd);
-		goto error_event;
+		struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
+		if (gbm_alloc == NULL) {
+			wlr_log(WLR_ERROR, "Failed to create GBM allocator");
+			close(drm_fd);
+			goto error_event;
+		}
+		x11->allocator = &gbm_alloc->base;
+	} else {
```
emersion, line 624: Should we error out if neither DRI3 nor SHM is available?

backend/x11/backend.c
```diff
@@ -606,22 +651,26 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
+	if (!query_formats(x11)) {
+		wlr_log(WLR_ERROR, "Failed to query supported DRM formats");
 		return false;
 	}
 
-	const struct wlr_drm_format *dri3_format =
-		wlr_drm_format_set_get(&x11->dri3_formats, x11->x11_format->drm);
-	if (dri3_format == NULL) {
+	const struct wlr_drm_format *x11_format =
+		wlr_drm_format_set_get(&x11->shm_formats, x11->x11_format->drm);
```
emersion, line 660: This effectively disables DRI3 format modifiers when SHM is available. `shm_formats` will never have modifiers.

Maybe we can move the logic to pick a `x11_format` up to the `if (x11->have_dri3)` block? `x11_format` is the format used for allocations, so it should come from DRI3 if we're using a GBM allocator, or from SHM if we're using the shm allocator.

In other words, we shouldn't use `shm_formats` when using DRI3, or vice-versa.

backend/x11/meson.build
```diff
@@ -5,6 +5,7 @@ x11_required = [
 	'xcb-present',
 	'xcb-render',
 	'xcb-renderutil',
+	'xcb-shm',
```
emersion, line 8: Can we add this dep to the README?

backend/x11/output.c
```diff
@@ -143,53 +145,99 @@ static void buffer_handle_buffer_destroy(struct wl_listener *listener,
 	destroy_x11_buffer(buffer);
 }
 
-static struct wlr_x11_buffer *create_x11_buffer(struct wlr_x11_output *output,
-		struct wlr_buffer *wlr_buffer) {
+static xcb_pixmap_t import_dmabuf(struct wlr_x11_output *output,
+		struct wlr_buffer *wlr_buffer, struct wlr_dmabuf_attributes *dmabuf) {
```
emersion, line 149: It seems like `wlr_buffer` is unused.

backend/x11/output.c
```diff
@@ -143,53 +145,99 @@ static void buffer_handle_buffer_destroy(struct wl_listener *listener,
...
-			attrs.stride[0], x11_fmt->depth, x11_fmt->bpp, dup_attrs.fd[0]);
+			dmabuf->height * dmabuf->stride[0], dmabuf->width, dmabuf->height,
+			dmabuf->stride[0], x11_fmt->depth, x11_fmt->bpp, dup_attrs.fd[0]);
+	}
+
+	return pixmap;
+}
+
+static xcb_pixmap_t import_shm(struct wlr_x11_output *output,
+		struct wlr_buffer *wlr_buffer, struct wlr_shm_attributes *shm) {
```
emersion, line 198: Ditto, `wlr_buffer` seems unused here.

emersion (changes requested):
Thanks for working on this, looks pretty good apart from these comments!

backend/x11/output.c
```diff
@@ -143,53 +145,99 @@ static void buffer_handle_buffer_destroy(struct wl_listener *listener,
...
+
+	// xcb closes the FD after sending it
+	int fd = fcntl(shm->fd, F_DUPFD_CLOEXEC, 0);
+	if (fd < 0) {
+		wlr_log_errno(WLR_ERROR, "fcntl(F_DUPFD_CLOEXEC) failed");
+		return XCB_PIXMAP_NONE;
+	}
+
+	xcb_shm_seg_t seg = xcb_generate_id(x11->xcb);
+	xcb_shm_attach_fd(x11->xcb, seg, fd, false);
```
emersion, line 215: I think it's desirable to set the read-only flag to true here. We don't want the X11 server to write to our buffers.

tadeokondrak: It looks like CreatePixmap needs a writable segment: https://gitlab.freedesktop.org/xorg/xserver/-/blob/7e142cb2a848acb6af986fa91d254d4c23963b24/Xext/shm.c#L1076

backend/x11/backend.c
tadeokondrak: I changed it to use a single format set for formats that can be used with either SHM or DRI3, not sure if `importable_formats` is a good name for that.

emersion: Hm, I'd prefer to have the two sets separate. In the future the allocator won't be initialized in the backend impl anymore, and I'd like to be able to switch the X11 window depth at runtime.

tadeokondrak: I think it's better now, still unsure about the name, having two `x11_format`s doesn't seem great so I went with `pixmap_format`.

backend/x11/backend.c
```diff
@@ -560,31 +597,65 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
-		goto error_event;
+	if (!query_formats(x11)) {
+		wlr_log(WLR_ERROR, "Failed to query supported DRM formats");
+		return false;
 	}
 
-	char *drm_name = drmGetDeviceNameFromFd2(x11->drm_fd);
-	wlr_log(WLR_DEBUG, "Using DRM node %s", drm_name);
-	free(drm_name);
+	const struct wlr_drm_format *pixmap_format = NULL;
```
emersion, line 605: Instead of a `struct wlr_drm_format`, we can declare a `struct wlr_drm_format_set` variable here, set it to either `dri3_formats` or `shm_formats` in each `if` branch, and de-duplicate the `wlr_drm_format_set_get` call. Does that make sense?

backend/x11/output.c
emersion: Ah, makes sense then!

emersion (approved):
LGTM, thanks for your contribution!

PR #2771, opened 2021-03-06
Implement vulkan renderer

nyorain:
Reworked version of #1214, building upon the new wlr_allocator abstraction. A lot of the old discussion points have been resolved.
- the renderer has a hard dependency on VK_EXT_image_drm_format_modifier. No undefined image importing hackery or linear image layouts anymore
- the old wlr_render_surface concept is removed, vulkan integrates very well with the wlr_allocator concept :)
- this does no API changes to wlroots at all
- The vulkan renderer is an optional compile-time feature. Only if compiled with vulkan and if the WLR_RENDERER=vulkan environment variable is set, the vulkan renderer will be created by wlr_renderer_autocreate.
- A couple of functions are currently not implemented: `wlr_renderer_read_pixels`, `wlr_renderer_blit_dmabuf`, `wlr_texture_to_dmabuf`. We can implement all of them but implementations would be poor, I would much rather propose interface changes to implement them efficiently, without completely blocking the compositor. These functions being unimplemented should only have an impact on multi-GPU support and wlr_screencopy_v1 not working under vulkan.
- There are still a couple of smaller improvements noted in the code but nothing of this seems too important to me, it's mostly performance improvements or simplifications.
- This support YCbCr textures (for shm as well as dmabuf texture importing) but I couldn't find a way to test that yet. It does not support rendering to such formats but I don't think that's ever needed.

As this isn't perfect and needs *a lot of* testing on all possible hardware, the renderer currently always enables the vulkan validation layers. You will need [this](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2610) patch as there were errors in the validation layers for VK_EXT_image_drm_format_modifier.
I tested this on an intel 8th gen GPU with the anv mesa MR (there isn't support for importing YCbCr dmabufs yet), using tinywl with the wayland and drm backend, running the sample weston-* applications as well as firefox and some GTK applications. Sadly, I don't have a >=GFX9 (vega) AMD GPU, and since radv already has support for the vulkan extension on those GPUs I'd be thankful for people testing it. Tests on any other driver/platform where the extension is available would certainly be useful as well.

If you can think of applications that this should be tested with, let me know (something that uses YCbCr shm buffers would be nice).

PR #2878, opened 2021-04-20
render: Assert that texture dimensions are > 0

nyorain:
Assert that the case found in #2771 does not happen.

### DIFF
--- a/render/wlr_texture.c
+++ b/render/wlr_texture.c
@@ -22,6 +22,10 @@ void wlr_texture_destroy(struct wlr_texture *texture) {
 struct wlr_texture *wlr_texture_from_pixels(struct wlr_renderer *renderer,
 		uint32_t fmt, uint32_t stride, uint32_t width, uint32_t height,
 		const void *data) {
+	assert(width > 0);
+	assert(height > 0);
+	assert(stride >= width);
+	assert(data);
 	return renderer->impl->texture_from_pixels(renderer, fmt, stride, width,
 		height, data);
 }


render/wlr_texture.c
```diff
@@ -22,6 +22,8 @@ void wlr_texture_destroy(struct wlr_texture *texture) {
 struct wlr_texture *wlr_texture_from_pixels(struct wlr_renderer *renderer,
 		uint32_t fmt, uint32_t stride, uint32_t width, uint32_t height,
 		const void *data) {
+	assert(width > 0);
+	assert(height > 0);
```
emersion, line 26: Maybe also assert that `stride > 0`?

nyorain: `assert(data)` as well?

nyorain: And `stride >= width` rather than `stride > 0`?

emersion: Yeah, I guess that makes sense too

emersion: I'm fine with `stride >= width` yeah, even if it won't catch all errors.

emersion (approved):
Thanks!

ggreer:
I'm not sure why, but this caused Firefox to crash sway (but only if I restored session, not if I created a new profile). I removed the asserts and added logging: https://gist.github.com/ggreer/48540d062c94db18de0a0809fee9c1fa

```
00:00:09.645 [ERROR] [wlr] [render/wlr_texture.c:29] fmt: 875713089 stride: 0 width: 39886 height: 32
```

I then removed the `assert(data)` line and Firefox started up just fine. I must have closed the tab(s) responsible, because even after going back to wlroots master, I can no longer cause the crash.

render/wlr_texture.c
```diff
@@ -22,6 +22,10 @@ void wlr_texture_destroy(struct wlr_texture *texture) {
...
+	assert(stride >= width);
```
emersion, line 27: Now that I think of it, the stride can be smaller than the width for formats with a number of bits per pixel lower than 8. We don't support any of these yet.

emersion:
Hm, a zero stride sounds like a client bug. But we shouldn't die on it. I'm surprised `wl_shm` doesn't catch it.

Issue #2865, opened 2021-04-17
render: introduce WLR_RENDERER env var


emersion:
This allows runtime selection of the renderer in `wlr_renderer_autocreate`.

cc @bl4ckb0ne

PR #2880, opened 2021-04-20
Implement WLR_RENDERER in renderer creation process

bl4ckb0ne:
Fixes https://github.com/swaywm/wlroots/issues/2865

### DIFF
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -1,7 +1,6 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stdlib.h>
-#include <gbm.h>
 #include <wlr/render/interface.h>
 #include <wlr/render/pixman.h>
 #include <wlr/render/wlr_renderer.h>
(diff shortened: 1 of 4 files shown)


render/wlr_renderer.c
```diff
@@ -228,34 +227,39 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
...
+		if (strcmp(name, "pixman") == 0) {
+			return wlr_pixman_renderer_create();
+		}
 
-	struct wlr_egl *egl = wlr_egl_create(EGL_PLATFORM_GBM_KHR, gbm_device);
-	if (egl == NULL) {
-		wlr_log(WLR_ERROR, "Could not initialize EGL");
-		gbm_device_destroy(gbm_device);
-		return NULL;
+		wlr_log(WLR_ERROR, "Invalid WLR_RENDERER value: '%s'", name);
```
emersion, line 241: We should probably return NULL here

render/wlr_renderer.c
```diff
@@ -228,34 +227,39 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
...
 	}
 
-	egl->gbm_device = gbm_device;
+	struct wlr_renderer *renderer = NULL;
+#if WLR_HAS_GLES2_RENDERER
+	renderer = wlr_gles2_renderer_create_with_drm_fd(drm_fd);
+	if (renderer) {
+		goto out;
```
emersion, line 248: This can just be `return renderer`, then no need for `goto`

render/wlr_renderer.c
```diff
@@ -228,33 +227,35 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
...
-	egl->gbm_device = gbm_device;
-
-	struct wlr_renderer *renderer = wlr_gles2_renderer_create(egl);
-	if (!renderer) {
-		wlr_log(WLR_ERROR, "Failed to create GLES2 renderer");
-		wlr_egl_destroy(egl);
-		return NULL;
+	struct wlr_renderer *renderer = NULL;
+#if WLR_HAS_GLES2_RENDERER
+	if((renderer = wlr_gles2_renderer_create_with_drm_fd(drm_fd)) != NULL) {
```
emersion, line 247: Style: missing space between `if` and `(`

render/wlr_renderer.c
```diff
@@ -228,33 +227,35 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
...
+		return renderer;
 	}
-
-	return renderer;
+	wlr_log(WLR_DEBUG, "Failed to create gles2 renderer");
 #endif
 
-	wlr_log(WLR_ERROR, "Failed to initialize any renderer");
+	if((renderer = wlr_pixman_renderer_create()) != NULL) {
```
emersion, line 253: Ditto

render/wlr_renderer.c
```diff
@@ -228,33 +227,35 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 }
 
 struct wlr_renderer *wlr_renderer_autocreate_with_drm_fd(int drm_fd) {
+	const char *name = getenv("WLR_RENDERER");
```
emersion, line 230: Oh. I was about to merge this but then realized: all of this happens in `wlr_renderer_autocreate_with_drm_fd`. This means `WLR_RENDERER` is ignored if the backend doesn't return a DRM FD. It would be nicer to parse `WLR_RENDERER` even when the backend doesn't return a DRM FD.

But let's fix this in a separate PR, it's not that of a big deal.

emersion (approved):
Thanks!

PR #2881, opened 2021-04-21
render: unconditionally read WLR_RENDERER

emersion:
Prior to this commit, WLR_RENDERER was only looked up when the
backend returned a DRM FD.

Make it so WLR_RENDERER is always looked up, so that running wlroots
on a system without a GPU and with WLR_RENDERER=gles2 fails as
expected instead of falling back to the Pixman renderer.

cc @bl4ckb0ne

### DIFF
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -231,6 +231,11 @@ struct wlr_renderer *wlr_renderer_autocreate_with_drm_fd(int drm_fd) {
 	if (name) {
 #if WLR_HAS_GLES2_RENDERER
 		if (strcmp(name, "gles2") == 0) {
+			if (drm_fd < 0) {
+				wlr_log(WLR_ERROR, "Cannot create GLES2 renderer: "
+					"no DRM FD available");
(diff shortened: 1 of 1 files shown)


bl4ckb0ne:
Small nit: I'd prefer the drm fd check to be done in the `wlr_gles2_renderer_create_with_drm_fd` function so the renderer autocreate function just has to deal with NULL wlr_renderers.

LGTM otherwise :)

emersion:
> Small nit: I'd prefer the drm fd check to be done in the wlr_gles2_renderer_create_with_drm_fd function so the renderer autocreate function just has to deal with NULL wlr_renderers.

I haven't gone with this approach because one code-path wants a `WLR_ERROR` log when there is no `drm_fd`, the other wants a `WLR_DEBUG` log. We shouldn't print a `WLR_ERROR` if we're going to try another renderer next.

bl4ckb0ne:
Seems good then, merge when you're ready

emersion:
Thanks for having a look!

PR #2883, opened 2021-04-21
render: destroy textures with renderer

emersion:
This prevents some leaks and stray `wlr_texture`s that don't have a renderer anymore.

cc @bl4ckb0ne for the pixman bits


bl4ckb0ne:
LGTM

Issue #2861, opened 2021-04-16
Improve backend, renderer and allocator initialization


emersion:
This proposal is about solving the following problem: given an existing backend, how do we create a renderer and an allocator suitable for it?

Right now the renderer and allocator are created inside the backend. Each backend can have its own logic to choose a renderer and an allocator. In the future we'd like to move the renderer and allocator initialization out of the backend, to allow compositors to have more control and to allow for third-party renderers and allocators.

The backend is created first. Then the renderer is created from the backend in `wlr_renderer_autocreate` (this allows to switch between hardware and software rendering in this function depending on the backend DRM FD, and allows to have a `WLR_RENDERER` env override). Then we need to create an allocator from the backend and the renderer.

An obvious solution is to have somewhere a list of (backend, renderer, allocator) that work fine together. However this list highly depends on a fixed list of backends and renderers -- it doesn't allow third-party implementations to blend in. The list also needs to record compat for each permutation, so can be quite long and tricky to maintain.

@bl4ckb0ne suggested having a `is_software_renderer` flag that renderers can set. This would allow us to choose between GBM and shm or DRM dumb, but doesn't completely solve the issue since we have no good way to figure out whether the backend wants DRM dumb buffers or shm buffers. There's also the edge-case of llvmpipe, a software renderer behind an OpenGL API that only groks DMA-BUFs.

Another solution would be to have in each backend and renderer a list of supported allocators. However the backends and renderers don't really care about which specific allocator is used, they only care about the capabilities of the `wlr_buffer`s submitted to them. By "capability" I mean the supported functions in `wlr_buffer_impl`.

Here's a recap table:

| Module | Supported buffer(s) |
|-|-|
| GLES2 renderer | dmabuf |
| Vulkan renderer | dmabuf |
| Pixman renderer | data_ptr |
| Wayland backend | dmabuf, shm |
| X11 backend | dmabuf (shm in the future) |
| Headless backend | _any_ (doesn't care) |
| GBM allocator | dmabuf |
| shm allocator | shm, data_ptr |
| DRM dumb buffer allocator | dmabuf, data_ptr |

This proposal is about adding a list of supported buffer capabilities in each backend and renderer. With the following new API:

```c
/**
 * Given a backend and a renderer, automatically create an allocator suitable
 * for both.
 */
struct wlr_allocator *wlr_allocator_autocreate(struct wlr_backend *backend,
	struct wlr_renderer *renderer);

/**
 * Buffer capabilities.
 *
[...]

PR #2884, opened 2021-04-22
wlr_allocator_autocreate

bl4ckb0ne:
Implements https://github.com/swaywm/wlroots/issues/2861

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -12,6 +12,7 @@
 #include <wlr/util/log.h>
 #include <xf86drm.h>
 #include "backend/drm/drm.h"
+#include "types/wlr_buffer.h"
 #include "util/signal.h"
 
 struct wlr_drm_backend *get_drm_backend_from_backend(
@@ -89,12 +90,17 @@ static int backend_get_drm_fd(struct wlr_backend *backend) {
 	}
 }
 
+static uint32_t backend_get_buffer_caps(struct wlr_backend *backend) {
+	return WLR_BUFFER_CAP_DMABUF;
+}
+
 static const struct wlr_backend_impl backend_impl = {
 	.start = backend_start,
 	.destroy = backend_destroy,
 	.get_renderer = backend_get_renderer,
 	.get_presentation_clock = backend_get_presentation_clock,
 	.get_drm_fd = backend_get_drm_fd,
+	.get_buffer_caps = backend_get_buffer_caps,
 };
 
 bool wlr_backend_is_drm(struct wlr_backend *b) {
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -13,6 +13,7 @@
 #include "render/drm_format_set.h"
 #include "render/gbm_allocator.h"
 #include "render/wlr_renderer.h"
+#include "types/wlr_buffer.h"
 #include "util/signal.h"
 
 struct wlr_headless_backend *headless_backend_from_backend(
@@ -92,11 +93,18 @@ static int backend_get_drm_fd(struct wlr_backend *wlr_backend) {
 	return backend->drm_fd;
 }
 
+static uint32_t backend_get_buffer_caps(struct wlr_backend *wlr_backend) {
+	return WLR_BUFFER_CAP_DATA_PTR
+		| WLR_BUFFER_CAP_DMABUF
+		| WLR_BUFFER_CAP_SHM;
+}
+
 static const struct wlr_backend_impl backend_impl = {
 	.start = backend_start,
 	.destroy = backend_destroy,
 	.get_renderer = backend_get_renderer,
 	.get_drm_fd = backend_get_drm_fd,
+	.get_buffer_caps = backend_get_buffer_caps,
 };
 
 static void handle_display_destroy(struct wl_listener *listener, void *data) {
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -220,21 +220,21 @@ struct wlr_backend *wlr_headless_backend_create(struct wl_display *display) {
 		goto error_dup;
 	}
 
-	struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
-	if (gbm_alloc == NULL) {
+	struct wlr_allocator *alloc = wlr_gbm_allocator_create(drm_fd);
+	if (alloc == NULL) {
 		wlr_log(WLR_ERROR, "Failed to create GBM allocator");
 		close(drm_fd);
 		goto error_dup;
 	}
 
-	if (!backend_init(backend, display, &gbm_alloc->base, NULL)) {
+	if (!backend_init(backend, display, alloc, NULL)) {
 		goto error_init;
 	}
 
 	return &backend->backend;
 
 error_init:
-	wlr_allocator_destroy(&gbm_alloc->base);
+	wlr_allocator_destroy(alloc);
 error_dup:
 	close(backend->drm_fd);
 error_drm_fd:
@@ -266,14 +266,14 @@ struct wlr_backend *wlr_headless_backend_create_with_renderer(
 		goto error_dup;
 	}
 
-	struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
-	if (gbm_alloc == NULL) {
+	struct wlr_allocator *alloc = wlr_gbm_allocator_create(drm_fd);
+	if (alloc == NULL) {
 		wlr_log(WLR_ERROR, "Failed to create GBM allocator");
 		close(drm_fd);
 		goto error_dup;
 	}
 
-	if (!backend_init(backend, display, &gbm_alloc->base, renderer)) {
+	if (!backend_init(backend, display, alloc, renderer)) {
 		goto error_init;
 	}
 
@@ -283,7 +283,7 @@ struct wlr_backend *wlr_headless_backend_create_with_renderer(
 	return &backend->backend;
 
 error_init:
-	wlr_allocator_destroy(&gbm_alloc->base);
+	wlr_allocator_destroy(alloc);
 error_dup:
 	close(backend->drm_fd);
 error_drm_fd:
--- a/backend/headless/backend.c
+++ b/backend/headless/backend.c
@@ -120,14 +120,11 @@ static void handle_renderer_destroy(struct wl_listener *listener, void *data) {
 }
 
 static bool backend_init(struct wlr_headless_backend *backend,
-		struct wl_display *display, struct wlr_allocator *allocator,
-		struct wlr_renderer *renderer) {
+		struct wl_display *display, struct wlr_renderer *renderer) {
 	wlr_backend_init(&backend->backend, &backend_impl);
 	backend->display = display;
 	wl_list_init(&backend->outputs);
 	wl_list_init(&backend->input_devices);
-
-	backend->allocator = allocator;
 
 	if (renderer == NULL) {
 		renderer = wlr_renderer_autocreate(&backend->backend);
@@ -137,6 +134,12 @@ static bool backend_init(struct wlr_headless_backend *backend,
 		}
 	}
 	backend->renderer = renderer;
+
+	backend->allocator = wlr_allocator_autocreate(&backend->backend, renderer);
+	if (!backend->allocator) {
+		wlr_log(WLR_ERROR, "Failed to create allocator");
+		return false;
+	}
 
 	const struct wlr_drm_format_set *formats =
 		wlr_renderer_get_render_formats(backend->renderer);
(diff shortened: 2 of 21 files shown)


include/render/wlr_renderer.h
```diff
@@ -21,5 +21,11 @@ bool wlr_renderer_bind_buffer(struct wlr_renderer *r, struct wlr_buffer *buffer)
  */
 const struct wlr_drm_format_set *wlr_renderer_get_render_formats(
 	struct wlr_renderer *renderer);
+/**
+ * Get the supported buffer capabilities.
+ *
+ * This functions returns a bitfield of supported wlr_buffer_cap.
+ */
+uint32_t renderer_get_buffer_caps(struct wlr_renderer *renderer);
```
emersion, line 29: I'd prefer to name this `renderer_get_render_buffer_caps`, to make it clear this queries information about buffers that will be rendered to, not buffers which are textured from.

backend/drm/backend.c
```diff
@@ -87,12 +88,17 @@ static int backend_get_drm_fd(struct wlr_backend *backend) {
 	}
 }
 
+static uint32_t backend_get_buffer_caps(void) {
+	return WLR_BUFFER_CAP_DATA_PTR | WLR_BUFFER_CAP_DMABUF;
```
emersion, line 92: The DRM backend doesn't need `get_data_ptr`.

backend/headless/backend.c
```diff
@@ -257,23 +255,16 @@ struct wlr_backend *wlr_headless_backend_create_with_renderer(
...
-		wlr_log_errno(WLR_ERROR, "fcntl(F_DUPFD_CLOEXEC) failed");
-		goto error_dup;
 	}
 
-	struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
-	if (gbm_alloc == NULL) {
-		wlr_log(WLR_ERROR, "Failed to create GBM allocator");
-		close(drm_fd);
-		goto error_dup;
+	struct wlr_allocator *alloc = wlr_allocator_autocreate(&backend->backend,
```
emersion, line 260: Maybe we can drop this, and let `backend_init` call `wlr_allocator_autocreate`?

backend/wayland/backend.c
```diff
@@ -436,42 +442,27 @@ struct wlr_backend *wlr_wl_backend_create(struct wl_display *display,
...
-			wlr_log(WLR_ERROR, "fcntl(F_DUPFD_CLOEXEC) failed");
-			goto error_drm_fd;
-		}
-
-		struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
-		if (gbm_alloc == NULL) {
-			wlr_log(WLR_ERROR, "Failed to create GBM allocator");
-			close(drm_fd);
-			goto error_drm_fd;
+			wl_event_source_remove(wl->remote_display_src);
```
emersion, line 451: I think this `if` block should returns in a fatal error.

backend/wayland/backend.c
```diff
@@ -495,36 +486,32 @@ struct wlr_backend *wlr_wl_backend_create(struct wl_display *display,
...
 	return &wl->backend;
 
-error_renderer:
-	wlr_renderer_destroy(wl->renderer);
 error_allocator:
 	wlr_allocator_destroy(wl->allocator);
-error_drm_fd:
-	close(wl->drm_fd);
-error_remote_display_src:
-	wl_event_source_remove(wl->remote_display_src);
```
emersion, line 527: We need to remove the event source in the error path. Is there a reason why this line has been removed?

backend/x11/backend.c
```diff
@@ -609,35 +615,17 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
-			wlr_log(WLR_ERROR, "fcntl(F_DUPFD_CLOEXEC) failed");
-			goto error_event;
-		}
-
-		struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
-		if (gbm_alloc == NULL) {
-			wlr_log(WLR_ERROR, "Failed to create GBM allocator");
-			close(drm_fd);
-			goto error_event;
+			if (x11->have_shm) {
```
emersion, line 618: TBH it's probably not worth it to fallback to shm when `query_dri3_drm_fd` fails. On a GPU-less system, `have_dri3` will be false.

emersion:
Can we import the doc-comments from the original issue https://github.com/swaywm/wlroots/issues/2861?

include/wlr/render/interface.h
```diff
@@ -56,6 +56,7 @@ struct wlr_renderer_impl {
 	bool (*init_wl_display)(struct wlr_renderer *renderer,
 		struct wl_display *wl_display);
 	int (*get_drm_fd)(struct wlr_renderer *renderer);
+	uint32_t (*get_buffer_caps)(void);
```
emersion, line 59: This should be renamed to `get_render_buffer_caps` too

render/allocator.c
```diff
@@ -9,6 +18,49 @@ void wlr_allocator_init(struct wlr_allocator *alloc,
...
+	const uint32_t shm_caps = WLR_BUFFER_CAP_SHM | WLR_BUFFER_CAP_DATA_PTR;
+	if ((backend_caps & shm_caps) && (renderer_caps & shm_caps)) {
+		wlr_log(WLR_DEBUG, "Trying to create shm allocator");
+		if ((alloc = wlr_shm_allocator_create()) != NULL) {
+			return alloc;
+		}
+		wlr_log(WLR_DEBUG, "Failed to create shm allocator");
+	}
+
+	/* TODO: drm_dumb allocator
```
emersion, line 51: Maybe remove this commented code for now, we don't have the DRM dumb allocator yet.

render/gbm_allocator.c
```diff
@@ -159,7 +159,11 @@ static struct wlr_gbm_allocator *get_gbm_alloc_from_alloc(
 	return (struct wlr_gbm_allocator *)alloc;
 }
 
-struct wlr_gbm_allocator *wlr_gbm_allocator_create(int fd) {
+struct wlr_allocator *wlr_gbm_allocator_create(int fd) {
+	char *drm_name = drmGetDeviceNameFromFd2(fd);
+	wlr_log(WLR_DEBUG, "Using DRM node %s", drm_name);
```
emersion, line 164: Maybe we can log this on the same line as "Created GBM allocator …" at the end of the function?

render/allocator.c
```diff
@@ -9,6 +18,49 @@ void wlr_allocator_init(struct wlr_allocator *alloc,
 	wl_signal_init(&alloc->events.destroy);
 }
 
+struct wlr_allocator *wlr_allocator_autocreate(struct wlr_backend *backend,
+		struct wlr_renderer *renderer) {
+	wlr_log(WLR_DEBUG, "Picking the appropriate allocator");
```
emersion, line 23: Maybe we can just drop this log line, it doesn't seem like it provides any useful information.

render/allocator.c
```diff
@@ -9,6 +18,49 @@ void wlr_allocator_init(struct wlr_allocator *alloc,
...
+	const uint32_t backend_caps = backend_get_buffer_caps(backend);
+	const uint32_t renderer_caps = renderer_get_render_buffer_caps(renderer);
+	const int drm_fd = wlr_backend_get_drm_fd(backend);
+
+	struct wlr_allocator *alloc = NULL;
+	const uint32_t gbm_caps = WLR_BUFFER_CAP_DMABUF;
+	if ((backend_caps & gbm_caps) && (renderer_caps & gbm_caps)
+			&& (drm_fd != -1)) {
```
emersion, line 31: Style: unnecessary parentheses

render/allocator.c
```diff
@@ -9,6 +18,49 @@ void wlr_allocator_init(struct wlr_allocator *alloc,
 	wl_signal_init(&alloc->events.destroy);
 }
 
+struct wlr_allocator *wlr_allocator_autocreate(struct wlr_backend *backend,
+		struct wlr_renderer *renderer) {
+	wlr_log(WLR_DEBUG, "Picking the appropriate allocator");
+	const uint32_t backend_caps = backend_get_buffer_caps(backend);
```
emersion, line 24: These `const` qualifiers aren't really useful, since these aren't pointers.

backend/wayland/backend.c
bl4ckb0ne: Shouldn't we fall back to software rendering if the drm render node fails to open?

emersion: If `wl->drm_render_name` is not NULL, then the parent compositor has advertised a DRM node. If a DRM node has been advertised but we aren't able to open it, it means that a GPU is available but something really bad happened. I'd prefer to complain loudly than to have a fallback. Users won't expect wlroots to use software rendering if a GPU is available. Users will notice the loud complaining, but won't notice the fallback.

backend/wayland/backend.c
```diff
@@ -495,34 +486,32 @@ struct wlr_backend *wlr_wl_backend_create(struct wl_display *display,
...
 	wl_display_add_destroy_listener(display, &wl->local_display_destroy);
 
 	return &wl->backend;
 
-error_renderer:
-	wlr_renderer_destroy(wl->renderer);
 error_allocator:
 	wlr_allocator_destroy(wl->allocator);
-error_drm_fd:
-	close(wl->drm_fd);
```
emersion, line 525: We still need to `close(wl->drm_fd)` in the error path, otherwise we leak a FD.

backend/wayland/backend.c
```diff
@@ -364,11 +365,16 @@ static int backend_get_drm_fd(struct wlr_backend *backend) {
 	return wl->drm_fd;
 }
 
+static uint32_t backend_get_buffer_caps(void) {
+	return WLR_BUFFER_CAP_DMABUF | WLR_BUFFER_CAP_SHM;
```
emersion, line 369: DMA-BUFs are only supported if `wl->zwp_linux_dmabuf_v1` is not NULL.

backend/x11/backend.c
```diff
@@ -223,11 +224,16 @@ static int backend_get_drm_fd(struct wlr_backend *backend) {
 	return x11->drm_fd;
 }
 
+static uint32_t backend_get_buffer_caps(void) {
+	return WLR_BUFFER_CAP_DMABUF | WLR_BUFFER_CAP_SHM;
```
emersion, line 228: Need to check `x11->have_dri3` and `x11->have_shm` here

backend/x11/backend.c
```diff
@@ -611,33 +617,9 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
-
-		struct wlr_gbm_allocator *gbm_alloc = wlr_gbm_allocator_create(drm_fd);
-		if (gbm_alloc == NULL) {
-			wlr_log(WLR_ERROR, "Failed to create GBM allocator");
-			close(drm_fd);
-			goto error_event;
-		}
-		x11->allocator = &gbm_alloc->base;
 		pixmap_formats = &x11->dri3_formats;
 	} else if (x11->have_shm) {
```
emersion, line 621: We can now support both DRI3 and SHM at the same time, so this can be an `if` instead of an `else if`.

backend/x11/backend.c
```diff
@@ -611,33 +617,9 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
 		x11->drm_fd = -1;
-		struct wlr_shm_allocator *shm_alloc = wlr_shm_allocator_create();
-		if (shm_alloc == NULL) {
-			wlr_log(WLR_ERROR, "Failed to create shared memory allocator");
-			goto error_event;
-		}
-		x11->allocator = &shm_alloc->base;
 		pixmap_formats = &x11->shm_formats;
 	} else {
 		wlr_log(WLR_ERROR,
```
emersion, line 625: No need for this error, `wlr_renderer_autocreate` should figure out on its own this case and return an error.

render/gbm_allocator.c
```diff
@@ -183,10 +183,13 @@ struct wlr_gbm_allocator *wlr_gbm_allocator_create(int fd) {
 		return NULL;
 	}
 
-	wlr_log(WLR_DEBUG, "Created GBM allocator with backend %s",
+	wlr_log(WLR_DEBUG, "Created GBM allocator with bacikend %s",
```
emersion, line 186: Typo: "backend"

backend/x11/backend.c
bl4ckb0ne: So we should store both pixman formats then right?

emersion: Oh. Hm that won't work too well. We must use the `dri3_formats` if the allocator is GBM, `shm_formats` if the allocator is shm. However we don't know whether we'll use GBM or shm prior to `wlr_allocator_autocreate`. Potential solutions:

1. Support either DMA-BUFs or shm, not both at the same time. But this will break if DMA-BUFs are available (disabling shm) and the user forces the Pixman renderer (via `WLR_RENDERER=pixman`).
2. Check the type of the allocator returned by `wlr_allocator_autocreate`, with new functions like `allocator_is_gbm` and `allocator_is_shm`.
3. Do something more involved like `allocator_get_buffer_caps` then check the result to decide which formats to select, but let's keep this for later (if ever), it's over-complicated when there are no third-party allocators.

emersion:
Also needs a rebase :)

backend/x11/backend.c
```diff
@@ -611,43 +619,30 @@ struct wlr_backend *wlr_x11_backend_create(struct wl_display *display,
...
 		pixmap_formats = &x11->dri3_formats;
-	} else if (x11->have_shm) {
-		x11->drm_fd = -1;
-		struct wlr_shm_allocator *shm_alloc = wlr_shm_allocator_create();
-		if (shm_alloc == NULL) {
-			wlr_log(WLR_ERROR, "Failed to create shared memory allocator");
-			goto error_event;
-		}
-		x11->allocator = &shm_alloc->base;
+	} else if (x11->have_shm && (caps & WLR_BUFFER_CAP_DATA_PTR)) {
```
emersion, line 635: The X11 backend needs `WLR_BUFFER_CAP_SHM`, it doesn't use/need/care about `CAP_DATA_PTR`

bl4ckb0ne: Those are renderer caps, pixman gives `CAP_DATA_PTR`.

emersion: Oh right. Hm, I guess that doesn't map that well, we could have buffers with `get_data_ptr` but not `get_shm`, like `wlr_wl_shm_buffer` in https://github.com/swaywm/wlroots/pull/2892. Not a big deal for now since allocators are hard-coded and no allocator will return such a buffer.

emersion (approved):
LGTM, thanks!

Issue #2886, opened 2021-04-24
VMware clipboard integration not working


tomKPZ:
When running wlroots in a VMware guest, clipboard integration is broken in both directions.  The integration works on Gnome/Wayland because the VMware guest additions uses the X11 clipboard to broadcast the host clipboard [1] which gnome-shell then forwards to wayland clients.  However, sway's XWayland blocks these requests [2].  Applying patch [3] fixes the issue.

[1] https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/services/plugins/dndcp/copyPasteUIX11.cpp
[2] https://github.com/swaywm/wlroots/commit/b884025558e750268a06818dc63bc46716c75843
[3] https://gist.github.com/tomKPZ/f4613cc767f1a07537d69c4c1ffaa19b

tomKPZ:
Due to wlr-data-control-unstable-v1 [1], clients without focus can already interact with the clipboard, so the checks seem obsolete and should be removed.

[1] https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-data-control-unstable-v1.xml

emersion:
As you mentioned, VMWare needs to use wlr-data-control-unstable-v1 to be able to set the clipboard. I think this is a legitimate use-case for wlr-data-control-unstable-v1, e.g. KDEConnect uses it too.

PR #2887, opened 2021-04-24
Allow unfocused xwayland clients to access clipboard

tomKPZ:
Fixes [1].  These checks were added in [2] in Dec 2017 to prevent
unfocused xwayland clients from accessing the clipboard.  But the
wlr-data-control-unstable-v1 protocol added in Dec 2018 [3] allows
unfocused clients to access the clipboard anyway
(eg. wl-copy/wl-paste), so the original checks seem obsolete.  Since
the checks are causing issues (eg. for open-vm-tools), this change
removes them.

[1] https://github.com/swaywm/wlroots/issues/2886
[2] https://github.com/swaywm/wlroots/commit/b884025558e750268a06818dc63bc46716c75843
[3] https://github.com/swaywm/wlr-protocols/commit/cdd0eddd54ae302c40f1110572b7e9908c289b2e

### DIFF
--- a/xwayland/selection/incoming.c
+++ b/xwayland/selection/incoming.c
@@ -467,13 +467,6 @@ void xwm_handle_selection_notify(struct wlr_xwm *xwm,
 			xwm_selection_transfer_destroy(transfer);
 		}
 	} else if (event->target == xwm->atoms[TARGETS]) {
-		// No xwayland surface focused, deny access to clipboard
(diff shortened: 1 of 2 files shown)


emersion:
NACK. wlr-data-control-unstable-v1 will be restricted to a few clients in the future. Regular clients which aren't clipboard managers must not use it.

PR #2891, opened 2021-04-26
render/egl: allow creating an llvmpipe EGL context

emersion:
There are use-cases for forcing an llvmpipe EGL context, e.g. CI. Allow users to force it.

Depends on: https://github.com/swaywm/wlroots/pull/2671

### DIFF
--- a/include/render/egl.h
+++ b/include/render/egl.h
@@ -1,0 +1,15 @@
+#ifndef RENDER_EGL_H
+#define RENDER_EGL_H
+
+#include <wlr/render/egl.h>
+
+/**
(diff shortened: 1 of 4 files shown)


emersion:
Hm. So even if we're able to create an llvmpipe EGL context without a DRM FD, we still won't have an allocator we can use with it. We need a new EGL ext to create an `EGLImageKHR` from an existing chunk of memory.

emersion:
We'll need https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/10744

PR #2892, opened 2021-04-26
render/pixman: avoid copy when creating textures

emersion:
This allows the Pixman renderer to stop `memcpy`'ing textures.

- [x] Address the shm client buffer SIGBUS issue

Depends on: https://github.com/swaywm/wlroots/pull/2851

cc @bl4ckb0ne

### DIFF
--- a/types/wlr_buffer.c
+++ b/types/wlr_buffer.c
@@ -214,23 +214,25 @@ struct wlr_client_buffer *wlr_client_buffer_import(
 	struct wlr_texture *texture = NULL;
 	bool resource_released = false;
 
-	struct wl_shm_buffer *shm_buf = wl_shm_buffer_get(resource);
-	if (shm_buf != NULL) {
-		enum wl_shm_format wl_shm_format = wl_shm_buffer_get_format(shm_buf);
-		uint32_t drm_format = convert_wl_shm_format_to_drm(wl_shm_format);
-		int32_t stride = wl_shm_buffer_get_stride(shm_buf);
-		int32_t width = wl_shm_buffer_get_width(shm_buf);
-		int32_t height = wl_shm_buffer_get_height(shm_buf);
+	if (wl_shm_buffer_get(resource) != NULL) {
+		struct wlr_shm_client_buffer *shm_client_buffer =
+			shm_client_buffer_create(resource);
+		if (shm_client_buffer == NULL) {
+			wlr_log(WLR_ERROR, "Failed to create shm client buffer");
+			return NULL;
+		}
 
-		wl_shm_buffer_begin_access(shm_buf);
-		void *data = wl_shm_buffer_get_data(shm_buf);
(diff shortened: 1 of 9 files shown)


emersion:
With this PR, the `memcpy` for client buffers is removed with the Pixman renderer. On my machine, CPU usage goes from 6% down to 3% when playing a fullscreen video.

emersion:
The last commit contains an attempt at fixing the SIGBUS issue. It adds guards to data ptr accesses to allow `wl_shm_buffer` to prevent compositor crashes. To test, start wlroots with the Pixman renderer and then run wleird's sigbus client.

Some questions:

- I'm not 100% happy with the additional state-tracking that `wlr_buffer` needs to do. OTOH, a more elaborate acquire/release mechanism would likely be over-engineering, and `wl_shm_buffer` offers the same `{begin,end}_access` API.
- Should the guard be separate from the data ptr attribute retrieval? e.g. `wlr_buffer_get_data_ptr(buffer, &data, &format, &stride)` + `wlr_buffer_begin_data_ptr_access(buffer)` 
- Should we transition to a more mmap-like interface? e.g. `wlr_buffer_map(buffer, &data, &format, &stride)` + `wlr_buffer_unmap(buffer, data)` (implementations would be free to retain the same mapping under the hood)
- Name bikeshedding?

cc @bl4ckb0ne

types/wlr_buffer.c
```diff
@@ -330,3 +362,88 @@ struct wlr_client_buffer *wlr_client_buffer_apply_damage(
...
+	.destroy = shm_client_buffer_destroy,
+	.begin_data_ptr_access = shm_client_buffer_begin_data_ptr_access,
+	.end_data_ptr_access = shm_client_buffer_end_data_ptr_access,
+};
+
+static void shm_client_buffer_resource_handle_destroy(
+		struct wl_listener *listener, void *data) {
+	struct wlr_shm_client_buffer *buffer =
+		wl_container_of(listener, buffer, resource_destroy);
+	buffer->resource = NULL;
```
emersion, line 412: The `buffer->shm_buffer` is destroyed with the resource. We might still be able to access the underlying data if we ref the shm pool and save all of the shm buffer params. But that's a little awkward, maybe it would be nicer to improve the libwayland API instead.

emersion: Fixed.

emersion:
> Should the guard be separate from the data ptr attribute retrieval?

No: in the case of `wl_shm_buffer`, the data pointer is not static and may change if the client resizes the shm pool.

> Should we transition to a more mmap-like interface?

We can't easily provide such an interface with `wl_shm_buffer`. We may be able to do it if we re-implement `wl_shm`.

emersion:
This is now ready for review.

bl4ckb0ne (approved):

Issue #2864, opened 2021-04-17
buffer: wlr_buffer_impl.get_data_ptr limitations


emersion:
We've introduced `wlr_buffer_impl.get_data_ptr` for the purposes of the Pixman renderer. This works well enough but has some limitations:

1. The caller has no easy way to figure out the format in which pixels are laid out: https://github.com/swaywm/wlroots/pull/2661#discussion_r614366805. We've worked around this in the Pixman renderer by calling `get_dmabuf`/`get_shm` just to get the format.
2.  It's not possible to safely wrap client `wl_shm` buffers inside a `wlr_buffer`. The commit "buffer: introduce wlr_shm_client_buffer" in https://github.com/swaywm/wlroots/pull/2851 does it anyways, but doesn't call `wl_shm_buffer_begin_access` nor `wl_shm_buffer_end_access`. Calling these before and after any access to the buffer data is necessary to prevent clients from SIGBUS'ing the compositor (by truncating their mmap'ed shm file). Wrapping a `wl_shm` client buffer in a `wlr_buffer` is desirable for screencopy purposes and for improving our buffer handling which is currently GL-centric.

cc @bl4ckb0ne

emersion:
Some previous thoughts on the matter: https://github.com/swaywm/wlroots/issues/2399#issuecomment-779742867

emersion:
(1) has been addressed in https://github.com/swaywm/wlroots/pull/2894. (2) remains.

PR #2894, opened 2021-04-26
buffer: add format param to get_data_ptr

emersion:
Allow wlr_buffer_impl.get_data_ptr to return a format.

This allows the Pixman renderer to not care about get_dmabuf/get_shm,
and only care about get_data_ptr. This will also help with [1], because
client wl_shm buffers can't implement get_shm.

[1]: https://github.com/swaywm/wlroots/pull/2892

References: https://github.com/swaywm/wlroots/issues/2864

cc @bl4ckb0ne

### DIFF
--- a/include/types/wlr_buffer.h
+++ b/include/types/wlr_buffer.h
@@ -2,14 +2,15 @@
 #define TYPES_WLR_BUFFER
 
 #include <wlr/types/wlr_buffer.h>
+
 /**
  * Access a pointer to the allocated data from the underlying implementation,
- * and its stride.
+ * its format and its stride.
  *
  * The returned pointer should be pointing to a valid memory location for read
  * and write operations.
  */
 bool buffer_get_data_ptr(struct wlr_buffer *buffer, void **data,
-	size_t *stride);
+	uint32_t *format, size_t *stride);
 
 #endif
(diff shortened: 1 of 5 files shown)


bl4ckb0ne:
Should the format be a member of `wlr_buffer`? Otherwise LGTM

emersion:
Oh, that's a good point, I forgot we discussed about that.

The only reason why it maybe doesn't make sense to have it as a member of `wlr_buffer` is if the buffer is opaque or if it doesn't have a single fixed format. Some potential edge-cases:

- A solid color `wl_buffer` without any backing storage.
- A 3D view `wl_buffer`, as seen in [`zxr_composite_buffer_v1`](https://git.sr.ht/~sircmpwn/wxrc/tree/master/item/protocol/zxr-shell-unstable-v1.xml#L135).
- A virtual `wl_buffer` which mirrors the contents of an output or toplevel. Could be useful for e.g. phosh's UI.

It's not yet clear to me all buffer types will have a format. Maybe it would be better to wait a little before baking this assumption into `wlr_buffer`.

bl4ckb0ne:
I think the `wlr_buffer` api is in dire need of more tinkering, but I agree that we don't know all of the edge cases for now. 

I'd be ok to merge this now as it is, because it fixes more than one issue. We can always come back later and improve :)

emersion:
Right. It's not a big deal because this is only a breaking change for internal wlroots APIs and for third-party `wlr_buffer_impl.get_data_ptr` implementors, and I'm not aware of any yet.

PR #2895, opened 2021-04-26
render/pixman: check format support

emersion:
Pixman doesn't necessarily support all formats on all architectures.

cc @bl4ckb0ne


bl4ckb0ne:
I think we should be good, [all pixman_format_code_t values are supported][1]. Unless we have support for [YUV][2] formats?

[1]: https://cgit.freedesktop.org/pixman/tree/pixman/pixman.c#n1009
[2]: https://cgit.freedesktop.org/pixman/tree/pixman/pixman.c#n1088

emersion:
Eh, indeed. Not worth the trouble then!

Issue #2896, opened 2021-04-27
Regression for WLR_DRM_NO_MODIFIERS after #2885


kennylevinsen:
https://github.com/swaywm/wlroots/commit/f9f90b417366581b58d806956c9b2099417ed4b3 from #2885 has lead to a regression for me when using WLR_DRM_NO_MODIFIERS=1.

```
00:00:00.909 [INFO] [wlr] [backend/drm/drm.c:1431] connector eDP-1: Requesting modeset
00:00:00.909 [DEBUG] [sway/desktop/output.c:840] New output 0x558589826400: eDP-1
00:00:00.909 [DEBUG] [sway/config/output.c:351] Turning on output eDP-1
00:00:00.909 [DEBUG] [sway/config/output.c:329] Output DPI: 336.331034x322.729412
00:00:00.909 [DEBUG] [sway/config/output.c:382] Auto-detected output scale: 2.000000
00:00:00.909 [DEBUG] [sway/config/output.c:385] Set eDP-1 scale to 2.000000
00:00:00.910 [DEBUG] [sway/config/output.c:412] Committing output eDP-1
00:00:00.910 [INFO] [wlr] [backend/drm/drm.c:778] connector eDP-1: Modesetting with '3840x2160@59997mHz'
00:00:00.910 [DEBUG] [wlr] [backend/drm/drm.c:671] connector eDP-1: Initializing renderer
00:00:00.910 [DEBUG] [wlr] [render/swapchain.c:105] Allocating new swapchain buffer
00:00:00.916 [DEBUG] [wlr] [render/gbm_allocator.c:123] Allocated 3840x2160 GBM buffer (format 0x34325241, modifier 0xFFFFFFFFFFFFFF)
00:00:00.916 [DEBUG] [wlr] [render/gles2/renderer.c:132] Created GL FBO for buffer 3840x2160
00:00:00.917 [ERROR] [wlr] [backend/drm/util.c:204] Failed to add DRM frame-buffer: BO has modifier 0x100000000000001 but KMS doesn't support modifiers
00:00:00.917 [DEBUG] [wlr] [backend/drm/renderer.c:417] Failed to import GBM BO in KMS
00:00:00.917 [ERROR] [wlr] [backend/drm/renderer.c:323] Failed to import buffer
00:00:00.917 [ERROR] [wlr] [backend/drm/drm.c:693] connector eDP-1: Failed to initialize renderer:initial page-flip failed
[...]

PR #2897, opened 2021-04-27
Revert "backend/drm: fail instead of stripping a modifier"

emersion:
This reverts commit f9f90b417366581b58d806956c9b2099417ed4b3.

gbm_bo_get_modifier may return a modifier in these cases:

- The kernel doesn't support modifiers but Mesa does
- WLR_DRM_NO_MODIFIERS=1 is set

However, in both of these cases, the gbm_bo has been allocated
without modifiers.

There is already a check in drm_fb_create for modifiers:
wlr_drm_format_set_has will make sure buffers with an explicit
modifier will be rejected if the DRM backend doesn't support them.
So no need for an additional check in get_fb_for_bo.

Closes: https://github.com/swaywm/wlroots/issues/2896

cc @kennylevinsen

### DIFF
--- a/backend/drm/util.c
+++ b/backend/drm/util.c
@@ -199,14 +199,6 @@ uint32_t get_fb_for_bo(struct gbm_bo *bo, bool with_modifiers) {
 			wlr_log_errno(WLR_ERROR, "Unable to add DRM framebuffer");
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


kennylevinsen (approved):

kennylevinsen:
Works for me™

PR #2898, opened 2021-04-28
backend/drm: always perform an atomic test-only commit on wlr_output_test

emersion:
~~Depends on: https://github.com/swaywm/wlroots/pull/2829~~

This allows callers to use wlr_output_test to check whether a mode
can be enabled, for instance.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -473,16 +473,19 @@ static bool drm_connector_test(struct wlr_output *output) {
 		}
 	}
 
-	if ((output->pending.committed & WLR_OUTPUT_STATE_BUFFER) && !conn->backend->parent) {
+	if (conn->backend->parent) {
+		// If we're running as a secondary GPU, we can't perform an atomic
+		// commit without blitting a buffer.
+		return true;
+	}
+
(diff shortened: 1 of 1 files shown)


emersion:
This is now ready for review.

backend/drm/drm.c
```diff
@@ -473,16 +473,19 @@ static bool drm_connector_test(struct wlr_output *output) {
 		}
 	}
 
-	if ((output->pending.committed & WLR_OUTPUT_STATE_BUFFER) && !conn->backend->parent) {
+	if (conn->backend->parent) {
+		// If we're running as a secondary GPU, we can't perform an atomic
+		// commit without blitting a buffer.
```
kennylevinsen, line 478: Hmm, so test commits aren't possible in that configuration?

emersion: We need the buffer which will be used for scan-out to do a test-only commit. So yeah, it's tricky to get one without doing a buffer copy. I'd rather disable any fancy features on secondary GPUs for now (which we don't have any right now, but in the future will be output layers, scaling/cropping, etc).

kennylevinsen (approved):
LGTM, thanks!

PR #2899, opened 2021-04-29
backend/drm: fix allocator DRM FD on multi-GPU setups

emersion:
On multi-GPU setups, there is a primary DRM backend and secondary
DRM backends. wlr_backend_get_drm_fd will always return the parent
DRM FD even on secondary backends, so that users always use the
primary device for rendering.

However, for our internal rendering we want to use the secondary
device. Use allocator_autocreate_with_drm_fd to make sure the
allocator will create buffers on the secondary device.

We do something similar to ensure our internal rendering will
happen on the secondary device with renderer_autocreate_with_drm_fd.

Fixes: cc1b663 ("backend: use wlr_allocator_autocreate")

cc @bl4ckb0ne

### DIFF
--- a/backend/drm/renderer.c
+++ b/backend/drm/renderer.c
@@ -35,8 +35,8 @@ bool init_drm_renderer(struct wlr_drm_backend *drm,
 		goto error_gbm;
 	}
 
-	renderer->allocator = wlr_allocator_autocreate(&drm->backend,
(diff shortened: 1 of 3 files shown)


bl4ckb0ne:
LGTM!

emersion:
Thanks for the review!

PR #2901, opened 2021-04-29
backend: add common renderer and allocator creation

emersion:
This is part of #2505 split out in a separate PR, and rebased on top of #2884.

Instead of explicitly creating the renderer and allocator inside the backends, do it in common code. The motivation is:

- Nothing of this is backend-specific. With #2505 backends won't need to care about renderers and allocators anymore (except for the edge case of DRM multi-GPU).
- Allow #2507 and #2505 to get access to a renderer and an allocator outside of the backend for creating swapchains and rendering.

With this PR, backends now get access to the renderer and allocator via `wlr_backend_get_renderer` and `backend_get_allocator`. This makes it pretty obvious that a lot of the backend logic no longer needs to live in the backend, and can be moved in common code as done in #2507 and #2505.

The DRM backend is left alone for now, because multi-GPU support complicates things: we can't easily get rid of the internal renderer/allocator on secondary DRM backends.

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -19,6 +19,7 @@
 #include <wlr/util/log.h>
 #include "backend/backend.h"
 #include "backend/multi.h"
+#include "render/allocator.h"
 #include "util/signal.h"
 
 #if WLR_HAS_X11_BACKEND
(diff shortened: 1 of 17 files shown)


emersion:
I'm not very happy with this PR, but I don't see good alternatives that make it possible to still support `wlr_output_attach_render`.

I've thought about making `wlr_multi_backend` responsible for doing this, but `wlr_output_attach_render` needs access to the allocator/renderer, and `wlr_output.backend` doesn't point to the multi backend (but points to the "real" underlying backend).

bl4ckb0ne:
Do you think we need a rework of the multi backend before this can land?

emersion:
No, as explained above, using the multi backend for this won't work.

kennylevinsen (approved):

kennylevinsen:
Thanks!

PR #2902, opened 2021-04-29
backend/headless: dup DRM FD in wlr_headless_backend_create_with_renderer

emersion:
We were not dup'ing the DRM FD coming from wlr_renderer_get_drm_fd,
resulting in a double-close on backend destroy.


kennylevinsen (approved):

kennylevinsen:
Thanks!

PR #2903, opened 2021-04-29
backend/drm: move primary wlr_swapchain handling out of backend

emersion:
This is https://github.com/swaywm/wlroots/pull/2505 but with the DRM backend changes too. This essentially completes the plan laid out in https://github.com/swaywm/wlroots/issues/1352, while keeping the existing compositor-facing `wlr_output` API intact.

- [x] Implement `get_primary_formats`
- [x] Allocate an empty buffer on modeset, if we don't have one
- [x] Allow direct scan-out with legacy KMS API as long as the format/modifier doesn't change
- [x] Perform a test-only commit with the buffer when mode-setting an output
- [x] Fallback to modifier-less buffers

~~Depends on: https://github.com/swaywm/wlroots/pull/2505~~
~~Depends on: https://github.com/swaywm/wlroots/pull/2829~~
~~Depends on: https://github.com/swaywm/wlroots/pull/3035~~

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -401,14 +401,14 @@ static bool drm_connector_set_pending_fb(struct wlr_drm_connector *conn,
 	struct wlr_buffer *local_buf;
 	if (drm->parent) {
 		struct wlr_drm_format *format =
-			drm_plane_pick_render_format(plane, &drm->renderer);
+			drm_plane_pick_render_format(plane, &drm->mgpu_renderer);
 		if (format == NULL) {
 			wlr_log(WLR_ERROR, "Failed to pick primary plane format");
 			return false;
 		}
 
 		// TODO: fallback to modifier-less buffer allocation
-		bool ok = init_drm_surface(&plane->mgpu_surf, &drm->renderer,
+		bool ok = init_drm_surface(&plane->mgpu_surf, &drm->mgpu_renderer,
 			state->buffer->width, state->buffer->height, format);
 		free(format);
 		if (!ok) {
@@ -632,14 +632,14 @@ static bool drm_connector_init_renderer(struct wlr_drm_connector *conn,
 		int height = mode.vdisplay;
 
 		struct wlr_drm_format *format =
-			drm_plane_pick_render_format(plane, &drm->renderer);
+			drm_plane_pick_render_format(plane, &drm->mgpu_renderer);
 		if (format == NULL) {
 			wlr_log(WLR_ERROR, "Failed to pick primary plane format");
 			return false;
 		}
 
 		// TODO: fallback to modifier-less buffer allocation
-		bool ok = init_drm_surface(&plane->mgpu_surf, &drm->renderer,
+		bool ok = init_drm_surface(&plane->mgpu_surf, &drm->mgpu_renderer,
 			width, height, format);
 		free(format);
 		if (!ok) {
@@ -834,13 +834,13 @@ static bool drm_connector_set_cursor(struct wlr_output *output,
 		struct wlr_buffer *local_buf;
 		if (drm->parent) {
 			struct wlr_drm_format *format =
-				drm_plane_pick_render_format(plane, &drm->renderer);
+				drm_plane_pick_render_format(plane, &drm->mgpu_renderer);
 			if (format == NULL) {
 				wlr_log(WLR_ERROR, "Failed to pick cursor plane format");
 				return false;
 			}
 
-			bool ok = init_drm_surface(&plane->mgpu_surf, &drm->renderer,
+			bool ok = init_drm_surface(&plane->mgpu_surf, &drm->mgpu_renderer,
 				buffer->width, buffer->height, format);
 			free(format);
 			if (!ok) {
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -466,8 +466,7 @@ static bool drm_connector_test(struct wlr_output *output) {
 		}
 	}
 
-	if ((output->pending.committed & WLR_OUTPUT_STATE_BUFFER) &&
-			output->pending.buffer_type == WLR_OUTPUT_STATE_BUFFER_SCANOUT) {
+	if ((output->pending.committed & WLR_OUTPUT_STATE_BUFFER) && !conn->backend->parent) {
 		if (!drm_connector_set_pending_fb(conn, &output->pending)) {
 			return false;
 		}
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -329,21 +329,6 @@ static struct wlr_drm_connector *get_drm_connector_from_output(
 	return (struct wlr_drm_connector *)wlr_output;
 }
 
-static bool drm_connector_attach_render(struct wlr_output *output,
-		int *buffer_age) {
-	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
-	return drm_surface_make_current(&conn->crtc->primary->surf, buffer_age);
-}
-
-static void drm_plane_set_committed(struct wlr_drm_plane *plane) {
-	drm_fb_move(&plane->queued_fb, &plane->pending_fb);
-
-	if (plane->queued_fb && plane->surf.swapchain) {
-		wlr_swapchain_set_buffer_submitted(plane->surf.swapchain,
-			plane->queued_fb->wlr_buf);
-	}
-}
-
 static bool drm_crtc_commit(struct wlr_drm_connector *conn,
 		const struct wlr_output_state *state, uint32_t flags, bool test_only) {
 	// Disallow atomic-only flags
@@ -353,9 +338,9 @@ static bool drm_crtc_commit(struct wlr_drm_connector *conn,
 	struct wlr_drm_crtc *crtc = conn->crtc;
 	bool ok = drm->iface->crtc_commit(conn, state, flags, test_only);
 	if (ok && !test_only) {
-		drm_plane_set_committed(crtc->primary);
+		drm_fb_move(&crtc->primary->queued_fb, &crtc->primary->pending_fb);
 		if (crtc->cursor != NULL) {
-			drm_plane_set_committed(crtc->cursor);
+			drm_fb_move(&crtc->cursor->queued_fb, &crtc->cursor->pending_fb);
 		}
 	} else {
 		drm_fb_clear(&crtc->primary->pending_fb);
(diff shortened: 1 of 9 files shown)


emersion:
Seems like I'm hitting synchronization issues with this PR, on both AMD and Intel. The buffers seem to be scanned out by KMS before the draw is complete, or cleared before KMS is done with them.

emersion:
> Seems like I'm hitting synchronization issues with this PR

Fixed by https://github.com/swaywm/wlroots/pull/3027

emersion:
This is now ready to review.

bl4ckb0ne (approved):

bl4ckb0ne:
Overall LGTM, I don't think I understand fully everything but I got the gist of it. Do you want another reviewer on it?

emersion:
@kennylevinsen, if you had a bit of time to double-check this PR, that would be helpful! Let me know if you don't, no worries. :)

@bl4ckb0ne, let me know if you have any specific questions about this PR.

backend/drm/legacy.c
```diff
@@ -8,13 +8,49 @@
...
+		struct wlr_drm_fb *prev_fb = crtc->primary->queued_fb;
+		if (!prev_fb) {
+			prev_fb = crtc->primary->current_fb;
+		}
+
+		/* Legacy is only guaranteed to be able to display a FB if it's been
+		 * allocated the same way as the previous one. */
+		if (prev_fb != NULL && !legacy_fb_props_match(prev_fb, pending_fb)) {
+			wlr_drm_conn_log(conn, WLR_DEBUG,
+				"Cannot use direct scan-out with legacy KMS API");
```
kennylevinsen, line 51: This message is no longer correct, as we will do direct scan-out on legacy when the buffer matches the previous one.

types/wlr_output.c
```diff
@@ -642,6 +657,74 @@ static void output_pending_resolution(struct wlr_output *output, int *width,
...
+
+	if (output->swapchain->format->len == 0) {
+		return false;
+	}
+
+	// The test failed for a buffer which has modifiers, try disabling
+	// modifiers to see if that makes a difference.
+	wlr_log(WLR_DEBUG, "Output modeset test failed, retrying without modifiers");
+
+	if (!output_create_swapchain(output, false)) {
```
kennylevinsen, line 714: So we set up a modifier-less swapchain here, but the code below will in some cases try to allocate a modifier...full swapchain again: `output_attach_empty_buffer` below calls `wlr_output_attach_render`, which in case of lacking `attach_render` implementation will call `output_attach_back_buffer` and ultimately attempt to allocate a swapchain with modifiers.

It would be better if we could avoid the split responsibility for swapchain setup.

types/wlr_output.c
```diff
@@ -706,6 +789,9 @@ static bool output_basic_test(struct wlr_output *output) {
 }
 
 bool wlr_output_test(struct wlr_output *output) {
+	if (!output_ensure_buffer(output)) {
```
kennylevinsen, line 792: Do we want to attach buffers to the output before doing the basic output test? And do we even want wlr_output_test to attach buffers? Seems a bit intrusive.

kennylevinsen (commented):
Commit message nit: "we don't need a renderer not an allocator." - I suspect you meant "nor".

Have this been run on multi-GPU for testing? I only have boring single-GPU machines.

kennylevinsen:
Fails in test:

```
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1096] Reallocating CRTCs
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1107] State before reallocation:
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1113]   'DP-1' crtc=0 state=1 desired_enabled=1
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1113]   'HDMI-A-1' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1113]   'DP-2' crtc=1 state=1 desired_enabled=1
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1113]   'HDMI-A-2' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1113]   'DP-3' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1113]   'HDMI-A-3' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1164] State after reallocation:
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1171]   'DP-1' crtc=0 state=1 desired_enabled=1
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1171]   'HDMI-A-1' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1171]   'DP-2' crtc=1 state=1 desired_enabled=1
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1171]   'HDMI-A-2' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1171]   'DP-3' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [DEBUG] [wlr] [backend/drm/drm.c:1171]   'HDMI-A-3' crtc=-1 state=0 desired_enabled=0
00:00:00.450 [INFO] [wlr] [backend/drm/drm.c:1453] connector DP-1: Requesting modeset
00:00:00.450 [DEBUG] [sway/desktop/output.c:839] New output 0x1ec6230: DP-1
00:00:00.450 [DEBUG] [sway/config/output.c:351] Turning on output DP-1
00:00:00.450 [DEBUG] [sway/config/output.c:329] Output DPI: 162.560000x161.364706
00:00:00.450 [DEBUG] [sway/config/output.c:382] Auto-detected output scale: 1.000000
00:00:00.450 [DEBUG] [sway/config/output.c:412] Committing output DP-1
00:00:00.450 [DEBUG] [wlr] [types/wlr_output.c:694] Attaching empty buffer to output for modeset
00:00:00.450 [DEBUG] [wlr] [types/wlr_output.c:513] Choosing primary buffer format 0x34325241 for output 'DP-1'
00:00:00.450 [DEBUG] [wlr] [render/swapchain.c:105] Allocating new swapchain buffer
00:00:00.454 [DEBUG] [wlr] [render/gbm_allocator.c:127] Allocated 3840x2160 GBM buffer (format 0x34325241, modifier 0x100000000000004)
00:00:00.454 [DEBUG] [wlr] [render/gles2/renderer.c:139] Created GL FBO for buffer 3840x2160
00:00:00.455 [DEBUG] [wlr] [backend/drm/atomic.c:35] connector DP-1: Atomic test failed (modeset): Invalid argument
00:00:00.455 [DEBUG] [wlr] [types/wlr_output.c:712] Output modeset test failed, retrying without modifiers
00:00:00.455 [DEBUG] [wlr] [types/wlr_output.c:513] Choosing primary buffer format 0x34325241 for output 'DP-1'
sway: types/wlr_output.c:661: output_attach_empty_buffer: Assertion `!(output->pending.committed & WLR_OUTPUT_STATE_BUFFER)' failed.
```

types/wlr_output.c
emersion: That's a good question. The reason a buffer is attached is that some backends might only be able to test a modeset if they have a buffer. This is the case of the DRM backend for instance, previously an empty black buffer was attached on modeset if none was provided by the compositor.

For the X11 and Wayland backends performing a modeset without a buffer doesn't really make sense either.

types/wlr_output.c
emersion: This shouldn't happen, since the `output_create_swapchain` call before ensures the swapchain is already allocated, so `output_attach_empty_buffer` won't do it again. I agree it's not ideal, but I don't have great ideas to simplify it.

kennylevinsen: Ah, you're right, it is correct although a bit subtle. Note to self: If you `output_create_swapchain` with `allow_modifiers = false`, then only modifier-less swapchains will be reused, while if called with `allow_modifiers = true`, then any swapchain will be reused rather than giving you a modifier-full one.

It wouldn't hurt with a comment at output_create_swapchain to describe this behavior, as it's definitely more subtle than the previous swapchain reuse logic.

types/wlr_output.c
kennylevinsen: Oh yeah, it's also needed to test the modeset.

However, does it make sense to perform `output_basic_test` *before* ensuring a buffer? This would let us bail out early from an obviously bad configuration without wasting time allocating and clearing buffers (possibly twice if we fall back to modifierless). I don't think `output_basic_test` relies on the output from `output_ensure_buffer`—the buffer tests have matching conditions in `output_ensure_buffer`, and the direct scan-out resolution test shouldn't be relevant if we had to allocate the buffer.

It should also be noted that we now issue up to 3 tests towards the output impl from `wlr_output_test`, as `output_ensure_buffer` performs 1 or 2 tests of its own. This shouldn't be a problem if the test is sufficiently fast.

types/wlr_output.c
emersion: Yes, that's a good point. Added a comment.

types/wlr_output.c
emersion: > However, does it make sense to perform output_basic_test before ensuring a buffer?

I've been wondering about this as well. Updated to allocate the buffer only if the basic test succeeds, we can always change this in the future if necessary.

Note, once the first test has allocated a swapchain+buffer, these are re-used in subsequent tests and commits, so the overhead shouldn't be too bad.

emersion:
Fixed up multi-GPU issues: with `drm_surface_lock` gone, the multi-GPU copy was gone too, and we were trying to use direct scanout without copying. Also we shouldn't strip LINEAR from the format if `allow_modifiers = false`.

kennylevinsen (approved):
LGTM. I'll take your word for multi-GPU being fixed. :)

backend/drm/drm.c
```diff
@@ -1004,6 +1004,21 @@ static void drm_connector_get_cursor_size(struct wlr_output *output,
...
 }
 
+static const struct wlr_drm_format_set *drm_connector_get_primary_formats(
+		struct wlr_output *output, uint32_t buffer_caps) {
+	if (!(buffer_caps & WLR_BUFFER_CAP_DMABUF)) {
+		return NULL;
+	}
+	struct wlr_drm_connector *conn = get_drm_connector_from_output(output);
+	if (!conn->crtc) {
+		return false;
```
bl4ckb0ne, line 973: Why not return `NULL`?

emersion: Good catch, seems like it's copypasta coming from `get_cursor_formats`. Sending a separate PR to fix these.

PR #2905, opened 2021-04-29
backend/drm: Do not require mode conmit on enable

kennylevinsen:
This was not required before, and the assert is tripped by simple
display disable/enable cycles.

Closes: https://github.com/swaywm/wlroots/issues/2904

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -729,9 +729,12 @@ static bool drm_connector_set_mode(struct wlr_drm_connector *conn,
 
 	struct wlr_output_mode *wlr_mode = NULL;
 	if (drm_connector_state_active(conn, state)) {
-		assert(state->committed & WLR_OUTPUT_STATE_MODE);
-		assert(state->mode_type == WLR_OUTPUT_STATE_MODE_FIXED);
-		wlr_mode = state->mode;
+		if (state->committed & WLR_OUTPUT_STATE_MODE) {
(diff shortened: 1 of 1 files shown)


emersion:
Hm, I think this will still be broken if the user sets a custom mode, then disables and enables the output.

If the state doesn't contain a mode, we should re-use the current mode. So maybe something like this will help?

```c
if (!(state->committed & WLR_OUTPUT_STATE_MODE)) {
    wlr_mode = conn->output.mode;
}
```

Note, I'd like to completely get rid of `drm_connector_set_mode` at some point.

kennylevinsen:
> Hm, I think this will still be broken if the user sets a custom mode, then disables and enables the output.

Don't we always convert custom mode to a fixed mode before calling set_mode?

https://github.com/swaywm/wlroots/blob/c85789a3a9f6f851e6fbc900495057ba91b3e255/backend/drm/drm.c#L503-L517

> So maybe something like this will help?

Yeah, it's a bit nicer to explicitly use the *current* mode rather than read from pending...

emersion:
> Don't we always convert custom mode to a fixed mode before calling set_mode?

We do it, but we don't mutate `output->pending.mode` (that would be surprising for users). Instead we copy over the pending state and mutate our copy, then pass that copy down.

kennylevinsen:
Ah, makes sense. The patch now uses the current mode anyway, as it's much more correct than reading from pending.

emersion (approved):
LGTM, thanks!

PR #2909, opened 2021-05-02
xwm: implement _NET_CLIENT_LIST_STACKING

### LINKED ISSUE #2889: X11 drag and drop sometimes drops "through" overlapping windows
Given something like `T[slack* discord]`, and initiating a drag in Slack:

* the drag image does not respond to movement, and is frozen a couple of pixels from where it started
* releasing the drag causes Discord to receive it, making it request urgent status

Video example:

https://user-images.githubusercontent.com/1403503/112929915-6027d380-90e7-11eb-8c68-3b66dc76ea31.mp4

The logs around this time look like:

```
14:50:42.736 [DEBUG] [wlr] [types/seat/wlr_seat_pointer.c:384] button_count=1 grab_serial=66649 serial=66651
14:50:42.737 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 401 (_NET_WM_USER_TIME) for window 20971527
14:50:42.968 [DEBUG] [sway/desktop/xwayland.c:761] New xwayland unmanaged surface
14:50:42.968 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 360 (CHROMIUM_TIMESTAMP) for window 20971528
14:50:42.968 [DEBUG] [wlr] [xwayland/selection/incoming.c:487] XCB_XFIXES_SELECTION_NOTIFY (selection=299, owner=20971621)
14:50:42.968 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 380 (XdndActionList) for window 20971527
14:50:42.968 [DEBUG] [sway/desktop/xwayland.c:761] New xwayland unmanaged surface
14:50:42.968 [DEBUG] [wlr] [xwayland/selection/incoming.c:453] XCB_SELECTION_NOTIFY (selection=299, property=282, target=283)
14:50:42.970 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 6291468
14:50:42.970 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 6291468
14:50:42.970 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 6291468
14:50:42.970 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 6291468
14:50:47.954 [DEBUG] [wlr] [types/seat/wlr_seat_pointer.c:384] button_count=0 grab_serial=66651 serial=66652
14:50:47.955 [DEBUG] [wlr] [xwayland/selection/incoming.c:487] XCB_XFIXES_SELECTION_NOTIFY (selection=299, owner=0)
14:50:47.955 [ERROR] [wlr] [xwayland/xwm.c:1386] xcb error: op 18:0, code 3, sequence 18214, value 20971622
14:50:48.804 [DEBUG] [sway/input/cursor.c:893] denying request to set cursor from unfocused client
14:50:49.025 [DEBUG] [sway/input/cursor.c:893] denying request to set cursor from unfocused client
14:50:49.026 [DEBUG] [sway/input/cursor.c:893] denying request to set cursor from unfocused client

--- releasing the drag here ---

14:50:51.119 [DEBUG] [sway/tree/arrange.c:264] Usable area for ws: 1440x2530@0,30
14:50:51.119 [DEBUG] [sway/tree/arrange.c:294] Arranging workspace '2' at 0.000000, 30.000000
14:50:51.119 [DEBUG] [sway/tree/arrange.c:77] Arranging 0x7fffb6e5b2b0 horizontally
14:50:51.119 [DEBUG] [sway/tree/arrange.c:156] Arranging 0x7fffb6e5b240 vertically
14:50:51.119 [DEBUG] [sway/tree/arrange.c:156] Arranging 0x7fffb6e5b160 vertically
14:50:51.119 [DEBUG] [sway/desktop/transaction.c:399] Transaction 0x559e6c8fb040 committing with 10 instructions
14:50:51.119 [DEBUG] [sway/desktop/transaction.c:294] Applying transaction 0x559e6c8fb040
14:50:51.119 [DEBUG] [wlr] [types/seat/wlr_seat_pointer.c:384] button_count=1 grab_serial=66651 serial=66662
```

In a floating layout like the below, with Chromium partially obscured by Slack and dragging from Slack:

image

...the drag image position updates with mouse movement until it overlaps with the Chromium window underneath, at which point it freezes.

At the time of intersection, the

```
14:55:23.241 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 35651585
14:55:23.242 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 35651585
14:55:23.242 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 35651585
14:55:23.242 [DEBUG] [wlr] [xwayland/xwm.c:759] unhandled X11 property 385 (_CHROMIUM_DRAG_RECEIVER) for window 35651585
```

lines get printed, but nothing else appears in the log.

Xyene:
This property is present on all modern X11 instances. The nonpresence of
it requires applications to fall back to XQueryTree-based logic to
determine stacking logic (e.g., to determine what surface should get
Xdnd events).

These code paths are effectively untested nowadays, so this makes it
more likely for wlroots to "break" applications. For instance, the
XQueryTree fallback path has been broken in Chromium for the last 10
years.

It's easy enough to maintain this property, so let's just do it.

Fixes #2889.


xwayland/xwm.c
```diff
@@ -1604,10 +1639,10 @@ void xwm_destroy(struct wlr_xwm *xwm) {
 	}
 #endif
 	struct wlr_xwayland_surface *xsurface, *tmp;
-	wl_list_for_each_safe(xsurface, tmp, &xwm->surfaces, link) {
+	wl_list_for_each_safe(xsurface, tmp, &xwm->surfaces_in_map_order, map_link) {
 		xwayland_surface_destroy(xsurface);
 	}
-	wl_list_for_each_safe(xsurface, tmp, &xwm->unpaired_surfaces, link) {
+	wl_list_for_each_safe(xsurface, tmp, &xwm->unpaired_surfaces, unpaired_link) {
```
Xyene, line 1645: Note for reviewer: this change isn't related to this patch, but it seemed broken so I changed it anyway.

xwayland/xwm.c
```diff
@@ -1518,7 +1543,17 @@ void wlr_xwayland_surface_restack(struct wlr_xwayland_surface *surface,
...
 	values[idx++] = mode;
 
-	xcb_configure_window(xwm->xcb_conn, surface->window_id, flags, values);
+	xcb_configure_window(xwm->xcb_conn, xsurface->window_id, flags, values);
+
+	if (mode == XCB_STACK_MODE_ABOVE) {
+		wl_list_remove(&xsurface->focus_link);
+		wl_list_insert(&xwm->surfaces_in_focus_order, &xsurface->focus_link);
+	} else {
+		// TODO
```
Xyene, line 1552: /cc @ammen99: Sway doesn't use this function so I'm not sure what the right behavior here is (or how I'd exercise this code path to test). Thoughts?

xwayland/xwm.c
```diff
@@ -186,7 +188,7 @@ static struct wlr_xwayland_surface *xwayland_surface_create(
 		return NULL;
 	}
 
-	wl_list_insert(&xwm->surfaces, &surface->link);
+	wl_list_insert(&xwm->surfaces_in_map_order, &surface->map_link);
```
Xyene, line 191: Note for reviewer: I'm pretty sure this logic is wrong. `surfaces_in_map_order` is used to maintain `_NET_CLIENT_LIST`.

EMWH describes `_NET_CLIENT_LIST` as:

> These arrays contain all X Windows managed by the Window Manager. _NET_CLIENT_LIST has initial mapping order, starting with the oldest window.

`xwm_set_net_client_list` filters this list by `surface->mapped`, but there's no hard guarantee that surfaces will be mapped in the order they're created, right?

Xyene:
It's worth noting that this needs a follow-up patch to Sway (and other compositors) to work correctly, maybe to wlroots too. In particular I don't think jumping to the top of the stack on focus is correct when taking into account mixed tiling/floating windows:

image

Here Chromium is focused, so the stacking order as determined by wlroots is Chromium > gedit, even though this isn't the case. So, drag & drop from Chromium to gedit fails over the region they overlap.

I think this PR still improves things in Sway's case (where things are mostly tiled, which this PR fixes), and Wayfire (where focus implies top-of-stack?)

Xyene:
I think this should be good for review now; I plan on addressing the remaining issues in a separate PR (since they're not introduced by this one).

emersion:
This sounds like a lot of trouble on our side to accommodate for a mis-behaving client…

Xyene:
It's a bit annoying, but I don't think maintaining a new list is too bad. i3 does the same thing; I suspect mutter does too. 

On principle, I dislike the idea of wlroots-based compositors being the "odd ones out" when it comes to supporting common EMWH properties, because we hit all the bitrotten code paths in clients (if they even exist). Fragmentation isn't good; there are better fights to fight. The bug in Chromium was there for 10 years and went unnoticed because everyone else already specifies `_NET_CLIENT_LIST_STACKING`.

User benefit aside, I'd argue the lifetime maintenance overhead of this patch is going to be strictly less than the contributor time spent fielding and looking into future "weird" Xwayland issues that aren't obviously caused by stacking issues (e.g., I only accidentally realized https://github.com/swaywm/sway/issues/5927 was caused by this).

The other points I mentioned to be fixed in another PR are already broken even for correct `XQueryTree`-based lookup, because we get the stacking order wrong.

Xyene:
Gentle ping :)

Another upside of supporting `_NET_CLIENT_LIST_STACKING` is performance; without it, an application sourcing a drag-and-drop operation needs, on every cursor frame, a round-trip through `XQueryTree` followed by a walk of the returned tree.

xwayland/xwm.c
```diff
@@ -243,6 +245,24 @@ static void xwm_set_net_client_list(struct wlr_xwm *xwm) {
 			XCB_ATOM_WINDOW, 32, mapped_surfaces, windows);
 }
 
+static void xwm_set_net_client_list_stacking(struct wlr_xwm *xwm) {
+	size_t num_surfaces = wl_list_length(&xwm->surfaces_in_stack_order);
+	xcb_window_t windows[num_surfaces + 1];
```
emersion, line 250: A client can blow up our stack by opening many windows.

xwayland/xwm.c
```diff
@@ -948,10 +970,11 @@ static void xwm_handle_map_request(struct wlr_xwm *xwm,
 	xsurface_set_wm_state(xsurface, XCB_ICCCM_WM_STATE_NORMAL);
 	xsurface_set_net_wm_state(xsurface);
 
-	uint32_t values[1];
-	values[0] = XCB_STACK_MODE_BELOW;
-	xcb_configure_window(xwm->xcb_conn, ev->window,
-			XCB_CONFIG_WINDOW_STACK_MODE, values);
+	xcb_configure_window(xwm->xcb_conn, ev->window, XCB_CONFIG_WINDOW_STACK_MODE,
```
emersion, line 973: Maybe we can use `wlr_xwayland_surface_restack` instead of doing this?

xwayland/xwm.c
```diff
@@ -1518,8 +1541,30 @@ void wlr_xwayland_surface_restack(struct wlr_xwayland_surface *surface,
 	}
 	values[idx++] = mode;
 
-	xcb_configure_window(xwm->xcb_conn, surface->window_id, flags, values);
-	xcb_flush(xwm->xcb_conn);
```
emersion, line 1522: This has disappeared.

xwayland/xwm.c
emersion: Yes, this sounds correct. So `surfaces_in_map_order` is actually not in map order.

Although it doesn't seem like i3 orders clients by map time in `_NET_CLIENT_LIST` either. So maybe it would be better to keep the current list with its current name...

xwayland/xwm.c
emersion: Can we isolate this change into a separate commit? That way we can safely revert other commits if necessary.

Xyene: Done.

xwayland/xwm.c
Xyene: Sounds reasonable, done.

xwayland/xwm.c
Xyene: Added a commit fixing this and the other instance for `_NET_CLIENT_LIST`.

xwayland/xwm.c
Xyene: i3 does order by map time, and has a test enforcing this:

https://github.com/i3/i3/blob/9db03797da3cea5dc6898adc79a68ba4523a409c/testcases/t/223-net-client-list.t

I don't feel strongly about this part of the change. I've backed it out and just left a FIXME as a big flag in case someone in the future stumbles upon this code while debugging some issue.

xwayland/xwm.c
Xyene: Oops, nice catch -- an earlier version of this flushed within `xwm_set_net_client_list_stacking`, but no longer does.

xwayland/xwm.c
emersion: Ah, I missed that, thanks

xwayland/xwm.c
```diff
@@ -244,6 +250,24 @@ static void xwm_set_net_client_list(struct wlr_xwm *xwm) {
 			XCB_ATOM_WINDOW, 32, mapped_surfaces, windows);
 }
 
+static void xwm_set_net_client_list_stacking(struct wlr_xwm *xwm) {
+	size_t num_surfaces = wl_list_length(&xwm->surfaces_in_stack_order);
+	xcb_window_t windows[num_surfaces + 1];
+
+	// We store surfaces in top-to-bottom order because this is easier to reason
```
emersion, line 265: Ah. This is inconsistent with the rest of wlroots: subsurfaces and popups are ordered in the other order, ie. the first item is underneath other items.

Maybe we should consider doing that too for Xwayland for consistency?

emersion (commented):
Apart from this one comment, looks good!

Xyene:
> Maybe we should consider doing that too for Xwayland for consistency?

Sure, done. This exposed a bug (not removing the surface from the list before determining where to re-add it, which would corrupt the list in some cases) -- this was also fixed.

xwayland/xwm.c
emersion: I removed the comment, because it's no longer relevant.

emersion (approved):
LGTM, thanks!

PR #2915, opened 2021-05-05
xwm: do not restack surfaces on activation 

Xyene:
Currently, upon activating a surface, wlroots restacks it on top of all
others.

This may not necessarily be correct from the calling compositor's point
of view, where having focus may not imply being top-of-stack (e.g.,
focusing a window under an always-on-top window).

In Sway's case, this means that focused tiling windows will always be on
top of floating windows, at least in the order communicated to X11 apps.
This breaks drag-and-drop from a focused tiling X11 window to a floating
X11 window which partially obscures the former.

This is a breaking change; to retain the previous behavior, users that
were calling

    wlr_xwayland_surface_activate(xsurface, true);

should now be calling

    wlr_xwayland_surface_activate(xsurface, true);
    wlr_xwayland_surface_restack(xsurface, NULL, XCB_STACK_MODE_ABOVE);

---

This is currently based off of #2909. I will send in a patch making Sway do
the "right thing" (i.e. not unconditionally restack tiling windows on top of
floating ones) soon.

I've tested this change with this patch to Sway:

```diff
diff --git a/sway/desktop/xwayland.c b/sway/desktop/xwayland.c
index 66cb3b02..1af8d248 100644
--- a/sway/desktop/xwayland.c
+++ b/sway/desktop/xwayland.c
@@ -254,6 +254,7 @@ static void set_activated(struct sway_view *view, bool activated) {
 	}
 
 	wlr_xwayland_surface_activate(surface, activated);
+	wlr_xwayland_surface_restack(surface, NULL, XCB_STACK_MODE_ABOVE);
 }
 
 static void set_tiled(struct sway_view *view, bool tiled) {
```

and running `watch -n.1 -d xwininfo -root -tree` while doing focusing
operations with JetBrains, Slack, and Chromium, with the results
appearing correct.

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -320,7 +320,6 @@ static void xwm_set_focus_window(struct wlr_xwm *xwm,
 		xwm->last_focus_seq = cookie.sequence;
 	}
 
-	wlr_xwayland_surface_restack(xsurface, NULL, XCB_STACK_MODE_ABOVE);
 	xsurface_set_net_wm_state(xsurface);
 }
 


emersion:
Should we also remove the restacking when a window is mapped, and let the compositor do it?

emersion:
Also cc @ammen99

Xyene:
> Should we also remove the restacking when a window is mapped, and let the compositor do it?

I don't have a strong opinion on this. I do think every mapped window should appear in `_NET_CLIENT_LIST_STACKING`, otherwise we _really_ break applications doing drag-and-drop as they'll take it as the source of truth over `XQueryTree` output.

I weakly think that requiring a call to `wlr_xwayland_surface_restack` on downstream `handle_map`, lest drag-and-drop break, would be a bit unfortunate because of how easy it'd be to mess up and not notice. Perhaps it's best not to provide this footgun.

The docs for [`XMapRaised`](https://tronche.com/gui/x/xlib/window/XMapRaised.html) seem to imply that there does exist a default stack position for [`XMapWindow`](https://tronche.com/gui/x/xlib/window/XMapWindow.html), but what this position is seems unspecified (at the very least, it's not top-of-stack), so it's not very instructive here.

ammen99:
> Also cc @ammen99

The proposed change makes sense to me, I too have wondered why input focus has anything to do with stack order.

emersion:
> The docs for XMapRaised seem to imply that there does exist a default stack position for XMapWindow, but what this position is seems unspecified (at the very least, it's not top-of-stack), so it's not very instructive here.

OK. In this case, maybe we can just add the window to the bottom (or top? or something else? see below) of the list when it's mapped, and update the position when it's re-stacked. That way, things won't be horribly broken if the compositor doesn't re-stack.

> The docs for XMapRaised seem to imply that there does exist a default stack position for XMapWindow, but what this position is seems unspecified (at the very least, it's not top-of-stack), so it's not very instructive here.

The `XCreateWindow` docs say: "The created window is placed on top in the stacking order with respect to siblings."

Looking at the Xorg impl, it doesn't seem like `XMapWindow` changes the stack position. However a client could probably restack the window before mapping it?

I'm fine with inserting the window at an arbitrary place in the list.

Thinking about it again, should we really put absolutely all windows in the `_NET_CLIENT_LIST*` lists? Or just the one managed by the WM? (some winows aren't, e.g. override-redirect windows)

> I lied, this won't be coming from me soon. disappointed I can send a patch adding the wlr_xwayland_surface_restack to maintain current broken behavior if this PR is merged, but updating the stack incrementally proved very painful and error-prone, and I think I'll end up rebuilding the ful stack on every focus/map (like i3) to avoid that. At any rate, I don't have enough bandwidth to look into this right now.

No problem. I also think re-sending the whole stack order is the sensible thing to do here.

emersion:
This needs a rebase.

Xyene:
Sway PR retaining the current behavior is up now.

When Sway is updated to use the scene graph API, I think we'll get proper maintenance of `_NET_CLIENT_LIST_STACKING` for free, and I can revisit things at that time.

emersion (approved):
LGTM, thanks!

PR #2918, opened 2021-05-07
virtual_keyboard: Emulate key release events on destroy

### LINKED ISSUE #2034: virtual-keyboard: release unreleased keys

yuja:
According to libinput, release events are generated when device is unplugged,
and libinput copies this behavior for device removal. Let's do the same for
our virtual keyboard.

https://github.com/wayland-project/libinput/commit/8f846a41fa0566fbd72ece676656e20e56ce43e6

This is another attempt to fix #2034 and the following sway issue:

https://github.com/swaywm/sway/issues/6254

Note that we have other key repeating issues in sway, which aren't addressed
by this patch. Since the virtual keyboard itself isn't destroyed when the
keyboard grab is destroyed, we'll probably need some trick to reset the state
of the corresponding virtual keyboard when the grab is released.

https://github.com/swaywm/sway/issues/6095
https://github.com/swaywm/sway/issues/6193

### DIFF
--- a/types/wlr_virtual_keyboard_v1.c
+++ b/types/wlr_virtual_keyboard_v1.c
@@ -8,6 +8,7 @@
 #include <wlr/util/log.h>
 #include <xkbcommon/xkbcommon.h>
 #include "util/signal.h"
+#include "util/time.h"
 #include "virtual-keyboard-unstable-v1-protocol.h"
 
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #2921, opened 2021-05-12
xwayland: actually use Xwayland from pkg-config

t-8ch:
eec2e1d3b16ba40c85483badcc7809277274ce5d introduced logic to use the Xwayland
binary discovered via pkg-config.
While the newly introduced checks correctly used the binary from pkg-config,
the actual execution still used the previous PATH-search logic.


emersion (commented):
Right. Since we have `WLR_XWAYLAND` now, I think it's a reasonable change.

Can we change the "failed to exec Xwayland" message below to "failed to exec %s" with the pkg-config path?

t-8ch:
> Can we change the "failed to exec Xwayland" message below to "failed to exec %s" with the pkg-config path?

I used the actual used path instead of the pkg-config one.

emersion (approved):
Thanks!

PR #2931, opened 2021-05-19
render/pixman: avoid sqrt() in render_quad without rotation

emersion:
When the matrix doesn't have a rotation, we can avoid a sqrt() call.

Tested with Sway's tabbed containers.

cc @bl4ckb0ne

### DIFF
--- a/render/pixman/renderer.c
+++ b/render/pixman/renderer.c
@@ -230,10 +230,14 @@ static void pixman_render_quad_with_matrix(struct wlr_renderer *wlr_renderer,
 	memcpy(m, matrix, sizeof(m));
 
 	// TODO get the width/height from the caller instead of extracting them
-	// TODO detect non rotation with matrix[1] == 0 and matrix[4] == 0 to avoid
-	// doing the calculation
-	float width = sqrt(matrix[0] * matrix[0] + matrix[1] * matrix[1]);
-	float height = sqrt(matrix[3] * matrix[3] + matrix[4] * matrix[4]);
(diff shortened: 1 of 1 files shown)


render/pixman/renderer.c
```diff
@@ -230,10 +230,14 @@ static void pixman_render_quad_with_matrix(struct wlr_renderer *wlr_renderer,
 	memcpy(m, matrix, sizeof(m));
 
 	// TODO get the width/height from the caller instead of extracting them
-	// TODO detect non rotation with matrix[1] == 0 and matrix[4] == 0 to avoid
-	// doing the calculation
-	float width = sqrt(matrix[0] * matrix[0] + matrix[1] * matrix[1]);
-	float height = sqrt(matrix[3] * matrix[3] + matrix[4] * matrix[4]);
+	float width, height;
+	if (matrix[1] == 0.0 && matrix[3] == 0.0) {
```
bl4ckb0ne, line 234: should be `matrix[4]` instead of `matrix[3]`

emersion: Nope: the matrix looks like this for width/height scaling:

````
[ w 0 0
  0 h 0
  0 0 0 ]
```

So:

```
m[0] = w
m[1] = 0
m[3] = 0
m[4] = h
```

bl4ckb0ne: I see, my comment was wrong.

bl4ckb0ne (approved):

PR #2933, opened 2021-05-20
meson: fix warning about build_root

bl4ckb0ne:
Fixes the following warning

```
WARNING: Project targeting '>=0.56.0' but tried to use feature deprecated since '0.56.0': meson.build_root. use meson.project_build_root() or meson.global_build_root() instead.
```

/cc @emersion


emersion:
This doesn't do the right thing. See https://github.com/mesonbuild/meson/pull/8629 for details.

PR #2934, opened 2021-05-21
Custom wlr_client_buffer from wl_resource

bl4ckb0ne:
This patch is a rewrite of the work done by ddevault in https://github.com/swaywm/wlroots/commit/c88abfbc23da485507add71534785b4d8141efdc

The patchset is currently used [here](https://git.sr.ht/~bl4ckb0ne/wxrc/commit/6e66418103cb2b866818adbdd99186408599fb48#src/xr-shell-protocol.c) in my own fork of wxrc.

I chose to not make the existing client buffer impl (shm, dma-buf and wl_drm) yet because of the pending status of [wl_drm in wlroots](https://github.com/swaywm/wlroots/pull/2708).

### DIFF
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -100,11 +100,17 @@ bool wlr_buffer_get_dmabuf(struct wlr_buffer *buffer,
 bool wlr_buffer_get_shm(struct wlr_buffer *buffer,
 	struct wlr_shm_attributes *attribs);
 
+struct wlr_renderer;
+struct wlr_client_buffer;
+struct wlr_client_buffer_impl;
+
 /**
  * A client buffer.
  */
 struct wlr_client_buffer {
 	struct wlr_buffer base;
+
+	const struct wlr_client_buffer_impl *impl;
 
 	/**
 	 * The buffer resource, if any. Will be NULL if the client destroys it.
@@ -124,8 +130,20 @@ struct wlr_client_buffer {
 	struct wl_listener release;
 };
 
-struct wlr_renderer;
+struct wlr_client_buffer_impl {
+	bool (*is_instance)(struct wl_resource *resource);
+	bool (*get_buffer_size)(struct wl_resource *resource, int *width,
+			int *height);
+	struct wlr_client_buffer *(*create)(struct wl_resource *resource,
+			struct wlr_renderer *renderer);
+	void (*destroy)(struct wlr_client_buffer *buffer);
+};
 
+
+/**
+ * Allows a client to registered a custom implementation of a client buffer
+ */
+void wlr_client_buffer_register_impl(const struct wlr_client_buffer_impl *impl);
 /**
  * Get a client buffer from a generic buffer. If the buffer isn't a client
  * buffer, returns NULL.
--- a/render/wlr_renderer.c
+++ b/render/wlr_renderer.c
@@ -4,6 +4,7 @@
 #include <wlr/render/interface.h>
 #include <wlr/render/pixman.h>
 #include <wlr/render/wlr_renderer.h>
+#include <wlr/types/wlr_buffer.h>
 #include <wlr/types/wlr_matrix.h>
 #include <wlr/util/log.h>
 
@@ -248,7 +249,12 @@ bool wlr_renderer_init_wl_display(struct wlr_renderer *r,
 	return true;
 }
 
+extern const struct wlr_client_buffer_impl shm_client_buffer_impl;
+
 struct wlr_renderer *renderer_autocreate_with_drm_fd(int drm_fd) {
+
+	wlr_client_buffer_register_impl(&shm_client_buffer_impl);
+
 	const char *name = getenv("WLR_RENDERER");
 	if (name) {
 		wlr_log(WLR_INFO, "Loading user-specified renderer due to WLR_RENDERER: %s",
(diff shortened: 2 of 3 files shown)


include/wlr/types/wlr_client_buffer.h
```diff
@@ -13,9 +13,25 @@
...
+struct wlr_client_buffer;
+
+struct wlr_client_buffer_impl {
+	bool (*is_instance)(struct wl_resource *resource);
+	bool (*get_buffer_size)(struct wl_resource *resource,
+			struct wlr_renderer *renderer, int *width, int *height);
+	bool (*initialize)(struct wlr_client_buffer *,
+			struct wlr_renderer *renderer);
+	bool (*apply_damage)(struct wlr_client_buffer *buffer,
+			struct wl_resource *resource, pixman_region32_t *damage);
```
emersion, line 26: Not yet sure it makes sense for this to be part of the interface. Ideally we should be able to rework `wlr_texture_write_pixels` and make it not depend on the client buffer implementation.

emersion:
I'd prefer the existing client buffer types (`wl_shm`, linux-dmabuf, `wl_drm`) to also use the `wlr_client_buffer_impl` infrastructure, instead of being special-cased. That would make it easier for compositors to override wlroots' implementations if need be.

include/wlr/types/wlr_client_buffer.h
```diff
@@ -13,9 +13,25 @@
 #include <wayland-server-core.h>
 #include <wlr/types/wlr_buffer.h>
 
+struct wlr_renderer;
+struct wlr_client_buffer;
+
+struct wlr_client_buffer_impl {
+	bool (*is_instance)(struct wl_resource *resource);
+	bool (*get_buffer_size)(struct wl_resource *resource,
```
emersion, line 21: I'd be nice if we could avoid implementing this function, and instead import the buffer before we need to know the size in `wl_surface.commit`'s handler.

include/wlr/types/wlr_client_buffer.h
```diff
@@ -13,9 +13,25 @@
...
+			struct wlr_renderer *renderer, int *width, int *height);
+	bool (*initialize)(struct wlr_client_buffer *,
+			struct wlr_renderer *renderer);
```
emersion, line 24: `initialize` and `terminate` are pretty weird here. The wording should at least be aligned with wlroots' (`init`/`finish`, see CONTRIBUTING).

Do client buffer impls need to store additional fields? Maybe it'd be better to have a function that looks like this?

```c
struct wlr_client_buffer *(*create)(struct wl_resource *resource, struct wlr_renderer *renderer);
```

include/wlr/types/wlr_client_buffer.h
```diff
@@ -13,9 +13,25 @@
 #include <wayland-server-core.h>
 #include <wlr/types/wlr_buffer.h>
 
+struct wlr_renderer;
+struct wlr_client_buffer;
+
+struct wlr_client_buffer_impl {
+	bool (*is_instance)(struct wl_resource *resource);
+	bool (*get_buffer_size)(struct wl_resource *resource,
+			struct wlr_renderer *renderer, int *width, int *height);
```
emersion, line 22: The `struct wlr_renderer *` argument is only required because of `wl_drm`. Would be nice to get rid of it somehow.

include/wlr/types/wlr_client_buffer.h
bl4ckb0ne: Yeah atm it's only used for shm buffers, what's your idea for `wlr_texture_write_pixels` ?

include/wlr/types/wlr_client_buffer.h
bl4ckb0ne: I don't have a use case for custom buffer storing additional. Maybe we could expose both `init` and `create`? IMO there's a lot of stuff that the user could miss by creating the buffer themselves.

include/wlr/types/wlr_client_buffer.h
bl4ckb0ne: As soon as https://github.com/swaywm/wlroots/pull/2708 lands

include/wlr/types/wlr_client_buffer.h
bl4ckb0ne: Maybe we could wrap it in the `init`? The function is also uses in [`wlr_surface`](https://github.com/swaywm/wlroots/blob/master/types/wlr_surface.c#L176), do you think we can do without by storing the size in the `wlr_client_buffer` ?

emersion: The size is in `wlr_client_buffer.base.{width,height}`.

bl4ckb0ne: (made the edit while you were commenting :p )

include/wlr/types/wlr_client_buffer.h
bl4ckb0ne: Seems like create/destroy makes the most sense, will do the changes.

emersion:
Hm, it seems like the most recent changes in this PR have been lost in a rebase. This PR looks like it's been reverted to a previous iteration. Or am I missing something?

emersion:
Ah. Wrong PR.

Let's close this one in favor of https://github.com/swaywm/wlroots/pull/3046.

PR #2936, opened 2021-05-22
Dispatch events in virtual pointer example

puffnfresh:


emersion:
This shouldn't matter, since the client will exit right after, so all client resources will get destroyed?

kennylevinsen:
> This shouldn't matter, since the client will exit right after, so all client resources will get destroyed?

Maybe the intent here is to flush the display, rather than dispatching?

puffnfresh:
@kennylevinsen yes thanks, I think you're right.

emersion:
Oh, I see, I forgot that this is a one-off CLI demo that just sends a bunch of requests then exits. Yes, please call `wl_display_flush` instead.

emersion (approved):
Thanks!

PR #2940, opened 2021-05-27
render: add renderer-specific functions

zccrs:
for wlr_renderer:
wlr_renderer_get_current_buffer

for pixmap:
wlr_renderer_is_pixman
wlr_texture_is_pixman
wlr_pixman_texture_get_attribs
wlr_pixman_buffer_get_attribs

for gles:
wlr_renderer_is_gles2
wlr_gles2_buffer_get_attribs

### DIFF
--- a/include/wlr/render/gles2.h
+++ b/include/wlr/render/gles2.h
@@ -21,6 +21,10 @@ struct wlr_renderer *wlr_gles2_renderer_create(struct wlr_egl *egl);
 struct wlr_egl *wlr_gles2_renderer_get_egl(struct wlr_renderer *renderer);
 bool wlr_gles2_renderer_check_ext(struct wlr_renderer *renderer,
 	const char *ext);
+/**
+ * Returns the OpenGL FBO of current buffer.
+ */
+GLuint wlr_gles2_renderer_get_current_fbo(struct wlr_renderer *wlr_renderer);
 
 struct wlr_gles2_texture_attribs {
 	GLenum target; /* either GL_TEXTURE_2D or GL_TEXTURE_EXTERNAL_OES */
--- a/include/wlr/render/gles2.h
+++ b/include/wlr/render/gles2.h
@@ -30,6 +30,7 @@ struct wlr_gles2_texture_attribs {
 	bool has_alpha;
 };
 
+bool wlr_renderer_is_gles2(struct wlr_renderer *wlr_renderer);
 bool wlr_texture_is_gles2(struct wlr_texture *texture);
 void wlr_gles2_texture_get_attribs(struct wlr_texture *texture,
 	struct wlr_gles2_texture_attribs *attribs);
--- a/include/wlr/render/pixman.h
+++ b/include/wlr/render/pixman.h
@@ -14,6 +14,11 @@
 #include <wlr/render/wlr_renderer.h>
 
 struct wlr_renderer *wlr_pixman_renderer_create(void);
+/**
+ * Returns the image of current buffer.
+ */
+pixman_image_t *wlr_pixman_renderer_get_current_image(
+	struct wlr_renderer *wlr_renderer);
 
 bool wlr_renderer_is_pixman(struct wlr_renderer *wlr_renderer);
 bool wlr_texture_is_pixman(struct wlr_texture *texture);
--- a/include/wlr/render/pixman.h
+++ b/include/wlr/render/pixman.h
@@ -9,9 +9,14 @@
 #ifndef WLR_RENDER_PIXMAN_H
 #define WLR_RENDER_PIXMAN_H
 
+#include <pixman.h>
 #include <wlr/backend.h>
 #include <wlr/render/wlr_renderer.h>
 
 struct wlr_renderer *wlr_pixman_renderer_create(void);
+
+bool wlr_renderer_is_pixman(struct wlr_renderer *wlr_renderer);
+bool wlr_texture_is_pixman(struct wlr_texture *texture);
+pixman_image_t *wlr_pixman_texture_get_image(struct wlr_texture *wlr_texture);
 
 #endif
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -919,3 +919,9 @@ bool wlr_gles2_renderer_check_ext(struct wlr_renderer *wlr_renderer,
 	struct wlr_gles2_renderer *renderer = gles2_get_renderer(wlr_renderer);
 	return check_gl_ext(renderer->exts_str, ext);
 }
+
+GLuint wlr_gles2_renderer_get_current_fbo(struct wlr_renderer *wlr_renderer) {
+	struct wlr_gles2_renderer *renderer = gles2_get_renderer(wlr_renderer);
+	assert(renderer->current_buffer);
+	return renderer->current_buffer->fbo;
+}
--- a/render/gles2/renderer.c
+++ b/render/gles2/renderer.c
@@ -28,9 +28,13 @@ static const GLfloat verts[] = {
 
 static const struct wlr_renderer_impl renderer_impl;
 
+bool wlr_renderer_is_gles2(struct wlr_renderer *wlr_renderer) {
+	return wlr_renderer->impl == &renderer_impl;
+}
+
 struct wlr_gles2_renderer *gles2_get_renderer(
 		struct wlr_renderer *wlr_renderer) {
-	assert(wlr_renderer->impl == &renderer_impl);
+	assert(wlr_renderer_is_gles2(wlr_renderer));
 	return (struct wlr_gles2_renderer *)wlr_renderer;
 }
 
--- a/render/pixman/renderer.c
+++ b/render/pixman/renderer.c
@@ -12,9 +12,13 @@
 
 static const struct wlr_renderer_impl renderer_impl;
 
+bool wlr_renderer_is_pixman(struct wlr_renderer *wlr_renderer) {
+	return wlr_renderer->impl == &renderer_impl;
+}
+
 static struct wlr_pixman_renderer *get_renderer(
 		struct wlr_renderer *wlr_renderer) {
-	assert(wlr_renderer->impl == &renderer_impl);
+	assert(wlr_renderer_is_pixman(wlr_renderer));
 	return (struct wlr_pixman_renderer *)wlr_renderer;
 }
 
@@ -31,9 +35,13 @@ static struct wlr_pixman_buffer *get_buffer(
 
 static const struct wlr_texture_impl texture_impl;
 
+bool wlr_texture_is_pixman(struct wlr_texture *texture) {
+	return texture->impl == &texture_impl;
+}
+
 static struct wlr_pixman_texture *get_texture(
 		struct wlr_texture *wlr_texture) {
-	assert(wlr_texture->impl == &texture_impl);
+	assert(wlr_texture_is_pixman(wlr_texture));
 	return (struct wlr_pixman_texture *)wlr_texture;
 }
 
@@ -553,4 +561,9 @@ struct wlr_renderer *wlr_pixman_renderer_create(void) {
 	}
 
 	return &renderer->wlr_renderer;
+}
+
+pixman_image_t *wlr_pixman_texture_get_image(struct wlr_texture *wlr_texture) {
+	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
+	return texture->image;
 }
--- a/render/pixman/renderer.c
+++ b/render/pixman/renderer.c
@@ -567,3 +567,10 @@ pixman_image_t *wlr_pixman_texture_get_image(struct wlr_texture *wlr_texture) {
 	struct wlr_pixman_texture *texture = get_texture(wlr_texture);
 	return texture->image;
 }
+
+pixman_image_t *wlr_pixman_renderer_get_current_image(
+		struct wlr_renderer *wlr_renderer) {
+	struct wlr_pixman_renderer *renderer = get_renderer(wlr_renderer);
+	assert(renderer->current_buffer);
+	return renderer->current_buffer->image;
+}


zccrs:
This may require adding some new interfaces.

For renderer:

````
wlr_output_attach_render(output);
buffer = wlr_renderer_get_current_buffer(render);

if (wlr_is_gles2_render(render)) {
    gles2_buffer = wlr_gles2_get_buffer(buffer);
    fbo = wlr_gles2_buffer_get_fbo(gles2_buffer);
    // use the fbo with Qt
    // do somethings.
    // .....
} else if (wlr_is_pixman_render(render)) {
    pixman_buffer = wlr_pixman_get_buffer(buffer);
    pixman_image = wlr_pixman_buffer_get_image(pixman_buffer);
    QImage image = pixmanImageToQImage(pixman_image);
    QPainter pa(&image);
    // do somethings
}

wlr_output_commit(output);
````

For surface:

````
texture = wlr_surface_get_texture(surface);

if (wlr_is_gles2_texture(texture)) {
    attr = wlr_gles2_texture_get_attribs(texture);
    // use the attr.tex with Qt
    // do somethings.
    // ...
} else if (wlr_is_pixman_texture(texture)) {
    attr = wlr_pixman_texture_get_attribs(texture);
    // use the attr.pixman_image with Qt
    // ...
}
````

I think should add the "wlr_renderer_get_current_buffer" "wlr_gles2_buffer_get_fbo" "wlr_pixman_buffer_get_image" "wlr_pixman_texture_get_attribs" interfaces(Function names need to be carefully designed ).

File: include/wlr/render/gles2.h [hunk 1]
```diff
@@ -30,8 +30,19 @@ struct wlr_gles2_texture_attribs {
 	bool has_alpha;
 };
 
+struct wlr_buffer;
+struct wlr_gles2_buffer_attribs {
+	struct wlr_buffer *buffer;
+
+	GLuint rbo;
```
ammen99, line 37: I'm wondering whether the `rbo` is going to be useful for compositors.

ammen99 (commented):
> wlr_renderer_is_gles2
> wlr_gles2_renderer_get_current_buffer_attribs

These two have +1 from me, I would like to have this in Wayfire, currently, we have some workarounds for that by using `glGetIntegerv(GL_FRAMEBUFFER_BINDING);`.

I cannot say anything about the pixman renderer.

include/wlr/render/gles2.h
zccrs: I am also not need the RBO, maybe remove it from wlr_gles2_buffer_attribs ?

emersion:
`wlr_renderer_is_pixman`, `wlr_texture_is_pixman` and `wlr_renderer_is_gles2` looks good to me.

include/wlr/render/pixman.h
```diff
@@ -9,9 +9,30 @@
...
+#include <pixman.h>
 #include <wlr/backend.h>
 #include <wlr/render/wlr_renderer.h>
 
 struct wlr_renderer *wlr_pixman_renderer_create(void);
 
+struct wlr_pixman_texture_attribs {
+	void *data;
+	pixman_image_t *image;
+	const struct wlr_pixel_format_info *format;
```
emersion, line 21: Do callers really need the `struct wlr_pixel_format_info`?

include/wlr/render/pixman.h
```diff
@@ -9,9 +9,30 @@
...
 #define WLR_RENDER_PIXMAN_H
 
+#include <pixman.h>
 #include <wlr/backend.h>
 #include <wlr/render/wlr_renderer.h>
 
 struct wlr_renderer *wlr_pixman_renderer_create(void);
 
+struct wlr_pixman_texture_attribs {
+	void *data;
```
emersion, line 19: What is this `data` pointer used for?

emersion:
I'm not sure about `wlr_gles2_renderer_get_current_buffer_attribs` and `wlr_pixman_renderer_get_current_buffer_attribs`. Though, I can see @ammen99's point about making it easier to extend the FBO with a depth buffer.

include/wlr/render/pixman.h
zccrs: No

include/wlr/render/pixman.h
zccrs: Not use in caller. This is copy from struct wlr_pixman_texture.

emersion: We add `data` pointers in structures allocated by wlroots to allow compositors to attach their own state to wlroots' objects. However this struct is not allocated by wlroots, so I'd rather not add this field.

zccrs: Ok, I  haved remove it form wlr_pixman_texture_attribs, it is only contains pixman_image_t now.

zccrs:
> I'm not sure about `wlr_gles2_renderer_get_current_buffer_attribs` and `wlr_pixman_renderer_get_current_buffer_attribs`. Though, I can see @ammen99's point about making it easier to extend the FBO with a depth buffer.

Using the wlr_*_renderer_get_current_buffer_attribs will allow the compositor without wlr_renderer for render.

zccrs:
@ammen99 Hi, Do you have some mean for this PR?

include/wlr/render/wlr_renderer.h
```diff
@@ -127,4 +127,10 @@ int wlr_renderer_get_drm_fd(struct wlr_renderer *r);
  */
 void wlr_renderer_destroy(struct wlr_renderer *renderer);
 
+/**
+ * Return the buffer of called 'wlr_renderer_bind_buffer' after.
+ * If no a active buffer then return NULL.
+ */
+struct wlr_buffer *wlr_renderer_get_current_buffer(struct wlr_renderer *renderer);
```
emersion, line 134: This is only useful if the caller doesn't know about the `wlr_buffer`. Instead of adding this function, I'd prefer to make it easier for compositors to allocate their own buffers and call `wlr_renderer_begin_with_buffer`.

include/wlr/render/gles2.h
```diff
@@ -30,8 +30,17 @@ struct wlr_gles2_texture_attribs {
 	bool has_alpha;
 };
 
+struct wlr_buffer;
```
emersion, line 33: This declaration is unused.

include/wlr/render/gles2.h
```diff
@@ -30,8 +30,17 @@ struct wlr_gles2_texture_attribs {
...
+struct wlr_gles2_buffer_attribs {
+	GLuint fbo;
+};
+
+bool wlr_renderer_is_gles2(struct wlr_renderer *wlr_renderer);
 bool wlr_texture_is_gles2(struct wlr_texture *texture);
 void wlr_gles2_texture_get_attribs(struct wlr_texture *texture,
 	struct wlr_gles2_texture_attribs *attribs);
+void wlr_gles2_buffer_get_attribs(struct wlr_renderer *wlr_renderer,
+	struct wlr_buffer *wlr_buffer,
```
emersion, line 43: Hm, this potentially operates on a buffer which isn't being used as a render target (via `wlr_renderer_bind_buffer`)... Is there a real use-case, or is it fine to operate on the currently bound buffer?

include/wlr/render/gles2.h
```diff
@@ -30,8 +30,17 @@ struct wlr_gles2_texture_attribs {
 	bool has_alpha;
 };
 
+struct wlr_buffer;
+struct wlr_gles2_buffer_attribs {
+	GLuint fbo;
```
emersion, line 35: Since the attribs struct only has a single field, maybe we can just have a function that directly returns the FBO?

emersion:
Do you have a compositor patch where these new functions are used?

include/wlr/render/wlr_renderer.h
zccrs: Yes, wlr_renderer_begin_with_buffer is better, do you have a plan about it? I look the https://github.com/swaywm/wlroots/pull/2903, when remove the wlr_output_attach_render, What kind of solution will replace it?

include/wlr/render/gles2.h
zccrs: This is to prevent the need to add new content in the future, I don't know if there will always be only fbo in the future

include/wlr/render/gles2.h
zccrs: What I can think of now is that it is only used after wlr_output_attach_render, because wlr_renderer_bind_buffer is the internal behavior of wlr_output.

zccrs:
> Do you have a compositor patch where these new functions are used?

````cpp
class DWOutput : public QObject
{
public:
    // begin slot function
    void on_frame(void *);
    // end slot function

    void init();
    void updateProjection();
    void maybeUpdatePaintDevice();

    inline wlr_renderer *renderer() const {
        return wlr_backend_get_renderer(handle->backend);
    }

    inline bool makeCurrent() {
        if (context)
            return context->makeCurrent(surface);

        return wlr_output_attach_render(handle, nullptr);
    }

    inline void doneCurrent() {
        if (context)
            context->doneCurrent();
        else
            handle->impl->rollback_render(handle);
    }

    bool doRender();

    inline void maybeRender() {
        if (Q_UNLIKELY(!renderable || !contentIsDirty))
            return;

        renderable = false;
        contentIsDirty = false;

        doRender();
    }

    inline void render() {
        contentIsDirty = true;
        maybeRender();
    }

    inline void maybeRenderCursor() {
        if (handle->hardware_cursor)
            return;
        contentIsDirty = true;
        maybeRender();
    }

    wlr_output *handle = nullptr;

    QQuickRenderControl *rc = nullptr;

    // for software renderer
    QImage paintDevice;
    pixman_image_t *currentBuffer = nullptr;

    bool renderable = false;
    bool contentIsDirty = false;

    QMatrix4x4 projection;
    QSurfaceFormat format;
    DWOpenGLContext *context;

    DWSignalConnector sc;
};

class DWOpenGLContext : public QPlatformOpenGLContext {
public:
    void swapBuffers(QPlatformSurface *surface) override {
        if (DWOutput *output = getOutput(surface)) {
            wlr_output_commit(output->handle);
        }
    }
    GLuint defaultFramebufferObject(QPlatformSurface *surface) const override {
        if (DWOutput *output = getOutput(surface)) {
            if (!wlr_renderer_is_gles2(output->renderer()))
                return 0;

            wlr_buffer *current_buffer = wlr_renderer_get_current_buffer(output->render());
            wlr_gles2_buffer_attribs attribs;
            wlr_gles2_buffer_get_attribs(output->renderer(), current_buffer, &attribs);

            return attribs.fbo;
        }

        return 0;
    }

    bool makeCurrent(QPlatformSurface *surface) override {
        if (DWOutput *output = getOutput(surface)) {
            return wlr_output_attach_render(output->handle, nullptr);
        }

        return false;
    }
    void doneCurrent() override {
        if (DWOutput *output = getOutput(m_context->surface()->surfaceHandle())) {
            output->handle->impl->rollback_render(output->handle);
        }
    }

    QFunctionPointer getProcAddress(const char *procName) override {
        return eglGetProcAddress(procName);
    }

    void setContext(QOpenGLContext *context) {
        m_context = context;
    }

private:
    QOpenGLContext *m_context = nullptr;
};

void DWOutput::on_frame(void *)
{
    renderable = true;
    maybeRender();
}

void DWOutput::init()
{
    sc.connect(&handle->events.frame,
               this, &DWOutput::on_frame);
    renderable = true;

    updateProjection();

    if (wlr_renderer_is_pixman(renderer())) {
        rc->initialize(nullptr);
    } else {
        format.setRenderableType(QSurfaceFormat::OpenGLES);
        auto egl = wlr_gles2_renderer_get_egl(renderer());
        auto eglConfig = q_configFromGLFormat(egl->display, format, false, EGL_WINDOW_BIT);
        format = q_glFormatFromConfig(egl->display, eglConfig, format);
        context = new DWOpenGLContext();

        if (!makeCurrent()) {
            qFatal("Failed on init DWOutput");
        }

        rc->initialize(context);
    }

    QObject::connect(rc, &QQuickRenderControl::renderRequested,
                     q, &DWOutput::render, Qt::QueuedConnection);
    QObject::connect(rc, &QQuickRenderControl::sceneChanged,
                     q, &DWOutput::render, Qt::QueuedConnection);

    // try do render for init
    render();
}

void DWOutput::updateProjection()
{
    projection = QMatrix4x4();
    projection.rotate(-90 * handle->pending.transform, 0, 0, 1);

    QSizeF size = handle->pending.transform % 2 ? QSize(handle->height, handle->width)
                                                       : this->size();

    if (handle->pending.scale > 0) {
        size /= handle->pending.scale;

        if (!paintDevice.isNull())
            paintDevice.setDevicePixelRatio(handle->pending.scale);
    }

    QRectF rect(QPointF(0, 0), size);

    if (Q_LIKELY(true)) {
        projection.ortho(rect.x(),
                         rect.x() + rect.width(),
                         rect.y(),
                         rect.y() + rect.height(),
                         1,
                         -1);
    } else {
        projection.ortho(rect.x(),
                         rect.x() + rect.width(),
                         rect.y() + rect.height(),
                         rect.y(),
                         1,
                         -1);
    }
}

void DWOutput::maybeUpdatePaintDevice()
{
    if (!wlr_renderer_is_pixman(renderer()))
        return;

    wlr_buffer *current_buffer = wlr_renderer_get_current_buffer(renderer());
    wlr_pixman_buffer_attribs attribs;
    wlr_pixman_buffer_get_attribs(renderer(), current_buffer, &attribs);
    void *data = pixman_image_get_data(attribs.image);

    if (Q_LIKELY(paintDevice.constBits() == data))
        return;

    D_QC(DWOutput);
    QQuickWindow *quick_window = attachedWindow();
    QQuickWindowPrivate *wd = QQuickWindowPrivate::get(quick_window);
    QSGSoftwareRenderer *qt_renderer = dynamic_cast<QSGSoftwareRenderer*>(wd->renderer);
    Q_ASSERT(qt_renderer);

    paintDevice = DWTools::fromPixmanImage(attribs.image, data);
    Q_ASSERT(!paintDevice.isNull());
    paintDevice.setDevicePixelRatio(handle->pending.scale);
    qt_renderer->setCurrentPaintDevice(&paintDevice);
}

bool DWOutput::doRender()
{
    if (Q_UNLIKELY(!makeCurrent())) {
        return false;
    }

    rc->sync();
    maybeUpdatePaintDevice();
    rc->render();

    if (!handle->hardware_cursor) {
        auto renderer = this->renderer();
        wlr_renderer_begin(renderer, handle->width, handle->height);
        /* Hardware cursors are rendered by the GPU on a separate plane, and can be
         * moved around without re-rendering what's beneath them - which is more
         * efficient. However, not all hardware supports hardware cursors. For this
         * reason, wlroots provides a software fallback, which we ask it to render
         * here. wlr_cursor handles configuring hardware vs software cursors for you,
         * and this function is a no-op when hardware cursors are in use. */
        wlr_output_render_software_cursors(handle, nullptr);
        wlr_renderer_end(renderer);
    }

    if (context)
        context->functions()->glFlush();

    return wlr_output_commit(handle);
}

````

File: include/wlr/render/gles2.h (hunk 5 above)
Comment (zccrs) [reply include/wlr/render/gles2.h]: Done

File: include/wlr/render/gles2.h (hunk 6 above)
Comment (zccrs) [reply include/wlr/render/gles2.h]: Done

File: include/wlr/render/gles2.h (hunk 7 above)
Comment (zccrs) [reply include/wlr/render/gles2.h]: Done

File: include/wlr/render/wlr_renderer.h (hunk 4 above)
Comment (emersion) [reply include/wlr/render/wlr_renderer.h:134]: > What kind of solution will replace it?

Instead, compositors will be able to create their own swapchain with `wlr_swapchain_create`, acquire a buffer from it, render to it via `wlr_renderer_begin_with_buffer`, and attach it to the output via `wlr_output_attach_buffer`. We may introduce additional helpers to make this whole dance easier.

Note, `wlr_output_attach_render` will likely stay around for a while, removing it would be a very hard breaking change for all compositors. But the old and new APIs can coexist.

Comment (zccrs) [reply include/wlr/render/wlr_renderer.h]: OK, I have remove the interface, wait the wlr_renderer_begin_with_buffer instead it.

File: include/wlr/render/pixman.h [hunk 8]
```diff
@@ -9,9 +9,24 @@
...
+pixman_image_t *wlr_pixman_renderer_get_current_image(
+	struct wlr_renderer *wlr_renderer);
+
+struct wlr_pixman_texture_attribs {
+	pixman_image_t *image;
+};
+
+bool wlr_renderer_is_pixman(struct wlr_renderer *wlr_renderer);
+bool wlr_texture_is_pixman(struct wlr_texture *texture);
+void wlr_pixman_texture_get_attribs(struct wlr_texture *wlr_texture,
```
emersion, line 29: One last comment: can we make this `wlr_pixman_texture_get_image` and return a `pixman_image_t` directly instead of having a struct with only one field?

zccrs: Done

emersion (approved):
I've slightly reworded the commit messages (we don't use the "feat" prefix).

This LGTM, thanks!

Issue #2941, opened 2021-05-28
Crash on types/wlr_buffer.c:52: wlr_buffer_unlock: Assertion `buffer->n_locks > 0' failed


zccrs:
Steps:

1. edit the `tinywl/tinywl.c` main function to:
add a new line code: "wlr_backend_destroy(server.backend);"
````
	wl_display_run(server.wl_display);

        wlr_backend_destroy(server.backend);
        
	wl_display_destroy_clients(server.wl_display);
	wl_display_destroy(server.wl_display);
	return 0;
````

2. run the tinywl
3. open a wayland application
4. press the "Alt+ESC" shortcut to quit the tinywl

the log:

````
00:00:00.000 [INFO] [backend/x11/backend.c:399] Creating X11 backend
00:00:00.211 [INFO] [render/egl.c:318] Using EGL 1.5
...........
00:00:00.216 [DEBUG] [render/allocator.c:30] Trying to create gbm allocator
00:00:00.218 [DEBUG] [GLES2] Shader Stats: SGPRS: 24 VGPRS: 28 Code Size: 116 LDS: 0 Scratch: 0 Max Waves: 0 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
00:00:00.218 [DEBUG] [GLES2] Shader Stats: SGPRS: 24 VGPRS: 28 Code Size: 124 LDS: 0 Scratch: 0 Max Waves: 0 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
00:00:00.219 [DEBUG] [GLES2] Shader Stats: SGPRS: 16 VGPRS: 8 Code Size: 132 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
00:00:00.219 [DEBUG] [GLES2] Shader Stats: SGPRS: 24 VGPRS: 8 Code Size: 152 LDS: 0 Scratch: 0 Max Waves: 10 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
00:00:00.219 [DEBUG] [GLES2] Shader Stats: SGPRS: 8 VGPRS: 28 Code Size: 44 LDS: 0 Scratch: 0 Max Waves: 0 Spilled SGPRs: 0 Spilled VGPRs: 0 PrivMem VGPRs: 0
[...]

PR #2944, opened 2021-05-29
render/gles2: fix texture cleanup on destroy

emersion:
When importing a DMA-BUF wlr_buffer as a wlr_texture, the GLES2
renderer caches the result, in case the buffer is used for texturing
again in the future. When the wlr_texture is destroyed by the caller,
the wlr_buffer is unref'ed, but the wlr_gles2_texture is kept around.
This is fine because wlr_gles2_texture listens for wlr_buffer's destroy
event to avoid any use-after-free.

However, with this logic wlr_texture_destroy doesn't "really" destroy
the wlr_gles2_texture. It just decrements the wlr_buffer ref'count.
Each wlr_texture_destroy call must have a matching prior
wlr_texture_create_from_buffer call or the ref'counting will go south.

Wehn destroying the renderer, we don't want to decrement any wlr_buffer
ref'count. Instead, we want to go through any cached wlr_gles2_texture
and destroy our GL state. So instead of calling wlr_texture_destroy, we
need to call our internal gles2_texture_destroy function.

Closes: https://github.com/swaywm/wlroots/issues/2941

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -128,6 +128,7 @@ struct wlr_texture *gles2_texture_from_dmabuf(struct wlr_renderer *wlr_renderer,
 	struct wlr_dmabuf_attributes *attribs);
 struct wlr_texture *gles2_texture_from_buffer(struct wlr_renderer *wlr_renderer,
 	struct wlr_buffer *buffer);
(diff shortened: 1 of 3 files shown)


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #2947, opened 2021-05-31
backend: stop using renderer to get the buffer type

emersion:
When picking a format, the backend needs to know whether the
buffers allocated by the allocator will be DMA-BUFs or shared
memory. So far, the backend used the renderer's supported
buffer types to guess this information.

This is pretty fragile: renderers in general don't care about the
SHM cap (they only care about the DATA_PTR one). Additionally,
nothing stops a renderer from supporting both DMA-BUFs and shared
memory, but this would break the backend's guess.

Instead, use wlr_allocator.buffer_caps. This is more reliable since
the buffers created with the allocator are guaranteed to have these
caps.

cc @bl4ckb0ne 

* * *

Not a breaking change since `wlr_allocator` is still a private wlroots API.

### DIFF
--- a/backend/wayland/backend.c
+++ b/backend/wayland/backend.c
@@ -17,6 +17,7 @@
 
 #include "backend/backend.h"
 #include "backend/wayland.h"
+#include "render/allocator.h"
 #include "render/drm_format_set.h"
 #include "render/pixel_format.h"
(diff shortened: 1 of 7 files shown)


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #2948, opened 2021-06-01
surface: allow placing subsurfaces below parent

emersion:
Prior to this commit, subsurfaces could only be placed above their
parent. Any place_{above,below} request involving the parent would
fail with a protocol error.

However the Wayland protocol allows using the parent surface in the
place_{above,below} requests, and allows subsurfaces to be placed
below their parent.

Weston's implementation adds a dummy wl_list node in the subsurface
list. However this is potentially dangerous: iterating the list
requires making sure the dummy wl_list node is checked for, otherwise
memory corruption will happen.

Instead, split the list in two: one for subsurfaces above the parent,
the other for subsurfaces below.

Tested with wleird's subsurfaces demo client.

Closes: https://github.com/swaywm/wlroots/issues/1865

* * *

Breaking change: `wlr_surface.subsurfaces` has been split into two lists: `subsurfaces_above` and `subsurfaces_below`.

Sway PR: https://github.com/swaywm/sway/pull/6302

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -139,10 +139,13 @@ struct wlr_surface {
 		struct wl_signal destroy;
 	} events;
 
-	struct wl_list subsurfaces; // wlr_subsurface::parent_link
+	// wlr_subsurface.parent_link
+	struct wl_list subsurfaces_below;
+	struct wl_list subsurfaces_above;
 
-	// wlr_subsurface::parent_pending_link
-	struct wl_list subsurface_pending_list;
+	// wlr_subsurface.parent_pending_link
+	struct wl_list subsurfaces_pending_below;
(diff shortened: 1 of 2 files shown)


kennylevinsen (approved):
LGTM

emersion:
Thanks for the review!

berylline:
I think the only problem that I have with this PR is just me nitpicking, I think:

There's another variable that would have to be checked and taken into consideration for dealing with subsurfaces.

However, this seems to help with organisation and this seems to make subsurfaces easier to deal with and maintain in a way, so I think this is for the best.

As this now-applied PR seems to fix the problem, I'll close #2676 and the delete the branch for it when I get around to doing so.

Thanks for fixing this issue, @emersion!

emersion:
Yeah, having to walk two lists isn't perfect, but the alternative solution was worse (because it's dangerous).

PR #2950, opened 2021-06-01
render: add wlr_renderer_begin_with_buffer

emersion:
This allows compositors to choose a wlr_buffer to render to. This
is a less awkward interface than having to call bind_buffer() before
and after begin() and end().

Closes: https://github.com/swaywm/wlroots/issues/2618

cc @dos1: would something like this be useful to phosh?

### DIFF
--- a/include/wlr/render/wlr_renderer.h
+++ b/include/wlr/render/wlr_renderer.h
@@ -27,6 +27,7 @@ struct wlr_renderer {
 	const struct wlr_renderer_impl *impl;
 
 	bool rendering;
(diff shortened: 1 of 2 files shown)


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

dos1:
> would something like this be useful to phosh?

Yes, looks good; with `wlr_allocator_autocreate` and `wlr_allocator_create_buffer` it allows to implement things like window thumbnails quite elegantly (although they still have to become public API before we're able to use it)

PR #2951, opened 2021-06-01
render: disallow wlr_renderer_destroy while rendering

emersion:
This probably already felt apart, but let's make it explicit that
this is not allowed.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #2954, opened 2021-06-08
render: drop wlr_ prefix from wlr_renderer_bind_buffer

emersion:
Make it clear this function is a private wlroots API and will stay that way. The only reason why it exists is to allow `wlr_output_attach_render` to continue to be supported.

Convert all other call sites to `wlr_renderer_begin_with_buffer`, which is the future-proof public API.


emersion:
Yup, seems to be fixed now!

PR #2955, opened 2021-06-08
output: add generic wlr_output_export_dmabuf implementation 

emersion:
When wlr_output manages its own swap-chain, there's no need to
hook into the backend to grab DMA-BUFs. Instead, maintain a
`wlr_output.front_buffer` field with the latest committed buffer.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks !

PR #2959, opened 2021-06-09
backend/drm: handle per-connector hotplug events

emersion:
When the kernel specifies a connector in the hotplug events, only update that connector. This should speed up hotplug event handling.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2959

Issue #1580, opened 2019-02-26
DRM connector tiling


emersion:
Some monitors might have multiple connectors with tiled screens.

DRM connectors will have a TILE property that can be used to retrieve tile information. See for example how GNOME Shell handles this: https://gitlab.gnome.org/GNOME/mutter/blob/master/src/backends/native/meta-output-kms.c#L170

soulfilleddirt:
Any word on this? 
Any way I can contribute?

emersion:
I think we'll want to expose this information to compositors, and let them decide how to handle these.

In the future when `wlr_output` doesn't have rendering primitives built-in we could maybe add `wlr_output_group`, a `wlr_output` that attaches a subset of its attached buffer to each sub-output.

emersion:
We'll also want to advertise a single logical `wl_output` to clients so that e.g. layer-shell clients work properly.

hanetzer:
I now have two of these displays now, so if someone needs a guinea pig
and perhaps help with the coding and design. I use gentoo so I'm used
to having system state in various forms of broke so feel free to try crazy
things.

coroner21:
It seems this is currently an obstacle for getting my LG 5k display to work with full resolution in sway. Happy to support in any way getting this to work. Any pointers?

Is my understanding correct that we first need to make wlroots conscious of such a specific display type? In a second part this would be exposed to sway and then handled accordingly?

emersion:
Yes. The first step is to parse the TILE property in the DRM backend.

The second step is to introduce something like `wlr_output_group` to make it easy for compositors to support this.

Alternatively we could add transparent support for tiled outputs in wlroots, but in the past we've tried doing magic under the hood and it turned out to be a bad idea. There are tricky details such as repaint timings to take into account. Also compositors may want to have their own custom logic for tiled displays instead of wlroots'.

ammen99:
> The second step is to introduce something like `wlr_output_group` to make it easy for compositors to support this.
> 
> Alternatively we could add transparent support for tiled outputs in wlroots, but in the past we've tried doing magic under the hood and it turned out to be a bad idea. There are tricky details such as repaint timings to take into account. Also compositors may want to have their own custom logic for tiled displays instead of wlroots'.

Can we somehow scan out parts of the buffers on each output? Or the compositor will have to really render the two parts separately?

emersion:
[...]

PR #2961, opened 2021-06-09
Support DRM connector tiling [WIP]

coroner21:
This first part enables parsing of TILE info for the DRM backend and exposing the info into `wlr_output`. Addresses #1580, the remainder is still WIP:

- Add a way to crop a buffer with wlr_output. Probably something like wlr_output_set_src_box? This should translate to the KMS SRC_* props.
- Add wlr_output_group which forwards buffers to multiple child outputs on commit.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -1369,6 +1369,12 @@ void scan_drm_connectors(struct wlr_drm_backend *drm) {
 			parse_edid(&wlr_conn->output, edid_len, edid);
 			free(edid);
 
+			size_t tile_len = 0;
+			uint8_t *tile = get_drm_prop_blob(drm->fd,
+				wlr_conn->id, wlr_conn->props.tile, &tile_len);
+			parse_tile(&wlr_conn->output, tile_len, tile);
+			free(tile);
+
 			char *subconnector = NULL;
 			if (wlr_conn->props.subconnector) {
 				subconnector = get_drm_prop_enum(drm->fd,
--- a/backend/drm/properties.c
+++ b/backend/drm/properties.c
@@ -24,6 +24,7 @@ static const struct prop_info connector_info[] = {
 	{ "DPMS", INDEX(dpms) },
 	{ "EDID", INDEX(edid) },
 	{ "PATH", INDEX(path) },
+	{ "TILE", INDEX(tile) },
 	{ "link-status", INDEX(link_status) },
 	{ "subconnector", INDEX(subconnector) },
 	{ "vrr_capable", INDEX(vrr_capable) },
--- a/backend/drm/util.c
+++ b/backend/drm/util.c
@@ -144,6 +144,41 @@ void parse_edid(struct wlr_output *restrict output, size_t len, const uint8_t *d
 	}
 }
 
+void parse_tile(struct wlr_output *restrict output, size_t len, const uint8_t *data) {
+	if (len > 0) {
+		int ret;
+		ret = sscanf((char*)data, "%d:%d:%d:%d:%d:%d:%d:%d",
+			&output->tile_info.group_id,
+			&output->tile_info.tile_is_single_monitor,
+			&output->tile_info.num_h_tile,
+			&output->tile_info.num_v_tile,
+			&output->tile_info.tile_h_loc,
+			&output->tile_info.tile_v_loc,
+			&output->tile_info.tile_h_size,
+			&output->tile_info.tile_v_size);
+		if(ret != 8)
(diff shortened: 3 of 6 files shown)


emersion:
To add `wlr_output_set_source_box`:

- Add a `source_box` field to `wlr_output_state`, with the matching enum entry. Populate it in `wlr_output_set_source_box`.
- In the DRM backend, expand `set_plane_props` to take a source box in addition to the existing parameters.
- When calling [`set_plane_props` for the primary plane](https://github.com/swaywm/wlroots/blob/103edde4810e35b5a95be4205436239a5ee724c9/backend/drm/atomic.c#L234), grab the source box from `state` if it's set there.

coroner21:
Thanks a lot Simon for the directions. Two questions came up:
- I have the enum and field for `source_box` in `wlr_output_state` and added `wlr_output_set_source_box` to populate that. When would we reset the flag in the `committed` field tracking the state? On my display there is only the full 5k resolution which requires both connectors, the lower scaled resolutions require only one. So probably on modeset we should in general reset the flag. Or should we just ensure the field is always populated for an active output (with default (0,0,width,height) in most cases)? 
- In `set_plane_props` how does the width and height from GBM relate to `source_box`? The GBM values would correspond to the full buffer and the box cuts out parts of it, right?

Note that it seems in `set_plane_props` the `wlr_drm_backend` is not used so I could just replace that parameter with the box. I still struggle to see how the same `set_plane_props` function can be used for both primary and cursor planes. I might make a proposal to have a bit different code layout here instead or just duplicate this function.

emersion:
> When would we reset the flag in the committed field tracking the state?

This is done automatically in `wlr_output_commit`.

> Or should we just ensure the field is always populated for an active output (with default (0,0,width,height) in most cases)?

The pattern so far has been that the fields have undefined value if the flag isn't set.

> In set_plane_props how does the width and height from GBM relate to source_box? The GBM values would correspond to the full buffer and the box cuts out parts of it, right?

Yes, that is correct. So no source box means use `(0, 0, buffer_width, buffer_height)`.

> Note that it seems in set_plane_props the wlr_drm_backend is not used

Indeed, it seems unused ATM. Feel free to submit a separate patch to remove that arg.

> I still struggle to see how the same set_plane_props function can be used for both primary and cursor planes.

The current buffer is taken from the `wlr_drm_plane`. ATM, the SRC_* coords are set to `(0, 0, buffer_width, buffer_height)`. The CRTC_* coords are set to `(x, y, buffer_width, buffer_height)` (x and y are provided as arguments to `set_plane_props`, they are zero for the primary plane and the cursor's position for the cursor plane).

> duplicate this function.

I don't think that's necessary.

coroner21:
OK, the second step should be also completed, now the only pending task is addition of `wlr_output_group`  to collect the tiles, setup the combined buffer and on commit set source boxes for different outputs related to the display and then call commit for the outputs.

Duplicating `wlr_output` I would add a `wlr_output_group` type. Would be great if you could provide some thoughts which parts I would need to duplicate.

Furthermore, how do we want to initialize the output group? Somewhere we need to detect the tiling display and then trigger output group initialization (probably by passing an array of `wlr_output` pointers and total number of tiles).

emersion:
> Duplicating wlr_output

It shouldn't be necessary to duplicate `wlr_output`. Instead, we should be able to design `wlr_output_group` as a new `wlr_output` implementation. Maybe look into the headless backend to see how to do this?

https://github.com/swaywm/wlroots/blob/7cbcc65ad00e195a821cf955dd2302b6b0ec1bf0/backend/headless/output.c

> Furthermore, how do we want to initialize the output group? Somewhere we need to detect the tiling display and then trigger output group initialization (probably by passing an array of wlr_output pointers and total number of tiles).

Yes. As a first step the compositor can manually do it. But ideally we could provide a helper to make it as easy as possible for compositors to support tiled outputs.

One annoying thing is that there's no easy way to list all outputs belonging to a backend, so we can't just have the helper walk the list. Maybe the compositor can collect all outputs with the same tile group and call a function which automatically configures and returns a new `wlr_output_group`? Not sure we can do much better with the current APIs.

emersion:
I've started work on `wlr_output_group` in https://github.com/swaywm/wlroots/pull/3067. It's completely untested, but maybe can help you get started? It'll only work with the DRM backend once https://github.com/swaywm/wlroots/pull/2903 is merged, but should already work fine with other backends. Maybe some bits need to be added for modesetting.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2961

PR #2964, opened 2021-06-11
output: reset back buffer on failed commit

emersion:
On commit failure, we need to unbind the back buffer from the
renderer.

This fixes assertions triggered on commits following a failed commit
where the compositor called wlr_output_attach_render.


kennylevinsen (approved):
LGTM, thanks!

PR #2966, opened 2021-06-14
backend/drm: populate cursor plane's current_fb

emersion:
The set_cursor() hook is a little bit special: it's not really
synchronized to commit() or test(). Once set_cursor() returns true,
the new cursor is part of the current state.

This fixes a state where wlr_drm_connector.cursor_enabled is true
but there is no FB available. This is triggered by set_cursor()
followed by a failed commit(), which resets pending_fb.

We should definitely fix the output interface to make the cursor part
of the pending state, but that's a more involved change.


kennylevinsen (approved):
LGTM, thanks!

Issue #2952, opened 2021-06-01
Static linking of wlroots


kennylevinsen:
wlroots is generally shipped as a shared library. This presents issues for packaging, as packagers must wait for all packaged wlroots-based compositors to be updated before they can bump wlroots, which in turn means that compositor releases depending on the latest release of wlroots end up waiting on other compositors to update. An example of such packaging issues can be seen in alpine here: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/20264.

The simplest way to solve this would be for compositors to ship wlroots as a statically linked subproject, bypassing the need for a shared `libwlroots.so` and thus allowing compositors to use whichever revision they support.

Current blockers for friction-less statically linked subproject:
1. Avoid wlroots installation when built as a subproject. See: https://github.com/mesonbuild/meson/issues/2550, which was resolved with a user flag rather than a meson.build option. This could be worked around in our meson.build file.
2. Policy or solution for internal symbol collisions, such as `get_current_time_msec` conflicting between sway and wlroots (https://github.com/swaywm/wlroots/issues/1892, https://github.com/swaywm/sway/issues/5514).

emersion:
I agree static linking makes sense for something like wlroots.

Regarding (1), packagers will need to add logic to setup the subproject anyways, so I think it's fine to ask them to use the `--skip-subprojects` flag.

emersion:
(Maybe we should update [the wiki page for packagers](https://github.com/swaywm/wlroots/wiki/Packaging-recommendations) once we have (2) addressed.)

kennylevinsen:
Reopened due to revert.

ericonr:
Please don't encourage projects to ship internal wlroots versions. This makes it harder to apply global patches and adds the need for distros to track which wlroots version each package is using. Furthermore, wlroots makes clear releases and bumps soname when it has ABI breaks: this means there are well tested points in time (instead of each project picking a random point in time to tag a release) and that two versions of the library can co-exist. Distros rushing to update things can

- pick up/develop patches for each compositor
- create a `wlroots0.x` package for the packages that haven't been updated yet (which I see is what was implemented in https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/20264#note_160337, and makes this discussion somewhat unnecessary)

emersion:
This issue is not about making it easy to bundle a random wlroots commit in a compositor. This issue is about making it easy to statically link to a wlroots release.

ammen99:
[...]

Issue #1892, opened 2019-11-01
Static library contains private symbols


emersion:
Follow-up issue for https://github.com/swaywm/sway/issues/4677

Potential solutions:

- Find a way to strip our private symbols for the static library, just like the version script we have for the dynamic library
- Prefix our private functions (but then the version script won't work anymore, at least if we prefix with `wlr_`)

cc @sheenobu

kennylevinsen:
- We can't strip private symbols from a static library. While shared libraries are linked binaries with a global symbol table that can be adjusted/filtered, static libraries aren't linked at all, and are just object files stuffed in an archive. Hiding a symbol also hides it from the other parts of wlroots that might need it.
  - Fixing this would require a tool that could "link" object files so that internal dependencies could be resolved and turned static, and output a new bare object file. Haven't found such tool so far.
- Renaming symbols to hide them seems to imply using objcopy to modify the library first, which seems a little nasty.
- Prefixing is possible, but either requires a different prefix or a changed version script as mentioned.

progandy:
> Prefix our private functions (but then the version script won't work anymore, at least if we prefix with wlr_)

Isn't it common to prefix private symbols with an underscore, i.e. ```_wlr_```?

emersion:
I'd be fine with a `_wlr` prefix FWIW.

PR #2969, opened 2021-06-14
meson: Make private static library symbols local

kennylevinsen:
Static libraries are not affected by our symbol file, so private symbols
are globally visible by default.

Use objcopy to make symbols that we do not want to expose local.

Closes: https://github.com/swaywm/wlroots/issues/1892
Closes: https://github.com/swaywm/wlroots/issues/2952

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -150,7 +150,29 @@ lib_wlr = library(
 	install: true,
 	link_args: symbols_flag,
 	link_depends: symbols_file,
+	prelink: true,
 )
+
+if get_option('default_library') != 'shared'
+	lib_target = lib_wlr
+	if get_option('default_library') == 'both'
+		lib_target = lib_wlr.get_static_lib()
+	endif
+	objcopy_prog = find_program('objcopy', native: true)
+	custom_target('libwlroots',
+		input: lib_target,
+		output: lib_target.name() + '.is-stripped',
+		capture: true,
+		command: [
+			objcopy_prog.full_path(), '-w',
+			'--localize-symbol=!wlr_*',
+			'--localize-symbol=!_wlr_*',
+			'--localize-symbol=*',
+			'@INPUT@',
+		],
+		build_by_default: true,
+	)
+endif
 
 wlr_vars = {}
 foreach name, have : features
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -19,7 +19,7 @@ sources:
 tasks:
   - setup: |
       cd wlroots
-      meson build -Dauto_features=enabled -Dlibseat=disabled -Dxcb-errors=disabled
+      meson build --default-library=both -Dauto_features=enabled -Dlibseat=disabled -Dxcb-errors=disabled
   - build: |
       cd wlroots
       ninja -C build
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -20,7 +20,7 @@ sources:
 tasks:
   - setup: |
       cd wlroots
-      CC=gcc meson build-gcc -Dauto_features=enabled --prefix /usr
+      CC=gcc meson build-gcc --default-library=both -Dauto_features=enabled --prefix /usr
       CC=clang meson build-clang -Dauto_features=enabled
   - gcc: |
       cd wlroots/build-gcc


meson.build
```diff
@@ -150,8 +150,25 @@ lib_wlr = library(
...
+	lib_target = lib_wlr
+	if get_option('default_library') == 'both'
+		lib_target = lib_wlr.get_static_lib()
+	endif
+	objcopy_prog = find_program('objcopy', native: true)
+	custom_target('libwlroots',
+		input: lib_wlr,
+		output: lib_wlr.name() + '.is-stripped',
+		command: [
+			'sh', '-c', 'objcopy -w -L "!wlr*" -L "!_wlr*" -L "*" @INPUT@; touch @OUTPUT@'.format(objcopy_prog.full_path())
```
emersion, line 166: Instead of hardcoding `objcopy`, we probably want to use `@0@`

kennylevinsen: Yeah, this format string was very broken. Side-stepped the shell entirely now.

meson.build
```diff
@@ -150,8 +150,30 @@ lib_wlr = library(
...
+		lib_target = lib_wlr.get_static_lib()
+	endif
+	objcopy_prog = find_program('objcopy', native: true)
+	custom_target('libwlroots',
+		input: lib_target,
+		output: lib_target.name() + '.is-stripped',
+		capture: true,
+		command: [
+			objcopy_prog.full_path(), '-w',
+			'--localize-symbol=!wlr*',
```
emersion, line 168: Maybe append another `_` to the prefix? That way silly stuff like `wlresource` doesn't get exported.

kennylevinsen: Good point

meson.build
```diff
@@ -150,8 +150,30 @@ lib_wlr = library(
...
+	custom_target('libwlroots',
+		input: lib_target,
+		output: lib_target.name() + '.is-stripped',
+		capture: true,
+		command: [
+			objcopy_prog.full_path(), '-w',
+			'--localize-symbol=!wlr_*',
+			'--localize-symbol=!_wlr_*',
+			'--localize-symbol=*',
+			lib_wlr.full_path(),
```
emersion, line 171: This should be `lib_target`. Maybe use `@INPUT@`?

kennylevinsen: Fixed

kennylevinsen:
Meson lacks support for prelinking on clang. See https://github.com/mesonbuild/meson/pull/8883.

emersion (commented):
LGTM with static libs disabled on Clang on BSD and Arch (waiting for the Meson patch to get in)

kennylevinsen:
> LGTM with static libs disabled on Clang on BSD and Arch (waiting for the Meson patch to get in)

Do you want to conditionally disable the feature so that existing static linking workarounds remain possible?

emersion:
> Do you want to conditionally disable the feature so that existing static linking workarounds remain possible?

As you prefer. I don't really like hardcoding specific compilers and Meson versions, but making all Clang builds fail when static libs are enabled isn't nice, especially if we're going to recommend static libs.

If we don't set `prelink: true`, we could always set `link_args: '-Wl,--relocatable'` I suppose, and that should work on all linkers of interest?

kennylevinsen:
> If we don't set `prelink: true`, we could always set `link_args: '-Wl,--relocatable'` I suppose, and that should work on all linkers of interest?

What would the idea be here?

Prelinking adds an additional linker step which links all objects into a single object, which runs before packing the final `.a` (which is not linking, but just making an`ar` archive of all final `.o`'s).

I'll make the new feature gcc-only for now, while I look at fixing meson.

emersion:
> Prelinking adds an additional linker step which links all objects into a single object, which runs before packing the final .a (which is not linking, but just making anar archive of all final .o's).

Hrm, indeed. The linker doesn't get invoked when building a static lib. It doesn't seem like `objcopy` has a mode where it can do what `ld --relocatable` does.

Let's just get the CI green for now then, just need to disable `--default-libraries=both` on Clang builds.

kennylevinsen:
Static linking is now only enabled for GCC in CI.

emersion (approved):
LGTM, thanks!

emersion:
Maybe we should add static linking to one of the Sway CI jobs?

PR #2972, opened 2021-06-15
backend/session: use DRM_PRIMARY_MINOR_NAME

emersion:
Instead of hardcoding the string "card", use DRM_PRIMARY_MINOR_NAME.
Some systems may use another prefix, e.g. OpenBSD uses "drm" instead.


kennylevinsen (approved):
LGTM, thanks!

PR #2975, opened 2021-06-18
backend: add output state allow-lists

emersion:
Right now, when a new output state field is added, all backends by
default won't reject it. This means we need to add new checks to
each and every backend when we introduce a new state field.

Instead, introduce a bitmask of supported output state fields in
each backend, and error out if the user has submitted an unknown
field.

Some fields don't need any backend involvment to work. These are
listed in WLR_OUTPUT_STATE_BACKEND_OPTIONAL as a convenience.

### DIFF
--- a/backend/drm/drm.c
+++ b/backend/drm/drm.c
@@ -31,6 +31,13 @@
 #include "types/wlr_buffer.h"
 #include "util/signal.h"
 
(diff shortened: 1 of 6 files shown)


kennylevinsen (approved):
LGTM, thanks!

PR #2976, opened 2021-06-18
output: fail commits if adaptive sync cannot be enabled

emersion:
Previously, adaptive sync was just a hint and wouldn't make any
atomic commit fail if the backend didn't support it. The main reason
is wlr_output_test wasn't supported at the time.

Now that we have a way for compositors to test whether a change can
work, let's remove the exception for adaptive sync and convert it to
a regular output state field.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2976

Issue #2957, opened 2021-06-08
Cached surface state for compositor atomic transactions


emersion:
For the purposes of `wl_subsurface`, the upcoming transactions protocol and explicit sync, we've recently introduced `wlr_surface_lock_pending` and `wlr_surface_unlock_cached`. With a few additions, I believe compositors could use these functions to implement atomic transactions.

The benefit is reduced complexity in compositors: they won't need to manually save the whole surface tree state and buffers anymore.

### What are atomic transactions in compositors?

Atomic transactions are used to wait for clients to submit new buffers without applying a layout change. For instance, wait for the client to submit a new buffer before moving a toplevel's coordinates during an interactive top-left-corner resize. Another example: wait for two clients to resize before showing the new buffers when resizing a tiled layout. This allows for a frame-perfect resize experience.

Atomic transactions typically use xdg-shell's `configure` and `ack_configure` events/requests to know when a client has replied to a compositor-side resize.

Atomic transactions can involve one toplevel (e.g. floating resize) or more (e.g. tiled resize). Depending on compositor policy, there may be multiple atomic transactions in progress at a given point in time (e.g. send 200x200 configure event, send 250x250 configure event, receive new 200x200 buffer, receive new 250x250 buffer).

Compositors known to implement atomic transactions: Sway and river. Wayfire is working on it. rootston had a very primitive mechanism to synchronize a toplevel's coordinates updates with xdg-shell configure events.

Typically transactions are used for resize operations, but are not limited to it. They can also be used for other state updates (example: xdg-decoration changes, but in general can be anything tied to `wl_surface.commit`).

### How do they work at the moment?

The _compositor state_ typically includes the layout (e.g. position of toplevels). The _client state_ is everything in `wlr_surface_state` (and add-on structs like `wlr_xdg_surface_state`). "Cached state" is a Wayland protocol term that refers to state committed by the client, but not yet made current by the compositor.

1. Accumulate pending compositor state changes.
2. The compositor sends a `configure` event with a new size to all surfaces involved in the transaction.
  - The compositor's state changes are not applied immediately, they are cached in the transaction.
  - For all surfaces involved in the transaction, save their client state (incl. the whole surface tree) and buffers.
  - From now on, rendering operations use the saved client state and the old compositor state (which hasn't been touched in this step).
3. The compositor waits for all clients to ACK the transaction with `ack_configure` (and to submit a new buffer with the right size). When this happens:
  - The compositor applies the cached compositor state.
[...]

PR #2977, opened 2021-06-18
xdg-shell: add support for cached state

emersion:
When the pending surface state becomes cached, also cache the
pending xdg_surface state. When a cached surface state becomes
current, also make the matching cached xdg_surface state current.

References: #2957

* * *

The `wlr_surface` API for cached state is not very ergonomic. Ideas welcome to improve it.


types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -346,19 +345,65 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
 	}
 }
 
+static void xdg_surface_handle_surface_cache(struct wl_listener *listener,
+		void *data) {
+	struct wlr_xdg_surface *surface =
+		wl_container_of(listener, surface, surface_cache);
+
+	if (!surface->pending.has_geometry) {
```
ammen99, line 353: I think we should still cache the state here:

- makes no sense if we have other properties too (fullscreen, tiled, etc.)
- some clients do not use window geometry, and leave it to the default of 'bounding box of surface+subsurfaces'. We'd still want to track configure serials for such surfaces, and for that we need to iterate the cached state of the xdg_surface.

emersion: > makes no sense if we have other properties too (fullscreen, tiled, etc.)

Fullscreen/tiled/etc are not part of the double-buffered toplevel state.

> some clients do not use window geometry, and leave it to the default of 'bounding box of surface+subsurfaces'. We'd still want to track configure serials for such surfaces, and for that we need to iterate the cached state of the xdg_surface.

The configure serials are missing from the cached xdg_surface state. We should add it, and only cache the state if the pending geometry or pending configure serial has changed.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -346,19 +345,65 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
...
+
+out:
+	surface->pending.seq = surface->surface->pending.seq;
+}
+
+static bool xdg_surface_pop_state(struct wlr_xdg_surface *surface,
+		uint32_t seq, struct wlr_xdg_surface_state *out) {
+	if (surface->pending.seq == seq) {
+		*out = surface->pending;
+		memset(&surface->pending, 0, sizeof(surface->pending));
```
ammen99, line 372: This thing here resets the `configure_serial` to 0, but it shouldn't.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2977

PR #2980, opened 2021-06-20
Revert "meson: Make private static library symbols local"

kennylevinsen:
This reverts commit 28d23ba6bda4f799b8d6689555cd33a40adda17e.

The prelinking and symbol filtering pass breaks builds with link-time
optimization enabled.

Building static libs in CI should be fine, so I didn't revert that.

cc @emersion


emersion (approved):
It's unfortunate we didn't manage to make this work, but it seems we won't be able to fix it quickly.

Thanks!

PR #2982, opened 2021-06-22
Revert "build: workaround for meson disabler object not working with if not"

emersion:
This reverts commit 9796abccedef881e99d293a2658c4da9466acd88.

This Meson issue has been fixed upstream for a while. We require 0.56.0 so we should never hit an unpatched Meson.


kennylevinsen (approved):
Checks out in a CI test without ffmpeg, thanks!

PR #2985, opened 2021-06-24
render/egl: make most functions private

emersion:
The wlr_egl functions are mostly used internally by the GLES2
renderer. Let's reduce our API surface a bit by hiding them. If
there are good use-cases for one of these, we can always make them
public again.

The functions mutating the current EGL context are not made private
because e.g. Wayfire uses them.

cc @ammen99

### DIFF
--- a/include/render/egl.h
+++ b/include/render/egl.h
@@ -1,0 +1,76 @@
+#ifndef RENDER_EGL_H
+#define RENDER_EGL_H
+
+#include <wlr/render/egl.h>
+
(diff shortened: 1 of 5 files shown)


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, I don't think wxrc uses these as well.

ammen99 (commented):
LGTM, wayfire doesn't use any of the newly hidden functions.

PR #2986, opened 2021-06-24
render/egl: replace wlr_egl_create with wlr_egl_create_with_drm_fd

emersion:
We never create an EGL context with the platform set to something
other than EGL_PLATFORM_GBM_KHR. Let's simplify wlr_egl_create by
taking a DRM FD instead of a (platform, remote_display) tuple.

This hides the internal details of creating an EGL context for a
specific device. This will allow us to transparently use the device
platform (#2671) when the time comes.

~~Depends on: https://github.com/swaywm/wlroots/pull/2985~~

cc @bl4ckb0ne

### DIFF
--- a/include/render/egl.h
+++ b/include/render/egl.h
@@ -11,10 +11,11 @@ struct wlr_egl_context {
 };
 
 /**
- * Initializes an EGL context for the given platform and remote display.
- * Will attempt to load all possibly required api functions.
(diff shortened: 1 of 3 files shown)


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, I'll rebase #2671

PR #2988, opened 2021-06-25
xwayland: terminate when no client is connected

emersion:
This PR takes advantage of [the new Xorg `-terminate <delay>`](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/631) argument to automatically shutdown Xwayland when all X11 clients are gone. We mark our XWM as a special client which shouldn't prevent from Xorg server from exiting.


kennylevinsen:
https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/merge_requests/15 is merged, which replaced https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/merge_requests/18.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/2988

PR #2989, opened 2021-06-25
ci: make Meson warnings fatal

emersion:
New warnings can be hard to notice in CI, since CI will just pass in
that case. Meson sometimes uses warnings for important mistakes, e.g.
invalid option.

Let's turn warnings into errors so that we can spot these more easily.

* * *

Note, we need to bump the Meson version to get rid of the `meson.build_root()` warning.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -19,7 +19,7 @@ sources:
 tasks:
   - setup: |
       cd wlroots
-      meson build --default-library=both -Dauto_features=enabled -Dlibseat=disabled -Dxcb-errors=disabled
+      meson build --default-library=both -Dauto_features=enabled -Dxcb-errors=disabled
   - build: |
       cd wlroots
       ninja -C build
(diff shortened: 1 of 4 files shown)


emersion:
Ah, this already caught a mistake in the Alpine CI build, good.

bl4ckb0ne (approved):

bl4ckb0ne:
Good catch on the alpine CI ;)

vcunat:
I don't mean to be rude, but isn't it a little aggressive to "hard-depend" on a meson version that was just released several days before?  Sure, people who find this (and understand it) can patch it back easily or they may even be able to update their distro's meson, but still...

kennylevinsen:
How is it aggressive that our master (i.e. bleeding edge, unreleased work) depends on other recent releases (not even the latest release at the time of writing this comment) whose features or fixes it needs? In my opinion it seems hypocritical to require the absolute latest unreleased work of a project, but then disallow it from depending on any recent releases.

If your distro does not have a recent meson release, it can easily be fetched with pip.

PR #2994, opened 2021-06-29
buffer: add data_ptr access flags

emersion:
This allows callers to specify the operations they'll perform on
the returned data pointer. The motivations for this are:

- The upcoming Linux MAP_NOSIGBUS flag may only be usable on
  read-only mappings.
- gbm_bo_map with GBM_BO_TRANSFER_READ hurts performance.

References: https://github.com/swaywm/wlroots/issues/2864


bl4ckb0ne (approved):

bl4ckb0ne:
I do think this could be reworked more finely in the future, but for now it's a viable solution to our issue, thanks!

PR #2998, opened 2021-06-30
surface: accept commits with buffer size not divisible by scale

emersion:
There are still many situations where the buffer scale is not
divisible by scale. The fix will require a tad more work, so
let's just log the client error for now and continue handling
the surface commit as usual.

Closes: https://github.com/swaywm/sway/issues/6352


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #3002, opened 2021-07-01
Nuke wlr_list

emersion:
This data structure is mis-named, as it can easily be confused with `wl_list`. Instead, use libwayland's `wl_array`.

* * *

Breaking change: `wlr_tablet.paths` and `wlr_tablet_pad.paths` are now `wl_array` instead of `wlr_list`.


ifreund (approved):
I gave this a good read and didn't find any typos or pointer screwups. There are a few places where allocation failure isn't handled (e.g. in tablet_pad.c) but those weren't handled before this patch either.

Getting rid of wlr_list is great!

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM thanks!

PR #3003, opened 2021-07-01
tablet-v2: fix pad and tool object versions

emersion:
These were hardcoded to 1. Instead, create the resource with the
version of the parent object.


bl4ckb0ne (approved):

bl4ckb0ne:
Good catch ;)

PR #3004, opened 2021-07-01
backend/libinput: use wl_array for wlr_libinput_tablet.tools

emersion:
Instead of using a single-field wl_list, let's just use a wl_array.

~~Depends on: https://github.com/swaywm/wlroots/pull/3002~~


emersion:
This is now ready for review.

PR #3012, opened 2021-07-01
buffer: make enum wlr_buffer_cap public

emersion:
Custom backends and renderers need to implement `wlr_backend_impl.get_buffer_caps` and `wlr_renderer_impl.get_render_buffer_caps`. They can't if `enum wlr_buffer_cap` isn't made public.

* * *

This was an oversight I made during the 0.14.0 release. I'd say it would make sense to backport to 0.14.1.


bl4ckb0ne (approved):

bl4ckb0ne:
Good catch ;)

PR #3023, opened 2021-07-05
Remove wlr_resource_get_buffer_size

### LINKED ISSUE #3015: buffer: get rid of wlr_resource_get_buffer_size
The only reason why we have `wlr_resource_get_buffer_size` is that we historically didn't always have a `wlr_buffer` for all types of `wl_buffer` objects coming from clients. With https://github.com/swaywm/wlroots/pull/2708, this time has passed.

I'd like to remove `wlr_resource_get_buffer_size` because it creates duplication with `wlr_client_buffer_import`: both functions need to handle all existing types of `wl_buffer` objects. This will help with enabling third-party `wl_buffer` implementations as well.

A possible plan is:

- Split `wlr_client_buffer_import` into two functions: one that turns a `wl_resource` into a `wlr_buffer` (e.g. named `wlr_buffer_from_resource`), and another that creates a `wlr_client_buffer` from a `wlr_buffer` (e.g. named `wlr_client_buffer_create`).
- In `wlr_surface_state`, add a new `struct wlr_buffer *buffer` field.
- In `surface_state_finalize`:
  - Before anything else, turn the `buffer_resource` into a `wlr_buffer` (via `wlr_buffer_from_resource`).
  - Get the buffer size from the `wlr_buffer`.
- In `surface_apply_damage`, turn the `wlr_buffer` into a `wlr_client_buffer` (via `wlr_client_buffer_create`), then unlock the `wlr_buffer` (so that the `n_locks` check in `wlr_client_buffer_apply_damage` continues to work).

cc @bl4ckb0ne

bl4ckb0ne:
Fixes #3015 

~~Turns out `wlr_surface_state` buffer member can be avoided, I'm fairly sure there's a way to make it hold only a `wlr_buffer` and get the `buffer_resource`, `buffer_width` and `buffer_height` from it without having to duplicate them. Let me know if you want that in this patch or in another one.~~

cc @emersion 

* * *

Breaking change:

- `wlr_resource_get_buffer_size` has been removed.
- `wlr_client_buffer_import` has been split into two functions: `wlr_buffer_from_resource` and `wlr_client_buffer_create`.

### DIFF
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -157,11 +157,6 @@ struct wlr_client_buffer *wlr_client_buffer_get(struct wlr_buffer *buffer);
  */
 bool wlr_resource_is_buffer(struct wl_resource *resource);
 /**
- * Get the size of a wl_buffer resource.
- */
-bool wlr_resource_get_buffer_size(struct wl_resource *resource,
-	int *width, int *height);
-/**
  * Try to update the buffer's content. On success, returns the updated buffer
  * and destroys the provided `buffer`. On error, `buffer` is intact and NULL is
  * returned.
--- a/include/wlr/types/wlr_buffer.h
+++ b/include/wlr/types/wlr_buffer.h
@@ -14,6 +14,7 @@
 #include <wlr/render/dmabuf.h>
 
 struct wlr_buffer;
+struct wlr_renderer;
 
 struct wlr_shm_attributes {
 	int fd;
@@ -111,6 +112,14 @@ bool wlr_buffer_get_dmabuf(struct wlr_buffer *buffer,
  */
 bool wlr_buffer_get_shm(struct wlr_buffer *buffer,
 	struct wlr_shm_attributes *attribs);
+/**
+ * Transforms a wl_resource into a wlr_buffer and locks it. Once the caller is
+ * done with the buffer, they must call wlr_buffer_unlock.
+ *
+ * The provided wl_resource must be a wl_buffer.
+ */
+struct wlr_buffer *wlr_buffer_from_resource(struct wlr_renderer *renderer,
+	struct wl_resource *resource);
 
 /**
  * A client buffer.
@@ -131,7 +140,12 @@ struct wlr_client_buffer {
 	struct wl_listener resource_destroy;
 };
 
-struct wlr_renderer;
+/**
+ * Creates a wlr_client_buffer from a given wlr_buffer by creating a texture
+ * from it, and copying its wl_resource.
+ */
+struct wlr_client_buffer *wlr_client_buffer_create(struct wlr_buffer *buffer,
+	struct wlr_renderer *renderer, struct wl_resource *resource);
 
 /**
  * Get a client buffer from a generic buffer. If the buffer isn't a client
@@ -147,13 +161,6 @@ bool wlr_resource_is_buffer(struct wl_resource *resource);
  */
 bool wlr_resource_get_buffer_size(struct wl_resource *resource,
 	int *width, int *height);
-/**
- * Import a client buffer and lock it.
- *
- * Once the caller is done with the buffer, they must call wlr_buffer_unlock.
- */
-struct wlr_client_buffer *wlr_client_buffer_import(
-	struct wlr_renderer *renderer, struct wl_resource *resource);
 /**
  * Try to update the buffer's content. On success, returns the updated buffer
  * and destroys the provided `buffer`. On error, `buffer` is intact and NULL is
--- a/types/wlr_buffer.c
+++ b/types/wlr_buffer.c
@@ -194,9 +194,9 @@ static void client_buffer_resource_handle_destroy(struct wl_listener *listener,
 	// which case we'll read garbage. We decide to accept this risk.
 }
 
-struct wlr_client_buffer *wlr_client_buffer_import(
-		struct wlr_renderer *renderer, struct wl_resource *resource) {
-	assert(wlr_resource_is_buffer(resource));
+struct wlr_buffer *wlr_buffer_from_resource(struct wlr_renderer *renderer,
+		struct wl_resource *resource) {
+	assert(resource && wlr_resource_is_buffer(resource));
 
 	struct wlr_buffer *buffer;
 	if (wl_shm_buffer_get(resource) != NULL) {
@@ -219,16 +219,16 @@ struct wlr_client_buffer *wlr_client_buffer_import(
 			wlr_drm_buffer_from_resource(resource);
 		buffer = wlr_buffer_lock(&drm_buffer->base);
 	} else {
-		wlr_log(WLR_ERROR, "Cannot upload texture: unknown buffer type");
-
-		// Instead of just logging the error, also disconnect the client with a
-		// fatal protocol error so that it's clear something went wrong.
-		wl_resource_post_error(resource, 0, "unknown buffer type");
+		wlr_log(WLR_ERROR, "Unknown buffer type");
 		return NULL;
 	}
 
+	return buffer;
+}
+
+struct wlr_client_buffer *wlr_client_buffer_create(struct wlr_buffer *buffer,
+		struct wlr_renderer *renderer, struct wl_resource *resource) {
 	struct wlr_texture *texture = wlr_texture_from_buffer(renderer, buffer);
-	wlr_buffer_unlock(buffer);
 	if (texture == NULL) {
 		wlr_log(WLR_ERROR, "Failed to create texture");
 		wl_buffer_send_release(resource);
@@ -307,10 +307,6 @@ struct wlr_client_buffer *wlr_client_buffer_apply_damage(
 	}
 
 	wl_shm_buffer_end_access(shm_buf);
-
-	// We have uploaded the data, we don't need to access the wl_buffer
-	// anymore
-	wl_buffer_send_release(resource);
 
 	wl_list_remove(&buffer->resource_destroy.link);
 	wl_resource_add_destroy_listener(resource, &buffer->resource_destroy);
--- a/types/wlr_buffer.c
+++ b/types/wlr_buffer.c
@@ -104,32 +104,6 @@ bool wlr_resource_is_buffer(struct wl_resource *resource) {
 	return strcmp(wl_resource_get_class(resource), wl_buffer_interface.name) == 0;
 }
 
-bool wlr_resource_get_buffer_size(struct wl_resource *resource,
-		int *width, int *height) {
-	assert(wlr_resource_is_buffer(resource));
-
-	struct wl_shm_buffer *shm_buf = wl_shm_buffer_get(resource);
-	if (shm_buf != NULL) {
-		*width = wl_shm_buffer_get_width(shm_buf);
-		*height = wl_shm_buffer_get_height(shm_buf);
-	} else if (wlr_dmabuf_v1_resource_is_buffer(resource)) {
-		struct wlr_dmabuf_v1_buffer *dmabuf =
-			wlr_dmabuf_v1_buffer_from_buffer_resource(resource);
-		*width = dmabuf->attributes.width;
-		*height = dmabuf->attributes.height;
-	} else if (wlr_drm_buffer_is_resource(resource)) {
-		struct wlr_drm_buffer *drm_buffer =
-			wlr_drm_buffer_from_resource(resource);
-		*width = drm_buffer->base.width;
-		*height = drm_buffer->base.height;
-	} else {
-		*width = *height = 0;
-		return false;
-	}
-
-	return true;
-}
-
 static const struct wlr_buffer_impl client_buffer_impl;
 
 struct wlr_client_buffer *wlr_client_buffer_get(struct wlr_buffer *buffer) {
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -365,12 +365,13 @@ static void surface_apply_damage(struct wlr_surface *surface) {
 		}
 	}
 
-	struct wlr_client_buffer *buffer =
-		wlr_client_buffer_import(surface->renderer, resource);
+	struct wlr_client_buffer *buffer = wlr_client_buffer_create(
+			surface->current.buffer, surface->renderer, resource);
 	if (buffer == NULL) {
 		wlr_log(WLR_ERROR, "Failed to upload buffer");
 		return;
 	}
+	wlr_buffer_unlock(surface->current.buffer);
 
 	if (surface->buffer != NULL) {
 		wlr_buffer_unlock(&surface->buffer->base);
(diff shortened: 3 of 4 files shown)


types/wlr_buffer.c
```diff
@@ -254,7 +228,7 @@ struct wlr_client_buffer *wlr_client_buffer_import(
 	wlr_buffer_lock(&client_buffer->base);
 	wlr_buffer_drop(&client_buffer->base);
 
-	return client_buffer;
+	return &client_buffer->base;
```
emersion, line 231: Returning the client buffer here is not okay. In the case of a `wl_shm` buffer and the GLES2 renderer, the buffer will be uploaded each time `wlr_buffer_import` is called. We should do this a single time per commit. but this patch does it more often since this function is used to get the buffer size.

emersion: Note, buffer upload == buffer copy == slow :P

bl4ckb0ne: I see what I got wrong, made a few changes and the wlr_buffer is imported only once now.

types/wlr_buffer.c
```diff
@@ -226,7 +205,12 @@ struct wlr_client_buffer *wlr_client_buffer_import(
 		wl_resource_post_error(resource, 0, "unknown buffer type");
```
emersion, line 205: The `wlr_log` message above is outdated: this function doesn't upload textures anymore. We can probably just drop the "Cannot upload texture" part.

types/wlr_surface.c
```diff
@@ -169,10 +169,16 @@ static void surface_state_viewport_src_size(struct wlr_surface_state *state,
 
 static void surface_state_finalize(struct wlr_surface *surface,
 		struct wlr_surface_state *state) {
+	if (state->buffer_resource) {
+		state->buffer = wlr_buffer_import(surface->renderer,
+				state->buffer_resource);
+		wlr_buffer_lock(state->buffer);
```
emersion, line 175: `wlr_buffer_import` returns a buffer which is already locked. No need to lock it a second time.

types/wlr_surface.c
```diff
@@ -255,6 +261,7 @@ static void surface_state_copy(struct wlr_surface_state *state,
 		struct wlr_surface_state *next) {
 	state->width = next->width;
 	state->height = next->height;
+	state->buffer = next->buffer;
```
emersion, line 264: You might notice that `surface_state_copy` is a bit weird: the `buffer_resource` doesn't get copied.

The reason is we have `wlr_surface.previous` which holds the state for the _previous_ commit. We use `surface_state_copy` to construct `previous`. However we don't want to populate the buffer fields of the previous state, because that would lock the buffer for longer than necessary (and force the client to allocate and use one more buffer).

I don't like `wlr_surface.previous`, I think we should get rid of it, but that's another battle (a few more details about this in https://github.com/swaywm/wlroots/issues/2098). In the meantime, move this logic into `surface_state_move` (where you can notice `buffer_resource` is handled as well).

types/wlr_surface.c
```diff
@@ -365,12 +372,13 @@ static void surface_apply_damage(struct wlr_surface *surface) {
...
 
-	struct wlr_client_buffer *buffer =
-		wlr_client_buffer_import(surface->renderer, resource);
+	struct wlr_client_buffer *buffer = wlr_client_buffer_create(
+			surface->current.buffer, surface->renderer, resource);
 	if (buffer == NULL) {
 		wlr_log(WLR_ERROR, "Failed to upload buffer");
 		return;
 	}
+	wlr_buffer_unlock(surface->current.buffer);
```
emersion, line 381: We should reset `surface->current.buffer` to NULL here, to avoid having someone read a free'd pointer.

types/wlr_buffer.c
```diff
@@ -194,8 +168,13 @@ static void client_buffer_resource_handle_destroy(struct wl_listener *listener,
 	// which case we'll read garbage. We decide to accept this risk.
 }
 
-struct wlr_client_buffer *wlr_client_buffer_import(
-		struct wlr_renderer *renderer, struct wl_resource *resource) {
+struct wlr_buffer *wlr_buffer_import(struct wlr_renderer *renderer,
+		struct wl_resource *resource) {
+	if (!resource) {
```
emersion, line 173: I think we can just let `wlr_resource_is_buffer` segfault instead. It's a caller error to provide a NULL `wl_resource`.

Segfaults are sometimes better than gracefully handling errors, because they complain loudly and result in user bug reports. Error logs can go unnoticed.

emersion (changes requested):
Thanks, looks much better! Here are a few comments.

include/wlr/types/wlr_buffer.h
```diff
@@ -111,6 +112,14 @@ bool wlr_buffer_get_dmabuf(struct wlr_buffer *buffer,
  */
 bool wlr_buffer_get_shm(struct wlr_buffer *buffer,
 	struct wlr_shm_attributes *attribs);
+/**
+ * Creates a wlr_buffer from a wl_resource and lock it. Once the caller is done
+ * with the buffer, they must call wlr_buffer_unlock.
+ *
+ * The wl_resource must be a wl_buffer.
+ */
+struct wlr_buffer *wlr_buffer_import(struct wlr_renderer *renderer,
```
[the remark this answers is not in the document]
bl4ckb0ne: `wlr_buffer_from_resource` makes the most sense.

include/wlr/types/wlr_buffer.h
```diff
@@ -111,6 +112,14 @@ bool wlr_buffer_get_dmabuf(struct wlr_buffer *buffer,
  */
 bool wlr_buffer_get_shm(struct wlr_buffer *buffer,
 	struct wlr_shm_attributes *attribs);
+/**
+ * Creates a wlr_buffer from a wl_resource and lock it. Once the caller is done
```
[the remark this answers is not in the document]
bl4ckb0ne: Is "imports" more suitable?

types/wlr_buffer.c
bl4ckb0ne: I merged it with the other assert to show that we know this is supposed to segfault.

include/wlr/types/wlr_buffer.h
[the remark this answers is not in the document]
emersion: Hm, maybe "obtains"?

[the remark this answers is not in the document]
bl4ckb0ne: "Transforms"?

types/wlr_buffer.c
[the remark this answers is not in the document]
bl4ckb0ne: Inside `wlr_surface_apply_damage`? It kinda makes sense to leave it here for the custom buffers.

include/wlr/types/wlr_buffer.h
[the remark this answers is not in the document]
emersion: I'm not a fan of "import" because in the case of DMA-BUFs, this function doesn't import anything. I'm fine with "transform". Other alternatives in the same vein could be "convert" or "turn a wl_resource into a wlr_buffer".

types/wlr_buffer.c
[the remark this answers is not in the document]
emersion: Yup, inside `surface_apply_damage`.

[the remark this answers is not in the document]
emersion: > It kinda makes sense to leave it here for the custom buffers.

I don't understand, can you elaborate?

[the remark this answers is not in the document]
bl4ckb0ne: Ah I see, I got confused up. You're saying to put this in `wlr_surface` in case it returns NULL. Nevermind.

types/wlr_surface.c
bl4ckb0ne: I'm hitting an assert without it, tracked it down to this one yesterday.

> Assertion failed: buffer->n_locks > 0 (types/wlr_buffer.c: wlr_buffer_unlock: 58)

include/wlr/types/wlr_buffer.h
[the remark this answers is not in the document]
bl4ckb0ne: I went with transform.

types/wlr_buffer.c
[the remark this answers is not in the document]
bl4ckb0ne: I ended up shoving it in `surface_state_finalize` since it's the only place that calls `wlr_buffer_from_resource`.

types/wlr_surface.c
emersion: In `surface_state_copy` the previous version was copying the pointer to the `wlr_buffer` without locking (`next->buffer = state->buffer` should have been `next->buffer = wlr_buffer_lock(state->buffer)`). I didn't mention it because I suggested to remove that code.

tl;dr I think it should work better with the new version of the PR. Can you try again removing it?

bl4ckb0ne: Yup, it works

types/wlr_buffer.c
```diff
@@ -219,14 +193,15 @@ struct wlr_client_buffer *wlr_client_buffer_import(
...
 		return NULL;
 	}
+	
+	return buffer;
+}
 
+struct wlr_client_buffer *wlr_client_buffer_create(struct wlr_buffer *buffer,
+		struct wlr_renderer *renderer, struct wl_resource *resource) {
 	struct wlr_texture *texture = wlr_texture_from_buffer(renderer, buffer);
 	wlr_buffer_unlock(buffer);
```
emersion, line 206: This `wlr_buffer_unlock` call makes `wlr_client_buffer_create` a bit weird: it decrements ref'count of the `wlr_buffer` passed in. This is surprising for callers, I think it'd be better to let the callers do it if they need to.

types/wlr_surface.c
```diff
@@ -303,8 +313,12 @@ static void surface_state_move(struct wlr_surface_state *state,
 	surface_state_copy(state, next);
 
 	if (next->committed & WLR_SURFACE_STATE_BUFFER) {
+		if (next->buffer) {
+			state->buffer = wlr_buffer_lock(next->buffer);
+		}
 		surface_state_set_buffer(state, next->buffer_resource);
 		surface_state_reset_buffer(next);
+		next->buffer = NULL;
```
emersion, line 321: Here we unset `next->buffer` without decrementing its ref'count.

types/wlr_buffer.c
bl4ckb0ne: Removed it

types/wlr_surface.c
bl4ckb0ne: Feels a bit weird to lock it when copying it to then unlock it a few lines beneath, but I get why.

emersion: If you lock it then unlock if immediately, it's possible to simplify to just "transfer" the ref'count.

e.g. it's perfectly fine to change

```c
b = wlr_buffer_lock(a);
wlr_buffer_unlock(a);
a = NULL;
```

to

```c
b = a;
a = NULL;
```

emersion:
I think we should also call `wlr_buffer_unlock` in `surface_state_finish`, to make sure we don't leave buffer refs behind when the surface is destroyed.

types/wlr_surface.c
```diff
@@ -169,10 +169,20 @@ static void surface_state_viewport_src_size(struct wlr_surface_state *state,
 
 static void surface_state_finalize(struct wlr_surface *surface,
 		struct wlr_surface_state *state) {
+	if (state->buffer_resource) {
```
emersion, line 172: I think we could move this to the `if ((state->committed & WLR_SURFACE_STATE_BUFFER))` below. No need to update `state->buffer` if it's still the same as before.

bl4ckb0ne: Done. I think I put it there when I was debugging and forgot to move it after.

types/wlr_surface.c
```diff
@@ -170,9 +170,17 @@ static void surface_state_viewport_src_size(struct wlr_surface_state *state,
 static void surface_state_finalize(struct wlr_surface *surface,
 		struct wlr_surface_state *state) {
 	if ((state->committed & WLR_SURFACE_STATE_BUFFER)) {
-		if (state->buffer_resource != NULL) {
-			wlr_resource_get_buffer_size(state->buffer_resource,
-				&state->buffer_width, &state->buffer_height);
+		if (state->buffer_resource) {
+			state->buffer = wlr_buffer_from_resource(surface->renderer,
```
emersion, line 174: When we overwrite a `state->buffer` field, we need to make sure it's NULL. If it's not, it means it's holding a `wlr_buffer` reference, so overwriting it will leak a reference.

types/wlr_surface.c
```diff
@@ -303,9 +311,14 @@ static void surface_state_move(struct wlr_surface_state *state,
 	surface_state_copy(state, next);
 
 	if (next->committed & WLR_SURFACE_STATE_BUFFER) {
+		if (next->buffer) {
+			state->buffer = wlr_buffer_lock(next->buffer);
```
emersion, line 315: Same here, if `state->buffer` is not NULL, we must make sure to drop the ref.

emersion:
Addressing the last two comments should make some `wl_shm` clients a bit confused, e.g. GTK clients will die in an assertion failure. This is due to the `wl_buffer_send_release` call in `wlr_client_buffer_apply_damage`. We now sometimes call `wlr_buffer_from_resource` followed by `wlr_client_buffer_apply_damage` followed by `wlr_buffer_unlock`. The last two calls send a `wl_buffer.release` event to the client.

To fix it, we can just remove the `wlr_buffer_from_resource` call from `wlr_client_buffer_apply_damage`, since we now take care of it ourselves.

types/wlr_surface.c
```diff
@@ -365,12 +378,14 @@ static void surface_apply_damage(struct wlr_surface *surface) {
...
-	struct wlr_client_buffer *buffer =
-		wlr_client_buffer_import(surface->renderer, resource);
+	struct wlr_client_buffer *buffer = wlr_client_buffer_create(
+			surface->current.buffer, surface->renderer, resource);
 	if (buffer == NULL) {
 		wlr_log(WLR_ERROR, "Failed to upload buffer");
 		return;
 	}
+	wlr_buffer_unlock(surface->current.buffer);
+	surface->current.buffer = NULL;
```
emersion, line 388: Here we take care of unlocking the buffer. But we don't do it in all branches of this function. We should also unlock the buffer above when `return` is used.

types/wlr_surface.c
```diff
@@ -170,9 +170,21 @@ static void surface_state_viewport_src_size(struct wlr_surface_state *state,
 static void surface_state_finalize(struct wlr_surface *surface,
 		struct wlr_surface_state *state) {
 	if ((state->committed & WLR_SURFACE_STATE_BUFFER)) {
-		if (state->buffer_resource != NULL) {
-			wlr_resource_get_buffer_size(state->buffer_resource,
-				&state->buffer_width, &state->buffer_height);
+		if (state->buffer_resource) {
+			if (state->buffer) {
+				wlr_buffer_unlock(state->buffer);
```
emersion, line 175: Nit: it's safe to pass NULL to `wlr_buffer_unlock`

emersion (approved):
LGTM, thanks!

PR #3024, opened 2021-07-06
surface: don't destroy surface while compositor has locks

emersion:
References: https://github.com/swaywm/wlroots/issues/2957

* * *

Breaking change: `wlr_surface.resource` may now be NULL if the client has destroyed it but the compositor still has locks.

TODO: check all `surface->resource` usage throughout wlroots, make sure we won't hit a NULL dereference.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3024

Issue #3020, opened 2021-07-04
Rapid window updates result in incorrect frame order in nested Sway sessions


vyivel:
- Sway version: 1.6.1
- Configuration: none/irrelevant (tested with `sway -c /dev/null`)

### Description
Sometimes Sway renders frames out of order when run with a Wayland/X11 backend, resulting in older frames being seen after newer ones. At least, that's what it looks like.

### Reproduction
1) Start Sway
2) Start `weston-terminal` (or any other program capable of updating its window contents rapidly)
3) Hold a key
4) Observe the incorrect rendering order, something like this:
Frame 1: `$ aaaaaaaaaa`
Frame 2: `$ aaaaaa` 
Frame 3: `$ aaaaaaaaaaaaaaaa`

This isn't reproducible with Sway 1.6 or when the DRM backend is used.

### Notes
- This might be a wlroots issue; https://github.com/swaywm/wlroots/commit/543f5b35d02c825347e4a2d6f032ddfa938c3fd7 is the commit that causes this behavior (under Wayland), compiling Sway with a commit before that fixes the problem.
- `-D damage=highlight` has no effect for nested Sway sessions, which may or may not be related to this. Also introduced with the wlroots commit mentioned above.
- It's easier to reproduce this under a high CPU load (can be simulated with `cpulimit -l 1 sway -c /dev/null`).

emersion:
Just to understand the bug buffer, does this wlroots patch help? https://l.sr.ht/qsvv.txt

vyivel:
Nope, the issue persists.

emersion:
Okay, new attempt: https://l.sr.ht/BSq5.txt

If this doesn't work, can you try `-Ddamage=rerender` and see if it helps? The flag disables damage tracking and forces the whole output to be re-rendered each frame.

vyivel:
This patch fixes `-Ddamage=highlight` (seems like nested sessions using `WLR_OUTPUT_STATE_BUFFER_SCANOUT` damage the whole output each frame anyway), but doesn't fix the flickering issue. `-Ddamage=rerender` has no effect.

vyivel:
`wlr_output_damage` doesn't seem to be relevant at all, as the issue is reproducible with tinywl.

emersion:
Thanks for confirming this isn't a damage tracking issue.

Here' yet another attempt: https://l.sr.ht/xcCP.txt. Can you try it? (Sorry, I'm having trouble reproducing reliably.)

PR #3027, opened 2021-07-07
output: detach buffer from renderer before commit

emersion:
Right now we rely entirely on implicit sync for synchronizing
access to GPU buffers. Implicit sync works by setting
synchronization points on the buffer in writers, and letting
readers wait on these sync points before accessing the buffer.

With OpenGL, sync points are created using functions such as
eglSwapBuffers or glFlush. If none of these special functions
are called, no sync point will be created and readers will
potentially access a buffer that hasn't finished rendering yet.

In the context of wlroots, OpenGL is the writer and the backend
(KMS or parent Wayland/X11 session) is the reader. After we're
done rendering a frame, and before passing that frame to the
backend, we need to call glFlush.

glFlush is called when the buffer is detached from the renderer.
This is a task done by output_clear_back_buffer. So let's call
this function before invoking the impl->commit hook, instead of
calling it after.

All of this is maybe a little tricky to get right with the
current renderer_bind_buffer API. The new
wlr_renderer_begin_with_buffer API is much better, because glFlush
is called on wlr_renderer_end, so it's more intuitive.

Closes: https://github.com/swaywm/wlroots/issues/3020

### DIFF
--- a/types/wlr_output.c
+++ b/types/wlr_output.c
@@ -736,8 +736,19 @@ bool wlr_output_commit(struct wlr_output *output) {
 	};
 	wlr_signal_emit_safe(&output->events.precommit, &pre_event);
 
-	if (!output->impl->commit(output)) {
+	// output_clear_back_buffer detaches the buffer from the renderer. This is
+	// important to do before calling impl->commit(), because this marks an
+	// implicit rendering synchronization point. The backend needs it to avoid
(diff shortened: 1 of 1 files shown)


vyivel:
Thanks for your work!

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM thanks!

PR #3028, opened 2021-07-07
output-damage: fix output swapchain handling

emersion:
When wlr_output.swapchain is used instead of the backend's, the
buffer_type will be set to SCANOUT even if wlr_output_attach_render
has been called. This tricks wlr_output_damage into thinking the
whole output needs to be repainted.

Workaround this issue by forcing buffer_type to RENDER when the
output has a back-buffer set.

Will clean all of that up when removing the precommit event handler
altogether.

This commit fixes damage tracking on the Wayland, X11 and headless
backends.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks :)

Issue #3030, opened 2021-06-27
Disappearing cursor on multi-GPU setup


IntinteDAO:
- **Sway Version:**
  - sway version 1.6-5cdce42d (Jun 26 2021, branch 'master')

- **Debug Log:**
https://0bin.net/paste/up-swxj-#PJPVTysHFaMxP3QLzyxcJ-EN4sFoSob1FtZtbqIm2FW

- **Configuration File:**
https://0bin.net/paste/eUn83bT7#nkNcFBESXx5ruLKva2F+zyzxJuhy5rw6Bu22csinUPT

- **Description:**
The error is probably due to disabling one monitor (using wdisplays). This causes problems with the cursor not displaying in Sway. Various actions turn it off and on. For example hover the mouse over wofi makes the cursor disappear and running wdisplays makes it come back.

The mouse is on button (it changed a color, but invisible)
![20210627_22h03m11s_grim](https://user-images.githubusercontent.com/56807194/123559385-0e512100-d79c-11eb-8b00-96db2d5efee9.png)

Here the mouse is near the menu, but also invisible)
![20210627_22h03m20s_grim](https://user-images.githubusercontent.com/56807194/123559387-0f824e00-d79c-11eb-9913-e637f3358059.png)

emersion:
Does this wlroots patch help by any chance? https://l.sr.ht/-ZAc.txt

IntinteDAO:
For me - works great! But it's nice to check @nwg-piotr also

nwg-piotr:
> check @nwg-piotr also

I'll check as soon as I'm back  home.

nwg-piotr:
Works for me, too. Great news!

emersion:
Thanks for testing!

nwg-piotr:
Just for curiosity: average CPU load measured on idle in the same conditions. Ryzen 7 + hybrid AMD graphics.

| version  | CPU load | MEM usage |
| ------------- | ------------- | ------------- |
| 0.12.0  | 1.88%  | 490 MiB |
| 0.13.0  | 3.50%  | 435 MiB |
| 0.14.0  | 3.15%  | 433 MiB |
| 0.14.1  | 2.82%  | 435 MiB |

PR #3033, opened 2021-07-08
backend/drm: force linear layout for multi-GPU buffers

emersion:
Some buffers need to be copied across GPUs. Such buffers need to be
allocated with a format and modifier suitable for both the source
and the destination.

When explicit modifiers aren't supported, we were forcing the buffers
to be allocated with a linear layout, because implicit modifiers
aren't portable across GPUs. All is well with this case.

When explicit modifiers are supported, we were advertising the whole
list of destination modifiers, in the hope that the source might
have some in common and might be able to allocate a buffer with a
more optimized layout. This works well if the source supports explicit
modifiers. However, if the source doesn't, then wlr_drm_format_intersect
will fallback to implicit modifiers, and everything goes boom: the
source uses a GPU-specific tiling and the destination interprets it
as linear.

To avoid this, just force linear unconditionally. We'll be able to
revert this once we have a good way to indicate that an implicit modifier
isn't supported in wlr_drm_format_set, see [1].

[1]: https://github.com/swaywm/wlroots/pull/2815

Closes: https://github.com/swaywm/wlroots/issues/3030

### DIFF
--- a/backend/drm/backend.c
+++ b/backend/drm/backend.c
@@ -248,21 +248,15 @@ struct wlr_backend *wlr_drm_backend_create(struct wl_display *display,
 			goto error_event;
 		}
 
+		// Force a linear layout. In case explicit modifiers aren't supported,
+		// the meaning of implicit modifiers changes from one GPU to the other.
+		// In case explicit modifiers are supported, we still have no guarantee
+		// that the buffer producer will support these, so they might fallback
+		// to implicit modifiers.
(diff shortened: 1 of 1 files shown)


emersion:
Note: wlroots 0.13.0 forces linear for all multi-gpu buffers as well (https://github.com/swaywm/wlroots/blob/69c71dbc8afecc5da5c800cdc1475185064b4ac4/backend/drm/renderer.c#L264).

Issue #2907, opened 2021-05-01
render/pixman: doesn't clear transparent regions


emersion:
1. Start Sway with the Pixman renderer
2. Start slurp
3. Start selecting a region
4. The selected region is full of previous rectangles. The area which isn't selected is supposed to be semi-transparent, but it's completely white.

![screenshot](https://user-images.githubusercontent.com/506932/116790554-04b97e80-aab5-11eb-9d94-670b4a318953.png)

cc @bl4ckb0ne

bl4ckb0ne:
I managed to reproduce on my side, gonna investigate it. Could you tell me more about slurp's rendering process please? I guess it's a layer shell, but how do you handle blanking the screen and the drawing of the rectange?

bl4ckb0ne:
I think it's because the renderer doesn't implement the texture `write_pixels` function.

emersion:
Unlikely. Re-using textures ils an optional optimization irrelevant for the Pixman renderer since there's no GPU upload going on.

emersion:
slurp just clears the whole buffer and renders the whole thing each frame. I think the Pixman renderer doesn't handle a non-opaque alpha channel properly.

PR #3036, opened 2021-07-09
render/pixman: fix texture_is_opaque()

vyivel:
A texture is opaque when it does *not* have alpha.

Fixes https://github.com/swaywm/wlroots/issues/2907


emersion (approved):
Very good catch! LGTM and works.

Thanks!

PR #3037, opened 2021-07-09
render/egl: add support for EGL_EXT_device_drm_render_node

emersion:
This EGL extension has been added in [1]. The upsides are:

- We directly get a render node, instead of having to convert the
  primary node name to a render node name.
- If EGL_DRM_RENDER_NODE_FILE_EXT returns NULL, that means there is
  no render node being used by the driver.

[1]: https://github.com/KhronosGroup/EGL-Registry/pull/127

Mesa patch: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/11797
libglvnd patch: https://gitlab.freedesktop.org/glvnd/libglvnd/-/merge_requests/245

### DIFF
--- a/render/egl.c
+++ b/render/egl.c
@@ -289,6 +289,8 @@ struct wlr_egl *wlr_egl_create_with_drm_fd(int drm_fd) {
 
 		egl->exts.EXT_device_drm =
 			check_egl_ext(device_exts_str, "EGL_EXT_device_drm");
+		egl->exts.EXT_device_drm_render_node =
+			check_egl_ext(device_exts_str, "EGL_EXT_device_drm_render_node");
 	}
 
 	if (!check_egl_ext(display_exts_str, "EGL_KHR_no_config_context") &&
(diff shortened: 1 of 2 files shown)


emersion:
The Mesa and libglvnd patches have been merged.

render/egl.c
```diff
@@ -712,23 +714,39 @@ static char *get_render_name(const char *name) {
...
 	}
 
-	const char *primary_name = egl->procs.eglQueryDeviceStringEXT(egl->device,
-		EGL_DRM_DEVICE_FILE_EXT);
-	if (primary_name == NULL) {
-		wlr_log(WLR_ERROR,
-			"eglQueryDeviceStringEXT(EGL_DRM_DEVICE_FILE_EXT) failed");
-		return -1;
+	char *render_name = NULL;
+#ifdef EGL_EXT_device_drm_render_node
```
bl4ckb0ne, line 723: Why the preprocessor? Isn't checking `egl->exts.EXT_device_drm_render_node` enough?

emersion: `EGL_EXT_device_drm_render_node` has been merged recently, so won't be widely available in the system headers for a while. Need to wait for new libglvnd and Mesa releases.

bl4ckb0ne: So we're waiting for a release before merging?

emersion: No need, the `#ifdef` ensures this code is disabled for older Mesa/libglvnd, and is enabled for recent ones.

emersion:
We may want to merge https://github.com/swaywm/wlroots/pull/2671 first.

render/egl.c
```diff
@@ -712,23 +714,39 @@ static char *get_render_name(const char *name) {
 }
 
 int wlr_egl_dup_drm_fd(struct wlr_egl *egl) {
-	if (egl->device == EGL_NO_DEVICE_EXT || !egl->exts.EXT_device_drm) {
+	if (egl->device == EGL_NO_DEVICE_EXT || (!egl->exts.EXT_device_drm &&
+			!egl->exts.EXT_device_drm_render_node)) {
```
bl4ckb0ne, line 718: Wont this check fail if the user doesnt have the ext?

emersion: If the user doesn't have `EXT_device_drm_render_node`, but has `EXT_device_drm`, then the `if` is not taken. The `if` is only taken if the user is missing both exts.

bl4ckb0ne: I should learn to read throughly the conditions for next time, sorry

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, sorry for the dumb comment

render/egl.c
emersion: No worries!

PR #3041, opened 2021-07-11
render/gles2: disable blending opportunistically

emersion:
We don't always need to enable blending: when the texture doesn't
have alpha or when the color is opaque, we can disable it.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #3042, opened 2021-07-11
render/{egl,gles2}: rename wlr_egl.exts to better match Khronos

emersion:
Khronos refers to extensions with their namespace as a prefix in
uppercase. Change our naming to align with Khronos conventions.
This also makes grepping easier.


bl4ckb0ne (approved):

bl4ckb0ne:
Very good idea! Thanks :)

PR #3046, opened 2021-07-12
Custom wlr_buffer from wl_resource

bl4ckb0ne:
Custom `wlr_buffer` from `wl_resource` take two

Supersedes https://github.com/swaywm/wlroots/pull/2934


include/wlr/types/wlr_buffer.h
```diff
@@ -24,6 +24,8 @@ struct wlr_shm_attributes {
 };
 
 struct wlr_buffer_impl {
+	bool (*is_instance)(struct wl_resource *resource);
+	struct wlr_buffer *(*create)(struct wl_resource *resource);
```
emersion, line 28: I'd prefer to introduce a separate struct for these, e.g. `wlr_wl_buffer_impl`. `wlr_buffer` itself is completely separate from the Wayland protocol and `wl_resource`.

emersion: I'd prefer to name this `buffer_from_resource` rather than `create`, for consistency with `wlr_buffer_from_resource`.

bl4ckb0ne: I was hesitating between the two and chose `create` to mimic [`shm_client_buffer_create`](https://github.com/swaywm/wlroots/blob/master/types/wlr_buffer.c#L167), will make the changes.

types/wlr_buffer.c
```diff
@@ -11,6 +11,33 @@
 #include "types/wlr_buffer.h"
 #include "util/signal.h"
 
+struct buffer_impl_registration {
+	const struct wlr_buffer_impl *impl;
+	struct wl_list link;
+};
+
+static struct wl_list impls = {0};
```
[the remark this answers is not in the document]
bl4ckb0ne: `wl_array` is very much what I needed here, thanks!

include/wlr/types/wlr_buffer.h
```diff
@@ -69,6 +69,11 @@ struct wlr_buffer {
 	} events;
 };
 
+struct wlr_buffer_resource_impl {
```
bl4ckb0ne, line 72: `wlr_buffer_resource_impl` or `wlr_buffer_resource_interface` ?

emersion: I'm fine with either.

bl4ckb0ne: I'll go with `interface` then.

types/wlr_buffer.c
```diff
@@ -157,6 +157,27 @@ static bool buffer_is_shm_client_buffer(struct wlr_buffer *buffer);
 static struct wlr_shm_client_buffer *shm_client_buffer_from_buffer(
 	struct wlr_buffer *buffer);
 
+static struct wl_array impls = {0}; // struct wlr_buffer_resource_impls
```
emersion, line 159: Maybe rename to `resource_impls`?

types/wlr_buffer.c
```diff
@@ -178,6 +178,18 @@ void wlr_buffer_register_resource_impl(
 	*impl_ptr = impl;
 }
 
+static struct wlr_buffer_resource_impl *get_buffer_resource_impl(
```
emersion, line 180: This function can return a `const` pointer

types/wlr_buffer.c
```diff
@@ -178,6 +178,18 @@ void wlr_buffer_register_resource_impl(
...
+		struct wl_resource *resource) {
+	struct wlr_buffer_resource_impl **impl_ptr;
```
emersion, line 182: Same here, can be `const`

emersion:
BTW, any reason why this is still a draft?

bl4ckb0ne:
No reasons, I'll put it as a standard patch.

emersion:
I've tried using this work to turn `wl_shm` into a buffer resource impl: https://github.com/swaywm/wlroots/pull/3061

Seems to work fine!

emersion:
I'm not a huge fan of introducing global state for this, OTOH avoiding global state would require more work. Maybe something like registering the resource impls in `wlr_compositor` would do the trick?

bl4ckb0ne:
what about `wlr_renderer` ?

emersion:
I'm suggesting `wlr_compositor`, since this is the parent object for surfaces, where the `wl_buffer` resources are turned into `wlr_buffer`. However I now realize turning a `wl_buffer` resource into a `wlr_buffer` happens elsewhere too, e.g. for screencopy.

emersion:
I'd rather not add this to the renderer: the renderer shouldn't really care about Wayland protocol related things.

include/wlr/types/wlr_buffer.h
```diff
@@ -69,6 +69,11 @@ struct wlr_buffer {
...
+	bool (*is_instance)(struct wl_resource *resource);
```
emersion, line 73: Maybe we could add a `const char *name` field here as well? That would make debugging easier, e.g. we could print the kind of the buffer resource in the error messages in the next commit, instead of just referring to a "custom buffer".

bl4ckb0ne: Sure!

bl4ckb0ne:
I've checked the code a bit and I think adding a parent object would add a lot of clutter to the current interface.

Mostly because the only place we create a custon `wlr_buffer` is `wlr_buffer_from_resource`, which we already cleaned to have just a `wl_resource` argument. Plus I can't find where to take the reference to the `wlr_compositor`.

emersion:
Yeah, it's probably fine for now. Something to think about before stabilization I guess.

emersion (approved):
This LGTM, thanks!

PR #3048, opened 2021-07-13
backend/wayland: properly cleanup wlr_wl_pointer

emersion:
We were missing destroy calls for gestures, and we were only
destroying the relative pointer on output destroy.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

Issue #2973, opened 2021-06-16
wl_pointer's axis_source event is sent multiple times per frame


Biolunar:
According to the Wayland spec, only one axis_source event is permitted per pointer frame. When scrolling on my laptops touchpad diagonally, two identical axis_source events are emitted.

wev output:
```
[…]
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 1 (finger)
[13:      wl_pointer] axis: time: 228155853; axis: 0 (vertical), value: -3.136719
[13:      wl_pointer] axis_source: 1 (finger)
[13:      wl_pointer] axis: time: 228155853; axis: 1 (horizontal), value: 3.089844
[13:      wl_pointer] frame
[…]
```

emersion:
We should fix this in the `wlr_seat` implementation:

- When sending a `wl_pointer.axis` event, cache the source.
- If there's already a cached source, don't send it again. If the event source and the cached source mismatch, error out.
- Clear the cached source when sending `wl_pointer.frame`.

PR #3050, opened 2021-07-17
input/pointer: send axis source once per frame

vyivel:
Only one wl_pointer.axis_source event is permitted per frame, according
to the Wayland specification.

Fixes https://github.com/swaywm/wlroots/issues/2973

### DIFF
--- a/types/seat/wlr_seat_pointer.c
+++ b/types/seat/wlr_seat_pointer.c
@@ -274,6 +274,15 @@ void wlr_seat_pointer_send_axis(struct wlr_seat *wlr_seat, uint32_t time,
 		return;
 	}
 
+	bool send_source = false;
+	if (wlr_seat->pointer_state.sent_axis_source) {
+		assert(wlr_seat->pointer_state.cached_axis_source == source);
+	} else {
+		wlr_seat->pointer_state.sent_axis_source = true;
+		wlr_seat->pointer_state.cached_axis_source = source;
+		send_source = true;
+	}
+
 	struct wl_resource *resource;
 	wl_resource_for_each(resource, &client->pointers) {
 		if (wlr_seat_client_from_pointer_resource(resource) == NULL) {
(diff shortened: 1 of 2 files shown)


vyivel:
@Biolunar, could you test this branch?

types/seat/wlr_seat_pointer.c
```diff
@@ -274,6 +274,18 @@ void wlr_seat_pointer_send_axis(struct wlr_seat *wlr_seat, uint32_t time,
 		return;
 	}
 
+	bool send_source = false;
+	if (wlr_seat->pointer_state.sent_axis_source) {
+		if (wlr_seat->pointer_state.cached_axis_source != source) {
+			wlr_log(WLR_DEBUG,
```
emersion, line 280: Maybe we should bump this to WLR_ERROR?

emersion (commented):
LGTM otherwise!

types/seat/wlr_seat_pointer.c
vyivel: Wouldn't that flood logs too much, writing an error message every pointer frame?

emersion: Yeah, but this indicates a compositor bug. I'm afraid of not noticing it when it happens.

vyivel: Hmm, I guess a similar check should be added to Sway then.

emersion: Hm, should it? The events come from the backends, I don't think there's a situation in which Sway arbitrarily changes the axis source or injects its own?

vyivel: You're right, there's no such situation, Sway just calls `wlr_seat_pointer_notify_axis()` which in turn calls `wlr_seat_pointer_send_axis()`. But it's a backend bug then, not compositor's.

vyivel: Maybe it should be changed to `assert()`?

emersion: Yeah, I'm wondering about this as well. In general I'm in favor of failing hard when we hit inconsistent state, because that is guaranteed to result in a bug report. We can always downgrade to a log message if time tells this was a bad idea.

vyivel:
Changed `wlr_log()` to `assert()`.

emersion (approved):
LGTM, thanks!

Biolunar:
@vyivel I tested it and it fixes the issue for me. Thank you for the patch!

PR #3053, opened 2021-07-19
backend/drm: fix NULL data in handle_drm_event

emersion:
wl_event_loop_add_fd was called with a NULL data argument, but the
function expects the data argument to be set to the wlr_drm_backend.

Fixes: 053ebe7c278b ("backend/drm: terminate display on drmHandleEvent failure")


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks :)

PR #3056, opened 2021-07-20
seat: fix keyboard/pointer grab end

vyivel:
- Call a grab's cancel() when another grab is started
- Clear focus on default grab cancel

This PR fixes the following issue (reproducible with Wayfire, not reproducible with Sway as it updates focus on every click):
1) Open a window with something to drag (tested with `weston-dnd`)
2) Create a layer surface with exclusive keyboard interactivity, so the focus won't switch on step 4 (tested with `bemenu` and `gtk-layer-demo`)
3) Press a key, typing it repeatedly to the layer surface
4) Holding the key, start a grab
6) Release the key
7) Observe that the layer surface didn't receive a key release event

Not sure if this is the right way to fix this though.


vyivel:
This causes infinite recursion in `xdg_popup_grab_end()`.

vyivel:
Closing as I'm not sure anymore this is the correct approach.

PR #3058, opened 2021-07-21
backend: make DRM and libinput backends optional

emersion:
Supersedes: https://github.com/swaywm/wlroots/pull/2635
Supersedes: https://github.com/swaywm/wlroots/pull/2809

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -6,10 +6,9 @@
 #include <string.h>
 #include <unistd.h>
 #include <wayland-server-core.h>
-#include <wlr/backend/drm.h>
+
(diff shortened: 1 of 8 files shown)


bl4ckb0ne:
Can the `noop` and `headless` backends be optional too?

emersion:
Yes, they could, but they won't allow us to make more deps optional (kind of similar to the pixman renderer). So I haven't bothered making these optional.

Same goes for the Wayland backend.

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks

PR #3062, opened 2021-07-22
render: drop wlr_renderer_impl.init_wl_display

emersion:
Now that we have our own wl_drm implementation, there's no reason
to provide custom renderer hooks to init a wl_display in the
interface. We can just initialize the wl_display generically,
depending on the renderer capabilities.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks :)

PR #3065, opened 2021-07-23
render/egl: set EGL_IMAGE_PRESERVED_KHR

emersion:
Without setting this the EGL implementation is allowed to perform
destructive actions on the buffer when imported: its contents
become undefined.

This is mostly a pedantic change, because Mesa processes the attrib
and does absolutely nothing with it.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM!

PR #3068, opened 2021-07-26
xdg-shell: add state caching support

vyivel:
>The wlr_surface API for cached state is not very ergonomic. Ideas welcome to improve it.

This is another approach to what https://github.com/swaywm/wlroots/pull/2977 tries to accomplish. Whether it's more ergonomic or not is up for debate.

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -146,6 +146,7 @@ struct wlr_surface {
 	struct {
 		struct wl_signal commit;
 		struct wl_signal cache;
+		struct wl_signal precommit;
 		struct wl_signal new_subsurface;
 		struct wl_signal destroy;
 	} events;
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -29,6 +29,13 @@ enum wlr_surface_state_field {
 	WLR_SURFACE_STATE_VIEWPORT = 1 << 8,
 };
 
+struct wlr_surface_state_addon {
+	void *state;
+	void *owner;
+	struct wl_list link;
+	void (*finish_state)(void *state);
+};
+
 struct wlr_surface_state {
 	uint32_t committed; // enum wlr_surface_state_field
 	// Sequence number of the surface state. Incremented on each commit, may
@@ -67,6 +74,8 @@ struct wlr_surface_state {
 	// Number of locks that prevent this surface state from being committed.
 	size_t cached_state_locks;
 	struct wl_list cached_state_link; // wlr_surface.cached
+
+	struct wl_list addons;
 };
 
 struct wlr_surface_role {
@@ -136,6 +145,7 @@ struct wlr_surface {
 
 	struct {
 		struct wl_signal commit;
+		struct wl_signal cache;
 		struct wl_signal new_subsurface;
 		struct wl_signal destroy;
 	} events;
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -330,6 +330,9 @@ static void surface_state_move(struct wlr_surface_state *state,
 
 	next->committed = 0;
 	next->cached_state_locks = 0;
+
+	wl_list_insert_list(&state->addons, &next->addons);
+	wl_list_init(&next->addons);
 }
 
 static void surface_damage_subsurfaces(struct wlr_subsurface *subsurface) {
@@ -432,6 +435,8 @@ static void surface_cache_pending(struct wlr_surface *surface) {
 	wl_list_insert(surface->cached.prev, &cached->cached_state_link);
 
 	surface->pending.seq++;
+
+	wlr_signal_emit_safe(&surface->events.cache, cached);
 }
 
 static void surface_commit_state(struct wlr_surface *surface,
@@ -656,6 +661,8 @@ static void surface_state_init(struct wlr_surface_state *state) {
 	pixman_region32_init(&state->opaque);
 	pixman_region32_init_rect(&state->input,
 		INT32_MIN, INT32_MIN, UINT32_MAX, UINT32_MAX);
+
+	wl_list_init(&state->addons);
 }
 
 static void surface_state_finish(struct wlr_surface_state *state) {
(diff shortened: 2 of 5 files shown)


vyivel:
wlr_surface.next must be set to NULL after being committed.

emersion:
Thanks for working on this!

One thing to note is that roles aren't the only "surface add-ons" which add more state to the surface. Other extensions such as viewporter, xdg-decoration, upcoming color management protocols also add more "surface state" tied to `wl_surface.commit`. Also, wlroots can't possibly know about all possible surface add-ons -- the compositor might have their custom protocols too.

I'd prefer to come up with a solution which works for the non-role surface add-on case as well, if possible.

vyivel:
Do we also need to preserve addon states in `wlr_surface.previous`?

emersion:
I want `wlr_surface.previous` to be gone tbh. The only reason it's still here is because it's used internally in wlr_surface.

vyivel:
Gotcha. I'll see what can be done once I get home.

vyivel:
Went with `wlr_surface.addons` list and `precommit` event. I think keeping `wlr_surface.previous` around isn't that bad, but maybe it should be labeled as an internal field like `pending`.

vyivel:
I think this is ready for a review. Using `precommit` event should be flexible enough to create addon states for any extension.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -347,19 +345,48 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
...
+static void xdg_surface_handle_surface_precommit(struct wl_listener *listener,
+		void *data) {
+	struct wlr_xdg_surface *surface =
+		wl_container_of(listener, surface, surface_precommit);
+	struct wlr_surface_state *pending = data;
+
+	struct wlr_surface_state_addon *addon = calloc(1, sizeof(*addon));
+	struct wlr_xdg_surface_state *state = calloc(1, sizeof(*state));
+	if (!addon || !state) {
+		free(addon);
```
bl4ckb0ne, line 357: Possibility of double free here.

include/wlr/types/wlr_surface.h
```diff
@@ -29,6 +29,12 @@ enum wlr_surface_state_field {
 	WLR_SURFACE_STATE_VIEWPORT = 1 << 8,
 };
 
+struct wlr_surface_state_addon {
+	void *state;
```
bl4ckb0ne, line 33: Why not `wlr_xdg_surface_state` ?

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -347,19 +345,48 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
...
+		free(state);
+		return;
+	}
+	memcpy(state, &surface->pending, sizeof(*state));
```
bl4ckb0ne, line 361: Why not make it a pointer?

include/wlr/types/wlr_surface.h
vyivel: The surface can have multiple extensions to it, not just `wlr_xdg_surface`.

types/xdg_shell/wlr_xdg_surface.c
vyivel: How so? Either an allocated pointer or NULL will be passed to `free()` in both cases.

types/xdg_shell/wlr_xdg_surface.c
vyivel: Sorry, not sure what do you mean. `wlr_xdg_surface.pending` is used to store, well, the pending state, just like `wlr_surface.pending`. It has no reason to be a pointer.

types/wlr_surface.c
```diff
@@ -437,6 +437,8 @@ static void surface_cache_pending(struct wlr_surface *surface) {
 	wl_list_insert(surface->cached.prev, &cached->cached_state_link);
 
 	surface->pending.seq++;
+
+	surface->last_committed = cached;
```
ammen99, line 444: Is this not problematic if a client commits two times and we have two cached events? Maybe the frame callback was in the first commit. This way we're losing the callbacks because the next commit is unrelated. IDK if real-world clients do this, but it is a possibility.

vyivel: Hmm, but we ony render the last state we got, so it would make sense to only call last commit callbacks, but that's just my opinion.

What does Weston do in this case?

ammen99: Weston seems to be accumulating all frames in a single list when committing. https://gitlab.freedesktop.org/wayland/weston/-/blob/main/libweston/compositor.c#L3859

Also, the wayland protocol indicates that this is the correct thing to do:

```
For a wl_surface, the notifications are posted in the order the frame requests were committed.
```

So to my understanding, we should accumulate.

vyivel: Okay, accumulate we will. Thanks for checking.

ammen99:
+1 for the precommit event. It is very useful in Wayfire, where we start a transaction if the client resizes itself. The precommit event lets us then block the resize commit and start a new transaction, so that every resize is done via transactions.

vyivel:
There is an obvious optimization possibility: no reason to create a state addon if the pending state didn't change. But before that: do we need to cache anything other than geometry?

emersion:
> do we need to cache anything other than geometry?

For xdg_surface, probably the ack configure serial? For xdg_toplevel, the min/max size should be part of the cached state.

vyivel:
Fixed some stuff, added `wlr_surface_state_addon.finish_state`, added toplevel state caching.

vyivel:
@ammen99 

>+1 for the precommit event. It is very useful in Wayfire, where we start a transaction if the client resizes itself. The precommit event lets us then block the resize commit and start a new transaction, so that every resize is done via transactions.

If you're going to use `wlr_xdg_surface_get_geometry()`, that won't work as the function operates on the current state. You would need to reimplement the function with pending state.

ammen99:
> @ammen99
> 
> > +1 for the precommit event. It is very useful in Wayfire, where we start a transaction if the client resizes itself. The precommit event lets us then block the resize commit and start a new transaction, so that every resize is done via transactions.
> 
> If you're going to use `wlr_xdg_surface_get_geometry()`, that won't work as the function operates on the current state. You would need to reimplement the function with pending state.

Yep, that's what I'm doing, but thanks for the heads up ;)

vyivel:
@emersion gentle ping

I'd be glad if you could give your thoughts on the current state of the PR; it might have some issues that I'm unaware of.

emersion:
There's potential to rebase this on top of master and use `wlr_addon`. But as discussed on IRC we'll hit issues because of the way the state is moved around.

One possible solution is to alloc all `wlr_surface_state` and make sure it's never "moved". Seems like the simplest solution.

Another solution would be to add a "move" hook to surface state addons, e.g.

```c
struct wlr_surface_state_addon_interface {
    struct wlr_addon_interface base;
    void (*move)(struct wlr_addon *addon, struct wlr_surface_state *dst);
};
```

vyivel:
Closing this in favor of https://github.com/swaywm/wlroots/pull/3107.

PR #3075, opened 2021-07-29
render/gles2: add support for some 24 and 16-bit formats

emersion:
On little-endian, we can enable pixel formats which don't use
gl_type = GL_UNSIGNED_BYTE. See [1].

[1]: https://afrantzis.com/pixel-format-guide/

### DIFF
--- a/render/gles2/pixel_format.c
+++ b/render/gles2/pixel_format.c
@@ -32,6 +32,45 @@ static const struct wlr_gles2_pixel_format formats[] = {
 		.gl_type = GL_UNSIGNED_BYTE,
 		.has_alpha = true,
 	},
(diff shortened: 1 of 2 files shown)


bl4ckb0ne:
Is there a reason more than "just in case" for the 24 ans 16 bits formats?

emersion:
These formats are mostly relevant in more memory-constrained machines, e.g. embedded or old GPUs. I'll have to admit I've just added them mostly for completeness and I don't really intend to use them in a client. More of a "GL gives us this for free, so why not make it available to clients" kind of thing.

bl4ckb0ne:
Seems good to me

PR #3076, opened 2021-07-30
Add support for some 10-bit and FP16 formats

mstoeckl:
In addition to adding support for XBGR2101010, ABGR2101010, XBGR16161616F, and ABGR16161616F, this PR introduces a mechanism to filter the list of available shm formats based on the GL extensions that the renderer supports. See also commit messages.

I have a modified weston-simple-shm which supports these formats at https://gitlab.freedesktop.org/mstoeckl/weston/-/tree/simple-shm-10 ; there is also https://gitlab.freedesktop.org/jadahl/wayland-test-clients/blob/master/gradient-test.c .

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -43,6 +43,8 @@ struct wlr_gles2_renderer {
 		bool KHR_debug;
 		bool OES_egl_image_external;
 		bool OES_egl_image;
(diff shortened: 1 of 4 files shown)


emersion (approved):
LGTM, thanks!

PR #3078, opened 2021-07-30
backend/drm: preserve mode order from kernel

emersion:
The kernel orders the mode list from highest to lowest. Preserve
this ordering in the wlr_output.modes list.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

PR #3080, opened 2021-08-01
README.md: Link to projects that satisfy the dependencies

mfwitten:
I would have found this information very useful in understanding what the dependencies are, and how they might be satisfied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,26 +40,32 @@ language.
 
 Install dependencies:
 
-* meson
-* wayland
-* wayland-protocols
-* EGL
-* GLESv2
-* libdrm
-* GBM
-* libinput
-* xkbcommon
-* udev
-* pixman
+* [meson]
+* [Wayland], for example:
+  * Both [libwayland] and [wayland-protocols]
+* [EGL] and [GLESv2], for example:
+  * *libEGL* and *libGLESv2*, as provided by:
+    * [mesa] alone
+    * or [libglvnd] combined with a vendor driver (perhaps [mesa] again)
+* [libdrm]
+* *libgbm*, for example:
+  * [mesa]
+  * or [minigbm]
+* [libinput]
+* [libxkbcommon]
+* *udev*, for example:
+  * [systemd]
+  * or [eudev]
+* [pixman]
 * [libseat]
 
 If you choose to enable X11 support:
 
-* xwayland (build-time only, optional at runtime)
-* libxcb
-* libxcb-render-util
-* libxcb-wm
-* libxcb-errors (optional, for improved error reporting)
+* [XWayland] (build-time only, optional at runtime)
+* [libxcb]
+* [libxcb-render-util]
+* [libxcb-wm]
+* [libxcb-errors] (optional, for improved error reporting)
 
 Run these commands:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,26 +40,31 @@ language.
...
-* xkbcommon
-* udev
-* pixman
+* [meson]
+* [Wayland], namely:
+  * [libwayland]
+  * [wayland-protocols]
+* [EGL], namely:
+  * [libglvnd]
+  * A vendor driver ([mesa], or something proprietary, *etc.*)
```
emersion, line 49: Something proprietary isn't officially supported.

README.md
```diff
@@ -40,26 +40,31 @@ language.
...
-* libinput
-* xkbcommon
-* udev
-* pixman
+* [meson]
+* [Wayland], namely:
+  * [libwayland]
+  * [wayland-protocols]
+* [EGL], namely:
+  * [libglvnd]
```
emersion, line 48: libglvnd is not required. Plain Mesa works fine too (and arguably works better, because one less component in the chain which might not support and ext or feature). We need an EGL implementation, and there are many.

README.md
```diff
@@ -40,26 +40,31 @@ language.
...
+  * A vendor driver ([mesa], or something proprietary, *etc.*)
+* [GLESv2](https://www.khronos.org/opengles/), namely:
+  * [libglvnd]
```
emersion, line 51: Same here.

README.md
```diff
@@ -78,6 +83,23 @@ See [CONTRIBUTING.md].
 [wiki]: https://github.com/swaywm/wlroots/wiki/Getting-started
 [#sway-devel on Libera Chat]: https://web.libera.chat/?channels=#sway-devel
 [Sway]: https://github.com/swaywm/sway
+[meson]: https://mesonbuild.com/
+[libwayland]: https://github.com/wayland-project/wayland
```
emersion, line 87: Please use the upstream gitlab.freedesktop.org repo. GitHub is just a mirror.

README.md
```diff
@@ -40,26 +40,31 @@ language.
...
+  * [libwayland]
+  * [wayland-protocols]
+* [EGL], namely:
+  * [libglvnd]
+  * A vendor driver ([mesa], or something proprietary, *etc.*)
+* [GLESv2](https://www.khronos.org/opengles/), namely:
+  * [libglvnd]
+  * A vendor driver ([mesa], or something proprietary, *etc.*)
+* [libdrm]
+* *libgbm* ([mesa])
```
emersion, line 54: Technically there are also multiple GBM implementations, e.g. minigbm.

emersion (changes requested):

README.md
```diff
@@ -40,26 +40,33 @@ language.
...
-* EGL
-* GLESv2
-* libdrm
-* GBM
-* libinput
-* xkbcommon
-* udev
-* pixman
+* [meson]
+* [Wayland], for example:
```
emersion, line 44: These aren't examples, these are separate dependencies.

README.md
```diff
@@ -40,26 +40,33 @@ language.
...
+  * [libwayland]
+  * [wayland-protocols]
+* [EGL] and [GLESv2], for example:
+  * *libEGL* and *libGLESv2*, as provided by:
```
emersion, line 48: This doesn't give more info than the previous line.

emersion (changes requested):
In general I'm not a fan of adding sub-lists to this list.

mfwitten:
This pull request has been updated with the following change:

    diff --git a/README.md b/README.md
    index 0205cac8..c2bcc2e9 100644
    --- a/README.md
    +++ b/README.md
    @@ -42,8 +42,7 @@ Install dependencies:
    
     * [meson]
     * [Wayland], for example:
    -  * [libwayland]
    -  * [wayland-protocols]
    +  * Both [libwayland] and [wayland-protocols]
     * [EGL] and [GLESv2], for example:
       * *libEGL* and *libGLESv2*, as provided by:
         * [mesa] alone

Regarding the discussion:

* > In general I'm not a fan of adding sub-lists to this list.

    You can't have it both ways, Simon; you can't have *both* generality *and* specificity without providing delineation.

* > This doesn't give more info than the previous line.

    Either  the previous request means something or it doesn't: "We need [an] implementation, and there are many".

    Perhaps Microsoft Windows supplies *bills1337graphiX.dll*, amirite?

* OK, so, this is all a lot of bikeshedding.

    * This pulll request satisfies *both* generality *and* specificity, as requested.

    * You are making vague remarks from the point of view of someone who *already* knows all this information very well; in contrast, this pull request represents the information desired by an actual *newbie* (me) who knows *nothing*.

    * Tinkering capriciously is pointless.

        * I'm very satisfied with this contribution.

        * Either this contribution offers little value to you, in which case you should just merge it and move on; or, you have very specific needs that this contribution is not meeting, in which case you should merge it and then scratch your itches as you see fit thereafter (with a further commit); here are some examples:

                P: This pull request (merged)
                Y: Your itch scratching
                M: Explicit merge to consolidate
                
                        P    Y
                ---o----o----o
                
                       or
                
                ---o--------o M
                    \      /
                     o----o
                     P    Y

            In this way, we can *both* be satisfied. I will have produced work that I like and that was accepted for the benefit of others, and you will have maintained expert control over the contents of your repo.

        * Put another way: When a pull request produces results that are demonstrably better than what came before, yet not as good as you think it could be, you would do well as a maintainer just to merge that pull request right away, and then make any additional changes thereafter on your own. Such an approach satisfies everyone's tastes, benefits other users, *and* makes everyone feel productive. I can't be spending a week tinkering with a README file, and neither can you.

ddevault:
You're being an asshole. Don't.

mfwitten:
I disgree. I'm being very helpful (and, I'm not the one calling people slurs).

AdrianVovk:
Wow! Being disrespectful to one of the core maintainers of a project you want to contribute to is a fast pass to getting your PR rejected. Nobody owes you the privalage of getting your work merged in. Produce work "you like" all you want; if the maintainers don't like it you either need to fix it or accept that your work won't get used. This is not your project, and you don't get to make demands

bl4ckb0ne:
Thanks for your time, but I don't think this patch is going anywhere.

PR #3081, opened 2021-08-01
surface: don't cache frame callback lists

vyivel:
While the compositor may render the locked state, it still needs to make clients think their latest state is being rendered. Achieve that by not caching frame callback lists.


vyivel:
cc @ammen99

ammen99 (commented):
Tested this and seems to work in Wayfire, so LGTM.

In case someone wonders, we're sending the frame events inside the cache event handler (in case that the desired state is not reached yet), and of course on `wlr_output.frame`

emersion (approved):
LGTM as well, thanks!

vyivel:
Ah, there was a typo in the comment. Sorry about that.

PR #3082, opened 2021-08-01
High-resolution scroll wheel support

JoseExposito:
## Description:

Starting with Linux Kernel v5.0 two new axes are available for high-resolution wheel scrolling: `REL_WHEEL_HI_RES` and `REL_HWHEEL_HI_RES`.
Both axes send data in fractions of 120 where each multiple of 120 amounts to one logical scroll event. Fractions of 120 indicate a wheel movement less than one detent. A "detent" is the named used in the Kernel for a mouse wheel click:
https://www.kernel.org/doc/html/v5.11-rc7/input/event-codes.html#ev-rel

Three new events are now available on libinput: `LIBINPUT_EVENT_POINTER_SCROLL_WHEEL`, `LIBINPUT_EVENT_POINTER_SCROLL_FINGER`, and `LIBINPUT_EVENT_POINTER_SCROLL_CONTINUOUS`. These events **replace** the `LIBINPUT_EVENT_POINTER_AXIS` event, so new clients should simply ignore that event.

This `LIBINPUT_EVENT_POINTER_SCROLL_WHEEL` event adds a new API `libinput_pointer_scroll_get_value_v120()`. The `libinput_pointer_scroll_get_value_v120()` is a mirror from the kernel API (itself a copy of the Windows API). The new event is sent for **all** wheel events, even those that don't technically support high-resolution scrolling and even on older kernels that don't have this feature. So callers can simply ignore any `LIBINPUT_EVENT_POINTER_AXIS` event and use the new event types only.

For `LIBINPUT_EVENT_POINTER_SCROLL_FINGER` and `LIBINPUT_EVENT_POINTER_SCROLL_CONTINUOUS`, as well as `LIBINPUT_EVENT_POINTER_SCROLL_WHEEL`, a new API  is available: `libinput_pointer_scroll_get_value()`. This API is similar to `libinput_pointer_axis_get_value()`.

## Status

- [libinput](https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/652/edit): Ready for merge, waiting on clients like this one ;) 
- [Wayland](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/72/): It was open about a year ago by Peter Hutterer, once we have a compositor and a framework it'd be nice to get it reviewed.
- wlroots: This PR
- GTK: I'm working on it

## How to test it:

Clone and install my libinput branch:
https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/652

Clone and install Peter's Wayland protocol:
https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/72/

Reboot.

`tinywl` is being updated in this PR, so you can test that scroll is still working with backends that support and don't support high-resolution scroll.

I haven't updated GTK yet, so for the moment there are no clients with `wl_seat` version 8 :disappointed: 

Of course, you'll need hardware to test it.
Logitech mice are supported, otherwise any mouse that shows a `ResolutionMultiplier` in the hid-recorder output (most Microsoft-branded mice of the last decade).
Or, if you happen to have an Apple Magic Mouse, I [added support in the Kernel](https://lore.kernel.org/linux-input/nycvar.YFH.7.76.2107152116510.8253@cbobk.fhfr.pm/T/). You'll need to compile the Kernel with my patches, let me know if you need help.

## For code reviewers

It's easier to review this PR commit by commit. 

I hesitated a lot about "seat: support lo-res clients using hi-res backends". Placing the accumulators in `wlr_seat_client` is _technically_ incorrect, there should be an accumulator for each pointer in `wlr_seat_client->pointers`.

However, given that the accumulator expires after one second of inactivity and given than using 2 or more mice at the same time is a niche use case, I decided to leave the accumulator there for the sake of simplicity. Also, the side effects of using multiple mice at the same time are **really** difficult to notice.

Keeping the PR as draft until there is a client to test with.

cc @emersion  @whot


emersion:
Thanks for working on this! In case this hasn't been noticed, @tokyovigilante made a PR for the same feature a few years (!) ago: https://github.com/swaywm/wlroots/pull/1628

JoseExposito:
Hi @emersion, thanks for pointing it out, I didn't noticed there was an existing PR. However, libinput's API has changed a bit. This PR should include every change made until now.

tokyovigilante:
Thanks for the @, I'm happy to hand this over to you @JoseExposito and review or test. As you note it had been both a moving target as well as a bit of a chicken and egg scenario waiting for libinput, wayland and GTK to all get their ducks in a row which is why I had parked this.

JoseExposito:
@tokyovigilante yes... This kind of features can be tricky, because of all the parts involved, hopefully this time we will get it merged.

@emersion GTK is ready:
https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3839

I updated the PR description with build instructions and removed the "Draft" tag. I'll drop a comment on the Wayland MR with all the links.

JoseExposito:
libinput has been merged and 1.19 will be released in less than a month including high-resolution scroll support.

@emersion  at this point your feedback on the the [Wayland protocol](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/72/) or in this PR would be very valuable, so I can perform the changes in time.

JoseExposito:
Updated to use `axis_value120` instead of `axis_v120` as decided in the [protocol MR](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/72#note_1064146).

emersion:
This introduces a new event, `wlr_pointer.events.axis_value120`. However unlike the kernel or libinput we don't need to guarantee ABI compatibility. Couldn't we just add a new `delta_value120` field to `struct wlr_event_pointer_axis` instead?

JoseExposito:
@emersion yes, that should also work. I'm not as familiarized with the code as you are, so please me correct me if I'm wrong.

If we remove the new event and add a filed in `wlr_event_pointer_axis`, we could check is `value120 != 0` in `wlr_seat_pointer_send_axis` and pick the right value, making transparent to clients the handling of high-resolution events.

However, what would happen with compositors using their own `wlr_event_pointer_axis` handler? Wouldn't this change break scroll because they'd be using the `wlr_event_pointer_axis.delta` and `wlr_event_pointer_axis.delta_discrete` fields?

emersion:
Hm, is the old axis event trigger less often than the new one? Is it bad to not look at axis120 and just use axis/axis_discrete?

JoseExposito:
> is the old axis event trigger less often than the new one?

Yes, when your hardware supports high-resolution scroll, the new event is sent more often. For example, a mouse with an extra mouse wheel "click" will send:

```
High Res 60
High Res 60
Low Res 120 (discrete 1)
```

> Is it bad to not look at axis120 and just use axis/axis_discrete?

When both compositor and client support a version of the protocol >= 8, clients will ignore the old event in favour of the new one.

That means that if the compositor's custom implementation of `wlr_seat_pointer_send_axis` doesn't take the right delta value and calls `wl_pointer_send_axis_discrete` or `wl_pointer_send_axis_value120` accordingly, client scroll will break silently.

I think that, sadly, there's no way to not break the API? 🤔

whot:
> a mouse with an extra mouse wheel "click" will send:

An important note here: the mouse **may** send. There really is no correlation between the two events once they come out of the kernel. logitech devices in particular have some in-kernel wheel handling where some high-resolution events get filtered, so it's not always the same number of high-res events per low-res event.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3082

PR #3083, opened 2021-08-01
render/pixman/pixel_format: add more formats

emersion:
Add a bunch of new formats for Pixman: a few missing 32-bit ones,
some 16-bit and 32-bit formats as well.

Mostly based on a Weston patch [1].

[1]: https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/664

cc @mstoeckl @bl4ckb0ne

### DIFF
--- a/render/pixman/pixel_format.c
+++ b/render/pixman/pixel_format.c
@@ -35,7 +35,68 @@ static const struct wlr_pixman_pixel_format formats[] = {
 #else
 		.pixman_format = PIXMAN_a8b8g8r8,
 #endif
-	}
+	},
+	{
+		.drm_format = DRM_FORMAT_RGBA8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_a8b8g8r8,
+#else
+		.pixman_format = PIXMAN_r8g8b8a8,
+#endif
+	},
+	{
+		.drm_format = DRM_FORMAT_RGBX8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_x8b8g8r8,
+#else
+		.pixman_format = PIXMAN_r8g8b8x8,
+#endif
+	},
+	{
+		.drm_format = DRM_FORMAT_BGRA8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_a8r8g8b8,
+#else
+		.pixman_format = PIXMAN_b8g8r8a8,
+#endif
+	},
+	{
+		.drm_format = DRM_FORMAT_BGRX8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_x8r8g8b8,
+#else
+		.pixman_format = PIXMAN_b8g8r8x8,
+#endif
+	},
+#if WLR_LITTLE_ENDIAN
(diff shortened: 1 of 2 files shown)


mstoeckl:
The 2101010 formats might not draw correctly right now; see https://gitlab.freedesktop.org/pixman/pixman/-/issues/48 .

render/pixman/pixel_format.c
```diff
@@ -35,7 +35,65 @@ static const struct wlr_pixman_pixel_format formats[] = {
...
+	},
+	{
+		.drm_format = DRM_FORMAT_BGRX8888,
+#if WLR_BIG_ENDIAN
+		.pixman_format = PIXMAN_x8r8g8b8,
+#else
+		.pixman_format = PIXMAN_b8g8r8x8,
+#endif
+	},
+#if WLR_LITTLE_ENDIAN
```
bl4ckb0ne, line 71: Why is there no big endian versions of these formats?

emersion: All of these formats don't have big-endian equivalent because the components (r, g, b, a) are in-between two bytes. Here's how `DRM_FORMAT_RGB565` looks like on all systems:

```
  0                1
  M              L M              L
  G₂G₁G₀B₄B₃B₂B₁B₀ R₄R₃R₂R₁R₀G₅G₄G₃
```

Here's how `PIXMAN_r5g6b5` looks like on a big endian system:

```
  R₄R₃R₂R₁R₀G₅G₄G₃ G₂G₁G₀B₄B₃B₂B₁B₀ 
```

Here's how `PIXMAN_b5g6r5` looks like on a big endian system:

```
  B₄B₃B₂B₁B₀G₅G₄G₃ G₂G₁G₀R₄R₃R₂R₁R₀
```

See https://afrantzis.com/pixel-format-guide/drm.html

bl4ckb0ne: I get it for the 565, but what about the 2101010? It also migh be worth leaving a comment to explain it.

emersion: `DRM_FORMAT_ARGB2101010` also has components that span across multiple bytes:

```
BBBBBBBB GGGGGGBB RRRRGGGG AARRRRRR
```

So on big-endian machines there's no equivalent.

bl4ckb0ne: FWIW the two formats arent exactly equal

```
#define PIXMAN_FORMAT(bpp,type,a,r,g,b)	(((bpp) << 24) |  \
					 ((type) << 16) | \
					 ((a) << 12) |	  \
					 ((r) << 8) |	  \
					 ((g) << 4) |	  \
					 ((b)))

    PIXMAN_r5g6b5 =	 PIXMAN_FORMAT(16,PIXMAN_TYPE_ARGB,0,5,6,5),
    PIXMAN_b5g6r5 =	 PIXMAN_FORMAT(16,PIXMAN_TYPE_ABGR,0,5,6,5),
```

emersion: Yes, these two formats have the r and b channels swapped.

I've added a comment to explain why these formats are a special case.

bl4ckb0ne: Got confused by endianness, everything is clear now

bl4ckb0ne (approved):

PR #3085, opened 2021-08-02
output: try skipping buffer allocation if the backend allows it

emersion:
When enabling an output, skip the empty buffer allocation if the
backend accepts modesets without a buffer.

This fixes mode-setting with the noop backend.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM :)

PR #3088, opened 2021-08-05
input/pointer: try harder to not send duplicate motion events

Xyene:
`wl_fixed_t` is a 32-bit data type, but our doubles are 64-bit. This meant
that two doubles that would map to the same `wl_fixed_t` could compare
unequal, and send a duplicate motion event.

Refs swaywm/sway#4632.

### DIFF
--- a/types/seat/wlr_seat_pointer.c
+++ b/types/seat/wlr_seat_pointer.c
@@ -230,18 +230,21 @@ void wlr_seat_pointer_send_motion(struct wlr_seat *wlr_seat, uint32_t time,
 		return;
 	}
 
-	if (wlr_seat->pointer_state.sx == sx && wlr_seat->pointer_state.sy == sy) {
-		return;
-	}
+	// Ensure we don't send duplicate motion events. Instead of comparing with an
+	// epsilon, chop off some precision by converting to a `wl_fixed_t` first,
+	// since that is what a client receives.
+	wl_fixed_t sx_fixed = wl_fixed_from_double(sx);
+	wl_fixed_t sy_fixed = wl_fixed_from_double(sy);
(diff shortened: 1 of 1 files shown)


types/seat/wlr_seat_pointer.c
```diff
@@ -230,18 +230,21 @@ void wlr_seat_pointer_send_motion(struct wlr_seat *wlr_seat, uint32_t time,
...
-	if (wlr_seat->pointer_state.sx == sx && wlr_seat->pointer_state.sy == sy) {
-		return;
-	}
+	// Ensure we don't send duplicate motion events. Instead of comparing with an
+	// epsilon, chop off some precision by converting to a `wl_fixed_t` first,
+	// since that is what a client receives.
+	wl_fixed_t sx_fixed = wl_fixed_from_double(sx);
+	wl_fixed_t sy_fixed = wl_fixed_from_double(sy);
+	if (wl_fixed_from_double(wlr_seat->pointer_state.sx) != sx_fixed ||
+			wl_fixed_from_double(wlr_seat->pointer_state.sy) != sy_fixed) {
```
emersion, line 239: Wrapping the loops inside the `if` changes behaviour: `wlr_seat_pointer_warp()` was previously not executed, and now is. Is this intended?

vyivel: The behavior is actually the same, calling `wlr_seat_pointer_warp()` previously would've been no-op anyway.

Xyene: Yeah, I figured we might as well continue to accumulate partial updates in the pointer state, since that seems most intuitive.

emersion: OK, makes sense.

emersion (approved):
Thanks!

PR #3090, opened 2021-08-05
ci: add smoke test

emersion:
Add a very basic smoke test which uses VKMS to fire up the DRM
backend.

It's a bit hacky atm:

- `sleep` could be replaced with polling for the socket to be created
- `chmod` is necessary because the dumb buffer allocator needs to open the card directly

cc @kennylevinsen

### DIFF
--- a/.builds/archlinux.yml
+++ b/.builds/archlinux.yml
@@ -31,3 +31,13 @@ tasks:
   - clang: |
       cd wlroots/build-clang
       ninja
+  - smoke-test: |
+      cd wlroots/tinywl
+      sudo modprobe vkms
+      sudo seatd -u "$USER" &
+      while ! [ -e /run/seatd.sock ]; do sleep 0.1; done
+      export WLR_BACKENDS=drm
+      export WLR_RENDERER=pixman
+      export WLR_DRM_DEVICES=/dev/dri/by-path/platform-vkms-card
+      sudo chmod ugo+rw /dev/dri/by-path/platform-vkms-card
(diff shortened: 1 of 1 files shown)


kennylevinsen (approved):
It works and is way more than we have already, so LGTM. :)

Next steps would be:
1. Maybe a tinywl variant that can make a few asserts (output count for example)
2. Writeback screenshot?
3. Remove the sleep when we have seatd-launch

emersion:
> Remove the sleep when we have seatd-launch

I've updated the sleep to align it with Weston, ie. sleep until the socket is available. Should be more robust on a busy system.

> Maybe a tinywl variant that can make a few asserts (output count for example)

I'd rather not bake CI stuff into tinywl. Maybe we could have a separate compositor for this, but it'll come with extra maintenance costs. Or maybe rely on e.g. Sway CI instead?

For simple things like output count, the Wayland protocol can be used directly (e.g. with something like wayland-info).

> Writeback screenshot?

This one will be tricky. Writeback connectors need to be setup by the DRM backend, and right now they are completely separate from regular connectors.

There were some ideas about providing a V4L2 device for vkms connectors, but this hasn't materialized yet.

kennylevinsen:
> I'd rather not bake CI stuff into tinywl.

Me neither, was trying to imply a barebone compositor for testing, although we could see how far we could get with just tinywl as is for now.

> There were some ideas about providing a V4L2 device for vkms connectors, but this hasn't materialized yet.

I'd very much prefer this to having to deal with writeback ourselves if possible. It's also better from a testing perspective anyway as the test jig sits outside the compositor under test.

PR #3091, opened 2021-08-06
seat: allow compositors to not load a keymap

emersion:
The protocol allows compositors to not send any keymap to Wayland
clients. Handle a keymap-less keyboard correctly by sending
WL_KEYBOARD_KEYMAP_FORMAT_NO_KEYMAP instead of erroring out in the
mmap call.

### DIFF
--- a/types/seat/wlr_seat_keyboard.c
+++ b/types/seat/wlr_seat_keyboard.c
@@ -370,6 +370,13 @@ static void seat_client_send_keymap(struct wlr_seat_client *client,
 			continue;
 		}
 
+		if (keyboard->keymap == NULL) {
+			wl_keyboard_send_keymap(resource,
+				WL_KEYBOARD_KEYMAP_FORMAT_NO_KEYMAP, keymap_fd, 0);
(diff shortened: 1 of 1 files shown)


types/seat/wlr_seat_keyboard.c
```diff
@@ -370,6 +370,13 @@ static void seat_client_send_keymap(struct wlr_seat_client *client,
 			continue;
 		}
 
+		if (keyboard->keymap == NULL) {
+			wl_keyboard_send_keymap(resource,
+				WL_KEYBOARD_KEYMAP_FORMAT_NO_KEYMAP, keymap_fd, 0);
```
kennylevinsen, line 375: I'm assuming that there isn't a way for us to skip sending an fd here (-1?), and that we're allocating this fd just to send a legal but useless value across?

emersion: Yes, libwayland errors out if we send -1.

kennylevinsen (approved):
LGTM, thanks!

PR #3092, opened 2021-08-07
seat: avoid copying the keymap for each client 

emersion:
If a client supports wl_keyboard version 7, no need to allocate
and copy a new keymap file. We can use a read-only file instead.

### DIFF
--- a/include/util/shm.h
+++ b/include/util/shm.h
@@ -1,7 +1,10 @@
 #ifndef UTIL_SHM_H
 #define UTIL_SHM_H
 
+#include <stdbool.h>
+
 int create_shm_file(void);
 int allocate_shm_file(size_t size);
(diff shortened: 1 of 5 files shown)


emersion:
TBH, I wonder if we even need `wl_seat` v7 to do this. v7 only mentions MAP_SHARED, not PROT_READ.

kennylevinsen:
I don't see anything mentioned, so let's do it unconditionally. Worst case we discover a faulty protocol assumption. :)

emersion:
Updated to always send a read-only FD. Tested with a few clients (GTK, Qt, foot, alacritty, etc), all seem to work fine.

Also opened https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/176 to make the spec clearer.

kennylevinsen (approved):
LGTM, thanks!

PR #3095, opened 2021-08-09
backend/drm: implement output layers

emersion:
Depends on #1985


emersion:
The ad-hoc copies of `wlr_output_state` are making things difficult, because they corrupt `wl_list` pointers.

I'll refactor the DRM backend a bit to avoid doing such copies.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3095

Issue #2927, opened 2021-05-17
Efficient hashmap for add-on structs


emersion:
Often times in wlroots, we need to extend "base" objects like `wlr_surface` or `wlr_buffer`:

- [`wlr_presentation_surface_sampled`](https://github.com/swaywm/wlroots/blob/6369f7093178f0d66414eb67c312a403bfbb84a4/include/wlr/types/wlr_presentation_time.h#L86)
- [`drm_fb_get`](https://github.com/swaywm/wlroots/blob/6369f7093178f0d66414eb67c312a403bfbb84a4/backend/drm/renderer.c#L447)
- [`gles2_texture_from_buffer`](https://github.com/swaywm/wlroots/pull/2851#discussion_r633555311)

Instead of using a `wl_list`, using a hashmap that uses a pointer for the keys would be more efficient. It would be nice to study from which number of element the list approach becomes an issue performance-wise.

bl4ckb0ne:
Fwiw I like the one mcf did for [samurai][1]. I cleaned the implementation in my [personal snippets][2]. 


[1]: https://git.sr.ht/~mcf/samurai/tree/master/item/htab.c] 
[2]: https://git.sr.ht/~bl4ckb0ne/tyr/tree/master/item/tyr/hash_table.c

emersion:
That implementations hashes string keys, though. We can probably drop `murmurhash64a` completely and replace it with a a function that shuffles the bits of the `uintptr_t` keys sufficiently.

Also, I'm not sure about that hashtable behaviour when two different entries end up with the same hash… It doesn't seem like this case is handled?

bl4ckb0ne:
I think the murmur hash has a pretty low collision rate.

emersion:
Low collision rate doesn't mean that collisions won't happen.

kennylevinsen:
> Also, I'm not sure about that hashtable behaviour when two different entries end up with the same hash… It doesn't seem like this case is handled?

It's handled in key_index. The hash is used to find the *initial* index, and key_index then simply increments the index until it either finds a matching key or an empty slot. It grows whenever usage reaches 50% capacity, at which point it rebalances.

Note that hash_table_put does *not* override the value if it already exists (that is, key_index found an entry). In this case it simply returns a pointer to the existing value.

> I think the murmur hash has a pretty low collision rate.

Murmur hash is a cheap and small hash and thus have a pretty high likelihood of collision, but it doesn't matter as we don't really need arbitrary string hashing.

[...]

PR #3096, opened 2021-08-09
util: add wlr_addon

vyivel:
Closes: https://github.com/swaywm/wlroots/issues/2927

### DIFF
--- a/include/wlr/util/addon.h
+++ b/include/wlr/util/addon.h
@@ -1,0 +1,43 @@
+/*
+ * This an unstable interface of wlroots. No guarantees are made regarding the
+ * future consistency of this API.
+ */
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_UTIL_ADDON_H
+#define WLR_UTIL_ADDON_H
+
+#include <wayland-server-core.h>
+
+struct wlr_addon_set {
+	// private state
+	struct wl_list addons;
+};
+
+struct wlr_addon;
+
+struct wlr_addon_interface {
+	const char *name;
+	void (*destroy)(struct wlr_addon *addon);
+};
+
+struct wlr_addon {
+	const struct wlr_addon_interface *impl;
+	// private state
+	const void *owner;
+	struct wl_list link;
+};
+
+void wlr_addon_set_init(struct wlr_addon_set *set);
+void wlr_addon_set_finish(struct wlr_addon_set *set);
+
+void wlr_addon_init(struct wlr_addon *addon, struct wlr_addon_set *set,
+		const void *owner, const struct wlr_addon_interface *impl);
+void wlr_addon_finish(struct wlr_addon *addon);
+
+struct wlr_addon *wlr_addon_find_by_owner(struct wlr_addon_set *set,
+		const void *owner);
+
+#endif
--- a/util/addon.c
+++ b/util/addon.c
@@ -1,0 +1,49 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <wayland-server-core.h>
+
+#include "wlr/util/addon.h"
+
+void wlr_addon_set_init(struct wlr_addon_set *set) {
+	wl_list_init(&set->addons);
+}
+
+void wlr_addon_set_finish(struct wlr_addon_set *set) {
+	struct wlr_addon *addon, *tmp;
+	wl_list_for_each_safe(addon, tmp, &set->addons, link) {
+		wlr_addon_finish(addon);
+		addon->impl->destroy(addon);
+	}
+}
+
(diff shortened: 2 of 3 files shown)


util/addon.c
```diff
@@ -0,0 +1,47 @@
...
+void wlr_addon_set_finish(struct wlr_addon_set *set) {
+	struct wlr_addon *addon, *tmp;
+	wl_list_for_each_safe(addon, tmp, &set->addons, link) {
+		wlr_addon_destroy(addon);
+	}
+}
+
+struct wlr_addon *wlr_addon_create(struct wlr_addon_set *set, void *data,
+		void *owner, wlr_addon_finish_data_func_t finish_data) {
+	struct wlr_addon *addon = calloc(1, sizeof(*addon));
```
emersion, line 19: Instead of allocating a `wlr_addon` here, can we make the caller responsible for embedding it in one of its struct?

util/addon.c
```diff
@@ -0,0 +1,47 @@
...
+
+void wlr_addon_set_finish(struct wlr_addon_set *set) {
+	struct wlr_addon *addon, *tmp;
+	wl_list_for_each_safe(addon, tmp, &set->addons, link) {
+		wlr_addon_destroy(addon);
+	}
+}
+
+struct wlr_addon *wlr_addon_create(struct wlr_addon_set *set, void *data,
+		void *owner, wlr_addon_finish_data_func_t finish_data) {
```
emersion, line 18: Maybe we should assert that we don't have another addon with the same owner?

util/addon.c
vyivel:
```c
struct wlr_xdg_surface_state {
    struct wlr_addon addon;
}
```
Like that?

util/addon.c
vyivel: Good idea.

util/addon.c
emersion: Yup. Doesn't necessarily need to be the first field, and users can get back the parent struct with something like `struct wlr_xdg_surface_state *xdg_state = wl_container_of(addon, xdg_state, addon)`.

kennylevinsen: Yeah, that would be better - then instead of creating, one would do `wlr_addon_set_insert(&surface->addon_set, &xdg_surface->addon)`.

kennylevinsen:
Do we have a good usage example to validate the interface?

util/addon.c
vyivel: Actually, if we want to hide the implementation, we don't want to expose `wlr_addon` with its `wl_list link` at all.

emersion:
Potential users: presentation-time, the DRM backend, the GLES2 renderer.

util/addon.c
emersion: We can mark the `wl_list` field as private state:

```c
struct wlr_addon {
    // private state
    struct wl_list link;
};
```

vyivel:
Pushed another rough attempt. Perhaps `wlr_addon_set_add()` could be merged with `wlr_addon_init()`.

include/wlr/util/addon.h
```diff
@@ -0,0 +1,38 @@
...
+
+struct wlr_addon;
+
+struct wlr_addon_interface {
+	const char *name;
+	void (*destroy)(struct wlr_addon *addon);
+};
+
+struct wlr_addon {
+	void *marker;
```
emersion, line 18: Maybe we can remove this field and use `impl` instead? Is there a use-case where we'd need to use a marker different from the `impl`?

vyivel: Only if the owner has more than one addon.

emersion: I'd prefer to wait a bit and see if someone really needs this. We can always add it later.

Note, one could do something like this and get the same behavior as `owner`/`marker`:

```c
struct my_protocol_impl {
    struct wlr_addon_interface addon_impl;
};
```

emersion:
Overall this sounds good. Would be nice to demonstrate its usage with an existing wlroots impl as @kennylevinsen said.

include/wlr/util/addon.h
```diff
@@ -0,0 +1,38 @@
...
+struct wlr_addon;
+
+struct wlr_addon_interface {
+	const char *name;
+	void (*destroy)(struct wlr_addon *addon);
+};
+
+struct wlr_addon {
+	const struct wlr_addon_interface *impl;
+	void *owner;
```
emersion, line 20: Maybe this could be made `const`, so that the caller can set `owner` to a pointer to a `struct wlr_addon_interface`. And maybe make this private state?

vyivel: Done, although I don't think `owner` should be set to impl. wlroots doesn't have anything global so there always will be something to set as owner anyway.

emersion (approved):
LGTM!

emersion:
Merged with additional changes in https://github.com/swaywm/wlroots/pull/3097.

PR #3097, opened 2021-08-09
wlr_output_layout: use wlr_addons 

vyivel:
This is an example of how https://github.com/swaywm/wlroots/pull/3096 can be used. Should speed up things a little bit in case a user has over 100 monitors.

### DIFF
--- a/types/wlr_output_layout.c
+++ b/types/wlr_output_layout.c
@@ -21,8 +21,9 @@ struct wlr_output_layout_output_state {
 
 	struct wl_listener mode;
 	struct wl_listener commit;
-	struct wl_listener output_destroy;
 };
+
+static const struct wlr_addon_interface addon_impl;
 
 struct wlr_output_layout *wlr_output_layout_create(void) {
 	struct wlr_output_layout *layout =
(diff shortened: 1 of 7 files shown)


types/wlr_output_layout.c
```diff
@@ -52,6 +69,7 @@ static void output_layout_output_destroy(
 	wl_list_remove(&l_output->state->commit.link);
 	wl_list_remove(&l_output->state->output_destroy.link);
```
emersion, line 70: Now that `l_output` is an addon, maybe we can get rid of the `output_destroy` listener?

types/wlr_output_layout.c
```diff
@@ -24,6 +24,23 @@ struct wlr_output_layout_output_state {
...
+		struct wlr_output_layout_output *l_output);
+
+static void addon_destroy(struct wlr_addon *addon) {
+	assert(addon->impl == &addon_impl);
+	struct wlr_output_layout_output *l_output =
+		wl_container_of(addon, l_output, addon);
+	output_layout_output_destroy(l_output);
+}
+
+static struct wlr_addon_interface addon_impl = {
```
emersion, line 39: This can be `const`

emersion (approved):
The commits "wlr_outputs: add wlr_addon_set" and  "wlr_output_layout: use wlr_addons" LGTM!

emersion:
Thanks!

PR #3099, opened 2021-08-10
backend/drm: add proper error handling to wlr_drm_backend_create

emersion:
Some listeners weren't removed and caused a use-after-free with
e.g. vkms when used as a secondary GPU.


kennylevinsen (approved):
LGTM, thanks!

PR #3101, opened 2021-08-11
Use addons in GLES2 renderer

emersion:
_See individual commits._

cc @vyivel

### DIFF
--- a/include/render/gles2.h
+++ b/include/render/gles2.h
@@ -108,8 +108,7 @@ struct wlr_gles2_texture {
 	uint32_t drm_format; // used to interpret upload data
 	// If imported from a wlr_buffer
 	struct wlr_buffer *buffer;
-
-	struct wl_listener buffer_destroy;
+	struct wlr_addon buffer_addon;
 };
 
 
--- a/render/gles2/texture.c
+++ b/render/gles2/texture.c
@@ -129,7 +129,9 @@ static bool gles2_texture_invalidate(struct wlr_gles2_texture *texture) {
 
 void gles2_texture_destroy(struct wlr_gles2_texture *texture) {
 	wl_list_remove(&texture->link);
-	wl_list_remove(&texture->buffer_destroy.link);
+	if (texture->buffer != NULL) {
+		wlr_addon_finish(&texture->buffer_addon);
+	}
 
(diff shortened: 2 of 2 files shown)


vyivel:
Could probably set `buffer_addon_impl.name` to something, although it's not strictly required. That aside, LGTM.

---

Maybe we could make `wlr_addon_find` take both owner and impl to avoid using lists as owners, if there will be more cases of one owner having multiple types of addons.

emersion:
> Could probably set buffer_addon_impl.name to something, although it's not strictly required

Oh, yes, completely forgot that. Done.

> Maybe we could make wlr_addon_find take both owner and impl to avoid using lists as owners, if there will be more cases of one owner having multiple types of addons. (Or add wlr_addon_find_by_impl.)

Yeah, I've been wondering about this too. The renderer's case seems pretty unique, I don't think there's another instance of it in the current wlroots codebase. I also like the fact that the owner pointer is a "primary key" for the addon. OTOH, it would be nice to take the `wlr_addon_interface` in the function which finds an addon, even if it's just to `assert(addon->impl == impl)`.

vyivel:
Hmm, maybe having to pass impl to the finder function isn't a good idea, as this only allows to find addons in the file where the implementation is defined (or the user has to declare in the header, which is ugly). Assertions can be done by the caller if desired.

emersion:
> this only allows to find addons in the file where the implementation is defined

Is there a case where an external user would want to use `wlr_addon_find_by_owner` without being the owner?

I'd much rather have the owner expose a safe function to get the addon from the base object e.g. 

```c
struct wlr_gles2_texture *wlr_gles2_buffer_from_buffer(struct wlr_buffer *buffer);
```

because the value of the owner field is an implementation detail.

So, in fact, I'd see this restriction as a feature.

> Assertions can be done by the caller if desired.

Right. Having the assert in `wlr_addon` would ensure callers don't forget it.

vyivel:
>Is there a case where an external user would want to use wlr_addon_find_by_owner without being the owner?

Unless they are doing some hacky things, I guess there isn't.

emersion:
Rebased on top of https://github.com/swaywm/wlroots/pull/3102.

PR #3102, opened 2021-08-11
util/addon: find both by owner and impl

vyivel:
This allows to have multiple addons of different types with the same
owner.


emersion:
Ref discussion in https://github.com/swaywm/wlroots/pull/3101#issuecomment-896632559

emersion (approved):
LGTM, thanks!

PR #3103, opened 2021-08-11
surface: kill wlr_surface.previous 

emersion:
This wlr_surface_state field was a special case because we don't
want to save the whole current state: for instance, the wlr_buffer
must not be saved or else wouldn't get released soon enough.

Let's just inline the state fields we need instead.

This patch should have zero logic changes, just refactoring.


vyivel:
Dropping `surface_state_copy()` is great. I'd put `previous_*` fields in a `struct previous {}`, would be a bit cleaner IMO.
LGTM otherwise.

emersion:
Nice idea, done!

PR #3106, opened 2021-08-11
xdg-shell: introduce wlr_xdg_surface_state

emersion:
This is preliminary work for properly handling cached state in
xdg-shell.

cc @vyivel

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -191,14 +191,11 @@ struct wlr_xdg_surface {
 	struct wl_list popups; // wlr_xdg_popup::link
 
 	bool added, configured, mapped;
-	uint32_t configure_serial;
 	struct wl_event_source *configure_idle;
 	uint32_t configure_next_serial;
 	struct wl_list configure_list;
 
-	struct wlr_box geometry;
-
-	struct wlr_xdg_surface_state pending;
+	struct wlr_xdg_surface_state current, pending;
 
 	struct wl_listener surface_destroy;
 	struct wl_listener surface_commit;
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -192,7 +192,7 @@ struct wlr_xdg_surface {
 
 	bool added, configured, mapped;
 	struct wl_event_source *configure_idle;
-	uint32_t configure_next_serial;
+	uint32_t scheduled_serial;
 	struct wl_list configure_list;
 
 	struct wlr_xdg_surface_state current, pending;
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -161,6 +161,11 @@ struct wlr_xdg_surface_configure {
 	struct wlr_xdg_toplevel_configure *toplevel_configure;
 };
 
+struct wlr_xdg_surface_state {
+	uint32_t configure_serial;
+	struct wlr_box geometry;
+};
+
 /**
  * An xdg-surface is a user interface element requiring management by the
  * compositor. An xdg-surface alone isn't useful, a role should be assigned to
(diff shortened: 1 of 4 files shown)


vyivel:
Ah, I added it myself already :D
Also, the state should probably have configure serial as well.

emersion:
I've updated this PR to just cherry-pick @vyivel's commit from https://github.com/swaywm/wlroots/pull/3107. Also added a small description of the change in the commit message.

vyivel:
Could you please update this to have `wlr_xdg_surface.current` as well? `surface_commit_state()` can be inlined.

emersion:
Done!

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -81,14 +81,13 @@ void unmap_xdg_surface(struct wlr_xdg_surface *surface) {
 	}
 
 	surface->configured = surface->mapped = false;
-	surface->configure_serial = 0;
 	if (surface->configure_idle) {
 		wl_event_source_remove(surface->configure_idle);
 		surface->configure_idle = NULL;
 	}
 	surface->configure_next_serial = 0;
```
vyivel, line 88: With `surface->configure_serial = 0` removed, this should be removed as well.

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -191,14 +191,11 @@ struct wlr_xdg_surface {
 	struct wl_list popups; // wlr_xdg_popup::link
 
 	bool added, configured, mapped;
-	uint32_t configure_serial;
 	struct wl_event_source *configure_idle;
 	uint32_t configure_next_serial;
```
vyivel, line 195: I'd rename this to `scheduled_serial` to follow xdg-toplevel/layer-surface fields naming. (Although this is maybe slightly out of scope of this PR.)

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -81,14 +81,13 @@ void unmap_xdg_surface(struct wlr_xdg_surface *surface) {
...
 
-	memset(&surface->geometry, 0, sizeof(struct wlr_box));
+	surface->current = surface->pending;
 	memset(&surface->pending, 0, sizeof(struct wlr_xdg_surface_state));
```
vyivel, line 91: >Once the window geometry of the surface is set, it is not possible to unset it, and it will remain the same until set_window_geometry is called again, even if a new subsurface or buffer is attached.

I don't think the state should be cleared here. Not sure why was this done in the first place though.

types/xdg_shell/wlr_xdg_surface.c
```diff
@@ -310,22 +309,14 @@ static void xdg_surface_handle_surface_commit(struct wl_listener *listener,
...
-
 void handle_xdg_surface_commit(struct wlr_surface *wlr_surface) {
 	struct wlr_xdg_surface *surface =
 		wlr_xdg_surface_from_wlr_surface(wlr_surface);
 	if (surface == NULL) {
 		return;
 	}
 
-	surface_commit_state(surface, &surface->pending);
+	surface->current = surface->pending;
```
vyivel, line 319: As `has_geometry` isn't checked here, that field could be removed.

types/xdg_shell/wlr_xdg_surface.c
emersion: Hm, I'm not sure. `surface->configure_serial = 0` is removed because the `memset` below takes care of resetting it in `current`.

include/wlr/types/wlr_xdg_shell.h
emersion: Done

types/xdg_shell/wlr_xdg_surface.c
emersion: I think I've removed all occurrences?

types/xdg_shell/wlr_xdg_surface.c
emersion: Ah, good catch.

types/xdg_shell/wlr_xdg_surface.c
vyivel: Sorry, somehow didn't notice that.

types/xdg_shell/wlr_xdg_surface.c
emersion: Hm, wait, I think I mixed things up. This is the unmap handler, it needs to reset all of the state.

Fixed.

vyivel (approved):
This iteration LGTM.
Thanks!

PR #3107, opened 2021-08-11
[obsolete] xdg-surface: use addons for state caching

vyivel:
~~Depends on https://github.com/swaywm/wlroots/pull/3103~~

cc @emersion

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -474,6 +474,11 @@ static void surface_commit_state(struct wlr_surface *surface,
 		}
 	}
 
+	wlr_signal_emit_safe(&surface->events.commit_addons, next);
+	// Reset the addon set
+	wlr_addon_set_finish(&next->addons);
+	wlr_addon_set_init(&next->addons);
+
 	// If we're committing the pending state, bump the pending sequence number
 	// here, to allow commit listeners to lock the new pending state.
 	if (next == &surface->pending) {
@@ -775,6 +780,7 @@ struct wlr_surface *surface_create(struct wl_client *client,
 	wl_signal_init(&surface->events.destroy);
 	wl_signal_init(&surface->events.new_subsurface);
 	wl_signal_init(&surface->events.prepare_addons);
+	wl_signal_init(&surface->events.commit_addons);
 	wl_list_init(&surface->subsurfaces_above);
 	wl_list_init(&surface->subsurfaces_below);
 	wl_list_init(&surface->subsurfaces_pending_above);
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -419,6 +419,7 @@ static void surface_cache_pending(struct wlr_surface *surface) {
 
 	surface_state_init(cached);
 	surface_state_move(cached, &surface->pending);
+	wlr_signal_emit_safe(&surface->events.prepare_addons, cached);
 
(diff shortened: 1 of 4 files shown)


vyivel:
Perhaps `apply` event could be removed so pending addon states would be applied on `commit`. libwayland notifies listeners in order of adding, so that should work for the same reasons things work now.

types/wlr_surface.c
```diff
@@ -491,6 +491,8 @@ static void surface_commit_pending(struct wlr_surface *surface) {
 		surface->role->precommit(surface);
 	}
 
+	wlr_signal_emit_safe(&surface->events.precommit, surface->pending);
```
emersion, line 496: "precommit" is confusing: it's not always triggered right before commit, it may be triggered right before the state is cached instead.

emersion:
It's a bit confusing that the `wlr_addon_set` gets "lost" when the surface state is made current. I understand this allows to side-step the issue of having to move the addon set…

types/wlr_surface.c
vyivel: Can be renamed to `prepare`.

vyivel:
>It's a bit confusing that the wlr_addon_set gets "lost" when the surface state is made current.

That can be fixed by having another set of addons just for `wlr_surface.current`.

vyivel:
Also, it's possible to avoid making `pending` a pointer at the cost of either allocating/deallocating a state on each commit or having something like `struct wlr_surface_state *to_commit` in `wlr_surface`.

vyivel:
TBH it seems a bit illogical to me that `wlr_surface_state` is used both to represent a complete state (`wlr_surface.current`) and a difference between two states (`wlr_surface.pending`, cached states).

Considering that `wlr_surface.current` is special in this case, maybe it would make sense to get rid of it and move the fields to `struct current {}` in `wlr_surface`. This change wouldn't break compositors that much.

emersion:
> TBH it seems a bit illogical to me that wlr_surface_state is used both to represent a complete state (wlr_surface.current) and a difference between two states (wlr_surface.pending, cached states).

Why? What's the difference?

vyivel:
A difference between two complete states isn't a complete state itself, it can't be used for rendering or calculating extends or stuff like that, what it actually is is a commit, not a surface state at a point in time. Likewise, while complete states could represent commits, they are not what commits actually are.

---

That was just a random thought, that's not an issue that should've had been dealt with or something.

vyivel:
To reiterate some thoughts and add some new ones:
- `apply` event isn't really required, libwayland notifies listeners in order of adding, so extension commit handlers will be always be the first to get the event and apply the state.
- `wlr_surface_state` is fine as it is. :P
- Making `pending` a pointer can (and should, imo) be avoided by adding `struct wlr_surface_state *next` field.
- In this PR, `surface_state_move()` is deleted and `surface_apply_state()` merges the state with the current one. I think `surface_state_move()` should be left as it is; in the future it can be used to implement state "squashing".
- `precommit` is named misleadingly; it should be renamed to something like `finalize_pending` or `prepare` and labeled as extension-only signal. For those users who might want to know when the client commtis, regardless of that commit being cached or applied, a `client_commit` event could be added.
- `surface->role->precommit()` should probably be called before actually applying the state.

vyivel:
Closed in favor of https://github.com/swaywm/wlroots/pull/3151.

PR #3108, opened 2021-08-12
layer-shell: replace close() with destroy() 

ifreund:
The protocol specifies that all requests (aside from destroy) are
ignored after the compositor sends the closed event. Therefore,
destroying the wlroots object and rendering the resource inert
when sending the closed event keeps things simpler for wlroots and
compositors.

### DIFF
--- a/include/wlr/types/wlr_layer_shell_v1.h
+++ b/include/wlr/types/wlr_layer_shell_v1.h
@@ -70,7 +70,7 @@ struct wlr_layer_surface_v1 {
 
 	char *namespace;
 
-	bool added, configured, mapped, closed;
+	bool added, configured, mapped;
 	uint32_t configure_serial;
 	uint32_t configure_next_serial;
 	struct wl_list configure_list;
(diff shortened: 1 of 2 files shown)


emersion:
Ideally we should make the layer-surface inert when the compositor decides to close it…

ifreund:
> Ideally we should make the layer-surface inert when the compositor decides to close it…

Yep, that's cleaner. Done.

ifreund:
Changed this to actually destroy the layer_surface in close() in keeping with the section on inert resources in CONTRIBUTING.md. 

This is now a breaking change and should be marked as such.

emersion (commented):
LGTM! I wonder if we should rename `wlr_layer_surface_v1_close` to `wlr_layer_surface_v1_destroy` to make it extra-clear that the function will make the layer pointer invalid.

ifreund:
@emersion good point, changing the function name will also make it easier for compositors to audit call sites when they update.

emersion (approved):
LGTM

emersion:
Thanks!

PR #3112, opened 2021-08-16
render: unify texture format enumeration

emersion:
Instead of having separate get_shm_texture_formats and
get_dmabuf_texture_formats functions, have a single unified
get_texture_formats function. This brings the renderer API in line
with wlr_backend_impl.


include/render/gles2.h
```diff
@@ -74,6 +74,8 @@ struct wlr_gles2_renderer {
 
 	struct wlr_gles2_buffer *current_buffer;
 	uint32_t viewport_width, viewport_height;
+
+	struct wlr_drm_format_set data_ptr_texture_formats;
```
bl4ckb0ne, line 78: Should this be stored if it's used only when calling `gles2_get_texture_formats`?

bl4ckb0ne: If it needs to be stored, it's missing a call to `wlr_drm_format_set_finish`

emersion: Yeah, we need to store it, otherwise we can't return it to the caller and let the caller use it after the function call has returned in `wlr_renderer_get_texture_formats`.

Good catch! Added the `wlr_drm_format_set_finish` call.

render/gles2/pixel_format.c
```diff
@@ -136,16 +136,12 @@ const struct wlr_gles2_pixel_format *get_gles2_format_from_gl(
...
-		size_t *len) {
-	static uint32_t shm_formats[sizeof(formats) / sizeof(formats[0])];
-	size_t j = 0;
+void init_gles2_data_ptr_formats(struct wlr_gles2_renderer *renderer) {
 	for (size_t i = 0; i < sizeof(formats) / sizeof(formats[0]); i++) {
 		if (!is_gles2_pixel_format_supported(renderer, &formats[i])) {
 			continue;
 		}
-		shm_formats[j++] = formats[i].drm_format;
+		wlr_drm_format_set_add(&renderer->data_ptr_texture_formats,
```
bl4ckb0ne, line 144: Why isnt the return value checked?

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3112

PR #3114, opened 2021-08-16
xwayland: do not free cursor in handle_server_ready()

michelesr:
If XWayland terminates for any reason, `xwm_set_cursor()` has to to be called again, so the cursor has to stick around.


emersion (approved):
Thanks!

PR #3116, opened 2021-08-17
surface: fix place_below handling

vyivel:
If a subsurface is being placed below a subsurface right above it, this
should be a noop. However, `node` pointed to the subsurface that was
moved, which resulted in `subsurface->parent_pending_link` being
inserted into itself, breaking parent's pending subsurface list.

This commit separates finding the requested node and getting it's `prev`
field, fixing the issue.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -961,7 +961,7 @@ static void subsurface_handle_place_below(struct wl_client *client,
 
 	struct wl_list *node;
 	if (sibling_surface == subsurface->parent) {
-		node = subsurface->parent->subsurfaces_pending_below.prev;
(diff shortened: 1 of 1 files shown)


emersion:
Is there the same issue with `place_above`?

vyivel:
No, `place_above` works correctly, as it doesn't require `prev` field.

emersion:
Hm. Do you have a reproducer for the issue?

vyivel:
`wleird/subsurfaces` works.
Run it and try placing green subsurface below blue one. Nothing should change, but it changes.

emersion (approved):
Good catch! LGTM, thanks for the fix!

PR #3117, opened 2021-08-17
scene: add support for damage tracking

emersion:
Introduce `wlr_scene_output` to have per-output state in the scene-graph.

I've decided to _not_ make `wlr_scene_output` a `wlr_scene_node` to make a clear cut between elements that are rendered on screen and viewports in the scene.

~~Depends on: https://github.com/swaywm/wlroots/pull/1966~~
Cage patch: https://github.com/Hjdskes/cage/pull/203

### DIFF
--- a/types/wlr_scene.c
+++ b/types/wlr_scene.c
@@ -426,7 +426,7 @@ void wlr_scene_render_output(struct wlr_scene *scene, struct wlr_output *output,
 			.output = output,
 			.damage = damage,
 		};
-		scene_node_for_each_node(&scene->node, lx, ly,
+		scene_node_for_each_node(&scene->node, -lx, -ly,
 			render_node_iterator, &data);
 		wlr_renderer_scissor(renderer, NULL);
 	}
--- a/types/wlr_scene.c
+++ b/types/wlr_scene.c
@@ -477,3 +477,27 @@ void wlr_scene_output_set_position(struct wlr_scene_output *scene_output,
 	scene_output->x = lx;
 	scene_output->y = ly;
 }
+
+bool wlr_scene_output_commit(struct wlr_scene_output *scene_output) {
+	struct wlr_output *output = scene_output->output;
+
+	if (!wlr_output_attach_render(output, NULL)) {
+		return false;
+	}
+
+	struct wlr_renderer *renderer = wlr_backend_get_renderer(output->backend);
+	assert(renderer != NULL);
+
+	int width, height;
+	wlr_output_effective_resolution(output, &width, &height);
+	wlr_renderer_begin(renderer, width, height);
+	wlr_renderer_clear(renderer, (float[4]){ 0.0, 0.0, 0.0, 0.0 });
+
+	wlr_scene_render_output(scene_output->scene, output,
+		scene_output->x, scene_output->y, NULL);
+	wlr_output_render_software_cursors(output, NULL);
+
+	wlr_renderer_end(renderer);
+
+	return wlr_output_commit(output);
+}
--- a/types/wlr_scene.c
+++ b/types/wlr_scene.c
@@ -71,6 +71,12 @@ void wlr_scene_node_destroy(struct wlr_scene_node *node) {
 	switch (node->type) {
 	case WLR_SCENE_NODE_ROOT:;
 		struct wlr_scene *scene = scene_root_from_node(node);
+
+		struct wlr_scene_output *scene_output, *scene_output_tmp;
+		wl_list_for_each_safe(scene_output, scene_output_tmp, &scene->outputs, link) {
+			wlr_scene_output_destroy(scene_output);
+		}
+
 		free(scene);
 		break;
 	case WLR_SCENE_NODE_SURFACE:;
@@ -91,7 +97,7 @@ struct wlr_scene *wlr_scene_create(void) {
 		return NULL;
 	}
 	scene_node_init(&scene->node, WLR_SCENE_NODE_ROOT, NULL);
-
+	wl_list_init(&scene->outputs);
 	return scene;
 }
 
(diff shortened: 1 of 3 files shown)


types/wlr_scene.c
```diff
@@ -0,0 +1,536 @@
...
+	struct wlr_surface *surface = scene_surface->surface;
+
+	if (!pixman_region32_not_empty(&surface->buffer_damage)) {
+		return;
+	}
+
+	struct wlr_scene *scene = scene_node_get_root(&scene_surface->node);
+
+	int lx, ly;
+	wlr_scene_node_coords(&scene_surface->node, &lx, &ly);
```
emersion, line 125: We can skip the damage step if any of the parent nodes is disabled.

types/wlr_scene.c
```diff
@@ -0,0 +1,536 @@
...
+}
+
+static void scene_node_damage_whole(struct wlr_scene_node *node) {
+	struct wlr_scene *scene = scene_node_get_root(node);
+	if (wl_list_empty(&scene->outputs)) {
+		return;
+	}
+
+	int lx, ly;
+	wlr_scene_node_coords(node, &lx, &ly);
```
emersion, line 228: Same here: if any of the parent nodes is disabled, we can skip the damage step.

emersion:
This is ready for review.

types/wlr_scene.c
```diff
@@ -105,16 +163,88 @@ struct wlr_scene_surface *wlr_scene_surface_create(struct wlr_scene_node *parent
...
+	}
+
+	_scene_node_damage_whole(node, scene, lx, ly);
+}
+
 void wlr_scene_node_set_enabled(struct wlr_scene_node *node, bool enabled) {
+	scene_node_damage_whole(node);
 	node->state.enabled = enabled;
+	scene_node_damage_whole(node);
 }
```
ifreund, line 241: This will damage the entire subtree even if the enabled state isn't changed, which I think should be avoided.

I also was very confused by the need for two `scene_node_damage_whole()` calls at first glance, perhaps that's worth a brief comment saying that one will short-circuit.

types/wlr_scene.c
```diff
@@ -105,16 +163,88 @@ struct wlr_scene_surface *wlr_scene_surface_create(struct wlr_scene_node *parent
...
 
 void wlr_scene_node_set_position(struct wlr_scene_node *node, int x, int y) {
+	scene_node_damage_whole(node);
 	node->state.x = x;
 	node->state.y = y;
+	scene_node_damage_whole(node);
 }
```
ifreund, line 248: same here, we still damage even if the position hasn't changed.

ifreund (commented):
I wonder if any compositor that doesn't already use wlr_output_layout will use the scene graph API. It seems like some integration there could save almost every compositor some boilerplate.

types/wlr_scene.c
emersion: Good point! Added an early return to the functions you've mentioned, plus place_above/place_below.

emersion:
> I wonder if any compositor that doesn't already use wlr_output_layout will use the scene graph API. It seems like some integration there could save almost every compositor some boilerplate.

Yes, likely, but I'd prefer to keep this as a separate patch (feel free to submit a PR built on top of this one!). One still unanswered question is how we should plug existing wlroots modules to the scene-graph. Should the scene-graph add support for `wlr_output_layout`, or should `wlr_output_layout` add optional support for the scene-graph? In other words, should we have `wlr_scene_create_with_output_layout(output_layout)` or `wlr_output_layout_create_with_scene(scene)`? (API subject to be tweaked a bit to allow better extensibility.)

ammen99:
> wlr_output_layout_create_with_scene(scene)

Would this support `NULL` scene? In Wayfire, I'd like to use `wlr_output_layout` but not scene for obvious reasons ..

emersion:
Yes, of course. output-layout and the scene-graph will both stay optional and compositors will still be able to use one without the other.

emersion:
Rebased on top of @djpohly's https://github.com/swaywm/wlroots/pull/3140.

types/wlr_scene.c
```diff
@@ -132,41 +195,142 @@ struct wlr_scene_rect *wlr_scene_rect_create(struct wlr_scene_node *parent,
 	scene_rect->height = height;
 	memcpy(scene_rect->color, color, sizeof(scene_rect->color));
 
+	scene_node_damage_whole(&scene_rect->node);
+
 	return scene_rect;
 }
 
 void wlr_scene_rect_set_size(struct wlr_scene_rect *rect, int width, int height) {
```
djpohly, line 203: Now that this adds damage, should it have an early return for no-ops like `wlr_scene_node_set_position` does?

djpohly:
Something is off with wlr_scene_output positioning. Here's the scene-graph example running Alacritty:
2021-09-06-000243_958x519_scrot
I added the line
```c
wlr_scene_output_set_position(output->scene_output, 30, 10);
```
and - if I had to guess - it looks like the damage rectangles are correctly translated with respect to the viewport, but the rendered surfaces are missing a pair of minus signs somewhere.:
2021-09-06-000105_958x519_scrot

types/wlr_scene.c
emersion: Good point, done!

emersion:
> Something is off with wlr_scene_output positioning

Indeed, must've missed something in a refactoring. Fixed in "scene: fix wlr_scene_render_output offset".

djpohly:
Here is a [MWE for another issue I'm seeing](https://gist.github.com/djpohly/85eefaf9e884d005953c25bc3f42e063). Run on DRM backend, move the (invisible) mouse cursor to generate redraws, notice the flickering rectangle, and click to exit.

I'm still not very confident in my understanding of double-buffering and damage, but it looks like... maybe existing bits of scene aren't ending up being rendered to one of the buffers when a new output is created?  I'm not actually sure whether the problem is specific to the scene-graph code.

emersion:
Indeed, what you're seeing is a `wlr_output_damage` issue, fixed here: https://github.com/swaywm/wlroots/pull/3175

djpohly:
Not sure whether this belongs here or there, but with #3175 applied, I still get the behavior when the [scene_output is created after the initial output_commit (MWE)](https://gist.github.com/djpohly/28ba17f484e874ffda5a4184e2bc48ed).

djpohly:
It's possible that I missed that push. I'll give it a full recheck and report back.

djpohly:
Guess this one was PEBKAC - sorry about that! (I'm still trying to track down another issue with what looks like the wrong regions being damaged on my second monitor, but I don't have anything too useful yet.)

types/wlr_scene.c
```diff
@@ -426,10 +608,114 @@ void wlr_scene_render_output(struct wlr_scene *scene, struct wlr_output *output,
...
+		return true;
+	}
+
+	wlr_renderer_begin(renderer, output->width, output->height);
+
+	int nrects;
+	pixman_box32_t *rects = pixman_region32_rectangles(&damage, &nrects);
+	for (int i = 0; i < nrects; ++i) {
+		scissor_output(output, &rects[i]);
+		wlr_renderer_clear(renderer, (float[4]){ 0.0, 0.0, 0.0, 1.0 });
```
bl4ckb0ne, line 697: Shouldnt the clear color be user defined?

vyivel: Eh, I don't see the need for it. If the caller really wants to set background color for whatever reason, they could create a solid color rectangle node, but again, I don't see the need for it.

emersion: Yeah, to customize it users should just create a rect node with the color they want.

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM!

PR #3119, opened 2021-08-17
Handle output enabled in wlr_output_send_frame

Emantor:
Alternative to #3118. We now check if the output is enabled before emitting the frame event in `wlr_output_send_frame` and revert the commit within the DRM backend which added a similar check but forgot to reset `frame_pending` to false.
 
Closes #3118


emersion (approved):
LGTM, thanks!

PR #3120, opened 2021-08-18
subsurface: move parent link to state

emersion:
Move the wlr_subsurface parent link to the subsurface state.

This is a dumb find/replace operation. This shouldn't result in
any behavior change.

* * *

We need to be a bit careful here, the `wlr_subsurface_state` is actually an "add-on" for the parent wl_surface's state. It's not applied on child commit, it's applied on parent commit.

```
      Sub-surfaces also have another kind of state, which is managed by
      wl_subsurface requests, as opposed to wl_surface requests. This
      state includes the sub-surface position relative to the parent
      surface (wl_subsurface.set_position), and the stacking order of
      the parent and its sub-surfaces (wl_subsurface.place_above and
      .place_below). This state is applied when the parent surface's
      wl_surface state is applied, regardless of the sub-surface's mode.
      As the exception, set_sync and set_desync are effective immediately.
```

cc @vyivel

### DIFF
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -170,6 +170,7 @@ struct wlr_surface {
  */
 struct wlr_subsurface_parent_state {
 	int32_t x, y;
+	struct wl_list link;
 };
 
 struct wlr_subsurface {
@@ -185,9 +186,6 @@ struct wlr_subsurface {
 	bool synchronized;
 	bool reordered;
 	bool mapped;
-
-	struct wl_list parent_link;
-	struct wl_list parent_pending_link;
 
 	struct wl_listener surface_destroy;
 	struct wl_listener parent_destroy;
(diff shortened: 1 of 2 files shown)


vyivel:
>the wlr_subsurface_state is actually an "add-on" for the parent wl_surface's state

That looks a bit confusing, considering that other state addons belong to the base surface state. I'd rename it to `wlr_subsurface_parent_state`, which later will have `addon` with the child subsurface as owner.

emersion:
Indeed, sounds like a good idea to me. Added a commit which does that, and also introduces a small comment to explain why it's special.

vyivel:
LGTM, but I'd like to leave this unmerged for now, until a PR for actual subsurface state caching is written.

vyivel:
https://github.com/swaywm/wlroots/pull/3125 took another approach, so I suppose this is not required anymore.

emersion:
Merged as part of https://github.com/swaywm/wlroots/pull/3169

PR #3121, opened 2021-08-18
surface: move frame callbacks out of state

vyivel:
As frame callbacks aren't cached anyway, they can be handled separately
from the state logic.


emersion:
The original change was reverted.

PR #3122, opened 2021-08-19
surface: cache frame callback lists again

vyivel:
Caching frame callback lists is actually the correct behavior, because
if a surface is locked because of e.g. subsurface synchronization,
clients would expect to receive frame done events only after the
pending state is actually committed.

---

If surface locking is going to be used for atomic layout updates by compositors, we would probably need to distinguish between locks required by protocols and locks made by compositors.

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -313,6 +313,11 @@ static void surface_state_move(struct wlr_surface_state *state,
 	if (next->committed & WLR_SURFACE_STATE_VIEWPORT) {
 		memcpy(&state->viewport, &next->viewport, sizeof(state->viewport));
 	}
+	if (next->committed & WLR_SURFACE_STATE_FRAME_CALLBACK_LIST) {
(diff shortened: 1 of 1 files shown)


emersion:
Can you expand on the motivation? Where in the protocol is this stated?

vyivel:
Hmm.

>The frame request will take effect on the next wl_surface.commit.

I've checked the protocol once again and apparently frame callbacks aren't part of the surface state at all?

As for the subsurfaces, synced ones probably don't care about frame callbacks anyway.

On the other hand, this seems to match Weston behavior IIRC.

BTW, there is a PR which moves callbacks out of state: https://github.com/swaywm/wlroots/pull/3121

emersion:
https://gitlab.freedesktop.org/wayland/wayland/-/issues/229

emersion:
Merged as part of https://github.com/swaywm/wlroots/pull/3169

PR #3124, opened 2021-08-19
xwayland: Allow to retrieve _NET_STARTUP_ID

agx:
This is use for startup notifications per startup-notifiation spec

https://specifications.freedesktop.org/startup-notification-spec/startup-notification-latest.txt


emersion:
Should we also add this atom to the list in `xwm_map_shell_surface`?

emersion:
Do you have an example usage in a compositor? What else does the compositor need to do?

Can it take advantage of the xdg-activation architecture for instance?

agx:
@emersion I'm currently using it with `gtk_shell1` (https://gitlab.gnome.org/World/Phosh/phoc/-/merge_requests/259/diffs). It's not the full protocol (which requires you to track `_NET_STARTUP_INFO_BEGIN`/`_NET_STARTUP_INFO` to the
root window too) but enough to track that the surface showed up . The token value comes via the `DESKTOP_STARTUP_ID` (mentioned in https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html).  I intend to expand on the root window bits later as follow up but we need to track changes to `_NET_STARTUP_ID` per spec anyway.

Regarding xdg-activation: are there any bits for x11 at all yet? I saw the wayland protocol (which is great) and the mention in the notification spec.

I've added it to `xwm_map_shell_surface` too since that makes sense.

emersion:
>  I'm currently using it with gtk_shell1

Ah. Would be much better to port GTK to xdg-activation, but oh well…

> It's not the full protocol (which requires you to track _NET_STARTUP_INFO_BEGIN/_NET_STARTUP_INFO to the
root window too) but enough to track that the surface showed up

Right. The root window messages don't strike me as very useful if used by the launchee. Or are these mainly used by the launcher client?

> Regarding xdg-activation: are there any bits for x11 at all yet?

I think it would be possible to re-use some bits here, or at least make it possible for compositors to re-use some bits. For instance generating new `DESKTOP_STARTUP_ID` tokens and storing them with a timeout. This could also be useful when the compositor starts a client on its own. But I guess it's pretty unrelated to this PR, just food for thought. :)

emersion (approved):
Thanks!

agx:
> Ah. Would be much better to port GTK to xdg-activation, but oh well…

GTK4 does that already but for GTK3 we need `gtk_shell1`. xdg-activiation is on the list once the other bits are in place since we I hope we get consistency with Qt then too.

agx:
Thanks for merging!

agx:
> Right. The root window messages don't strike me as very useful if used by the launchee. Or are these mainly used by the launcher client?

Both: "new" and "change" messages are sent by the launcher code and the launchee sends "remove" so it would make sense to track these. My goal would be to be able to have the compositor in a position where it can inform interested clients (panels/desktop shells)  about a clients startup sequence using either the above for xwayland or `xdg-activation` for wayland since the only the compositor gets informed about those. That could well live within wlroots if that makes sense.

emersion:
I see. Yeah, I think it makes sense for at least the root window message parsing parts to be handled by wlroots.

PR #3128, opened 2021-08-20
scene: add support for sub-surfaces

emersion:
Representing each sub-surface as a scene-graph node allows us to remove a bunch of special cases. Damage tracking for instance works without any special casing for sub-surfaces.

~~Depends on: https://github.com/swaywm/wlroots/pull/1966~~

### DIFF
--- a/include/wlr/types/wlr_scene.h
+++ b/include/wlr/types/wlr_scene.h
@@ -224,4 +224,11 @@ bool wlr_scene_output_commit(struct wlr_scene_output *scene_output);
 bool wlr_scene_attach_output_layout(struct wlr_scene *scene,
 	struct wlr_output_layout *output_layout);
 
+/**
+ * Add a node displaying a surface and all of its sub-surfaces to the
+ * scene-graph.
+ */
+struct wlr_scene_node *wlr_scene_subsurface_tree_create(
+	struct wlr_scene_node *parent, struct wlr_surface *surface);
+
 #endif
--- a/include/wlr/types/wlr_scene.h
+++ b/include/wlr/types/wlr_scene.h
@@ -28,6 +28,7 @@ struct wlr_output_layout;
 
 enum wlr_scene_node_type {
 	WLR_SCENE_NODE_ROOT,
+	WLR_SCENE_NODE_TREE,
 	WLR_SCENE_NODE_SURFACE,
 	WLR_SCENE_NODE_RECT,
 };
@@ -59,6 +60,11 @@ struct wlr_scene {
 	struct wlr_scene_node node;
 
 	struct wl_list outputs; // wlr_scene_output.link
+};
+
+/** A sub-tree in the scene-graph. */
+struct wlr_scene_tree {
+	struct wlr_scene_node node;
 };
 
 /** A scene-graph node displaying a single surface. */
@@ -156,6 +162,11 @@ struct wlr_scene *wlr_scene_create(void);
  */
 void wlr_scene_render_output(struct wlr_scene *scene, struct wlr_output *output,
 	int lx, int ly, pixman_region32_t *damage);
+
+/**
+ * Add a node displaying nothing but its children.
+ */
+struct wlr_scene_tree *wlr_scene_tree_create(struct wlr_scene_node *parent);
 
 /**
  * Add a node displaying a single surface to the scene-graph.
--- a/types/scene/subsurface_tree.c
+++ b/types/scene/subsurface_tree.c
@@ -1,0 +1,218 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <wlr/types/wlr_scene.h>
+#include <wlr/util/addon.h>
+
+/**
+ * A tree for a surface and all of its child sub-surfaces.
+ *
+ * `tree` contains `scene_surface` and one node per sub-surface.
+ */
+struct wlr_scene_subsurface_tree {
+	struct wlr_scene_tree *tree;
+	struct wlr_surface *surface;
+	struct wlr_scene_surface *scene_surface;
+
+	struct wlr_scene_subsurface_tree *parent; // NULL for the top-level surface
+	struct wlr_addon surface_addon; // only set if there's a parent
+
+	struct wl_listener tree_destroy;
+	struct wl_listener surface_destroy;
+	struct wl_listener surface_commit;
+	struct wl_listener surface_new_subsurface;
+};
+
+static void subsurface_tree_destroy(struct wlr_scene_subsurface_tree *subsurface_tree) {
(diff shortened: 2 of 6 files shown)


emersion:
This is now ready for review.

emersion:
Rebased on top of master and applied https://github.com/swaywm/wlroots/pull/3160#issuecomment-915876046.

types/scene/surface_tree.c
```diff
@@ -0,0 +1,206 @@
...
+}
+
+static void surface_tree_addon_destroy(struct wlr_addon *addon) {
+	struct wlr_scene_surface_tree *surface_tree =
+		wl_container_of(addon, surface_tree, surface_addon);
+	wlr_scene_node_destroy(&surface_tree->tree->node);
+}
+
+static const struct wlr_addon_interface surface_tree_addon = {
+	.name = "wlr_scene_surface_tree",
```
ifreund, line 116: Not really related to this PR, but why does this `name` field exist? Just to make poking around at `wlr_addon_set` with a debugger easier?

types/scene/surface_tree.c
```diff
@@ -0,0 +1,206 @@
...
+
+	struct wlr_subsurface *subsurface;
+	wl_list_for_each(subsurface, &surface->current.subsurfaces_below,
+			current.link) {
+		surface_tree_create_subsurface(surface_tree, subsurface);
+	}
+	wl_list_for_each(subsurface, &surface->current.subsurfaces_above,
+			current.link) {
+		surface_tree_create_subsurface(surface_tree, subsurface);
+	}
```
ifreund, line 176: Hmm, bubbling up the OOM here would be a massive pain, especially without leaking memory, but silently ignoring subtrees if allocation fails doesn't feel good either.

ifreund (approved):
Everything here seems reasonable, I think this is mergeable as-is but I left a couple of comments.

include/wlr/types/wlr_scene.h
```diff
@@ -213,4 +224,11 @@ bool wlr_scene_output_commit(struct wlr_scene_output *scene_output);
 bool wlr_scene_attach_output_layout(struct wlr_scene *scene,
 	struct wlr_output_layout *output_layout);
 
+/**
+ * Add a node displaying a surface and all of its sub-surfaces to the
+ * scene-graph.
+ */
+struct wlr_scene_node *wlr_scene_surface_tree_create(
```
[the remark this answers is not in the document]
emersion: Sounds good!

types/scene/surface_tree.c
emersion: Yeah, it's mainly for making it easier to figure out what an addon represents when debugging (addons are kind of opaque).

types/scene/surface_tree.c
emersion: Hm yeah it's a bit annoying, especially since we can't easily iterate over the stuff we allocate. Pushed a commit to bubble up the OOM without cleaning up everything.

types/scene/subsurface_tree.c
```diff
@@ -0,0 +1,215 @@
...
+	struct wlr_scene_subsurface_tree *parent; // NULL for the top-level surface
+	struct wlr_addon surface_addon; // only set if there's a parent
+
+	struct wl_listener tree_destroy;
+	struct wl_listener surface_destroy;
+	struct wl_listener surface_commit;
+	struct wl_listener surface_new_subsurface;
+};
+
+static void surface_tree_destroy(struct wlr_scene_subsurface_tree *surface_tree) {
```
ifreund, line 25: I think this file could use a `s/surface_tree/subsurface_tree/g` after the rename of the struct.

types/scene/subsurface_tree.c
```diff
@@ -0,0 +1,215 @@
...
+
+	return true;
+}
+
+static void surface_tree_handle_surface_new_subsurface(
+		struct wl_listener *listener, void *data) {
+	struct wlr_scene_subsurface_tree *surface_tree =
+		wl_container_of(listener, surface_tree, surface_new_subsurface);
+	struct wlr_subsurface *subsurface = data;
+	surface_tree_create_subsurface(surface_tree, subsurface);
```
ifreund, line 144: This is the one remaining place I see where allocation failure isn't handled.

types/scene/surface_tree.c
ifreund: Hmm, I'm looking at this more closely and I think your current code does clean up everything correctly in fact. The `wlr_scene_node_destroy()` call will trigger any successfully allocated subtrees to be recursively destroyed.

types/scene/subsurface_tree.c
emersion: Indeed, added a `wl_resource_post_no_memory` call.

emersion:
> I think this file could use a s/surface_tree/subsurface_tree/g after the rename of the struct.

Done!

> Hmm, I'm looking at this more closely and I think your revised code does clean up everything correctly in fact. The wlr_scene_node_destroy() call will trigger any successfully allocated subtrees to be recursively destroyed.

Oh right, godo catch. Well I guess for once chance is on our side. :P

types/scene/subsurface_tree.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+static void subsurface_tree_handle_surface_destroy(struct wl_listener *listener,
+		void *data) {
+	struct wlr_scene_subsurface_tree *subsurface_tree =
+		wl_container_of(listener, subsurface_tree, surface_destroy);
+	wlr_scene_node_destroy(&subsurface_tree->tree->node);
+}
+
+static const struct wlr_addon_interface subsurface_tree_addon;
```
vyivel, line 51: Nit: every other use of `wlr_addon_interface` has `*_addon_impl` name pattern AFAIK

vyivel:
Having `wlr_scene_tree` is kinda weird as every node can have children and therefore be tree. Maybe remove `wlr_scene_node_state.children` and make it that only `wlr_scene_tree` and `wlr_scene_root` can have child nodes?

emersion:
Or maybe make it so `wlr_scene` is a node with a single children of type `wlr_scene_tree` which represents the root? That way only `wlr_scene_tree` would have children?

In any case I'd prefer to do this in a separate PR, since `wlr_scene_node_state.children` is already merged.

types/scene/subsurface_tree.c
emersion: Fixed!

vyivel:
Yeah, that would work too.
As for this PR, LGTM!

vyivel (approved):

emersion:
Thank you both for the reviews!

Issue #2916, opened 2021-05-05
render/drm_dumb_allocator: doesn't work on amdgpu


emersion:
The GBM imports are failing. See https://github.com/swaywm/wlroots/pull/2700#issuecomment-832508102.

Fixing this requires side-stepping GBM.

We probably have two choices here:

- Pass GEM handles around directly. In theory this is simpler, but also more dangerous if there's no safe-guard. In particular, on multi-GPU systems, it's pretty easy to mix up GEM handles between devices.
- Open a separate DRM FD in the DRM backend. Manually perform the GEM handle ref'counting.

sjnewbury:
I'm getting severe corruption on fullscreen windows, for example alacritty with "super-f".  I've bisected it to this commit.

Using current git master/main libdrm/mesa/wlroots/sway, amdgpu (POLARIS10).

emersion:
Please open a new bug report with debug logs and dmesg.

PR #3131, opened 2021-08-23
backend/drm: introduce wlr_drm_bo_handle_table

emersion:
Using GBM to import DRM dumb buffers tends to not work well. By
using GBM we're calling some driver-specific functions in Mesa.
These functions check whether Mesa can work with the buffer.
Sometimes Mesa has requirements which differ from DRM dumb buffers
and the GBM import will fail (e.g. on amdgpu).

Instead, drop GBM and use drmPrimeFDToHandle directly. But there's
a twist: BO handles are not ref'counted by the kernel and need to
be ref'counted in user-space [1]. libdrm usually performs this
bookkeeping and is used under-the-hood by Mesa.

We can't re-use libdrm for this task without using driver-specific
APIs. So let's just re-implement the ref'counting logic in wlroots.
The wlroots implementation is inspired from amdgpu's in libdrm [2].

Closes: #2916

[1]: https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/110
[2]: https://gitlab.freedesktop.org/mesa/drm/-/blob/1a4c0ec9aea13211997f982715fe5ffcf19dd067/amdgpu/handle_table.c

### DIFF
--- a/backend/drm/bo_handle_table.c
+++ b/backend/drm/bo_handle_table.c
@@ -1,0 +1,43 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <wlr/util/log.h>
+#include "backend/drm/bo_handle_table.h"
+
+static size_t align(size_t val, size_t align) {
+	size_t mask = align - 1;
+	return (val + mask) & ~mask;
+}
+
+void drm_bo_handle_table_finish(struct wlr_drm_bo_handle_table *table) {
+	free(table->nrefs);
+}
+
+bool drm_bo_handle_table_ref(struct wlr_drm_bo_handle_table *table,
+		uint32_t handle) {
+	assert(handle != 0);
+
+	if (handle > table->len) {
+		// Grow linearily, because we don't expect the number of BOs to explode
+		size_t len = align(handle + 1, 512);
+		size_t *nrefs = realloc(table->nrefs, len * sizeof(size_t));
+		if (nrefs == NULL) {
+			wlr_log_errno(WLR_ERROR, "realloc failed");
+			return false;
+		}
+		memset(&nrefs[table->len], 0, (len - table->len) * sizeof(size_t));
+		table->len = len;
+		table->nrefs = nrefs;
+	}
+
+	table->nrefs[handle]++;
+	return true;
+}
+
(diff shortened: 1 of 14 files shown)


backend/drm/bo_handle_table.c
```diff
@@ -0,0 +1,43 @@
...
+	free(table->nrefs);
+}
+
+bool drm_bo_handle_table_ref(struct wlr_drm_bo_handle_table *table,
+		uint32_t handle) {
+	assert(handle != 0);
+
+	if (handle > table->len) {
+		// Grow linearily, because we don't expect the number of BOs to explode
+		size_t len = align(handle + 1, 512);
```
bl4ckb0ne, line 22: Why 512?

emersion: Arbitrary number. This happens to be `4096 / 8`, which is what amdgpu is using.

include/backend/drm/bo_handle_table.h
```diff
@@ -0,0 +1,21 @@
+#ifndef BACKEND_DRM_BO_HANDLE_TABLE_H
+#define BACKEND_DRM_BO_HANDLE_TABLE_H
+
+/**
+ * Table performing reference counting for buffer object handles.
```
bl4ckb0ne, line 5: Would be worth mentionning the [inspiration](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/amdgpu/handle_table.c)

backend/drm/bo_handle_table.c
bl4ckb0ne: Could the value of a memory page size change depending on the arch or bitness?

emersion: Maybe, but it's not worth the effort. Growing by a reasonable amount of memory should be enough. Using the page table size (as amdgpu does) is micro-optimization IMHO.

bl4ckb0ne: Sounds good

include/backend/drm/bo_handle_table.h
emersion: Ack, added a link

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks :)

PR #3132, opened 2021-08-23
Viewport+transform fix

vyivel:
Partially fixes https://github.com/swaywm/sway/issues/6322

### DIFF
--- a/include/wlr/util/box.h
+++ b/include/wlr/util/box.h
@@ -78,4 +78,9 @@ bool wlr_box_empty(const struct wlr_box *box);
 void wlr_box_transform(struct wlr_box *dest, const struct wlr_box *box,
 	enum wl_output_transform transform, int width, int height);
 
+/**
+ * Transforms a floating-point box inside a (0, 0, width, height) box.
+ */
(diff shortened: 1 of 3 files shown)


include/wlr/util/box.h
```diff
@@ -78,4 +78,9 @@ bool wlr_box_empty(const struct wlr_box *box);
 void wlr_box_transform(struct wlr_box *dest, const struct wlr_box *box,
 	enum wl_output_transform transform, int width, int height);
 
+/**
+ * Transforms a floating-point box inside a (0, 0, width, height) box.
+ */
+void wlr_fbox_transform(struct wlr_fbox *dest, const struct wlr_fbox *box,
+	enum wl_output_transform transform, int width, int height);
```
emersion, line 85: We should probably turn `width` and `height` into `double` here.

emersion (commented):
Apart from this nit, LGTM!

include/wlr/util/box.h
vyivel: Makes sense. Done.

emersion (approved):
Thanks for the fix!

PR #3133, opened 2021-08-25
Move allocator stuff into new directory

emersion:
Add render/allocator/ and include/render/allocator/ to hold
everything allocator-related.


bl4ckb0ne:
Would it also be a good place to move all that is buffer related as well?

emersion:
I don't think so, buffers can exist without allocators, and existed before allocators were introduced.

bl4ckb0ne:
LGTM then

bl4ckb0ne (approved):

PR #3136, opened 2021-08-25
xdg-activation: Pass on activation id when we don't find a token

agx:
If the activation happened via another protocol (such as startup
notification protocol,  gtk_shell1 or other compositor specific
mechanisms the compositor might know about the id but don't have a valid
token. Just pass on the id in this case so it can match up things against
it's token list.

See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/xdg-activation/x11-interoperation.rst


emersion:
Maybe a better approach would be to allow compositors to generate fresh new xdg-activation tokens.

emersion:
Has this been superseded by https://github.com/swaywm/wlroots/pull/3146?

emersion:
Please re-open if necessary.

PR #3140, opened 2021-08-26
scene: add RECT node type

djpohly:
Adds a new scene-graph node type for rendering solid-colored rectangles, which may be useful for simple window decorations. A little reworking was needed in the scene-graph implementation, since it would break the assumption that every rendered node has an associated surface.

Thoughts/questions:
1. Should there be a `wlr_scene_rect_set_size()` and/or `wlr_scene_rect_set_color()` function, or is it straghtforward enough to change the fields directly? (Maybe damage tracking considerations tip the scale toward functions?)
1. How should RECTs interact with `wlr_scene_node_surface_at()`?
1. Would it complicate the scene-graph example too much if I were to add a handler for wlr_surface.commit? It would allow us to get the surface dimensions and draw something less ugly than what I currently have.
1. Is there a point at which it makes more sense to implement additional node types with a `scene_node_impl` rather than switching on a type field?

### DIFF
--- a/examples/scene-graph.c
+++ b/examples/scene-graph.c
@@ -20,6 +20,8 @@
  *
  * New surfaces are stacked on top of the existing ones as they appear. */
 
+static const int border_width = 3;
+
 struct server {
 	struct wl_display *display;
 	struct wlr_backend *backend;
@@ -35,8 +37,10 @@ struct server {
 struct surface {
 	struct wlr_surface *wlr;
 	struct wlr_scene_surface *scene_surface;
+	struct wlr_scene_rect *border;
 	struct wl_list link;
 
+	struct wl_listener commit;
 	struct wl_listener destroy;
 };
 
@@ -102,9 +106,17 @@ static void server_handle_new_output(struct wl_listener *listener, void *data) {
 	wlr_output_create_global(wlr_output);
 }
 
+static void surface_handle_commit(struct wl_listener *listener, void *data) {
+	struct surface *surface = wl_container_of(listener, surface, commit);
+	wlr_scene_rect_set_size(surface->border,
+			surface->wlr->current.width + 2 * border_width,
+			surface->wlr->current.height + 2 * border_width);
+}
+
 static void surface_handle_destroy(struct wl_listener *listener, void *data) {
 	struct surface *surface = wl_container_of(listener, surface, destroy);
 	wlr_scene_node_destroy(&surface->scene_surface->node);
+	wlr_scene_node_destroy(&surface->border->node);
 	wl_list_remove(&surface->destroy.link);
 	wl_list_remove(&surface->link);
 	free(surface);
@@ -119,13 +131,22 @@ static void server_handle_new_surface(struct wl_listener *listener,
 
 	struct surface *surface = calloc(1, sizeof(struct surface));
 	surface->wlr = wlr_surface;
+	surface->commit.notify = surface_handle_commit;
+	wl_signal_add(&wlr_surface->events.commit, &surface->commit);
 	surface->destroy.notify = surface_handle_destroy;
 	wl_signal_add(&wlr_surface->events.destroy, &surface->destroy);
+
+	/* Border dimensions will be set in surface.commit handler */
+	surface->border = wlr_scene_rect_create(&server->scene->node,
+			0, 0, (float[4]){ 0.5f, 0.5f, 0.5f, 1 });
+	wlr_scene_node_set_position(&surface->border->node, pos, pos);
+
 	surface->scene_surface =
 		wlr_scene_surface_create(&server->scene->node, wlr_surface);
 	wl_list_insert(server->surfaces.prev, &surface->link);
 
-	wlr_scene_node_set_position(&surface->scene_surface->node, pos, pos);
+	wlr_scene_node_set_position(&surface->scene_surface->node,
+			pos + border_width, pos + border_width);
 }
 
 int main(int argc, char *argv[]) {
--- a/types/wlr_scene.c
+++ b/types/wlr_scene.c
@@ -1,5 +1,6 @@
 #include <assert.h>
 #include <stdlib.h>
+#include <string.h>
 #include <wlr/backend.h>
 #include <wlr/render/wlr_renderer.h>
 #include <wlr/types/wlr_matrix.h>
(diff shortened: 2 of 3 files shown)


djpohly:
Added setters, updated the example, and created a `wlr_scene_node_at` function.

I went ahead and removed `wlr_scene_node_surface_at` after thinking about it a bit. It could either treat rects as opaque and return NULL even if there are surfaces at the given coordinates further down the tree, or it could treat them as transparent and continue searching, in which case you couldn't tell whether the returned surface was obscured or visible. Given a scene_surface node, it's easy enough to retrieve the associated surface, so I'd contend that the new function subsumes existing functionality.

examples/scene-graph.c
```diff
@@ -102,9 +104,16 @@ static void server_handle_new_output(struct wl_listener *listener, void *data) {
 	wlr_output_create_global(wlr_output);
 }
 
+static void surface_handle_commit(struct wl_listener *listener, void *data) {
+	struct surface *surface = wl_container_of(listener, surface, commit);
+	wlr_scene_rect_set_size(surface->border, surface->wlr->current.width + 6,
```
emersion, line 109: Maybe add a constant for the border size, to avoid magic numbers?

types/wlr_scene.c
```diff
@@ -108,6 +118,35 @@ struct wlr_scene_surface *wlr_scene_surface_create(struct wlr_scene_node *parent
...
+	if (scene_rect == NULL) {
+		return NULL;
+	}
+	scene_node_init(&scene_rect->node, WLR_SCENE_NODE_RECT, parent);
+
+	scene_rect->width = width;
+	scene_rect->height = height;
+	for (int i = 0; i < 4; ++i) {
+		scene_rect->color[i] = color[i];
+	}
```
emersion, line 134: Nit: this loop can be replaced with a `memcpy`

types/wlr_scene.c
```diff
@@ -108,6 +118,35 @@ struct wlr_scene_surface *wlr_scene_surface_create(struct wlr_scene_node *parent
...
+
+void wlr_scene_rect_set_size(struct wlr_scene_rect *rect, int width, int height) {
+	rect->width = width;
+	rect->height = height;
+}
+
+void wlr_scene_rect_set_color(struct wlr_scene_rect *rect, const float color[static 4]) {
+	for (int i = 0; i < 4; ++i) {
+		rect->color[i] = color[i];
+	}
```
emersion, line 147: Ditto

types/wlr_scene.c
```diff
@@ -226,6 +276,27 @@ static void scissor_output(struct wlr_output *output, pixman_box32_t *rect) {
...
+	pixman_region32_t damage;
+	pixman_region32_init(&damage);
+	pixman_region32_init_rect(&damage, box->x, box->y, box->width, box->height);
+	pixman_region32_intersect(&damage, &damage, output_damage);
+	if (pixman_region32_not_empty(&damage)) {
+		int nrects;
+		pixman_box32_t *rects = pixman_region32_rectangles(&damage, &nrects);
+		for (int i = 0; i < nrects; ++i) {
+			scissor_output(output, &rects[i]);
+			wlr_render_rect(renderer, box, color, matrix);
```
emersion, line 294: Technically we can just use clear() here

emersion: If we do that, we should be able to remove the `matrix` arg.

vyivel: clear() as in `wlr_renderer_clear()`? Won't that cause issues with semi-transparent RECT nodes?

emersion: Derp. Indeed, we need to always use`wlr_render_rect`.

types/wlr_scene.c
```diff
@@ -171,23 +210,34 @@ struct wlr_surface *wlr_scene_node_surface_at(struct wlr_scene_node *node,
...
-				*sx = lx;
+			if (nx != NULL) {
+				*nx = lx;
+			}
+			if (ny != NULL) {
+				*ny = ly;
+			}
+			return &scene_surface->node;
+		}
+	} else if (node->type == WLR_SCENE_NODE_RECT) {
```
emersion, line 231: Style nit: we can use a `switch` here

types/wlr_scene.c
```diff
@@ -226,6 +276,27 @@ static void scissor_output(struct wlr_output *output, pixman_box32_t *rect) {
...
+		pixman_region32_t *output_damage, const float color[static 4],
+		const struct wlr_box *box, const float matrix[static 9]) {
+	struct wlr_renderer *renderer = wlr_backend_get_renderer(output->backend);
+	assert(renderer);
+
+	pixman_region32_t damage;
+	pixman_region32_init(&damage);
+	pixman_region32_init_rect(&damage, box->x, box->y, box->width, box->height);
+	pixman_region32_intersect(&damage, &damage, output_damage);
+	if (pixman_region32_not_empty(&damage)) {
```
emersion, line 289: Technically, no need for this condition. If the region is empty the `for` loop will be a no-op, and the `if` doesn't avoid intermediate computations.

emersion (commented):
Overall this looks pretty good!

I'm still wondering about `wlr_scene_node_surface_at`: it doesn't seem like compositors can re-implement it with `wlr_scene_node_surface_at`, or at least not without making `scene_surface_from_node` public.

types/wlr_scene.c
djpohly: Does that mean it can be removed from the existing `render_texture()` as well?

emersion: Very likely yes! Feel free to add a separate commit to do this, either in this PR or in a separate one.

djpohly:
Ah, you're completely right. Somehow in my mind `scene_surface_from_node` was already public. Fixed.

emersion:
Looks good to me!

Can you squash the commits into logical commits? See https://github.com/swaywm/wlroots/blob/master/CONTRIBUTING.md#commit-log

e.g. "scene: use memcpy to copy color arrays" can be squashed into "scene: add RECT node type"

djpohly:
I left the simplification to `render_texture()` as a separate commit since it wasn't actually part of implementing rects - let me know if it'd be preferable to have that merged with one of the others.

emersion (approved):
This is perfect, thanks!

PR #3141, opened 2021-08-26
scene: add node reparent function

djpohly:
If nodes are arranged in a tree rather than at a single level (e.g. if #3128 is merged), then it makes sense to have a way to move them to a completely different parent in addition to moving up or down among siblings.

### DIFF
--- a/types/wlr_scene.c
+++ b/types/wlr_scene.c
@@ -133,6 +133,20 @@ void wlr_scene_node_place_below(struct wlr_scene_node *node,
 	wl_list_insert(sibling->state.link.prev, &node->state.link);
 }
 
+void wlr_scene_node_reparent(struct wlr_scene_node *node,
+		struct wlr_scene_node *new_parent) {
+	if (node->parent == new_parent) {
+		return;
+	}
+
+	wl_list_remove(&node->state.link);
+
+	node->parent = new_parent;
+	if (new_parent != NULL) {
+		wl_list_insert(new_parent->state.children.prev, &node->state.link);
+	}
+}
+
 static void scene_node_for_each_surface(struct wlr_scene_node *node,
 		int lx, int ly, wlr_surface_iterator_func_t user_iterator,
 		void *user_data) {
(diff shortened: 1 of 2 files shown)


emersion:
Without this, compositors have to teardown and rebuild the tree from scratch. I think it makes sense to make this operation easier for compositors.

It would be nice to make sure bad things don't happen, e.g. it's impossible to reparent a node to one of its children and end up with a corrupted tree.

djpohly:
I've taken @vyivel's suggestion of asserting in case of a potential cycle, which makes sense given the behavior of other nearby functions.

types/wlr_scene.c
```diff
@@ -133,6 +133,25 @@ void wlr_scene_node_place_below(struct wlr_scene_node *node,
...
+	/* Ensure that a node cannot become its own ancestor */
+	for (struct wlr_scene_node *ancestor = new_parent; ancestor != NULL;
+			ancestor = ancestor->parent) {
+		assert(ancestor != node);
+	}
+
+	wl_list_remove(&node->state.link);
+
+	node->parent = new_parent;
+	if (new_parent != NULL) {
```
emersion, line 150: Do we want to allow a node without any parent? I think I'd rather not.

emersion:
One extra safe-guard we could add is `assert(node->type != ROOT)`, I think.

types/wlr_scene.c
djpohly: OK. I wasn't sure if there might be a case where it would be convenient for a compositor to be able to detach a subtree for a period of time. I guess the existing code already requires you to attach non-root nodes to a tree when you create them, so it would be consistent with current invariants to prevent detaching.

(If the distinction between ROOT and TREE ever disappears, this might be worth revisiting.)

emersion: Some code currently assumes each node has a root, especially with the damage tracking PRs. I'm not necessarily against allowing detached nodes, but need to be careful about edge cases, and I think `wlr_scene_node_set_enabled` can be used instead in most (?) cases.

emersion:
Err, should've asked you to squash the commits before merging. Oh well, too late.

Issue #2705, opened 2021-02-01
Add support for GL_EXT_robustness


emersion:
Would allow us to recover from GPU resets. We'll need to trash all of our renderer state and re-create it.

See:

* https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_create_context_robustness.txt
* https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_robustness.txt

emersion:
On amdgpu, a manual GPU reset can be triggered with `/sys/kernel/debug/dri/0/amdgpu_gpu_recover`.

zzag:
You will have to trash renderer state and also client buffer textures. This means that you will need to keep a wl-shm client buffer referenced even after uploading data to the opengl texture.

emersion:
Support for robustness could look like this:

- In `wlr_renderer_bind_buffer`, check `glGetGraphicsResetStatusEXT`. If it doesn't return `NO_ERROR`, busy-wait with some sleep interval and a timeout to wait for the GPU to be ready again.
- Re-initialize the renderer's internal state: EGL contexts, shaders
- Destroy all `wlr_gles2_buffer`
- Re-import all `wlr_texture` from their original `wlr_buffer`, if any (needs https://github.com/swaywm/wlroots/pull/2851). Otherwise make the texture "inert": destroy the GL texture but keep the `wlr_texture` alive to avoid crashing compositors.

emersion:
> This means that you will need to keep a wl-shm client buffer referenced even after uploading data to the opengl texture.

Indeed. We can't re-upload the buffer after we've released the buffer, because the client might be in the process of rendering to it, so its contents can be garbage.

emersion:
> Emit a wlr_renderer.events.reset event so that compositors can […] re-create their own GL state if any

This might be slightly more complicated: we need compositors to destroy their old GL state before we destroy the EGL context, and re-create it after we've established a new EGL context. We might need two events (meh API), or to keep the old EGL context alive up to `wlr_renderer.events.reset`.

zzag:
> Indeed. We can't re-upload the buffer after we've released the buffer, because the client might be in the process of rendering to it, so its contents can be garbage.

Beware though, there are applications in the wild that assume the compositor will release a shm buffer after uploading its data to an opengl texture, most prominent example is Firefox. https://bugzilla.mozilla.org/show_bug.cgi?id=1693472

emersion:
[...]

PR #3143, opened 2021-08-27
[do not merge] Surface refactor

vyivel:
_Also see individual commits._

Includes https://github.com/swaywm/wlroots/pull/3122 and https://github.com/swaywm/wlroots/pull/3120

## State structure

~~Every state is split in two parts: absolute state and relative state. Absolute state is surface-local (e.g. viewport, buffer), and relative state is commit-local (e.g. buffer damage).~~ _State structure is left as it is._

## State squashing

(The term `squashing` is taken from Git, but maybe another word would fit here better.)

"Squashing" refers to an action of
1) copying the absolute state over previous absolute state,
2) merging the relative state into previous relative state, and
3) clearing the relative state,

where "previous state" means the state right before the state being squashed in the queue.

## Changes

- ~~`wlr_surface_state.committed` is removed~~
- `wlr_surface.cached` is replaced with `wlr_surface.states` which also includes `wlr_surface.current` and `wlr_surface.pending`

These changes make squashing easier, removing potential corner cases with `wlr_surface.{current,pending}`.

- `wlr_subsurface.parent_link` is removed

Instead, use `wlr_subsurface.current.link`; same goes for `wlr_subsurface.parent_pending_link`.

- `wlr_surface.subsurfaces_{above,below}` is moved to `wlr_surface_state`

Use `wlr_surface.current.subsurfaces_{above,below}`.

## Data structure

The following data structure is maintained:

state_structure

When a state is cached, a column of `cached`<sub>`n+1`</sub> states is added right before the last one (`pending`).
When a surface extension is added, a row of `extension`<sup>`k+1`</sup> states is added. Note that rows have undefined order; `wlr_surface_state::extensions` must be treated as an unordered set.

## Notes

- ~~Using `wlr_client_buffer` has an (unintentional?) interesting side-effect: upon creation, the original buffer is unlocked, which results in using 0 client `wl_buffer`s (i.e. there is no busy `wl_buffer`). This PR nihilates this optimization, as buffers, being absolute state, can be present in multiple `wlr_surface_state`s.~~
  - ~~This makes [Cloudef/bemenu](https://github.com/Cloudef/bemenu) crash occasionally. That's their fault probably though, as I haven't encountered this problem with anything else~~.
- ~~This PR unintentionally fixes all Firefox issues with `gfx.webrender.compositor` I had before.~~ _Unrelated to subsurface caching._

## Issues

- ~~Resizing windows with subsurfaces still causes flickering, probably due to `surface_damage_subsurfaces()` doing something wrong (should it even exist?).~~ _Turned out to be Sway issue._
- ~~During a few days of testing, Firefox has crashed twice due to a Wayland error. Not sure what's the issue yet.~~ _Outdated information._


vyivel:
>Maybe it's possible to re-use wlr_addon instead of re-implementing it.

Making `wlr_surface_extension_state` an addon wouldn't make much sense as we need to a) store the order in `wlr_surface_extension::states` and b) iterate over `wlr_surface_state::extensions` to squash every extension state. Iterating over `wlr_surface::extensions` is required to initialize newly created cached state properly. `wlr_addon_set` isn't really fit to be iterated over.

>this isn't great because this forces clients to re-render

Hmm, how so? I don't really see why would that change make clients re-render.

>Maybe the client's wl_buffer can be "relative state"?

Well, a buffer isn't a difference between two objects, so it's not relative. Absolute/relative state could be renamed to surface-local/commit-local state respectively though, and `wl_buffer` would simply be moved from one state to another then (and not copied).

>wlr_client_buffer shouldn't be part of the state and should stay as a mere helper to upload/import the buffer to the renderer.

Yes, the only `wlr_client_buffer` is in `wlr_surface`.

>For instance, if the sub-surface order hasn't changed, no need to re-arrange the scene-graph on parent commit.

Fair point. I originally thought that change would make state squashing simpler, but actually it doesn't really add much complexity so I guess it could be left.

vyivel:
Merged extension interfaces, returned `wlr_surface_state.committed` back, updated the PR description to match the changes.

emersion:
This needs a rebase, now that the first part of this PR has been merged in #3169.

> >  this isn't great because this forces clients to re-render
>
> Hmm, how so? I don't really see why would that change make clients re-render.

See the discussion in https://github.com/swaywm/wlroots/issues/2705#issuecomment-833382113, and the comment a bit below with a performance comparison table.

"Immediate release" means the compositor uploads the shm buffer to the GPU on commit and then releases the buffer.

> Well, a buffer isn't a difference between two objects, so it's not relative. Absolute/relative state could be renamed to surface-local/commit-local state respectively though, and wlr_buffer would simply be moved from one state to another then (and not copied).

Right, I think that would make sense. We just need a way to make the buffer non-sticky.

vyivel:
Closed by accident.
Will be rebased soon™.

As for buffer thing, none of that absolute/relative state matters anymore now that `wlr_surface_state.committed` is back.

vyivel:
Rebased on top of master, removed outdated info from the description.

vyivel:
Rebased against https://github.com/swaywm/wlroots/pull/3191.

vyivel:
Closing as an outdated part of https://github.com/swaywm/wlroots/pull/3151.

Issue #2968, opened 2021-06-14
xwayland: add support for the startup notification protocol


emersion:
Ref https://specifications.freedesktop.org/startup-notification-spec/startup-notification-latest.txt

emersion:
Ref https://github.com/swaywm/wlroots/pull/3124

agx:
I think we don't need much more than #3146 here since handling the lanuchee part on the X side looks out of scope for wlroots.

agx:
Do we want anything more here besides https://github.com/swaywm/wlroots/pull/3146 and https://github.com/swaywm/wlroots/pull/3124 or can it be closed?

emersion:
I think this can be closed. It would be nice to expose a function to manually request xdg-activation tokens to pass down to X11 clients, and another function to validate a token. But that's separate I suppose, opened https://github.com/swaywm/wlroots/issues/3211.

PR #3146, opened 2021-08-29
xwayland: Allow to retrieve startup-id via _NET_STARTUP_INFO

agx:
A launchee notifies with a "remove"¹ message when done starting up.
Catch these and forward to the compositor. This allows the compositor to
end the startup sequence that might have been started by another
protocol like xdg-activation.

We don't handle other messages since we expect the launcher to use a
wayland protocol like xdg-activation.

While `_NET_STARTUP_ID` helps to associate toplevels with startup-ids
this signals the end of the startup sequence.

This would e.g. allow to remove tokens that we got via xdg-activation and where the launchee started an x11 application. 

1) https://specifications.freedesktop.org/startup-notification-spec/startup-notification-latest.txt

Related: #2968 , #3136

### DIFF
--- a/xwayland/xwm.c
+++ b/xwayland/xwm.c
@@ -61,6 +61,8 @@ const char *const atom_map[ATOM_LAST] = {
 	[TIMESTAMP] = "TIMESTAMP",
 	[DELETE] = "DELETE",
 	[NET_STARTUP_ID] = "_NET_STARTUP_ID",
+	[NET_STARTUP_INFO] = "_NET_STARTUP_INFO",
+	[NET_STARTUP_INFO_BEGIN] = "_NET_STARTUP_INFO_BEGIN",
 	[NET_WM_WINDOW_TYPE_NORMAL] = "_NET_WM_WINDOW_TYPE_NORMAL",
 	[NET_WM_WINDOW_TYPE_UTILITY] = "_NET_WM_WINDOW_TYPE_UTILITY",
 	[NET_WM_WINDOW_TYPE_TOOLTIP] = "_NET_WM_WINDOW_TYPE_TOOLTIP",
@@ -87,6 +89,14 @@ const char *const atom_map[ATOM_LAST] = {
 	[DND_ACTION_PRIVATE] = "XdndActionPrivate",
 	[NET_CLIENT_LIST] = "_NET_CLIENT_LIST",
 	[NET_CLIENT_LIST_STACKING] = "_NET_CLIENT_LIST_STACKING",
+};
+
+#define STARTUP_INFO_REMOVE_PREFIX "remove: ID="
+struct pending_startup_id {
+	char *msg;
+	size_t len;
+	xcb_window_t window;
+	struct wl_list link;
 };
 
 static const struct wlr_surface_role xwayland_surface_role;
@@ -1361,6 +1371,73 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
 	wlr_signal_emit_safe(&surface->events.request_activate, surface);
 }
 
+static void pending_startup_id_destroy(struct pending_startup_id *pending) {
+	wl_list_remove(&pending->link);
+	free(pending->msg);
+	free(pending);
+}
+
+static void xwm_handle_net_startup_info_message(struct wlr_xwm *xwm,
+		xcb_client_message_event_t *ev) {
+	struct pending_startup_id *pending, *curr = NULL;
+	wl_list_for_each(pending, &xwm->pending_startup_ids, link) {
+		if (pending->window == ev->window) {
+			curr = pending;
+			break;
+		}
+	}
+
+	char *start;
+	size_t buf_len = sizeof(ev->data);
+	if (curr) {
+		curr->msg = realloc(curr->msg, curr->len + buf_len);
+		if (!curr->msg) {
+			pending_startup_id_destroy(curr);
+			return;
+		}
+		start = curr->msg + curr->len;
+		curr->len += buf_len;
+	} else {
+		curr = calloc(1, sizeof(struct pending_startup_id));
+		if (!curr)
+			return;
+		curr->window = ev->window;
+		curr->msg = malloc(buf_len);
+		if (!curr->msg) {
+			free(curr);
+			return;
+		}
+		start = curr->msg;
+		curr->len = buf_len;
+		wl_list_insert(&xwm->pending_startup_ids, &curr->link);
+	}
+
+	char *id = NULL;
+	const char *data = (const char *)ev->data.data8;
+	for (size_t i = 0; i < buf_len; i++) {
+		start[i] = data[i];
+		if (start[i] == '\0') {
+			if (strncmp(curr->msg, STARTUP_INFO_REMOVE_PREFIX,
+					strlen(STARTUP_INFO_REMOVE_PREFIX)) == 0 &&
+					strlen(curr->msg) > strlen(STARTUP_INFO_REMOVE_PREFIX)) {
+				id = curr->msg + strlen(STARTUP_INFO_REMOVE_PREFIX);
+				break;
+			} else {
+				wlr_log(WLR_ERROR, "Unhandled message '%s'\n", curr->msg);
+				pending_startup_id_destroy(curr);
+				return;
+			}
+		}
+	}
+
+	if (id) {
+		struct wlr_xwayland_remove_startup_info_event data = { id, ev->window };
+		wlr_log(WLR_DEBUG, "Got startup id: %s", id);
+		wlr_signal_emit_safe(&xwm->xwayland->events.remove_startup_info, &data);
+		pending_startup_id_destroy(curr);
+	}
+}
+
 static void xwm_handle_wm_change_state_message(struct wlr_xwm *xwm,
 		xcb_client_message_event_t *ev) {
 	struct wlr_xwayland_surface *xsurface = lookup_surface(xwm, ev->window);
@@ -1399,6 +1476,9 @@ static void xwm_handle_client_message(struct wlr_xwm *xwm,
 		xwm_handle_wm_protocols_message(xwm, ev);
 	} else if (ev->type == xwm->atoms[NET_ACTIVE_WINDOW]) {
 		xwm_handle_net_active_window_message(xwm, ev);
+	} else if (ev->type == xwm->atoms[NET_STARTUP_INFO] ||
+			ev->type == xwm->atoms[NET_STARTUP_INFO_BEGIN]) {
+		xwm_handle_net_startup_info_message(xwm, ev);
 	} else if (ev->type == xwm->atoms[WM_CHANGE_STATE]) {
 		xwm_handle_wm_change_state_message(xwm, ev);
 	} else if (!xwm_handle_selection_client_message(xwm, ev)) {
@@ -1718,6 +1798,11 @@ void xwm_destroy(struct wlr_xwm *xwm) {
 	wl_list_remove(&xwm->compositor_destroy.link);
 	xcb_disconnect(xwm->xcb_conn);
 
+	struct pending_startup_id *pending, *next;
+	wl_list_for_each_safe(pending, next, &xwm->pending_startup_ids, link) {
+		pending_startup_id_destroy(pending);
+	}
+
 	xwm->xwayland->xwm = NULL;
 	free(xwm);
 }
@@ -1958,6 +2043,7 @@ struct wlr_xwm *xwm_create(struct wlr_xwayland *xwayland, int wm_fd) {
 	wl_list_init(&xwm->surfaces);
 	wl_list_init(&xwm->surfaces_in_stack_order);
 	wl_list_init(&xwm->unpaired_surfaces);
+	wl_list_init(&xwm->pending_startup_ids);
 	xwm->ping_timeout = 10000;
 
 	xwm->xcb_conn = xcb_connect_to_fd(wm_fd, NULL);
--- a/include/wlr/xwayland.h
+++ b/include/wlr/xwayland.h
@@ -72,6 +72,7 @@ struct wlr_xwayland {
 	struct {
 		struct wl_signal ready;
 		struct wl_signal new_surface;
+		struct wl_signal remove_startup_info;
 	} events;
 
 	/**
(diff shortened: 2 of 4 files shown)


xwayland/xwm.c
```diff
@@ -723,6 +734,8 @@ static void read_surface_normal_hints(struct wlr_xwm *xwm,
 #define MWM_DECOR_BORDER (1 << 1)
 #define MWM_DECOR_TITLE (1 << 3)
 
+#define STARTUP_ID_PREFIX "remove: ID="
```
emersion, line 737: This is unused.

xwayland/xwm.c
```diff
@@ -1718,6 +1791,12 @@ void xwm_destroy(struct wlr_xwm *xwm) {
 	wl_list_remove(&xwm->compositor_destroy.link);
 	xcb_disconnect(xwm->xcb_conn);
 
+	struct pending_startup_id *pending, *next;
+	wl_list_for_each_safe(pending, next, &xwm->pending_startup_ids, link) {
+		wl_list_remove(&pending->link);
+		pending_startup_id_destroy (pending);
```
emersion, line 1797: Style: extra space before `(`

xwayland/xwm.c
```diff
@@ -1399,6 +1469,9 @@ static void xwm_handle_client_message(struct wlr_xwm *xwm,
 		xwm_handle_wm_protocols_message(xwm, ev);
 	} else if (ev->type == xwm->atoms[NET_ACTIVE_WINDOW]) {
 		xwm_handle_net_active_window_message(xwm, ev);
+	} else if (ev->type == xwm->atoms[NET_STARTUP_INFO] ||
+			ev->type == xwm->atoms[NET_STARTUP_INFO_BEGIN]) {
+                xwm_handle_net_startup_info_message(xwm, ev);
```
emersion, line 1474: Style: s/spaces/tabs/

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
 	wlr_signal_emit_safe(&surface->events.request_activate, surface);
 }
 
+
+static void pending_startup_id_destroy(struct pending_startup_id *pending)
+{
```
emersion, line 1379: Style: `{` on the previous line

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
 	wlr_signal_emit_safe(&surface->events.request_activate, surface);
 }
 
+
```
emersion, line 1377: Style: extra newline here

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+			if (strncmp (curr->msg, STARTUP_INFO_REMOVE_PREFIX,
+					strlen (STARTUP_INFO_REMOVE_PREFIX)) == 0 &&
+					strlen (curr->msg) > strlen(STARTUP_INFO_REMOVE_PREFIX)) {
+				id = curr->msg + strlen(STARTUP_INFO_REMOVE_PREFIX);
+				break;
+			} else {
+				wlr_log (WLR_ERROR, "Unhandled message '%s'\n", curr->msg);
+			}
+		}
+		start++; data++;
```
emersion, line 1423: Maybe consider using `data[i]` instead of pointer arithmetic?

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+		curr = calloc(sizeof(struct pending_startup_id), 1);
+		curr->window = ev->window;
+		curr->msg = malloc(STARTUP_INFO_BUF_LEN);
+		start = curr->msg;
+		curr->len = STARTUP_INFO_BUF_LEN;
+		wl_list_insert(&xwm->pending_startup_ids, &curr->link);
+	}
+
+	const char *data = (const char *)ev->data.data8;
+	for (int i = 0; i < STARTUP_INFO_BUF_LEN; i++) {
```
emersion, line 1411: Nit: s/int/size_t/

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+		*start = *data;
+		if (*start == '\0') {
+			if (strncmp (curr->msg, STARTUP_INFO_REMOVE_PREFIX,
```
emersion, line 1414: Style: extra spaces before `(`, applies to the whole block.

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+static void pending_startup_id_destroy(struct pending_startup_id *pending)
+{
+	free(pending->msg);
+	free(pending);
+}
+
+
+static void xwm_handle_net_startup_info_message(struct wlr_xwm *xwm,
+		xcb_client_message_event_t *ev) {
+	char *start, *id = NULL;
```
emersion, line 1387: Style nit: these can be declared before the `if` below

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+	struct pending_startup_id *pending, *curr = NULL;
+	wl_list_for_each(pending, &xwm->pending_startup_ids, link) {
+		if (pending->window == ev->window) {
+			curr = pending;
+			break;
+		}
+	}
+
+	if (curr) {
+		curr->msg = realloc(curr->msg, curr->len + STARTUP_INFO_BUF_LEN);
```
emersion, line 1398: We should handle alloc failures for `realloc`, `calloc` and `malloc`

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+		start = curr->msg + curr->len;
+		curr->len += STARTUP_INFO_BUF_LEN;
+	} else {
+		curr = calloc(sizeof(struct pending_startup_id), 1);
```
emersion, line 1402: Nit: the arguments for `calloc` are reversed

xwayland/xwm.c
```diff
@@ -89,6 +91,15 @@ const char *const atom_map[ATOM_LAST] = {
 	[NET_CLIENT_LIST_STACKING] = "_NET_CLIENT_LIST_STACKING",
 };
 
+#define STARTUP_INFO_BUF_LEN 20
```
emersion, line 94: Where is this coming from? I don't see it in the spec.

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+				wlr_log (WLR_ERROR, "Unhandled message '%s'\n", curr->msg);
+			}
+		}
+		start++; data++;
+	}
+
+	if (id) {
+		wlr_log(WLR_DEBUG, "Got startup id: %s", id);
+		wlr_signal_emit_safe(&xwm->xwayland->events.remove_startup_info, id);
+		wl_list_remove(&(curr->link));
```
emersion, line 1429: Style: unnecessary parentheses

xwayland/xwm.c
```diff
@@ -1361,6 +1374,63 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+	for (int i = 0; i < STARTUP_INFO_BUF_LEN; i++) {
+		*start = *data;
+		if (*start == '\0') {
+			if (strncmp (curr->msg, STARTUP_INFO_REMOVE_PREFIX,
+					strlen (STARTUP_INFO_REMOVE_PREFIX)) == 0 &&
+					strlen (curr->msg) > strlen(STARTUP_INFO_REMOVE_PREFIX)) {
+				id = curr->msg + strlen(STARTUP_INFO_REMOVE_PREFIX);
+				break;
+			} else {
+				wlr_log (WLR_ERROR, "Unhandled message '%s'\n", curr->msg);
```
emersion, line 1420: In this case, it doesn't seem like we ever cleanup the `pending_startup_id`?

emersion:
Hm, but the tokens are removed when used to activate an X11 window, so why do we need to have a separate "remove" event?

xwayland/xwm.c
agx: https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/blob/master/src/xproto.xml#L1032

agx:
> Hm, but the tokens are removed when used to activate an X11 window, so why do we need to have a separate "remove" event?

I don't understand that question. Where is something removed? The client signals via `remove ID=...` that it's up which should then e.g. trigger the removal of the xdg-activation token (which isnt' in that MR but would go into the bits of https://github.com/swaywm/wlroots/pull/3136 that allow to create/destroy tokens since the compositor should do that explicitly). It can't be certain it also gets a `NET_STARTUP_ID` startup id with a toplevel as far as i understand the spec.  There are clients out there which send the removal message but never `NET_STARTUP_ID` as far as I can tell.

emersion:
Hm, sorry, I should've phrased that better.

xdg-activation lacks the "remove" request X11 has. Instead, tokens expire with a compositor-specific mechanism (e.g. a timeout).

Which is why I'm wondering why the X11 "remove" request exists at all. If the launchee just ignores the activation token available in its environment, then it isn't a big deal to just let it expire. If the launchee does open an X11 connection but doesn't want to activate any window, then the same applies (although I'm not sure why an X11 client would choose to do this).

>  There are clients out there which send the removal message but never NET_STARTUP_ID as far as I can tell.

Interesting, do you know which kind of clients and under which circumstances they do it?

> As far as i understand things: NET_STARTUP_INFO with remove: ID= signals client up. NET_STARTUP_ID with startup-id set signals this topelvel belogs to this client

Oh. So compositors shouldn't activate an X11 window when the client sets NET_STARTUP_ID? They should wait for "remove"?

agx:
> Which is why I'm wondering why the X11 "remove" request exists at all.

It is my understanding that the `remove` is the "i'm starup sequence done" request. I think he key here is to look at two things: the startup sequence (launch the client and have it signal "i'm spawned") and activation (focus a toplevel). xdg-activation only looks at the later (handling startup sequence implicitly) while x11 seems to split that.

>  If the launchee just ignores the activation token available in its environment, then it isn't a big deal to just let it expire.

It can become an issue if you e.g. tie a launch splash to it that you hide once the app is up (ideally when activation happens but in the x11 case you can only rely on startup sequence being complerte) so it would be my aim (if this gets merged) to add enough to xdg-activation so you can inject and remove tokens so `events.request_activate` becomes the norm and expiry only happens when the client really doesn't show up (or is broken and doesn't request activation).

> Oh. So compositors shouldn't activate an X11 window when the client sets NET_STARTUP_ID? They should wait for "remove"?

I think activation should happen based on `NET_STARTUP_ID` but the end of the startup sequence is indicated via `remove:` in `NET_STARTUP_INFO`. Note that this is from observation and reading of the protocol.

So far i've seen qt clients like e.g. qutebrowser only send the `removed:` but never a `NET_STARTUP_ID` while gtk apps like firefox send both.

emersion:
> So far i've seen qt clients like e.g. qutebrowser only send the removed: but never a NET_STARTUP_ID while gtk apps like firefox send both.

Hm. Indeed, Qt doesn't seem to support `_NET_STARTUP_ID` at all. This is quite unfortunate.

https://github.com/qt/qtbase/blob/9db7cc79a26ced4997277b5c206ca15949133240/src/plugins/platforms/xcb/qxcbatom.h#L152

Oh well. I guess we get to do all of this fun parsing stuff then.

emersion:
Opened https://bugreports.qt.io/browse/QTBUG-96276 to track this.

xwayland/xwm.c
emersion: Maybe consider using `sizeof(ev->data)` instead?

xwayland/xwm.c
```diff
@@ -1361,6 +1372,70 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+		wl_list_insert(&xwm->pending_startup_ids, &curr->link);
+	}
+
+	char *id = NULL;
+	const char *data = (const char *)ev->data.data8;
+	for (size_t i = 0; i < STARTUP_INFO_BUF_LEN; i++) {
+		start[i] = data[i];
+		if (start[i] == '\0') {
+			if (strncmp(curr->msg, STARTUP_INFO_REMOVE_PREFIX,
+					strlen (STARTUP_INFO_REMOVE_PREFIX)) == 0 &&
```
emersion, line 1420: Style: no space between `strlen` and `(`

xwayland/xwm.c
```diff
@@ -1361,6 +1372,70 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+			curr = pending;
+			break;
+		}
+	}
+
+	char *start;
+	if (curr) {
+		curr->msg = realloc(curr->msg, curr->len + STARTUP_INFO_BUF_LEN);
+		if (!curr->msg) {
+			pending_startup_id_destroy(curr);
```
emersion, line 1395: Need a `return` here

xwayland/xwm.c
```diff
@@ -1361,6 +1372,70 @@ static void xwm_handle_net_active_window_message(struct wlr_xwm *xwm,
...
+				wlr_log(WLR_ERROR, "Unhandled message '%s'\n", curr->msg);
+				pending_startup_id_destroy(curr);
+				return;
+			}
+		}
+	}
+
+	if (id) {
+		wlr_log(WLR_DEBUG, "Got startup id: %s", id);
+		wlr_signal_emit_safe(&xwm->xwayland->events.remove_startup_info, id);
```
emersion, line 1434: We might want to provide more info to compositors here. Can we introduce `struct wlr_xwayland_remove_startup_info_event` with a `const char *id` field and a window field?

This will also allow us to extend the event in the future if need be.

xwayland/xwm.c
```diff
@@ -89,6 +91,15 @@ const char *const atom_map[ATOM_LAST] = {
 	[NET_CLIENT_LIST_STACKING] = "_NET_CLIENT_LIST_STACKING",
 };
 
+#define STARTUP_INFO_BUF_LEN 20
+#define STARTUP_INFO_REMOVE_PREFIX "remove: ID="
+struct pending_startup_id {
+	char *msg;
+	int len;
```
emersion, line 98: Nit: this should be `size_t`

emersion (approved):
LGTM otherwise!

emersion (changes requested):
Err, I meant to request changes.

emersion (approved):
Thanks!

Issue #803, opened 2018-04-02
Remove wlr_seat_set_keyboard


emersion:
This seems to be only used to update the keymap. We could just require the compositor to listen for keymap changes in wlr_keyboard and forward it to the wlr_seat.

PR #3152, opened 2021-08-31
tinywl: fix crash if there is no keyboard

ifreund:
To reproduce, run

WLR_BACKENDS=headles tinywl -s foot


emersion:
Do we need to give keyboard focus and activate the toplevel in any particular order? Do clients care about it?

ifreund:
> Do we need to give keyboard focus and activate the toplevel in any particular order? Do clients care about it?

I don't think it matters: 
```xml
      <entry name="activated" value="4" summary="the surface is now activated">
	<description summary="the surface is now activated">
	  Client window decorations should be painted as if the window is
	  active. Do not assume this means that the window actually has
	  keyboard or pointer focus.
	</description>
      </entry>
```

ifreund:
Though I guess it would make more sense to do absolutely nothing in this function if there is no keyboard.

emersion:
Hm, but the user could still want to focus a toplevel with the pointer, even if there is no keyboard?

ifreund:
> Hm, but the user could still want to focus a toplevel with the pointer, even if there is no keyboard?

Yeah, that better matches what sway and river at least actually do as well. Reverted the latest force push.

ifreund:
Now we still send the enter even if there is no actual keyboard, but with no keys or modifiers pressed.

tinywl/tinywl.c
```diff
@@ -125,8 +124,13 @@ static void focus_view(struct tinywl_view *view, struct wlr_surface *surface) {
...
 	 */
-	wlr_seat_keyboard_notify_enter(seat, view->xdg_surface->surface,
-		keyboard->keycodes, keyboard->num_keycodes, &keyboard->modifiers);
+	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat);
+	if (keyboard != NULL) {
+		wlr_seat_keyboard_notify_enter(seat, view->xdg_surface->surface,
+			keyboard->keycodes, keyboard->num_keycodes, &keyboard->modifiers);
+	} else {
+		wlr_seat_keyboard_notify_enter(seat, surface, NULL, 0, NULL);
+	}
```
ifreund, line 133: This kinda makes me wonder while wlroots makes compositors do this check themselves, instead `wlr_seat_keyboard_notify_enter()` could just take a pointer to a `wlr_keyboard`.

tinywl/tinywl.c
```diff
@@ -125,8 +124,13 @@ static void focus_view(struct tinywl_view *view, struct wlr_surface *surface) {
...
 	 * clients without additional work on your part.
 	 */
-	wlr_seat_keyboard_notify_enter(seat, view->xdg_surface->surface,
-		keyboard->keycodes, keyboard->num_keycodes, &keyboard->modifiers);
+	struct wlr_keyboard *keyboard = wlr_seat_get_keyboard(seat);
+	if (keyboard != NULL) {
+		wlr_seat_keyboard_notify_enter(seat, view->xdg_surface->surface,
+			keyboard->keycodes, keyboard->num_keycodes, &keyboard->modifiers);
+	} else {
+		wlr_seat_keyboard_notify_enter(seat, surface, NULL, 0, NULL);
```
emersion, line 132: Hm, I wonder if it really makes sense to do this…?

Sway may do it because of data device related things (copy/paste relies on keyboard focus).

tinywl/tinywl.c
emersion: Hm, yes, that's a good point. At some point we wanted to get rid of `wlr_seat_set_keyboard` and de-couple `wlr_keyboard` from `wlr_seat` (just like `wlr_cursor`/`wlr_pointer` is decoupled from `wlr_seat`), but stopped halfway through. Ref https://github.com/swaywm/wlroots/issues/803

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3152

PR #3153, opened 2021-09-01
render/allocator: re-open GBM FD

emersion:
Using the same DRM file description for the DRM backend and for the
GBM allocator will result in GEM handle ref'counting issues [1].
Re-open the DRM FD to fix these issues.

[1]: https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/110

### DIFF
--- a/render/allocator/allocator.c
+++ b/render/allocator/allocator.c
@@ -1,5 +1,8 @@
+#define _POSIX_C_SOURCE 200809L
 #include <assert.h>
+#include <fcntl.h>
 #include <stdlib.h>
+#include <unistd.h>
 #include <wlr/util/log.h>
 #include <xf86drm.h>
 #include "backend/backend.h"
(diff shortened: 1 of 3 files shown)


subdiff (approved):
Looks good to me. You're moving out functionality from `wlr_drm_dumb_allocator_create` into the `reopen_drm_node` as you wanna reuse it for the `wlr_gbm_allocator_create` case in the `allocator_autocreate_with_drm_fd` call too.

bl4ckb0ne (approved):

bl4ckb0ne:
LGTM

render/allocator/allocator.c
```diff
@@ -17,6 +20,26 @@ void wlr_allocator_init(struct wlr_allocator *alloc,
 	wl_signal_init(&alloc->events.destroy);
 }
 
+/* Re-open the DRM node to avoid GEM handle ref'counting issues. See:
+ * https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/110
+ * TODO: don't assume we have the permission to just open the DRM node,
+ * find another way to re-open it.
```
quantum5, line 26: Does `dup(2)` work or do we have to reopen by name?

emersion: Unfortunately dup(2) doesn't work here, because the GEM handlers are per file description. dup(2) only gives a new file descriptor, but shares the same file description on the kernel side. See the linked libdrm discussion above.

Issue #3154, opened 2021-08-05
Crash when fullscreening


J0nnyMak0:
I'm seeing a crash occasionally when fullscreening a window. Log attached.
```
sway version 1.6-f4cda515 (Aug  4 2021, branch 'master')

```
[sway.log](https://github.com/swaywm/wlroots/files/6940166/sway.log)

J0nnyMak0:
Unfortunately that didn't help. New log and trace attached.
[sway.log](https://github.com/swaywm/wlroots/files/7093976/sway.log)
[gdb.txt](https://github.com/swaywm/wlroots/files/7093977/gdb.txt)

emersion:
OK, I don't know why it took me so long to realize it, but the fix is pretty trivial: https://github.com/swaywm/wlroots/pull/3155

Can you give it a go?

J0nnyMak0:
That fixes it! Thanks!

PR #3155, opened 2021-09-01
backend/drm: handle drm_surface_blit errors

emersion:
drm_surface_blit returns NULL on error. This can happen e.g. when
the source buffer cannot be imported into EGL.

Closes: https://github.com/swaywm/wlroots/issues/3154


J0nnyMak0:
This fixes https://github.com/swaywm/wlroots/issues/3154

Issue #3156, opened 2021-09-02
[render/allocator/gbm.c:63] gbm_bo_get_fd failed


jbeich:
Regressed by #3153. Starting Sway on DRM console hangs without leaving console (Ctrl+Alt+Fn works fine). Tested on FreeBSD with i915 ported from Linux 5.4.92.

From [debug log](https://github.com/swaywm/wlroots/files/7096553/sway.log):
```diff
 [DEBUG] [wlr] [types/wlr_output.c:689] Attaching empty buffer to output for modeset
 [DEBUG] [wlr] [types/wlr_output.c:524] Choosing primary buffer format 0x34325241 for output 'DP-1'
 [DEBUG] [wlr] [render/swapchain.c:105] Allocating new swapchain buffer
-[DEBUG] [wlr] [render/allocator/gbm.c:129] Allocated 3840x2160 GBM buffer (format 0x34325241, modifier 0x100000000000004)
-[DEBUG] [wlr] [render/gles2/renderer.c:143] Created GL FBO for buffer 3840x2160
-[DEBUG] [sway/config/output.c:346] Output DPI: 162.560000x161.364706
+[ERROR] [wlr] [render/allocator/gbm.c:63] gbm_bo_get_fd failed
+[ERROR] [wlr] [render/swapchain.c:109] Failed to allocate buffer
+[DEBUG] [sway/config/output.c:388] Preferred mode rejected, falling back to another mode
+[DEBUG] [wlr] [backend/drm/drm.c:464] connector DP-1: Can't enable an output without a buffer
```

emersion:
Hm, weird. Seems like the buffer was successfully created but *something* prevents it from being exported as a DMA-BUF… Is there anything in `dmesg`?

emersion:
Please also try enabling the bit 0x08 in the DRM logs (https://github.com/swaywm/wlroots/wiki/DRM-Debugging).

jbeich:
dmesg shows nothing from kernel, even after `sysctl hw.dri.debug=0x08`.

[DRM debug log](https://github.com/swaywm/wlroots/files/7099872/sway.drm-debug.log) with every bit enabled.

emersion:
Hm, `DRM_IOCTL_PRIME_HANDLE_TO_FD` fails with `EACCES` it seems. May be due to the use of a primary node without authentication.

jbeich:
On FreeBSD devices of the same type may have different major numbers which may confuse DRM authentication.
```
[...]

PR #3157, opened 2021-09-02
render/allocator/gbm: add log message for gbm_bo_get_fd_for_plane

emersion:
Makes it easier to find out which branch is taken when debugging
issues like [1].

[1]: https://github.com/swaywm/wlroots/issues/3156


kennylevinsen (approved):
LGTM

PR #3158, opened 2021-09-02
render/allocator: use empty DRM lease to re-open node

emersion:
This allows us to obtain a new DRM file description without relying
on filesystem permissions.


emersion:
The kernel part of this has been merged, so this is now ready for review.

PR #3160, opened 2021-09-02
scene: add wlr_scene_attach_output_layout

emersion:
This is a helper to integrate wlr_scene with wlr_output_layout.

Depends on: https://github.com/swaywm/wlroots/pull/3117

* * *

cc @ifreund @kennylevinsen @ammen99 

This uses the approach discussed on IRC.


kennylevinsen (approved):
LGTM. Don't have a super strong opinion about the location, but maybe there's some value in making the distinction.

emersion:
I've moved the new helper implementation to a separate source file, but kept the declaration in the `wlr_scene.h` header for now.

PR #3165, opened 2021-09-04
Add wlr_screenlocker

### LINKED ISSUE #2706: Improved lock screen support
I'm looking for some feedback and ideas about the current state of lock screens under wlroots.

Right now, it looks like lock screens are implemented the same way they were under X11, specifically: fullscreen window on top of everything else and input inhibit. While this works, it has a number of problems, and I'm hoping that it can be improved on.

### Problems

1) If anything causes the lock screen process to exit, the screen becomes unlocked. This issue continues to pop up (eg. [xscreensaver](https://www.jwz.org/blog/2021/01/i-told-you-so-2021-edition/), [swaylock](https://github.com/swaywm/swaylock/issues/162)), and likely can't be solved ([thanks, oomk](https://www.jwz.org/xscreensaver/faq.html#oom)) without compositor support.

2) Programs can [unintentionally](https://github.com/swaywm/swaylock/issues/107) end up on the lock screen. Programs that want to end up on the lock screen have no way to ensure they do.

3) Race conditions: when resuming from sleep or messing with [dpms](https://github.com/swaywm/swaylock/issues/99), the desktop sometimes shows briefly before the lock screen does.

### Protocol design goals

1) Provide a mechanism to request the screen be locked / unlocked. The lock state should persist even if the requesting client dies.

2) Provide clients with lock screen state: `enum {UNLOCKED,LOCKING,LOCKED,UNLOCKING}`
    * `UNLOCKED`: the unlocked desktop is visible

    * `LOCKING`: the compositor has received a LOCK request, but the unlocked desktop is still visible
    
    * `LOCKED`: the unlocked desktop is not visible
    
    * `UNLOCKING`: the locked desktop is still visible

3) Allow clients to configure which surfaces can appear where

### Client usage

* A wallpaper process could display different images when locked / unlocked (similar to how Android and iOS do it).

* A volume/brightness OSD could ensure it appears when locked and unlocked.

* A notification daemon might want to hide notification content and only display icons with numbers next to them when locked.

* A panel could hide all workspace buttons/menus/trays but continue to show current time.

* A lock screen process (eg. swaylock --daemonize) could wait for a `LOCKED` event, ensuring that the desktop is not visible once the parent exits.

### Implementation

I haven't worked much with wayland, but here's what I'm thinking, and I'd appreciate any feedback.

This should probably be a new protocol. It looks like it would fit nicely into layer_shell, but with kwin implementing layer_shell, I doubt it would be a good idea to force a lockscreen on them since they already handle locking their own way.

I'm not entirely sure of the best way to communicate to clients which surfaces the compositor is capable of rendering on a lock screen.

danieldg:
Fixes https://github.com/swaywm/wlroots/issues/2706

### DIFF
--- a/include/wlr/types/wlr_screenlocker_v1.h
+++ b/include/wlr/types/wlr_screenlocker_v1.h
@@ -1,0 +1,60 @@
+/*
+* This an unstable interface of wlroots. No guarantees are made regarding the
+* future consistency of this API.
+*/
+#ifndef WLR_USE_UNSTABLE
+#error "Add -DWLR_USE_UNSTABLE to enable unstable wlroots features"
+#endif
+
+#ifndef WLR_TYPES_WLR_SCREENLOCKER_V1_H
+#define WLR_TYPES_WLR_SCREENLOCKER_V1_H
+
+struct wlr_screenlock_manager_v1 {
+	struct wl_global *global;
+
+	struct wl_list locker_globals;
+	struct wl_list lock_surfaces;
+
+	bool locked;
+
+	// This is NULL if a lock was abandoned
+	struct wl_resource *lock_resource;
+
+	struct {
+		struct wl_signal change_request; // arg: wlr_screenlock_change
+	} events;
+};
+
+struct wlr_screenlock_lock_surface {
+	struct wl_resource *resource;
+	struct wlr_surface *surface;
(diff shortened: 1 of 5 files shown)


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3165

PR #3169, opened 2021-09-06
Minor surface refactor

vyivel:
A bunch of commits from #3143, including #3120 and #3122.

cc @emersion 

## Changes

- `wlr_subsurface.parent_link` is removed

Instead, use `wlr_subsurface.current.link`; same goes for `wlr_subsurface.parent_pending_link`.

- `wlr_surface.subsurfaces_{above,below}` is moved to `wlr_surface_state`

Use `wlr_surface.current.subsurfaces_{above,below}`.


emersion (approved):
LGTM, thanks!

PR #3170, opened 2021-09-06
surface: add WLR_SURFACE_STATE_SUBSURFACES

emersion:
This allows to figure out when the subsurface list has changed:
subsurface added, removed, or re-ordered. In the case where
the subsurface list hasn't changed, some book-keeping operations
can be skipped.

Useful for https://github.com/swaywm/wlroots/pull/3128

cc @vyivel

### DIFF
--- a/types/wlr_surface.c
+++ b/types/wlr_surface.c
@@ -698,6 +698,7 @@ static void subsurface_destroy(struct wlr_subsurface *subsurface) {
 	wl_list_remove(&subsurface->surface_destroy.link);
 
 	if (subsurface->parent) {
+		subsurface->parent->pending.committed |= WLR_SURFACE_STATE_SUBSURFACES;
 		wl_list_remove(&subsurface->current.link);
 		wl_list_remove(&subsurface->pending.link);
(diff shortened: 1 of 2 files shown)


types/wlr_surface.c
```diff
@@ -698,6 +701,7 @@ static void subsurface_destroy(struct wlr_subsurface *subsurface) {
 	wl_list_remove(&subsurface->surface_destroy.link);
 
 	if (subsurface->parent) {
+		subsurface->parent->pending.committed |= WLR_SURFACE_STATE_SUBSURFACES;
```
vyivel, line 704: Subsurfaces are destroyed instantly, their removal isn't double-buffered.

emersion: Yes, but I'd still like to give the caller a chance to figure out that the subsurface list changed…

vyivel: I think callers should simply listen to new_subsurface/destroy signals, and WLR_SURFACE_STATE_SUBSURFACES should be left as an optimization thing.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3170

PR #3175, opened 2021-09-07
output-damage: fix damage on modeset

emersion:
On modeset wlr_output will internally allocate a buffer. The
backend will emit a "mode" output event, then wlr_output will
emit a "commit" event.

wlr_output_damage handles the "mode" event by damaging the whole
output, and then handles the "commit" event. However the commit
event has a buffer, so wlr_output_damage rotates the damage in its
ring buffer, thinking the compositor has rendered a frame. The
compositor hasn't rendered a frame, what wlr_output_damage sees is
the internal wlr_output black buffer used for the modeset.

Let's fix this by damaging the whole output in the "commit" event
handler if the mode has changed. Additionally, damage the whole
output after rotating the damage ring buffer.

### DIFF
--- a/types/wlr_output_damage.c
+++ b/types/wlr_output_damage.c
@@ -61,31 +61,30 @@ static void output_handle_commit(struct wl_listener *listener, void *data) {
 		wl_container_of(listener, output_damage, output_commit);
 	struct wlr_output_event_commit *event = data;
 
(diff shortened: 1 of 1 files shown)


kennylevinsen (approved):
LGTM

PR #3176, opened 2021-09-07
backend/drm: drop attempt_enable_needs_modeset

emersion:
Modesets require a buffer. The DRM backend tried to auto-enable
outputs when a CRTC becomes available in the past, but now that
fails because no buffer is available.

Instead of having this magic inside the DRM backend, a better
approach is to do it in the compositor or in an optional helper.


kennylevinsen (approved):
LGTM

PR #3193, opened 2021-09-13
ci: switch to seatd-launch

emersion:
cc @kennylevinsen


emersion:
Added an `udevadm settle` invocation, see commit message.

kennylevinsen (approved):
> udevadm settle

Ugh. But LGTM!

PR #3195, opened 2021-09-13
External renderer

bl4ckb0ne:
Allows for user to plug their renderer instead of creating one under the hood.


emersion:
"render/egl: make most functions public" -- hmm, I don't know if we want to go down this path and build a custom renderer framework library. Do you have example code using it?

"types/buffer: make {begin,end}_data_ptr_access part of the public API" -- ACK

"render/pixel_format: make wlr_pixel_format_info part of the public API" -- what is this used for? Example downstream code maybe?

"render: wlr_renderer_init_wl_display fails if no renderer is provided" -- I think I'd rather just let it crash like it is today...

This is missing the main ingredient: stop initializing the renderer and allocator in the backend init functions.

bl4ckb0ne:
https://git.sr.ht/~bl4ckb0ne/wxrc/commit/4d4f63055d68c696520fa0dcc6004450f5fd53d3 very WIP, `wlr_egl` saves me a bunch of code. But it would be totally doable on my own.

emersion:
"types/buffer: make {begin,end}_data_ptr_access part of the public API" -- pushed to master

I think the rest is still very much WIP.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3195

PR #3197, opened 2021-09-15
output-swapchain: new helper

emersion:
- [ ] Perform test-only output commits with format modifiers
- [ ] Figure out interactions with `wlr_output_damage`
- [ ] Figure out if we can use it internally in `wlr_output`

Closes: https://github.com/swaywm/wlroots/issues/3079


types/wlr_output_swapchain.c
```diff
@@ -0,0 +1,236 @@
...
+#include "render/wlr_renderer.h"
+
+struct wlr_output_swapchain_manager *wlr_output_swapchain_manager_create(
+		struct wlr_backend *backend) {
+	struct wlr_output_swapchain_manager *manager = calloc(1, sizeof(*manager));
+	if (manager == NULL) {
+		return NULL;
+	}
+
+	manager->renderer = wlr_renderer_autocreate(backend);
```
bl4ckb0ne, line 23: Should the helper "own" the renderer? `wlr_backend_get_renderer` seems more appropriate for external renderers.

emersion: I want to get rid of `wlr_backend_get_renderer` eventually, because the backend should be completely separate from the renderer.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3197

PR #3198, opened 2021-09-15
build: simplify get_variable calls

emersion:
With recent-ish Meson we can stop repeating the variable name for
each provider.


bl4ckb0ne (approved):

bl4ckb0ne:
LGTM, thanks :)

PR #3151, opened 2021-08-31
Use wlr_surface_extension everywhere

vyivel:
Depends on #3143

- [x] `wlr_subsurface` (done in #3143)
- [x] `wlr_xdg_surface`
- [ ] `wlr_xdg_toplevel` (states)
- [ ] `wlr_layer_surface_v1`
- [ ] Something else?

PR #3199, opened 2021-09-16
xdg-toplevel: refactor configure/state flow

vyivel:
*See individual commits.*

Fixes https://github.com/swaywm/wlroots/issues/2330

---

Preliminary work for https://github.com/swaywm/wlroots/pull/3151.
If this is fine, I'll make a similar PR for `wlr_layer_surface`.

This is a breaking change: compositors that used to check `wlr_xdg_toplevel.client_pending` will now have to check `wlr_xdg_toplevel.requested`.

### DIFF
--- a/include/wlr/types/wlr_xdg_shell.h
+++ b/include/wlr/types/wlr_xdg_shell.h
@@ -105,10 +105,16 @@ struct wlr_xdg_toplevel_state {
 	uint32_t width, height;
 	uint32_t max_width, max_height;
 	uint32_t min_width, min_height;
+};
 
-	// Since the fullscreen request may be made before the toplevel's surface
-	// is mapped, this is used to store the requested fullscreen output (if
-	// any) for wlr_xdg_toplevel::client_pending.
+struct wlr_xdg_toplevel_configure {
+	bool maximized, fullscreen, resizing, activated;
+	uint32_t tiled; // enum wlr_edges
+	uint32_t width, height;
+};
+
+struct wlr_xdg_toplevel_requested {
+	bool maximized, minimized, fullscreen;
 	struct wlr_output *fullscreen_output;
 	struct wl_listener fullscreen_output_destroy;
 };
@@ -121,10 +127,15 @@ struct wlr_xdg_toplevel {
 	struct wlr_xdg_surface *parent;
 	struct wl_listener parent_unmap;
 
-	struct wlr_xdg_toplevel_state client_pending;
-	struct wlr_xdg_toplevel_state server_pending;
-	struct wlr_xdg_toplevel_state last_acked;
-	struct wlr_xdg_toplevel_state current;
+	struct wlr_xdg_toplevel_state current, pending;
+
+	// Properties to be sent to the client in the next configure event.
+	struct wlr_xdg_toplevel_configure scheduled;
+
+	// Properties that the client has requested. Intended to be checked
+	// by the compositor on surface map and handled accordingly
+	// (e.g. a client might want to start already in a fullscreen state).
+	struct wlr_xdg_toplevel_requested requested;
 
 	char *title;
 	char *app_id;
@@ -147,7 +158,7 @@ struct wlr_xdg_surface_configure {
 	struct wl_list link; // wlr_xdg_surface::configure_list
 	uint32_t serial;
 
-	struct wlr_xdg_toplevel_state *toplevel_state;
+	struct wlr_xdg_toplevel_configure *toplevel_configure;
 };
 
 /**
--- a/types/xdg_shell/wlr_xdg_toplevel.c
+++ b/types/xdg_shell/wlr_xdg_toplevel.c
@@ -11,50 +11,67 @@ void handle_xdg_toplevel_ack_configure(
 		struct wlr_xdg_surface *surface,
 		struct wlr_xdg_surface_configure *configure) {
 	assert(surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL);
-	assert(configure->toplevel_state != NULL);
 
-	surface->toplevel->last_acked = *configure->toplevel_state;
+	struct wlr_xdg_toplevel_configure *acked = configure->toplevel_configure;
+	assert(acked != NULL);
+
+	surface->toplevel->pending.maximized = acked->maximized;
+	surface->toplevel->pending.fullscreen = acked->fullscreen;
+	surface->toplevel->pending.resizing = acked->resizing;
+	surface->toplevel->pending.activated = acked->activated;
+	surface->toplevel->pending.tiled = acked->tiled;
+
+	surface->toplevel->pending.width = acked->width;
+	surface->toplevel->pending.height = acked->height;
 }
 
-bool compare_xdg_surface_toplevel_state(struct wlr_xdg_toplevel *state) {
-	// is pending state different from current state?
-	if (!state->base->configured) {
+bool compare_xdg_surface_toplevel_state(struct wlr_xdg_toplevel *toplevel) {
+	// Is the scheduled configure different from the last sent one?
+	if (!toplevel->base->configured) {
 		return false;
 	}
 
-	struct wlr_xdg_toplevel_state *configured = NULL;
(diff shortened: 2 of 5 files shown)


include/wlr/types/wlr_xdg_shell.h
```diff
@@ -121,10 +125,11 @@ struct wlr_xdg_toplevel {
...
 	struct wl_listener parent_unmap;
 
-	struct wlr_xdg_toplevel_state client_pending;
-	struct wlr_xdg_toplevel_state server_pending;
-	struct wlr_xdg_toplevel_state last_acked;
-	struct wlr_xdg_toplevel_state current;
+	struct wlr_xdg_toplevel_state current, pending;
+	struct wlr_xdg_toplevel_state scheduled;
+
+	// Intended to be checked on map to handle events before it
```
ifreund, line 131: This comment could be a bit more clear that the compositor author using wlroots is the one intended to check this, as opposed to the check being done internally to wlroots.

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -121,10 +125,11 @@ struct wlr_xdg_toplevel {
 	struct wlr_xdg_surface *parent;
 	struct wl_listener parent_unmap;
 
-	struct wlr_xdg_toplevel_state client_pending;
-	struct wlr_xdg_toplevel_state server_pending;
-	struct wlr_xdg_toplevel_state last_acked;
-	struct wlr_xdg_toplevel_state current;
+	struct wlr_xdg_toplevel_state current, pending;
+	struct wlr_xdg_toplevel_state scheduled;
```
ifreund, line 129: I think some more verbose docs here would go a long way towards making this code easier to understand for wlroots contributors and compositor authors alike. Feel free to expand/tweak as you see fit, or just copy this verbatim:
```suggestion
	// The current and pending state of the toplevel. The pending state
	// includes values sent by the compositor in a configure as well as
	// the max/min width/height set by the client. When the client commits
	// the surface, the entire pending state is copied as-is into the
	// current state.
	struct wlr_xdg_toplevel_state current, pending;
	// State scheduled by the compositor to be sent to the client in
	// the next configure event. The max/min width/height fields are ignored
	struct wlr_xdg_toplevel_state scheduled;
```

ifreund (approved):
LGTM! Splitting off the things *not* tied to the surface commit into a separate struct is a great idea and much cleaner. I'm also quite happy with the new names you've chosen, I find "scheduled" and "requested" much more clear.

include/wlr/types/wlr_xdg_shell.h
vyivel: Added a hopefully more clear comment.

include/wlr/types/wlr_xdg_shell.h
vyivel: Added a comment for `scheduled` state. I don't think there is a reason to describe `current`/`pending`, as it basically copies `wlr_surface.{current,pending}` semantics.

vyivel:
Giving it some more thoughts, allowing `requested` to be accessed only on map event feels hacky. Also, considering that requested properties are double-buffered, I suppose they must a part of state.

```c
struct wlr_xdg_toplevel_state {
    ...
    struct wlr_xdg_toplevel_requested requested;
};
```

vyivel:
Technically `scheduled` isn't a state too, so `wlr_xdg_toplevel_configure` is added. This also avoids having useless size constraint fields in `scheduled`.

include/wlr/types/wlr_xdg_shell.h
```diff
@@ -121,10 +127,16 @@ struct wlr_xdg_toplevel {
...
 
-	struct wlr_xdg_toplevel_state client_pending;
-	struct wlr_xdg_toplevel_state server_pending;
-	struct wlr_xdg_toplevel_state last_acked;
-	struct wlr_xdg_toplevel_state current;
+	struct wlr_xdg_toplevel_state current, pending;
+
+	// Properties to be sent to the client in the next configure event.
+	struct wlr_xdg_toplevel_configure scheduled;
+	struct wlr_xdg_toplevel_configure last_acked;
```
ifreund, line 134: Storing this `last_acked` configure is entirely redundant with the `pending` state. I think we should instead construct this configure on the stack from the `pending` state when needed for comparision in `compare_xdg_surface_toplevel_state()`.

ifreund (commented):
Separating `configure` and `state` into different structs seems like a great idea to me!

include/wlr/types/wlr_xdg_shell.h
vyivel: Makes sense. Will remove it a bit later.

types/xdg_shell/wlr_xdg_toplevel.c
```diff
@@ -11,50 +11,67 @@ void handle_xdg_toplevel_ack_configure(
...
+	surface->toplevel->pending.activated = acked->activated;
+	surface->toplevel->pending.tiled = acked->tiled;
+
+	surface->toplevel->pending.width = acked->width;
+	surface->toplevel->pending.height = acked->height;
 }
 
-bool compare_xdg_surface_toplevel_state(struct wlr_xdg_toplevel *state) {
+bool compare_xdg_surface_toplevel_state(struct wlr_xdg_toplevel *toplevel) {
 	// is pending state different from current state?
```
ifreund, line 29: This comment is not accurate, this function checks if the scheduled state is different from the pending state. Would you mind correcting it since you're already touching this function?

ifreund (approved):
LGTM!

types/xdg_shell/wlr_xdg_toplevel.c
vyivel: Replaced with a comment about comparing the scheduled configure to the last sent one.

vyivel: Actually the comment was incorrect, as the scheduled configure was compared to the last *acked* one, but it doesn't matter as the whole (badly named btw) function is removed now.

ifreund (approved):
The latest iteration looks great to me, it simplifies the code fixing #2330 and the problems discussed in #3206. It also makes subtle, hard to debug, bugs such as the one fixed by #2342 impossible.

types/xdg_shell/wlr_xdg_toplevel.c
```diff
@@ -65,49 +82,49 @@ void send_xdg_toplevel_configure(struct wlr_xdg_surface *surface,
...
 		return;
 	}
-	*configure->toplevel_state = surface->toplevel->server_pending;
+	*configure->toplevel_configure = surface->toplevel->scheduled;
 
 	struct wl_array states;
 	wl_array_init(&states);
-	if (surface->toplevel->server_pending.maximized) {
+	if (surface->toplevel->scheduled.maximized) {
 		uint32_t *s = wl_array_add(&states, sizeof(uint32_t));
```
emersion, line 43: Unrelated note: seems like this code could be cleaner by filling a `uint32_t states[32]`, then creating a `wl_array` pointing to it. Something like:

```c
uint32_t states[32];
size_t states_len = 0;

if (scheduled.maximized) {
    states[states_len++] = XDG_TOPLEVEL_STATE_MAXIMIZED;
}
…

struct wl_array states_arr = {
    .data = states,
    .size = states_len * sizeof(uint32_t),
};
```

vyivel: Seems like a pretty unconventional way to create a wl_array, but sure, why not. Also avoids potential OOM errors.

emersion: Yeah… Ideally `wl_array` could be marked `const` so that wayland-util functions like `wl_array_add` can't be called, but [it's a bit too late](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/107). Here `wl_array` is just a read-only structure used by libwayland to send events.

emersion (approved):
This looks good to me! Thanks a bunch, and also thanks @ifreund for the review!

Issue #3200, opened 2021-09-18
libseat/backend/seatd: fatal error when switching VTs


Dudemanguy:
To reproduce this, you need to be running only pure `seatd` on your system with no systemd-logind or elogind daemon running at all. It will only happen if libseat uses its `seatd` backend and not `logind`. Launch any wlroots-based compositor (e.g. sway), switch VTs and it will crash. This behavior was actually caused by this libinput [commit](https://gitlab.freedesktop.org/libinput/libinput/-/commit/b7aac3c1482410a82e1fc091d643fb1b2d08ff6b) (it works normally if you revert it). I'm not really sure of the details, but I guess it messes with the FDs somehow which makes libseat unable to properly close it. libseat sends back an error to wlroots which then sends a fatal error over wayland display. Here's the relevant part of the sway debug log.

```
00:00:02.807 [INFO] [wlr] [libseat] [libseat/backend/seatd.c:222] Disabling seat
00:00:03.543 [ERROR] [wlr] [backend/session/session.c:326] Failed to close device 11: Bad file descriptor
00:00:03.633 [INFO] [wlr] [backend/drm/backend.c:127] DRM fd paused
Gdk-Message: 21:46:21.624: Error reading events from display: Broken pipe
(EE) failed to read Wayland events: Broken pipe
```
[Debug Log](https://0x0.st/-3ah.log)

Not sure if wlroots, seatd, or libinput needs fixing, but I'd figure I'd start by dropping this here.

vyivel:
Not sure about the cause, but can confirm the crash.

emersion:
The error isn't fatal I think. My guess would be a crash? Can you try to see if there's a coredump produced?

vyivel:
#0  0x00007ff552d54b97 in wl_list_remove () at /usr/lib/libwayland-server.so.0
#1  0x00007ff5534086f2 in wlr_session_close_file (session=0x558d6c812cb0, dev=0x558d6cf55fc0) at ../subprojects/wlroots/backend/session/session.c:329
#2  0x00007ff5533f450d in libinput_close_restricted (fd=38, _backend=0x558d6c813550) at ../subprojects/wlroots/backend/libinput/backend.c:39
#3  0x00007ff5532051db in  () at /usr/lib/libinput.so.10
#4  0x00007ff55320547e in  () at /usr/lib/libinput.so.10
#5  0x00007ff553205924 in  () at /usr/lib/libinput.so.10
[...]

PR #3203, opened 2021-09-18
backend/session: Run libseat callbacks on idle

kennylevinsen:
libseat callbacks can be triggered from the stack of a libseat call if
they were read while waiting for a return value. This is necessary as
dispatch will not be called when the file descriptor has already
been drained.

This leads to issues if we're called by libinput to close a device, and
then handle a disable event that leads to libinput closing all its
devices.

As workaround, break the stack on libseat events by executing them as
idle events.

Fixes: https://github.com/swaywm/wlroots/issues/3200

TODO:
- [ ] queued idle events could fire after wlr_session_destroy (use-after-free)

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -130,6 +130,8 @@ static struct wlr_session *session_create_and_wait(struct wl_display *disp) {
 		struct wl_event_loop *event_loop =
 			wl_display_get_event_loop(session->display);
 
+		// Session enable might already have been registered as idle callback
+		wl_event_loop_dispatch_idle(event_loop);
 		while (!session->active) {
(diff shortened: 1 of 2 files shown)


kennylevinsen:
An "easy" way to solve the wlr_session_destroy thing would be to just call wl_event_loop_dispatch_idle first-thing there, but not sure if we want to do that. On the other hand, the alternative would be to store the idle event sources on the wlr_session so we can destroy them, but seems like a lot of house-keeping for a corner-case...

kennylevinsen:
Superseded by https://git.sr.ht/~kennylevinsen/seatd/commit/2eee9aa445e3f9dc6a7ca115489f87b10f60b9ba.

Issue #3161, opened 2021-09-02
scene: add a way to render arbitrary textures


emersion:
Some compositors like Sway need to render arbitrary textures. This includes use-cases such as:

- Titlebar textures
- Saved surface buffers for atomic layout transactions

The scene-graph API doesn't yet allow to do this. Potential solutions include:

- Add a new `TEXTURE` node type, which takes a `wlr_texture`. Probably the most straightforward, but ties the node to a specific renderer and doesn't allow for direct scan-out.
- Add a new `BUFFER` node type, which takes a `wlr_buffer`. The scene-graph would be responsible for uploading/importing the buffer as a texture. However, this potentially duplicates the work with `wlr_surface`'s client buffers in the case of displaying a saved surface buffer.
- Add a new `CUSTOM` node type, which takes an arbitrary rendering function and the extents of the node. This would allow compositors to still use the scene-graph even when they do custom rendering (e.g. 3D effects). This doesn't allow direct scan-out and could go overboard.

PR #3207, opened 2021-09-20
scene: add wlr_scene_buffer

emersion:
This new scene-graph node displays a wlr_buffer.

Closes: https://github.com/swaywm/wlroots/issues/3161

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -266,6 +266,23 @@ struct wlr_scene_buffer *wlr_scene_buffer_create(struct wlr_scene_node *parent,
 	return scene_buffer;
 }
 
+void wlr_scene_buffer_set_source_box(struct wlr_scene_buffer *scene_buffer,
+		const struct wlr_fbox *box) {
+	struct wlr_fbox *cur = &scene_buffer->src_box;
+	if ((wlr_fbox_empty(box) && wlr_fbox_empty(cur)) ||
+			(box != NULL && memcmp(cur, box, sizeof(*box)) == 0)) {
+		return;
+	}
+
+	if (box != NULL) {
+		memcpy(cur, box, sizeof(*box));
+	} else {
+		memset(cur, 0, sizeof(*cur));
+	}
+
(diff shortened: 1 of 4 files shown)


emersion:
To make this usable for rendering saved buffers we also need to add a few more functions for scaling/cropping.

emersion:
Pushed two more commits to add scaling and cropping.

emersion:
One thing worth noting is that it should be possible to implement `wlr_scene_surface` on top of `wlr_scene_buffer`. Maybe we should do that at some point. Need to investigate how that would impact surface-specific integrations with the scene-graph, such as auto `wl_surface.enter`/`leave`.

types/scene/wlr_scene.c
```diff
@@ -220,6 +232,119 @@ void wlr_scene_rect_set_color(struct wlr_scene_rect *rect, const float color[sta
...
+		return client_buffer->texture;
+	}
+
+	if (scene_buffer->texture != NULL) {
+		return scene_buffer->texture;
+	}
+
+	scene_buffer->texture =
+		wlr_texture_from_buffer(renderer, scene_buffer->buffer);
+	if (scene_buffer->texture == NULL) {
```
vyivel, line 304: Is this check really necessary?

emersion: Yes -- the name might not be well-chosen but `wlr_texture_from_buffer` will import/upload the buffer. This can fail.

vyivel: Ah, sorry, that's not what I meant. Returning NULL if the value is NULL is redundant if said value is returned otherwise.

emersion: Oh, right, fixed!

emersion:
> I think a struct wlr_scene_node_impl would be a nice addition in the future to clear the big switches here and there.

This has been discussed here: https://github.com/swaywm/wlroots/pull/3140#issuecomment-907100708

tl;dr not sure it'd simplify things.

bl4ckb0ne:
Good point, it does make sense to keep the impl indeed.

Don't forget to rebase the patchset!

PR #3215, opened 2021-09-24
Decouple the backend and the renderer

bl4ckb0ne:
Removes the hard coupling between the backend and the renderer to improve third party backends/renderers

Related to #3195


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3215

PR #3218, opened 2021-09-25
Rename wlr_surface_get_extends() to wlr_surface_get_bounding_box()

vyivel:
Extend (verb): cause to cover a wider area; make larger.


tiosgz:
I think the original intention was `extents`.

(might make the change a little less confusing in case someone is used to the name; though i personally have nothing against `bounding_box`)

vyivel:
>I think the original intention was `extents`.

At first I thought "extents" was a bit wrong (and should have been just "extent" instead), but now I see that `pixman_region32_extents()` exists, so I'll just rename the function to `wlr_surface_get_extents()`.

emersion:
Similar to https://github.com/swaywm/wlroots/pull/2434. Not sure the breakage is worth it.

PR #3220, opened 2021-09-26
for_each_surface and surface_at: skip unmapped surfaces

tiosgz:
These functions are used mostly for rendering, where including unmapped surfaces is undesired.

This is a breaking change. However, few to no usages will have to be updated.

Tested in kiwmi (with https://github.com/buffet/kiwmi/commit/d69ccc7bdd6d2e6d2930bdbeadbc6c1d679e55a0 reverted) using [`tiosgz/wleird@subsubsurface:subsubsurface.c`](https://github.com/tiosgz/wleird/blob/subsubsurface/subsubsurface.c) (when the black surface hides, its subsurface \[the red one\] should also be hidden). Can also be tested in tinywl.

### DIFF
--- a/include/wlr/types/wlr_layer_shell_v1.h
+++ b/include/wlr/types/wlr_layer_shell_v1.h
@@ -143,7 +143,10 @@ bool wlr_surface_is_layer_surface(struct wlr_surface *surface);
 struct wlr_layer_surface_v1 *wlr_layer_surface_v1_from_wlr_surface(
 		struct wlr_surface *surface);
 
-/* Calls the iterator function for each sub-surface and popup of this surface */
+/**
+ * Calls the iterator function for each mapped sub-surface and popup of this
+ * surface (whether or not this surface is mapped).
+ */
 void wlr_layer_surface_v1_for_each_surface(struct wlr_layer_surface_v1 *surface,
 		wlr_surface_iterator_func_t iterator, void *user_data);
 
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -264,9 +264,9 @@ void wlr_surface_get_extends(struct wlr_surface *surface, struct wlr_box *box);
 struct wlr_surface *wlr_surface_from_resource(struct wl_resource *resource);
 
 /**
- * Call `iterator` on each surface in the surface tree, with the surface's
- * position relative to the root surface. The function is called from root to
- * leaves (in rendering order).
+ * Call `iterator` on each mapped surface in the surface tree (whether or not
+ * this surface is mapped), with the surface's position relative to the root
+ * surface. The function is called from root to leaves (in rendering order).
  */
 void wlr_surface_for_each_surface(struct wlr_surface *surface,
 	wlr_surface_iterator_func_t iterator, void *user_data);
--- a/include/wlr/types/wlr_surface.h
+++ b/include/wlr/types/wlr_surface.h
@@ -233,7 +233,8 @@ bool wlr_surface_point_accepts_input(struct wlr_surface *surface,
 		double sx, double sy);
 
 /**
- * Find a surface in this surface's tree that accepts input events at the given
+ * Find a surface in this surface's tree that accepts input events and has all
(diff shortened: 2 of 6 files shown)


vyivel:
I wonder if the name is too misleading. Also, some users might want to iterate over *all* surfaces (e.g. to create a custom object for each one). IMO, it would better to introduce `wlr_surface_for_each_mapped_surface()`.

tiosgz:
Let me remind you of something ;-)

> ```
> 2021-09-25 11:59 vyivel    when iterating over all (including unmapped) surfaces is required?
> ```

I mean, i’ve got no problem with either variant (and adding the `mapped` function is what i initially suggested), but first i’d like to make sure what the final one is going to be.

Since all `subsurfaces_below` are iterated before their parent, creating custom objects for them with `for_each_surface` only makes sense (is less complicated than a custom implementation) as long as the custom object isn’t aware of its parent. (E.g. [sway’s `view_init_subsurfaces`](https://github.com/swaywm/sway/blob/5fd5d6434e9ebbc453dde33bf58fea931ea2975a/sway/tree/view.c#L1029) walks the tree itself, also skipping the top-most surface.)

vyivel:
I'd check how compositors other than Sway use the function to make the better decision.

emersion:
Hm I'd also favor a `_mapped` variant, if only to make it clear what the function does. Internally we don't need to have two implementations, we can just add a `bool only_mapped` arg to a private function.

emersion:
We could even go for a rename to `_mapped` (removing the old function) if there isn't any use-case for iterating over all surfaces regardless of their mapped state.

ifreund:
As another data point, river also only uses for_each_surface for rendering. 

There are two other things we need to keep in mind here:
- `wlr_surface_surface_at()` needs to be kept in sync with `wlr_surface_for_each_surface()` and if a mapped-only for_each_surface variant is added we need a mapped-only surface_at() variant.
- If we add a mapped variant instead of changing `wlr_surface_for_each_surface()`, we need to add mapped variants for `wlr_xdg_surface_for_each_surface()`, `wlr_xdg_surface_for_each_popup_surface()`, `wlr_xdg_surface_surface_at()`, `wlr_xdg_surface_popup_surface_at()`, and the layer shell variants as well.

emersion:
> wlr_surface_surface_at() needs to be kept in sync with wlr_surface_for_each_surface() and if a mapped-only for_each_surface() variant is added we need a mapped-only surface_at() variant.

`wlr_surface_surface_at` is already a bit special in that it specifically checks for input event coords (see `wlr_surface_accepts_input`). Hm.

Regardless, if a surface isn't mapped, it's not visible, so it can't be intersected with a coord. So I don't think it's an issue.

> If we add a mapped variant instead of changing wlr_surface_for_each_surface(), we need to add mapped variants for wlr_xdg_surface_for_each_surface(), wlr_xdg_surface_for_each_popup_surface(), wlr_xdg_surface_surface_at(), wlr_xdg_surface_popup_surface_at(), and the layer shell variants as well.

Right.

Okay, I'm not against just changing the current function behavior I guess, especially given that `wlr_surface_surface_at` already does more than the name suggests...

tiosgz:
> > wlr_surface_surface_at() needs to be kept in sync with wlr_surface_for_each_surface() and if a mapped-only for_each_surface() variant is added we need a mapped-only surface_at() variant.
>
> Regardless, if a surface isn't mapped, it's not visible, so it can't be intersected with a coord. So I don't think it's an issue.

Right now i don’t see surface_at checking if anything is mapped, so unless i’ve missed anything, it can return a "mapped" subsurface of an unmapped surface. So it needs an update too.

tiosgz:
From my side this is ready to go (been wondering if this has to be explicitly stated in wlroots, or if there’s something else blocking this – either way, i just want to be sure).

The comments could possibly be improved, but it’s a fight between clarity & brevity.

ifreund (approved):

vyivel (approved):

emersion (approved):
LGTM

emersion:
Thanks!

PR #3222, opened 2021-09-28
backend: create renderer and allocator in wlr_backend_autocreate

emersion:
Instead of ensuring the renderer and allocator are initialized in each
backend, do it in wlr_backend_autocreate. This allows compositors to
create backends without any renderer/allocator if they side-step
wlr_backend_autocreate.

Since the wlr_backend_get_renderer and backend_get_allocator end up
calling wlr_renderer_autocreate and wlr_allocator_autocreate, it sounds
like a good idea to centralize all of the opimionated bits in one place.

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -209,6 +209,21 @@ static size_t parse_outputs_env(const char *name) {
 	return outputs;
 }
 
+static struct wlr_backend *ensure_backend_renderer_and_allocator(
+		struct wlr_backend *backend) {
+	struct wlr_renderer *renderer = wlr_backend_get_renderer(backend);
+	if (renderer == NULL) {
+		wlr_backend_destroy(backend);
+		return NULL;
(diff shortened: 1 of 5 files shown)


bl4ckb0ne:
The should be a way for users to use let wlroots choose the backend without creating the renderer and allocator.
In wxrc I call this function, then add the XR backend and create the renderer after. 

Maybe a `wlr_backend_create` function that will attempt all backend without ensuring a renderer?

emersion:
This PR doesn't intend to fix everything, it's just a small step in the right direction.

> Maybe a wlr_backend_create function that will attempt all backend without ensuring a renderer?

That would still be "auto" because it would have an hardcoded list of backends in it and it would pick the one it thinks is best.

bl4ckb0ne:
Alright, LGTM :)

PR #3226, opened 2021-09-29
Make allocator handle the swapchain managment

bl4ckb0ne:
Another try to decouple the backend/renderer/output/swapchain

Here the allocator is not made into a public interface required to be created by the user. It silently listens to the backend new output event to wire up the new `request_swapchain` event that the `wlr_outpu` fires when requesting a new swapchain.

Related to #3222 #3215 and #3197


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3226

PR #3231, opened 2021-10-01
Require INVALID for implicit format modifiers

emersion:
See this for motivation: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/75

- [ ] This breaks DRM cursors when modifiers aren't supported: the plane has only { LINEAR } supported and receives a buffer with INVALID.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3231

PR #3233, opened 2021-10-02
render/allocator: use render node if available in reopen_drm_node

emersion:
If we aren't trying to create a dumb buffer allocator, and if the
DRM device has a render node (ie, not a split render/display SoC),
then we can use the render node instead of the primary node. This
should allow wlroots to run under seatd when the current user
doesn't have the permission to open primary nodes (logind has a
quirk to allow physically logged in users to open primary nodes).

### DIFF
--- a/render/allocator/allocator.c
+++ b/render/allocator/allocator.c
@@ -25,11 +25,19 @@ void wlr_allocator_init(struct wlr_allocator *alloc,
  * TODO: don't assume we have the permission to just open the DRM node,
  * find another way to re-open it.
  */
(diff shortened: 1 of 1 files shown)


kennylevinsen (approved):
LGTM, although I wonder what we should do with respect to the dumb allocator. Just require that the primary node is freely accessible?

emersion:
The plan for this is https://github.com/swaywm/wlroots/pull/3158, but it requires an unreleased kernel version at the moment.

PR #3234, opened 2021-10-02
backend/drm: use drmCloseBufferHandle

emersion:
This has been added in [1] and allows us to close buffer handles
without manually calling drmIoctl.

[1]: https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/192


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3234

PR #3240, opened 2021-10-05
backend: unbreak build without libinput package

jbeich:
See [error log](https://github.com/swaywm/wlroots/files/7282201/wlroots-0.14.0.288.log) (`-Dbackends= -Drenderers= -Dxwayland=disabled`).


emersion (approved):
Thanks!

PR #3241, opened 2021-10-05
render: drop libEGL dependency if gles2 is disabled

jbeich:
Disabling `gles2` currently violates reproducibility due to opportunistic lookup of libEGL even when auto-detection is disabled.
```
$ meson setup --auto-features=disabled _build
$ readelf -d _buildlibwlroots.so | grep NEEDED.\*GL
 0x0000000000000001 NEEDED               Shared library: [libEGL.so.1]
$ nm --undefined-only -D _build/libwlroots.so | fgrep egl
                 U eglBindAPI
                 U eglCreateContext
                 U eglDestroyContext
                 U eglGetCurrentContext
                 U eglGetCurrentDisplay
                 U eglGetCurrentSurface
                 U eglGetError
                 U eglGetProcAddress
                 U eglInitialize
                 U eglMakeCurrent
                 U eglQueryContext
                 U eglQueryString
                 U eglReleaseThread
                 U eglTerminate
```


render/meson.build
```diff
@@ -14,13 +14,10 @@ wlr_files += files(
 	'wlr_texture.c',
 )
 
-egl = dependency('egl', required: 'gles2' in renderers)
-if egl.found()
+if 'gles2' in renderers or 'auto' in renderers
+	egl = dependency('egl')
```
emersion, line 18: Hm, this isn't quite right. If running with `auto`, we'll want to look for the `egl` dep, and disable the `gles2` renderer if it's missing. Here the `egl` dep is mandatory.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3241

PR #3243, opened 2021-10-06
Remove redundant NULL buffer checks

vyivel:
From what I understand, (confusingly named) `surface_apply_damage()` doesn't update `wlr_surface.buffer` if the upload has failed.

### DIFF
--- a/types/wlr_layer_shell_v1.c
+++ b/types/wlr_layer_shell_v1.c
@@ -339,15 +339,28 @@ static void layer_surface_role_commit(struct wlr_surface *wlr_surface) {
 		surface->mapped = true;
 		wlr_signal_emit_safe(&surface->events.map, surface);
 	}
-	if (surface->configured && !wlr_surface_has_buffer(surface->surface) &&
-			surface->mapped) {
-		layer_surface_unmap(surface);
+}
+
+static void layer_surface_role_precommit(struct wlr_surface *wlr_surface) {
+	struct wlr_layer_surface_v1 *surface =
+		wlr_layer_surface_v1_from_wlr_surface(wlr_surface);
+	if (surface == NULL) {
+		return;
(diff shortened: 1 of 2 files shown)


include/wlr/types/wlr_surface.h
```diff
@@ -91,8 +91,8 @@ struct wlr_surface {
 	/**
 	 * The surface's buffer, if any. A surface has an attached buffer when it
 	 * commits with a non-null buffer in its pending state. A surface will not
-	 * have a buffer if it has never committed one, has committed a null buffer,
-	 * or something went wrong with uploading the buffer.
```
emersion, line 95: If the first attached buffer failed during import/upload, this can still happen.

vyivel: Right. Although, if that were to happen, the surface would've been unmapped already, so there's no reason to have those checks anyway.
Should I change it to "something went wrong with uploading the *first* buffer"?

types/wlr_layer_shell_v1.c
```diff
@@ -339,10 +339,6 @@ static void layer_surface_role_commit(struct wlr_surface *wlr_surface) {
 		surface->mapped = true;
 		wlr_signal_emit_safe(&surface->events.map, surface);
 	}
-	if (surface->configured && !wlr_surface_has_buffer(surface->surface) &&
```
emersion, line 342: This is not done in the precommit handler for layer-shell.

vyivel: Hm. It should be, though.

vyivel: Added role precommit handler for `layer-surface`.

emersion (approved):
LGTM, thanks!

PR #3246, opened 2021-10-09
presentation-time: use a surface addon 

vyivel:
Preliminary work for https://github.com/swaywm/wlroots/pull/3151.

---

This PR introduces a surface-synced object which acts as an "owner" of presentation feedbacks, which allows switching to the usual `current`/`pending` pattern.

### Notes 

- While there is no way to destroy `wlr_presentation_surface` without destroying the surface itself, it shouldn't be a problem as a surface which uses presentation feedbacks is likely to use them for its entire lifetime.
- This has only been tested with `weston-presentation-shm`.

### DIFF
--- a/include/wlr/types/wlr_presentation_time.h
+++ b/include/wlr/types/wlr_presentation_time.h
@@ -14,12 +14,13 @@
 #include <time.h>
 #include <wayland-server-core.h>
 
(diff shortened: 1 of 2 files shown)


vyivel:
Rebased against master.

PR #3251, opened 2021-10-12
scene: add functions to place node on top/bottom

ifreund:
Currently these functions remove the node from the scene if the sibling
argument is the same node as the node. Making this case a noop is much
more predictable behavior.


ifreund:
Renamed the new functions to `wlr_scene_node_place_on_top` and `wlr_scene_node_place_on_bottom`. This should be ready to go now barring further feedback.

Issue #2257, opened 2020-06-05
Udev dependency


illiliti:
Is it possible to avoid udev dependency by using fixed(not autodetected) configuration for input/output devices ?

emersion:
What's your use-case? What are you trying to achieve?

illiliti:
I'm trying to use mdev[d] as my main device manager. In Xorg is it possible to avoid udev dependecy by using static xorg.conf configuration. Libinput can be also used without udev with this patch https://github.com/kisslinux/repo/blob/master/extra/libinput/patches/libinput-optional-udev.patch

illiliti:
> Write a library that exposes the udev interface but uses mdev under-the-hood

There is one - https://github.com/jcnelson/vdev/tree/master/libudev-compat , but it's a bit outdated and doesn't work

> Patch wlroots to use mdev instead of udev

mdev[d] does not expose any information about input/output by default. Instead it's relies on /etc/mdev.conf which is depends on user configuration

Would be nice to have static configuration to be independent of device managers at all.

myfreeweb:
[libudev-devd](https://github.com/FreeBSDDesktop/libudev-devd) basically enumerates devices by looking at directories, and only uses devd for hotplug. Should be easy to fork it to not do any hotplug at all and make it work on Linux.

illiliti:
it wasn't so easy as you said but i did it. i had to reimplement udev api from scratch because libudev-devd depends on bsd specific code. i don't know why wlroots decided to use this over-engineered library, but i can't blame choices. anyway i did it, now it's possible to use wlroots-based compositors with any device manager or without at all! i'm currently working on hotplugging support, so for now you need to restart compositor to hotplug devices. i still hope that wlroots provide ability to disable libudev(i did it through ripping out udev code and using WLR_DRM_DEVICES). also someone reported that mouse doesn't work with libudev-zero but i can't reproduce. suggestions about code, reports about which software works with it, feature requests and bug reports are highly wanted!

https://github.com/illiliti/libudev-zero

akvadrako:
This is interesting. I run sway in a user namespace (for an unprivileged chroot) and libudev doesn't work in that case because it only allows messages from uid = 0, but real root is mapped to nobody in the namespace. So I patched libudev to allow messages from any uid. This is still secure, but it may not get accepted upstream (see  https://github.com/systemd/systemd/pull/16307). So I'm interested in alternatives to libudev.

[...]

PR #3256, opened 2021-10-14
Make seat backend optional

DemiMarie:
It adds a dependency on libseat and not all compositors need it.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -18,7 +18,7 @@ packages:
   - xwayland
   - libseat-dev
 sources:
-  - https://github.com/swaywm/wlroots
+  - https://gitlab.freedesktop.org/wlroots/wlroots.git
 tasks:
   - setup: |
       cd wlroots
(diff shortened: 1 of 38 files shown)


DemiMarie:
@emersion It allows building a smaller binary, and reduces the total size of the dependency tree for those (such as Qubes OS) who will only be using software rendering and virtual inputs.

DemiMarie:
> `-Dsession=disabled` doesn't need libudev as well, see #2257. This may help NetBSD and OpenBSD to make an initial port of wlroots as nested-only + headless-only.

I can see a rootless nested backend being quite useful in practice.

DemiMarie:
> I don't really buy the "smaller binary" argument, libseat is a few KiBs on my system, probably even less with everything disabled.

That is still a bit less code, and for projects that need to review all of the code they use, it is less code to review.

emersion:
Yes, I haven't considered the udev dependency, and it's definitely a good point.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3256

PR #3261, opened 2021-10-17
Add error handling to backend creation

AnthonySuper:
This commit adds two error-handling cases to the function
attempt_dmr_backend. Specifically:

- In the case where the number of found GPUs is zero, we now
  print a log message indicating this and return a NULL pointer
- In the case where we could not successfully create a backend
  on any GPU, we now log a message indicating this and return
  a NULL pointer

This allows us to provide more descriptive error messages,
as well as avoid a SEGFAULT (the function
`ensure_primary_backend_renderer_and_allocator` dereferences the pointer
given, which could be NULL until this patch) when these cases arise.

### DIFF
--- a/backend/backend.c
+++ b/backend/backend.c
@@ -293,7 +293,12 @@ static struct wlr_backend *attempt_drm_backend(struct wl_display *display,
 		return NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


AnthonySuper:
@emersion styles should be fixed!

PR #3271, opened 2021-10-18
render/vulkan: Refactor image usages for modifiers

misyltoad:
Makes this much more extensible so we could add a storage type in future for compute.

Signed-off-by: Joshua Ashton <joshua@froggi.es>


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3271

PR #3279, opened 2021-10-20
scene: add support for direct scan-out

emersion:
Check if only a single node intersects with the output viewport
and all of its properties match. In this case, attempt direct
scan-out.


kennylevinsen:
Will https://github.com/swaywm/sway/issues/6149 also be a problem here?

emersion:
Hm, yeah, forgot about that. Done.

PR #3282, opened 2021-10-20
Add explicit synchronization, with timelines

emersion:
This PR implements explicit synchronization. Unlike previous attempts, it uses a timeline abstraction based on drm_syncobj like Vulkan.

To inter-operate with other APIs (KMS, EGL, linux-explicit-synchronization-v1), the timeline points can be converted from/to sync_file FDs.

Test with: `build/examples/explicit-sync -s weston-simple-dmabuf-egl`. Use `WAYLAND_DEBUG=server` to check that fences are exchanged. To test the DMA-BUF sync_file extraction, use `weston-simple-egl` as test client (requires [kernel patch](https://lore.kernel.org/dri-devel/20210610210925.642582-5-jason@jlekstrand.net/)).

TODO:

- [x] Write an example compositor
- [x] Extract sync_file from DMA-BUF when client doesn't support linux-explicit-synchronization-v1
- [ ] Add `wlr_texture` upload synchronization
- [x] Figure out if this API is okay for Vulkan
- [ ] Detect support in backends and renderers
- [ ] Add support for cached state to linux-explicit-synchronization-v1
- [ ] Documentation
- [ ] Sanity checks (e.g. in `wlr_output_commit`)
- [ ] Consider ref'counting `wlr_render_timeline`
- [x] ~~Output cursors~~ let's wait until output layers for this

Previous work:

- https://github.com/swaywm/wlroots/pull/2070
- https://github.com/swaywm/wlroots/pull/1685

Future work:

- Multi-GPU support in the DRM backend
- Wayland backend support

### DIFF
--- a/types/wlr_linux_explicit_synchronization_v1.c
+++ b/types/wlr_linux_explicit_synchronization_v1.c
@@ -327,13 +327,20 @@ bool wlr_linux_explicit_synchronization_v1_signal_surface_timeline(
 		uint64_t dst_point) {
 	struct wlr_linux_surface_synchronization_v1 *surface_sync =
 		surface_sync_from_surface(explicit_sync, surface);
-	if (!surface_sync) {
-		// TODO: fallback to DMA-BUF fence export
-		return false;
+	if (surface_sync) {
+		return wlr_render_timeline_import_sync_file(timeline, dst_point,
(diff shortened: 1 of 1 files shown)


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3282

PR #3286, opened 2021-10-21
linux-explicit-synchronization-v2: new protocol implementation

emersion:
See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/90

Depends on: https://github.com/swaywm/wlroots/pull/3282


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3286

PR #3292, opened 2021-10-22
Use the pending resolution when rendering

DemiMarie:
This is necessary for rootless compositors, where the output resolution
can be changed by a client at any time.  This requires wlr_scene to have
access to `output_pending_resolution`, which other renderers will need
as well.  Make that function a public API.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3292

PR #3298, opened 2021-10-25
scene: add wlr_scene_xdg_popup_create

emersion:
This allows compositors to easily add an xdg_popup to the
scene-graph while retaining the ability to unconstraint the popup
and decide its final position.

Compositors can handle new popups with the wlr_xdg_shell.new_surface
event, get the parent scene-graph node via wlr_xdg_popup.parent.data,
create a new scene-graph node via wlr_scene_xdg_popup_tree_create,
and unconstraint the popup if they want to.

### DIFF
--- a/include/wlr/types/wlr_scene.h
+++ b/include/wlr/types/wlr_scene.h
@@ -25,6 +25,7 @@
 
 struct wlr_output;
 struct wlr_output_layout;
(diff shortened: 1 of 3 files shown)


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3298

PR #3306, opened 2021-10-27
Make wlr_output_enable idempotent

DemiMarie:
Enabling or disabling an output should be idempotent.  Previously, this
was not the case, causing bugs.


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3306

PR #3311, opened 2021-10-28
output-layout: emit add before change event

zsugabubus:
It seems more logical to receive add first.


zsugabubus:
Not incorrect but an unconfigured/default one. I imagine `add` event is just about adding (hey, I have a new output and that's it), and `change` event is about reconfiguration.

emersion:
I think that's a bit confusing, and the existing behavior is less misleading.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3311

PR #3312, opened 2021-10-28
scene: fix scaling of negative coordinates

zsugabubus:
round() rounds away from zero that can cause unwanted lengthening of
dimensions when sides of box differ in their sign:

scale_length(2, -1, 1.5) == 4 (-1.5 -> -2; 1.5 -> 2)
scale_length(2,  1, 1.5) == 3

### DIFF
--- a/types/scene/wlr_scene.c
+++ b/types/scene/wlr_scene.c
@@ -364,14 +364,14 @@ static void scene_node_get_size(struct wlr_scene_node *node,
 }
 
 static int scale_length(int length, int offset, float scale) {
-	return round((offset + length) * scale) - round(offset * scale);
+	return (int)((offset + length) * scale + .5f) - (int)(offset * scale + .5f);
 }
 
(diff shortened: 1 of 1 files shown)


vyivel (approved):
The logic seems right to me.

emersion:
Hm, maybe we could have a `round_up` function to make this more explicit?

ascent12:
I haven't really looked into the issue at all, but would using the other rounding functions like `ceil()`/`floor()`/`trunc()`/`rint()` work?

zsugabubus:
Sorry, I have made a ~big~ huge mistake here with truncation (towards zero) and adding half (scale = 1, offset < 0 => bleh).

Thinking it over, I would go with `ceil()`. It has the advantage over `round()` that moving a line on the screen, does not change its size.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3312

Issue #3283, opened 2021-10-20
vulkan: Remove hard dependency on validation layer


nyorain:
We currently require the validation layer in the vulkan renderer. If that is not installed, creating the renderer fails. While the validation layer is useful at the moment we probably don't want to abort because it's not installed and (sooner or later) we should disable it by default as it has a significant performance impact.

See https://github.com/swaywm/wlroots/commit/8e346922508aa3eaccd6e12f2917f6574f349843#commitcomment-58239638

PR #3318, opened 2021-10-31
render/vulkan: get only available validation layers

bl4ckb0ne:
Fixes https://github.com/swaywm/wlroots/issues/3283

/cc @nyorain


render/vulkan/vulkan.c
```diff
@@ -33,6 +33,81 @@ static const char *find_extensions(const VkExtensionProperties *avail,
...
+		goto layers_err;
+	}
+
+	vkEnumerateInstanceLayerProperties(&count, layer_props);
+	for (size_t i = 0; i < (size_t)count; ++i) {
+		wlr_log(WLR_DEBUG, "Vulkan instance validation layer %s v%"PRIu32,
+			layer_props[i].layerName, layer_props[i].implementationVersion);
+		for (size_t j = 0; j < layers_len; ++j) {
+			if (strcmp(layer_props[i].layerName, layers[j]) == 0) {
+				found_layers[found_layers_count] = calloc(
```
nyorain, line 78: The alloc and strcpy isn't needed. You know that `layer_props[i].layerName` is the same as `layers[j]` and the string in `layers[j]` is a string literal so you can just store that here. Then you can also remove the free calls for the layer names below.

render/vulkan/vulkan.c
```diff
@@ -163,21 +238,21 @@ struct wlr_vk_instance *vulkan_instance_create(size_t ext_count,
...
+	}
 
 	VkInstanceCreateInfo instance_info = {0};
 	instance_info.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO;
 	instance_info.pApplicationInfo = &application_info;
 	instance_info.enabledExtensionCount = ini->extension_count;
 	instance_info.ppEnabledExtensionNames = ini->extensions;
 	instance_info.enabledLayerCount = layer_count;
-	instance_info.ppEnabledLayerNames = layers;
+	instance_info.ppEnabledLayerNames = (const char *const *)layers;
```
nyorain, line 255: get_validation_layers should already return that type. We don't ever want to modify the layer names and it's needed if you want to use the string literals as suggested above.

render/vulkan/vulkan.c
```diff
@@ -33,6 +33,81 @@ static const char *find_extensions(const VkExtensionProperties *avail,
...
+	static const size_t layers_len = sizeof(layers) / sizeof(layers[0]);
+
+	VkLayerProperties *layer_props = NULL;
+	char **found_layers = NULL;
+	uint32_t found_layers_count = 0;
+
+	uint32_t count;
+	vkEnumerateInstanceLayerProperties(&count, NULL);
+
+	if (count == 0) {
```
nyorain, line 53: Could also check that the return value from the `vkEnumerateInstanceLayerProperties` above is `VK_SUCCESS`.

nyorain:
This only addresses the layer availability part of #3283. When people really start using the Vulkan renderer, they probably want to run it without the validation layer and so we should offer an option to not enable it (even though I don't think it should be the default yet). Note that currently we consider the `debug` parameter passed to `vulkan_instance_create` for `layer_count`, your change ignores it and always forces the layers when they are available.
With the comments addressed, I'd still be in favor of merging this, we can think about explicitly disabling the layers later on.

emersion:
When the Vulkan renderer is deemed production-ready, we can remove all of this code: people can always enable it again with `VK_INSTANCE_LAYERS`.

bl4ckb0ne:
We could do two iterations of the validation layers, first one to enable them if present, then another one to enable them on demand. I tried to do a bit of both here but it's not very good. I'll rework the patch with your feedback @nyorain and I'll open a new issue for `WLR_VK_INSTANCE_LAYERS` to do later.

render/vulkan/vulkan.c
```diff
@@ -33,6 +33,73 @@ static const char *find_extensions(const VkExtensionProperties *avail,
 	return NULL;
 }
 
+static const char* const *get_instance_layers(uint32_t *layer_count) {
+	// TODO request layer via env var WLR_VK_INSTANCE_LAYER
```
emersion, line 37: The env var I mentioned doesn't need any action from wlroots. It's called `VK_INSTANCE_LAYERS` and is handled by the Vulkan loader.

bl4ckb0ne: Oh that's nice, I wasn't aware of that. I'll remove the comment.

emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3318

PR #3319, opened 2021-10-31
scene/subsurface_tree: hide unmapped subsurfaces

tiosgz:
This approach leaves two unused listeners on each root surface, but i don’t know of a better solution.


emersion (approved):
LGTM, thanks!

PR #3320, opened 2021-10-31
zext-screencopy-unstable-v1

any1:
This is an implementation of a new screencopy protocol which integrates most of the requirements which we have discussed in the past. It should also be possible to extend it to capture top-levels.

The new protocol is modelled after `wl_surface`. This model should make it easier to extend the protocol in the future.

wayland-protocols PR: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/124
Supersedes: #2329, #3166, https://github.com/swaywm/wlr-protocols/pull/105

cc @emersion, @columbarius, @quantum5


emersion:

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3320

